PR #5940, opened 2020-04-03
Added Safari version for ResizeObserver

jashsayani:
Source: https://webkit.org/blog/10247/new-webkit-features-in-safari-13-1/

### DIFF
--- a/api/ResizeObserver.json
+++ b/api/ResizeObserver.json
@@ -29,7 +29,7 @@
             "version_added": "47"
           },
           "safari": {
-            "version_added": false
+            "version_added": "13.1"
           },
           "safari_ios": {
             "version_added": false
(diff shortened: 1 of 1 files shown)


api/ResizeObserver.json
```diff
@@ -29,7 +29,7 @@
             "version_added": "47"
           },
           "safari": {
-            "version_added": false
+            "version_added": "13.1"
           },
           "safari_ios": {
             "version_added": false
```
ddbeck, line 35: It looks like these changes have also shipped to Safari for iOS. Would you mind setting the iOS values to `"13.4"`, as well? Like this:

```suggestion
            "version_added": "13.4"
```

jashsayani: Done.

PR #5942, opened 2020-04-04
IE doesn't support changes to animation-duration

vinyldarkscratch:
Apparently Internet Explorer doesn't support changes to `animation-duration`.  This PR adds a note to indicate as such.

Part of work for #5933.

### DIFF
--- a/css/properties/animation-duration.json
+++ b/css/properties/animation-duration.json
@@ -82,7 +82,8 @@
             ],
             "ie": {
               "version_added": "10",
-              "notes": "Internet Explorer does not respond to changes to this property, whether programmatic or via developer tools."
+              "partial_implementation": true,
+              "notes": "Once the element has loaded, changing the value of this property has no effect."
             },
             "opera": [
               {
--- a/css/properties/animation-duration.json
+++ b/css/properties/animation-duration.json
@@ -81,7 +81,8 @@
               }
             ],
             "ie": {
-              "version_added": "10"
+              "version_added": "10",
+              "notes": "Internet Explorer does not respond to changes to this property, whether programmatic or via developer tools."
             },
             "opera": [
               {


css/properties/animation-duration.json
```diff
@@ -81,7 +81,8 @@
               }
             ],
             "ie": {
-              "version_added": "10"
+              "version_added": "10",
+              "notes": "Internet Explorer does not respond to changes to this property, whether programmatic or via developer tools."
```
ddbeck, line 85: This is a tough one. The way this is written is a little ambiguous. I originally read it as, IE recognizes this property but doesn't do anything with it, so animations don't work. But searching about this, it's a bit [more subtle than that](https://stackoverflow.com/questions/32380632/changing-rotation-speed-of-a-div-not-working-in-ie-11-using-js-to-update-css-an). Plus, I like to avoid mentioning dev tools, since they're not part of any standard behavior.

What do you think of this rewrite?

```suggestion
              "notes": "The value of this property cannot be changed after an animation has started."
```

ddbeck: And on additional thought, another attempt:

```suggestion
              "notes": "If an animation has already started, then changing the value of this property has no effect."
```

Because I guess there's nothing stopping you from changing it, it's just that nothing happens.

vinyldarkscratch: One of the things I found is that it doesn’t matter if the animation has started or not — it’s when the element has loaded, the animation duration’s locked in.

Maybe: “Once the element has loaded, changing the value of this property has no effect.”?

ddbeck: Ugh, that's worse then right? If the CSS loads late, then the animation never goes off? Or you can't animate an element after the fact (e.g., applying a class to an element)?

Do we need to mark this as `"partial_implementation": true`?

vinyldarkscratch: I think the nature of this issue would qualify as partial implementation, yeah...

ddbeck: OK, one last question about this: is this behavior the same between IE 10 and 11? Or does the note apply to both?

vinyldarkscratch: The behavior’s the same for both I’ve found.

ddbeck: Great, thank you!

PR #5949, opened 2020-04-06
Add row for CSS content property alt-text

vinyldarkscratch:
Part of work for #5933.  It was determined that Firefox doesn't have support for alt text in the `content` property, so this PR adds a new feature for alt text.  All data was determined via manual testing.

### DIFF
--- a/css/properties/content.json
+++ b/css/properties/content.json
@@ -48,6 +48,54 @@
             "deprecated": false
           }
         },
+        "alt_text": {
+          "__compat": {
+            "description": "Alt text",
+            "support": {
+              "chrome": {
+                "version_added": "1"
+              },
+              "chrome_android": {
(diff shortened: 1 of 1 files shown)


css/properties/content.json
```diff
@@ -48,6 +48,54 @@
             "deprecated": false
           }
         },
+        "alt_text": {
+          "__compat": {
+            "description": "Alt text",
```
ddbeck, line 53: The way this works is so cryptic. Maybe we can relate this to the specification a little more?

```suggestion
            "description": "Alternative text after <code>/</code>",
```

vinyldarkscratch:
Not a problem, description's updated!

> How? It's not clear to me where this data comes from

I actually do mention at the end of my initial comment: "All data was determined via manual testing."

PR #5952, opened 2020-04-06
Chromium has no unprefixed support for text-emphasis-*

vinyldarkscratch:
After some research, it appears that Chrome oddly has no support for the `text-emphasis` properties without a prefix, yet Safari does.  This PR updates our data to indicate as such.

Part of work for #5933.

### DIFF
--- a/css/properties/text-emphasis-color.json
+++ b/css/properties/text-emphasis-color.json
@@ -53,13 +53,13 @@
               "version_added": false
             },
             "opera": {
-                "prefix": "-webkit-",
-                "version_added": "15"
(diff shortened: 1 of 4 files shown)


ddbeck:
> After some research

Any links for this? I don't doubt it, but a record is nice to have, should we ever revisit this data.

vinyldarkscratch:
I found this out through manual testing (forgot to specify as such, sorry), so I didn't have any links initially, but a quick search on CanIUse got me to this page: https://caniuse.com/#feat=text-emphasis

PR #5953, opened 2020-04-06
Update Chromium support for min-intrinsic

vinyldarkscratch:
Chrome dropped support for `min-intrinsic` in M48.  This PR updates our data accordingly.  (Determined via manual testing, as well as https://www.chromestatus.com/feature/5758434351775744.)

Furthermore, it appears there was a sight issue in our data.  `min-intrinsic` was labeled as supported since Chrome 22, however this conflicts with the fact that `intrinsic` was supported since Chrome 1 (see #5962), and from manual testing that reveals support for `intrinsic` in Chrome 1.

Part of work for #5933.

### DIFF
--- a/css/properties/width.json
+++ b/css/properties/width.json
@@ -329,7 +329,7 @@
                 },
                 {
                   "alternative_name": "min-intrinsic",
-                  "version_added": "22",
+                  "version_added": "25",
                   "version_removed": "48"
                 }
               ],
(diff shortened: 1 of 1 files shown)


ddbeck:
[This Chrome status entry](https://www.chromestatus.com/feature/5758434351775744) suggests both the `intrinsic` and `min-intrinsic` keywords were removed. Would it be too much to mark all of them as removed in this PR?

vinyldarkscratch:
Huh, it seems that we don't have an entry for `intrinsic` for Chrome, only a prefixed version, though Safari mentions `intrinsic`...  I think it makes more sense to update the data for `intrinsic` in a separate PR due to the complexity, so I'm going to hop right on it!

PR #5961, opened 2020-04-07
Make data for width: fit-content consistent

vinyldarkscratch:
This PR makes the data for the `fit-content` feature consistent across Chromium browsers.  `fit-content` is supported both with and without a prefix in Chrome, however this wasn't reflected in Android Chromium browsers (this was confirmed in manual testing).  Furthermore, Samsung Internet was labeled as supported under `-webkit-fill-available`, which is inaccurate (that's for `stretch`).

### DIFF
--- a/css/properties/width.json
+++ b/css/properties/width.json
@@ -157,9 +157,15 @@
                   "version_added": "22"
                 }
               ],
-              "chrome_android": {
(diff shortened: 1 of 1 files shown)


vinyldarkscratch:
Superseded by #5962.

PR #5967, opened 2020-04-09
Safari macOS 13.1/iOS 13.4 supports optional chaining

Wevah:
It was enabled in JavaScriptCore by default in November, but probably only became available in public Safari with the latest updates.

Also verified in the JS console.


Wevah:
Looks like it's not passing 100% of the tests in Test262; not sure if that's a requirement.

PR #5976, opened 2020-04-13
Update DataTransfer.json

Yaffle:
A small details about Opera 12

### DIFF
--- a/api/DataTransfer.json
+++ b/api/DataTransfer.json
@@ -946,7 +946,8 @@
               ]
             },
             "opera": {
-              "version_added": true
+              "version_added": true,
+              "notes": "As of Opera 12, <code>Text</code> is returned instead of <code>text/plain</code>"
             },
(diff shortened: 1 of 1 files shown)


ddbeck:
Thanks for this PR, @Yaffle. Can you tell me more about how you came to know this? Do you have a source for this note?

Yaffle:
@ddbeck , manual testing in Opera 12

ddbeck:
Thanks for the update, @Yaffle. One more question, then: did you try it in a version before Opera 12. In other words, do you know that Opera 12 was the first version to exhibit this behavior or is it possible it did so before that?

Yaffle:
@ddbeck , DataTransfer is not supported at all before this version according to https://caniuse.com/#feat=dragndrop

Yaffle:
@ddbeck, thanks for merging

PR #5989, opened 2020-04-16
Set correct compatibility for storage event in safari

koddsson:
Set `StorageEvent() constructor` compatibility to `true` in Safari. I tried looking at test in the [`Web Platform Tests`](wpt.fyi) project but couldn't find any appropriate to link here.

I just tested this in my local browser like so:

```js
> new StorageEvent()
< TypeError: Not enough arguments
> new StorageEvent('storage')
< StorageEvent {isTrusted: false, key: null, oldValue: null, newValue: null, url: "", …}
```

### DIFF
--- a/api/StorageEvent.json
+++ b/api/StorageEvent.json
@@ -77,10 +77,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": null
+              "version_added": true
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": true
             },
             "samsunginternet_android": {
               "version_added": true
--- a/api/StorageEvent.json
+++ b/api/StorageEvent.json
@@ -77,10 +77,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "6"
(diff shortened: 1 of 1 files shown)


api/StorageEvent.json
```diff
@@ -77,10 +77,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": null
+              "version_added": true
```
ddbeck, line 80:
```suggestion
              "version_added": "6"
```

api/StorageEvent.json
```diff
@@ -77,10 +77,10 @@
...
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": true
```
ddbeck, line 83:
```suggestion
              "version_added": "6"
```

koddsson:
>I did a little digging and it looks like this shipped in Safari 6 and iOS 6: https://trac.webkit.org/changeset/101591/webkit.

That's great! Thanks @ddbeck, I made those changes ✨

PR #5990, opened 2020-04-16
feat(html): Add global enterkeyhint attribute

eps1lon:
## Summarize your changes

Add global enterkeyhint attribute

## Data
- [`enterkeyhint` in the WHATWG spec](https://html.spec.whatwg.org/multipage/interaction.html#input-modalities:-the-enterkeyhint-attribute)
- [`enterkeyhint` chrome platform status](https://www.chromestatus.com/feature/5654529808793600)
- https://codepen.io/eps1lon/pen/bGVEZep?editors=1111 if it logs `undefined` then the attribute isn't supported yet

## Link to related issues or pull requests
https://github.com/mdn/sprints/issues/3109

### DIFF
--- a/html/global_attributes.json
+++ b/html/global_attributes.json
@@ -877,7 +877,7 @@
               "version_added": "13.1"
             },
             "safari_ios": {
-              "version_added": "13.1"
+              "version_added": "13.4"
             },
             "samsunginternet_android": {
               "version_added": false
--- a/html/global_attributes.json
+++ b/html/global_attributes.json
@@ -874,10 +874,10 @@
               "version_added": false
             },
             "safari": {
-              "version_added": false
+              "version_added": "13.1"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "13.1"
             },
             "samsunginternet_android": {
               "version_added": false
--- a/html/global_attributes.json
+++ b/html/global_attributes.json
@@ -846,6 +846,54 @@
           }
         }
       },
+      "enterkeyhint": {
+        "__compat": {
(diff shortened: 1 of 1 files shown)


html/global_attributes.json
```diff
@@ -846,6 +846,54 @@
           }
         }
       },
+      "enterkeyhint": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/HTML/Global_attributes/enterkeyhint",
```
sideshowbarker, line 851:
```suggestion
```
IMHO if there’s not yet any MDN article for a feature, it’s better to omit the `mdn_url`

html/global_attributes.json
```diff
@@ -846,6 +846,53 @@
...
+            },
+            "opera_android": {
+              "version_added": false
+            },
+            "safari": {
+              "version_added": false
+            },
+            "safari_ios": {
+              "version_added": false
+            },
```
orta, line 881: https://developer.apple.com/documentation/safari_release_notes/safari_13_1_beta_release_notes

This is in safari 👍

eps1lon: Whoops. I thought I added it to all versions.

html/global_attributes.json
```diff
@@ -846,6 +846,53 @@
...
+            "opera_android": {
+              "version_added": false
+            },
+            "safari": {
+              "version_added": "13.1"
+            },
+            "safari_ios": {
+              "version_added": "13.1"
```
vinyldarkscratch, line 880: Safari iOS versions are based upon iOS versions rather than Safari versions.  Can we fix that?

```suggestion
              "version_added": "13.4"
```

PR #5991, opened 2020-04-16
No WorkerNavigator.permissons in Firefox yet

asutherland:
https://bugzilla.mozilla.org/show_bug.cgi?id=1193373 tracks implementing this.  This can be checked by browsing to https://worker-playground.glitch.me/ and pasting `navigator.permissions` into the top box.  (Also, we don't define the `partial interface WorkerNavigator` in https://searchfox.org/mozilla-central/source/dom/webidl/WorkerNavigator.webidl.)

PR #5992, opened 2020-04-16
No WorkerNavigator.permissons in Firefox yet

asutherland:
https://bugzilla.mozilla.org/show_bug.cgi?id=1193373 tracks implementing this.  This can be checked by browsing to https://worker-playground.glitch.me/ and pasting `navigator.permissions` into the top box.  (Also, we don't define the `partial interface WorkerNavigator` in https://searchfox.org/mozilla-central/source/dom/webidl/WorkerNavigator.webidl.)

### DIFF
--- a/api/WorkerNavigator.json
+++ b/api/WorkerNavigator.json
@@ -109,10 +109,10 @@
               "version_added": "≤79"
             },
             "firefox": {
-              "version_added": null
+              "version_added": false
             },
             "firefox_android": {
-              "version_added": null
+              "version_added": false
             },
             "ie": {
               "version_added": null
--- a/api/WorkerNavigator.json
+++ b/api/WorkerNavigator.json
@@ -109,10 +109,10 @@
               "version_added": "≤79"
             },
             "firefox": {
-              "version_added": "46"
+              "version_added": null
             },
             "firefox_android": {
-              "version_added": "46"
+              "version_added": null
             },
             "ie": {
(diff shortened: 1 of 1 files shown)


asutherland:
https://github.com/mdn/browser-compat-data/pull/5991 was a variant of this that wasn't valid JSON.

api/WorkerNavigator.json
```diff
@@ -109,10 +109,10 @@
               "version_added": "≤79"
             },
             "firefox": {
-              "version_added": "46"
+              "version_added": null
```
sideshowbarker, line 112:
```suggestion
              "version_added": false
```
https://github.com/mdn/browser-compat-data/blob/master/schemas/compat-data-schema.md#version_added says that `null` means *“Support unknown”* while `false` means *“No support”*. So this should be `false`, right?

api/WorkerNavigator.json
```diff
@@ -109,10 +109,10 @@
...
             },
             "firefox_android": {
-              "version_added": "46"
+              "version_added": null
```
sideshowbarker, line 115:
```suggestion
              "version_added": false
```
Ditto

asutherland:
Yes, thank you for both the corrections and the documentation link!

pouyajabbarisani:
navigator.permissions is still undefined for me in service worker on firefox...

PR #5994, opened 2020-04-16
Add Network Error Logging compatibility data

lol768:
**Spec**: https://www.w3.org/TR/network-error-logging/
**Chrome version source**: https://www.chromestatus.com/feature/5391249376804864
**Edge version source**: x-ref https://en.wikipedia.org/wiki/Microsoft_Edge#Release_history to find first Blink-based build > Chrome 69

**Documentation**: https://wiki.developer.mozilla.org/en-US/docs/Web/HTTP/Network_Error_Logging

> Network Error Logging is a mechanism that can be configured via the NEL HTTP response header. This experimental header allows web sites and applications to opt-in to receive reports about failed (and, if desired, successful) network fetches from supporting browsers.

### DIFF
--- a/http/headers/nel.json
+++ b/http/headers/nel.json
@@ -1,0 +1,54 @@
+{
+  "http": {
+    "headers": {
+      "NEL": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/HTTP/Headers/NEL",
+          "support": {
+            "chrome": {
+              "version_added": "69"
+            },
+            "chrome_android": {
+              "version_added": "69"
+            },
+            "edge": {
+              "version_added": "79"
+            },
+            "firefox": {
+              "version_added": false
+            },
+            "firefox_android": {
+              "version_added": false
+            },
+            "ie": {
+              "version_added": false
+            },
+            "opera": {
+              "version_added": false
+            },
+            "opera_android": {
+              "version_added": false
+            },
+            "safari": {
+              "version_added": false
+            },
+            "safari_ios": {
+              "version_added": false
+            },
+            "samsunginternet_android": {
+              "version_added": false
+            },
+            "webview_android": {
+              "version_added": "69"
+            }
+          },
+          "status": {
+            "experimental": true,
+            "standard_track": true,
+            "deprecated": false
+          }
+        }
+      }
+    }
+  }
+}
--- a/http/headers/nel.json
+++ b/http/headers/nel.json
@@ -5,11 +5,23 @@
         "__compat": {
           "mdn_url": "https://developer.mozilla.org/docs/Web/HTTP/Headers/NEL",
           "support": {
-            "chrome": {
-              "version_added": "69"
-            },
+            "chrome": [
+              {
+                "version_added": "71"
+              },
+              {
+                "version_added": "69",
+                "flags": [
+                  {
+                    "type": "runtime_flag",
+                    "name": "Reporting",
+                    "value_to_set": "true"
+                  }
+                ]
+              }
+            ],
             "chrome_android": {
-              "version_added": "69"
+              "version_added": "71"
             },
             "edge": {
               "version_added": "79"
@@ -39,7 +51,7 @@
               "version_added": false
             },
             "webview_android": {
-              "version_added": "69"
+              "version_added": "71"
             }
           },
           "status": {
--- a/http/headers/nel.json
+++ b/http/headers/nel.json
@@ -36,10 +36,10 @@
               "version_added": false
             },
             "opera": {
-              "version_added": false
+              "version_added": "58"
             },
             "opera_android": {
-              "version_added": false
+              "version_added": "50"
             },
             "safari": {
               "version_added": false
@@ -48,7 +48,7 @@
               "version_added": false
             },
             "samsunginternet_android": {
-              "version_added": false
+              "version_added": "10.2"
             },
             "webview_android": {
               "version_added": "71"


lol768:
It looks like Chrome 69 shipped the reporting API behind a CLI flag. Unclear to me when it actually became available

sideshowbarker:
> It looks like Chrome 69 shipped the reporting API behind a CLI flag. Unclear to me when it actually became available...
> 
> NEL seems to pre-date the reporting API though https://www.w3.org/TR/2015/WD-network-error-logging-20150305/
> 
> In which case maybe the version should be even earlier?

See https://github.com/mdn/browser-compat-data/blob/master/schemas/compat-data-schema.md#flags.

If it no longer requires a flag, then you want to make the value for the `"chrome"` key an array with two objects: the first object is just `{ "version_added": "NN" }`, where `NN` is whatever version it shipped in without the need for a runtime flag — and the second object is `{"version_added": "MM" ,"flags": [{"type": "runtime_flag","name": "browser.flag.name","value_to_set": "true"}]}`, where `MM` is whatever version it first shipped in, when the runtime flag was required.

But if it still requires a flag, then rather than an array, you can just make the `"chrome"` value `{"version_added": "MM" ,"flags": [{"type": "runtime_flag","name": "browser.flag.name","value_to_set": "true"}]}` object.

lol768:
> Shouldn’t the filename here be NEL.json? (rather than network-error-logging.json)

Yes, this is fixed.

>See https://github.com/mdn/browser-compat-data/blob/master/schemas/compat-data-schema.md#flags.

Sure, but to be clear I don't know if the reporting API requiring a flag implies (at the time of Chrome 69) that NEL did too. In particular, there's a version of the spec that pre-dates the reporting API entirely and [Google's own paper](https://www.usenix.org/system/files/nsdi20spring_burnett_prepub.pdf) suggests that NEL has existed in some form or another since 2014 which would be much earlier than Chrome 69.

lol768:
Okay - the information on https://www.chromestatus.com/feature/5391249376804864 appears to be complete nonsense (and not for the first time..). In my testing w/ Browserstack only versions of Chrome >= 71 worked out of the box. This is despite the "Enabled by default" version reading 69 on that page.

I _think_ this is the relevant commit which makes it work by default:

```
commit 3dc6bcb5c5fff2cd697f2d4aa0983cd96b07160d
Author: Douglas Creager <dcreager@chromium.org>
Date:   Mon Sep 17 13:20:06 2018 +0000

    Enable Reporting and NEL by default
    
    We received approval to launch Reporting and Network Error Logging to
    100% on Stable.
    
    Bug: 676024, 799253
    Cq-Include-Trybots: luci.chromium.try:linux_mojo
    Change-Id: I2e502ffc340ae3d3a9c72c30791c7b71bbf8ab70
    Reviewed-on: https://chromium-review.googlesource.com/1224657
    Reviewed-by: Matt Menke <mmenke@chromium.org>
    Commit-Queue: Douglas Creager <dcreager@chromium.org>
    Cr-Commit-Position: refs/heads/master@{#591655}
```

And `git describe` gives us an answer which backs up my testing:

```
$ git describe --contains 3dc6bcb5c5fff2cd697f2d4aa0983cd96b07160d
71.0.3555.0~207
```

But the commit message references two bugs which I can't read so ¯\_(ツ)_/¯:

Screen Shot 2020-04-20 at 00 53 10

I've updated the PR nonetheless to make it clear that this didn't ship in 69.

http/headers/nel.json
```diff
@@ -0,0 +1,66 @@
...
+              "version_added": false
+            },
+            "firefox_android": {
+              "version_added": false
+            },
+            "ie": {
+              "version_added": false
+            },
+            "opera": {
+              "version_added": false
```
sideshowbarker, line 39:
```suggestion
              "version_added": "58"
```
…per the following:

https://github.com/mdn/browser-compat-data/blob/ee68a6df245329b16136f89fa69cd94627f9dc9e/browsers/opera.json#L475-L480

http/headers/nel.json
```diff
@@ -0,0 +1,66 @@
...
+            },
+            "opera_android": {
+              "version_added": false
```
sideshowbarker, line 42:
```suggestion
              "version_added": "50"
```
…per the following:

https://github.com/mdn/browser-compat-data/blob/ee68a6df245329b16136f89fa69cd94627f9dc9e/browsers/opera_android.json#L255-L260

http/headers/nel.json
```diff
@@ -0,0 +1,66 @@
...
+            },
+            "safari": {
+              "version_added": false
+            },
+            "safari_ios": {
+              "version_added": false
+            },
+            "samsunginternet_android": {
+              "version_added": false
```
sideshowbarker, line 51:
```suggestion
              "version_added": "10.2"
```
…per the following:

https://github.com/mdn/browser-compat-data/blob/ee68a6df245329b16136f89fa69cd94627f9dc9e/browsers/samsunginternet_android.json#L150-L154

sideshowbarker:
Kudos on the research

> I _think_ this is the relevant commit which makes it work by default:
> 
> ```
> commit 3dc6bcb5c5fff2cd697f2d4aa0983cd96b07160d
> Author: Douglas Creager <dcreager@chromium.org>
> Date:   Mon Sep 17 13:20:06 2018 +0000
> 
>     Enable Reporting and NEL by default

Looks right to me too.

> And `git describe` gives us an answer which backs up my testing:
> 
> ```
> $ git describe --contains 3dc6bcb5c5fff2cd697f2d4aa0983cd96b07160d
> 71.0.3555.0~207
> ```

Yes — and that can be confirmed online too, using the same hash and form at https://storage.googleapis.com/chromium-find-releases-static/index.html

https://storage.googleapis.com/chromium-find-releases-static/3dc.html#3dc6bcb5c5fff2cd697f2d4aa0983cd96b07160d

> But the commit message references two bugs which I can't read so ¯\_(ツ)_/¯:

Yeah, me neither…

lol768:
>Yes — and that can be confirmed online too, using the same hash and form at https://storage.googleapis.com/chromium-find-releases-static/index.html

Ah great, wish I'd known about this _before_ downloading the 20GB or so from Git :laughing: 

Thanks a bunch for sorting out the other Blink-based browsers - have applied those changes. I forget that Opera isn't Presto anymore..

sideshowbarker:
> > Yes — and that can be confirmed online too, using the same hash and form at https://storage.googleapis.com/chromium-find-releases-static/index.html
> 
> Ah great, wish I'd known about this _before_ downloading the 20GB or so from Git 😆

heh :) Yeah that form is pretty handy.

On the plus side, for your BCD contributions going forward, you now have Chrome source checkout that you can search through the logs of in the future ;)

> Thanks a bunch for sorting out the other Blink-based browsers - have applied those changes. I forget that Opera isn't Presto anymore..

Thanks for doing the research and testing

PR #5995, opened 2020-04-17
Update Safari support for TypedArray.copyWithin

lorenzosim:
Added in https://bugs.webkit.org/show_bug.cgi?id=148035 which is part of Safari 10.

### DIFF
--- a/javascript/builtins/TypedArray.json
+++ b/javascript/builtins/TypedArray.json
@@ -344,10 +344,10 @@
                 "version_added": false
               },
               "safari": {
-                "version_added": false
+                "version_added": "10"
               },
               "safari_ios": {
-                "version_added": false
+                "version_added": "10"
               },
               "samsunginternet_android": {
                 "version_added": false
--- a/javascript/builtins/TypedArray.json
+++ b/javascript/builtins/TypedArray.json
@@ -344,7 +344,7 @@
                 "version_added": false
               },
               "safari": {
-                "version_added": "10"
+                "version_added": "9.1"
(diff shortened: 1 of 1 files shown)


javascript/builtins/TypedArray.json
```diff
@@ -344,10 +344,10 @@
                 "version_added": false
               },
               "safari": {
-                "version_added": false
+                "version_added": "10"
```
sideshowbarker, line 347:
```suggestion
                "version_added": "9.1"
```
Per https://trac.webkit.org/log/webkit/tags/Safari-601.2.7.1 and the following: https://github.com/mdn/browser-compat-data/blob/17a48dcd70cc8c804ffc53d904441417a01263d5/browsers/safari.json#L107-L112

…it looks like this might have first gone into Safari 9.1, before Safari 10

javascript/builtins/TypedArray.json
```diff
@@ -344,10 +344,10 @@
...
               },
               "safari_ios": {
-                "version_added": false
+                "version_added": "10"
```
sideshowbarker, line 350:
```suggestion
                "version_added": "9.3"
```
…per https://trac.webkit.org/log/webkit/tags/Safari-601.2.7.1 and the following:

https://github.com/mdn/browser-compat-data/blob/17a48dcd70cc8c804ffc53d904441417a01263d5/browsers/safari_ios.json#L96-L100

lorenzosim:
Thanks!

PR #6004, opened 2020-04-18
Fixed #5649 - set CSS logical properties shorthands as flag enabled in Chromium

### LINKED ISSUE #5649: No support for margin-inline|block and padding-inline|block in Chrome
Testing with Chrome 79 (stable):

Property | Status | Shown on MDN as
--- | --- | ---
margin-inline | **Not supported** | Supported (69)
margin-inline-start | Supported | Supported (69)
margin-inline-end | Supported | Supported (69)
margin-block | **Not supported** | Supported (69)
margin-block-start | Supported | Supported (69)
margin-block-end | Supported | Supported (69)
padding-inline | **Not supported** | Supported (69)
padding-inline-start | Supported | Supported (69)
padding-inline-end | Supported | Supported (69)
padding-block | **Not supported** | Supported (69)
padding-block-start | Supported | Supported (69)
padding-block-end | Supported | Supported (69)

Chrome doesn't seem to support the `-inline` and `-block` shorthands.

phistuck:
Fixed https://github.com/mdn/browser-compat-data/issues/5649.
`margin-block`, `padding-block`, `margin-inline` and `padding-inline` are not enabled by default yet.
Sources -
https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/core/css/css_properties.json5;l=5276?q=padding-block%20-file:test%20file:json
https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/platform/runtime_enabled_features.json5;l=445?q=CSSLogical%20file:json5

### DIFF
--- a/css/properties/margin-block.json
+++ b/css/properties/margin-block.json
@@ -9,8 +9,8 @@
               "version_added": "69",
               "flags": [
                 {
-                "type": "runtime_flag",
-                "name": "--enable-experimental-web-platform-features"
+                 "type": "runtime_flag",
+                  "name": "--enable-experimental-web-platform-features"
                 }
               ]
             },
@@ -18,8 +18,8 @@
               "version_added": "69",
               "flags": [
                 {
-                "type": "runtime_flag",
-                "name": "--enable-experimental-web-platform-features"
+                  "type": "runtime_flag",
+                  "name": "--enable-experimental-web-platform-features"
                 }
               ]
             },
@@ -27,8 +27,8 @@
               "version_added": "79",
               "flags": [
                 {
-                "type": "runtime_flag",
-                "name": "--enable-experimental-web-platform-features"
+                  "type": "runtime_flag",
+                  "name": "--enable-experimental-web-platform-features"
                 }
               ]
             },
--- a/css/properties/margin-block.json
+++ b/css/properties/margin-block.json
@@ -9,8 +9,9 @@
               "version_added": "69",
               "flags": [
                 {
-                  "type": "runtime_flag",
-                  "name": "--enable-experimental-web-platform-features"
+                  "type": "preference",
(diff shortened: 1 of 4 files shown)


css/properties/margin-block.json
```diff
@@ -6,13 +6,31 @@
...
           "support": {
             "chrome": {
-              "version_added": "69"
+              "version_added": "69",
+              "flags": [
+                {
+                  "type": "runtime_flag",
+                  "name": "--enable-experimental-web-platform-features"
+                }
+              ]
```
ddbeck, line 15: I've opened #6009 to address this more broadly, but we can do our part here to document this flag more consistently. Pending a decision on that, I'd suggest using this form for all the flags listed here:

```
"flags": [
  {
    "type": "preference",
    "name": "Experimental Web Platform features",
    "value_to_set": "Enabled"
  }
]
```

ddbeck:
Hi @phistuck, do you plan to revisit this PR soon?

css/properties/margin-block.json
phistuck: I changed it. I also shared my opinion (spoiler - I do not agree with this) at https://github.com/mdn/browser-compat-data/issues/6009.

phistuck:
> Hi @phistuck, do you plan to revisit this PR soon?

I am so sorry, I must have missed your comments! I definitely would have planned to revisit this and now I have.

css/properties/margin-block.json
ddbeck: Thank you for the changes and your input on #6009. I understand your disagreement and I am sympathetic to your point. I don't expect this to be the final form of the data and by aligning with data we already have, you're making it easier to get consistency with whatever we do end up with. I appreciate your help on this!

PR #6012, opened 2020-04-20
Structure ES6 classes data for Chromium browsers

ddbeck:
In the course of reviewing #5997, I discovered some oddities in the way the data for ES6 classes is recorded for the Chromium browsers. I've attempted to structure some data that was cleverly hidden in notes.

@Elchi3 I've tagged you for review because you've been working on some the JS stuff recently, but it's not urgent.

## Before

Before this PR, the data said that ES6 classes were fully supported in some version and unsupported in previous versions. There was a note, however, that said it _was_ supported in earlier versions, but in strict mode only and non-strict mode was also supported behind a flag. In other words, the note said there was considerable support before the version reported in the data. 😬

## After

This PR restructures the data as follows:

* Supported, fully, in some version X
* Supported in some version Y with a note describing the strict mode only caveat
* Supported, fully, in some version Y behind a flag (for the browsers that support Chromium flags)

This makes the diff look rather large, but it's mostly a copy-and-paste operation. Once you've seen one feature, you've seen 'em all.

Incidentally, this also fixes:

* A few more notes (like in #5997) that had the wrong browser name
* An inconsistency in the Samsung Internet data, between the `javascript.classes` feature and the `javascript.statements.class` feature.

### DIFF
--- a/javascript/classes.json
+++ b/javascript/classes.json
@@ -5,14 +5,48 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/JavaScript/Reference/Classes",
         "spec_url": "https://tc39.es/ecma262/#sec-class-definitions",
         "support": {
-          "chrome": {
-            "version_added": "49",
-            "notes": "From Chrome 42 to 48, strict mode is required. Non-strict mode support can be enabled using the flag \"Enable Experimental JavaScript\"."
-          },
-          "chrome_android": {
-            "version_added": "49",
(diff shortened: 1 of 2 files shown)


ddbeck:
Oh yeah, one more thing: I did check that [the Chrome numbers made sense](https://www.chromestatus.com/features/4633745457938432), so I trusted that the other browser versions were correct too.

PR #6024, opened 2020-04-22
Mark all CSS Shadow Parts features as supported in Safari 13.1

sideshowbarker:
Along with the `::part` pseudo-element marked as Safari-supported in #5941, the `part` and `exportparts` markup attributes and `Element.prototype.part` IDL/DOM property are all included in CSS Shadow Parts, which as noted in #5941 and https://developer.apple.com/documentation/safari_release_notes/safari_13_1_beta_release_notes shipped in Safari 13.1

See also https://bugs.webkit.org/show_bug.cgi?id=202409 and https://bugs.webkit.org/show_bug.cgi?id=202520 and https://bugs.webkit.org/show_bug.cgi?id=202644.

So the whole set of CSS Shadow Parts features was turned on in https://trac.webkit.org/changeset/250889/webkit/, which was after https://trac.webkit.org/log/webkit/tags/Safari-608.2.11 (=Safari 13) and before https://trac.webkit.org/log/webkit/tags/Safari-609.1.20 (=Safari 13.1).

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -5954,10 +5954,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": false
+              "version_added": "13.1"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "13.4"
             },
             "samsunginternet_android": {
               "version_added": true
--- a/html/global_attributes.json
+++ b/html/global_attributes.json
@@ -919,10 +919,10 @@
               "version_added": "60"
             },
             "safari": {
-              "version_added": false
+              "version_added": "13.1"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "13.4"
             },
             "samsunginternet_android": {
               "version_added": false
@@ -1531,10 +1531,10 @@
               "version_added": "60"
             },
             "safari": {
-              "version_added": false
+              "version_added": "13.1"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "13.4"
             },
             "samsunginternet_android": {
               "version_added": false


sideshowbarker:
> I wonder if there would be a way for us to better deal with updates to entire feature sets instead of piecemeal updates to these things.

I think one thing which makes that a challenge is that the descriptions in the release notes don’t have a lot of granularity; for example, in this case, the release notes just said, *“Added support for CSS Shadow Parts.”*, with no further details.

And even digging a little further into the change history for the browser source, and finding https://trac.webkit.org/changeset/250889/webkit/, that just says, *“[CSS Shadow Parts] Enable by default”*, with no further details there either. That however does at least link to a bug, https://bugs.webkit.org/show_bug.cgi?id=202644 — but if you look at that bug, you won’t find any more details there either. The only useful think you will find is that it shows that bug was blocked on another bug at https://bugs.webkit.org/show_bug.cgi?id=202644.

Only after you’ve managed to finally find your way that https://bugs.webkit.org/show_bug.cgi?id=202644 bug will you start to find the kind of details we need — because the **Depends on** field there lists about 15 other bugs, and if you open all those, you start to find mention of the specific CSS feature and the HTML attribute names. But even from there, it’s not immediately clear there was also an affected DOM interface.

In short, I think a reason we sometimes end up doing piecemeal updates is, one of us sees something mentioned in release notes or somewhere and realizes it affects some one specific BCD feature they were already aware of or had seen before in another change, but without realizing it also affects some other BCD feature(s) — because making that connection can require a lot of time spent digging through commit histories and issue/bug trackers to research what specifically actually changed in the browser source, and exactly when.

PR #6026, opened 2020-04-24
Add another caveat related to CORP in Chrome

lol768:
Since this has bitten me in a production system, and it would've been really useful to get a warning about on the MDN docs page.

https://bugs.chromium.org/p/chromium/issues/detail?id=1074261

Version 80 comes from:

> https://crbug.com/1063034#c14 indicates that, due to a bug, partial PDF loading has not been used for some time. Now that it seems to be working again in M80, we're getting a number of new bugs involving partial loading, linearized PDFs, etc.

### DIFF
--- a/http/headers/cross-origin-resource-policy.json
+++ b/http/headers/cross-origin-resource-policy.json
@@ -7,11 +7,17 @@
           "support": {
             "chrome": {
               "version_added": "73",
-              "notes": "Until version 75, downloads for files with this header would fail in Chrome. See <a href='https://crbug.com/952834'>bug 952834</a>."
+              "notes": [
+                "Until version 75, downloads for files with this header would fail in Chrome. See <a href='https://crbug.com/952834'>bug 952834</a>.",
+                "From version 80, linearized PDFs served inline with this header fail to render properly. See <a href='https://crbug.com/1074261'>bug 1074261</a>."
+              ]
             },
             "chrome_android": {
               "version_added": "73",
-              "notes": "Until version 75, downloads for files with this header would fail in Chrome. See <a href='https://crbug.com/952834'>bug 952834</a>."
+              "notes": [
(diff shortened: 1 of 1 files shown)


http/headers/cross-origin-resource-policy.json
```diff
@@ -7,11 +7,17 @@
           "support": {
             "chrome": {
               "version_added": "73",
-              "notes": "Until version 75, downloads for files with this header would fail in Chrome. See <a href='https://crbug.com/952834'>bug 952834</a>."
+              "notes": [
+                "Until version 75, downloads for files with this header would fail in WebView. See <a href='https://crbug.com/952834'>bug 952834</a>.",
```
vinyldarkscratch, line 11:
```suggestion
                "Until version 75, downloads for files with this header would fail in Chrome. See <a href='https://crbug.com/952834'>bug 952834</a>.",
```

http/headers/cross-origin-resource-policy.json
```diff
@@ -7,11 +7,17 @@
...
+                "From version 80, linearized PDFs served inline with this header fail to render properly. See <a href='https://crbug.com/1074261'>bug 1074261</a>."
+              ]
             },
             "chrome_android": {
               "version_added": "73",
-              "notes": "Until version 75, downloads for files with this header would fail in Chrome. See <a href='https://crbug.com/952834'>bug 952834</a>."
+              "notes": [
+                "Until version 75, downloads for files with this header would fail in WebView. See <a href='https://crbug.com/952834'>bug 952834</a>.",
```
vinyldarkscratch, line 18:
```suggestion
                "Until version 75, downloads for files with this header would fail in Chrome. See <a href='https://crbug.com/952834'>bug 952834</a>.",
```

lol768:
Thanks for the review!

> Looks like there was an accidental change of words in the existing notes, however

:see_no_evil: That'll teach me not to copy/paste them. Should be all sorted now, I think/

PR #3530, opened 2019-02-26
setting periodicsync support data to false, as it was never implement…

chrisdavidmills:
…ed anywhere

After a [conversation on Twitter](https://twitter.com/jaffathecake/status/1100307442477486080), we realised that `periodicSync` was never implemented anywhere. It was mentioned in the explainer briefly at some point, but that's about as far as it got. I suspect it was in the spec at some point, otherwise we never would've documented it...

Supporting evidence for this change:

* Chrome folks have confirmed it is not available in Chrome
* it isn't in Gecko: https://dxr.mozilla.org/mozilla-central/source/dom/webidl/ServiceWorkerRegistration.webidl
* It isn't in Webkit: https://trac.webkit.org/browser/webkit/releases/Apple/Safari%2012.0.3/WebCore/workers/service/ServiceWorkerRegistration.idl 

Anyway, I am setting all the support info to `false` rather than just removing it, because people have talked about the idea now, and need to be set straight, plus it might be explored in the future.

PR #6030, opened 2020-04-26
Update support for periodicSync of ServiceWorkerRegistration

webmaxru:
Based on the status update 'version_added' was set to '80':
- "Enabled by default": https://chromestatus.com/features/5689383275462656
- "Fixed (Closed)": https://bugs.chromium.org/p/chromium/issues/detail?id=925297
- "Launch": https://web.dev/periodic-background-sync/

Also, 'standard_track' was changed to 'true': as this is a relevant status now: "becoming a standard (waiting for all vendor implementation)"

### DIFF
--- a/api/ServiceWorkerRegistration.json
+++ b/api/ServiceWorkerRegistration.json
@@ -519,7 +519,7 @@
           },
           "status": {
             "experimental": true,
-            "standard_track": true,
+            "standard_track": false,
             "deprecated": false
           }
         }
--- a/api/ServiceWorkerRegistration.json
+++ b/api/ServiceWorkerRegistration.json
@@ -481,13 +481,13 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/ServiceWorkerRegistration/periodicSync",
           "support": {
             "chrome": {
-              "version_added": false
+              "version_added": "80"
             },
             "chrome_android": {
-              "version_added": false
+              "version_added": "80"
             },
             "edge": {
-              "version_added": false
+              "version_added": "80"
             },
             "firefox": {
               "version_added": false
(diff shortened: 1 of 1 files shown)


api/ServiceWorkerRegistration.json
```diff
@@ -514,12 +514,12 @@
...
             },
             "webview_android": {
-              "version_added": false
+              "version_added": "80"
             }
           },
           "status": {
             "experimental": true,
-            "standard_track": false,
+            "standard_track": true,
```
ddbeck, line 522: I dug into this a bit. This feature has had a complicated specification history (see also #3530). It doesn't look like there's a spec with `periodicSync` and it's not clear what spec it may ultimately make its way into (e.g., see https://github.com/WICG/BackgroundSync/issues/156). Until there's something a bit more definitive, I'd like to hold off marking this as `true`.

```suggestion
            "standard_track": false,
```

webmaxru: Hello Daniel! It's my pleasure to send my first PR to BCD :)
I based my decision on `standard_track` value on 3 points:
- Explanation of this property https://github.com/mdn/browser-compat-data/issues/4650#issuecomment-523094023
- Spec https://wicg.github.io/BackgroundSync/spec/PeriodicBackgroundSync-index.html
- This spec is implemented in Chromium-based browsers and shipped in the stable channel

For me, it looked like the case number 2 from the `standard_track` description. On the other hand, this spec is still a "_Draft Community Group Report_". So I agree with you - let's wait for some extra signals to call it a part of the standard.

ddbeck: Yes, I admit there's a bit of a judgement call on the standards question. Our [schema docs](https://github.com/mdn/browser-compat-data/blob/master/schemas/compat-data-schema.md#status-information) officially say the value should be `true` if the feature is "part of an active […] specification process". But we could probably use a clearer definition of "active." 😄

PR #6049, opened 2020-04-29
AudioContext.p.state|onstatechange in Safari 9

sideshowbarker:
https://trac.webkit.org/changeset/182141/webkit added support for `AudioContext.p.state` and `AudioContext.p.onstatechange` to WebKit, which per https://trac.webkit.org/log/webkit/tags/Safari-601.1.25 went into WebKit 601.1.25 and so into Safari 9 and iOS Safari 9.

### DIFF
--- a/api/BaseAudioContext.json
+++ b/api/BaseAudioContext.json
@@ -1638,10 +1638,12 @@
               "version_added": true
             },
             "safari": {
-              "version_added": null
+              "version_added": "9",
+              "prefix": "webkit"
             },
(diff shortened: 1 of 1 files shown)


sideshowbarker:
The lint failure is due to this depending on https://github.com/mdn/browser-compat-data/pull/6050. So if https://github.com/mdn/browser-compat-data/pull/6050 is merged first and we rebase this branch, the lint failure here would go away.

PR #6053, opened 2020-04-29
removal of Firefox desktop support for window.onappinstalled

chrisdavidmills:
As per https://bugzilla.mozilla.org/show_bug.cgi?id=1625384

Basically, `window.onappinstalled` has been removed in Fx 76.

We'll do the same for Fx mobile when the new version is available.

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -4440,22 +4440,17 @@
             "edge": {
               "version_added": "≤79"
             },
-            "firefox": [
-              {
-                "version_added": false
-              },
-              {
-                "version_added": "49",
-                "version_removed": "76",
-                "flags": [
-                  {
-                    "type": "preference",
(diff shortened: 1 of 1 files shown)


api/Window.json
```diff
@@ -4440,16 +4440,22 @@
...
-                {
-                  "type": "preference",
-                  "name": "dom.manifest.onappinstall",
-                  "value_to_set": "true"
-                }
-              ]
-            },
+            "firefox": [
+              {
+                "version_added": false
```
ddbeck, line 4445: This support statement isn't really necessary. The `"version_removed"` in the next statement has the same effect. I'd suggest removing this one.

chrisdavidmills:
OK, update made as suggested. Thanks @ddbeck !

PR #6054, opened 2020-04-29
Correct Chrome + Safari data for Element.getAttribute()

vinyldarkscratch:
While working on getting all real values for the `Attr` API, I found some errors in the data for `Element.getAttribute()`.  Chrome and Safari had support for this feature for MUCH, much longer than was initially mentioned.  Determined via manual testing.

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -29,7 +29,7 @@
             "version_added": "10.1"
           },
           "safari": {
-            "version_added": "3"
+            "version_added": "1.3"
           },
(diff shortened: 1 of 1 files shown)


vinyldarkscratch:
Oops, thought I checked for linter errors -- fixed now!

sideshowbarker:
👍 I think the data for core DOM features is especially important — so I think it’s great when we get these kinds of accuracy refinements made

PR #3403, opened 2019-02-06
Fix version inconsistencies in api/* (N through S)

vinyldarkscratch:
Sub-PR of #3368 for the fixes of the files themselves.

PR #6057, opened 2020-04-30
fix(Request): Safari 12.1+ supports `"same-origin"` default

keithamus:
This sets support for the `Request` "Default value `same-origin`" for Safari 12.1+.

I ran this: `(new Request(window.location.href)).credentials` code in Browserstack for all Safari versions, plus a local copy of Safari 13.1, here are the results:

 - 9.1 `ReferenceError: Can't find variable: Request`
 - 10.1 `"omit"`
 - 11.1 `"omit"`
 - 12.1 `"same-origin"`
 - 13: `"same-origin"`
 - 13.1: `"same-origin"`

Safari iOS:

 - 9 `ReferenceError: Can't find variable: Request`
 - 10: `"omit"`
 - 11: `"omit"`
 - 12.1: `"same-origin"`

image
image

The change was documented in Safari Tech Preview 61 (https://developer.apple.com/safari/technology-preview/release-notes/#r61), 

It looks like this was originally set to `12.1` in https://github.com/mdn/browser-compat-data/pull/3283. But support was accidentally removed in https://github.com/mdn/browser-compat-data/pull/3403 As far as I can see Safari 12.0 was based off of TP58, but I don't have 12.0 to test, so I have set this to 12.1 because I can test that.

### DIFF
--- a/api/Request.json
+++ b/api/Request.json
@@ -830,7 +830,7 @@
                 "version_added": false
               },
               "safari": {
-                "version_added": false
+                "version_added": "12.1"
               },
               "safari_ios": {
                 "version_added": false
--- a/api/Request.json
+++ b/api/Request.json
@@ -783,10 +783,10 @@
               }
             ],
             "safari": {
-              "version_added": false
+              "version_added": "10.1"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "10.1"
             },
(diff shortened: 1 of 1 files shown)


keithamus:
@Elchi3 I've updated it, and also tested in iOS and added the support there also. 👍

PR #6058, opened 2020-04-30
Arrow functions don't work on IOS.

LucasLelaidier:
On both ios safari and ios chrome i got this error :

`SyntaxError: Unexpected token '='. Expected an opening '(' before a method's parameter list.`

for this line :

`setLoading = (bool) => {`

I had to change all my arrows function with normal functions and replace this with self, and all works fine.

I tested on a real iPhone X and a real iPhone 6, and also on lambdaTest that emulate iPhones.

Hope this PR could help !

### DIFF
--- a/javascript/functions.json
+++ b/javascript/functions.json
@@ -307,7 +307,7 @@
               "version_added": "10"
             },
             "safari_ios": {
-              "version_added": "10"
+              "version_added": false
             },
             "samsunginternet_android": {
               "version_added": "5.0"


jpmedley:
Which version of Safari is running on your system?

LucasLelaidier:
I tried on ios 13.3 (so Safari 13 i gess)

LucasLelaidier:
Hum, it seems that the problem is more complex than just arrow function don't work, because
```
<div id="test">
    Don't work
</div>

<script>

// Test of arrow functions on IOS

const hello = () => {
    console.log('Works');
    document.getElementById("test").innerHTML = "Works !"
}

hello();

</script>`
```
print works.

I don't know what was is the problem with my previous code.

vinyldarkscratch:
I didn't really think that Safari for iOS/iPadOS would be lacking such a core syntax feature when Safari for macOS had support -- and after some quick tests, my suspicions were confirmed.  From both some quick testing on my iPad, as well as the comments above, it sounds like the issue you're experiencing isn't a compatibility issue, but some sort of issue with your code.

I'm going to close this pull request as a result.  If you'd like, I'd be willing to help you diagnose what's going on within your code in private -- feel free to reach out to me through any of my contact methods!  😉

PR #6073, opened 2020-05-04
DataTransferItemList.p.remove support in Safari 6

sideshowbarker:
https://trac.webkit.org/changeset/220829/webkit added support for `DataTransferItemList.p.remove` to WebKit. So https://trac.webkit.org/log/webkit/tags/Safari-604.2.4, which means Safari 6.


sideshowbarker:
Lint failing because this depends on #6072

PR #6074, opened 2020-05-04
Update DataTransferItemList Chrome support data

sideshowbarker:
https://source.chromium.org/chromium/chromium/src/+/9f23921 and https://src.chromium.org/viewvc/blink?revision=156015&view=revision show that `DataTransferItemList.p.remove()` support was added to the Blink sources in 2013-08-13. Based on that date, that’d seem to mean it shipped in Chrome 30.

---

For the days and weeks after `DataTransferItemList.p.remove()` landed in the sources on 2013-08-13, here are the release dates:

 * Chrome 29: 2013-08-20
 * Chrome 30: 2013-10-01
 * Chrome 31: 2013-11-12
 * Chrome 32: 2014-01-14

So it’s not impossible it shipped in Chrome 29 — just three days after being merged — but that seems unlikely. I guess it’s also possible it shipped as late as 5 months or so later, in Chrome 32. But it seems most likely it shipped in Chrome 30, around 45 days after it landed.

In general, for Blink/Chromium changes from that era — which I guess amounts to the range of some years after Blink was first split off from WebKit but before systematic association of tags with revisions started happening — I don’t know how to deterministically identify what Chrome release a particular revision shipped in.

I tried https://storage.googleapis.com/chromium-find-releases-static/index.html with 9f239214f3c903fd7ad4491a67372070d116325c but it didn’t give me anything…

cc @jpmedley for guidance

### DIFF
--- a/api/DataTransferItemList.json
+++ b/api/DataTransferItemList.json
@@ -5,10 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/DataTransferItemList",
         "support": {
           "chrome": {
-            "version_added": "4"
+            "version_added": "13"
           },
           "chrome_android": {
-            "version_added": false
+            "version_added": "18"
           },
           "edge": {
             "version_added": "≤18"
@@ -35,10 +35,10 @@
             "version_added": "6"
           },
           "samsunginternet_android": {
-            "version_added": false
+            "version_added": "1.0"
           },
           "webview_android": {
-            "version_added": false
+            "version_added": "4.4"
           }
         },
         "status": {
@@ -53,10 +53,10 @@
           "description": "<code>DataTransferItemList[]</code>",
           "support": {
             "chrome": {
-              "version_added": "4"
+              "version_added": "13"
             },
             "chrome_android": {
-              "version_added": false
+              "version_added": "18"
             },
             "edge": {
               "version_added": "≤79"
@@ -83,10 +83,10 @@
               "version_added": "6"
             },
             "samsunginternet_android": {
-              "version_added": false
+              "version_added": "1.0"
             },
             "webview_android": {
-              "version_added": false
+              "version_added": "4.4"
             }
           },
           "status": {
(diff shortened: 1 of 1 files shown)


sideshowbarker:
The lint failure will be gone after #6080 is merged (which in turn has a lint failure that will be resolved after this PR is merged, so it’s circular dependency…).

jpmedley:
@sideshowbarker Stable shipping dates aren't what I look at when I have no other way to tell when something shipped. Feature freeze is what I care about. 

For example, Chrome 28 feature freeze was on April 19, 2013. If a feature laded on April 20, 2013, it's more likely to be in Chrome 29 than in Chrome 28.

I just shared a spreadsheet with listing feature freezes for Chrome.

sideshowbarker:
> @sideshowbarker Stable shipping dates aren't what I look at when I have no other way to tell when something shipped. Feature freeze is what I care about.

OK thanks yeah I had figured you must have a systematic way like that.

> For example, Chrome 28 feature freeze was on April 19, 2013. If a feature laded on April 20, 2013, it's more likely to be in Chrome 29 than in Chrome 28.

That’s exactly the kind of more-precise info I had been hoping for. I’ll update the patch

> I just shared a spreadsheet with listing feature freezes for Chrome.

w00t — thanks super much 🎉 I feel like I’ve just unlocked a new secret level in this adventure 😋

sideshowbarker:
OK, given that the code for `DataTransferItemList.remove` landed in the Chrome sources on 2013-08-13, that was too late to make the Chrome 30 feature freeze, which was on 2013-07-26. The next feature freeze was 2013-09-06, for Chrome 31. So it seems this shipped in Chrome 31.

I’ve updated the version data here accordingly.

vinyldarkscratch:
Marking as "not ready" for now, due to blockage by #6080.

vinyldarkscratch:
Actually never mind, I just noticed the circular dependency!  I removed the "not ready" label accordingly.

Since both PRs are related to one another and modify the same API, it might be more reasonable to combine the two?

sideshowbarker:
> Marking as "not ready" for now, due to blockage by #6080.

OK, I’ve now merged that #6080 patch into this PR branch — and the lint no longer fails.

So think is now ready to be merged.

ddbeck:
@jpmedley are you permitted to share that sheet of Chrome feature freeze dates more widely? I imagine that would be a valuable resource for reviewing PRs related to older features.

jpmedley:
Check your inbox.

PR #6081, opened 2020-05-04
Mark DataTransferItem supported in Chrome 11

sideshowbarker:
https://trac.webkit.org/changeset/80536/webkit added `DataTransferItem` support to WebKit. https://trac.webkit.org/log/webkit/tags/Safari-534.24 is where that puts it, which means it shipped in Chrome 11.


sideshowbarker:
@jpmedley thanks!

PR #6113, opened 2020-05-07
Update for Samsung Internet 11.2 and 12.0

AdaRoseCannon:
Made using the mirroring script


Elchi3:
Thanks for the PR @AdaRoseCannon! :tada: Can you comment on why we should have 11.2 as opposed to mark things as supported in 11.0? Isn't 11.2 just a minor version without new features?
This came up in https://github.com/mdn/browser-compat-data/pull/6099

AdaRoseCannon:
In this case it is us implementing a feature early. It would otherwise be
out in 12.0.

api/WebGLRenderingContext.json
```diff
@@ -5693,7 +5693,7 @@
               "version_added": false
             },
             "samsunginternet_android": {
-              "version_added": false
+              "version_added": "12.0"
```
Elchi3, line 5696: This is another XR API, is it 12.0 or 11.2?

AdaRoseCannon: Nice spot, thank you!!

api/Navigator.json
```diff
@@ -1051,6 +1051,7 @@
             },
             "samsunginternet_android": {
               "version_added": true,
+              "version_removed": false,
```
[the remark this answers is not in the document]
AdaRoseCannon: It seems it was only available in the GearVR version of the browser, I think perhaps it's best to mark it `version_added: false`

PR #6143, opened 2020-05-13
Document and element animation events support in IE

### LINKED ISSUE #6135: animationend is supported by IE11
I'm not sure what to do here. The browser compat data doesn't say when IE got support for the `animationend` event, so I tried it in IE11 and it worked. (It doesn't work in IE9.) I don't have IE10 to try it on.

Do I do a PR for `/api/HTMLElement.json` saying: `"version_added": "11"` even though I don't know whether it was added in 10 or 11?

tjcrowder:
Update IE support for document and element animation events. Fixes #6135.

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -425,7 +425,7 @@
               "version_added": "54"
             },
             "ie": {
-              "version_added": null
+              "version_added": false
             },
             "opera": {
               "version_added": false
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -164,7 +164,7 @@
               "version_added": "54"
             },
             "ie": {
-              "version_added": null
+              "version_added": false
             },
             "opera": {
               "version_added": false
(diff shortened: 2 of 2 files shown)


tjcrowder:
@ddbeck - That's a good point. Unfortunately, I downloaded the Edge v12 VM from [here](https://drive.google.com/drive/folders/1AcgKWuSAPpWCfClrrggtDUU2hl_6ETv9) but (unlike the IE10 VM I got from there), it won't boot for me under VirtualBox. :-| I *can* attest to the fact that Edge v44 supports `animationstart`, `animationend`, and `animationiteration` but not `animationcancel` just like IE10-11 do. :-) Should I at least add that?

ddbeck:
I tested Edge myself (in Sauce Labs) and confirmed that `animationstart, `animationend`, and `animationiteration` are supported from Edge version 12 and up. If you'd add that info, it'd be very helpful.

tjcrowder:
@ddbeck - Great, thanks for doing that. Done!

PR #6147, opened 2020-05-13
Update versions for Attr API

vinyldarkscratch:
This PR updates various versions for the `Attr` API based upon both mirroring and manual testing.  Data is as follows:

```
api.Attr
	- Chrome - 1
	- IE - 8
	- Opera - 8
	- Safari - 1.3
api.Attr.localName
	- IE - false
	- Opera - Blink
	- Safari - false
api.Attr.namespaceURI
	- IE - false
	- Opera - Blink
	- Safari - false
api.Attr.prefix
	- IE - false
	- Opera - Blink
	- Safari - false
```

### DIFF
--- a/api/Attr.json
+++ b/api/Attr.json
@@ -5,11 +5,11 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/Attr",
         "support": {
           "chrome": {
-            "version_added": true,
+            "version_added": "1",
             "notes": "As of Chrome 45, this property no longer inherits from Node."
           },
           "chrome_android": {
-            "version_added": true,
+            "version_added": "18",
             "notes": "As of Chrome 45, this property no longer inherits from Node."
           },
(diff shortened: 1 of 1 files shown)


sideshowbarker:
I’d like to understand the Safari changes in this revision. As far as I can see from testing in Safari 13.1, it does in fact expose all the following:

* `Attr.prototype.localName`
* `Attr.prototype.namespaceURI`
* `Attr.prototype.prefix`

Safari 13.1 also passes the related `Attr` tests in https://wpt.live/dom/idlharness.window.html

vinyldarkscratch:
Good catch!  I just took a look and I'm not sure what led me to say "false" initially, but looking again I can confirm there's support.  Huh...

I double-checked IE and confirmed that it's truly got no support, so that just leaves Safari.  I'll get a follow-up going ASAP!

PR #6153, opened 2020-05-14
Safari supports Rest in arrays

vinyldarkscratch:
Part of #5934. Safari supports "rest in arrays" destructuring since version 9.1 (based upon manual testing). This PR updates the data accordingly.

### DIFF
--- a/javascript/operators/destructuring.json
+++ b/javascript/operators/destructuring.json
@@ -144,10 +144,10 @@
                 "version_added": "36"
               },
               "safari": {
-                "version_added": "9"
+                "version_added": "9.1"
               },
               "safari_ios": {
-                "version_added": "9"
+                "version_added": "9.3"
               },
               "samsunginternet_android": {
                 "version_added": "5.0"
--- a/javascript/operators/destructuring.json
+++ b/javascript/operators/destructuring.json
@@ -144,10 +144,10 @@
                 "version_added": "36"
               },
               "safari": {
-                "version_added": false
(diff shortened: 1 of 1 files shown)


javascript/operators/destructuring.json
```diff
@@ -144,10 +144,10 @@
                 "version_added": "36"
               },
               "safari": {
-                "version_added": false
+                "version_added": "9"
```
sideshowbarker, line 147:
```suggestion
                "version_added": "9.1"
```
It looks like https://trac.webkit.org/changeset/192671/webkit, 19 Nov 2015, was the revision which landed support for rest parameters. I think that means it couldn’t have gone into Safari 9, which shipped on 30 Sep 2015, but instead went into Safari 9.1, which shipped on 21 Mar 2016.

javascript/operators/destructuring.json
```diff
@@ -144,10 +144,10 @@
...
               },
               "safari_ios": {
-                "version_added": false
+                "version_added": "9"
```
sideshowbarker, line 150:
```suggestion
                "version_added": "9.3"
```

javascript/operators/destructuring.json
vinyldarkscratch: Agh, I forgot: Sauce Labs' "Safari 9" is actually "Safari 9.1"...

sideshowbarker:
👍 Very nice to see all these (from #5934 getting landed

PR #6163, opened 2020-05-15
Update api.lFileSystem(Sync) status and URLs

bershanskiy:
This updates two files to match existing status descriptions and MDN URL patterns.

Mark these interfaces:
 - "experimental" since only Chromium ever implemented them.
 - "standard_track" since there was a spec.
 - "deprecated" since the spec was abandoned.


bershanskiy:
Now when I had time to think about this a bit longer, I think `LocalFileSystem` members should be documented under `Window` like [`Window.requestFileSystem`](https://developer.mozilla.org/en-US/docs/Web/API/Window/requestFileSystem). Then [`LocalFileSystem.requestFileSystem`](https://github.com/mdn/browser-compat-data/blob/master/api/LocalFileSystem.json) would be merged with `Window.requestFileSystem` and `LocalFileSystem.resolveLocalFileSystemURL` would be moved to `Window.resolveLocalFileSystemURL`.
Does this make sense?

PR #6165, opened 2020-05-15
Remove flag from viewport.

jpmedley:
https://storage.googleapis.com/chromium-find-releases-static/37a.html#37aaaed4e1e12d7e05ddb898acc43c1abb56a24c

### DIFF
--- a/css/at-rules/viewport.json
+++ b/css/at-rules/viewport.json
@@ -8,7 +8,7 @@
           "support": {
             "chrome": [
               {
-                "version_added": "84"
+                "version_removed": "84"
               },
               {
                 "version_added": "29",
@@ -27,7 +27,7 @@
             ],
             "chrome_android": [
               {
-                "version_added": "84"
+                "version_removed": "84"
               },
               {
                 "version_added": "29",
@@ -117,9 +117,14 @@
             "samsunginternet_android": {
               "version_added": "2.0"
             },
-            "webview_android": {
-              "version_added": "4.4"
-            }
+            "webview_android": [
+              {
+                "version_removed": "84"
+              },
+              {
+                "version_added": "4.4"
+              }
+            ],
           },
           "status": {
             "experimental": false,
--- a/css/at-rules/viewport.json
+++ b/css/at-rules/viewport.json
@@ -6,44 +6,36 @@
           "description": "<code>@viewport</code>",
           "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/@viewport",
           "support": {
-            "chrome": [
-              {
-                "version_removed": "84"
-              },
-              {
-                "version_added": "29",
-                "flags": [
-                  {
-                    "type": "runtime_flag",
-                    "name": "--enable-experimental-webkit-features"
-                  },
-                  {
-                    "type": "runtime_flag",
-                    "name": "--enable-viewport"
-                  }
-                ],
-                "notes": "See Chromium <a href='https://crbug.com/235457'>bug 235457</a>."
-              }
-            ],
-            "chrome_android": [
-              {
-                "version_removed": "84"
-              },
-              {
-                "version_added": "29",
-                "flags": [
-                  {
-                    "type": "runtime_flag",
-                    "name": "--enable-experimental-webkit-features"
-                  },
-                  {
-                    "type": "runtime_flag",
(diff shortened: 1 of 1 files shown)


ddbeck:
@jpmedley reading the linked change and associated bug, it looks like `@viewport` has been removed. Is there another commit you meant to link to or have I grossly misunderstood something?

jpmedley:
@ddbeck Sorry for the delay. The answer to this is so far back in Git history that it's hard to get to on our web interface. I'm having to install our source code on my machine and have run into some issues. I am working on it though.

jpmedley:
Sorry for the delay. I had a little trouble getting this one sorted. 

Since you're likely to ask, yes, this is right. This feature never got out from behind its runtime flag. It went straight to removal.

ddbeck:
@jpmedley thank you for coming back to this!

One last question here: if it never get out from behind a flag before being removed, then would it be more correct to say that this never shipped on WebView? I ask because I think previously you've mentioned that WebView doesn't have flags, unlike Chrome. If that's the case, then all of the WebView entries for `@viewport` could be replaced with `false` here.

css/at-rules/viewport.json
```diff
@@ -97,6 +110,7 @@
               "version_added": "2.0"
             },
             "webview_android": {
+              "version_removed": "84",
```
jpmedley, line 113:
```suggestion
              "version_removed": "37",
```

css/at-rules/viewport.json
```diff
@@ -159,6 +175,7 @@
                 "version_added": "2.0"
               },
               "webview_android": {
+                "version_removed": "84",
```
jpmedley, line 178:
```suggestion
                "version_removed": "37",
```

css/at-rules/viewport.json
```diff
@@ -222,6 +241,7 @@
...
```
jpmedley, line 244:
```suggestion
                "version_removed": "37",
```

css/at-rules/viewport.json
```diff
@@ -285,6 +307,7 @@
...
```
jpmedley, line 310:
```suggestion
                "version_removed": "37",
```

css/at-rules/viewport.json
```diff
@@ -336,6 +361,7 @@
...
                 "version_added": "37"
```
jpmedley, line 365:
```suggestion
                "version_added": false
```

css/at-rules/viewport.json
```diff
@@ -399,6 +427,7 @@
                 "version_added": "2.0"
               },
               "webview_android": {
+                "version_removed": "84",
```
jpmedley, line 430:
```suggestion
                "version_removed": "37",
```

css/at-rules/viewport.json
```diff
@@ -462,6 +493,7 @@
...
```
jpmedley, line 496:
```suggestion
                "version_removed": "37",
```

css/at-rules/viewport.json
```diff
@@ -513,6 +547,7 @@
...
                 "version_added": "37"
```
jpmedley, line 551:
```suggestion
                "version_added": false
```

css/at-rules/viewport.json
```diff
@@ -618,6 +655,7 @@
...
```
jpmedley, line 659:
```suggestion
                "version_added": false
```

css/at-rules/viewport.json
```diff
@@ -730,6 +770,7 @@
                 "version_added": "2.0"
               },
               "webview_android": {
+                "version_removed": "84",
```
jpmedley, line 773:
```suggestion
                "version_removed": "37",
```

css/at-rules/viewport.json
```diff
@@ -781,6 +824,7 @@
                 "version_added": "8.0"
               },
               "webview_android": {
+                "version_removed": "84",
```
jpmedley, line 827:
```suggestion
                "version_removed": "37",
```

jpmedley:
You're right. Somebody thinks some fo these features were in pre-chromium WebView. I've marked those as removed in 37, which was the first Chromium WebView.

PR #6166, opened 2020-05-15
Move a version-based test from test-browsers to test-versions

queengooborg:
I was looking at `test-browsers.js` and found the inclusion of a test that checks if there are multiple support statements with only a `version_added` and nothing else.  However, this is clearly not a _browser_ test, which means the test was placed in the wrong location.  This PR moves it to `test-versions.js` instead.


foolip (commented):
Are there any tests for the lint code that ensures that they actually fail for certain input?

queengooborg:
No, we actually only have one unittest for one linter file at this point in time.  I plan to spend some time in the near future adding them though!

caugner (approved):
LGTM

PR #6167, opened 2020-05-15
Add linter for browser data

### LINKED ISSUE #6160: Browsers cannot have multiple current versions
A couple of times, we’ve accidentally had multiple `current` versions marked for browsers.  However, the rule for the `current` version is that it must always be the most recent stable release.  We should write a linter to allow only one `current`, `beta`, `nightly`, and `esr` version per browser.  (For NodeJS, it can have multiple `current` and `esr` versions.)

queengooborg:
Fixes #6160.  This PR adds a linter script for the browser data to test if browsers have multiple `current`, `beta`, `nightly`, and `esr` versions (aside from NodeJS which is only tested for multiple `beta` and `nightly`).  Since there was already a `test-browsers.js`, this moves the existing file to `test-browsers-presence.js`, creating a new `test-browsers-data.js` file.


test/linter/test-browsers-data.js
```diff
@@ -0,0 +1,67 @@
...
+    esr: null,
+  };
+
+  for (let releaseVersion in releases) {
+    const releaseData = releases[releaseVersion];
+
+    if (
+      (browser == 'nodejs'
+        ? ['beta', 'nightly']
+        : ['current', 'beta', 'nightly', 'esr']
```
saschanaz, line 33: Does this mean we allow multiple `current` on Node.js?

vinyldarkscratch: Based upon the linter, we do, however I realize that NodeJS wouldn't have multiple `current` releases, only multiple `esr` releases.  Lemme fix that up real quick...

saschanaz: Looking at the Firefox ESR release cycle, It seems Firefox can also have multiple ESR releases for a short time. https://support.mozilla.org/en-US/kb/firefox-esr-release-cycle

test/linter/test-browsers-data.js
```diff
@@ -0,0 +1,68 @@
...
+    esr: null,
+  };
+
+  for (let releaseVersion in releases) {
+    const releaseData = releases[releaseVersion];
+
+    if (['current', 'beta', 'nightly', 'esr'].includes(releaseData.status)) {
+      if (
+        releaseData.status === 'esr' &&
+        ['nodejs', 'firefox'].includes(browser)
```
saschanaz, line 33: Can we even ban the `esr` value from all other browsers? It seems there are two cases here: multiple ESR releases or no ESR at all.

vinyldarkscratch: To be honest, NodeJS and Firefox are really the only two that _have_ ESR versions to begin with.  I guess we don't really need to add the special-case logic because of that.

test/linter/test-browsers-data.js
```diff
@@ -0,0 +1,62 @@
...
+ */
+function processData(data, logger) {
+  const browser = Object.keys(data.browsers)[0];
+  const releases = data.browsers[browser].releases;
+
+  let releaseByStatus = {
+    current: null,
+    beta: null,
+    nightly: null,
+    esr: null,
```
saschanaz, line 24: (This also is redundant then)

foolip:
Does the lint pass now after rebase?

queengooborg:
Yep, the lint now passes!

test/linter/test-browsers-data.js
```diff
@@ -0,0 +1,57 @@
...
+ * @typedef {import('../utils').Logger} Logger
+ */
+
+/**
+ * @param {Identifier} data
+ * @param {Logger} logger
+ * @returns {boolean}
+ */
+function processData(data, logger) {
+  const browser = Object.keys(data.browsers)[0];
```
foolip, line 17: Why is only one browser tested?

test/linter/test-browsers-data.js
```diff
@@ -0,0 +1,57 @@
...
+  const releases = data.browsers[browser].releases;
+
+  let releaseByStatus = {
```
foolip, line 20: Is this unused?

test/linter/test-browsers-data.js
queengooborg: This test is run per each file, which only contains one browser!

test/linter/test-browsers-data.js
queengooborg: Looks like that was left over from a former revision and is no longer used -- removed!

queengooborg:
@foolip Once the lint passes again (depends on #16151), can we merge this soon before another issue occurs in our browser data?

test/linter/test-browsers-data.js
foolip: I see! But could we just loop over the browsers anyway so that it would also work if linting all of BCD?

Or if you'd prefer to keep it as is, a comment or an assert to explain?

foolip (approved):

test/linter/test-browsers-data.js
queengooborg: I've added a comment to explain the reasoning -- I'm expecting this to change once we get into the next-gen linter!

PR #6170, opened 2020-05-15
Add target version to mirroring script

vinyldarkscratch:
This PR adds the ability for us to specify a version to isolate when running the mirroring script.  When a new browser version is released and we update the data, this will allow us to selectively mirror data that only affects this new browser version.

### DIFF
--- a/scripts/mirror.js
+++ b/scripts/mirror.js
@@ -41,6 +41,13 @@ const { argv } = require('yargs').command(
           'Specify when to perform mirroring, whether on true/null ("nonreal", default), true/null/false ("bool"), or always ("always")',
         type: 'string',
         default: 'nonreal',
+      })
+      .option('target_version', {
+        alias: 't',
+        describe:
+          "Only perform mirroring if it affects this destination browser's release",
+        type: 'string',
+        default: undefined,
       });
   },
 );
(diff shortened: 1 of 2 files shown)


alshamiri5 (approved):

scripts/mirror.js
```diff
@@ -138,6 +145,46 @@ const getSource = (browser, forced_source) => {
...
+/**
+ * @param {SupportStatement} compatData
+ * @param {string | null} versionToCheck
+ * @return {string | null}
+ */
+const isVersionAdded = (compatData, versionToCheck) => {
+  if (typeof compatData.version_added === 'string')
+    return compatData.version_added == versionToCheck;
+
+  if (compatData.constructor === Array) {
```
foolip, line 157:
```suggestion
  if (Array.isArray(compatData)) {
```

scripts/mirror.js
```diff
@@ -138,6 +145,46 @@ const getSource = (browser, forced_source) => {
...
+    for (var i = compatData.length - 1; i >= 0; i--) {
+      if (compatData[i].version_added == versionToCheck) {
+        return true;
```
foolip, line 160: How about https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/some instead of this loop?

scripts/mirror.js
```diff
@@ -531,8 +578,15 @@ const bumpGeneric = (originalData, sourceData, source) => {
...
  * @param {SupportStatement} originalData
+ * @param {string} target_version
  */
-const bumpVersion = (data, destination, source, originalData) => {
+const bumpVersion = (
+  data,
+  destination,
+  source,
+  originalData,
+  target_version,
```
foolip, line 588:
```suggestion
  targetVersion,
```

scripts/mirror.js
queengooborg: I keep forgetting `Array.some()` is a thing... :P

scripts/mirror.js
```diff
@@ -51,6 +51,13 @@ const { argv } = require('yargs').command(
           'Specify when to perform mirroring, whether on true/null ("nonreal", default), true/null/false ("bool"), or always ("always")',
         type: 'string',
         default: 'nonreal',
+      })
+      .option('target_version', {
```
caugner, line 55: Looking at `git`s and `yarn`s command line options, a dash seems to be common to separate words in options rather than underscores:

```suggestion
      .option('target-version', {
```

scripts/mirror.js
```diff
@@ -584,9 +652,18 @@ const bumpVersion = (data, destination, source, originalData) => {
...
  */
-const doSetFeature = (data, newData, rootPath, browser, source, modify) => {
+const doSetFeature = (
+  data,
+  newData,
+  rootPath,
+  browser,
+  source,
+  modify,
+  targetVersion,
```
caugner, line 665: 💡 Refactoring opportunity for the future: Functions with this many parameters could be refactored to accept an object instead. There's probably an ESLint rule to enforce this. 🤔

scripts/mirror.js
```diff
@@ -574,6 +631,17 @@ const bumpVersion = (data, destination, source, originalData) => {
     newData = bumpFunction(originalData, data, source);
   }
 
+  if (targetVersion) {
+    if (
+      !(
+        isVersionAdded(newData, targetVersion) ||
+        isVersionRemoved(newData, targetVersion)
+      )
+    ) {
```
caugner, line 640: Wondering whether this might be more readable:

```suggestion
    if (
        !isVersionAdded(newData, targetVersion) &&
        !isVersionRemoved(newData, targetVersion)
    ) {
```

caugner (changes requested):
LGTM, just two and a half nits or so.

PR #6172, opened 2020-05-15
Added `transform-origin`

jmswisher:
- New in Fx 77, per https://bugzil.la/1581691
- Description of that bug says both Blink and Webkit support it.


svg/attributes/presentation.json
```diff
@@ -2882,6 +2882,54 @@
             }
           }
         },
+        "transform-origin": {
+          "__compat": {
+            "mdn_url": "https://developer.mozilla.org/docs/Web/SVG/Attribute/transform-origin",
+            "support": {
+              "chrome": {
+                "version_added": "yes"
```
vinyldarkscratch, line 2890: Looks like you were trying to use `"yes"` instead of `true`.  😛

jmswisher:
Nevermind. I'm not going to chase down versions for other browsers, given that data for SVG are not displayed on MDN.

vinyldarkscratch:
We can chase down the exact versions another time!  Actually what I was trying to say what that if you're not certain what the version is, then we should put `true` as the value.  `"yes"` was the older syntax.

younggun0509:
Thank you for assisting me with this

And what should I do?

Thank you so much
Nizar

PR #6176, opened 2020-05-18
summary tag - set version_added to true for safari on ios

ynn1k:
works fine on an iPhone SE 2020 with iOS 13.4.1


ynn1k:
You were right, I've made my changes in the wrong line! I made another commit.

PR #6179, opened 2020-05-18
updating firefox support data for :is(); adding support data for :where()

chrisdavidmills:
As per https://bugzilla.mozilla.org/show_bug.cgi?id=1509418

This is supported in Fx 77+, but preffed off in all versions except Nightly.

I couldn't find any evidence of support for :where() in any other browsers.


chrisdavidmills:
@Elchi3 updated!

PR #6186, opened 2020-05-19
Add desktop value for autocapitalize.

jpmedley:
I confirmed with the feature owner that this is not yet available on desktop. 

The interface is visible on desktop, but doesn't do anything. I don't remember how we handled this before.


ddbeck:
Making a note here in case anyone else was thinking something similar: I thought [our guidelines suggest partial_implementation](https://github.com/mdn/browser-compat-data/blob/master/docs/data-guidelines.md#non-functional-defined-names-imply-partial_implementation), but reading the background it's clear that would be the case if we had a version number already. In this case we don't, so false is fine.

PR #6187, opened 2020-05-19
updating compat data for pageshow and pagehide events

chrisdavidmills:
As per https://github.com/mdn/sprints/issues/3141.

TL;DR is that the compat data for the `PageTransitionEvent` object looks right, but the related `pageshow` and `pagehide` event pages have incomplete compat data.

To fix this, I've just updated the compat data for the latter two, to match the compat data for the former.

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -6044,46 +6044,96 @@
           "description": "<code>pagehide</code> event",
           "support": {
             "chrome": {
-              "version_added": null
+              "version_added": true
             },
             "chrome_android": {
-              "version_added": null
+              "version_added": true
             },
             "edge": {
-              "version_added": null
+              "version_added": "12"
             },
             "firefox": {
-              "version_added": "1.5"
+              "version_added": true
             },
             "firefox_android": {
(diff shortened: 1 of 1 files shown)


api/Window.json
```diff
@@ -6044,47 +6044,97 @@
...
+              "version_added": true
             }
           },
           "status": {
             "experimental": false,
             "standard_track": true,
             "deprecated": false
           }
+        },
+        "persisted": {
```
ddbeck, line 6089: I don't think we typically duplicate event object properties in `Window.json` like this. Since `PageTransitionEvent` has this feature already, I think we can remove this feature.

api/Window.json
```diff
@@ -6093,47 +6143,97 @@
...
```
ddbeck, line 6188: Like before, I don't think we typically duplicate event object properties in `Window.json` like this. I think we can remove this feature.

chrisdavidmills: Good call @ddbeck - I've removed the `persisted` event object property data from here.

PR #6193, opened 2020-05-20
IE11: mark Intl.NumberFormat support as partial

wil93:
I was trying to set the option `style` to "unit" in order to use Intl.NumberFormat to format kilobytes / megabytes / and so on.

I got this error on IE11:

Screenshot 2020-05-20 at 19 13 05

It would seem that the only allowed values for the `style` option are:
- decimal
- percent
- currency

### DIFF
--- a/javascript/builtins/intl/NumberFormat.json
+++ b/javascript/builtins/intl/NumberFormat.json
@@ -77,7 +77,7 @@
                 "ie": {
                   "version_added": "11",
                   "partial_implementation": true,
-                  "notes": "The `style` option only supports the values: 'decimal', 'percent', and 'currency'."
+                  "notes": "The <code>style</code> option only supports the values <code>'decimal'</code>, <code>'percent'</code>, and <code>'currency'</code>."
                 },
                 "nodejs": {
                   "version_added": true
--- a/javascript/builtins/intl/NumberFormat.json
+++ b/javascript/builtins/intl/NumberFormat.json
@@ -23,9 +23,7 @@
                 "version_added": "56"
               },
               "ie": {
-                "version_added": "11",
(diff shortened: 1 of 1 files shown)


ddbeck:
Hi @wil93 and welcome to the BCD project. Thanks for opening this PR.

Is it possible you can share the code you're using which results in the `RangeError` in your screenshot? Really, I'd like to know whether this error happens when you call the constructor (e.g., `new Intl.NumberFormat("…", { "style": "unit", … })`) or if it's happening when you call an instance method.

I suspect the `"partial_implementation"` flag and note should move to the constructor subfeature, but I wanted to double check. Again, thank you!

wil93:
Hi, it seems to happen in the constructor. If I test it like this, it only shows the first console.log:

```js
  console.log('Just before instancing class');
  const instance = new Intl.NumberFormat('it', {
    style: 'unit',
    unit: 'byte',
    unitDisplay: 'long',
  });
  console.log('Right after instancing class');
  const retval = instance.format(12345);
  console.log('Right after calling method');
  return retval;
```

javascript/builtins/intl/NumberFormat.json
```diff
@@ -75,7 +75,9 @@
                   "version_added": "56"
                 },
                 "ie": {
-                  "version_added": "11"
+                  "version_added": "11",
+                  "partial_implementation": true,
+                  "notes": "The `style` option only supports the values: 'decimal', 'percent', and 'currency'."
```
ddbeck, line 80:
```suggestion
                  "notes": "The <code>style</code> option only supports the values <code>'decimal'</code>, <code>'percent'</code>, and <code>'currency'</code>."
```

PR #6209, opened 2020-05-26
Update Safari compat data for ::marker

torelafes:
Added data for Safari Desktop
Tested through Safari 10.1 to 13.0
Safari 10.1 (macOS Sierra)
Safari 11.1 (macOS High Sierra)
Safari 12 (macOS Mojave)
Safari 13 (macOS Catalina)

### DIFF
--- a/css/selectors/marker.json
+++ b/css/selectors/marker.json
@@ -49,7 +49,7 @@
               "version_added": false
             },
             "safari": {
-              "version_added": false
+              "version_added": "11.1"
             },
             "safari_ios": {
               "version_added": false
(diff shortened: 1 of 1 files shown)


css/selectors/marker.json
```diff
@@ -49,7 +49,7 @@
               "version_added": false
             },
             "safari": {
-              "version_added": false
+              "version_added": "11.1"
             },
             "safari_ios": {
               "version_added": false
```
ddbeck, line 55: Since we're here, I did a little digging and found the changeset for this: https://trac.webkit.org/changeset/220207/webkit That maps to Safari for iOS 11.3:

```suggestion
              "version_added": "11.3"
```

PR #6221, opened 2020-05-28
Add Chromium numbers for pagehide and pageshow.

jpmedley:
Addition to #6187. 

Here's the commit that landed these: https://source.chromium.org/chromium/chromium/src/+/b74de5a09c2a9e851cc0ac1a02f7cd963cd39a03?originalUrl=https%2F:%2F%2F%2Fcs.chromium.org%2F

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -6107,7 +6107,7 @@
               "version_added": true
             },
             "webview_android": {
-              "version_added": "37"
+              "version_added": "≤37"
             }
           },
           "status": {
@@ -6156,7 +6156,7 @@
               "version_added": true
             },
             "webview_android": {
-              "version_added": "37"
+              "version_added": "≤37"
             }
           },
           "status": {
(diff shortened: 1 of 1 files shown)


api/Window.json
```diff
@@ -6107,7 +6107,7 @@
               "version_added": true
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "37"
```
ddbeck, line 6110: Thank you for this PR, Joe. Are these definitely 37, not `"≤37"`? The linked commit didn't help me answer this question

jpmedley: I know it was in Chromium WebView for sure starting in 37. I don't know anything about non-Chromium WebView. I'm rusty what covers that?

vinyldarkscratch: I think that `"≤37"` would probably be more appropriate, since we're not sure when it was implemented and that it _could_ be in pre-Chromium WebView?  Considering the age of the feature, I'd be surprised if it was only just implemented in WebView 37, and not a version like WebView for Android 4.4 (Blink 30).

jpmedley:
Fixed. Please take another look.

PR #5591, opened 2020-01-28
Enforce specific HTML elements and single spaces in notes

queengooborg:
This PR introduces a brand new linter that enforces the HTML elements within the notes.  A while ago, #4259 was opened that presents a regex-based solution, however it was closed as we wanted to use a parser-based solution instead.  That's what this PR does.  This uses the `html-validate` package to validate the HTML and ensure there's no broken HTML first.  Then, this uses the `node-html-parser` package to parse the HTML and check to ensure the elements within the notes are only one of the five allowed element types.  This also checks the attributes on tags, and ensures that `<a>` tags only have an `href` attribute, while all other tags have no attributes.  Finally, this enforces single-spacing in notes for consistency.

This PR fixes #2799, fixes #3985, fixes #4033, and fixes #4829.  ~~This linter will fail for the time being, as we currently have conflicting data.~~

PR #6223, opened 2020-05-31
Remove double-space in api.Location.reload Edge notes

vinyldarkscratch:
Preliminary work for #5591.

### DIFF
--- a/api/Location.json
+++ b/api/Location.json
@@ -596,7 +596,7 @@
             },
             "edge": {
               "version_added": "12",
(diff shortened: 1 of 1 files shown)


ddbeck:
Hey Vinyl, the proper thing is to rewrite the note to avoid this two-sentence EdgeHTML/Chromium thing. But this PR alerted me to several instances of this problem, so I've gone and cleaned them all up with #6228, which supersedes this PR. Thanks for bringing this to my attention!

PR #2863, opened 2018-09-23
Add AudioTrackList and VideoTrackList

sideshowbarker:
Fixes https://github.com/mdn/browser-compat-data/issues/2855

PR #6224, opened 2020-06-01
Update AudioTrack and VideoTrack support data

sideshowbarker:
This change updates the support data for `AudioTrack` and `VideoTrack`. The data was confirmed by doing manual testing and by checking browser-engine change/revision histories. It mirrors data from https://github.com/mdn/browser-compat-data/pull/2863 (https://github.com/mdn/browser-compat-data/commit/b378e47) for `AudioTrackList` and `VideoTrackList` (when we first added those).

### DIFF
--- a/api/AudioTrack.json
+++ b/api/AudioTrack.json
@@ -26,9 +26,6 @@
           },
           "edge": [
             {
-              "version_added": "12"
-            },
-            {
               "version_added": "79",
               "flags": [
                 {
@@ -37,6 +34,10 @@
                   "value_to_set": "enabled"
                 }
               ]
+            },
+            {
+              "version_added": "12",
+              "version_removed": "79"
             }
           ],
           "firefox": {
@@ -127,9 +128,6 @@
             },
             "edge": [
               {
-                "version_added": "12"
-              },
-              {
                 "version_added": "79",
                 "flags": [
                   {
@@ -138,6 +136,10 @@
                     "value_to_set": "enabled"
                   }
                 ]
+              },
+              {
+                "version_added": "12",
+                "version_removed": "79"
               }
             ],
             "firefox": {
@@ -229,9 +231,6 @@
             },
             "edge": [
               {
-                "version_added": "12"
-              },
-              {
                 "version_added": "79",
                 "flags": [
                   {
@@ -240,6 +239,10 @@
                     "value_to_set": "enabled"
                   }
                 ]
+              },
+              {
+                "version_added": "12",
+                "version_removed": "79"
               }
             ],
             "firefox": {
@@ -331,9 +334,6 @@
             },
             "edge": [
               {
-                "version_added": "12"
-              },
-              {
                 "version_added": "79",
                 "flags": [
                   {
@@ -342,6 +342,10 @@
                     "value_to_set": "enabled"
                   }
                 ]
+              },
+              {
+                "version_added": "12",
+                "version_removed": "79"
               }
             ],
             "firefox": {
@@ -433,9 +437,6 @@
             },
             "edge": [
               {
-                "version_added": "12"
-              },
-              {
                 "version_added": "79",
                 "flags": [
                   {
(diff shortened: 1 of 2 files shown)


ddbeck:
I haven't had a chance to give this a full review yet, but it looks like https://github.com/mdn/browser-compat-data/pull/6225#pullrequestreview-421901818 will also apply here. Thank you!

sideshowbarker:
> I haven't had a chance to give this a full review yet, but it looks like [#6225 (review)](https://github.com/mdn/browser-compat-data/pull/6225#pullrequestreview-421901818) will also apply here.

Yes — thanks, made that change

sideshowbarker:
> This applies to Chrome Android and Android WebView.

Can you clarify what exactly you mean applies to Chrome Android and Android WebView? What should I change?

> Please change. https://www.chromestatus.com/features/5748496434987008

When I open https://www.chromestatus.com/features/5748496434987008 it doesn’t see to show any feature related to this PR… It shows the feature *'noopener' should always disown the opener* at the top of a list of *no active development* features

jpmedley:
> Can you clarify what exactly you mean applies to Chrome Android and Android WebView? What should I change?

`"chrome_android"` and `"webview_android"` should be identical to `"chrome"`

> When I open https://www.chromestatus.com/features/5748496434987008 it doesn’t seem to show any feature related to this PR… It shows the feature _'noopener' should always disown the opener_ at the top of a list of _no active development_ features

Chrome Status isn't behaving correctly just now. Try this one: https://www.chromestatus.com/feature/5748496434987008

(The difference in the URL is `feature/` vs. `features/`.)

sideshowbarker:
> Chrome Status isn't behaving correctly just now. Try this one: https://www.chromestatus.com/feature/5748496434987008
> 
> (The difference in the URL is `feature/` vs. `features/`.)

Aha — thanks

> `"chrome_android"` and `"webview_android"` should be identical to `"chrome"`

Ah OK — so the `"chrome_android"` entries are already identical. It’s just the diff kind of obscures that. https://github.com/mdn/browser-compat-data/blob/sideshowbarker/AudioTrack-VideoTrack-update/api/AudioTrack.json makes it more clear.

But as far as the `"webview_android"` entries, I already set the identical release number for those, but not the flags info. That’s because when this came up previously in https://github.com/mdn/browser-compat-data/pull/2863#issuecomment-425243882, I thought the resolution was we shouldn’t add flags for `"webview_android"` entries — because it’s not possible to set flags when using a Webview? Or maybe I simply misunderstood, or maybe that’s since changed? (That discussion was from 19 months ago.)

jpmedley:
>But as far as the "webview_android" entries, I already set the identical release number for those, but not the flags info. That’s because when this came up previously in #2863 (comment), I thought the resolution was we shouldn’t add flags for "webview_android" entries — because it’s not possible to set flags when using a Webview?

You are correct. I misspoke because I wasn't thinking about the notes when I made my comment.

(Footnote: Sometimes Android features aren't in WebView. You can find a list [here](https://source.chromium.org/chromium/chromium/src/+/master:android_webview/tools/system_webview_shell/test/data/webexposed/not-webview-exposed.txt?originalUrl=https:%2F%2Fcs.chromium.org%2Fchromium%2Fsrc%2Fandroid_webview%2Ftools%2Fsystem_webview_shell%2Ftest%2Fdata%2Fwebexposed%2Fnot-webview-exposed.txt).)

sideshowbarker:
> > But as far as the "webview_android" entries, I already set the identical release number for those, but not the flags info. That’s because when this came up previously in #2863 (comment), I thought the resolution was we shouldn’t add flags for "webview_android" entries — because it’s not possible to set flags when using a Webview?
> 
> You are correct. I misspoke because I wasn't thinking about the notes when I made my comment.

ah OK — thanks for clearing that up

> (Footnote: Sometimes Android features aren't in WebView. You can find a list [here](https://source.chromium.org/chromium/chromium/src/+/master:android_webview/tools/system_webview_shell/test/data/webexposed/not-webview-exposed.txt?originalUrl=https:%2F%2Fcs.chromium.org%2Fchromium%2Fsrc%2Fandroid_webview%2Ftools%2Fsystem_webview_shell%2Ftest%2Fdata%2Fwebexposed%2Fnot-webview-exposed.txt).)

oh wow! This is really great — thanks for the heads-up about it

PR #6225, opened 2020-06-01
Update (Audio|Video)TrackList Edge support data

sideshowbarker:
This change updates that Edge support data for `AudioTrackList` and `VideoTrackList`, to indicate they’re supported in Blink-based Edge 79+ behind the same flag as Chrome, and to retain the indication that they’re supported in legacy Edge since version 12.

### DIFF
--- a/api/AudioTrackList.json
+++ b/api/AudioTrackList.json
@@ -24,9 +24,21 @@
               }
             ]
           },
-          "edge": {
-            "version_added": "12"
-          },
+          "edge": [
+            {
+              "version_added": "12"
+            },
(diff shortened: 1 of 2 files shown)


sideshowbarker:
> … we ought to flip the data and edit a little:
> 
> 1. Added in 79, with flags data
> 2. Added in 12, edited to show removed in 79
> 
> This way the current version is the first entry and the end of the on-by-default status is accurately portrayed.

Thanks — yeah, I had wondered that the best practice was for this kind of case. Now I know :)

PR #6231, opened 2020-06-02
Update Element.attributes IE value

markcarver:
Fixes mistake introduced in #5740

https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/platform-apis/cc288590(v=vs.85)

Also, how would any of us ever have been able to access attributes way back when? LOL

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -1215,7 +1215,7 @@
               "version_added": "26"
             },
             "edge": {
-              "version_added": "16"
+              "version_added": "12"
             },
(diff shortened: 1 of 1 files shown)


vinyldarkscratch:
Happy to review this one!  I’ve found that IE/EdgeHTML have some really weird things with the Element API...but of course, it wouldn’t be IE if it didn’t do weird things.  😛

markcarver:
Done 😄

PR #6234, opened 2020-06-03
adding in compat data for ParentNode querySelector and replaceChildren

chrisdavidmills:
As per https://bugzilla.mozilla.org/show_bug.cgi?id=1626015

I did a bit of testing around, and it doesn't seem to be supported anywhere else yet. I tested in Safari, and Chrome Canary, and it wasn't in either of those, so by extension, none of the other Chromiums will support it either.

I also took this opportunity to add in support for `ParentNode.querySelector`, which should just have the same support as `ParentNode.querySelectorAll`. Any reason why this wasn't added?

### DIFF
--- a/api/ParentNode.json
+++ b/api/ParentNode.json
@@ -632,7 +632,7 @@
               "version_added": "1"
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": "18"
             },
             "edge": {
               "version_added": "12"
@@ -663,7 +663,7 @@
               "version_added": true
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "≤37"
             }
           },
           "status": {
--- a/api/ParentNode.json
+++ b/api/ParentNode.json
@@ -624,6 +624,55 @@
           }
         }
       },
+      "querySelector": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/ParentNode/querySelector",
+          "support": {
+            "chrome": {
+              "version_added": "1"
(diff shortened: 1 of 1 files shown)


ddbeck:
> I also took this opportunity to add in support for `ParentNode.querySelector`, which should just have the same support as `ParentNode.querySelectorAll`. Any reason why this wasn't added?

Maybe because mixins are weird? `Element`, `Document`, and `DocumentFragment` all have `querySelectorAll` data. 🤷‍♂️

api/ParentNode.json
```diff
@@ -624,6 +624,55 @@
...
       },
+      "querySelector": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/ParentNode/querySelector",
+          "support": {
+            "chrome": {
+              "version_added": "1"
+            },
+            "chrome_android": {
+              "version_added": true
```
jpmedley, line 635:
```suggestion
              "version_added": "18"
```

api/ParentNode.json
```diff
@@ -624,6 +624,55 @@
...
+              "version_added": "4"
+            },
+            "safari_ios": {
+              "version_added": "3.2"
+            },
+            "samsunginternet_android": {
+              "version_added": true
+            },
+            "webview_android": {
+              "version_added": true
```
jpmedley, line 666:
```suggestion
              "version_added": "≤37"
```

chrisdavidmills:
@jpmedley thanks Joe! Changes made.

PR #6237, opened 2020-06-04
Update previousSibling IE compatability

### LINKED ISSUE #6236: Node.previousSibling for IE
Currently documented as '?' for IE, but has been available since at least IE8 (I don't have access to any older).

Node.nextSibling is already documented as having support starting in IE8 (#5116).

Edit: Support goes back to at least IE5.5 according to https://quirksmode.org/dom/core/#t74

edg2s:
Fixes #6236

### DIFF
--- a/api/Node.json
+++ b/api/Node.json
@@ -1901,7 +1901,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": "8"
+              "version_added": "5.5"
             },
             "opera": {
               "version_added": true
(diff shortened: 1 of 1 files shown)


api/Node.json
```diff
@@ -1901,7 +1901,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": null
+              "version_added": "8"
```
ddbeck, line 1904:
```suggestion
              "version_added": "5.5"
```

edg2s:
Thanks, I've updated parentNode as well based on https://quirksmode.org/dom/core/#t74

PR #6240, opened 2020-06-04
Add Opera support for location.hostname

### LINKED ISSUE #6239: Location.hostname for Opera
Currently it is claimed there is no support for this in Opera. I am able to test back to version 11, but I suspect it should be version 3 like the other properties of the Location API.

edg2s:
Fixes #6239

### DIFF
--- a/api/Location.json
+++ b/api/Location.json
@@ -262,7 +262,7 @@
               "version_added": true
             },
             "opera": {
-              "version_added": false
+              "version_added": "11"
             },
             "opera_android": {
               "version_added": false


ddbeck:
@edg2s Thank you for this PR. Since you mentioned in #6239 that you were only able test back to version 11, I think the correct value here would be `true` (especially if you suspect it's older still). With that in mind, I plan to merge #6286. But if you would like to rebase and see if you can find evidence of a definitive version, then I'd happily reopen this PR or review a new PR. Thank you again!

PR #6253, opened 2020-06-06
Move Intl.Collator caseFirst option under the constructor

gilmoreorless:
This is a follow-up to the discussion in #6202. There was an inconsistency in the data structure regarding options to `Intl` object constructors. The options for `Intl.DateTimeFormat()` are nested under the constructor, but the option for `Intl.Collator()` was at the same level as the constructor.

As discussed with @ddbeck, this PR moves the `caseFirst` option so that it's nested underneath the `Intl.Collator()` constructor. This might be considered a breaking change, though.

### DIFF
--- a/javascript/builtins/intl/Collator.json
+++ b/javascript/builtins/intl/Collator.json
@@ -112,56 +112,56 @@
                 "standard_track": true,
                 "deprecated": false
               }
-            }
-          },
-          "caseFirst": {
-            "__compat": {
-              "description": "<code>caseFirst</code> option",
-              "support": {
-                "chrome": {
-                  "version_added": "24"
+            },
+            "caseFirst": {
+              "__compat": {
+                "description": "<code>caseFirst</code> option",
+                "support": {
+                  "chrome": {
+                    "version_added": "24"
+                  },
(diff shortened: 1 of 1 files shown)


gilmoreorless:
Ooh, looks like the tests have picked up other inconsistencies with the pre-existing data for Android, now that it's nested. The current data says that `webview_android` supports an option for a constructor that it doesn't have, and there's a version mismatch for `chrome_android` too.

Now that I look closely, I see the data for the `caseFirst` option is also incorrect for Safari. The data file says it's not supported, but it works when I try it locally. Likewise for Edge.

I think I'll need to do some digging to find out the correct versions before I can fix up this PR.

gilmoreorless:
I've raised #6260 to correct the data. Once (_if_) that's merged, I'll rebase this PR to get it passing the tests. (Sorry, I thought I'd run the tests locally first, but clearly I didn't.)

gilmoreorless:
I've rebased this on top of the fixed data from #6260 so hopefully it's good to go now. I also found some prior work in moving constructor options to be nested underneath the constructor in #5750, so I'm slightly more confident in the change. 😉

PR #6278, opened 2020-06-09
Add CSSPropertyRule.json.

jpmedley:
https://www.chromestatus.com/feature/5640265926705152

### DIFF
--- a/api/CSSPropertyRule.json
+++ b/api/CSSPropertyRule.json
@@ -1,0 +1,244 @@
+{
+  "api": {
+    "CSSPropertyRule": {
+      "__compat": {
+        "mdn_url": "https://developer.mozilla.org/docs/Web/API/CSSPropertyRule",
+        "support": {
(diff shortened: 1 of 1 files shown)


jpmedley:
I was looking at the wrong Chrome Status entry. Sorry about that. It is 78. I've corrected the PR and the link in the description.

foolip:
Due to an apparent discrepancy between Chrome and Edge in https://github.com/mdn/browser-compat-data/pull/9189 I took a second look at this. This interface is for the `@property` rule, and was shipped in Chrome 85:
https://www.chromestatus.com/feature/5193698449031168
https://storage.googleapis.com/chromium-find-releases-static/9ab.html#9abf13eb1926e80fddd30a505c4370499433e236

foolip:
I've sent https://github.com/mdn/browser-compat-data/pull/9201 to fix this.

PR #6291, opened 2020-06-11
Add Chrome file system data.

jpmedley:
* I got the alternative names from the Chrome product manager working on the new Native File System API.
* I cannot confirm that the existing prefix version number are correct, nor can I find a when the prefixes were removed. 
* I know it's suspect that the unprefixed versions are older than the prefixed versions, but I can find nothing that disproves it.

### DIFF
--- a/api/FileSystem.json
+++ b/api/FileSystem.json
@@ -4,28 +4,14 @@
       "__compat": {
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/FileSystem",
         "support": {
-          "chrome": [
-            {
-              "version_removed": true,
-              "version_added": "13",
-              "prefix": "webkit"
-            },
-            {
-              "version_added": "7",
-              "alternative_name": "DOMFileSystem"
-            }
-          ],
-          "chrome_android": [
-            {
-              "version_removed": true,
-              "version_added": "18",
-              "prefix": "webkit"
-            },
-            {
-              "version_added": "18",
-              "alternative_name": "DOMFileSystem"
-            }
-          ],
+          "chrome": {
+            "version_added": "7",
+            "alternative_name": "DOMFileSystem"
+          },
+          "chrome_android": {
+            "version_added": "18",
+            "alternative_name": "DOMFileSystem"
+          },
           "edge": {
             "version_added": "≤18",
             "prefix": "WebKit",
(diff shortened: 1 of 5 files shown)


ddbeck:
Hi Joe, thanks for this PR. A couple things:

First, it looks like there are a bunch of Chrome for Android values which are below the first release. Can you hoist those numbers to 18 or later?

> I cannot confirm that the existing prefix version number are correct, nor can I find a when the prefixes were removed.

Do I understand correctly that the prefixes were removed? If so, I'd prefer to see `"version_removed: true` added to those entries (e.g., with both version added and removed: `"version_added": "123", "version_removed: true`), even if we don't have a specific version.

Wshao62:
Hello hacker, do you have Teshu C vvto sell?  Or data

Daniel D. Beck <notifications@github.com> 於 2020年6月16日週二 00:25 寫道：

jpmedley:
>First, it looks like there are a bunch of Chrome for Android values which are below the first release. Can you hoist those numbers to 18 or later?

Fixed. Sorry about that. 

>Do I understand correctly that the prefixes were removed? If so, I'd prefer to see "version_removed: true added to those entries (e.g., with both version added and removed: "version_added": "123", "version_removed: true), even if we don't have a specific version.

I didn't do this because all I can tell you about the interface members is that the prefixes are not there now. I do not know whether they were ever there. I've seen APIs where the interfaces were prefixed, but its members were not. Without direct evidence, I'm hesitant to do that here. Since the members are available on a non-prefixed interface, calls should work regardless.

ddbeck:
Interesting. Thanks for that detail, @jpmedley. I tried to see if I could dig up any more detail on this.

First, I decided to try out the API in Saucelabs, to see if I could narrow the range at all. In Chrome 26 (the oldest supported) running something like `window.webkitRequestFileSystem(Window.TEMPORARY, 1024, function (fs) { console.log(fs) })` returns a `DOMFileSystem` object (likewise, you can get a `DirectoryEntry` object, not `FileSystemDirectoryEntry`). This is the same in Chrome 84, so the alternative names are genuine and current; this suggests the webkit prefixes have at least been long removed.

Since this predates the Blink fork, I searched webkit bugs and the [webkit source mirror](https://github.com/WebKit/webkit/). I could only turn up evidence of `webkitRequestFileSystem`, but not any evidence of `FileSystem` or `FileSystemDirectoryEntry` prefixes (or prefixes for the alternative names). My guess here is that the `webkit` prefix data here is a phantom.

I suppose my conclusion is this: we ought to drop the `prefix: "webkit"` entries and only keep the unprefixed alternative name entries.

jpmedley:
Done.

PR #6300, opened 2020-06-16
Fix copy/paste errors.

jpmedley:
Correct copy/paste errors introduced when adding Chrome's alternative names for file/directory interfaces.


jpmedley:
@ddbeck Submitted with apologies. 

#6291 had a cut and paste error. Can I ask you to merge this quickly so that it does not end up in Thursday's deployment?

PR #6303, opened 2020-06-16
Note dropped support for "xbl" ResourceType value

wbamberg:
This PR is for the first part of https://bugzilla.mozilla.org/show_bug.cgi?id=1638581#c9, in which we remove support for `"xbl"` as a value for [`webRequest.ResourceType`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/ResourceType).

Rather than mark it as deprecated, I have marked it as unsupported from Firefox 78, which seems to capture the case more accurately (deprecated I would expect to mean "works but complains and will eventually stop working").

### DIFF
--- a/webextensions/api/webRequest.json
+++ b/webextensions/api/webRequest.json
@@ -445,7 +445,8 @@
                   "version_added": false
                 },
                 "firefox": {
-                  "version_added": "45"
+                  "version_added": "45",
+                  "version_removed": "78"
                 },
                 "firefox_android": {
                   "version_added": "48"


wbamberg:
@rpl , please let me know if this looks accurate to you.

rpl:
> @rpl , please let me know if this looks accurate to you.

yeah, sounds good to me. (technically we removed xbl completely in 78 and so there shouldn't be even any webRequest event fired with that resource type, but I'm not sure if it is something actually worth to capture in a note, `version_removed: "78"` could be likely enough).

Issue #6229, opened 2020-06-01
Document style for notes


ddbeck:
There are certain aspects of notes that we can't lint for. We ought to document a lightweight style guide for writing notes. We could probably start by codifying a few common practices:

* Use the present tense (the inspiration for this issue, #6228, plus a few others like #2151 and #505)
* Use browser names and avoid engine names where possible (e.g., "Chrome" not "Chromium", "Firefox" not "Gecko")
* When a note applies only to some subset of versions of a support statement, then begin the note with the limitation (e.g., "Before X", "After Y", "From Z to W")
* Avoid imprecision by specifying versions, not dates or times (e.g., avoid "now", "currently")

There are probably others I'm missing.

We could probably write up the guide and illustrate with a few bad/better/best examples.

PR #6305, opened 2020-06-17
Document behavior when unrecognized resource types are present

wbamberg:
This PR updates BCD for the second part of https://bugzilla.mozilla.org/show_bug.cgi?id=1638581#c9:

> in WebRequest/RequestFilterWebRequest/RequestFilter API doc page on MDN, mention on the types property description that an invalid/unsupported value is going to log a warning starting from Firefox 78 (but it does throw a validation error in Firefox < 78)

I believe the situation is: in Firefox < 78, and other browsers (although I have only tested in Chrome), if you pass unrecognized values in the `types` property of [`RequestFilter`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/WebRequest/RequestFilter), then this is treated as an error and the operation in which you try to use the filter fails.

For example, if you try to redirect web requests with some code like:

```js
browser.webRequest.onBeforeRequest.addListener(
  redirect,
  {urls:[pattern], types:["bibble", "image"]},
  ["blocking"]
);
```

...this will fail because `"bibble"` is not a valid value for [`ResourceType`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/ResourceType).

But from Firefox 78 onwards, if unrecognized values are found, Firefox will log a warning and just discard the unrecognized values, applying the rest of them.

It's tricky to know how to document things like this. In general, it seems we should document the standard behavior in the main MDN page, and document browser-specific deviations from that in BCD. But with WebExtensions, there is no standard, so which do we choose as the standard behavior? I've chosen to describe the new Firefox behavior as the standard here, because Firefox is really the main driver of the WebExtension APIs.

The BCD for Firefox in this PR will render like:

Screen Shot 2020-06-16 at 5 13 28 PM

@rpl , please let me know if you think this is good, and if you do I'll update the [`RequestFilter`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/WebRequest/RequestFilter) page to describe the "standard" behavior (logging a warning and discarding unrecognized types).

### DIFF
--- a/webextensions/api/webRequest.json
+++ b/webextensions/api/webRequest.json
@@ -127,7 +127,7 @@
               "firefox": {
                 "notes": [
                   "From Firefox 78 onwards, if a filter contains unrecognized values in its <code>types</code> property, then these values will be ignored and <code>addListener()</code> will proceed.",
-                  "Before Firefox 78, if a filter contained unrecognized values in its <code>types</code> property, <code>addListener()</code> threw an exception."
+                  "Before Firefox 78, if a filter contains unrecognized values in its <code>types</code> property, <code>addListener()</code> throws an exception."
                 ],
                 "version_added": "45"
               },
--- a/webextensions/api/webRequest.json
+++ b/webextensions/api/webRequest.json
@@ -117,18 +117,29 @@
             "mdn_url": "https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/RequestFilter",
             "support": {
               "chrome": {
+                "notes": "If a filter contains unrecognized values in its <code>types</code> property, this is treated as an error.",
                 "version_added": true
               },
               "edge": {
+                "notes": "If a filter contains unrecognized values in its <code>types</code> property, this is treated as an error.",
                 "version_added": "14"
               },
-              "firefox": {
-                "version_added": "45"
-              },
+              "firefox": [
+                {
+                  "version_added": "45"
+                },
+                {
+                  "notes": "If a filter contains unrecognized values in its <code>types</code> property, this is treated as an error.",
+                  "version_added": "45",
+                  "version_removed": "78"
+                }
+              ],
               "firefox_android": {
+                "notes": "If a filter contains unrecognized values in its <code>types</code> property, this is treated as an error.",
                 "version_added": "48"
               },
               "opera": {
+                "notes": "If a filter contains unrecognized values in its <code>types</code> property, this is treated as an error.",
                 "version_added": true
               }
             }
--- a/webextensions/api/webRequest.json
+++ b/webextensions/api/webRequest.json
@@ -117,7 +117,7 @@
             "mdn_url": "https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/RequestFilter",
             "support": {
               "chrome": {
-                "notes": "If a filter contains unrecognized values in its <code>types</code> property, <code>addListener()</code> will throw an exception.",
+                "notes": "If a filter contains unrecognized values in its <code>types</code> property, <code>addListener()</code> throws an exception.",
                 "version_added": true
               },
               "edge": {
--- a/webextensions/api/webRequest.json
+++ b/webextensions/api/webRequest.json
@@ -121,7 +121,7 @@
                 "version_added": true
               },
               "edge": {
-                "notes": "If a filter contains unrecognized values in its <code>types</code> property, <code>addListener()</code> will throw an exception.",
+                "notes": "If a filter contains unrecognized values in its <code>types</code> property, <code>addListener()</code> throws an exception.",
                 "version_added": "14"
               },
               "firefox": {
--- a/webextensions/api/webRequest.json
+++ b/webextensions/api/webRequest.json
@@ -117,29 +117,26 @@
             "mdn_url": "https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/RequestFilter",
             "support": {
               "chrome": {
-                "notes": "If a filter contains unrecognized values in its <code>types</code> property, this is treated as an error.",
+                "notes": "If a filter contains unrecognized values in its <code>types</code> property, <code>addListener()</code> will throw an exception.",
                 "version_added": true
               },
               "edge": {
-                "notes": "If a filter contains unrecognized values in its <code>types</code> property, this is treated as an error.",
+                "notes": "If a filter contains unrecognized values in its <code>types</code> property, <code>addListener()</code> will throw an exception.",
                 "version_added": "14"
               },
-              "firefox": [
-                {
-                  "version_added": "45"
-                },
-                {
-                  "notes": "If a filter contains unrecognized values in its <code>types</code> property, this is treated as an error.",
-                  "version_added": "45",
-                  "version_removed": "78"
-                }
-              ],
+              "firefox": {
+                "notes": [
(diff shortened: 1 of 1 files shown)


webextensions/api/webRequest.json
```diff
@@ -117,18 +117,29 @@
...
-                "version_added": "45"
-              },
+              "firefox": [
+                {
+                  "version_added": "45"
+                },
+                {
+                  "notes": "If a filter contains unrecognized values in its <code>types</code> property, this is treated as an error.",
+                  "version_added": "45",
+                  "version_removed": "78"
```
rpl, line 134: @wbamberg`version_removed` makes me think that ResourceType isn't supported anymore in 78, like if we removed it completely, but it is still supported in "Firefox > 78" and it just changed its behavior in case of unsupported values.

is there a way to express this without making it looks like we removed it in that version?

I see that in the userScripts.json to express that the API was originally locked behind a pref we do use to version_added entries one with a flags property and one without it: 
https://github.com/mdn/browser-compat-data/blob/790eeb364cf5f19e589830b5242547890fbf7039/webextensions/api/userScripts.json#L16-L28

and so I'm wondering if it should be ok to use to version_added entries, and one with version_added and the note?
To be fair this change sounds more like a "version_changed" entry, but I think that a "version_changed" property is not in the schema we use for this metadata, at least not at the moment.

webextensions/api/webRequest.json
```diff
@@ -117,18 +117,29 @@
             "mdn_url": "https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/RequestFilter",
             "support": {
               "chrome": {
+                "notes": "If a filter contains unrecognized values in its <code>types</code> property, this is treated as an error.",
```
ddbeck, line 120: I'm sorry I'm familiar enough with this API to provide a more direct suggestion, but can we be more explicit about what "treated as an error" means? Something like

* If a filter contains unrecognized values in its <code>types</code> property, then the callback throws an exception.
* …, then the `addListener` method returns an error value.

Or whatever makes sense in this case.

webextensions/api/webRequest.json
```diff
@@ -117,18 +117,29 @@
...
               "edge": {
+                "notes": "If a filter contains unrecognized values in its <code>types</code> property, this is treated as an error.",
                 "version_added": "14"
               },
-              "firefox": {
-                "version_added": "45"
-              },
+              "firefox": [
+                {
+                  "version_added": "45"
```
ddbeck, line 129: Yeah, structuring this is a bit challenging. In this case, because there's no real standard and Firefox's new behavior seems to have gotten more permissive, I'm inclined to stuff this into notes. Maybe two notes, like this:

1. `From Firefox 78, unrecognized <code>types</code> values are discarded with a warning.`
1. `Before Firefox 78, if a filter contains unrecognized values…`

webextensions/api/webRequest.json
```diff
@@ -117,18 +117,29 @@
...
+                {
+                  "notes": "If a filter contains unrecognized values in its <code>types</code> property, this is treated as an error.",
+                  "version_added": "45",
+                  "version_removed": "78"
+                }
+              ],
               "firefox_android": {
+                "notes": "If a filter contains unrecognized values in its <code>types</code> property, this is treated as an error.",
                 "version_added": "48"
               },
```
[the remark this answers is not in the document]
wbamberg: @rpl , only Fennec AFAIK. I've been out of the WebExtensions world for a little while though, I don't know what the plans are for documenting GeckoView WebExtension support. Good question.

wbamberg:
> version_removed makes me think that ResourceType isn't supported anymore in 78

As in the screenshot I posted above, I think the table makes it clear enough that this feature is still supported in Firefox.

I think there's a question I'm not sure of about BCD. I like to use `version_added` and `version_removed` like this, to indicate version ranges in which some aspect of a feature is applicable (like a pref, or just some behavior change like this). I like this because it's more structured than sticking it all in notes. But I'm not sure if this is considered a legitimate thing to do, so I've removed it from this PR and stuck everything in notes. 

> can we be more explicit about what "treated as an error" means?

Yes, it means `addListener()` will throw. I've tried to clarify this.

webextensions/api/webRequest.json
```diff
@@ -117,18 +117,26 @@
             "mdn_url": "https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/RequestFilter",
             "support": {
               "chrome": {
+                "notes": "If a filter contains unrecognized values in its <code>types</code> property, <code>addListener()</code> will throw an exception.",
```
ddbeck, line 120: It's sort of an unwritten rule (see #6229) but I've been steering people toward all notes being written in the present tense (for reviewability and maintainability)

```suggestion
                "notes": "If a filter contains unrecognized values in its <code>types</code> property, <code>addListener()</code> throws an exception.",
```

webextensions/api/webRequest.json
```diff
@@ -117,18 +117,26 @@
...
                 "version_added": true
               },
               "edge": {
+                "notes": "If a filter contains unrecognized values in its <code>types</code> property, <code>addListener()</code> will throw an exception.",
```
ddbeck, line 124:
```suggestion
                "notes": "If a filter contains unrecognized values in its <code>types</code> property, <code>addListener()</code> throws an exception.",
```

webextensions/api/webRequest.json
```diff
@@ -117,18 +117,26 @@
...
                 "version_added": "14"
               },
               "firefox": {
+                "notes": [
+                  "From Firefox 78 onwards, if a filter contains unrecognized values in its <code>types</code> property, then these values will be ignored and <code>addListener()</code> will proceed.",
```
ddbeck, line 129:
```suggestion
                  "From Firefox 78 onwards, if a filter contains unrecognized values in its <code>types</code> property, then these values are ignored and <code>addListener()</code> proceeds.",
```

webextensions/api/webRequest.json
```diff
@@ -117,18 +117,26 @@
...
+                  "Before Firefox 78, if a filter contained unrecognized values in its <code>types</code> property, <code>addListener()</code> threw an exception."
```
ddbeck, line 130:
```suggestion
                  "Before Firefox 78, if a filter contains unrecognized values in its <code>types</code> property, <code>addListener()</code> throws an exception."
```

webextensions/api/webRequest.json
```diff
@@ -117,18 +117,26 @@
...
+                ],
                 "version_added": "45"
               },
               "firefox_android": {
+                "notes": "If a filter contains unrecognized values in its <code>types</code> property, <code>addListener()</code> will throw an exception.",
```
ddbeck, line 135:
```suggestion
                "notes": "If a filter contains unrecognized values in its <code>types</code> property, <code>addListener()</code> throws an exception.",
```

webextensions/api/webRequest.json
```diff
@@ -117,18 +117,26 @@
...
                 "version_added": "48"
               },
               "opera": {
+                "notes": "If a filter contains unrecognized values in its <code>types</code> property, <code>addListener()</code> will throw an exception.",
```
ddbeck, line 139:
```suggestion
                "notes": "If a filter contains unrecognized values in its <code>types</code> property, <code>addListener()</code> throws an exception.",
```

PR #6314, opened 2020-06-18
Drop now-nonexistent Streams API constructors

sideshowbarker:
The constructors below are not in the current Streams spec.

* `ReadableByteStreamController()`
* `ReadableStreamBYOBRequest()`
* `ReadableStreamDefaultController()`
* `WritableStreamDefaultController()`

See https://github.com/whatwg/streams/issues/1043 for confirmation from one of the spec editors that the constructors should be de-documented.

All of the corresponding MDN articles have been deleted.

The BCD data showed that none of them had ever been implemented in any browser engines — with the exception of `ReadableStreamDefaultController()`, which the data indicated as being implemented in Firefox. But I think that may have been incorrect — because I can’t find any evidence that constructor was in fact ever actually implemented in Firefox.


ddbeck:
I noticed this needed to be labeled for release notes, but then realized that it landed in v1.0.27, so I updated the release notes for that too.

PR #6316, opened 2020-06-20
fix navigator.registerProtocolHandler mobile browser support

Kukks:
I tested `navigator.registerProtocolHandler` on Android against Chrome(83.0.4103.106) and Firefox (68.9.0) and neither worked.

I used some simple JS to test(https://github.com/Kukks/kukks.org/blob/648fd5dc97e348185cd762c7d104c638a5053c07/test.html):
Chrome reported that `navigator.registerProtocolHandler` is `undefined`
Firefox had the function set, but upon calling it to set a  `bitcoin:` scheme as per [MDN](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/registerProtocolHandler), nothing happens: (no exception thrown, no response given) 

CanIUse also reports that android support is absent: https://caniuse.com/#feat=registerprotocolhandler

### DIFF
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -1799,7 +1799,7 @@
               "notes": "Allowed schemes include <code>mailto</code>, <code>mms</code>, <code>nntp</code>, <code>rtsp</code>, and <code>webcal</code>. Custom protocols must be prefixed with <code>web+</code>."
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": false
             },
             "edge": {
               "version_added": "≤79",
@@ -1809,7 +1809,7 @@
               "version_added": "3"
             },
             "firefox_android": {
-              "version_added": "4"
+              "version_added": false
             },
             "ie": {
               "version_added": null
@@ -1847,7 +1847,7 @@
                 "version_added": "80"
               },
               "chrome_android": {
-                "version_added": "80"
+                "version_added": false
               },
               "edge": {
                 "version_added": "≤79"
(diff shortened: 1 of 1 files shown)


api/Navigator.json
```diff
@@ -1799,7 +1799,7 @@
               "notes": "Allowed schemes include <code>mailto</code>, <code>mms</code>, <code>nntp</code>, <code>rtsp</code>, and <code>webcal</code>. Custom protocols must be prefixed with <code>web+</code>."
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": false
```
jpmedley, line 1802: Support of this feature in Chrome is a little more complicated.

https://www.chromestatus.com/features#registerProtocolHandler

I can help you work through this. Please start by mining the data in that link.

ddbeck: Can you be more specific about what you mean here? When I tried `registerProtocolHandler` in Chrome for Android, it was `undefined` (plus I found [this open bug](https://bugs.chromium.org/p/chromium/issues/detail?id=178097)).

ddbeck:
Welcome to BCD, @Kukks! This is an interesting PR; thank you for opening it.

api/Navigator.json
jpmedley: I was not aware that it is not supported on Android. On desktop it requires a secure context [since Chrome 80](https://www.chromestatus.com/features/5756636801007616). [Since 77](https://www.chromestatus.com/features/5066079862784000) the URL argument only accepts 'http' or 'https'.

PR #6318, opened 2020-06-20
Chrome 83+ supports CSS selector: :where() behind a flag

Schweinepriester:
Fixes Fyrd/caniuse#5481.

I've tested using <https://jsfiddle.net/simevidas/d1tLszx9/> from <https://bugs.chromium.org/p/chromium/issues/detail?id=808904#c7>:

image

Chromestatus: https://www.chromestatus.com/feature/4600991135563776

### DIFF
--- a/css/selectors/where.json
+++ b/css/selectors/where.json
@@ -16,7 +16,7 @@
                     "value_to_set": "enabled"
                   }
                 ]
-              },
+              }
             ],
             "chrome_android": {
               "version_added": false
--- a/css/selectors/where.json
+++ b/css/selectors/where.json
@@ -11,7 +11,7 @@
               "flags": [
                 {
                   "type": "preference",
-                  "name": "Experimental Web Platform Features",
+                  "name": "#enable-experimental-web-platform-features",
                   "value_to_set": "enabled"
                 }
               ]
--- a/css/selectors/where.json
+++ b/css/selectors/where.json
@@ -17,7 +17,14 @@
               ]
             },
             "chrome_android": {
-              "version_added": false
+              "version_added": "72",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "#enable-experimental-web-platform-features",
+                  "value_to_set": "Enabled"
+                }
+              ]
             },
             "edge": {
               "version_added": false
--- a/css/selectors/where.json
+++ b/css/selectors/where.json
@@ -6,9 +6,18 @@
           "description": "<code>:where()</code>",
           "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/:where",
           "support": {
-            "chrome": {
-              "version_added": false
-            },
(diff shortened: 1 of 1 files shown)


css/selectors/where.json
```diff
@@ -7,7 +7,14 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/:where",
           "support": {
             "chrome": {
-              "version_added": false
+              "version_added": "83",
```
ddbeck, line 10: This feature seems to be somewhat older than your test shows. It looks like [this change](https://storage.googleapis.com/chromium-find-releases-static/e4a.html#e4a8dda8926d36a02e3f3bc3d20d839bb71ede5c) renamed it from `is()` to `:where()` in Chrome 72.

```suggestion
              "version_added": "72",
```

css/selectors/where.json
```diff
@@ -7,7 +7,14 @@
...
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "Experimental Web Platform Features",
```
ddbeck, line 14: For consistency, I'd suggest changing this. This hasn't been formalized yet, but see #6009 for discussion.

```suggestion
                  "name": "#enable-experimental-web-platform-features",
```

css/selectors/where.json
```diff
@@ -7,7 +7,14 @@
...
+                  "value_to_set": "enabled"
```
ddbeck, line 15:
```suggestion
                  "value_to_set": "Enabled"
```

css/selectors/where.json
```diff
@@ -7,7 +7,14 @@
...
+                }
+              ]
             },
             "chrome_android": {
               "version_added": false
```
ddbeck, line 20: Can you duplicate the data point above to Chrome for Android? They're nearly always the same

css/selectors/where.json
Schweinepriester: You found the change, nice! Tried to but didnt find it...

Schweinepriester:
Just to double check regarding the inconsistencies, I copied those from `:is()`: https://github.com/mdn/browser-compat-data/blob/master/css/selectors/is.json#L9

Given the tendency of #6009 I should change them? And maybe on top in another PR correct those in `:is()`?

css/selectors/where.json
jpmedley: If they're not, that will be reflected in the Chrome Status entry.

ddbeck:
I thought that, while we're editing the data anyway, it'd be nice to get some more consistency (I had done a little bit of work to fix #6009, but there's no PR yet). You don't have to change the  `flags` data (for this PR or for `:is()`), if you don't want to, but I would like to see the data copied to `chrome_android` before merging. Thank you!

Schweinepriester:
Done I think :)

PR #6322, opened 2020-06-24
Add more accuracy to notes on document.readyState support for IE

Fyrd:
Hey look, my first BCD PR. :)

This is in response to a [caniuse issue](https://github.com/Fyrd/caniuse/issues/5489) which I investigated and came to the conclusion that the data seems ambiguous on which versions the notes refer to. As a result the assumption (on caniuse) is that the the notes refer to these versions and beyond, resulting in incorrect information shown on the [caniuse support table](https://caniuse.com/#feat=mdn-api_document_readystate).

Also took the liberty of flagging IE8's support as partial, which I think matches what the note describes.

Happy to discuss this further if my understanding is incorrect and this does not seem like the correct change to make.

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -9212,11 +9212,17 @@
             },
             "ie": [
               {
+                "version_added": "11"
+              },
+              {
                 "version_added": "9",
+                "version_removed": "11",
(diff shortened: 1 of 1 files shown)


sideshowbarker:
Looks great to me — but will leave it unmerged until EU work hours roll around, to give it a chance for a second opinion (third opinion? 😄)

PR #6339, opened 2020-06-26
Safari 14 support private class fields of Javascript

jimmy-collazos:
Safari Technology Preview 109 add support to private class fields. This is a beta version of Safari 14.

Add notes in Javascript Classes.

Release Note: https://webkit.org/blog/10875/release-notes-for-safari-technology-preview-109-with-safari-14-features/
Change set: https://trac.webkit.org/changeset/262613/webkit/
Test: https://repl.it/@JimmyCollazos/Javascript-private-class-fields#script.js

### DIFF
--- a/javascript/classes.json
+++ b/javascript/classes.json
@@ -507,7 +507,7 @@
               "version_added": "53"
             },
             "safari": {
-              "version_added": false
+              "version_added": "14"
             },
             "safari_ios": {
               "version_added": false
--- a/javascript/classes.json
+++ b/javascript/classes.json
@@ -507,10 +507,12 @@
               "version_added": "53"
             },
             "safari": {
-              "version_added": "14"
+              "version_added": false,
+              "notes": "Added in Safari Technology Preview 109 - Safari Beta 14"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": false,
+              "notes": "Added in Safari Technology Preview 109 - Safari Beta 14"
             },
             "samsunginternet_android": {
               "version_added": false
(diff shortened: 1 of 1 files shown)


jonasjelonek:
This also works on Safari for iOS and iPadOS in the iOS 14 and iPadOS 14 beta versions, should also be included in the update.

jimmy-collazos:
@jonasjelonek you are right; maybe it's much better set to false and just add note; because now is a beta version.

javascript/classes.json
```diff
@@ -507,10 +507,12 @@
               "version_added": "53"
             },
             "safari": {
-              "version_added": false
+              "version_added": false,
```
ddbeck, line 510: BCD accepts Safari 14 as a value and it's tracked as a beta version: https://github.com/mdn/browser-compat-data/blob/master/browsers/safari.json#L170. You can remove the note and set `"version_added": "14"`

javascript/classes.json
```diff
@@ -507,10 +507,12 @@
...
+              "notes": "Added in Safari Technology Preview 109 - Safari Beta 14"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": false,
```
ddbeck, line 514: Likewise for Safari for iOS (which has slightly different versioning, historically, though in this case the version numbers are the same) https://github.com/mdn/browser-compat-data/blob/master/browsers/safari_ios.json#L161

jimmy-collazos:
@ddbeck  Thanks for the info; I didn't know that file existed.

Updated!

PR #6355, opened 2020-06-30
Move feature description links to mdn_url

gsouquet:
The compat data schema (`schemas/compat-data-schema.md`) specifies the use of `<code>`, `<kbd>`, `<em>`, and `<strong>` HTML elements only.

A few `description` that contain `<a>` HTML elements lingered. I have moved all internal link to the preferred `mdn_url` entry.

However there are a few entries that reference external documentation. Would appreciate some guidance on how to change those. They are definitely relevant information but they probably should live inside MDN. The `notes` entry is not really applicable in this use case AFAIK

Below is a comprehensive list of all references to external documentations from a `description` entry

- `css/properties/display.json` references https://drafts.csswg.org/css-display/#unbox
- `css/selectors/visited.json` references https://groups.google.com/forum/#!msg/mozilla.dev.platform/1NP6oJzK6zg/ftAz_TajAAAJ
- `javascript/builtins/Function.json` references http://tc39.es/Function-prototype-toString-revision/
- `javascript/builtins/JSON.json` references https://github.com/tc39/proposal-well-formed-stringify

### DIFF
--- a/api/Animation.json
+++ b/api/Animation.json
@@ -1339,7 +1339,7 @@
       },
       "remove_filling_animation": {
         "__compat": {
-          "description": "Browsers automatically remove <a href='https://developer.mozilla.org/docs/Web/API/Animation#Automatically_removing_filling_animations'>indefinite filling animations</a>.",
+          "description": "Browsers automatically remove indefinite filling animations.",
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Animation#Automatically_removing_filling_animations",
           "support": {
             "chrome": {
--- a/css/selectors/invalid.json
+++ b/css/selectors/invalid.json
@@ -51,7 +51,8 @@
         },
         "form": {
           "__compat": {
-            "description": "Applies to <a href='https://developer.mozilla.org/docs/Web/HTML/Element/form'>&lt;form&gt;</a> elements",
+            "mdn_url": "https://developer.mozilla.org/docs/Web/HTML/Element/form",
+            "description": "Applies to <code>&lt;form&gt;</code> elements",
             "support": {
               "chrome": {
                 "version_added": "40"
--- a/css/selectors/valid.json
+++ b/css/selectors/valid.json
@@ -51,7 +51,8 @@
         },
         "form": {
           "__compat": {
-            "description": "Applies to <a href='https://developer.mozilla.org/docs/Web/HTML/Element/form'>&lt;form&gt;</a> elements",
+            "mdn_url": "https://developer.mozilla.org/docs/Web/HTML/Element/form",
+            "description": "Applies to <code>&lt;form&gt;</code> elements",
             "support": {
               "chrome": {
                 "version_added": "40"
--- a/api/CharacterData.json
+++ b/api/CharacterData.json
@@ -49,7 +49,8 @@
       },
       "ChildNode": {
         "__compat": {
-          "description": "Implements the <a href='https://developer.mozilla.org/docs/Web/API/ChildNode'><code>ChildNode</code></a> interface",
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/ChildNode",
+          "description": "Implements the <code>ChildNode</code> interface",
           "support": {
             "chrome": {
               "version_added": true
--- a/api/ShadowRoot.json
+++ b/api/ShadowRoot.json
@@ -123,7 +123,8 @@
       },
       "documentorshadowroot": {
         "__compat": {
-          "description": "Features included from the <a href='https://developer.mozilla.org/docs/Web/API/DocumentOrShadowRoot'><code>DocumentOrShadowRoot</code></a> mixin",
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/DocumentOrShadowRoot",
+          "description": "Features included from the <code>DocumentOrShadowRoot</code> mixin",
           "support": {
             "chrome": {
               "version_added": "57"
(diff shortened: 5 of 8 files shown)


api/Animation.json
```diff
@@ -1339,7 +1339,7 @@
       },
       "remove_filling_animation": {
         "__compat": {
-          "description": "Browsers automatically remove <a href='https://developer.mozilla.org/docs/Web/API/Animation#Automatically_removing_filling_animations'>indefinite filling animations</a>.",
+          "description": "Browsers automatically remove indefinite filling animations.",
```
ddbeck, line 1342: While we're here, we should drop the punctuation

```suggestion
          "description": "Browsers automatically remove indefinite filling animations",
```

css/selectors/invalid.json
```diff
@@ -51,7 +51,8 @@
         },
         "form": {
           "__compat": {
-            "description": "Applies to <a href='https://developer.mozilla.org/docs/Web/HTML/Element/form'>&lt;form&gt;</a> elements",
+            "mdn_url": "https://developer.mozilla.org/docs/Web/HTML/Element/form",
```
ddbeck, line 54: I think the link here might be confusing. Really, the feature relates to the selector, not the form element. On MDN, this link makes sense, but it might be confusing in other consumers' contexts

```suggestion
```

css/selectors/valid.json
```diff
@@ -51,7 +51,8 @@
         },
         "form": {
           "__compat": {
-            "description": "Applies to <a href='https://developer.mozilla.org/docs/Web/HTML/Element/form'>&lt;form&gt;</a> elements",
+            "mdn_url": "https://developer.mozilla.org/docs/Web/HTML/Element/form",
```
ddbeck, line 54:
```suggestion
```

ddbeck:
For the other items mentioned, but not yet(?) changed in the PR:

https://github.com/mdn/browser-compat-data/blob/e2c50c4aba195b5bcbdf74cd2650bf286a42a7a3/javascript/builtins/Function.json#L833

This is the hardest one. Honestly, I'm not sure what to do about this, simply because the "revision" does a lot of different things at once. I'd like to get @Elchi3's take on this one in particular.

The others are a bit more tractable.

https://github.com/mdn/browser-compat-data/blob/e2c50c4aba195b5bcbdf74cd2650bf286a42a7a3/css/properties/display.json#L160

[The MDN page covers this](https://developer.mozilla.org/en-US/docs/Web/CSS/display#Box), so maybe just dropping the link is inconsequential. And making sure the section of the page links to the spec.

https://github.com/mdn/browser-compat-data/blob/e2c50c4aba195b5bcbdf74cd2650bf286a42a7a3/css/selectors/visited.json#L54

Now this one is kinda interesting.

First, there's a content problem: there's [a page about `:visited`: and privacy](https://developer.mozilla.org/en-US/docs/Web/CSS/Privacy_and_the_:visited_selector), but that's significant information, which is barely mentioned in the reference page. [`:visited`](https://developer.mozilla.org/en-US/docs/Web/CSS/:visited) deserves a "Privacy" section, IMO. I'll open a content issue on this.

In terms of this feature, I think the description is phrased unclearly and lets the link do the work. We can probably revise that and do without the link successfully. Perhaps something like `Privacy: <code>:visited</code> does not match <code>&lt;link&gt;</code> elements`.

Finally, that feature's data is a mess (e.g., the notes about "never" having done something should be discard and the `"version_added"` values set to the first release). It deserves a clean up.

https://github.com/mdn/browser-compat-data/blob/e2c50c4aba195b5bcbdf74cd2650bf286a42a7a3/javascript/builtins/JSON.json#L212

This one can also survive removing the link if we revise the description text. It's probably better as something like `String values are in well-formed UTF-8` or `Strings are escaped to well-formed UTF-8`. We could also set the `spec_url` to https://tc39.es/ecma262/#sec-quotejsonstring.

gsouquet:
That is correct, it is not yet changed because I wanted to gather a bit of feedback before pushing this forward.
I'll wait for @Elchi3 's review and update all at once accordingly

ddbeck:
@gsouquet I don't expect @Elchi3 to have time to come back to this soon, so I think we should go ahead without him and continue removing the links from `"description"` fields.

To summarize myself above, with some more clarity or specificity:

| feature | new description | notes |
| --- | --- | --- |
| `javascript.builtins.Function.toString.toString_revision` | `Implements <code>Function.prototype.toString</code> revision` | "`Function.prototype.toString` revision" is the title of the proposal, the phrase Chrome Platform Status uses, and if you search for this, the spec is very likely to be the first result
| `css.properties.contents.contents_unusual` |`Specific behavior of unusual elements when <code>display: contents</code> is applied to them` | drop the link
| `css.selectors.visited.not_match_link` | `<code>:visited</code> privacy: selector does not match <code>&lt;link&gt;</code> elements` | |
| `javascript.builtins.JSON.stringify.well_formed_stringify` | `Strings are escaped to well-formed UTF-8` | |

The other things I rambled on about are either resolved (open content issue) or not that significant (the messy notes).

alshamiri5 (approved):

ddbeck:
I opened #7154 to pick up where we left on this. When it gets merged, it will close this PR (though you'll get credit on the superseding PR). Thank you, @gsouquet!

gsouquet:
Thank you @ddbeck , that completely fell off my radar. Anything I can do to help here?

ddbeck:
@gsouquet yeah, it's tough to keep up with everything! I don't think there's anything left to do for this, unless you'd like to review #7154 yourself.

PR #6362, opened 2020-07-02
Safari 13 supports Intl.PluralRules

Constellation:
This PR adds support information of Intl.PluralRules in Safari 13.
bugzilla: https://bugs.webkit.org/show_bug.cgi?id=199288
caniuse: https://caniuse.com/#feat=intl-pluralrules


Constellation:
Thanks for merging!!! :D

PR #6364, opened 2020-07-02
adding note about loadstart event firing async in Fx 79 onwards

chrisdavidmills:
As per https://bugzilla.mozilla.org/show_bug.cgi?id=1502403


chrisdavidmills:
@ddbeck  Yup, I agree with you; updated!

chrisdavidmills:
@ddbeck How now brown cow? ;-)

PR #6366, opened 2020-07-02
adding data about implicit rel=noopener behavior

chrisdavidmills:
As per https://bugzilla.mozilla.org/show_bug.cgi?id=1522083

Enabled for Firefox 79.

I got the data for Safari desktop and Chromiums from https://www.fxsitecompat.dev/en-CA/docs/2020/target-blank-on-anchors-now-implies-rel-noopener/

### DIFF
--- a/html/elements/a.json
+++ b/html/elements/a.json
@@ -353,7 +353,7 @@
         },
         "implicit_noopener": {
           "__compat": {
-            "description": "Setting <code>target=\"_blank\"</code> also provides implicit <code>rel=\"noopener\"</code> behavior",
+            "description": "<code>target=\"_blank\"</code> implies <code>rel=\"noopener\"</code> behavior",
             "support": {
               "chrome": {
                 "version_added": false
--- a/html/elements/a.json
+++ b/html/elements/a.json
@@ -351,6 +351,54 @@
             }
           }
         },
+        "implicit_noopener": {
+          "__compat": {
+            "description": "Setting <code>target=\"_blank\"</code> also provides implicit <code>rel=\"noopener\"</code> behavior",
+            "support": {
+              "chrome": {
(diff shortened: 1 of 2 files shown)


html/elements/a.json
```diff
@@ -351,6 +351,54 @@
             }
           }
         },
+        "implicit_noopener": {
+          "__compat": {
+            "description": "Setting <code>target=\"_blank\"</code> also provides implicit <code>rel=\"noopener\"</code> behavior",
```
ddbeck, line 356: Thank you for this Chris. The data is good. I had a thought that the description could be a bit punchier, like this:

```suggestion
            "description": "<code>target=\"_blank\"</code> implies <code>rel=\"noopener\"</code> behavior",
```

Less wordy still would be this…

```suggestion
            "description": "<code>target=\"_blank\"</code> implies <code>rel=\"noopener\"</code>",
```

…but I'm not sure it's right: `someElement.getAttribute("rel") === "noopener"` but there's no actual `rel="noopener"` attribute. I'm not sure if that's a distinction without a difference. I'd be interested in your take.

chrisdavidmills:
@ddbeck thanks for the advice on wording. I've gone for the first one, as I think being more explicit is useful = it doesn't actually set the attribute, but it implies the behavior.

mkurz:
@chrisdavidmills @ddbeck Starting with version 88, Chrome (and other chromium based browsers) will imply `rel=noopener` if anchor `target=_blank`, the corresponding issue just got closed: https://bugs.chromium.org/p/chromium/issues/detail?id=898942#c29 Can you update the data in BCD?

ddbeck:
@mkurz I opened an issue, #7324, based on your comment. Thank you!

PR #6394, opened 2020-07-13
add `export * as ns` support data

JLHwung:
- [x] Summarize your changes

**Context**
Babel 7.11 will detect whether `export * as ns` should be transformed according to compat-data here: https://github.com/babel/babel/pull/11849

**Sources:**
V8: https://v8.dev/blog/v8-release-72#module-namespace-exports
SM: https://bugzilla.mozilla.org/show_bug.cgi?id=1496852, https://groups.google.com/g/mozilla.dev.platform/c/NSzWY92Ut0o?pli=1
JSC: https://bugs.webkit.org/show_bug.cgi?id=214379

Environments Mapping: https://github.com/mdn/browser-compat-data/tree/master/browsers

### DIFF
--- a/javascript/statements.json
+++ b/javascript/statements.json
@@ -799,7 +799,7 @@
         },
         "namespace": {
           "__compat": {
-            "description": "<code>export * as ns</code>",
+            "description": "<code>export * as namespace</code>",
             "mdn_url": "https://developer.mozilla.org/docs/Web/JavaScript/Reference/Statements/export",
             "spec_url": "https://tc39.es/ecma262/#sec-exports",
             "support": {
--- a/javascript/statements.json
+++ b/javascript/statements.json
@@ -813,7 +813,7 @@
                 "version_added": "79"
               },
               "firefox": {
-                "version_added": false
+                "version_added": "80"
               },
               "firefox_android": {
                 "version_added": false
--- a/javascript/statements.json
+++ b/javascript/statements.json
@@ -821,22 +821,9 @@
               "ie": {
                 "version_added": false
               },
-              "nodejs": [
-                {
-                  "version_added": "13.2.0",
-                  "notes": "Modules must either have a filename ending in <code>.mjs</code>, or the nearest parent <code>package.json</code> file must contain <code>\"type\": \"module\"</code>. See Node's <a href='https://nodejs.org/api/esm.html#esm_enabling'>ECMAScript Modules documentation</a> for more details."
-                },
-                {
-                  "version_added": "12.0.0",
-                  "flags": [
-                    {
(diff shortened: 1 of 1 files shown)


javascript/statements.json
```diff
@@ -796,6 +796,72 @@
...
+              "firefox": {
+                "version_added": false
+              },
+              "firefox_android": {
+                "version_added": false
+              },
+              "ie": {
+                "version_added": false
+              },
+              "nodejs": [
```
ddbeck, line 824: As a subfeature of `export`, we won't need the flags data here. It can be just be `"version_added": "12.0.0"`

javascript/statements.json
```diff
@@ -796,6 +796,72 @@
               "deprecated": false
             }
           }
+        },
+        "namespace": {
+          "__compat": {
+            "description": "<code>export * as ns</code>",
```
ddbeck, line 802:
```suggestion
            "description": "<code>export * as namespace</code>",
```

JLHwung:
> I think the feature should be a direct subfeature of export (at the same level as default, just before the current feature)

`namespace` is indeed a subfeature of `export`.

> we won't need the flags data here. It can be just be "version_added": "12.0.0"

Did you mean `version_added: "13.2.0"`? Otherwise as a compat-table consumer, the data will be

- `export` available (without flags) from 13.2.0
- `export namespace` available (no flags informations) from 12.0.0.

Also `export.default` has flags data, too. I can remove them if decision is made, but would want to clarify a bit.

ddbeck:
> `namespace` is indeed a subfeature of `export`.

Ah, my apologies. I misread the folding in my editor.

> Did you mean `version_added: "13.2.0"`? Otherwise as a compat-table consumer, the data will be
>
> - `export` available (without flags) from 13.2.0
> - `export namespace` available (no flags informations) from 12.0.0.

No, I meant from 12.0.0. When it comes to flags, subfeatures present a slightly different story. With APIs, as opposed to language features, it's a little more obvious: if some `SomeAPI` is behind a flag, then `SomeAPI.someMethod()` is _also_ behind the flag, but there's typically no additional flag needed for `someMethod()` to be enabled. Duplicating the flag data on `someMethod()` is a bit misleading because you _can't_ switch on `someMethod()` alone. It's only gated through the parent feature.

I understood this to be a similar situation: you can't turn on the namespace feature independent of the parent modules feature.

This is an area where we could stand to improve the documentation for the schema and contributions.

> Also `export.default` has flags data, too. I can remove them if decision is made, but would want to clarify a bit.

Yeah, this is a bit inconsistent. I'd welcome a PR fixing that too. It doesn't have to be part of this change, though.

PR #6403, opened 2020-07-15
Update MediaQueryList with details about Safari

a2sheppy:
This adds to MediaQueryList.json notes about the lack of
`EventTarget` support in the MQL implementation on Safari,
but also that this is coming in iOS 14 and macOS 11 Big Sur.
Also assorted other corrections including fixing an
experimental flag that was `true` by mistake and adding
description fields to the methods to show them properly in
the tables.

Sources:
* https://trac.webkit.org/changeset/261012/webkit
  - Changeset fixing this for WebKit
* https://bugs.webkit.org/show_bug.cgi?id=203288
  - WebKit bug

### DIFF
--- a/api/MediaQueryList.json
+++ b/api/MediaQueryList.json
@@ -175,10 +175,12 @@
               "version_added": true
             },
             "safari": {
-              "version_added": "5.1"
+              "version_added": "5.1",
+              "notes": "Prior to Safari 14, <code>MediaQueryList</code> was not based on <code>EventTarget</code>, so you had to use <code>addListener()</code> and <code>removeListener()</code> to observe media query lists."
             },
             "safari_ios": {
-              "version_added": "5"
+              "version_added": "5",
+              "notes": "Prior to Safari 14, <code>MediaQueryList</code> was not based on <code>EventTarget</code>, so you had to use <code>addListener()</code> and <code>removeListener()</code> to observe media query lists."
             },
             "samsunginternet_android": {
               "version_added": "1.0"
@@ -372,6 +374,7 @@
               "notes": "Prior to Safari 14, <code>MediaQueryList</code> was not based on <code>EventTarget</code>, so you had to use <code>addListener()</code> and <code>removeListener()</code> to observe media query lists."
             },
             "safari_ios": {
+              "version_added": "5",
               "notes": "Prior to Safari 14, <code>MediaQueryList</code> was not based on <code>EventTarget</code>, so you had to use <code>addListener()</code> and <code>removeListener()</code> to observe media query lists."
             },
             "samsunginternet_android": {
--- a/api/MediaQueryList.json
+++ b/api/MediaQueryList.json
@@ -29,10 +29,12 @@
             "version_added": true
           },
           "safari": {
-            "version_added": "5.1"
+            "version_added": "5.1",
+            "notes": "In Safari, <code>MediaQueryList</code> is not based on <code>EventTarget</code>, so you must use <code>addListener()</code> and <code>remoteListener()</code> to observe media query lists. This is expected to change in macOS 11 Big Sur."
           },
           "safari_ios": {
-            "version_added": "5"
+            "version_added": "5",
+            "notes": "In Safari, <code>MediaQueryList</code> is not based on <code>EventTarget</code>, so you must use <code>addListener()</code> and <code>remoteListener()</code> to observe media query lists. This is expected to change in iOS 14."
           },
           "samsunginternet_android": {
             "version_added": "1.0"
@@ -42,7 +44,7 @@
           }
         },
         "status": {
-          "experimental": true,
+          "experimental": false,
           "standard_track": true,
           "deprecated": false
         }
@@ -97,7 +99,7 @@
       },
       "EventTarget_inheritance": {
         "__compat": {
-          "description": "<code>MediaQueryList</code> inheriting from <code>EventTarget</code>",
+          "description": "<code>MediaQueryList</code> is based on <code>EventTarget</code>",
           "support": {
             "chrome": {
               "version_added": "45"
@@ -146,6 +148,7 @@
       "addListener": {
         "__compat": {
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/MediaQueryList/addListener",
+          "description": "<code>addListener()</code>",
           "support": {
             "chrome": {
               "version_added": "9"
@@ -338,6 +341,7 @@
       "removeListener": {
         "__compat": {
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/MediaQueryList/removeListener",
+          "description": "<code>removeListener()</code>",
           "support": {
             "chrome": {
               "version_added": "9"
--- a/api/MediaQueryList.json
+++ b/api/MediaQueryList.json
@@ -192,7 +192,7 @@
           "status": {
             "experimental": false,
             "standard_track": true,
-            "deprecated": true
+            "deprecated": false
           }
         }
       },
@@ -387,7 +387,7 @@
           "status": {
             "experimental": false,
             "standard_track": true,
-            "deprecated": true
+            "deprecated": false
           }
         }
       }
--- a/api/MediaQueryList.json
+++ b/api/MediaQueryList.json
@@ -30,11 +30,11 @@
           },
           "safari": {
             "version_added": "5.1",
-            "notes": "In Safari, <code>MediaQueryList</code> is not based on <code>EventTarget</code>, so you must use <code>addListener()</code> and <code>remoteListener()</code> to observe media query lists. This is expected to change in macOS 11 Big Sur."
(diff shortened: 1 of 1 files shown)


api/MediaQueryList.json
```diff
@@ -29,10 +29,12 @@
             "version_added": true
           },
           "safari": {
-            "version_added": "5.1"
+            "version_added": "5.1",
+            "notes": "In Safari, <code>MediaQueryList</code> is not based on <code>EventTarget</code>, so you must use <code>addListener()</code> and <code>remoteListener()</code> to observe media query lists. This is expected to change in macOS 11 Big Sur."
```
ddbeck, line 33: A style note that probably applies in any case: don't introduce the browser if there's no other caveats (e.g., "In Safari" is unnecessary, but "Before Safari 14" would be).

The real issue here is a) whether this bit about macOS 11 is relatively speculative or certain and b) whether it's specific to the version of macOS. If you're not sure it's going to ship, then drop macOS thing entirely, like this:

```suggestion
            "notes": "<code>MediaQueryList</code> is not based on <code>EventTarget</code>, so you must use <code>addListener()</code> and <code>remoteListener()</code> to observe media query lists."
```

If you want the data to be revisited in the future, an issue is the right place for that information.

If you're certain it's going to ship, then I'd be very surprised if it had different behavior based on the version of macOS rather than Safari (e.g., that equivalent Safari versions will have different behavior on macOS 10.15 than macOS 11). There's nothing in the linked bug that suggests that. I'm guessing it's more like this:

```suggestion
            "notes": "Before Safari 14, <code>MediaQueryList</code> is not based on <code>EventTarget</code>, so you must use <code>addListener()</code> and <code>remoteListener()</code> to observe media query lists."
```

If it really is operating-system version dependent, then we can revisit that. But that would be very unusual and I'd want us to document that in this PR for posterity.

api/MediaQueryList.json
```diff
@@ -29,10 +29,12 @@
...
           },
           "safari_ios": {
-            "version_added": "5"
+            "version_added": "5",
+            "notes": "In Safari, <code>MediaQueryList</code> is not based on <code>EventTarget</code>, so you must use <code>addListener()</code> and <code>remoteListener()</code> to observe media query lists. This is expected to change in iOS 14."
```
ddbeck, line 37: Similar issues here. Probably one of these is more accurate:

More certain:

```suggestion
            "notes": "Before iOS 14, <code>MediaQueryList</code> is not based on <code>EventTarget</code>, so you must use <code>addListener()</code> and <code>remoteListener()</code> to observe media query lists."
```

More speculative:

```suggestion
            "notes": "<code>MediaQueryList</code> is not based on <code>EventTarget</code>, so you must use <code>addListener()</code> and <code>remoteListener()</code> to observe media query lists."
```

api/MediaQueryList.json
```diff
@@ -97,7 +99,7 @@
       },
       "EventTarget_inheritance": {
         "__compat": {
-          "description": "<code>MediaQueryList</code> inheriting from <code>EventTarget</code>",
+          "description": "<code>MediaQueryList</code> is based on <code>EventTarget</code>",
```
ddbeck, line 102: Can you elaborate on your thinking by the change here a little? I see that the original is a little awkward, but reading the spec it seems like inheritance is the right term. Perhaps something like this is good?

```suggestion
          "description": "<code>MediaQueryList</code> inherits <code>EventTarget</code>",
```

Or maybe I'm missing something and the relationship is weirder than I thought.

api/MediaQueryList.json
a2sheppy: My intent here is to note that Safari 14, which currently includes this change, is expected to ship with Big Sur. However, you're right that this is extraneous information for our purposes, so I'll remove it.

Historically I've often included the name of the browser the note applies to because it helps in use cases in which a tool is simply displaying all the notes as footnotes. Readability of the note is much improved. However, if our style guide now prefers otherwise, I'll endeavor to change my practices on this.

api/MediaQueryList.json
a2sheppy: I don’t recall making that change, though clearly  I did. I’ll undo it.

api/MediaQueryList.json
```diff
@@ -30,11 +30,11 @@
           },
           "safari": {
             "version_added": "5.1",
-            "notes": "In Safari, <code>MediaQueryList</code> is not based on <code>EventTarget</code>, so you must use <code>addListener()</code> and <code>remoteListener()</code> to observe media query lists. This is expected to change in macOS 11 Big Sur."
+            "notes": "Prior to Safari 14, <code>MediaQueryList</code> was not based on <code>EventTarget</code>, so you had to use <code>addListener()</code> and <code>removeListener()</code> to observe media query lists."
```
ddbeck, line 33: Can we roll these back to present tense? Safari 14 isn't even released yet, for starters. Like this:

```suggestion
            "notes": "Prior to Safari 14, <code>MediaQueryList</code> is based on <code>EventTarget</code>, so you must use <code>addListener()</code> and <code>removeListener()</code> to observe media query lists."
```

api/MediaQueryList.json
```diff
@@ -30,11 +30,11 @@
...
           },
           "safari_ios": {
             "version_added": "5",
-            "notes": "In Safari, <code>MediaQueryList</code> is not based on <code>EventTarget</code>, so you must use <code>addListener()</code> and <code>remoteListener()</code> to observe media query lists. This is expected to change in iOS 14."
+            "notes": "Prior to Safari 14, <code>MediaQueryList</code> was not based on <code>EventTarget</code>, so you had to use <code>addListener()</code> and <code>removeListener()</code> to observe media query lists."
```
ddbeck, line 37:
```suggestion
            "notes": "Prior to Safari 14, <code>MediaQueryList</code> is based on <code>EventTarget</code>, so you must use <code>addListener()</code> and <code>removeListener()</code> to observe media query lists."
```

api/MediaQueryList.json
```diff
@@ -175,10 +175,12 @@
               "version_added": true
             },
             "safari": {
-              "version_added": "5.1"
+              "version_added": "5.1",
+              "notes": "Prior to Safari 14, <code>MediaQueryList</code> was not based on <code>EventTarget</code>, so you had to use <code>addListener()</code> and <code>removeListener()</code> to observe media query lists."
```
ddbeck, line 179:
```suggestion
              "notes": "Prior to Safari 14, <code>MediaQueryList</code> is based on <code>EventTarget</code>, so you must use <code>addListener()</code> and <code>removeListener()</code> to observe media query lists."
```

api/MediaQueryList.json
```diff
@@ -175,10 +175,12 @@
...
             },
             "safari_ios": {
-              "version_added": "5"
+              "version_added": "5",
+              "notes": "Prior to Safari 14, <code>MediaQueryList</code> was not based on <code>EventTarget</code>, so you had to use <code>addListener()</code> and <code>removeListener()</code> to observe media query lists."
```
ddbeck, line 183:
```suggestion
              "notes": "Prior to Safari 14, <code>MediaQueryList</code> is based on <code>EventTarget</code>, so you must use <code>addListener()</code> and <code>removeListener()</code> to observe media query lists."
```

api/MediaQueryList.json
```diff
@@ -190,7 +192,7 @@
           "status": {
             "experimental": false,
             "standard_track": true,
-            "deprecated": false
+            "deprecated": true
```
ddbeck, line 195: Was this actually deprecated? This is unexpected—it'd be nice to have a link in this PR to some substantiation of this.

api/MediaQueryList.json
```diff
@@ -383,7 +387,7 @@
...
```
ddbeck, line 390: Similar deprecation issue here

api/MediaQueryList.json
```diff
@@ -368,10 +370,12 @@
               "version_added": true
             },
             "safari": {
-              "version_added": "5.1"
+              "version_added": "5.1",
+              "notes": "Prior to Safari 14, <code>MediaQueryList</code> was not based on <code>EventTarget</code>, so you had to use <code>addListener()</code> and <code>removeListener()</code> to observe media query lists."
```
ddbeck, line 374:
```suggestion
              "notes": "Prior to Safari 14, <code>MediaQueryList</code> is based on <code>EventTarget</code>, so you must use <code>addListener()</code> and <code>removeListener()</code> to observe media query lists."
```

api/MediaQueryList.json
```diff
@@ -368,10 +370,12 @@
...
             },
             "safari_ios": {
-              "version_added": "5"
+              "version_added": "5",
+              "notes": "Prior to Safari 14, <code>MediaQueryList</code> was not based on <code>EventTarget</code>, so you had to use <code>addListener()</code> and <code>removeListener()</code> to observe media query lists."
```
ddbeck, line 378:
```suggestion
              "notes": "Prior to Safari 14, <code>MediaQueryList</code> is based on <code>EventTarget</code>, so you must use <code>addListener()</code> and <code>removeListener()</code> to observe media query lists."
```

a2sheppy:
@ddbeck Should be good to go now but if you want to double check that’s fine.

PR #6407, opened 2020-07-16
Remove compat data for -ms prefixed selectors

wbamberg:
As discussed in https://github.com/mdn/sprints/issues/3491 I am archiving MDN pages that document old `-ms` prefixed CSS selectors, that are no longer supported in Edge. As noted in https://github.com/mdn/sprints/issues/3491#issuecomment-659597268, this PR removes the associated BCD.


ddbeck:
Technically, this doesn't meet [the irrelevance criteria](https://github.com/mdn/browser-compat-data/blob/HEAD/docs/data-guidelines.md#removal-of-irrelevant-features)—it was supported too recently. But maybe the criteria should be less strict for features which were only ever supported in one engine?

PR #6411, opened 2020-07-18
Mark methods unavailable in Manifest V3 deprecated

bershanskiy:
Chrome Manifest V3 deprecates a few methods that allow synchronous DOM access and thus are not compatible with asynchronous service workers. [The documentation](https://developer.chrome.com/extensions/migrating_to_manifest_v3#api_removals) explicitly asks extension developers to migrate to asynchronous alternatives.

- [ ] Summarize your changes

### DIFF
--- a/webextensions/api/extension.json
+++ b/webextensions/api/extension.json
@@ -56,9 +56,11 @@
             "mdn_url": "https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/API/extension/getBackgroundPage",
             "support": {
               "chrome": {
+                "notes": "Not supported within service workers (Manifest V3).",
                 "version_added": true
               },
               "edge": {
+                "notes": "Not supported within service workers (Manifest V3).",
                 "version_added": "14"
               },
               "firefox": {
@@ -68,11 +70,17 @@
                 "version_added": "48"
               },
               "opera": {
+                "notes": "Not supported within service workers (Manifest V3).",
                 "version_added": true
               },
               "safari": {
                 "version_added": "14"
               }
+            },
+            "status": {
+              "experimental": false,
+              "standard_track": true,
+              "deprecated": true
             }
           }
         },
@@ -81,9 +89,11 @@
             "mdn_url": "https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/API/extension/getExtensionTabs",
             "support": {
               "chrome": {
+                "notes": "Not supported within service workers (Manifest V3).",
                 "version_added": true
               },
               "edge": {
+                "notes": "Not supported within service workers (Manifest V3).",
                 "version_added": "79"
               },
               "firefox": {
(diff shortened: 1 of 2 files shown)


ddbeck:
Hi @wbamberg, do you plan to review this? If you don't want or expect to get to it, feel free to remove yourself as reviewer

wbamberg:
Sorry for being slow to look at this.

For reference, this PR updates BCD for three features:
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/extension/getBackgroundPage
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/extension/getExtensionTabs
https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/runtime/getBackgroundPage

I have a couple of concerns here.

First, I think "Not supported within service workers (Manifest V3)" is rather an unclear way to express this. I'm not very familiar with Manifest v3, so I didn't know about https://developer.chrome.com/extensions/migrating_to_service_workers. So I think "service workers? what have they got to do with anything?". And just "(Manifest V3)" is unclear. What about Manifest 3? 

Second, standards status for WebExtensions APIs is tricky, because there isn't really a standard (yes I know there is community WG but it's dormant (https://browserext.github.io/)). See also here: https://github.com/mdn/browser-compat-data/issues/618 for a previous attempt at this. So it's not clear how we should decide "deprecated" status for them.

Having said that, we do sometimes use "deprecated" (as above, for [`extension.getExtensionTabs`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/extension/getExtensionTabs#Browser_compatibility)). I think if the Firefox add-ons team would also like people to stop using all three of these APIs (and we already know that `getBackgroundPage` is broken in private windows) then I think it would be helpful to mark them deprecated in the data.

So I think there are two options here:

1. if the Firefox team would also like people to stop using all three of these APIs then I think it would be helpful to mark them deprecated in the data. Then we could add a note to the Chromiums like "Not supported in Manifest V3.".

2. if the Firefox team would like people to keep using `extension.getBackgroundPage` and `runtime.getBackgroundPage`, then don't update the `status` object, but add a note to the Chromiums like "Deprecated. Not supported in Manifest V3.".

@kewisch , do you have a view here? Or about this PR in general?

kewisch:
I do think it makes sense to mark those APIs as deprecated in preparation for Manifest v3. 

These APIs are all around accessing the background page that WebExtensions using Manifest v2 have. Manifest v3 doesn't have the concept of a background page, but instead uses a service worker that has access to the APIs necessary. This will be the same in Firefox with Manifest v3, so I think it makes sense to deprecate these APIs ahead of time. 

For Firefox there is a [deprecation policy](https://wiki.mozilla.org/WebExtensions/DeprecationPolicy) that we'd have to follow, we could start the process for that soon. This way we could feasibly add a more general deprecated message.

Generally, I think it would make sense to have a document that explains manifest versioning, that we could link to from messages like these.

/cc @jvillalobos

kewisch:
A note on the deprecation policy, we wouldn't actually be removing support for v2, but I think it would make sense to have the linter make note about this so that extension developers creating new add-ons are prepared and can avoid its use.

PR #6430, opened 2020-07-23
Fix data for allowFullscreen on HTMLIFrameElement

foolip:
This is messy because there's also a content attribute, and prefixed
variants of both, but the content attributes and IDL attributes didn't
always change at the same time. No webkit-prefixed variant of the IDL
attribute was ever exposed, and the moz-prefixed IDL attribute had a
different capitalization.

Version numbers found by bisecting on BrowserStack with this test:
https://software.hixie.ch/utilities/js/live-dom-viewer/saved/8300

The commits that added the final unprefixed form in each engine:
https://chromium.googlesource.com/chromium/src/+/9206c8a832a4ead0c9129fc110727f9155228d0c
https://hg.mozilla.org/mozilla-central/rev/91d78336d372
https://trac.webkit.org/changeset/205686/webkit

Because these changes are in bindings only, it's safe to assume that
these were supported at the same time in different browsers using the
same engine, including desktop/mobile.

The miscapitalized variant "allowfullscreen" that Gecko briefly had was
not added to the data.

### DIFF
--- a/api/HTMLIFrameElement.json
+++ b/api/HTMLIFrameElement.json
@@ -142,66 +142,62 @@
           }
         }
       },
-      "allowFullScreen": {
+      "allowFullscreen": {
         "__compat": {
           "support": {
             "chrome": {
-              "prefix": "webkit",
-              "version_added": "17",
-              "notes": "Daily test builds only."
+              "version_added": "38"
             },
             "chrome_android": {
-              "prefix": "webkit",
-              "version_added": "18",
-              "notes": "Daily test builds only."
+              "version_added": "38"
             },
             "edge": {
-              "version_added": "≤18"
+              "version_added": "≤15"
             },
             "firefox": [
               {
-                "version_added": "18"
+                "version_added": "22"
               },
               {
-                "prefix": "moz",
-                "version_added": "9"
+                "alternative_name": "mozAllowFullScreen",
+                "version_added": "9",
+                "version_removed": "18"
               }
             ],
             "firefox_android": [
               {
-                "version_added": "18"
+                "version_added": "22"
               },
               {
-                "prefix": "moz",
-                "version_added": "9"
+                "alternative_name": "mozAllowFullScreen",
+                "version_added": "9",
+                "version_removed": "18"
               }
             ],
             "ie": {
               "version_added": false
             },
             "opera": {
-              "version_added": false
+              "version_added": "25"
             },
             "opera_android": {
-              "version_added": false
+              "version_added": "25"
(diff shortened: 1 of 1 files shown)


api/HTMLIFrameElement.json
```diff
@@ -142,66 +142,62 @@
...
             },
             "chrome_android": {
-              "prefix": "webkit",
-              "version_added": "18",
-              "notes": "Daily test builds only."
+              "version_added": "38"
             },
             "edge": {
-              "version_added": "≤18"
+              "version_added": "≤15"
```
foolip, line 155: I couldn't go back further than Edge 15 on BrowserStack.

sideshowbarker: > I couldn't go back further than Edge 15 on BrowserStack.

I also haven’t been able to find any way to get any closer than that, so I reckon that is close enough (and if/when somebody later able to get more-precise version info, the data can be further updated then)

sideshowbarker:
```diff
-             "version_added": "≤18"
+             "version_added": "≤15"
```
> I couldn't go back further than Edge 15 on BrowserStack.
>> I also haven’t been able to find any way to get any closer than that, so I reckon that is close enough (and if/when somebody later able to get more-precise version info, the data can be further updated then)

Actually, this change is causing the linter to fail, because `≤15` isn’t a valid/recognized version value for Edge. The closest allowed value for this case is actually `≤18`, so we need to just revert this part of the change.

api/HTMLIFrameElement.json
```diff
@@ -142,66 +142,62 @@
...
-              "version_added": false
+              "version_added": "25"
             },
             "safari": {
-              "version_added": false
+              "version_added": "10.0"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "10.0"
```
sideshowbarker, line 190:
```diff
             "safari_ios": {
-              "version_added": null
+              "version_added": "10.0"
```

`10.0` isn’t an allowed version value for iOS Safari — but `10` is, so we need to change this to thate

sideshowbarker:
tip: You can run `npm test` locally in the working directory for you clone, and that will run the same automated checks that CI here runs — including checking the browser-version-number data in the change with the linter. (You can also run just the linter separately, with `npm run lint` — but there’s usually not much point in doing that rather than running `npm test`, because the other tests that then get run take almost no additional time at all; most of the time is consumed by the linter, and either way it only takes about 10 seconds.

PR #6433, opened 2020-07-23
HTMLDialogElement cancel event is supported on Firefox

plonibarploni:
This was added in Firefox 78, see the [bug tracker](https://bugzilla.mozilla.org/show_bug.cgi?id=1322947).
I have tested it using Firefox 78.0.2 on Ubuntu 20.04.

### DIFF
--- a/api/HTMLDialogElement.json
+++ b/api/HTMLDialogElement.json
@@ -78,7 +78,7 @@
               "version_added": "79"
             },
             "firefox": {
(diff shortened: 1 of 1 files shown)


chrisdavidmills:
@plonibarploni @ddbeck I know this has been merged, but `<dialog>` is disabled behind a pref, in its entirety. So the data for `cancel` should probably reflect this too?

PR #6435, opened 2020-07-23
Remove the code and documentation for Web Confluence integration

foolip:
These scripts are likely no longer being used by anyone, as we no longer
update Web Confluence on any reliable schedule, and are now taking a
different approach to collecting data for updating BCD:
https://github.com/foolip/mdn-bcd-collector


foolip:
@jpmedley says he's used this. Let's hold off on merging this until we sort that out :)

ddbeck:
Hi @foolip, when might I expect a follow up on this? I can be patient, of course, but if this is more speculative than likely, we might want to park this as an issue or something.

foolip:
I owe @jpmedley a reply to an email about this when I'm back from vacation next week.

I think it would be fine to close this if it's cluttering up reviews, I can reopen later.

ddbeck:
Sorry, @foolip I didn't notice your status. I'll let this sit for now—enjoy your vacation!

jpmedley:
I was not trying to prevent the removal of this. I just wanted to know
what to use instead.
Joe Medley | Technical Writer, Chrome DevRel | jmedley@google.com |
 816-678-7195
*If an API's not documented it doesn't exist.*

foolip:
@jpmedley, @vinyldarkscratch is working on the code to take the results in https://github.com/foolip/mdn-bcd-results and update BCD. However, I don't think it will be integrated into BCD itself the way that Confluence was.

Per https://github.com/GoogleChromeLabs/confluence/issues/584 we're no longer updating the Confluence data. Before we drop the code it might be good look at all the differences that `npm run confluence` still produces. Unfortunately it no longer works as it fails with "Error: Invalid argument not valid semver ('≤79' received)". (It wasn't updated to handle ranged versions.) I'll try hacking it a bit to see what changes are being suggested.

foolip:
I've released mdn-confluence 1.0.4 with a one-liner fix, and running `npm up mdn-confluence` in Confluence will now make `npm run confluence` work.

foolip:
I've sent https://github.com/mdn/browser-compat-data/pull/6526 to show the remaining changes suggested by Confluence. It will take a long time, but I'd be willing to go through those changes and extract what's valuable.

@jpmedley do you think consulting that PR would work for you, or would you like the tooling to remain in the repo until all differences have been consolidated?

foolip:
https://github.com/mdn/browser-compat-data/pull/6527 was opened automatically. Whether that's landed or not isn't super important.

ddbeck:
> I've sent #6526 to show the remaining changes suggested by Confluence. It will take a long time, but I'd be willing to go through those changes and extract what's valuable.

I really like this approach. Thank you for taking this on, @foolip!

foolip:
@ddbeck I've rebased this on top of the merged https://github.com/mdn/browser-compat-data/pull/6527 now.

foolip:
@ddbeck any chance this will makes today's release? I suspect many will be pleased with the many fewer dependencies :)

ddbeck:
@foolip I'm sorry, I did the release before seeing this. I thought we will still awaiting Joe's confirmation before merging. 😬

foolip:
That's alright, next release it is :)

foolip:
I have released mdn-confluence 2.0.0, which includes https://github.com/mdittmer/mdn-confluence/pull/47, which is effectively a rewrite of how the Confluence data is interpreted. The results are very different with these changes and I believe more correct. Previously it seems like some releases got mixed up, resulting in spurious data. No doubt there are bugs in the new code as well, but it's now contained to just a few files.

I dropped as many dependencies as possible, which results in a BCD only installing 88 packages instead of 228.

@ddbeck there ought to be a bot update for this version. Getting that merged before the removal would be great.

Note that the tool can also be used standalone, so removing the integration won't make it impossible to use.

foolip:
The 2.0.0 upgrade is up at https://github.com/mdn/browser-compat-data/pull/6555 now.

foolip:
Since I filed this issue, I've simplified https://github.com/mdittmer/mdn-confluence a lot, and I think it's within reach to have no dependencies beyond the compare-versions and yargs dependencies that BCD already has.

Part of my original motivation for this was to unburden BCD from all those extra dependencies. If mdn-confluence is stripped of its final extra dependency (foam2) I wonder if it might be acceptable to leave it in here until we've extracted all of the value via https://github.com/mdn/browser-compat-data/pull/6526?

Alternatively, I could remove it as a dependency and update the documentation to show how one uses it as a standalone tool.

@ddbeck to advise :)

ddbeck:
@foolip I'm happy to leave it in place. To be honest, I'm not too bothered by extra dev dependencies, though I am appreciative of fewer dependencies.

Do you want to close this? Or leave it open as blocked by #6526?

foolip:
Let's close it. I am a bit bothered by a lot of dependencies myself, especially when they result in non audit warnings. So I'll continue to trim dependencies if it annoys me, but not make it a priority :)

PR #6436, opened 2020-07-23
Update allowfullscreen/webkitallowfullscreen data for Chromium

foolip:
The important part is making it clear that Edge now supports the
unprefixed allowfullscreen attribute.

Both of these attributes were added in the distant past, before Blink
forked from WebKit. The version_added dates were not confirmed or
changed.

Commit where the webkitallowfullscreen attribute was removed:
https://chromium.googlesource.com/chromium/src/+/bc6619f6cb0a8e2bad1b9721cbc63b8ee1909aa0

Based on the date this must have been M38.

All Chromium-based browsers were updated based on this removal time, and
where the unprefixed allowfullscreen wasn't represented in the data it
was added. In the case of Edge and Opera, it's somewhat plausible they
had support for the unprefixed attribute before the switch to Chromium,
so a ranged version was used to avoid having to confirm that.

### DIFF
--- a/html/elements/iframe.json
+++ b/html/elements/iframe.json
@@ -155,7 +155,8 @@
                 },
                 {
                   "prefix": "webkit",
-                  "version_added": "17"
+                  "version_added": "17",
+                  "version_removed": "38"
                 }
               ],
               "chrome_android": [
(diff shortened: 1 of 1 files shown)


sideshowbarker:
Looks great 👍

PR #6438, opened 2020-07-24
Add info regarding support of module workers

aduh95:
That's the `type` option (defined in the [HTML living standard](https://html.spec.whatwg.org/multipage/workers.html#workertype)), and more precisely the support for `"module"` value. I have also updated the Node.js section to add ES modules support (in addition to CJS modules).

Ressources:
- [Chomium 80 changelog](https://blog.chromium.org/2019/12/chrome-80-content-indexing-es-modules.html)
- [Webkit ticket](https://bugs.webkit.org/show_bug.cgi?id=164860)
- [Firefox ticket](https://bugzilla.mozilla.org/show_bug.cgi?id=1247687)

### DIFF
--- a/api/Worker.json
+++ b/api/Worker.json
@@ -178,7 +178,7 @@
                 "version_added": null
               },
               "webview_android": {
-                "version_added": null
+                "version_added": "80"
               }
             },
             "status": {
--- a/api/Worker.json
+++ b/api/Worker.json
@@ -288,56 +288,56 @@
               "deprecated": false
             }
           }
-        }
-      },
-      "type": {
-        "__compat": {
-          "description": "Constructor <code>type</code> option",
-          "support": {
-            "chrome": {
-              "version_added": "80"
+        },
+        "type": {
+          "__compat": {
+            "description": "Constructor <code>type</code> option",
+            "support": {
+              "chrome": {
+                "version_added": "80"
+              },
+              "chrome_android": {
+                "version_added": "80"
+              },
+              "edge": {
+                "version_added": "80"
+              },
+              "firefox": {
+                "version_added": false
+              },
+              "firefox_android": {
+                "version_added": false
+              },
+              "ie": {
+                "version_added": false
+              },
+              "nodejs": {
+                "version_added": false
+              },
+              "opera": {
+                "version_added": "67"
+              },
+              "opera_android": {
+                "version_added": "57"
+              },
+              "safari": {
+                "version_added": false
+              },
+              "safari_ios": {
+                "version_added": false
+              },
+              "samsunginternet_android": {
+                "version_added": false
+              },
+              "webview_android": {
+                "version_added": "80"
+              }
             },
-            "chrome_android": {
-              "version_added": "80"
-            },
-            "edge": {
-              "version_added": "80"
-            },
-            "firefox": {
-              "version_added": false
-            },
-            "firefox_android": {
-              "version_added": false
-            },
-            "ie": {
-              "version_added": false
-            },
-            "nodejs": {
-              "version_added": false
-            },
-            "opera": {
-              "version_added": "67"
-            },
-            "opera_android": {
-              "version_added": "57"
-            },
-            "safari": {
-              "version_added": false
-            },
-            "safari_ios": {
-              "version_added": false
-            },
-            "samsunginternet_android": {
-              "version_added": false
-            },
-            "webview_android": {
-              "version_added": "80"
+            "status": {
+              "experimental": false,
+              "standard_track": true,
+              "deprecated": false
             }
-          },
-          "status": {
-            "experimental": false,
-            "standard_track": true,
-            "deprecated": false
           }
         }
       },
--- a/api/Worker.json
+++ b/api/Worker.json
@@ -28,7 +28,7 @@
               "partial_implementation": true,
               "notes": [
                 "Is a Node <code>EventEmitter</code> instead of DOM <code>EventTarget</code>.",
-                "Worker script environment expects CommonJS modules or ECMAScript2015 modules.",
+                "Worker script environment expects CommonJS modules or ECMAScript modules.",
                 "Must be imported from the <code>worker_threads</code> module."
               ]
             },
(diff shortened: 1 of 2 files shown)


api/Worker.json
```diff
@@ -126,6 +135,59 @@
...
+                "version_added": false
+              },
+              "safari_ios": {
+                "version_added": false
+              },
+              "samsunginternet_android": {
+                "version_added": null
+              },
+              "webview_android": {
+                "version_added": null
```
jpmedley, line 181:
```suggestion
                "version_added": "80"
```

jpmedley: https://www.chromestatus.com/features/5761300827209728

api/Worker.json
```diff
@@ -23,6 +23,15 @@
             "version_added": "10"
           },
           "nodejs": [
+            {
+              "version_added": "12.17.0",
```
ddbeck, line 27: CI is failing on this line because we have not yet recorded 12.17 as a valid release. To make this PR pass the tests, you'll need to add 12.17 to [`nodejs.json`](https://github.com/mdn/browser-compat-data/blob/master/browsers/nodejs.json).

api/Worker.json
```diff
@@ -23,6 +23,15 @@
...
+              "partial_implementation": true,
+              "notes": [
+                "Is a Node <code>EventEmitter</code> instead of DOM <code>EventTarget</code>.",
+                "Worker script environment expects CommonJS modules or ECMAScript2015 modules.",
```
ddbeck, line 31: The Node.js docs refer to these as "ECMAScript modules" and MDN as "modules". I think we should also avoid the spec year/number stuff too.

```suggestion
                "Worker script environment expects CommonJS modules or ECMAScript modules.",
```

api/Worker.json
```diff
@@ -126,6 +135,59 @@
             "deprecated": false
           }
         },
+        "es2015_module": {
```
ddbeck, line 138:
```suggestion
        "ecmascript_modules": {
```

api/Worker.json
```diff
@@ -126,6 +135,59 @@
...
+          "__compat": {
+            "description": "Support for ECMAScript2015 module worker",
```
ddbeck, line 140:
```suggestion
            "description": "Support for ECMAScript modules",
```

api/Worker.json
```diff
@@ -126,6 +135,59 @@
...
+              "firefox_android": {
+                "version_added": false
+              },
+              "ie": {
+                "version_added": false
+              },
+              "nodejs": {
+                "version_added": "12.17.0",
+                "partial_implementation": true,
+                "notes": "ES2015 is enabled based on file extension rather than type option."
```
ddbeck, line 163:
```suggestion
                "notes": "ECMAScript modules are enabled based on the file extension <code>.mjs</code> rather than the <code>type</code> option."
```

ddbeck: Of course, to create more work for you, this also suggests we should add a sibling feature, `type` (description:  `Constructor <code>type</code> option`), to capture which engines support the option. I'd welcome this addition to the PR, but if you don't want to take on that work, let me know—we can open an issue for it.

api/Worker.json
```diff
@@ -126,6 +135,59 @@
...
+              },
+              "opera": {
+                "version_added": null
```
ddbeck, line 166: Mapped based on the underlying engine version

```suggestion
                "version_added": "67"
```

api/Worker.json
```diff
@@ -126,6 +135,59 @@
...
+              },
+              "opera_android": {
+                "version_added": null
```
ddbeck, line 169:
```suggestion
                "version_added": "57"
```

api/Worker.json
```diff
@@ -126,6 +135,59 @@
...
+              },
+              "safari": {
+                "version_added": false
+              },
+              "safari_ios": {
+                "version_added": false
+              },
+              "samsunginternet_android": {
+                "version_added": null
```
ddbeck, line 178:
```suggestion
                "version_added": false
```

api/Worker.json
aduh95: Note that Node.js supports ESM with `.js` extension when setting `"type": "module"` in the `package.json` file. That being said I think the current phrasing is good enough, as `.mjs` works no matter what `package.json` contains.

aduh95: Regarding the `Constructor <code>type</code> option`, I'm not ure that would add much more value since the `"classic"` value is the default, and the only alternative value `"module"` is covered by this PR. Do you think it's relevant to add a separate entry?

ddbeck: Ah, I misunderstood your original note. I interpreted it to mean that Node doesn't support the constructor's `type` option and uses the file extension to determine the module type instead. The Node.js docs don't seem to show support for the `type` option, but I haven't tested it.

It sounds like the real behavior is rather more complex: it checks the `"type"` field in `package.json` or, if unset, the file extension. We can rewrite the note to reflect that (I'll open a new suggestion).

I'd say this reinforces the need for a type option feature: we have a case where ESM is supported but the mechanism for opting in to that support has different compatibility data from the support itself.

api/Worker.json
```diff
@@ -126,6 +135,59 @@
...
+              "firefox_android": {
+                "version_added": false
+              },
+              "ie": {
+                "version_added": false
+              },
+              "nodejs": {
+                "version_added": "12.17.0",
+                "partial_implementation": true,
+                "notes": "ECMAScript modules are enabled based on the file extension <code>.mjs</code> rather than the <code>type</code> option."
```
ddbeck, line 163:
```suggestion
                "notes": "ECMAScript modules are enabled for files ending with <code>.mjs</code> and for files ending with <code>.js</code> when the nearest parent <code>package.json</code> file contains a top-level field <code>"type"</code> with a value of <code>"module"</code>."
```

I've borrowed part of the phrasing here [from the Node.js docs](https://nodejs.org/api/esm.html#esm_package_json_type_field).

aduh95:
Not sure why the CI is failing this time..

ddbeck:
@aduh95 I checked it out locally and found that it was a sorting message from the linter. This is caused by the `"type"` object not being a sibling of the other `Worker()` subfeatures. If you bump that up one level, CI will pass. Let me know if you need any help with that.

PR #6442, opened 2020-07-27
appinstalled was mistakenly added

AdaRoseCannon:
- Removes appinstalled from Samsung Internet,

This feature has it's endpoints accidentally added for one version but were never hooked up. They were removed but mdn was not updated to reflect this, this PR rectifies that.

An upcoming Samsung Internet version will correctly add this feature I will make a new PR then.

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -4557,7 +4557,7 @@
               "version_added": null
             },
             "samsunginternet_android": {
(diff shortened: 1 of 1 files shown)


B4LQ15:
Please support and support because I'm in the learning phase

PR #6443, opened 2020-07-27
Make note about how IE normalizes rules in cssText

eoghanmurray:

### DIFF
--- a/api/CSSRule.json
+++ b/api/CSSRule.json
@@ -67,7 +67,8 @@
               "version_added": "4"
             },
             "ie": {
-              "version_added": null
+              "version_added": null,
+              "notes": "Does not return the original verbatim rule text. E.g. 'border-width:1;border-style:solid;border-color:white;' will be returned as 'border: 1px solid white;'"
             },
             "opera": {
               "version_added": true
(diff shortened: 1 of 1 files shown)


api/CSSRule.json
```diff
@@ -67,7 +67,8 @@
               "version_added": "4"
             },
             "ie": {
-              "version_added": null
+              "version_added": null,
```
ddbeck, line 70: Your note implies this feature has some support. We should mark this as `true`:

```suggestion
              "version_added": true,
```

api/CSSRule.json
```diff
@@ -67,7 +67,8 @@
...
+              "notes": "Does not return the original verbatim rule text. E.g. 'border-width:1;border-style:solid;border-color:white;' will be returned as 'border: 1px solid white;'"
```
ddbeck, line 71: Suggesting some edits here for clarity and style:

```suggestion
              "notes": "The property normalizes rules and does not return the original verbatim rule text. For example, <code>border-width: 1px; border-style:solid; border-color:white;</code> returns <code>border: 1px solid white;</code>."
```

eoghanmurray:
I'm very confused by the github UI! Not sure if I successfully accepted your changes or not @ddbeck

PR #6456, opened 2020-07-30
Add data for Periodic Background Sync

jpmedley:
https://www.chromestatus.com/feature/5689383275462656

### DIFF
--- a/api/PeriodicSyncEvent.json
+++ b/api/PeriodicSyncEvent.json
@@ -35,7 +35,7 @@
             "version_added": false
           },
           "samsunginternet_android": {
-            "version_added": true
+            "version_added": false
           },
           "webview_android": {
             "version_added": "80"
@@ -83,7 +83,7 @@
               "version_added": false
             },
             "samsunginternet_android": {
-              "version_added": true
+              "version_added": false
             },
             "webview_android": {
               "version_added": "80"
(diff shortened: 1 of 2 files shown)


api/PeriodicSyncEvent.json
```diff
@@ -34,10 +35,10 @@
             "version_added": false
           },
           "samsunginternet_android": {
-            "version_added": false
+            "version_added": true
```
ddbeck, line 38: Where did the Samsung Internet data come from? I'm not saying it's wrong, but it doesn't exactly square with the browser version mappings we have in `/browsers/samsung_internet.json`.

The Chrome, WebView, and MDN URLs look fine, so this is the only part I'm concerned about.

jpmedley: This is me forgetting that Samsung doesn't fork the current version of Chrome. I've changed this to false everywhere.

vinyldarkscratch: Looks like a few entries are still left as `true` in PeriodicSyncManager?

jpmedley: Sorry about that. I think I got them this time.

vinyldarkscratch: Not a problem, looks good now!

Issue #6431, opened 2020-07-23
Linter: prevent an all "no" row


queengooborg:
If a feature is not implemented in any browsers, and never was, we have chosen not to keep data here in BCD, as it could be a part of a changing spec.  However, it seems we've got a number of features that are all "false".  We should prevent them in the form of a linter.

ddbeck:
There's a bit to unpack here. Some initial thoughts:

First, we should be explicit about what we're trying to prevent here. The title here is a bit ambiguous. I'd be fine with preventing features which were never supported by any browser. I'd be opposed to having the linter prevent features where there is no current support. I'm going to assume it's former, but if it's the latter, there needs to be a serious discussion about that.

Based on #6463, it appears there are many files and features impacted by this proposal. Implementing a linter change is really the end of the line. Before that, we need to:

* Discourage the introduction of new data that goes against the aims of the proposal (i.e., write a data guideline and note it in the next release notes).
* Get a better understanding of the features affected. Are they features we should remove (i.e., we know they were never supported)? Or are they features that need better data (i.e., they're erroneous or incomplete)?
* Clean up the existing data, as appropriate. If we're removing features, we need to do in a way that can be shown in the release notes like other data removals.

Then we could merge a new linter check. (At the completion of this process, we might want to increment the minor version number too.)

bershanskiy:
I think it's a great idea to remove the "no" rows, too. However, I think linter should fault only "no" rows of deprecated features.

The majority of "false positives" of the linter in #6463 comes from upcoming features that already launched (but data wasn't updated yet) or will launch within the next couple releases of at least one browser (usually Chromium). Apparently, sometimes it's preferable to add "stubs" describing the structure of the feature data but not specific version numbers. Letting non-deprecated rows through would facilitate this use case.

A separate but similar linter might also remove [irrelevant features](https://github.com/mdn/browser-compat-data/blob/master/docs/data-guidelines.md#removal-of-irrelevant-features). However, such linter would rely on browser data to determine the old browsers (to correlate version numbers and release dates).

If anyone is interested in reviewing a patch (for either idea), I'd be happy to create one.

jculver050:
Close

PR #6463, opened 2020-08-02
Lint against a browser having all "false"

vinyldarkscratch:
This PR adds an additional feature to the version linter to check when a feature has all `false` for the support, to prevent the addition of features in a changing spec that hasn't shipped in any browsers.

Currently, this linter will **fail** as we have some data with all `false`.


ddbeck:
Is this to address #6431? If so, I'd like to propose that we close this PR for now and take up a discussion there. This feels like the end of a long process, not the beginning. There's several questions to answer and intermediate steps to carry out before we could merge a PR like this.

vinyldarkscratch:
Okay, yeah!  I'll close this PR for now then, and we can come back to it another time.  😃

PR #6468, opened 2020-08-04
Add inputmode for Firefox Mobile

test137E29B:
This adds data for Firefox Mobile's support of the inputmode attribute's usage as of Version 68, with a flag change.

Sourced from: [https://caniuse.com/#feat=input-inputmode](https://caniuse.com/#feat=input-inputmode)

### DIFF
--- a/html/global_attributes.json
+++ b/html/global_attributes.json
@@ -1108,7 +1108,14 @@
               }
             ],
             "firefox_android": {
-              "version_added": false
+              "version_added": "68",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "dom.forms.inputmode",
+                  "value_to_set": "true"
+                }
+              ]
             },
             "ie": {
               "version_added": false


evilpie:
Shouldn't `inputmode` be shipping in the new Firefox for Android 79? https://bugzilla.mozilla.org/show_bug.cgi?id=1631681

test137E29B:
> Shouldn't inputmode be shipping in the new Firefox for Android 79? https://bugzilla.mozilla.org/show_bug.cgi?id=1631681

From the pushed changes, it would appear that it's simply defaulted true as of an upcoming version if the Version is Android, rather than using the flag like it currently does, though I'm not sure of the version. (See [https://hg.mozilla.org/integration/autoland/rev/f98800d012f0](https://hg.mozilla.org/integration/autoland/rev/f98800d012f0))

There is currently no Firefox for Android v79 as the current is 68.11.0, released end of July this year. v79 is the Desktop version, which this change doesn't affect.

makotokato:
See https://github.com/mdn/browser-compat-data/issues/4983 for Firefox/Android version. 79 is stable release and this attribute supports on 79, but I don't know whether this is ready for this data.

html/global_attributes.json
```diff
@@ -1108,7 +1108,14 @@
               }
             ],
             "firefox_android": {
-              "version_added": false
+              "version_added": "68",
```
ddbeck, line 1111: Based on the feature in the desktop Firefox, I'm guessing this has been in the Android version for some time. Let's set this to true:

```suggestion
              "version_added": true,
```

evilpie: Why? I haven't seen any indication that this was enabled before: https://bugzilla.mozilla.org/show_bug.cgi?id=1631681#c0

ddbeck: Ah, thank you @evilpie. I misunderstood the original PR description ("as of Version 68") to mean that it was tested in the current version and no others. I'll merge it as is.

PR #6482, opened 2020-08-05
Update availability of the `revert` keyword

daformat:
[bug 579788](https://bugs.chromium.org/p/chromium/issues/detail?id=579788) was fixed and shipped on Chrome, Chrome Android and Edge

Added version info based on [caniuse support table](https://caniuse.com/#feat=css-revert-value)

### DIFF
--- a/css/properties/all.json
+++ b/css/properties/all.json
@@ -53,13 +53,13 @@
             "description": "<code>revert</code>",
             "support": {
               "chrome": {
-                "version_added": false
+                "version_added": "84"
               },
               "chrome_android": {
-                "version_added": false
+                "version_added": "84"
               },
               "edge": {
-                "version_added": false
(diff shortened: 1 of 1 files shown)


vsn4ik:
@daformat please, add `opera@70` too 🙏 Thank you!

daformat:
@Elchi3 thanks, adding this right now
@vsn4ik I only found a page mentioning that `opera@70` is based on `chromium@84` but without mention of the `revert` value, is it enough proof?

vsn4ik:
@daformat May be https://blogs.opera.com/desktop/changelog-for-70/:
```
CHR-8026 Update chromium on desktop-stable-84-3728 to 84.0.4147.105
```
i checked locally:
image

daformat:
Added both in last commit, thanks guys!

PR #6491, opened 2020-08-06
Update XMLHttpRequest.responseType for Opera

Yaffle:
Add XMLHttpRequestion#responseType === 'json' support info in Opera 12:
See http://www.oldversion.com/windows/opera-12-00-beta-1
Tested manually.


Yaffle:
@ddbeck thanks for checking

PR #6508, opened 2020-08-10
Add "depth" argument to traverse script

vinyldarkscratch:
This PR adds a "depth" argument to the traverse script to specify the maximum depth to traverse.  Useful to filter out behavioral subfeatures on attributes/operators (such as `.returns_promise`).

### DIFF
--- a/scripts/traverse.js
+++ b/scripts/traverse.js
@@ -19,6 +19,13 @@ const { argv } = require('yargs').command(
         describe: 'The value(s) to test against',
         type: 'array',
         default: ['null', 'true'],
+      })
+      .option('depth', {
+        alias: 'd',
+        describe:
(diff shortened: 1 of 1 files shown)


scripts/traverse.js
```diff
@@ -19,6 +19,13 @@ const { argv } = require('yargs').command(
         describe: 'The value(s) to test against',
         type: 'array',
         default: ['null', 'true'],
+      })
+      .option('depth', {
+        alias: 'd',
+        describe:
+          'Depth of features to go (ex. "2" will capture "api.CSSStyleSheet.insertRule" but not "api.CSSStyleSheet.insertRule.optional_index")',
```
foolip, line 26:
```suggestion
          'Depth of features to traverse (ex. "2" will capture "api.CSSStyleSheet.insertRule" but not "api.CSSStyleSheet.insertRule.optional_index")',
```

Issue #6510, opened 2020-08-10
Case mismatch in fullscreen api for IE 11


hyperd:
Some vendor specific events and properties for the full-screen api are not correct for IE11.
The ones I've identified are:
- MSFullscreenChange (should be msfullscreenchange)
- msFullScreenEnabled (should be msFullscreenEnabled)
- MSFullscreenError (should be msfullscreenerror)
- onMSFullscreenChange (should be onmsfullscreenchange)
- onMSFullscreenError (should be onmsfullscreenerror)

I'm happy to create a PR to fix this.

PR #6511, opened 2020-08-10
Fix IE 11 fullscreen api docs

hyperd:
Fixes https://github.com/mdn/browser-compat-data/issues/6510

JsBin to test this: https://jsbin.com/wizuwejocu/1/edit?html,js,output

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -4911,7 +4911,7 @@
             ],
             "ie": {
               "version_added": "11",
-              "alternative_name": "MSFullscreenChange"
+              "alternative_name": "msfullscreenchange"
             },
             "opera": {
               "version_added": "32"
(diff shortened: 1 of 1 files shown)


api/Document.json
```diff
@@ -4911,7 +4911,7 @@
             ],
             "ie": {
               "version_added": "11",
-              "alternative_name": "MSFullscreenChange"
+              "alternative_name": "msfullscreenchange"
```
foolip, line 4914: I believe that this and the "MSFullscreenError" change was mistaken, and have sent https://github.com/mdn/browser-compat-data/pull/6684 to change them back.

https://jsbin.com/wizuwejocu/1/edit?html,js,output tested the event handler attributes, but not the event types. These are almost always the same apart from the "on" prefix, but when the event type isn't all lowercase, the event handler attribute usually still is. At least I don't know of any exceptions to that.

hyperd: Ah sorry for misreading. Thank you for getting this fixed.

PR #6515, opened 2020-08-13
Add Safari version for response.clone

derduher:
relevant links:
tag: 604.2.10 https://trac.webkit.org/timeline?from=2017-09-12T16%3A52%3A44-07%3A00&precision=second
exact commit: https://trac.webkit.org/changeset/221772/webkit
test code https://codesandbox.io/s/zealous-currying-wljo8?file=/src/index.js
verified in Safari version 13.0.5

Response.body Response.bodyUsed, Body may also need to be updated

### DIFF
--- a/api/Response.json
+++ b/api/Response.json
@@ -389,7 +389,7 @@
               "version_added": "11.1"
             },
             "safari_ios": {
-              "version_added": "11"
+              "version_added": "11.3"
             },
             "samsunginternet_android": {
               "version_added": false
--- a/api/Response.json
+++ b/api/Response.json
@@ -386,10 +386,10 @@
               }
             ],
             "safari": {
-              "version_added": false
+              "version_added": "11.1"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "11"
             },
             "samsunginternet_android": {
               "version_added": false


api/Response.json
```diff
@@ -386,10 +386,10 @@
               }
             ],
             "safari": {
-              "version_added": false
+              "version_added": "11.1"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "11"
```
ddbeck, line 392: GitHub lost my comment. Let's try this again.

Based on the commit date and the release date in `/browsers/safari_ios.json`, this probably shipped in the next release in our inventory:

```suggestion
              "version_added": "11.3"
```

derduher:
updated, @ddbeck, Thanks!

ddbeck (approved):
Excellent, @derduher. Thank you! 🎉

foolip:
I came across this because it conflicts with a change (among many incorrect ones) in https://github.com/mdn/browser-compat-data/pull/6526 when rebasing. The data from Web Confluence suggests this was added earlier.

Testing [`'clone' in Response.prototype`](https://software.hixie.ch/utilities/js/live-dom-viewer/saved/8365) in BrowserStack and Sauce I find support for it in Safari 10.1.2 (WebKit 603.3.8).

I wonder if there is an error in the methodology that was used here? In particular, where did the tag 604.2.10 come from?

derduher:
The tag came from looking for the earliest commit that referenced clone in the source (see link above) It is possible that it was supported even earlier as the commit message does suggest an earlier method of cloning. Let me know If you'd like to dig deeper, I may have gotten it wrong.

foolip:
@derduher I see, thank you. It looks like `clone()` was indeed introduced prior to that commit, as you suspected. The way I usually check when things were added is to figure out which file has the IDL for it and use `git log -S` in a checkout. In this case, `git log -S clone Source/WebCore/Modules/fetch/FetchResponse.idl` led to https://trac.webkit.org/changeset/197049/webkit. How can I go from there to a WebKit tag, like you did to find 604.2.10?

@ddbeck would it be worth documenting how to do this sort of thing for each engine? I know how to do it for Chromium, but not for Gecko or WebKit.

Anyway, the `Conditional=FETCH_API` in FetchResponse.idl means one can't know for sure when it was enabled in any WebKit port. It should be together with the rest of the `fetch()` API though. Based on https://wiki.developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch and https://wiki.developer.mozilla.org/en-US/docs/Web/API/Response, that should be Safari 10.1 and Safari iOS 10.3. I'll send a PR.

derduher:
@foolip I looked at the the date of the commit date and then referenced the [taglist](https://trac.webkit.org/browser/webkit/tags) then the [timeline](https://trac.webkit.org/timeline?from=2017-09-12T16%3A52%3A44-07%3A00&precision=second) and found the commit. From there I went forward in time until the next tag.
I started all of this by looking at https://developer.mozilla.org/en-US/docs/MDN/Contribute/Processes/Matching_features_to_browser_version for safari. 
If IDL is a better way to go about it I'd definitely like a doc. Even tracking down what I did was pretty hard. I haven't used svn/trac in a solid decade.

derduher:
I think I found the commit by looking at the blame for the FetchResponse clone in the source.

foolip:
Ah, I hadn't seen https://developer.mozilla.org/en-US/docs/MDN/Contribute/Processes/Matching_features_to_browser_version#Safari, that's great. The "go through the branches or tags until you narrow down the one that introduced it" isn't very specific, but if there's a mechanical process to follow perhaps that can be added there.

ddbeck:
The linked MDN page is a good resource, but digging up details for Safari is hard. The WebKit version numbers aren't always reliable (IIRC, Safari has frozen the WebKit version in the user-agent string). One thing I find very useful: searching [WebKit/webkit](https://github.com/WebKit/webkit/) instead of Trac. I often end up finding a relevant commit, comparing the date of the commit and the dates of Safari releases, and making a best guess based on that.

Issue #6522, opened 2020-08-17
Label PRs with GitHub Actions


ddbeck:
[GitHub Actions now supports a `pull_request_target` event](https://github.blog/2020-08-03-github-actions-improvements-for-fork-and-pull-request-workflows/) that gives us the permissions we need to run [`@actions/labeler`](https://github.com/actions/labeler) with PRs from forks.

We attempted this in the past and it didn't work. Previous issues and PRs:

* #5119 (original proposal to use GitHub Actions)
* #5493 (added labeler using `pull_request` event)
* #5496 (revert #5493)

It's now been reported that this works using the new event type: https://github.com/actions/labeler/issues/12#issuecomment-670967607.

There's a few things to do after this is merged:

- [x] Shut down the Probot-based bot
- [x] Confirm that the thing works on a new fork-originating PR
- [x] Retire the Probot-based app

ddbeck:
Before merging #6523, I switched the existing bot to stop respond to PR webooks. After that, I opened #6565 to show that the labeler does indeed work with PRs from a fork. 🎉

Next steps are to delete the old app from the MDN org. I'm contacting the MDN dev team to do this now.

ddbeck:
And 💨 it's gone! The bot is dead, long live the bot. Thanks to everyone who participated on this, or the previous attempt to do this.

ddbeck:
One last note here: because the compat-data-helper has been removed, it shows up as a "ghost" user in old PRs.

PR #6523, opened 2020-08-17
Replace Probot labeler with GitHub Actions labeler

ddbeck:
This PR removes the Probot labeler and reinstates the official `@actions/labeler` action using the new `pull_request_target` event. See #6522 for background. It _should_ work this time. 🤞


.github/workflows/labeler.yml
```diff
@@ -0,0 +1,11 @@
+name: 'Pull Request Labeler'
+on:
+  - pull_request_target
```
foolip, line 3: I didn't know about https://docs.github.com/en/actions/reference/events-that-trigger-workflows#pull_request_target before, seems quite useful!

foolip (approved):

foolip:
I guess the plan is to land this and see if it works. @ddbeck have you tried running it on a fork of the repository, to shake out any problems? (It has to be a fork because the config will have to be on the default branch. Or maybe it could be done with two branches in this repo, but there's no need to try it here.)

schalkneethling:
@ddbeck This looks good to me. From what I can see and read, this should indeed do the trick.

ddbeck:
Thank you @foolip and @schalkneethling for the reviews! I'm going to deem your reviews sufficient to merge here; I'll be opening a demonstration PR to verify shortly. I'll report progress on #6522.

ddbeck:
Ugh, @foolip sorry I forgot to actually answer your question: yes, I did it test it, but because of the GitHub Actions permissions model and because GitHub won't let you have two forks of the same repo, I had to test it on a copy of this repo and a fork on my account (it worked).

foolip:
Cool, thanks for getting this done!

Issue #6414, opened 2020-07-19
file.name works in Safari on iOS and Firefox on Android


lgommans:
It is indicated that [Safari on iOS and Firefox on Android have no support at all for `file.name`](https://github.com/mdn/browser-compat-data/blob/e4b9484aed0a116a47f71efe1baf0de06838d135/api/File.json#L227-L242). If I understand it correctly, this should not work:

    <input id=x type=file>
    <script>
        $("#x").onchange = function(){
            doSomethingWith( $("#x").files[0].name );
        };
    </script>

However, it does work when testing with Fennec 68.10.1esr (which is Firefox but recompiled on F-Droid, requiring it to be rebranded due to trademark stuff) as well as on Safari 13.1.2 running on an iOS device.

Do we need to figure out the exact version when it was added before it can be fixed, or can we fix the table to at least be correct for the current versions? The current data seems to have been introduced in [this commit](https://github.com/mdn/browser-compat-data/commit/a85a9f499da574ca4078dbb3b322394bddf6a619) but I don't see any source referenced for those values.

(I also reported this through the MDN page and was linked to the general issue tracker to track my report, but they're not showing up here so I don't know into which fifo that went.)

ddbeck:
If you've tested this API and it works, I'd welcome a PR updating it. Some of the data—especially older data as in the case—was converted from old, hand-written tables on MDN pages, which we didn't closely verify at the time. Based on the age of the API, I'd expect the Android data to map from the desktop Firefox at version 5.

PR #6524, opened 2020-08-17
Correct support for File.name in safari_ios and firefox_android

lgommans:
The `File.name` API (accessed through `$('input[type="file"]').files[0].name`) works both in Safari on iOS and Firefox on Android; the current browser compat data table claims the contrary. This pull request just adds the browser versions I tested, I don't know when this was added.

See also #6414 for a PoC that is similar to what I tested in the versions added in this pull request. The real test was with my site https://dro.pm which uses this API.

### DIFF
--- a/api/File.json
+++ b/api/File.json
@@ -224,7 +224,7 @@
               "version_added": "3.6"
             },
             "firefox_android": {
-              "version_added": "≤68.10.1"
+              "version_added": "4"
             },
             "ie": {
               "version_added": "10"
--- a/api/File.json
+++ b/api/File.json
@@ -224,7 +224,7 @@
               "version_added": "3.6"
             },
             "firefox_android": {
-              "version_added": false
+              "version_added": "≤68.10.1"
             },
             "ie": {
               "version_added": "10"
@@ -239,7 +239,7 @@
               "version_added": true
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "≤13.1.2"
             },
             "samsunginternet_android": {
               "version_added": true
--- a/api/File.json
+++ b/api/File.json
@@ -239,7 +239,7 @@
               "version_added": true
             },
             "safari_ios": {
-              "version_added": "≤13.1.2"
+              "version_added": "8"
             },
             "samsunginternet_android": {
               "version_added": true


lgommans:
> Valid firefox_android versions are: true, false, null, 4, 5, [...], 67, 68

Afaik 68.x versions are not all equal, but sure I can dumb that down to just 68. Or we could add 68.{1..10} as possible values.

> Valid safari_ios versions are: true, false, null, 1, 2, 3, 4, 5, 6, 7, 8, 9, 10, 11, 12, 13, 14, 3.1, 3.2, 4.2, 4.3, 5.1, 6.1, 9.3, 10.3, 11.3, 12.2, 13.3, 13.4

Well that's interesting, the version in the user agent string says it really is 13.1.2:  
`Mozilla/5.0 (iPhone; CPU iPhone OS 13_6 like Mac OS X) AppleWebKit/605.1.15 (KHTML, like Gecko) Version/13.1.2 Mobile/15E148 Safari/604.1`  
Other websites say that OS version == Safari version, which seems to have been 13.6, but that's not in the whitelisted values either.

I can also re-test with an updated phone which now has Safari 14.0 on iOS 14.0 ([user agent string](https://lgms.nl/p/cau/?Mozilla%2F5.0+%28iPhone%3B+CPU+iPhone+OS+14_0+like+Mac+OS+X%29+AppleWebKit%2F605.1.15+%28KHTML%2C+like+Gecko%29+Version%2F14.0+Mobile%2F15E148+Safari%2F604.1)), but it doesn't look like that'll fare any better given that set of allowed versions...

Should I add 13.1 as a Safari version and add it for that, since that's the earliest known version where it works? (I assume you'll want to drop the patch level `.2`.) And what about Firefox: dumb down to 68 or add the 68.x minor versions?

api/File.json
```diff
@@ -239,7 +239,7 @@
               "version_added": true
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "≤13.1.2"
```
foolip, line 242:
```suggestion
              "version_added": "8"
```

I have tested `'name' in File.prototype` in Safari iOS on BrowserStack, and found that it's present in Safari iOS 8 but wasn't present in 7.

foolip (commented):
Thanks for fixing this! I think I found the right version number for Safari, but I'm not sure about Firefox.

foolip:
@ddbeck can you advise on how to handle Firefox for Android for features that have likely been around since the very beginning (at least 5 years in this case) but where it's not possible to test on a real device?

ddbeck:
Hi @lgommans and welcome to BCD!

> Afaik 68.x versions are not all equal, but sure I can dumb that down to just 68.

Yes, simplifying this to version 68 is correct. We generally track major versions only (there's some more nuance to this, but that's the gist of it), since it's very difficult to track down minor version differences for the major browsers. We have a data guideline that touches on this [in our contributing docs](https://github.com/mdn/browser-compat-data/blob/master/docs/data-guidelines.md#release-lines-and-backported-features).

> version in the user agent string

For Safari for iOS, we use the iOS version number, not the Safari version number. This is a historical quirk of our data set (one we hope contributors from Apple may help us rectify some day) but for now, that's what we're doing. Again, there's a little bit more information in [the data guidelines](https://github.com/mdn/browser-compat-data/blob/master/docs/data-guidelines.md#safari-for-ios-versioning). Though in this case, Philip has found a more accurate version number for us.

api/File.json
```diff
@@ -224,7 +224,7 @@
               "version_added": "3.6"
             },
             "firefox_android": {
-              "version_added": false
+              "version_added": "≤68.10.1"
```
ddbeck, line 227: Both a suggestion here and a response to @foolip:

> can you advise on how to handle Firefox for Android for features that have likely been around since the very beginning (at least 5 years in this case) but where it's not possible to test on a real device?

Firefox has pretty good desktop-mobile parity. Based on the the rest of this API and a quick search in Bugzilla, I suspect that the Android features landed in the same versions (or nearest first release, for pre-version 4 stuff) as the desktop browser.

```suggestion
              "version_added": "4"
```

ddbeck (changes requested):

lgommans:
@ddbeck @foolip Thanks for the help and suggestions! First time using this commit proposal feature, hope I did that correctly -- let me know if not.

It looks like the checks pass now :)

foolip (approved):
This look good to me now!

foolip:
Oh, actually, one thing I noticed. https://developer.mozilla.org/en-US/docs/Web/API/File#Browser_compatibility says that there was a non-standard implementation in versions 4-7. How do we know if the `name` property was part of that or not?

lgommans:
Personally I don't think it matters that much which version it was introduced in exactly. One of the reasons I dropped a feature from some hobby software I was writing was lacking JS filename access that, now, turns out wasn't lacking at all. (There were still other annoyances like a maximum possible file size, but this was a contributing factor.) Knowing that it's currently supported, even if we were to mark it as introduced today rather than in the past, would tell devs it'll eventually be broadly available rather than not knowing if it'll ever be implemented. Spending time figuring out exact versions for historic cases... I'm personally not sure that's really used by people that wouldn't do original research anyway.

I'd propose to just use the `<=` syntax for the version we're sure about. Shall I change it to version <=8 to avoid the non-standard implementation issue? (Or should that be <=7 then?)

foolip:
I'll have to defer to @ddbeck on how to handle this. My question in https://github.com/mdn/browser-compat-data/pull/6524#issuecomment-676737319 was for him, sorry for not being clear.

ddbeck:
@foolip 

> How do we know if the name property was part of that or not?

OK, so the truthful answer here is: I don't know. I searched Bugzilla and tried plumbing the depths of [dxr](https://dxr.mozilla.org/) to figure it out, but I couldn't successfully follow the the renames from the pre-IDL implementation to say definitively. But I also couldn't find any evidence for it being implemented later either. I ultimately based it on the desktop data (which I can't completely trust either 🤷‍♂️), on the assumption that 95% of the time, Firefox supports the same APIs across platforms.

But for features that are nearly a decade old (at worst), I'm willing to accept a little bit of uncertainty. We're probably better serving web developers by saying this landed in version 4 (or 7! it doesn't _really_ matter—we're off by at most 6 months' worth of support) than giving an ambiguous `true` value.

(It's possible to add `≤7` to the allowed versions list, but I'm opposed to doing this for one-offs.)

I hope that explains my thinking. Happy to talk about this more, if you'd like. In the meantime, I think I'm going to merge this. 😄

ddbeck (approved):
Thank you, @lgommans! 🎉

foolip:
Thanks for the explanation @ddbeck! I think that approach makes sense, and will probably end up making guesses for some very old things as well. That's pragmatic, and I'm glad it's not frowned upon :)

PR #6526, opened 2020-08-18
Burndown list: all changes from npm run confluence

foolip:
This branch/PR is in support of https://github.com/mdn/browser-compat-data/pull/6435 and not intended to be merged. Rather, I will use it as a burndown list of what changes are suggested by Confluence, to make sure we get the remaining value out of the data. I will split out PRs or revert changes on the branch until no changes remain and then close this PR.


foolip:
https://github.com/mdn/browser-compat-data/pull/6528 is the first PR I extracted from this. Confluence was correct, but a lot more checking and fixing was needed.

foolip:
I've rerun after https://github.com/mdn/browser-compat-data/pull/6555 and pushed those results. There are lots of system errors still, so I've fixed those and released mdn-confluence 2.1.0. That will get rid of all the system problems I'd previously handled by reverting some changes on the branch.

queengooborg:
This is now our oldest open PR.  Should this be updated with the latest Confluence changes?

queengooborg:
Ah, it looks like it's actually broken because of our TypeScript migration, which makes sense.  I wonder if it would be worthwhile to update `mdn-confluence`, though I've heard in passing that Confluence is going out of support?

foolip:
I'm not collecting any new data or doing releases for Web Confluence, so I think we should remove the scripts for it rather than fixing them.

foolip:
So, contrary to what I said, I went ahead and burned down this whole set of changes. There are PRs merged or open for everything now.

PR #6528, opened 2020-08-18
Update support data for Animation and KeyframeEffect

foolip:
The AnimationEffect data is already good, but Animation's effect
attribute and KeyframeEffect with constructor and target attribute
weren't updated.

These changes landed together in Chromium M75:
https://chromium-review.googlesource.com/c/chromium/src/+/1564568

The Animation and KeyframeEffect interfaces were verified to not exist 
in Edge 18, so the Edge version was set/updated to "75". These data 
points support this as well:
https://caniuse.com/#feat=web-animation
https://developer.microsoft.com/en-us/microsoft-edge/status/webanimationsjavascriptapi/

### DIFF
--- a/api/Animation.json
+++ b/api/Animation.json
@@ -25,7 +25,7 @@
             }
           ],
           "edge": {
-            "version_added": "≤79"
+            "version_added": "79"
           },
           "firefox": [
             {
@@ -106,7 +106,7 @@
               "version_added": "61"
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "79"
             },
             "firefox": [
               {
@@ -180,7 +180,7 @@
               "version_added": "39"
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "79"
             },
             "firefox": [
               {
@@ -306,7 +306,7 @@
               "version_added": "39"
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "79"
             },
             "firefox": [
               {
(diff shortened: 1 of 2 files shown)


foolip:
Note that only Chrome on Windows has been tested. The fact that the controlling flags are dropped in https://chromium-review.googlesource.com/c/chromium/src/+/1564568 is used to assume that this is in all browsers based on Chromium M75 or later, since extra work would be needed to disable it.

api/Animation.json
```diff
@@ -374,13 +374,13 @@
...
-              "version_added": false
+              "version_added": "75"
             },
             "chrome_android": {
-              "version_added": false
+              "version_added": "75"
             },
             "edge": {
-              "version_added": false
+              "version_added": "≤79"
```
foolip, line 383: I stuck with the ranged versions already used in Animation and KeyframeEffect, but have verified no support for these or AnimationEffect in Edge 18. Happy to fix the data, but would prefer to not touch too much in a single PR.

foolip:

foolip:
Thanks @vinyldarkscratch! I've made those changes and updated the description.

api/Animation.json
```diff
@@ -25,7 +25,7 @@
             }
           ],
           "edge": {
-            "version_added": "≤79"
+            "version_added": "79"
```
jpmedley, line 28: I think Florian and co. were using this value because they didn't know whether this was supported in non-Chromium Edge. What do you know that I don't?

vinyldarkscratch: As mentioned above, it was found that neither interface was supported in pre-Chrome Edge versions.￼￼

foolip: Indeed, I tested it and it's not there. The "test" is simply logging the expressions `Animation` and `KeyframeEffect`, so I didn't link to a test in the description.

PR #6530, opened 2020-08-18
Drop incorrect note about XHR via ActiveXObject for Edge

foolip:
This source of this information is mirroring in
https://github.com/mdn/browser-compat-data/pull/5568.

ActiveXObject isn't supported in Edge per archived documentation:
https://developer.mozilla.org/en-US/docs/Archive/Web/JavaScript/Microsoft_Extensions/ActiveXObject

Also tested manually, there's no ActiveXObject global.

### DIFF
--- a/api/XMLHttpRequest.json
+++ b/api/XMLHttpRequest.json
@@ -10,16 +10,9 @@
           "chrome_android": {
             "version_added": "18"
           },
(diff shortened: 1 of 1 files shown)


foolip:
@vinyldarkscratch can you review?

PR #6531, opened 2020-08-19
Update Window.json - onbeforeinstallprompt

8Thomas8:
Hi.

onbeforeinstallprompt isn't working on firefox desktop and firefox mobile. I've tested on desktop and mobile, while working on PWA.

I just changed null to false for this compatibility.

On https://developer.mozilla.org/en-US/docs/Web/API/BeforeInstallPromptEvent it's said that onbeforeinstallprompt event doesn't work with firefox desktop and mobile.

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -4584,10 +4584,10 @@
               "version_added": "≤79"
             },
             "firefox": {
-              "version_added": null
+              "version_added": false
             },
             "firefox_android": {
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Hi @8Thomas8 and welcome to BCD! Thanks for catching and correcting this inconsistency. 🎉

PR #6533, opened 2020-08-19
Mark Event constructor as supported in Edge 12

foolip:
Confirmed by running `new Event('type')` in Edge 13 in Sauce Labs. Edge
12 isn't available and therefore not tested.

### DIFF
--- a/api/Event.json
+++ b/api/Event.json
@@ -59,7 +59,7 @@
               "version_added": "18"
             },
             "edge": {
-              "version_added": "≤18"
+              "version_added": "≤13"
             },
             "firefox": {
               "version_added": "11"
--- a/api/Event.json
+++ b/api/Event.json
@@ -59,7 +59,7 @@
               "version_added": "18"
             },
             "edge": {
-              "version_added": "≤13"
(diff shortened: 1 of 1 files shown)


foolip:

foolip:
I see that this is failing the lint because ≤13 is not allowed. What then to do when one can only test back to 13? Guess that it's also in 12, which is likely?

api/Event.json
```diff
@@ -59,7 +59,7 @@
               "version_added": "18"
             },
             "edge": {
-              "version_added": "≤18"
+              "version_added": "≤13"
```
ddbeck, line 62: Looks like it was supported [from the first release](https://github.com/MicrosoftEdge/Status/pull/190)

```suggestion
              "version_added": "12"
```

ddbeck (changes requested):
Went digging about the ≤13 question and found a likely answer for you. Thanks!

foolip:
Thanks @ddbeck, I've applied the suggestion!

ddbeck (approved):
Thanks, @foolip! 🎉

PR #6534, opened 2020-08-19
Update when WeakSet.clear was removed from Chromium

foolip:
Extracted from https://github.com/mdn/browser-compat-data/pull/6526

https://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8359 was
tested in Chrome 35-44 and Opera 22-31 on Windows 8.1 on BrowserStack
to confirm, and the mobile browser data was mirrored.


foolip:
@vinyldarkscratch can you review?

PR #6535, opened 2020-08-19
Update BudgetService/BudgetState to reflect that they're gone

foolip:
Only the reserve() method was shipped, and later removed:
https://groups.google.com/a/chromium.org/d/msg/blink-dev/yBtmc-4xl_o/GE0vneAVDQAJ
https://groups.google.com/a/chromium.org/d/msg/blink-dev/18r3whCBv0I/b8qrtFTsDAAJ

https://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8362 was
tested on Chrome and Opera on Windows to verify the ranges. Android
WebView was never supported per the blink-dev intent, and the other data
was mirrored.

### DIFF
--- a/api/BudgetService.json
+++ b/api/BudgetService.json
@@ -5,34 +5,39 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/BudgetService",
         "support": {
           "chrome": {
-            "version_added": "55"
+            "version_added": "60",
+            "version_removed": "70"
           },
           "chrome_android": {
-            "version_added": "55"
+            "version_added": "60",
+            "version_removed": "70"
           },
           "edge": {
-            "version_added": "≤79"
+            "version_added": false
           },
           "firefox": {
-            "version_added": null
+            "version_added": false
           },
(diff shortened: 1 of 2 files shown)


foolip:

ddbeck:
Opened #6536 to eventually remove this outright, provided this PR checks out.

foolip:
I agree with removing for sure. Given that there are MDN pages, what's the process for doing that? First remove the pages on MDN, then the compat data? Can anyone remove pages, or can one request it somehow?

ddbeck:
Pages on MDN ought to be archived. I haven't actually done it myself though. I believe it requires a special permission to do, but Chris Mills can grant it if the person doing the work doesn't have it already. (Still a couple months off though, since the guideline calls for waiting 2 years from when all browsers stop shipping the feature.)

foolip:
I see, thanks @ddbeck! I guess BudgetState could be archived right away since it never shipped, but it doesn't matter much if it stays around until BudgetService is removed.

PR #1400, opened 2018-03-16
Update structure of BCD data for fetch interfaces

dontcallmedom:
Generated via https://github.com/dontcallmedom/webidl2mdnbcd from https://fetch.spec.whatwg.org/

PR #6543, opened 2020-08-20
Remove the api.Response.trailer entry

foolip:
It was added in https://github.com/mdn/browser-compat-data/pull/1400
because it was then in https://fetch.spec.whatwg.org/.

It was removed from the spec in https://github.com/whatwg/fetch/pull/979
and apparently never implemented.

There's no MDN page for this so nothing more to clean up.

### DIFF
--- a/api/Response.json
+++ b/api/Response.json
@@ -917,54 +917,6 @@
           }
         }
       },
-      "trailer": {
-        "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/Response/trailer",
-          "support": {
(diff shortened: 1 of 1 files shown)


foolip:
As an FYI, I'm removing an entry added here in https://github.com/mdn/browser-compat-data/pull/6543. There may well be other entries in BCD that need removing on similar grounds, which came from a change generated from spec IDL.

Elchi3 (approved):
Thanks, this is the reason why we later came up with the policy to at least have one implementation before things are allowed to be added to BCD.

PR #6544, opened 2020-08-20
Fix the Safari versions for api.Response.clone

foolip:
See https://github.com/mdn/browser-compat-data/pull/6515#issuecomment-676752274
for how this was tested in BrowserStack and Sauce, for macOS only.

Additional evidence is https://trac.webkit.org/changeset/197049/webkit,
where the whole Response interface was added, controlled by a single
flag. For the attributes/methods included in that commit, they were
almost certainly first shipped in the same version of Safari that
shipped Response and the overall Fetch API. This was 10.1 / 10.3 per
other existing compat data:
https://wiki.developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/fetch#Browser_compatibility
https://developer.mozilla.org/en-US/docs/Web/API/Request#Browser_compatibility
https://developer.mozilla.org/en-US/docs/Web/API/Response#Browser_compatibility

There's still a lot clearly wrong with the data for Safari, but this
fixes one small piece of it.

### DIFF
--- a/api/Response.json
+++ b/api/Response.json
@@ -386,10 +386,10 @@
               }
             ],
             "safari": {
-              "version_added": "11.1"
+              "version_added": "10.1"
             },
             "safari_ios": {
-              "version_added": "11.3"
+              "version_added": "10.3"
             },
             "samsunginternet_android": {
               "version_added": false


foolip:
@ddbeck @vinyldarkscratch can you review? There's so much more to fix here, but I fear that if I start pulling the threads I'll be reimplementing what Vinyl and I are working on in https://github.com/foolip/mdn-bcd-collector. I'd prefer to try and update the data with that process and see how much of the errors it can fix. (Unfortunately, the mixin issue will pop up because `Body` is a mixin...)

ddbeck (approved):

ddbeck:
Thank you, @foolip. As for this:

> There's so much more to fix here

Yeah, I get it. Don't feel like you have to fix everything all at once. I'm always happy to review a one-line PR.

PR #6546, opened 2020-08-20
Remove inaccurate notes about basic Attr support

foolip:
These notes and those for the attributes were added in
https://github.com/mdn/browser-compat-data/pull/1727.

The notes on api.Attr (basic support) don't make sense, because the
prototype chain (inheritance) of Attr never changed. What did change was
which prototype the 3 attributes localName/namespaceURI/prefix are on.

See https://github.com/mdn/browser-compat-data/issues/6545 for context.

### DIFF
--- a/api/Attr.json
+++ b/api/Attr.json
@@ -5,12 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/Attr",
         "support": {
           "chrome": {
-            "version_added": "1",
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Thank you, @foolip. I'm going to merge this to get it into today's release and I'll tackle the questions in your issue another day. 🎉

foolip:
Thanks @ddbeck!

jeanMS741:
Thanks for taking this path. Apologies for the confusion.

PR #6547, opened 2020-08-20
Add data for ContentIndex and ContentIndexEvent APIs

vinyldarkscratch:
This PR adds the data for the new [Content Indexing APIs](https://wicg.github.io/content-index/spec).  Data comes from https://www.chromestatus.com/features/5658416729030656 and https://bugs.chromium.org/p/chromium/issues/detail?id=973844.

CC @jpmedley

### DIFF
--- a/api/ContentIndex.json
+++ b/api/ContentIndex.json
@@ -4,7 +4,7 @@
       "__compat": {
         "support": {
           "chrome": {
-            "version_added": "84"
+            "version_added": false
           },
           "chrome_android": {
             "version_added": "84"
@@ -22,7 +22,7 @@
             "version_added": false
           },
           "opera": {
-            "version_added": "71"
+            "version_added": false
           },
           "opera_android": {
             "version_added": false
@@ -50,7 +50,7 @@
         "__compat": {
           "support": {
             "chrome": {
-              "version_added": "84"
+              "version_added": false
             },
             "chrome_android": {
               "version_added": "84"
@@ -68,7 +68,7 @@
               "version_added": false
             },
             "opera": {
-              "version_added": "71"
+              "version_added": false
             },
             "opera_android": {
               "version_added": false
@@ -97,7 +97,7 @@
         "__compat": {
           "support": {
             "chrome": {
-              "version_added": "84"
+              "version_added": false
             },
             "chrome_android": {
               "version_added": "84"
@@ -115,7 +115,7 @@
               "version_added": false
             },
             "opera": {
-              "version_added": "71"
+              "version_added": false
             },
             "opera_android": {
               "version_added": false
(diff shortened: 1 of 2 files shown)


vinyldarkscratch:
Marking as not ready for the time being; awaiting a response in an email thread to confirm data.

jpmedley:
Let's wait to merge this until there's an answer to my email. Because this interacts with the operating system, it's entirely possible that desktop and Android were on different timelines.

cc: @foolip

foolip:
I haven't gotten an email, FWIW.

https://chromium-review.googlesource.com/c/chromium/src/+/2181264 is the change that enabled this, and *only* on Android.

https://source.chromium.org/chromium/chromium/src/+/master:android_webview/tools/system_webview_shell/test/data/webexposed/global-interface-listing-expected.txt would suggest this isn't enabled on WebView.

That it's not supported and desktop can be quickly confirmed by just evaluating `ContentIndex` in Chrome Canary.

vinyldarkscratch:
Hmm, I had tested this using the webpage linked in the [post on web.dev](https://web.dev/content-indexing-api/), specifically https://contentindex.dev, and it seemed to operate on Chrome macOS 84.0.4147.135...

jpmedley:
@foolip I sent the email before I knew that Vinyl had opened this PR. I cc'd you here to make sure you didn't merge this.

foolip:
@vinyldarkscratch maybe there are polyfills or something involved that make the demo work?

api/ContentIndex.json
```diff
@@ -0,0 +1,192 @@
...
+            "version_added": false
+          },
+          "ie": {
+            "version_added": false
+          },
+          "opera": {
+            "version_added": false
+          },
+          "opera_android": {
+            "version_added": false
```
foolip, line 28: It's likely it will be in Opera for Android 60. How can we remember to update this later?

api/ContentIndex.json
```diff
@@ -0,0 +1,192 @@
...
+          },
+          "safari": {
+            "version_added": false
+          },
+          "safari_ios": {
+            "version_added": false
+          },
+          "samsunginternet_android": {
+            "version_added": false
```
foolip, line 37: This too is likely to be supported soon.

api/ContentIndex.json
foolip: I've filed https://github.com/mdn/browser-compat-data/issues/6601

api/ContentIndex.json
foolip: I've filed https://github.com/mdn/browser-compat-data/issues/6602

foolip (approved):

PR #6548, opened 2020-08-20
Update Microsoft Edge support for forced-colors media feature

melanierichards:
Update `forced-colors` media query support for Microsoft Edge, which shipped awhile back in 79.

For verification: our [feature availability announcement](https://www.microsoftedgeinsider.com/en-us/whats-new) for whatever reason does not have a version number on it, but I'm part of the feature team that contributed towards support for forced color modes in Microsoft Edge and in web standards.

### DIFF
--- a/css/at-rules/media.json
+++ b/css/at-rules/media.json
@@ -604,8 +604,7 @@
                 "notes": "See <a href='https://crbug.com/970285'>bug 970285</a>."
               },
               "edge": {
-                "version_added": false,
-                "notes": "See <a href='https://crbug.com/970285'>bug 970285</a>."
+                "version_added": "79"
               },
               "firefox": {
                 "version_added": "81",
--- a/css/at-rules/media.json
+++ b/css/at-rules/media.json
@@ -596,7 +596,14 @@
             "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/@media/forced-colors",
             "support": {
               "chrome": {
-                "version_added": false,
+                "version_added": "79",
+                "flags": [
+                  {
+                    "type": "preference",
+                    "name": "#forced-colors",
+                    "value_to_set": "Enabled"
+                  }
+                ],
                 "notes": "See <a href='https://crbug.com/970285'>bug 970285</a>."
               },
               "chrome_android": {


css/at-rules/media.json
```diff
@@ -596,16 +596,22 @@
             "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/@media/forced-colors",
             "support": {
               "chrome": {
-                "version_added": false,
+                "version_added": "79",
+                "flags": [
```
ddbeck, line 600: Thank you very much for the update (and bringing in additional reviewers). One request here, could you duplicate this data in chrome_android below? I checked and this pref appears to be supported on both desktop and Android. Thanks again!

ddbeck (changes requested):

css/at-rules/media.json
ddbeck: Hi @melanierichards, do you plan to return to this PR? If not, let me know—we can take this over

melanierichards: Sorry @ddbeck, got pulled into some other tasks during the feedback loop!

I think it might be better not to add chrome_android. Though the flag for forced color modes is technically supported on all platforms, it wouldn't be meaningful on platforms which don't have a forced color mode (like Windows High Contrast) to trigger the media query. Might be confusing to authors if we list it as supported behind a flag.

ddbeck: Understood. Thank you for coming back to this, @melanierichards. I'll merge this shortly! 🎉

ddbeck (approved):

css/at-rules/media.json
melanierichards: Thank you!! And thanks for your patience :)

css/at-rules/media.json
```diff
@@ -596,16 +596,22 @@
...
+                  {
+                    "type": "preference",
+                    "name": "#forced-colors",
+                    "value_to_set": "Enabled"
+                  }
+                ],
                 "notes": "See <a href='https://crbug.com/970285'>bug 970285</a>."
```
vsn4ik, line 607: Still need a note?

melanierichards: I think this is good to have until it's out from behind a flag in Chrome :)

PR #6554, opened 2020-08-25
Added in support for nowrap in white-space CSS

Yash-Singh1:
# 
# Overview
This resolves issue #6389 
It doesn't resolve it completely just adding in `nowrap`
# Data
Data was received from w3schools: https://www.w3schools.com/cssref/pr_text_white-space.asp
Also, other data was received from css-tricks.com: https://css-tricks.com/almanac/properties/w/whitespace/
# Issue
The issue was that there were => "Missing white-space values"
The missing values were `white-space`, `normal`, and `nowrap`. The issue is tagged as `data:css`.
Here is the issue: #6389 
# Testing
I tested and all browsers listed support the `white-space`'s `nowrap` option.

# Checklist
- [x] Summarize your changes
- [x] Link to related issues or pull requests if any
#

### DIFF
--- a/css/properties/white-space.json
+++ b/css/properties/white-space.json
@@ -39,7 +39,7 @@
               "version_added": "1.0"
             },
             "webview_android": {
-              "version_added": "37"
+              "version_added": "1"
             }
           },
           "status": {
--- a/css/properties/white-space.json
+++ b/css/properties/white-space.json
@@ -134,7 +134,7 @@
                 "version_added": "1.0"
               },
               "webview_android": {
-                "version_added": "37"
+                "version_added": "1"
               }
             },
             "status": {
--- a/css/properties/white-space.json
+++ b/css/properties/white-space.json
@@ -134,7 +134,7 @@
                 "version_added": "1.0"
               },
               "webview_android": {
-                "version_added": "1"
+                "version_added": "37"
               }
             },
             "status": {
--- a/css/properties/white-space.json
+++ b/css/properties/white-space.json
@@ -110,31 +110,31 @@
                 "version_added": "12"
               },
               "firefox": {
-                "version_added": "3.5"
+                "version_added": "1"
               },
               "firefox_android": {
                 "version_added": "4"
               },
               "ie": {
-                "version_added": "8"
+                "version_added": "5.5"
               },
               "opera": {
-                "version_added": "9.5"
+                "version_added": "4"
               },
               "opera_android": {
-                "version_added": "46"
+                "version_added": "10.1"
               },
               "safari": {
-                "version_added": "3"
+                "version_added": "1"
               },
               "safari_ios": {
-                "version_added": "3"
+                "version_added": "1"
               },
               "samsunginternet_android": {
-                "version_added": "11.2"
+                "version_added": "1.0"
               },
               "webview_android": {
-                "version_added": "37"
+                "version_added": "1"
               }
             },
             "status": {
--- a/css/properties/white-space.json
+++ b/css/properties/white-space.json
@@ -96,6 +96,54 @@
             }
           }
         },
+        "nowrap": {
+          "__compat": {
+            "description": "<code>nowrap</code>",
+            "support": {
+              "chrome": {
+                "version_added": "1"
+              },
+              "chrome_android": {
+                "version_added": "18"
+              },
+              "edge": {
+                "version_added": "12"
+              },
+              "firefox": {
+                "version_added": "3.5"
+              },
+              "firefox_android": {
+                "version_added": "4"
+              },
+              "ie": {
+                "version_added": "8"
+              },
+              "opera": {
+                "version_added": "9.5"
+              },
+              "opera_android": {
+                "version_added": "9.5"
+              },
+              "safari": {
+                "version_added": "3"
+              },
+              "safari_ios": {
+                "version_added": "3"
+              },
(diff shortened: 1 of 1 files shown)


Yash-Singh1:
Made a mistake in the information, the tests should work now ...

css/properties/white-space.json
```diff
@@ -96,6 +96,54 @@
...
+                "version_added": "1"
+              },
+              "chrome_android": {
+                "version_added": "18"
+              },
+              "edge": {
+                "version_added": "12"
+              },
+              "firefox": {
+                "version_added": "3.5"
```
ddbeck, line 113: There's some evidence that this predates Firefox 1 https://bugzilla.mozilla.org/show_bug.cgi?id=45621

```suggestion
                "version_added": "1"
```

css/properties/white-space.json
```diff
@@ -96,6 +96,54 @@
...
+              },
+              "firefox_android": {
+                "version_added": "4"
+              },
+              "ie": {
+                "version_added": "8"
```
ddbeck, line 119: It looks like this showed up a bit earlier too. For example, some bug reports:

https://www.quirksmode.org/bugreports/archives/2005/02/White_space_IE55_bug.html
https://www.quirksmode.org/bugreports/archives/2005/07/Wrapped_inlinenowrap_elements_lose_padding_left_in.html

I think it's possible it showed up even earlier, but I wasn't able to find any mentions of `nowrap` that wasn't the HTML attribute for IEs before version 5.5. This is also the version for the property as a whole (the parent feature). So let's go with that:

```suggestion
                "version_added": "5.5"
```

css/properties/white-space.json
```diff
@@ -96,6 +96,54 @@
...
+              },
+              "opera": {
+                "version_added": "9.5"
+              },
+              "opera_android": {
+                "version_added": "46"
```
ddbeck, line 125: Since `nowrap` long predates Android itself, I feel safe putting this one as Opera's first release:

```suggestion
                "version_added": "10.1"
```

css/properties/white-space.json
```diff
@@ -96,6 +96,54 @@
...
+              },
+              "safari": {
+                "version_added": "3"
```
ddbeck, line 128: Safari's first release is also sometime after this property was standardized. All of the WebKit bugs I could find for `nowrap` were about bugs in the implementation, but none of them seemed to indicate the feature was in any way new. Seems safe to set this as the first release as well:

```suggestion
                "version_added": "1"
```

css/properties/white-space.json
```diff
@@ -96,6 +96,54 @@
...
+              },
+              "safari_ios": {
+                "version_added": "3"
```
ddbeck, line 131: Safari for iOS also comes later, so it's certainly had it since inception

```suggestion
                "version_added": "1"
```

css/properties/white-space.json
```diff
@@ -96,6 +96,54 @@
...
+              },
+              "samsunginternet_android": {
+                "version_added": "11.2"
```
ddbeck, line 134: Since Chrome has always had this feature and Samsung Internet is derived from Chrome, we can go with the first release here too!

```suggestion
                "version_added": "1.0"
```

css/properties/white-space.json
```diff
@@ -96,6 +96,54 @@
...
+              },
+              "webview_android": {
+                "version_added": "37"
```
ddbeck, line 137: Since Android's original browser was WebKit derived, this was also almost certainly supported from the first release:

```suggestion
                "version_added": "1"
```

css/properties/white-space.json
```diff
@@ -96,6 +96,54 @@
...
+                "version_added": "3.5"
+              },
+              "firefox_android": {
+                "version_added": "4"
+              },
+              "ie": {
+                "version_added": "8"
+              },
+              "opera": {
+                "version_added": "9.5"
```
ddbeck, line 122: The CSS-Tricks page shows version 4, which is a plausible date compared to the others. This is also the first version of Opera to support the property in general, some time after it was standardized.

```suggestion
                "version_added": "4"
```

ddbeck (changes requested):
Hi @Yash-Singh1 and welcome to BCD! This is a really nice start. Because this feature has been around a long time, it's a bit tougher to research than most. I've made some suggestions to change a few of these values to earlier versions. Please take a look. Thank you!

Yash-Singh1:
One test has failed. I think it is saying that `nowrap` can't be added before `white-space`. Should we shift `nowrap` to the release of `white-space`, 37 in `webview_android`?

Yash-Singh1:
Alright, all the checks have succeeded.

css/properties/white-space.json
ddbeck: Ah, I see what you mean here. I think the correct thing to do is change both this version to `"1"` and the parent feature to `"1"` (around line 42)

Yash-Singh1: Sure

Yash-Singh1:
Ok, I changed the parent and child feature.

Yash-Singh1:
I don't see a review, is there any review which you are requesting?

ddbeck (approved):
Thank you for the updates, @Yash-Singh1! Looks great. 🎉 

(I think the review you saw was my stale, original review.)

PR #6558, opened 2020-08-25
Make the 'subfeature_earlier_implementation' lint stricter

foolip:
Previously it wouldn't compare versions if either was a ranged value
starting with '≤'. However, '≤18' is definitely earlier than '79', to
take Edge numbers as an example, so the comparison can be made in one
direction.

There are no violations of this tighter check.


alshamiri5 (approved):

foolip:
Thanks for putting together some tests @ddbeck, looks good, and I'm happy to review in detail later.

PR #6559, opened 2020-08-25
Mark APIs (interfaces) as suppported in Edge 12 if any subfeature is

foolip:
For all of these, some subfeature is already marked as supported in Edge
12. The feature itself must then also be, although the interface object
might not be exposed.

This is a mechanical change, none of them were verified in Edge. If the
subfeature support statement is incorrect, the main feature will now
also be incorrectly marked as supported.

### DIFF
--- a/api/AnalyserNode.json
+++ b/api/AnalyserNode.json
@@ -11,7 +11,7 @@
             "version_added": "18"
           },
           "edge": {
-            "version_added": "≤18"
+            "version_added": "12"
           },
           "firefox": {
             "version_added": "25"
--- a/api/AudioBuffer.json
+++ b/api/AudioBuffer.json
@@ -11,7 +11,7 @@
             "version_added": "18"
           },
           "edge": {
-            "version_added": "≤18"
+            "version_added": "12"
           },
           "firefox": {
             "version_added": "25"
--- a/api/AudioBufferSourceNode.json
+++ b/api/AudioBufferSourceNode.json
@@ -11,7 +11,7 @@
             "version_added": "18"
           },
           "edge": {
-            "version_added": "≤18"
+            "version_added": "12"
           },
           "firefox": {
             "version_added": "25"
--- a/api/AudioContext.json
+++ b/api/AudioContext.json
@@ -25,7 +25,7 @@
             }
           ],
           "edge": {
-            "version_added": "≤18"
+            "version_added": "12"
           },
           "firefox": {
             "version_added": "25"
--- a/api/AudioDestinationNode.json
+++ b/api/AudioDestinationNode.json
@@ -11,7 +11,7 @@
             "version_added": "18"
           },
           "edge": {
-            "version_added": "≤18"
+            "version_added": "12"
           },
           "firefox": {
             "version_added": "25"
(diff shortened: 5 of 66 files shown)


foolip:
@ddbeck this is a fairly mechanical change, can you review? Re-verifying all of the information seems like too much effort to be worth it, but is there some part that would be worth verifying to be more confident in this?

ddbeck:
This is a good idea, @foolip! I agree, that where we're inferring data points based on data we've already got, we shouldn't validate them one-by-one. Our data already implies that this is true and this is merely making it explicit.

Before I actually review this, a quick point of order: did you do this by hand or did you write a script? If you did it by hand, I'll just review it. But if you did it with a script, I'd like to alert you to our ["migrations" process](https://github.com/mdn/browser-compat-data/blob/master/docs/migrations.md).

We've tried this a few times before and it works really well for mechanistic changes like this. The main benefit is that we sort of abstract the review a bit: instead of checking to make sure that each individual change is correct, we review the script and presume that the application of the script is correct (after a spot check or two).

Like I said though, if you did this by hand, I can do a direct review (though I would advise against doing anything along these lines that results in a diff much bigger than this one).

foolip:
Wow, https://github.com/mdn/browser-compat-data/blob/master/docs/migrations.md is great stuff, I'm glad to be working in a project taking such a deliberate approach to large-scale changes.

This change I'm afraid wasn't the result of a migration script, but from locally tweaking the lint in a way similar to https://github.com/mdn/browser-compat-data/pull/6558 to just strip '≤' from both versions to see what it would flag. This is the diff against the *current* code:
```diff
--- a/test/linter/test-consistency.js
+++ b/test/linter/test-consistency.js
@@ -180,6 +180,9 @@ class ConsistencyChecker {
 
     subfeatures.forEach(subfeature => {
       supportInParent.forEach(browser => {
+        if (browser !== 'edge') {
+          return;
+        }
         if (
           data[subfeature].__compat.support[browser] != undefined &&
           this.isVersionAddedGreater(
@@ -319,10 +322,11 @@ class ConsistencyChecker {
       typeof a_version_added === 'string' &&
       typeof b_version_added === 'string'
     ) {
-      if (a_version_added.startsWith('≤') || b_version_added.startsWith('≤')) {
-        return false;
-      }
-      return compareVersions.compare(a_version_added, b_version_added, '<');
+      return compareVersions.compare(
+        a_version_added.replace('≤', ''),
+        b_version_added.replace('≤', ''),
+        '<',
+      );
     }
 
     return false;
```

Not all of the resulting errors are real, but it did reveal this class of problems for Edge which can be fixed. I went through the lint errors by hand and fixed the ones involving Edge 12. There is some chance of manual error because of this.

If you prefer, I can still do this as a migration.

ddbeck (approved):
Thanks for the background on this, @foolip! I don't think you need to go through the migration process for this. I checked a most of these carefully (and confirmed that all of them only rewrite `≤18` entries), so I'm reasonably confident that this a smart change. I'm going to merge. Thank you!

(Also, I'm about to wrap things up for the day, so this is likely to be my last review until after the Monday bank holiday in the UK.)

PR #6563, opened 2020-08-26
Fix data for File() constructor

bodqhrohro:
I tested it myself in Opera 12 and Opera Mini; the constructor throws `DomException: NOT_SUPPORTED_ERR`

The addition of the API was mentioned in [Opera 25 release notes](https://dev.opera.com/blog/opera-25), so it wasn't available in earlier Chromium-based versions of Opera as well.

[Can I use...](https://caniuse.com/#feat=fileapi) seems to have more clarified data, so I copied them there, for Opera and some other browsers too, as the MDN data for the constructor seem to be just derived from the whole File API.

### DIFF
--- a/api/File.json
+++ b/api/File.json
@@ -67,28 +67,28 @@
           "description": "<code>File()</code> constructor",
           "support": {
             "chrome": {
-              "version_added": "13"
+              "version_added": "38"
             },
             "chrome_android": {
-              "version_added": "18"
+              "version_added": "38"
             },
             "edge": {
               "version_added": "79"
             },
             "firefox": {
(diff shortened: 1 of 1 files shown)


ddbeck (changes requested):
Hi @bodqhrohro and welcome to BCD! This is a great PR—thank you for updating this info. I've got two suggestions to extend this that are outside of GitHub diff range:

* The Chromium value also implies Samsung Internet `"3.0"`
* The [Chrome status entry](https://www.chromestatus.com/feature/5731244088229888) says that WebView is `"38"` as well

Would you mind making these changes too? Thank you again!

PR #6564, opened 2020-08-26
Fix Chrome versions for DOMTokenList API

vinyldarkscratch:
This PR fixes version numbers for the DOMTokenList API based upon manual testing.  The API was actually implemented in Chrome 8; this PR corrects the data to match.  All the features listed as supported in Chrome 1 were tested in Chrome 8 and confirmed functional.

Found during the creation of PRs from mdn-bcd-collector.

### DIFF
--- a/api/DOMTokenList.json
+++ b/api/DOMTokenList.json
@@ -5,7 +5,7 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/DOMTokenList",
         "support": {
           "chrome": {
-            "version_added": "1"
+            "version_added": "8"
           },
           "chrome_android": {
             "version_added": "18"
@@ -52,10 +52,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/DOMTokenList/add",
           "support": {
             "chrome": {
-              "version_added": "1"
+              "version_added": "8"
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": "18"
             },
             "edge": {
               "version_added": "12"
@@ -82,10 +82,10 @@
               "version_added": "5.1"
             },
             "samsunginternet_android": {
-              "version_added": true
+              "version_added": "1.0"
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "≤37"
             }
           },
           "status": {
@@ -148,10 +148,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/DOMTokenList/contains",
           "support": {
             "chrome": {
-              "version_added": "1"
+              "version_added": "8"
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": "18"
             },
             "edge": {
               "version_added": "12"
@@ -178,10 +178,10 @@
               "version_added": "5.1"
             },
             "samsunginternet_android": {
-              "version_added": true
+              "version_added": "1.0"
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "≤37"
             }
           },
           "status": {
@@ -292,10 +292,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/DOMTokenList/item",
           "support": {
             "chrome": {
-              "version_added": "1"
+              "version_added": "8"
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": "18"
             },
             "edge": {
               "version_added": "12"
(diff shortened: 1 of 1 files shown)


vinyldarkscratch:
Sure thing, fixed!

foolip:
@vinyldarkscratch it would make review easier if you could put some more details into the commit message about what you tested, even when testing manually. Here I suspected (incorrectly) that originally the `DOMTokenList` interface object wasn't exposed, or maybe even was called something else. I wrote http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8382 to check it out.

This was implemented in WebKit in https://trac.webkit.org/changeset/68440/webkit. That was WebKit version 534.10, verified by the process in https://wiki.developer.mozilla.org/en-US/docs/MDN/Contribute/Processes/Matching_features_to_browser_version#Safari (which I just updated).

Chrome 8 was using precisely WebKit 534.10, so this checks out.

api/DOMTokenList.json
```diff
@@ -38,7 +38,7 @@
             "version_added": "1.0"
           },
           "webview_android": {
-            "version_added": "1"
+            "version_added": "≤37"
```
foolip, line 41:
```suggestion
            "version_added": "3"
```

Based on the fact that WebView 3 uses WebKit 534.13.

api/DOMTokenList.json
```diff
@@ -82,10 +82,10 @@
               "version_added": "5.1"
             },
             "samsunginternet_android": {
-              "version_added": true
+              "version_added": "1.0"
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "≤37"
```
foolip, line 88:
```suggestion
              "version_added": "3"
```

api/DOMTokenList.json
```diff
@@ -178,10 +178,10 @@
...
```
foolip, line 184:
```suggestion
              "version_added": "3"
```

api/DOMTokenList.json
```diff
@@ -322,10 +322,10 @@
...
```
foolip, line 328:
```suggestion
              "version_added": "3"
```

api/DOMTokenList.json
```diff
@@ -418,10 +418,10 @@
...
```
foolip, line 424:
```suggestion
              "version_added": "3"
```

api/DOMTokenList.json
```diff
@@ -466,10 +466,10 @@
...
```
foolip, line 472:
```suggestion
              "version_added": "3"
```

api/DOMTokenList.json
```diff
@@ -759,10 +759,10 @@
...
```
foolip, line 765:
```suggestion
              "version_added": "3"
```

foolip:
Another thing I notice here is the inconsistency between api.DOMTokenList and api.Element.classList. These were introduced at the same time.

foolip:
I've filed https://github.com/mdn/browser-compat-data/issues/6571 about a possible lint that would have flagged inconsistency between api.DOMTokenList and api.Element.classList.

foolip:
Another little wrinkle. https://developer.mozilla.org/en-US/docs/Web/API/Element/classList#Browser_compatibility notes that this was originally not supported for SVG elements in Internet Explorer and Edge. This was actually the case in WebKit as well, and was fixed in https://trac.webkit.org/changeset/125092/webkit. That was WebKit 537.4, which ought to have been Chrome 22, Safari 6.1. and Safari for iOS 7.

I don't really think it'd be valuable to document this, but what should the criteria be for something like this?

api/DOMTokenList.json
vinyldarkscratch: May I ask where you sourced the information about WebView's engine versions?

foolip: I figured it out from https://wiki.developer.mozilla.org/en-US/docs/MDN/Contribute/Processes/Matching_features_to_browser_version#Safari and updated that a bit in the process. Basically, I have a checkout of WebKit where I can find the right commit easily using `git log -S` and other git tricks. Given the commit, get the approximate WebKit version from Source/WebCore/Configurations/Version.xcconfig, and from there determine if the right releases can be determined or if the exact WebKit version has to be found. That last bit I haven't done, it would involve stepping through branches and looking at the content of each. Would be much faster with the branches in a local checkout, but I don't have that.

foolip: Oh, and of course, with this approach one has to check if the feature was behind an ifdef or a runtime flag. If it is then I wouldn't have confidence in determining the release for Safari. (I do know how Chromium controls features well enough to do this though.)

foolip (approved):

foolip:
Thanks for the fixes, merging this now!

PR #6565, opened 2020-08-26
Exercise new labeler

ddbeck:
Don't merge this PR. It's for verification of #6522. Success is labels are automatically applied; failure is an error in the triage workflow.


ddbeck:
Success! 🎉

PR #4902, opened 2019-09-30
Update DOMMatrix[ReadOnly] Chromium compatibility

vinyldarkscratch:
This PR intends to fix some inconsistencies within the `DOMMatrix` and `DOMMatrixReadOnly` data for Chromium.  As I found in #3743, both APIs were implemented in Chrome 45, however in Chrome 60, it's reported as undefined.  And I found out why:

DOMMatrix and DOMMatrixReadOnly were implemented in [Chrome 45](https://storage.googleapis.com/chromium-find-releases-static/588.html#5889261bab473ef399160e39c6b48d5eb21b4499) behind a flag (`--enable-blink-features=GeometryInterfaces`).  The flag was then enabled by default in [Chrome 61](https://storage.googleapis.com/chromium-find-releases-static/065.html#06592e080fd1cf4a188265ed4f9bcf826952671a).

DOMMatrix was available to workers in [Chrome 57](https://storage.googleapis.com/chromium-find-releases-static/0b4.html#0b469432cc5bd5db8422ce5a8100f8fcae148a2c).

PR #6566, opened 2020-08-26
Update Chromium versions for DOMRect + DOMRectReadOnly

vinyldarkscratch:
This PR updates the Chromium versions for DOMRect and DOMRectReadOnly.  When working on results from mdn-bcd-collector. I found some discrepancies with the reported data, which said the interfaces were implemented in 61, even though its subfeatures mentioned earlier implementation.  I remember a similar situation in #4902, so I tracked it down, finding it was the same case.  This API was implemented in [Chrome 45](https://storage.googleapis.com/chromium-find-releases-static/cbf.html#cbfb8e068dc4f423f3c710ce0dd2a96caddb6d16) behind a runtime flag, of which was enabled by default in Chrome 61.  This, and the subfeatures mentioned below, were proven in testing via SauceLabs (with a little tweaking to enable the flag).

Sources:

https://storage.googleapis.com/chromium-find-releases-static/cbf.html#cbfb8e068dc4f423f3c710ce0dd2a96caddb6d16 (Chrome 45)
api.DOMRect
api.DOMRect.DOMRect
api.DOMRectReadOnly
api.DOMRectReadOnly.DOMRectReadOnly
api.DOMRectReadOnly.x
api.DOMRectReadOnly.y
api.DOMRectReadOnly.width
api.DOMRectReadOnly.height
api.DOMRectReadOnly.top
api.DOMRectReadOnly.right
api.DOMRectReadOnly.bottom
api.DOMRectReadOnly.left

### DIFF
--- a/api/DOMRectReadOnly.json
+++ b/api/DOMRectReadOnly.json
@@ -157,10 +157,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/DOMRectReadOnly/bottom",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "45"
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": "45"
             },
             "edge": {
               "version_added": "79"
@@ -176,10 +176,10 @@
               "notes": "Implemented on the proprietary <code><a href='https://msdn.microsoft.com/en-us/library/hh826029(VS.85).aspx'>ClientRect</a></code> interface."
             },
             "opera": {
-              "version_added": true
+              "version_added": "32"
             },
             "opera_android": {
-              "version_added": true
+              "version_added": "32"
             },
             "safari": {
               "version_added": "10.1"
@@ -188,10 +188,10 @@
               "version_added": "10.3"
             },
             "samsunginternet_android": {
-              "version_added": true
+              "version_added": "8.0"
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "61"
             }
           },
           "status": {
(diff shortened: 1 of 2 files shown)


foolip:
I'm afraid that https://chromium.googlesource.com/chromium/src/+/cbfb8e068dc4f423f3c710ce0dd2a96caddb6d16 is not the commit that shipped these APIs. Note that `[RuntimeEnabled=GeometryInterfaces]` in the IDL files.

Rather, https://chromium.googlesource.com/chromium/src/+/06592e080fd1cf4a188265ed4f9bcf826952671a is the commit that enabled it, and that was shipped in M61.

You can see the full set of interfaces and their members that were enabled together in https://chromium.googlesource.com/chromium/src/+/06592e080fd1cf4a188265ed4f9bcf826952671a%5E%21/#F10.

foolip:
Sorry! I only saw the link in your description and started digging, but you already pointed out how this came about. Linking to the commit that enabled in instead or in addition would be good though.

api/DOMRectReadOnly.json
```diff
@@ -65,16 +65,16 @@
...
             "chrome_android": {
-              "version_added": true
+              "version_added": "61"
             },
             "edge": {
               "version_added": "79"
             },
             "firefox": {
-              "version_added": true
+              "version_added": "62"
```
foolip, line 77: This changes Firefox but isn't mentioned in the description. Was this a mistake?

vinyldarkscratch: Yeah, that was a mistake, probably during the cherry-picking from the collector.  Reverting!

foolip (approved):

PR #4823, opened 2019-09-13
Update DOMMatrix for Firefox 69

a2sheppy:
Update DOMMatrix* for Firefox 69

Lots of changes to these files.
* Note `scaleNonUniform()` and `scaleNonUniformSelf()` removed in Firefox 69 (source: [bug 1560119](https://bugzilla.mozilla.org/show_bug.cgi?id=1560119)
* Added missing `fromMatrix()` method. Sources: [bug 1560462](https://bugzilla.mozilla.org/show_bug.cgi?id=1560462) for Firefox, [WebKit bug 110001](https://bugs.webkit.org/show_bug.cgi?id=110001)
* Note that Firefox correctly labels certain parameters as optional now (source: [bug 1397949](https://bugzilla.mozilla.org/show_bug.cgi?id=1397949)).
* Note updated syntax support for `scale()` and `scale3d()` (Source: [bug 1397945](https://bugzilla.mozilla.org/show_bug.cgi?id=1397945))
* Note updated parameter list for `translate()` (Source: bug 1397949](https://bugzilla.mozilla.org/show_bug.cgi?id=1397949))
* Note support for workers added (Source: [bug 1420580](https://bugzilla.mozilla.org/show_bug.cgi?id=1420580))
* Add descriptions with formatted labels for all functions

PR #6568, opened 2020-08-26
Results from mdn-bcd-collector for DOM* APIs

vinyldarkscratch:
This PR provides results collected from [mdn-bcd-collector](https://github.com/foolip/mdn-bcd-collector) for the `DOM*` APIs.  All versions have been verified manually in SauceLabs and through searching Chromium Code Search to verify the commits.  (All invalid data was removed from this PR and either corrected on the collector or separated into a new PR.)


api/DOMRectReadOnly.json
```diff
@@ -74,7 +74,7 @@
               "version_added": "79"
             },
             "firefox": {
-              "version_added": true
+              "version_added": "62"
```
foolip, line 77: Why is only the constructor updated here? I think it'd be better to split out the Firefox changes for geometry APIs into its own PR for review by @ddbeck.

api/DOMStringList.json
```diff
@@ -148,7 +148,7 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/DOMStringList/length",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "44"
```
foolip, line 151: This is a tricky one. `DOMStringList` is legacy by now, but is used for `location.ancestorOrigins` and 3 different IndexedDB attributes, `objectStoreNames`, `indexNames`, and `objectStoreNames`.

I think some of those were available before Chrome 44, so I doubt this change. That this is close to Chrome 43 also makes me suspicious, see https://github.com/mdittmer/mdn-confluence/commit/5e173b18caedd5dd4aec38675712d514c9fe4784.

Can you dig around a bit?

api/DOMStringList.json
```diff
@@ -172,7 +172,7 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "10"
```
foolip, line 175: I doubt this one too, can you provide details?

api/DOMTokenList.json
```diff
@@ -196,13 +196,13 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/DOMTokenList/entries",
           "support": {
             "chrome": {
-              "version_added": "45"
+              "version_added": "42"
```
foolip, line 199:
```suggestion
              "version_added": "45"
```

45 was correct, see https://storage.googleapis.com/chromium-find-releases-static/734.html#734997b4ae07d305c47087ada6754a0bf8597a9e and follow links to https://chromium.googlesource.com/chromium/src/+/734997b4ae07d305c47087ada6754a0bf8597a9e%5E%21/#F4 to see what was introduced in M45.

Can you file a bug for what went wrong with the pipeline here?

api/DOMTokenList.json
```diff
@@ -196,13 +196,13 @@
...
             },
             "chrome_android": {
               "version_added": "45"
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "16"
```
foolip, line 205: Nice! I won't reconfirm this since you've already tested in Sauce.

api/DOMTokenList.json
```diff
@@ -220,7 +220,7 @@
               "version_added": "32"
             },
             "safari": {
-              "version_added": "10"
+              "version_added": "10.1"
```
foolip, line 223: `iterable<DOMString>` is from https://trac.webkit.org/changeset/203728/webkit, WebKit 603.1.1. That confirms this change. But safari_ios should then be "10.3". I'll suggest edits.

api/DOMTokenList.json
```diff
@@ -244,7 +244,7 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/DOMTokenList/forEach",
           "support": {
             "chrome": {
-              "version_added": "45"
+              "version_added": "42"
```
foolip, line 247:
```suggestion
              "version_added": "45"
```

api/DOMTokenList.json
```diff
@@ -340,13 +340,13 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/DOMTokenList/keys",
           "support": {
             "chrome": {
-              "version_added": "45"
+              "version_added": "42"
```
foolip, line 343:
```suggestion
              "version_added": "45"
```

api/DOMTokenList.json
```diff
@@ -879,13 +879,13 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/DOMTokenList/values",
           "support": {
             "chrome": {
-              "version_added": "45"
+              "version_added": "42"
```
foolip, line 882:
```suggestion
              "version_added": "45"
```

api/DOMTokenList.json
```diff
@@ -220,7 +220,7 @@
               "version_added": "32"
             },
             "safari": {
-              "version_added": "10"
+              "version_added": "10.1"
             },
             "safari_ios": {
               "version_added": "10"
```
foolip, line 226:
```suggestion
              "version_added": "10.3"
```

api/DOMTokenList.json
```diff
@@ -268,7 +268,7 @@
...
```
foolip, line 274: 10.3

api/DOMTokenList.json
```diff
@@ -364,7 +364,7 @@
...
```
foolip, line 370:
```suggestion
              "version_added": "10.3"
```

api/DOMTokenList.json
```diff
@@ -705,7 +705,7 @@
               "version_added": "36"
             },
             "safari": {
-              "version_added": "10"
+              "version_added": "10.1"
             },
             "safari_ios": {
               "version_added": "10"
```
foolip, line 711:
```suggestion
              "version_added": "10.3"
```

api/DOMTokenList.json
```diff
@@ -833,10 +833,10 @@
               "notes": "Before Chrome 50, this property was part of the deprecated child <code>DOMSettableTokenList</code> interface."
             },
             "edge": {
-              "version_added": "≤18"
+              "version_added": "17"
```
foolip, line 836: Trusting this change. Just a note for anyone following along that `value` is not `values`, it's not implied by `iterable<T>` in IDL. It's not strange that it was added separately.

api/DOMTokenList.json
```diff
@@ -833,10 +833,10 @@
...
             },
             "firefox": {
-              "version_added": true
+              "version_added": "47"
```
foolip, line 839: Can you mirror this to firefox_android?

api/DOMTokenList.json
```diff
@@ -903,7 +903,7 @@
               "version_added": "32"
             },
             "safari": {
-              "version_added": "10"
+              "version_added": "10.1"
             },
             "safari_ios": {
               "version_added": "10"
```
foolip, line 909:
```suggestion
              "version_added": "10.3"
```

api/DOMException.json
```diff
@@ -77,7 +77,7 @@
               "version_added": true
             },
             "safari": {
-              "version_added": null
+              "version_added": "10.1"
             },
             "safari_ios": {
               "version_added": null
```
foolip, line 83:
```suggestion
              "version_added": "10.3"
```

api/DOMPointReadOnly.json
```diff
@@ -175,7 +175,7 @@
               "version_added": "45"
             },
             "safari": {
-              "version_added": "10.1"
+              "version_added": "11"
             },
             "safari_ios": {
               "version_added": "10.3"
```
foolip, line 181:
```suggestion
              "version_added": "11"
```

api/DOMQuad.json
```diff
@@ -424,7 +424,7 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "11"
             },
             "safari_ios": {
               "version_added": true
```
foolip, line 430:
```suggestion
              "version_added": "11"
```

api/DOMException.json
```diff
@@ -53,7 +53,7 @@
           "description": "<code>DOMException()</code> constructor",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "46"
```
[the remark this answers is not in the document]
vinyldarkscratch: Sure thing, I'll perform mirroring along with the process!

api/DOMException.json
```diff
@@ -101,7 +101,7 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/DOMException/code",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "43"
```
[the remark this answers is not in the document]
vinyldarkscratch: Hmm, this one's a bit odd.  When testing in a few different Chrome versions (1, 26, and 42), the `code` variable is `undefined` and not present in the prototype.

Screen Shot 2020-08-28 at 08 20 24
Screen Shot 2020-08-28 at 08 19 26
Screen Shot 2020-08-28 at 08 09 51
Screen Shot 2020-08-28 at 07 43 05

However, the attribute __is__ present in Chrome's source since the initial implementation of `DOMException` -- see https://source.chromium.org/chromium/chromium/src/+/8d75ff802c1d2d66f5224851415b955343a18c42:third_party/WebKit/WebCore/dom/DOMCoreException.idl;bpv=0.  I'm wondering if it's `undefined` upon initialization, one of those cases that makes testing quite frustrating...

api/DOMMatrixReadOnly.json
```diff
@@ -62,7 +62,7 @@
               "version_added": "79"
             },
             "firefox": {
-              "version_added": "33"
+              "version_added": "62"
```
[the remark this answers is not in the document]
vinyldarkscratch: Looks like the old data was just imported from the old wiki tables.  From what I can tell, the constructor is indeed invalid until Firefox 62.

api/DOMMatrixReadOnly.json
```diff
@@ -1758,8 +1758,7 @@
               "version_added": "79"
             },
             "firefox": {
-              "version_added": "33",
-              "version_removed": "69"
```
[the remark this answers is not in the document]
vinyldarkscratch: It looks like this was a mistake in #4823.  Firefox removed `scaleNonUniformSelf`, but not `scaleNonUniform` (see https://bugzilla.mozilla.org/show_bug.cgi?id=1560119).

I apparently reviewed and merged the PR a bit too hastily, heh...

api/DOMPointReadOnly.json
```diff
@@ -62,7 +62,7 @@
               "version_added": "79"
             },
             "firefox": {
-              "version_added": "31"
```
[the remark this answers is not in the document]
vinyldarkscratch: Looks like it was from a wiki import and has not been touched since.

api/DOMRectReadOnly.json
vinyldarkscratch: It's probably because the APIs themselves were implemented in Firefox 32 or lower (perhaps 31 if I had to guess), which we don't have data for in the collector.  I can fix that, though, run a few collectors on older Firefox versions.

api/DOMStringList.json
vinyldarkscratch: This seems to be another case just like `api.DOMException.code` up above -- property was around in the [first available commit for the interface](https://source.chromium.org/chromium/chromium/src/+/8d75ff802c1d2d66f5224851415b955343a18c42:third_party/WebKit/WebCore/dom/DOMCoreException.idl;bpv=0), but isn't defined in the prototype.  Hmmmm...

api/DOMTokenList.json
vinyldarkscratch: That doesn't make much sense...  Checking the prototype for `entries` on Chrome 42 shows it was implemented:
Screen Shot 2020-08-28 at 10 06 00

api/DOMException.json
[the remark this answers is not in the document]
vinyldarkscratch: Ah, yeah.  When working on the `Element` API, I found that until Chrome 43, properties were sometimes left as `undefined`, which was probably the case here.  Not sure how to initiate a `DOMException`, will have to research that, and then define tests/overrides in the collector...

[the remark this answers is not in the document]
vinyldarkscratch: Reverting these changes.  We can come back to them another time.

api/DOMException.json
```diff
@@ -212,7 +212,7 @@
               "version_added": "4"
             },
             "ie": {
-              "version_added": null
+              "version_added": "10"
```
foolip, line 215: Is this (api.DOMException.name) based on mdn-bcd-collector, or manual testing? If it's the collector, can you confirm it's correct? It seems a bit late for this old API...

api/DOMStringList.json
```diff
@@ -20,7 +20,7 @@
             "version_added": true
           },
           "ie": {
-            "version_added": true
+            "version_added": "10"
```
foolip, line 23: This matches https://developer.mozilla.org/en-US/docs/Web/API/Element/classList#Browser_compatibility, so seems right.

api/DOMException.json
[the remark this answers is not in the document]
foolip: Some more details about what happened in comments in https://github.com/mdittmer/mdn-confluence/commit/5e173b18caedd5dd4aec38675712d514c9fe4784. It should affect all attributes on prototypes, with the exception of global scopes like Window, where attributes are still on the instance and Window.prototype, which is correct per Web IDL.

api/DOMMatrixReadOnly.json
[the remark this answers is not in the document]
foolip: Thanks for confirming!

api/DOMMatrixReadOnly.json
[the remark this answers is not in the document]
foolip: Thanks for checking it out.

api/DOMRect.json
```diff
@@ -74,10 +74,10 @@
               "version_added": "79"
             },
             "firefox": {
-              "version_added": true
+              "version_added": "31"
```
foolip, line 77: This matches https://developer.mozilla.org/en-US/docs/Web/API/DOMPoint#Browser_compatibility so is almost certainly correct. (Or both are wrong, but you've already confirmed manually.)

api/DOMRectReadOnly.json
foolip: I see. I've confirmed this change with https://software.hixie.ch/utilities/js/live-dom-viewer/saved/8398 in Firefox 61+62.

api/DOMRect.json
```diff
@@ -4,12 +4,34 @@
...
-          "chrome_android": {
-            "version_added": true
-          },
+          "chrome": [
+            {
+              "version_added": "61"
+            },
+            {
+              "version_added": "45",
+              "flags": [
```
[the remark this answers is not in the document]
foolip: (No changes left to review now.)

api/DOMRectReadOnly.json
foolip: I see. I've confirmed this change with https://software.hixie.ch/utilities/js/live-dom-viewer/saved/8398 in Firefox 61+62.

api/DOMException.json
vinyldarkscratch: This was from the collector, I just checked it in manual testing and confirmed that `name` is not present in the object's prototype.  It's possible that this suffers the same issue as described in https://github.com/foolip/mdn-bcd-collector/issues/380.

api/DOMMatrixReadOnly.json
```diff
@@ -497,7 +497,7 @@
               "version_added": "79"
             },
             "firefox": {
-              "version_added": "33"
+              "version_added": "69"
```
foolip, line 500: Can you mirror firefox_android for this and the rest of the file?

api/DOMPointReadOnly.json
```diff
@@ -62,7 +62,7 @@
...
+              "version_added": "62"
```
foolip, line 65: Can you mirror firefox_android for this file too?

api/DOMQuad.json
```diff
@@ -151,7 +151,7 @@
           "description": "<code>fromRect()</code> static function",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "61"
```
foolip, line 154: Can you mirror all Chromium browsers for this file?

api/DOMTokenList.json
foolip: This is strange. I've confirmed M42 with http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8402. Then I used bisect-builds.py to narrow it down to this range:
https://chromium.googlesource.com/chromium/src/+log/a8a9e2cd46440abd315cc876a8f6740ded878a0b..e75311d05a5839a37e20dbf8998080f2db9627ab

I think the issue is probably that this is before Blink was merged into Chromium.

foolip: And the commit we had found was in that Blink roll:
https://chromium.googlesource.com/chromium/blink/+log/0a1a0a2..a109415

So 42 is correct. I'm not sure why the omaha proxy tools says the commit was in 45, but in any case, next time I'll need to be extra careful with Chromium changes before https://chromium.googlesource.com/chromium/src/+/b59b6df51a249895fbba24f92b661f744e031546

api/DOMException.json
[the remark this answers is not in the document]
foolip: This still needs reverting.

api/DOMException.json
```diff
@@ -125,7 +125,7 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "11.1"
```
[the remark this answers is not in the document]
foolip: Still needs reverting.

api/DOMTokenList.json
foolip: The Blink merge (commit b59b6df51a249895fbba24f92b661f744e031546) was in Chromium M47 so the general rule of thumb here would be to be extra skeptical when Chromium source code archeology suggesting 47 or earlier. The Blink commits at the time were in another repository) and the mapping to Chromium milestones is even trickier.

foolip:
@vinyldarkscratch I've reviewed this again now. The only remaining issues are mirroring. If you do that I can merge tomorrow.

api/DOMMatrixReadOnly.json
```diff
@@ -642,10 +642,10 @@
               "version_added": "79"
             },
             "firefox": {
-              "version_added": "33"
+              "version_added": "59"
```
foolip, line 645: Confirmed by https://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8403

foolip (approved):
Alright, I've gone over everything again. I'm confident this is a huge improvement in the data, but acknowledge it's possible some detail of this ends up being wrong. We will eventually find out with more automation :)

ddbeck:
I just wanted to thank you both for carrying this along. I looked at some of this briefly and it seems like a big effort. Way to go!

PR #3743, opened 2019-03-30
Update api.DOMMatrixReadOnly for Chromium and Safari

vinyldarkscratch:
Whilst working on #3368, I stumbled across #3644, and noticed that there was an inconsistency in the new data it provided.  So, I tracked it down, and found out that all the version numbers that were initially set were incorrect for Chromium and Safari.  This fixes said version numbers using data from the Chromium Git repo and Confluence.

Safari data was all imported from Confluence.

<details><summary>Chromium data is severely incorrect in Confluence and states that the API was added in version 61.  This was proven false by the following commits:</summary><p>

https://storage.googleapis.com/chromium-find-releases-static/588.html#5889261bab473ef399160e39c6b48d5eb21b4499
api.DOMMatrixReadOnly.a
api.DOMMatrixReadOnly.b
api.DOMMatrixReadOnly.c
api.DOMMatrixReadOnly.d
api.DOMMatrixReadOnly.e
api.DOMMatrixReadOnly.f
api.DOMMatrixReadOnly.m11
api.DOMMatrixReadOnly.m12
api.DOMMatrixReadOnly.m13
api.DOMMatrixReadOnly.m14
api.DOMMatrixReadOnly.m21
api.DOMMatrixReadOnly.m22
api.DOMMatrixReadOnly.m23
api.DOMMatrixReadOnly.m24
api.DOMMatrixReadOnly.m31
api.DOMMatrixReadOnly.m32
api.DOMMatrixReadOnly.m33
api.DOMMatrixReadOnly.m34
api.DOMMatrixReadOnly.is2D
api.DOMMatrixReadOnly.isIdentity

https://storage.googleapis.com/chromium-find-releases-static/2b3.html#2b3ee51b8aa11c3fcb57ca34abdf65929d0f4e0b
api.DOMMatrixReadOnly.translate

https://storage.googleapis.com/chromium-find-releases-static/3cf.html#3cf007564c0562b24113dc6aeda42b9a5553b597
api.DOMMatrixReadOnly.scale
api.DOMMatrixReadOnly.scale3d

Added: https://storage.googleapis.com/chromium-find-releases-static/3cf.html#3cf007564c0562b24113dc6aeda42b9a5553b597
Removed: https://storage.googleapis.com/chromium-find-releases-static/890.html#89064304d942709fa79a51b51e540317f4bb7d9b
api.DOMMatrixReadOnly.scaleNonUniform

[...]

PR #3644, opened 2019-03-15
Adds support information for DOMMatrixReadOnly.translate

gingerchris:
There's currently no data for this method, so I've tried to add what I can.

Chrome data pulled from https://www.chromestatus.com/features/6015941766807552

Other browsers I set to `true` if I could verify they are working right now by running
```new DOMMatrixReadOnly().translate```
and verifying that the console output is a function.

I struggled to find much version information specifically for the `translate` method and would definitely appreciate any help anyone can offer on how better to find this info!

PR #1617, opened 2018-03-18
Adding DOMMatrix and DOMMatrixReadOnly data to BCD

chrisdavidmills:
As seen at

* https://developer.mozilla.org/en-US/docs/Web/API/DOMMatrix
* https://developer.mozilla.org/en-US/docs/Web/API/DOMMatrixReadOnly

PR #6573, opened 2020-08-27
Fix mistakes in the Chromium data for DOMMatrix/DOMMatrixReadOnly

foolip:
First, remove the GeometryInterfaces flag, removed over 2 years ago:
https://chromium.googlesource.com/chromium/src/+/3e51c4becac41fe61dbd02a68bd875cc4a340457

Lint errors made it apparent that a lot of the versions were wrong.

The support data here comes mainly from the following:
https://github.com/mdn/browser-compat-data/pull/1617
https://github.com/mdn/browser-compat-data/pull/3644
https://github.com/mdn/browser-compat-data/pull/3743
https://github.com/mdn/browser-compat-data/pull/4902
https://github.com/mdn/browser-compat-data/pull/4936

The root mistake was looking at commits which added/updated these APIs
while it was still behind a flag, and some BCD entries using those
releases without documenting it as behind a flag.

In short, all of this shipped together in M61:
https://storage.googleapis.com/chromium-find-releases-static/065.html#06592e080fd1cf4a188265ed4f9bcf826952671a

The only exceptions are scaleNonUniform which shipped in M73:
https://storage.googleapis.com/chromium-find-releases-static/57f.html#57f350e070536c7870cad53b3edf74f58cc51d50

And scaleNonUniformSelf isn't in the spec and was never in Chromium:
https://github.com/w3c/fxtf-drafts/issues/214
https://github.com/whatwg/compat/issues/19

The data for opera, opera_android, samsung_internet and webview_android
was mirrored.

### DIFF
--- a/api/DOMMatrix.json
+++ b/api/DOMMatrix.json
@@ -4,48 +4,15 @@
       "__compat": {
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/DOMMatrix",
         "support": {
-          "chrome": [
-            {
-              "version_added": "61"
-            },
-            {
-              "version_added": "45",
-              "flags": [
-                {
-                  "type": "runtime_flag",
-                  "name": "--enable-blink-features=GeometryInterfaces"
-                }
-              ]
-            }
-          ],
-          "chrome_android": [
-            {
-              "version_added": "61"
-            },
-            {
-              "version_added": "45",
-              "flags": [
-                {
-                  "type": "runtime_flag",
-                  "name": "--enable-blink-features=GeometryInterfaces"
-                }
-              ]
-            }
-          ],
-          "edge": [
-            {
-              "version_added": "79"
-            },
-            {
-              "version_added": "79",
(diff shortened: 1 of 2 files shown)


foolip:
Yep, I'm fixing that now, it revealed other issues too.

foolip:
Alright, updated now. It became a pretty big change. I will squash into a single commit and update the description to match.

vinyldarkscratch:
Oh no need to worry about squashing the commits on your end!  This repository is configured so that only squash merges are allowed, so it'll all be one commit anyways.  😉

foolip:
> This repository is configured so that only squash merges are allowed, so it'll all be one commit anyways.

Yes, but that only concatenates commit messages in a pretty ugly format. Here I wanted to combine the two commit messages to tell a cleaner story about what has happened here.

foolip:
I missed one thing: https://github.com/mdn/browser-compat-data/pull/6575

PR #6574, opened 2020-08-27
Fix safari Request.mode compat data.

wanderview:
This file had the correct data under navigate_mode in one place, but was missing from other instances mode and navigate_mode.  I copied the version numbers from line 272.

I verified Request.mode is set correctly in a FetchEvent handler in safari by using fetch-event-echo.glitch.me and then opening the console for the service worker.  I tested as far back as 11.1 since it appears safari 10.1 devtools cannot show the web console for the service worker.

Given that Request.mode was in the spec from the beginning I am confident safari has supported this since they first implemented service workers.

### DIFF
--- a/api/Request.json
+++ b/api/Request.json
@@ -1211,10 +1211,10 @@
               "version_added": false
             },
             "safari": {
-              "version_added": false
+              "version_added": "10.1"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "10.3"
             },
             "samsunginternet_android": {
               "version_added": "5.0"
@@ -1258,10 +1258,10 @@
                 "version_added": false
               },
               "safari": {
-                "version_added": false
+                "version_added": "10.1"
               },
               "safari_ios": {
-                "version_added": false
+                "version_added": "10.3"
               },
               "samsunginternet_android": {
                 "version_added": "5.0"


foolip:
Welcome to BCD, @wanderview, great to see you here! I'll take a stab at reviewing this.

foolip:
> Given that Request.mode was in the spec from the beginning I am confident safari has supported this since they first implemented service workers.

Further evidence for this is the commit that introduced `Request`, indeed it includes `mode` and the `RequestMode` enum has a "navigate" value: https://trac.webkit.org/changeset/195954/webkit

The WebKit version on trunk the time was 602.1.19. (Although that scarcely matter in this case, what matters is when this was enabled.)

api/Request.json
```diff
@@ -1211,10 +1211,10 @@
               "version_added": false
             },
             "safari": {
-              "version_added": false
+              "version_added": "10.1"
```
foolip, line 1214: 10.1 and 10.3 matches when `fetch()` is said to have shipped, which makes me quite confident this is correct: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API#Browser_compatibility

api/Request.json
```diff
@@ -1258,10 +1258,10 @@
                 "version_added": false
               },
               "safari": {
-                "version_added": false
+                "version_added": "10.1"
```
foolip, line 1261: This is the "navigate_mode" entry. I've filed https://github.com/mdn/browser-compat-data/issues/6580 about why it's not shown on MDN.

foolip (approved):

PR #6575, opened 2020-08-27
Mark api.DOMMatrixReadOnly.scaleNonUniform as shipped in Edge 79

foolip:
This was overlooked in https://github.com/mdn/browser-compat-data/pull/6573
because the mirror script doesn't really work for Edge.

### DIFF
--- a/api/DOMMatrixReadOnly.json
+++ b/api/DOMMatrixReadOnly.json
@@ -1755,7 +1755,7 @@
               "version_added": "73"
             },
             "edge": {
-              "version_added": false
+              "version_added": "79"
             },
             "firefox": {
               "version_added": "33",


foolip:
> the mirror script doesn't really work for Edge

What I mean here is that `npm run mirror -- -m always edge api/DOMMatrixReadOnly.json` replaces everything with null. I guess it's because the thing was shipped in 61, before the first Edge.

foolip:
Ah, thanks, I'll try `--source=chrome` next time I mirror something.

Perhaps we could switch Edge to use Chrome as its source now though?

vinyldarkscratch:
Actually, since Edge is already kind of special cased, perhaps it makes sense to refactor the mirroring script a little bit and structure Edge to account for both IE and Edge in a single run.  I will get right on that!

PR #6576, opened 2020-08-27
Mention lack of support for SVG elements in WebKit for Element.classList

vinyldarkscratch:
This PR mentions that in WebKit, `api.Element.classList` also had no support for SVG elements at first.  Data is sourced from https://github.com/mdn/browser-compat-data/pull/6564#issuecomment-681891026 and https://github.com/mdn/browser-compat-data/pull/6564#discussion_r478278571.

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -1474,12 +1474,28 @@
         "__compat": {
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Element/classList",
           "support": {
-            "chrome": {
-              "version_added": "8"
-            },
-            "chrome_android": {
-              "version_added": "18"
-            },
+            "chrome": [
+              {
+                "version_added": "22"
+              },
+              {
+                "version_added": "8",
+                "version_removed": "22",
+                "notes": "Not supported for SVG elements.",
+                "partial_implementation": true
+              }
+            ],
+            "chrome_android": [
+              {
+                "version_added": "22"
+              },
+              {
+                "version_added": "18",
+                "version_removed": "22",
+                "notes": "Not supported for SVG elements.",
+                "partial_implementation": true
+              }
+            ],
             "edge": [
               {
                 "version_added": "16"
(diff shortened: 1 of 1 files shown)


api/Element.json
```diff
@@ -1508,18 +1524,50 @@
...
+              {
+                "version_added": "1.0",
+                "version_removed": "1.5",
+                "notes": "Not supported for SVG elements.",
+                "partial_implementation": true
+              }
+            ],
+            "webview_android": [
+              {
+                "version_added": "≤37"
```
foolip, line 1562:
```suggestion
                "version_added": "4.4"
```

The change we're looking for was in WebKit 537.4. The last WebKit-based WebView was 4.2 using WebKit 534.30, then 4.4 was based on Chromium M30 which had the change.

api/Element.json
```diff
@@ -1508,18 +1524,50 @@
...
+              },
+              {
+                "version_added": "3",
+                "version_removed": "≤37",
```
foolip, line 1566:
```suggestion
                "version_removed": "4.4",
```

foolip (approved):
These changes are technically correct, and we've now put a lot of effort into researching them.

Nevertheless, I'd like to ask what @ddbeck and @jpmedley think before merging this. https://github.com/mdn/browser-compat-data/blob/master/docs/data-guidelines.md#removal-of-irrelevant-features is relevant, but because the feature as a whole still exists, there's not a good way to remove this data, without also making it incorrect about when Element.classList was added.

vinyldarkscratch:
Well, if we were to remove this little bit of data, what we could do is remove the partial implementation, or just keep it as a note?

foolip:
I agree this doesn't fall under the irrelevant criteria. But it's also not really useful to a reader of MDN to click the little arrow and expand, only to learn that there was a period ~2010-2013 where this didn't work on SVG elements.

Still, I think we should merge this. Would still like to get a second opinion from @ddbeck or @jpmedley though.

ddbeck:
OK, I don't have my heart set on this—and I'll explain my thinking in more detail below, if you'd like something to push back on—but I think my take is this: the structure with multiple support statements is fine. But if you were to move this to a note, you'd need to do it for all the browsers (including IE).

My more complete thinking on the subject:

I think the structure for the data is basically correct. The only sensible alternative would be one entry with `"version_added"` from the earliest supported version, no partial flag, and a note in the style of, "From version 8 to 22, SVG elements (`SVGElement`) do not support this property."

We don't have guidelines for when something should get its own entry and when it's merely a note. When I've reviewed things like this, I've tried to ask myself a few questions:

* As a web developer, how significant is this departure from the contemporary specified behavior? Would this Just Work™ anyway?
* Is this an edge case or pervasive difference?
* Did the spec change or did the behavior change?
* For single-engine changes, what are we saying for other engines? Is this stylistic consistent?

Sometimes this is really obvious when you need multiple entries (e.g., an API was originally specified to accept a callback is now specified to return a promise). Sometimes it's obvious when it's a note (a buggy method returns a slightly wrong but coherent type). Sometimes, like this case, it's a bit fuzzy.

But we've got two things going for us here: one is that the other browsers (e.g., IE) present the data the same way. Changing one means changing all and what we've got here doesn't scream "overhaul all the data" to me.

Another thing is that it's basically historic. Apart from IE, this has not been a live concern in 7 or 8 years, so I'm not too stressed by the idea of the most recent entry suggesting that the feature hasn't been supported long (or showing the expansion on MDN). It's not great how BCD consumers treat arrays of support statements (including MDN itself), but I'd rather be rigorous about the data than workaround some of the rougher edges of consumers' presentation, especially when no one is really going to notice but us. 😄

foolip:
Thanks @ddbeck, those "guidelines" resonate with me. I agree the data is correct here, and it's unfortunate that it can seem to MDN readers like there's something they should care about here, but with the structure of the data is it is, any data consumer could filter this out because it's too old to care about.

I'll go ahead and merge this now.

foolip (approved):

PR #2858, opened 2018-09-20
updating Referrer Policy data

chrisdavidmills:
To find the data, I firstly looked at making it more consistent between pages, but I also looked up values/information at:

* https://www.chromestatus.com/features/5639972996513792
* https://caniuse.com/#search=referrer
* https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/7119603/

PR #6581, opened 2020-08-27
Update versions for the initial WebKit implementation of Request

foolip:
This is fixing issues similar to these recent PRs:
https://github.com/mdn/browser-compat-data/pull/6544
https://github.com/mdn/browser-compat-data/pull/6574

Fix the versions for those parts that were in WebKit's initial
implementation:
https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Modules/fetch/FetchRequest.idl?rev=195954

That the updated members were present in Safari 10.1 for macOS and
Safari for iOS 10.3 was confirmed with an ad-hoc test in BrowserStack:
https://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8385

Note that `keepalive` wasn't in the initial implementation and added
some time later. It's not updated in this change for ease of review.

### DIFF
--- a/api/Request.json
+++ b/api/Request.json
@@ -496,10 +496,10 @@
               "version_added": "47"
             },
             "safari": {
-              "version_added": "11"
+              "version_added": "10.1"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "10.3"
             },
             "samsunginternet_android": {
               "version_added": "9.0"
@@ -636,10 +636,10 @@
               }
             ],
             "safari": {
-              "version_added": false
+              "version_added": "10.1"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "10.3"
             },
             "samsunginternet_android": {
               "version_added": false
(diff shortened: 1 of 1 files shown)


api/Request.json
```diff
@@ -496,10 +496,10 @@
               "version_added": "47"
             },
             "safari": {
-              "version_added": "11"
```
foolip, line 499: This "11" came from https://github.com/mdn/browser-compat-data/pull/937 but I see no discussion of it so am guessing it was a mistake.

api/Request.json
```diff
@@ -1436,10 +1436,10 @@
               "version_added": "41"
             },
             "safari": {
-              "version_added": "11.1"
```
foolip, line 1439: This "11.1" came from https://github.com/mdn/browser-compat-data/pull/2858/files#diff-51792689c76b160a652e7eb8956b5e4f where a lot of other `referrerPolicy` attributes updated. However, as far as I can tell from `git log -S referrerPolicy -- '*.idl'` in WebKit, those were implemented separately, they're not really connected.

ddbeck (approved):
Thanks for cleaning this up, @foolip! 🎉

PR #6582, opened 2020-08-27
Mirroring script: Edge mirrors from both IE AND Chrome at the same time

queengooborg:
This PR refactors the mirroring script, adding new functions (and a few bug fixes), to mirror data for Edge a little bit differently in light of its engine change.  Now, rather than mirroring from just IE or just Chrome, it will mirror the data from _both_ browsers at once.

This includes the addition of a `combineStatements()` function, which takes an array of support statements and combines them down to their simplest form possible.  This will be helpful for future multi-source mirroring to come.  Accordingly, this also provides a structure for multi-source mirroring.

Inspired by #6575.


docs/contributing.md
```diff
@@ -74,18 +74,18 @@ If the feature you're interested in is a JavaScript API, you can cross-reference
...
-| ---------------- | ----------------- |
-| Chrome Android   | Chrome            |
-| Edge             | Internet Explorer |
-| Firefox Android  | Firefox           |
-| Opera            | Chrome            |
-| Opera Android    | Chrome Android    |
-| Safari iOS       | Safari            |
-| Samsung Internet | Chrome Android    |
-| WebView          | Chrome Android    |
+The <browser> argument is the destination browser that values will be copied to. The script automatically determines what browser to copy from based upon the destination (see table below), but manual specification is possible through the `--source=""` argument. (Mirroring data for Edge ignores this argument.)
```
foolip, line 77: Where is the "Mirroring data for Edge ignores this argument" bit in the implementation? I can't spot it, but in any case I don't think that's the behavior we should have. Let's have `--source` override the special mirroring logic for Edge if it's used. (If it's a no-op, it would be better to have the command fail with an error message.)

scripts/mirror.js
```diff
@@ -219,71 +219,61 @@ const bumpChromeAndroid = (originalData, sourceData, source) => {
...
-    if (sourceData.notes) {
-      newData.notes = updateNotes(
-        sourceData.notes,
-        /Internet Explorer/g,
-        'Edge',
-      );
-    }
-  } else if (source == 'chrome') {
-    newData = originalData == undefined ? sourceData : originalData;
+  if (ieData.version_removed && ieData.version_removed !== null) {
```
foolip, line 231: There's a bug here. Running `npm run mirror -- --m always edge` throws "TypeError: Cannot read property 'version_removed' of undefined" on this line. Seems like it's failing in webextensions/api/alarms.json where there is no IE data.

foolip:
I tried running `npm run mirror -- --m always edge` (with workaround for a bug in `bumpEdge` added) and then `git diff | grep '^+ ' | sed 's/^+ *//;s/,$//' | sort | uniq -c | sort -nr`  to see what kinds of changes this is mostly making.

It looks pretty reasonable, but it makes me suspect there are fixes that would improve this a lot. Most of the changes are adding "≤79", and very few "79" are added. There should be *lots* of APIs the data says was not in IE but was in Chrome ≤79, and where the value should be set to "79". The first cases in the diff of this is AbortController.json. Can you experiment with checking for false in the IE data, if that's not already being done?

Then there are lots of nulls being added. Sometimes that makes sense, if it was added in Edge 12-18 then obviously we can't infer the data from IE+Chrome. There are so many nulls that I suspect that there is still some bug here, but I've failed to find an incorrect change by skimming through the diffs. I can take another look after the PR has been updated.

scripts/mirror.js
```diff
@@ -219,71 +219,61 @@ const bumpChromeAndroid = (originalData, sourceData, source) => {
...
-const bumpEdge = (originalData, sourceData, source) => {
-  let newData = {};
-
-  if (source == 'ie') {
-    if (sourceData.version_removed && sourceData.version_removed !== null) {
-      newData.version_added = false;
-    } else if (sourceData.version_added !== null) {
-      newData.version_added = sourceData.version_added ? '12' : null;
-    }
+const bumpEdge = comp => {
```
foolip, line 225: Can you call this `data` instead? The compat statement objects are mostly called that in this file, and `comp` is more easily misread is being a comparator, like for a sort. (Not a misunderstanding that would stick for long, of course.)

scripts/mirror.js
```diff
@@ -219,71 +219,61 @@ const bumpChromeAndroid = (originalData, sourceData, source) => {
...
+  let newData = comp['edge'];
```
foolip, line 226: Assigning the same object to `newData` and `originalData` can't be right. Should `newData` be an empty object, or a copy of `comp['edge']`?

scripts/mirror.js
foolip, line 231:
```suggestion
  if (ieData.version_removed !== null) {
```

This is an equivalent check. But it still looks a bit funny. Should it be `ieData.version_added && !ie_data.removed_removed`?

scripts/mirror.js
```diff
@@ -219,71 +219,61 @@ const bumpChromeAndroid = (originalData, sourceData, source) => {
...
+    newData.version_added = ieData.version_added ? '12' : null;
+  }
 
-    let chromeFalse =
-      sourceData.version_added === false ||
-      sourceData.version_removed !== undefined;
-    let chromeNull = sourceData.version_added === null;
+  let chromeFalse =
+    chromeData.version_added === false ||
+    chromeData.version_removed !== undefined;
```
foolip, line 239: This bit makes me wonder what happens for a feature behind a flag. This script doesn't seem to handle flags at all. The data for Edge seems to mostly have flags copied, but it seems to me there's still a but waiting here: if the feature was added behind a flag in Chrome, `chromeFalse` will still end up false here, and we'll mark it as supported in Edge. I've verified this by modifying Bluetooth.json locally.

I think we'll need a step where support statements are pre-processed to apply flags, alternative_name, and perhaps more.

It's fine to do that in a follow-up though, it's a generic problem, not specific to mirroring Edge.

scripts/mirror.js
```diff
@@ -219,71 +219,61 @@ const bumpChromeAndroid = (originalData, sourceData, source) => {
...
-            if (originalData.version_added == false) {
-              newData.version_added = '79';
-            } else if (originalData.version_added == null) {
-              newData.version_added = '≤79';
-            }
-          } else {
-            newData.version_added == sourceData.version_added;
+  if (originalData === undefined) {
+    newData.version_added = chromeFalse ? false : chromeNull ? null : '≤79';
+  } else {
```
foolip, line 244: You can turn this into an `else if` and save some indentation in the following.

scripts/mirror.js
```diff
@@ -219,71 +219,61 @@ const bumpChromeAndroid = (originalData, sourceData, source) => {
...
-            Number(sourceData.version_added) <= 79
-          ) {
-            if (originalData.version_added == false) {
-              newData.version_added = '79';
-            } else if (originalData.version_added == null) {
-              newData.version_added = '≤79';
-            }
-          } else {
-            newData.version_added == sourceData.version_added;
+  if (originalData === undefined) {
```
foolip, line 242: In my testing `originalData === undefined` never happens. It was assigned to an object above...

scripts/mirror.js
```diff
@@ -219,71 +219,61 @@ const bumpChromeAndroid = (originalData, sourceData, source) => {
...
+    newData.version_added = chromeFalse ? false : chromeNull ? null : '≤79';
+  } else {
+    if (!chromeFalse && !chromeNull) {
+      if (originalData.version_added == true) {
```
foolip, line 246: Looking at the original data doesn't quite make sense together with `--m always`. Should this code not derive a statement from IE + Chrome, then use the command line argument to decide whether or not to update?

scripts/mirror.js
```diff
@@ -219,71 +219,61 @@ const bumpChromeAndroid = (originalData, sourceData, source) => {
...
+        newData.version_added = '≤18';
+      } else {
+        if (
+          chromeData.version_added == true ||
```
foolip, line 250: Increasingly, I think version_added true for Chrome means we don't know enough. What we *could* infer if Chrome is true and IE is false is really "≥79", but that's not an option. In the absence of that, I think we just have to set Edge to true as well?

scripts/mirror.js
```diff
@@ -527,9 +514,6 @@ const bumpVersion = (data, destination, source, originalData) => {
       case 'chrome_android':
         bumpFunction = bumpChromeAndroid;
         break;
-      case 'edge':
```
foolip, line 530: This dispatch was kinda nice though. I think we can keep it by doing `bumpFunction = () => bumpEdge(extra, stuff, here)` and then avoid special-casing Edge below.

This is a bit it would be good to have @ddbeck check out, though.

foolip (commented):
Sweet stuff! I have a lot of comments, but that's not because it ain't sweet :)

docs/contributing.md
vinyldarkscratch: Actually, thinking about it, maybe we should remove this argument entirely?  In all honesty, the sole purpose of this argument was to allow for mirroring in Edge from either IE or Chrome -- it's an argument that's not exactly well-supported and allows for mirroring across different engines.  We wouldn't want to encourage someone trying to mirror Internet Explorer's data onto Safari, for example.

scripts/mirror.js
vinyldarkscratch: I think we should do this in a follow-up, considering that it's a problem that applies to all browsers.

scripts/mirror.js
vinyldarkscratch: I'm happy to do this, but I'm thinking it would be better to follow this up in another PR, since it involves a bit more of a rewrite than I feel comfortable with in this PR.

foolip: Hmm, I see, it's because the argument for `bumpEdge` isn't available here, right?

scripts/mirror.js
foolip: Yep! Can you file an issue for that?

docs/contributing.md
foolip: I've never used it, so I wouldn't mind. It does seem a little bit plausible that someone might want to mirror Opera to Opera Android instead of using Chrome Android, in case a feature was added (or removed) from both. I'll defer to @ddbeck on whether this should stay or not.

scripts/mirror.js
vinyldarkscratch: I removed my earlier comment because I realized there actually __is__ a way to work it back in!  😅

scripts/mirror.js
```diff
@@ -219,55 +219,54 @@ const bumpChromeAndroid = (originalData, sourceData, source) => {
 };
 
 /**
- * @param {SupportStatement} comp
+ * @param {SupportStatement} compData
```
foolip, line 222:
```suggestion
 * @param {SupportStatement} originalData
```

scripts/mirror.js
vinyldarkscratch: Filed as #6625!

docs/contributing.md
foolip: Resolved by commit d6a5b11af19cb22f0726f49c34bcbd550a60deec.

foolip:
@vinyldarkscratch is this ready for final review now? It would be good to first verify that the mirror script still produces exactly the same changes for all other browsers, with/without `--source` and with/without `--modify=always`. It would also be good to review a diff of what the mirroring script would do for Edge before and after these changes, to see that there aren't any large regressions.

foolip:
@vinyldarkscratch can you ping me when you'd like review of this again?

vinyldarkscratch:
This is all good to go now, yep!

alshamiri5 (approved):

foolip:
Oops, I didn't notice I was pinged for review two weeks ago :)

foolip:
I tried running `npm run mirror -- --modify=always edge` but some of the suggested changes don't seem right. For example, in api/AbortController.json, version_added is being set to false, even though it was in Chrome 66. The right answer if ignoring the existing data should be "79". (It not using `--modify=always` it isn't modified.)

@vinyldarkscratch can you tweak the script until the output when using `--modify=always` is reasonable when spot checking?

vinyldarkscratch:
This is all ready to go; everything's been double-checked for functionality.  I would love to get this PR merged sooner than later, as it paves the way for two new PRs I have planned (one that fixes the WebView mirroring, another that has Opera Android account for data from both Opera Presto and Chrome Android).

scripts/mirror.js
```diff
@@ -220,70 +218,59 @@ const bumpChromeAndroid = (originalData, sourceData, source) => {
...
- * @param {SupportStatement} sourceData
- * @param {string} source
+ * @param {SupportStatement} chromeData
+ * @param {SupportStatement} ieData
  * @returns {SupportStatement}
  */
-const bumpEdge = (originalData, sourceData, source) => {
-  let newData = {};
+const bumpEdge = (originalData, chromeData, ieData) => {
+  let newData = copyStatement(originalData);
```
foolip, line 226: There's a bug here, or in `copyStatement`, which I don't think is really new, but it didn't manifest before. If you run `npm run mirror -- edge --modify=bool`, this change will be produced:

```diff
--- a/css/selectors/file-selector-button.json
+++ b/css/selectors/file-selector-button.json
@@ -26,11 +26,24 @@
             ],
             "edge": [
               {
+                "0": {
+                  "version_added": false
+                },
+                "1": {
+                  "version_added": "79",
+                  "alternative_name": "::-webkit-file-upload-button"
+                },
                 "version_added": false
               },
               {
-                "version_added": "79",
-                "alternative_name": "::-webkit-file-upload-button"
+                "0": {
+                  "version_added": false
+                },
+                "1": {
+                  "version_added": "79",
+                  "alternative_name": "::-webkit-file-upload-button"
+                },
+                "version_added": "79"
               }
             ],
             "firefox": {
```

It looks like Chrome's array of support statements is being treated as a single support statement, with broken results. It seems to me this might also be happening for other uses of `copyStatement`, but for this PR it's important to at least not regress what `npm run mirror` does.

foolip (commented):
Sorry I left this sitting for months! I've taken another look and tried it a bit myself, unfortunately found a bug in need of fixing.

scripts/mirror.js
vinyldarkscratch: I don't know why it took me this long to finally get a fix out, especially one so easy.  Probably just needed a fresh pair of eyes...  Anyways, this is all fixed and ready to go!

scripts/mirror.js
```diff
@@ -223,6 +223,10 @@ const bumpChromeAndroid = (originalData, sourceData, source) => {
  * @returns {SupportStatement}
  */
 const bumpEdge = (originalData, chromeData, ieData) => {
+  if (Array.isArray(originalData)) {
+    return originalData.map(d => bumpEdge(d));
```
foolip, line 227: What about the `chromeData` and `ieData` arguments in this case? If they're arrays, should the data be used? How about if they're not arrays? Just ignoring the data seems odd, although I don't remember now how any of this works :)

scripts/mirror.js
```diff
@@ -220,70 +218,63 @@ const bumpChromeAndroid = (originalData, sourceData, source) => {
...
-            newData.version_added == sourceData.version_added;
-          }
-        }
-      } else if (chromeFalse) {
-        if (originalData.version_added && !originalData.version_removed) {
-          newData.version_removed = '79';
-        }
+  let chromeFalse =
+    chromeData.version_added === false ||
+    chromeData.version_removed !== undefined;
```
foolip, line 242: Could this be `chromeData.version_removed || chromeData.version_added === false`, more like the `ieData` case above?

scripts/mirror.js
```diff
@@ -220,70 +218,62 @@ const bumpChromeAndroid = (originalData, sourceData, source) => {
...
-    }
+  let newData = copyStatement(originalData);
 
-    if (sourceData.notes) {
-      newData.notes = updateNotes(
-        sourceData.notes,
-        /Internet Explorer/g,
-        'Edge',
-      );
+  if (ieData) {
```
foolip, line 232: `ieData` can be an array here, verified by adding an assert. Can you explicitly handle that case, even if it's by bailing and doing nothing?

scripts/mirror.js
```diff
@@ -220,70 +218,62 @@ const bumpChromeAndroid = (originalData, sourceData, source) => {
...
-          } else {
-            newData.version_added == sourceData.version_added;
-          }
-        }
-      } else if (chromeFalse) {
-        if (originalData.version_added && !originalData.version_removed) {
-          newData.version_removed = '79';
-        }
+  let chromeFalse =
+    chromeData.version_removed || chromeData.version_added === false;
```
foolip, line 241: It should be possible that `chromeData` is an array here, but adding an assert I don't see it happening. Can it happen, and if so can you handle it?

scripts/mirror.js
```diff
@@ -220,70 +218,62 @@ const bumpChromeAndroid = (originalData, sourceData, source) => {
...
-      newData.notes = updateNotes(
-        sourceData.notes,
-        /Internet Explorer/g,
-        'Edge',
-      );
+  if (ieData) {
+    if (ieData.version_removed || ieData.version_added === false) {
+      newData.version_added = false;
+    } else if (ieData.version_added) {
+      newData.version_added = ieData.version_added ? '12' : null;
```
foolip, line 236:
```suggestion
      newData.version_added = '12';
```

`ieData.version_added` is always truthy here.

scripts/mirror.js
```diff
@@ -220,70 +218,62 @@ const bumpChromeAndroid = (originalData, sourceData, source) => {
...
-        if (originalData.version_added && !originalData.version_removed) {
-          newData.version_removed = '79';
-        }
+  let chromeFalse =
+    chromeData.version_removed || chromeData.version_added === false;
+  let chromeNull = chromeData.version_added === null;
+
+  if (!chromeFalse && !chromeNull) {
+    if (originalData.version_added === true) {
+      newData.version_added = '≤18';
```
foolip, line 246: This is an existing condition, but it doesn't make sense to me. If the original data was true, and Chrome is neither false nor null, then I don't think anything at all follows, that depends on what the Chrome data is.

Can you find a case where this branch is taken and see if it actually makes sense there? I'm skeptical :)

scripts/mirror.js
```diff
@@ -220,70 +218,62 @@ const bumpChromeAndroid = (originalData, sourceData, source) => {
...
+    } else if (chromeData.version_added === true) {
+      newData.version_added = true;
+    } else if (Number(chromeData.version_added) <= 79) {
+      if (
+        originalData.version_added === false ||
+        newData.version_added === false
```
foolip, line 252: OK, so `newData.version_added === false` can happen here if it was set to false because it wasn't in IE. This makes sense, but took some pondering to understand :)

A comment for each case would be nice, but I won't insist on it.

scripts/mirror.js
```diff
@@ -220,70 +218,62 @@ const bumpChromeAndroid = (originalData, sourceData, source) => {
...
       }
+    } else {
+      newData.version_added = chromeData.version_added;
     }
+  } else if (chromeFalse) {
+    if (originalData.version_added && !originalData.version_removed) {
+      newData.version_removed = '79';
+    }
+  } else if (chromeNull) {
+    newData.version_added = null;
```
foolip, line 266: There soon won't be any null Chrome values, but can't this case end up overwriting non-null data sourced from IE? Is it needed?

foolip (changes requested):
This is pretty close to done I think, let's get it across the finish line!

scripts/mirror.js
vinyldarkscratch: I've just pushed a commit that includes something that solves this issue and beyond!

scripts/mirror.js
vinyldarkscratch: This is for if Chrome is `true`, though the way the code is structured doesn't help with legibility.  Fixing that up!

scripts/mirror.js
vinyldarkscratch: I've made some changes which will prevent this from occurring!

vinyldarkscratch:
Got an improved structure for multi-source mirroring put into place!  I think this is all ready to go now!

scripts/mirror.js
```diff
@@ -219,74 +280,93 @@ const bumpChromeAndroid = (originalData, sourceData, source) => {
...
+      newData.push(bumpEdgeFromIE(ieData));
     }
   }
 
-  return newData;
+  if (chromeData) {
+    if (Array.isArray(chromeData)) {
+      newData += chromeData.map(d => bumpEdgeFromChrome(d, originalData));
+    } else {
+      newData.push(bumpEdgeFromChrome(chromeData, originalData));
```
foolip, line 365: If you run `npm run -- mirror edge -m always` an exception will be thrown here: "newData.push is not a function"

scripts/mirror.js
```diff
@@ -219,74 +280,93 @@ const bumpChromeAndroid = (originalData, sourceData, source) => {
...
+    return originalData.map(d => bumpEdge(d, chromeData, ieData));
+  }
 
-    if (newNotes) {
-      newData.notes = newNotes;
+  let newData = [];
+
+  if (ieData) {
+    if (Array.isArray(ieData)) {
+      newData += ieData.map(d => bumpEdgeFromIE(d));
```
foolip, line 355: This and the the similar branch won't work, arrays can't be extended like this in JS.

foolip (commented):
Found some bugs. It would be good to try running the script with all permutations of arguments and seeing if the result always makes sense, passes the lint, and whatever else can be easily checked.

scripts/mirror.js
vinyldarkscratch: Ah, right, sometimes I forget Python and JavaScript are two separate languages.  :P

foolip: Me too :)

foolip:
@vinyldarkscratch have you tried running without various arguments to see what kinds of changes are made? Can you summarize for ease of review?

foolip:
@teoli2003 ran into the fact that Edge is mirrored only from IE in https://github.com/mdn/browser-compat-data/pull/11063#discussion_r653635715 so I have a renewed desire to get this landed :)

vinyldarkscratch:
Running the mirroring script now without arguments (other than `edge` of course), I've noticed a few cases that may still need to be resolved.  Looking at the [generated diff](https://github.com/mdn/browser-compat-data/files/6735831/mirror.diff.txt), these include:

- IE is `false`, Chrome is `null` -> Edge becomes `false` (should be `null`)
- IE is truthy, Chrome has a `version_added` and `version_removed` -> Edge becomes multi-statement and Chrome's versions don't get bumped

vinyldarkscratch:
Alright, this is now ready to go, I'm happy to say!  All the major bugs have been worked out, and I'm confident in the results yielded by this script now!

scripts/mirror.js
```diff
@@ -187,13 +183,91 @@ const copyStatement = data => {
...
+  const ignored_keys = ['version_added', 'notes'];
+
+  let flattenedData = data.flat(2);
+  let sections = {};
+  let newData = [];
+
+  for (const d of flattenedData) {
+    let key = Object.keys(d)
+      .filter(k => !ignored_keys.includes(k))
+      .join('');
```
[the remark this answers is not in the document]
queengooborg: This `key` parameter here is to make sure that multiple statements with the same set of keys are properly merged, so this is fine!  Joining them together just makes it a little easier to work with.

queengooborg:
I mentioned this during our last meeting -- although there are still some unresolved comments, I feel this is ready enough to merge now, and then any more changes can be followed up later.  This PR's blocking some other mirroring updates, and I think it's ready enough to use!

foolip:
@queengooborg I just tried `npm run mirror -- -m bool edge` and that throws "TypeError: Invalid argument expected string". Can you check that out?

queengooborg:
Slight logic problem, fixed now though!

foolip:
`npm run mirror -- -m bool edge` now removes notes which it didn't do before. Example:
```diff
--- a/api/CSSStyleDeclaration.json
+++ b/api/CSSStyleDeclaration.json
@@ -163,8 +163,7 @@
               "notes": "See <a href='https://crbug.com/331608'>bug 331608</a>."
             },
             "edge": {
-              "version_added": false,
-              "notes": "See <a href='https://crbug.com/331608'>bug 331608</a>."
+              "version_added": false
             },
             "firefox": {
               "version_added": "1",
```

foolip:
I'm trying to figure out how the Edge team could use this script after each Edge release to keep the data up-to-date, after a discussion with @Reezaali last week.

Unfortunately, the script is going to make incorrect changes based on things that were in IE but were removed in Edge 12. The natural way to avoid that would be `--source chrome`, but that doesn't work now, and `npm run mirror -- --source chrome -m bool edge` prints this warning:

> Warning: Edge does not support --source parameter.

@queengooborg that did work previously. Is there an alternative command that would be suitable for running for every release now?

queengooborg:
I was going to suggest the `--target_version` parameter but looking through the code I realize that's not a thing yet... :D  I have a PR open, #6170, that adds that functionality in -- it's created specifically for the purpose of mirroring data for a new release for Samsung Internet, Opera, and now Edge!

queengooborg:
Even if we might be considering removing Internet Explorer, I'd still like to get this landed as it provides important infrastructure changes to this script that I definitely expect us to use for other browsers as well.

foolip:
@queengooborg I just tried `npm run mirror -- -m bool edge` again and it's making many changes that don't pass the lint. Two of them are:

```diff
--- a/api/Document.json
+++ b/api/Document.json
@@ -4946,9 +4946,7 @@
             "chrome_android": {
               "version_added": false
             },
-            "edge": {
-              "version_added": false
-            },
+            "edge": [],
             "firefox": {
               "version_added": false
             },

```

```diff
--- a/webextensions/manifest/manifest_version.json
+++ b/webextensions/manifest/manifest_version.json
@@ -33,9 +33,7 @@
                 "version_added": "4",
                 "version_removed": "21"
               },
-              "edge": {
-                "version_added": false
-              },
+              "edge": {},
               "firefox": {
                 "version_added": false
               },
```

Running `npm run mirror -- -m always edge` also makes changes that don't pass the lint, like this one:

```diff
--- a/svg/elements/discard.json
+++ b/svg/elements/discard.json
@@ -15,7 +15,7 @@
               "version_removed": "81"
             },
             "edge": {
-              "version_added": "79",
+              "version_added": "34",
               "version_removed": "81"
             },
             "firefox": {
```

Some manual fixing after running the script is expected, but the above seem like errors the script should be able to avoid.

foolip:
So what should we do with Edge mirroring?

I think that with https://github.com/mdn/browser-compat-data/issues/15083 we will not want to supporting mirroring from both IE and Chrome, but just Chrome. I would really prefer to make that same change here, so that we can reuse more of the code.

queengooborg:
> I think that with #15083 we will not want to supporting mirroring from both IE and Chrome, but just Chrome.

I opened up #16130 to change the default to Chrome, which should help with #15083!

queengooborg:
I just fixed up the logic in regards to those three odd statements, and I have confirmed they don't negatively impact other statements as well!

foolip:
I've tried `npm run mirror -- -m always edge` again, now it throws with an error.

It seems like the approach here is very brittle since there are new kinds of errors with each iteration. Can we simplify this to supporting mirroring from IE _or_ Chrome, not both at the same time? In other words, support `--source`, and have it default to Chrome.

Then it would be straightforward to review the changes that `npm run mirror -- --source ie edge` would make, and after that we could pretty much forget about that code path.

queengooborg:
I just pushed a few changes that allow for selecting a source for Edge, and fixed the error I think you might've experienced!  I still have the dual browsers selected as the default for mirroring, but a `--source` argument is now allowed once again, though this time with a validation to ensure it's only one of three options we've scripted.

queengooborg:
Made the default as just `chrome` now actually, with build-time mirroring coming up it's less likely that we'll want to mirror IE data.

OrToR-KJM:
내 Galaxy에서 보냄
-------- 원본 이메일 --------발신: queengooborg ***@***.***> 날짜: 22/5/18  15:50  (GMT+09:00) 받은 사람: mdn/browser-compat-data ***@***.***> 참조: Subscribed ***@***.***> 제목: Re: [mdn/browser-compat-data] Mirroring script: Edge mirrors from both IE AND Chrome at the same time (#6582) 
Merged #6582 into main.

PR #6586, opened 2020-08-28
Add Firefox Android (Fenix) 79.0

vinyldarkscratch:
The brand new Firefox Android browser is now out as of last month!  This PR adds it to the browser data -- a follow-up to mirror all the data will be coming in a separate PR.

Fenix-specific release notes: https://support.mozilla.org/en-US/kb/whats-new-firefox-android-79

### DIFF
--- a/browsers/firefox_android.json
+++ b/browsers/firefox_android.json
@@ -378,7 +378,7 @@
         "60": {
           "release_date": "2018-05-09",
           "release_notes": "https://developer.mozilla.org/docs/Mozilla/Firefox/Releases/60",
-          "status": "esr",
+          "status": "retired",
           "engine": "Gecko",
           "engine_version": "60"
         },
--- a/browsers/firefox_android.json
+++ b/browsers/firefox_android.json
@@ -434,9 +434,16 @@
         "68": {
           "release_date": "2019-07-09",
           "release_notes": "https://developer.mozilla.org/docs/Mozilla/Firefox/Releases/68",
-          "status": "current",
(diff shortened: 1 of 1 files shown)


browsers/firefox_android.json
```diff
@@ -378,7 +378,7 @@
         "60": {
           "release_date": "2018-05-09",
           "release_notes": "https://developer.mozilla.org/docs/Mozilla/Firefox/Releases/60",
-          "status": "esr",
+          "status": "retired",
```
foolip, line 381: This ESR change I think would be best reviewed by @ddbeck.

foolip (approved):
Looks good apart from the ESR changes which I dare not review.

browsers/firefox_android.json
vinyldarkscratch: I've separated this into #6627!

foolip:
I'm suddenly in doubt if I should have merged this per https://github.com/mdn/browser-compat-data/blob/master/GOVERNANCE.md#peers...

@ddbeck to confirm, is it OK for peers to merge changes to browser releases, or only to the support data?

vinyldarkscratch:
@foolip I've merged a number of browser data PRs and haven't heard anything negative about it, so I'd say you're all good!

ddbeck:
@foolip Yep, @vinyldarkscratch has it right. Browser data is data too. (Adding a _new_ browser would be something else, though that would require some significant changes to the linter and such.)

PR #6587, opened 2020-08-28
Mirror Firefox -> Fenix for 79

vinyldarkscratch:
Created using the mirroring script, validated for any problematic entries.


ddbeck:
> validated for any problematic entries

Did you make any changes as a consequence? Or did the mirroring check out across the board?

Also, perhaps a question worth exploring: does anything here conflict with #6590? (I'm somewhat inclined to finish that one _first_ and see if this one clobbers anything in webext, actually.)

vinyldarkscratch:
Mirroring checked out without any problems!  I just had to adjust some notes here and there, but otherwise it worked like a charm.

In regards to the WebExt data, I will be reverting those changes and diverting to #6590.  I trust that PR more than I do a simple mirroring.

ddbeck (approved):
Thanks, @vinyldarkscratch! 🎉

PR #6589, opened 2020-08-28
Update for Samsung Internet 13

AdaRoseCannon:
Uses the script to update Samsung Internet for 13 which is at Chromium m83

### DIFF
--- a/api/AudioParam.json
+++ b/api/AudioParam.json
@@ -164,7 +164,7 @@
               {
                 "version_added": true,
                 "version_removed": "6.0",
-                "alternative_name": "cancelValuesAndHoldAtTime()"
+                "alternative_name": "cancelValuesAndHoldAtTime"
               }
             ],
             "webview_android": [
--- a/api/AudioParam.json
+++ b/api/AudioParam.json
@@ -162,9 +162,9 @@
                 "version_added": "7.0"
               },
               {
-                "alternative_name": "cancelValuesAndHoldAtTime()",
                 "version_added": true,
-                "version_removed": "6.0"
+                "version_removed": "6.0",
+                "alternative_name": "cancelValuesAndHoldAtTime()"
               }
             ],
             "webview_android": [
--- a/api/BarcodeDetector.json
+++ b/api/BarcodeDetector.json
@@ -35,7 +35,7 @@
             "version_added": false
           },
           "samsunginternet_android": {
-            "version_added": true
+            "version_added": "13.0"
           },
           "webview_android": {
             "version_added": "83"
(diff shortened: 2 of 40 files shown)


api/AudioParam.json
```diff
@@ -162,9 +162,9 @@
                 "version_added": "7.0"
               },
               {
-                "alternative_name": "cancelValuesAndHoldAtTime()",
                 "version_added": true,
-                "version_removed": "6.0"
+                "version_removed": "6.0",
+                "alternative_name": "cancelValuesAndHoldAtTime()"
```
foolip, line 167: I saw the parens here and sent https://github.com/mdn/browser-compat-data/pull/6631. Sorry that this will conflict if landed first, but if you revert the changes to api/AudioParam.json it'll work it.

foolip:
@AdaRoseCannon I'm new to BCD as a reviewer but am happy to help. I wanted to try reproducing the changes and tried running `npm run mirror samsunginternet_android` but don't get quite the same results. Something that I find helpful both when making changes and as a reviewer is to have one commit with the scripted changes with no changes, and then to separate make the fixes and manual edits. When reviewing, I'd then mostly focus on the manual changes.

foolip:
I also see that the updates are a mix of things that were really changed in Samsung Internet 13.0 and mirror lots of changes that have failed to be mirrored in the repo since the last update. I don't know if separating those categories is practical, but it would be useful if it is simple. Perhaps some hack to the mirror script could produce the changes separately. (This is not a request, just an idea.)

api/AudioParam.json
AdaRoseCannon: Fixed

AdaRoseCannon:
Reviewing the changes I make to the scripted output is just about as long as the final output. Here are my steps:

1. Run `./scripts/mirror.js samsunginternet_android --modify=bool */`

This adds the support information when we didn't support it before i.e. `false` -> `13.0`

This only mirrors boolean changes, so brings in more details when it has been added i.e. `null|true` -> `6.0`

2. Go through every touched file and remove/fix:

* Incorrectly copied notes
* APIs we do not support
* APIs behind flags (we do not use flags)
* Newly added APIs we added earlier than the comparable Chrome release

Here is the diff between the raw output and after the changes: https://github.com/SamsungInternet/browser-compat-data/compare/samsunginternet-no-fixes...SamsungInternet:samsunginternet-with-fixes

foolip:
Thanks for the details @AdaRoseCannon! The extra argument to only update boolean values is what I was missing. I'll review this in a bit when I'm at my computer.

foolip (approved):
I've gone through all the changes and spot checked some things that seemed interesting. Thanks for the updates!

PR #6592, opened 2020-08-29
Results from mdn-bcd-collector for Element API

vinyldarkscratch:
This PR provides results collected from [mdn-bcd-collector](https://github.com/foolip/mdn-bcd-collector) for the `Element` API.  All versions have been verified manually in SauceLabs and local VMs.  (All invalid data was removed from this PR and either corrected on the collector or separated into a new PR.)

_Note: although the collector has more results, I stripped it down to about half to keep the changeset down._

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -677,19 +677,24 @@
             "opera_android": {
               "version_added": "24"
             },
-            "safari": {
-              "version_added": true,
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "Web Animations"
-                },
-                {
-                  "type": "preference",
-                  "name": "CSS Animations via Web Animations"
-                }
-              ]
-            },
+            "safari": [
+              {
+                "version_added": "13.1"
+              },
+              {
+                "version_added": true,
+                "flags": [
+                  {
+                    "type": "preference",
+                    "name": "Web Animations"
+                  },
+                  {
+                    "type": "preference",
+                    "name": "CSS Animations via Web Animations"
+                  }
+                ]
+              }
+            ],
             "safari_ios": {
               "version_added": "13.4"
             },
@@ -3311,6 +3316,9 @@
           "support": {
             "chrome": [
               {
+                "version_added": "84"
+              },
+              {
                 "version_added": "79",
                 "flags": [
                   {
@@ -3359,6 +3367,9 @@
             ],
             "chrome_android": [
               {
+                "version_added": "84"
+              },
+              {
                 "version_added": "79",
                 "flags": [
                   {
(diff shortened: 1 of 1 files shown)


api/Element.json
```diff
@@ -5189,7 +5209,7 @@
             ],
             "edge": [
               {
-                "version_added": "≤18"
+                "version_added": "15"
               },
               {
                 "version_added": "≤18",
```
foolip, line 5215: I've confirmed using Sauce that `webkitMatchesSelector` was present in Edge 13. It was probably in Edge 12, but I don't have my computer with the Edge 12 on it. Would be good to fix this too, but not necessary in this PR.

api/Element.json
```diff
@@ -5874,7 +5894,7 @@
               }
             ],
             "edge": {
-              "version_added": "≤79"
+              "version_added": "79"
```
foolip, line 5897: IE and Edge both had support for an ms-prefixed and/or a webkit-prefixed version of the fullscreen API, but never shipped an unprefixed version. Can you revert the fullscreen-related changes and deal with them separately? It will be pretty messy and take a lot of iteration to get just right.

I will add a comment that says just "revert" to each line, so that it's easy to see if anything was missed, not because I'm very rude :)

api/Element.json
```diff
@@ -5906,7 +5926,7 @@
               "version_added": true
             },
             "safari": {
-              "version_added": null
+              "version_added": false
```
foolip, line 5929: Let's revert this too, Safari was in fact the first browser to ship a fullscreen API, but webkit-prefixed.

api/Element.json
```diff
@@ -5897,7 +5917,7 @@
               }
             ],
             "ie": {
-              "version_added": null
+              "version_added": false
```
foolip, line 5920: revert

api/Element.json
```diff
@@ -5954,7 +5974,7 @@
               }
             ],
             "edge": {
-              "version_added": "≤79"
+              "version_added": "79"
```
foolip, line 5977: revert

api/Element.json
```diff
@@ -5977,7 +5997,7 @@
               }
             ],
             "ie": {
-              "version_added": null
+              "version_added": false
```
foolip, line 6000: revert

api/Element.json
```diff
@@ -5820,7 +5840,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": null
+              "version_added": true
```
foolip, line 5843: revert

api/Element.json
```diff
@@ -6326,7 +6346,7 @@
               "notes": "This API was previously available on the <code>Node</code> API."
             },
             "edge": {
-              "version_added": "13"
+              "version_added": "12"
```
foolip, line 6349: revert

api/Element.json
```diff
@@ -6337,7 +6357,7 @@
               "notes": "This API was previously available on the <code>Node</code> API."
             },
             "ie": {
-              "version_added": null
+              "version_added": true
```
foolip, line 6360: revert

foolip (approved):
With some of the changes split out, I'd be happy to merge this.

api/Element.json
vinyldarkscratch: It looks like it's present in Edge 12 as well -- patch coming right up!

api/Element.json
```diff
@@ -5189,10 +5209,10 @@
             ],
             "edge": [
               {
-                "version_added": "≤18"
+                "version_added": "15"
               },
               {
-                "version_added": "≤18",
+                "version_added": "12",
```
foolip, line 5215: Sweet!

foolip (approved):

PR #6593, opened 2020-08-29
fix(css): safari order for 'max-inline-size'

kripod:
See: https://github.com/mdn/browser-compat-data/blob/master/schemas/compat-data-schema.md#the-support_statement-object

> If there is an array, the `simple_support_statement` objects should be sorted with the most relevant and general entries first. In other words, sort such arrays with entries applying to the most recent browser releases first and sort entries with prefixes or flags after those without.

- [x] Summarize your changes

### DIFF
--- a/css/properties/max-inline-size.json
+++ b/css/properties/max-inline-size.json
@@ -57,11 +57,11 @@
             },
             "safari": [
               {
-                "prefix": "-webkit-",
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Looks good. Thank you! 🎉

PR #6596, opened 2020-08-29
fix(css): prefix order of user-select in Firefox

kripod:
On Firefox for Android, `-webkit-user-select: none` takes effect, while `-moz-user-select: none` doesn't.

- [x] Summarize your changes

### DIFF
--- a/css/properties/user-select.json
+++ b/css/properties/user-select.json
@@ -38,12 +38,12 @@
                 "version_added": "69"
               },
               {
-                "prefix": "-moz-",
-                "version_added": "1"
-              },
-              {
                 "prefix": "-webkit-",
                 "version_added": "49"
+              },
+              {
+                "prefix": "-moz-",
+                "version_added": "1"
               },
               {
                 "prefix": "-webkit-",
@@ -59,12 +59,12 @@
             ],
             "firefox_android": [
               {
-                "prefix": "-moz-",
-                "version_added": "4"
-              },
-              {
                 "prefix": "-webkit-",
                 "version_added": "49"
+              },
+              {
+                "prefix": "-moz-",
+                "version_added": "4"
               },
               {
                 "prefix": "-webkit-",


ddbeck:
Hi @kripod, I assume you tested this yourself. Can you share your test case? (I got… weird results when I tried it.) We might want to mark the prefix as unsupported on Android, if that's the case.

The other change, to the sort for Firefox on the desktop, that looks fine to me. 👍

kripod:
Sure, I can share how I replicated my results. Here are the CodePens I'm testing with:

- `-moz-user-select: none`: [pen](https://codepen.io/kripod97/pen/bGpYZGj), [debug](https://codepen.io/kripod97/debug/bGpYZGj)
- `-webkit-user-select: none`: [pen](https://codepen.io/kripod97/pen/bGpYZGj), [debug](https://codepen.io/kripod97/debug/VwarRaW)

I downloaded [Android Studio](https://developer.android.com/studio) with its emulator included, and then opened the AVD Manager to run a Pixel 9 API 19 with [Firefox 49](https://www.apkmirror.com/apk/mozilla/firefox/firefox-49-0-release/firefox-49-0-android-apk-download/). On that version, both `-webkit-user-select: none` and `-moz-user-select: none` work fine. However, on a more recent version of Firefox for Android (possibly the latest), only `-webkit-user-select: none` seems to work.

kripod:
After trying with [Firefox Mobile 80](https://www.apkmirror.com/apk/mozilla/firefox/firefox-80-1-2-release/) on an emulated Pixel 3 (API 30, Android 10), it seems that both `-moz-user-select` and `webkit-user-select` work fine. The `-webkit-` prefix is still more relevant, as it's more widely supported by other browsers. So, this PR shouldn't be invalidated by the fact that both non-standard properties work.

ddbeck (approved):
Thanks for the follow-up! This does seem right. I'll merge this now. 🎉

PR #6599, opened 2020-08-30
fix(css): grid-template alternatives for IE 10+

kripod:
Please see the note about the alternative properties in [this CSS-Tricks article](https://css-tricks.com/almanac/properties/g/grid-rows-columns/) for further details.

- [x] Summarize your changes


ddbeck (changes requested):
Thanks for opening this PR, @kripod. I think this is the right thing to do in general, but there's some specifics here that need some attention.

First, we seem to have dedicated files for `-ms-grid-columns` and `-ms-grid-rows`. I think those ought to be removed as  well, since the new data here duplicates that.

Second, I'm pretty sure the `-ms-` prefixes went away in Edge 79. Would you be willing to confirm that they're no longer supported and set the version numbers (12-??) for Edge accordingly?

Thank you again for your help on this!

kripod:
Thank you for the review! I've updated the PR as requested, incorporating previously available data for Edge 12–16.

ddbeck (changes requested):
Sorry, @kripod I think I explained myself poorly. It looks like Edge continued to support the `-ms-` prefixed properties until Edge 79 (that is, `"version_added": "12" and "version_removed": "79"). Can you reinstate that data for Edge? Then we can merge this. Thank you!

kripod:
> Sorry, @kripod I think I explained myself poorly. It looks like Edge continued to support the `-ms-` prefixed properties until Edge 79 (that is, `"version_added": "12"` and `"version_removed": "79"`). Can you reinstate that data for Edge? Then we can merge this. Thank you!

Oh I see, you're right, thank you! I've just updated the PR to fix those.

ddbeck (approved):
Perfect. Thank you, @kripod! 🎉

PR #5779, opened 2020-03-04
svg/discard will be removed in chrome 81

HolgerJeromin:
https://www.chromestatus.com/feature/4870172764536832

PR #6600, opened 2020-08-30
Update SVG <discard> support data (only implementation removed)

foolip:
This was removed in Chromium 81: https://github.com/mdn/browser-compat-data/pull/5779

Using a simple `window.SVGDiscardElement` test, I determined that the
interface object was added in Chrome 34. It's possible that the markup
was supported earlier than that, but it's not important.

Since it's not possible for anyone to depend on this feature and the
data in BCD should be removed in ~2 years, just delete the subfeatures
which had no useful data, rather than researching the attribute support
in similar detail.

### DIFF
--- a/svg/elements/discard.json
+++ b/svg/elements/discard.json
@@ -6,15 +6,16 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/SVG/Element/discard",
           "support": {
             "chrome": {
-              "version_added": true,
+              "version_added": "34",
               "version_removed": "81"
             },
             "chrome_android": {
-              "version_added": true,
+              "version_added": "34",
               "version_removed": "81"
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "79",
+              "version_removed": "81"
             },
             "firefox": {
               "version_added": false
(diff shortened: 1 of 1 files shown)


foolip:
@ddbeck can you review this? It's a bit opinionated and it would be possible to do better, but I just don't think it's worth it. I wanted to do this because of the href attribute, as I was researching the difference between href and xlink:href.

svg/elements/discard.json
```diff
@@ -26,22 +27,25 @@
...
+              "version_added": false
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": false
             },
             "samsunginternet_android": {
-              "version_added": true
+              "version_added": "2.0",
+              "version_removed": false
```
foolip, line 45: I'll file an issue about updating this in the future if the change is accepted.

ddbeck, line 45: You can omit the non-removal:

```suggestion
```

ddbeck (changes requested):
Thanks for this PR! One concrete suggestion inline, but let me see if I understand the situation correctly:

1. There was a `<discard>` element, only implemented in Chrome (etc.), which is being removed in 81.
2. It (probably?) implemented some attributes at some point, but BCD never had data about them.
3. This PR marks the removal of `<discard>` and removes the subfeatures (on the basis that the data is not good and not worth digging up, not that the features never existed).

If that correctly characterizes the changes here, then I'm fine with it (with the inclusion of the line edit). I would probably accept a future PR that added back the subfeatures with reliable data, but that strikes me as unlikely to happen.

(I'm labeling this for release notes, which I try to do for all feature removals.)

svg/elements/discard.json
foolip: Will do, but shouldn't this fail the lint? :)

foolip:
@ddbeck your summary in https://github.com/mdn/browser-compat-data/pull/6600#pullrequestreview-479930316 is correct to the best of my understanding. If somebody does want to come along and fill in more of the data here, then indeed we should accept that. Most likely, though, this entry will be deleted in the future.

ddbeck (approved):
Thank you! 🎉

PR #5135, opened 2019-11-15
MediaSession (Partial) added to Firefox 71

a2sheppy:
Add to BCD the parts of MediaSession API that Firefox 71 supports:
* `Navigator.mediaSession`
* Added missing `setPositionState()` to the `MediaSession` data
* `MediaSession` (all but two members)
* `MediaMetadata`

Source: https://bugzilla.mozilla.org/show_bug.cgi?id=1580602

PR #6611, opened 2020-08-31
Fix Firefox data for MediaMetadata API

vinyldarkscratch:
This PR fixes the Firefox data for the MediaMetadata API.  In #5135, support data was added for the interface.  However, from manual testing, I found that it was actually supported behind the `dom.media.mediasession.enabled` flag.  This PR corrects that.

### DIFF
--- a/api/MediaMetadata.json
+++ b/api/MediaMetadata.json
@@ -36,7 +36,7 @@
             "version_added": false
           },
           "safari": {
-            "version_added": false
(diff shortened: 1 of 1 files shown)


sideshowbarker (approved):

foolip (approved):
Confirmed in local testing that MediaSession still isn't enabled in Firefox, and this matches api.MediaSession.

Issue #6475, opened 2020-08-05
Do Chrome & Safari support <rb>?


fasiha:
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/rb

The first snippet, with `<rb>` tags inside `<ruby>`, renders incorrectly in both Safari and Chrome (looks fine in Firefox). The second snippet, without `<rb>`, looks right in all three.

Chrome and Safari also incorrectly render all the examples on this 2015 [Mozilla blog post](https://hacks.mozilla.org/2015/03/ruby-support-in-firefox-developer-edition-38/).

More details on my [Stack Overflow question](https://stackoverflow.com/q/63257758).

Looks like this was added to BCD in 2018, #2422.

sideshowbarker:
While Chrome and Safari have implemented support for parsing the `rb` element , I think they haven’t implemented the support necessary for styling it as expected.

So I’ve raised #6613 to mark `rb` with `"partial_implementation": true` for Chrome and Safari.

Incidentally, the `rb` element isn’t actually part of the HTML standard, so implementors aren’t required to do anything with it.

So I’ve also raised #6614 to mark `rb` as non-standard.

fasiha:
@sideshowbarker thank you! I of course defer to your knowledge, but could you help me understand, I see in the [HTML living standard](https://html.spec.whatwg.org/) several references to the `rb` tag, which led me to believe it was indeed part of the spec, am I misreading this?

ddbeck:
@fasiha You're right there are some mentions of `rb`, but it's also explicitly named [as obsolete and non-conforming](https://html.spec.whatwg.org/multipage/obsolete.html#non-conforming-features).

I guess it's a little philosophical whether something in the spec described as being out-of-spec is standards track or not. I'll defer to @sideshowbarker on that point. 😄

fasiha:
Ahh, I see! Many thanks for that clarification @ddbeck!, though it makes my tears flow 😭.

(Crying because, without `rb`, pages with `ruby` break Control-F searching and copy-paste in major browsers 😢.)

PR #6613, opened 2020-08-31
Mark the <rb> element with partial_implementation for Chrome and Safari

sideshowbarker:
This change marks the `rb` element with `partial_implementation` for Safari and Chrome and all Blink/Chrome-based browsers. It also adds a note explaining what’s partially implemented and what’s not.

Fixes https://github.com/mdn/browser-compat-data/issues/6475

### DIFF
--- a/html/elements/rb.json
+++ b/html/elements/rb.json
@@ -11,7 +11,7 @@
               "notes": "Blink has support for parsing the <code>rb</code> element, but not for rendering <code>rb</code> content as expected."
             },
             "chrome_android": {
-              "version_added": true,
+              "version_added": "18",
               "partial_implementation": true,
               "notes": "Blink has support for parsing the <code>rb</code> element, but not for rendering <code>rb</code> content as expected."
             },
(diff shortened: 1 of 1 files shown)


html/elements/rb.json
```diff
@@ -6,13 +6,19 @@
...
           "support": {
             "chrome": {
-              "version_added": "5"
+              "version_added": "5",
+              "partial_implementation": true,
+              "notes": "Blink has support for parsing the <code>rb</code> element, but not for rendering <code>rb</code> content as expected."
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": true,
```
jpmedley, line 14:
```suggestion
              "version_added": "18",
```

html/elements/rb.json
```diff
@@ -24,22 +30,34 @@
...
             },
             "samsunginternet_android": {
-              "version_added": true
+              "version_added": true,
+              "partial_implementation": true,
+              "notes": "Blink has support for parsing the <code>rb</code> element, but not for rendering <code>rb</code> content as expected."
             },
             "webview_android": {
-              "version_added": true
+              "version_added": true,
```
jpmedley, line 58:
```suggestion
              "version_added": "37",
```

html/elements/rb.json
jpmedley: If this was in Chrome 5 then it would have been in Chrome Android 18.

foolip (approved):

PR #6614, opened 2020-08-31
Mark the <rb> element as non-standard

sideshowbarker:
This change marks the `rb` element with `"standard_track": false` — because the `rb` element isn’t in fact actually part of the HTML standard.

### DIFF
--- a/html/elements/rb.json
+++ b/html/elements/rb.json
@@ -44,7 +44,7 @@
           },
           "status": {
             "experimental": false,
-            "standard_track": true,
+            "standard_track": false,
             "deprecated": false
           }
         }


foolip:
Should this go together with some changes in https://developer.mozilla.org/en-US/docs/Web/HTML/Element/rb? Should developers avoid this?

sideshowbarker:
> Should this go together with some changes in https://developer.mozilla.org/en-US/docs/Web/HTML/Element/rb?

Yes — so I went ahead just now and added `{{Non-standard_Header}}` there

> Should developers avoid this?

It seems so, given that it Chrome and Safari don’t actually render it as expected.

See the specific wording of the warning the `{{Non-standard_Header}}` adds to https://developer.mozilla.org/en-US/docs/Web/HTML/Element/rb:

> This feature is non-standard and is not on a standards track. Do not use it on production sites facing the Web: it will not work for every user. There may also be large incompatibilities between implementations and the behavior may change in the future.

foolip:
> Chrome and Safari don’t actually render it as expected

If there are any notes that could be added to the data that would help web developers understand the situation, that'd be great. But I'll go ahead and merge this now.

foolip (approved):

Issue #6466, opened 2020-08-04
<dialog> support in Firefox


plonibarploni:
Firefox has enabled `<dialog>` in Nightly beginning with Firefox 80 (see [bug 1645046](https://bugzilla.mozilla.org/show_bug.cgi?id=1645046)). How should this be reflected in the compat data, if at all?

foolip:
I've tested https://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8406 in Firefox 80 on BrowserStack and Firefox Developer Edition 81 locally shows that it's still a HTMLUnknownElement. It doesn't look like this feature is enabled by default, after all.

sideshowbarker:
At least in current Nightly — 82.0a1 (2020-08-31) — for https://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8406 I get `log: object "[object HTMLDialogElement]"…` (not `HTMLUnknownElement`).

So I wonder if it’s the case that maybe the first release branch this went into was the version 82 branch.

sideshowbarker:
Sorry, ignore my previous comment. I’d forgotten that I have the `dom.dialog_element.enabled` preference set to true in my Nightly. When I set that to false in v82 current Nightly, I also get `HTMLUnknownElement`.

So yeah I agree it doesn’t seem to actually be enabled yet — not even in 82, let alone in 80.

smaug----:
dialog is currently enabled on Nightly only.dom. https://searchfox.org/mozilla-central/rev/0c682c4f01442c3de0fa6cd286e9cadc8276b45f/modules/libpref/init/StaticPrefList.yaml#1590,1593

sideshowbarker:
For the record here, in https://freenode.logbot.info/whatwg/20200901#c4947970 @smaug---- also pointed out to me that `dom.dialog_element.enabled` is set to true by default in Nightly. So manually setting it to false isn’t reflective of the default support.

sideshowbarker:
Paging in @sefeng211 (assignee for https://bugzilla.mozilla.org/show_bug.cgi?id=1645046)

queengooborg:
If something is enabled by default only in Nightly/Aurora/Dev builds, we'd been just mentioning as such in notes.  However, considering that we're working to remove flag data from BCD, I don't feel like this is something we wish to record, since most users will be on the normal builds anyways.

PR #6615, opened 2020-08-31
Mark <dialog> element as supported in Firefox 80

sideshowbarker:
This change marks the `dialog` element and the `HTMLDialogElement` interface as supported in Firefox 80+.

Fixes https://github.com/mdn/browser-compat-data/issues/6466

### DIFF
--- a/api/HTMLDialogElement.json
+++ b/api/HTMLDialogElement.json
@@ -13,17 +13,22 @@
           "edge": {
             "version_added": "79"
           },
-          "firefox": {
-            "version_added": "53",
-            "flags": [
-              {
-                "type": "preference",
-                "name": "dom.dialog_element.enabled",
-                "value_to_set": "true"
-              }
-            ],
-            "notes": "See <a href='https://bugzil.la/840640'>bug 840640</a>."
-          },
+          "firefox": [
+            {
+              "version_added": "80"
+            },
+            {
+              "version_added": "53",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "dom.dialog_element.enabled",
+                  "value_to_set": "true"
+                }
+              ],
+              "notes": "See <a href='https://bugzil.la/840640'>bug 840640</a>."
+            }
+          ],
           "firefox_android": {
             "version_added": "53",
             "flags": [
@@ -125,17 +130,22 @@
             "edge": {
               "version_added": "79"
             },
-            "firefox": {
-              "version_added": "53",
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "dom.dialog_element.enabled",
-                  "value_to_set": "true"
-                }
-              ],
-              "notes": "See <a href='https://bugzil.la/840640'>bug 840640</a>."
-            },
+            "firefox": [
+              {
+                "version_added": "80"
+              },
+              {
+                "version_added": "53",
+                "flags": [
+                  {
+                    "type": "preference",
+                    "name": "dom.dialog_element.enabled",
+                    "value_to_set": "true"
+                  }
+                ],
+                "notes": "See <a href='https://bugzil.la/840640'>bug 840640</a>."
+              }
+            ],
             "firefox_android": {
               "version_added": "53",
               "flags": [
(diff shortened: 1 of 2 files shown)


foolip (approved):
Looks good, but I'd like @ddbeck to confirm this is how we want to handle flags. At what point in the future should we remove them?

sideshowbarker:
> At what point in the future should we remove [the flags]?

We won’t need to remove them later, right? — because the state of the data after this patch is such that it indicates the flag was needed from Firefox 53 up until Firefox 80, and that from Firefox 80 on, no flag is needed.

foolip:
@sideshowbarker just a heads up that https://github.com/mdn/browser-compat-data/blob/2834c7f5844698a992dca6b328f2328f2df283fd/GOVERNANCE.md#peers says "work on public branches of their forks". That's something I'm not used to either, takes extra care to avoid it especially when using the web UI.

foolip:
> We won’t need to remove them later, right?

We don't strictly need to, but is there any use in having information about flags for features that have already shipped? If there is, should we also try to document when the flags themselves are removed from the codebase? (This makes it impossible to disable the feature.)

sideshowbarker:
> there any use in having information about flags for features that have already shipped

It’s useful information for developers who need to target users who are still on older versions of browsers

foolip:
Do you mean they would disable the feature in the most recent version of the browser to emulate an older version?

sideshowbarker:
> Do you mean they would disable the feature in the most recent version of the browser to emulate an older version?

No, I just mean that in general at any given time, some number of users aren’t using the latest browser release but instead running older releases. So in general, developers continue to want to know what the behavior for a particular feature is in an any older browser versions that their users might be using.

foolip:
I've found https://github.com/mdn/browser-compat-data/issues/3318 which is a long discussion about feature flags. I will make my case there, if I find I have a strong opinion :)

foolip:
On this change itself, testing https://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8406 in Firefox 80 on BrowserStack and Firefox Developer Edition 81 locally shows that it's still a HTMLUnknownElement. I'll comment on https://github.com/mdn/browser-compat-data/issues/6466 as well.

sideshowbarker:
> On this change itself, testing https://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8406 in Firefox 80 on BrowserStack and Firefox Developer Edition 81 locally shows that it's still a HTMLUnknownElement.

Agreed. So this PR definitely shouldn’t be merged yet. And so I’ve now flipped it to Draft status.

ddbeck:
For a little more on #3318, I put forward a proposal — https://github.com/mdn/browser-compat-data/issues/3318#issuecomment-460185392 — but it never got traction. We still have this semi-weird state of having no official policy about how you're supposed to handle "dead flags". But I _like_ my proposal of marking on-by-default flags as "removed" 😄

foolip:
@ddbeck I like that proposal too, you have my thumbs up. I would like it even more if flag data was allowed to be removed after a feature has been enabled by default, but recognize that the data could be filtered like that by a simple script.

alshamiri5 (approved):

ddbeck:
The chief question here: do we have any indication there's movement on this feature? This has been in draft status for a while, but it's not clear to me if this would be resolved any time soon. Any thoughts, @sideshowbarker?

A couple of minor miscellaneous details below:

---

@foolip I just noticed this about the governance doc, from https://github.com/mdn/browser-compat-data/pull/6615#issuecomment-683721836

> just a heads up that https://github.com/mdn/browser-compat-data/blob/2834c7f5844698a992dca6b328f2328f2df283fd/GOVERNANCE.md#peers says "work on public branches of their forks". That's something I'm not used to either, takes extra care to avoid it especially when using the web UI.

I'm not sure why we have that in the governance doc. I don't think we've ever actually enforced it. I think using `username/` branches is OK (though I personally work in a fork because I dread accidentally pushing to `master`) and is in the spirit of the governance doc, if not the letter.

---

I think the whole prefs question was settled by the adoption of [the irrelevant flag data guideline](https://github.com/mdn/browser-compat-data/blob/master/docs/data-guidelines.md#removal-of-irrelevant-flag-data) in #6670. Looking here, it does not appear that the flag data can be removed.

sideshowbarker:
> The chief question here: do we have any indication there's movement on this feature? This has been in draft status for a while, but it's not clear to me if this would be resolved any time soon.

Yeah I’m going ahead and closing this. It can always be re-raised later. I don’t personally have any idea what the status is on the Gecko implementation. To revisit this, we’d need somebody from Firefox product development to weigh in with info.

sideshowbarker:
> @foolip I just noticed this about the governance doc, from [#6615 (comment)](https://github.com/mdn/browser-compat-data/pull/6615#issuecomment-683721836)
> 
> > just a heads up that https://github.com/mdn/browser-compat-data/blob/2834c7f5844698a992dca6b328f2328f2df283fd/GOVERNANCE.md#peers says "work on public branches of their forks". That's something I'm not used to either, takes extra care to avoid it especially when using the web UI.
> 
> I'm not sure why we have that in the governance doc. I don't think we've ever actually enforced it. I think using `username/` branches is OK

I think I will start doing that again, then

> (though I personally work in a fork because I dread accidentally pushing to `master`)

I think that risk could be avoided by configuring the GitHub settings for the repo such that the only way to get merges into the main branch is from PRs.

That can be set at https://github.com/mdn/browser-compat-data/settings/branches under **Branch protection rules**, with the **Require pull request reviews before merging** checkbox.

ddbeck:
I had never looked at branch protection rules before. Good idea, @sideshowbarker! I opened #7171 to track introducing that and revising GOVERNANCE accordingly.

Issue #6520, opened 2020-08-16
`request.body.formData` don't work in safari or TP


jimmywarting:
I have tried it in safari's console but it don't work. the compatibility table says it is supported...

https://github.com/mdn/browser-compat-data/blob/4c4326016dc01976d0eb1bd5acca35a0116cab2c/api/Body.json#L489

(tested on safari v14)
![image](https://user-images.githubusercontent.com/1148376/90340000-4e306f00-dff5-11ea-81ab-8b4ae4853f14.png)

There is even an open issue about this here:
https://bugs.webkit.org/show_bug.cgi?id=212858

jimmywarting:
Hey 
btw, it later turned out that safari did support `res.formData()` 
but only if the Form has at least one entry (the formdata could not be empty)

```js
// Works
fd = new FormData()
fd.set('a', 'b')
new Response(fd).fromData()
```

```js
// Do not work
fd = new FormData()
new Response(fd).fromData()
```

I later created a WPT test for this and it turns out chrome have this bug too

- wpt: https://github.com/web-platform-tests/wpt/issues/28607
- chrome bug: https://bugs.chromium.org/p/chromium/issues/detail?id=1204594 (still affect chrome 104)
- safari bug: https://bugs.webkit.org/show_bug.cgi?id=225238 (still affect safari 15.6.1)

foolip:
@jimmywarting if you want that to be reflected in BCD, can you file a new issue? A comment on a closed issue will only show up as a notification, and will easily be lost.

PR #6616, opened 2020-08-31
Mark Request.p.formData partial in Safari

sideshowbarker:
Per https://webkit.org/b/212858, `Request.prototype.formData` isn’t fully supported in Safari. So this change marks it with `"partial_implementation": true` and adds a note with a link to the relevant WebKit bug.

Fixes https://github.com/mdn/browser-compat-data/issues/6520

### DIFF
--- a/api/Body.json
+++ b/api/Body.json
@@ -486,10 +486,14 @@
               "version_added": "44"
             },
             "safari": {
-              "version_added": "10"
+              "version_added": "10",
+              "partial_implementation": true,
+              "notes": "See <a href='https://webkit.org/b/212858'>WebKit bug 212858</a>"
             },
             "safari_ios": {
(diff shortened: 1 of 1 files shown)


foolip (approved):

foolip:
I came across https://bugs.webkit.org/show_bug.cgi?id=215671 just now, fixed very recently.

sideshowbarker:
> I came across https://bugs.webkit.org/show_bug.cgi?id=215671 just now, fixed very recently.

Nice find. So after it’s clear that’s gone into a stable release (or will be), we can re-test and update the data here.

foolip:
I've filed https://github.com/mdn/browser-compat-data/issues/6620 to follow up on that.

PR #6617, opened 2020-08-31
Update Body mixin support data for Safari

foolip:
The initial implementation was in WebKit 602.1.19:
https://trac.webkit.org/changeset/195954/webkit

However, it was all [Conditional=FETCH_API] and didn't ship until Safari
10.1 / iOS 10.3 together with the rest of fetch().

The body attribute was added in WebKit 605.1.4:
https://trac.webkit.org/changeset/221329/webkit

Note that the commit message talks about Request, not Response.

### DIFF
--- a/api/Body.json
+++ b/api/Body.json
@@ -70,10 +70,10 @@
             }
           ],
           "safari": {
-            "version_added": "10"
+            "version_added": "10.1"
           },
           "safari_ios": {
-            "version_added": "10"
+            "version_added": "10.3"
           },
           "samsunginternet_android": {
             "version_added": "4.0"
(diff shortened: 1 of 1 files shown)


foolip:
This is a follow-up to https://github.com/mdn/browser-compat-data/pull/6616, where I thought I could quickly remove the true.

I suspect this is still wrong, that some of these things didn't work until later, like `formData`, and that some of them might have worked on `Response` but not `Request`.

@jakearchibald @youennf, looking at https://developer.mozilla.org/en-US/docs/Web/API/Body#Browser_compatibility, is there more that's clearly wrong here?

api/Body.json
```diff
@@ -427,10 +427,10 @@
               }
             ],
             "safari": {
-              "version_added": false
```
foolip, line 430: This is `bodyUsed`. It was there from the start, just with `[ImplementedAs=isDisturbed]` which might have misled some previous attempt here.

sideshowbarker (approved):

PR #6618, opened 2020-08-31
Complete nodejs data for RegExp

gilmoreorless:
Part of #6249. This fills in all the `true` and `null` values for Node in `RegExp`.

Most of the entries were supported in the first version of Node, because they were also in Chrome v1. For the ones that remained I checked the data against Chrome/V8 releases and https://node.green, as well as some manual testing.

There was one oddity I noticed in the browser data though, for `javascript.builtins.RegExp.source.empty_regex_string`. This was only added to V8 in version 3.10.6 ([release notes](https://chromium.googlesource.com/v8/v8/+/refs/tags/3.10.6/ChangeLog)), which means Node 0.8.0 (backed up by manual testing). This equates to Chrome 20 ([according to Wikipedia, anyway](https://en.wikipedia.org/wiki/Google_Chrome_version_history)), but the data currently lists Chrome 6 for that entry. I've left the Chrome entry alone here, but it might be worth a separate PR.

### DIFF
--- a/javascript/builtins/RegExp.json
+++ b/javascript/builtins/RegExp.json
@@ -25,7 +25,7 @@
               "version_added": "4"
             },
             "nodejs": {
-              "version_added": true
+              "version_added": "0.1.100"
             },
             "opera": {
               "version_added": "5"
@@ -77,7 +77,7 @@
                 "version_added": "4"
               },
               "nodejs": {
-                "version_added": true
+                "version_added": "0.1.100"
               },
               "opera": {
                 "version_added": "5"
@@ -129,7 +129,7 @@
                 "version_added": "4"
               },
               "nodejs": {
-                "version_added": true
+                "version_added": "0.1.100"
               },
               "opera": {
                 "version_added": "6"
@@ -243,7 +243,7 @@
                 "version_added": "4"
               },
               "nodejs": {
-                "version_added": true
+                "version_added": "0.1.100"
               },
               "opera": {
                 "version_added": "5"
@@ -347,7 +347,7 @@
                 "version_added": "5.5"
               },
               "nodejs": {
-                "version_added": true
+                "version_added": "0.1.100"
               },
               "opera": {
                 "version_added": "5"
@@ -397,7 +397,7 @@
                   "version_added": "5.5"
                 },
                 "nodejs": {
-                  "version_added": true
+                  "version_added": "6.0.0"
                 },
                 "opera": {
                   "version_added": "35"
@@ -450,7 +450,7 @@
                 "version_added": "5.5"
               },
               "nodejs": {
-                "version_added": true
+                "version_added": "0.1.100"
               },
               "opera": {
                 "version_added": "5"
@@ -500,7 +500,7 @@
                   "version_added": "5.5"
                 },
                 "nodejs": {
-                  "version_added": true
+                  "version_added": "6.0.0"
                 },
                 "opera": {
                   "version_added": "35"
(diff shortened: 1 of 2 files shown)


vinyldarkscratch:
Hey @gilmoreorless, thanks for tackling this!  Regarding the Node.js browser data, I'm thinking that it may make sense to change 0.9.1 to 0.8.0, since they're both the same V8 version (confirmed on https://nodejs.org/en/download/releases/).  It may be better to follow it up in another PR -- I'm going to do some more digging if there were some notable changes between the two versions.

gilmoreorless:
I agree with consolidating versions, but I had a memory that 0.9.1 was added specifically for a non-V8-related feature.
Digging around found #3160. The `setImmediate` and `clearImmediate` methods were added in Node 0.9.1, as listed in the API docs: https://nodejs.org/api/timers.html#timers_setimmediate_callback_args

I also have an annoyance with 0.1.100 being the first version listed in the BCD files. I believe it's because `console.log` was added in that version ([ref](https://nodejs.org/api/console.html#console_console_log_data_args)), but many features were in Node from earlier than that. It mostly annoys me just because of how it displays on MDN ;) 
image

ddbeck:
> I also have an annoyance with 0.1.100 being the first version listed in the BCD files. I believe it's because `console.log` was added in that version ([ref](https://nodejs.org/api/console.html#console_console_log_data_args)), but many features were in Node from earlier than that. It mostly annoys me just because of how it displays on MDN ;)

This is a fair point, @gilmoreorless. I've only though about it for a minute here, but it seems like permitting `≤0.1.00` to cover such cases would be appropriate. If you want to work on fixing this, I'd invite you to open an issue on the topic and we can plan it out. 😃

ddbeck:
Hi @gilmoreorless, I'm sorry this seems to have fallen through the cracks. Do you know if any of the data here is affected by the distinction between `≤0.1.100` / `0.1.100`? If that's a blocker for this data, then I'm happy to help work on that problem. If none of these features predated `0.1.100`, then we can move toward merging instead. Thank you!

gilmoreorless:
@ddbeck Don't worry, it slipped off my radar too! I've raised #6861 to cover the discussion about the minimum version.

For this PR, all the `RegExp` features that are listed as `0.1.100` were in the very first version of Node, as they were also in Chrome version 1. However, I don't think this PR should be blocked on the minimum version decision, as plenty of other data files are going to have to be updated anyway.

javascript/builtins/RegExp.json
```diff
@@ -1765,7 +1765,8 @@
                 "version_added": false
               },
               "nodejs": {
-                "version_added": true
+                "version_added": "6.0.0",
+                "notes": "Case folding is implemented in version 8.6.0"
```
ddbeck, line 1769: Can we reframe this a little, to front-load the fact that it's a historic deficiency?

```suggestion
                "notes": "Before 8.6.0, case folding is not supported."
```

ddbeck (changes requested):
Thank you for your incredible patience on this, @gilmoreorless. I have one suggestion on a comment. Otherwise, this looks great.

(Expect an answer on the Node.js version soon!)

javascript/builtins/RegExp.json
gilmoreorless: Good suggestion. I think I chose that wording to be consistent with the existing note for `edge` a few lines above. I'm happy to change it, but probably both notes should be changed together for consistency.

ddbeck: Ah, I missed that. Yeah, let's do that in a separate PR. I'll approve and merge this now. Thank you! 🎉

ddbeck (approved):
🚀

PR #6621, opened 2020-08-31
CSS Update easing-function jump keyword

vsn4ik:
Safari 14.0 supports it. In Safari TP, it is included in 106:
https://webkit.org/blog/10580/release-notes-for-safari-technology-preview-106/
https://trac.webkit.org/changeset/261046/webkit/

Ref: #6461


sideshowbarker (approved):
Also confirmed the Opera change included in this patch

Issue #6344, opened 2020-06-27
WebKit supports WindowOrWorkerGlobalScope#origin


saschanaz:
https://developer.mozilla.org/en-US/docs/Web/API/WindowOrWorkerGlobalScope/origin

MDN says it's not supported but the support was added three years ago: https://bugs.webkit.org/show_bug.cgi?id=168023

PR #6634, opened 2020-09-02
Mark WindowOrWorkerGlobalScope.p.origin as supported in Safari 11+

sideshowbarker:
This change marks `WindowOrWorkerGlobalScope.prototype.origin` as supported in Safari 11 — per https://bugs.webkit.org/show_bug.cgi?id=168023, which per https://trac.webkit.org/browser/webkit/tags/Safari-604.2.4/Source/WebCore/page/WindowOrWorkerGlobalScope.idl#L31 was first added in Safari 11, because per https://trac.webkit.org/browser/webkit/tags/Safari-603.2.1/Source/WebCore/page/WindowOrWorkerGlobalScope.idl it was not in the tag for Safari 10.1

Fixes https://github.com/mdn/browser-compat-data/issues/6344

### DIFF
--- a/api/WindowOrWorkerGlobalScope.json
+++ b/api/WindowOrWorkerGlobalScope.json
@@ -1108,10 +1108,10 @@
               "version_added": false
             },
             "safari": {
-              "version_added": false
+              "version_added": "11"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "11"
             },
             "samsunginternet_android": {
               "version_added": "7.0"


foolip (approved):
This was added in https://trac.webkit.org/changeset/214147/webkit, WebKit 604.1.13. That matches Safari 11 and Safari for iOS 11. I've also confirmed it's in Safari 11.1 on BrowserStack, but don't have access to Safari 11 to validate that. Still, I'm 99% confident this is right.

Issue #6102, opened 2020-05-06
Fix SourceBuffer.audioTracks/videoTracks/textTracks support version number 


saschanaz:
From https://github.com/mdn/browser-compat-data/issues/5781#issuecomment-599274380:

>* Chrome added flagged support for audioTracks/videoTracks in [March 2016](https://chromium.googlesource.com/chromium/src/+/92333c0e5ca6b7768e5bcf6d41ff8fc733ebbfc1), but no textTracks 
>* Safari added support for all of them in [January 2015](https://trac.webkit.org/changeset/178172/webkit)

So the Chrome version number should be at least 49 instead of current 23 and the Safari one should be at least 11 instead of current 8.

SnyderTheOne:
On https://developer.mozilla.org/en-US/docs/Web/API/SourceBuffer, 'Firefox for Android' lists `SourceBuffer` as an unsupported feature, even though this seems to be implemented. Support for MSE is listed correctly on https://developer.mozilla.org/en-US/docs/Web/API/Media_Source_Extensions_API.

jculver050:
Close

saschanaz:
Fixed by #6644, #6655, etc.

Issue #5781, opened 2020-03-04
SourceBuffer.audioTracks/videoTracks/textTracks are not actually supported


saschanaz:
[BCD says it's supported for all browsers for a long time](https://developer.mozilla.org/en-US/docs/Web/API/SourceBuffer/audioTracks) but actually both Firefox and Chrome does not have them.

saschanaz:
* Firefox never supported them https://searchfox.org/mozilla-central/source/dom/webidl/SourceBuffer.webidl
* Chrome added flagged support for audioTracks/videoTracks in [March 2016](https://chromium.googlesource.com/chromium/src/+/92333c0e5ca6b7768e5bcf6d41ff8fc733ebbfc1), but no textTracks
* Safari added support for all of them in [January 2015](https://trac.webkit.org/changeset/178172/webkit)

sideshowbarker:
I’m closing this issue as being fixed by #5828. I’m aware of the comment there which says:

> This only removes unsupported features without fixing version numbers for supported ones

…but *“fixing version numbers for supported ones”* can be handled by another PR or issue.

foolip:
https://github.com/mdn/browser-compat-data/pull/5828 only partially fixed this, the support data for audioTracks/videoTracks is still wrong for at least Chrome.

saschanaz:
Ah, I opened #6102 for that.

sideshowbarker:
It seems like we could really use some guidance and insight from @jpmedley on this one.

foolip:
I've looked up where these things were implemented and sent https://github.com/mdn/browser-compat-data/pull/6655 to sort it out.

Rant: I kinda wish we didn't have flags in BCD, it's *a lot* of work to sort out when it's wrong.

PR #6635, opened 2020-09-02
Update audioTracks/videoTracks/textTracks versions for Chrome

sideshowbarker:
This change corrects the Chrome version data for the `audioTracks`, `videoTracks`, and `textTracks` members of the `SourceBuffer` interface to reflect, per https://chromium.googlesource.com/chromium/src/+/92333c0e5ca6b7768e5bcf6d41ff8fc733ebbfc1, the actual Chrome versions which shipped support for them.

Fixes https://github.com/mdn/browser-compat-data/issues/6102

### DIFF
--- a/api/SourceBuffer.json
+++ b/api/SourceBuffer.json
@@ -446,10 +446,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/SourceBuffer/audioTracks",
           "support": {
             "chrome": {
-              "version_added": "23"
+              "version_added": "49"
             },
             "chrome_android": {
-              "version_added": "33"
+              "version_added": "49"
             },
             "edge": {
               "version_added": "12"
@@ -465,10 +465,10 @@
               "notes": "Only works on Windows 8+."
             },
             "opera": {
-              "version_added": "15"
+              "version_added": "36"
             },
             "opera_android": {
-              "version_added": "14"
+              "version_added": "36"
             },
             "safari": {
               "version_added": "8"
@@ -477,10 +477,10 @@
               "version_added": false
             },
             "samsunginternet_android": {
-              "version_added": "3.0"
+              "version_added": "5.0"
             },
             "webview_android": {
-              "version_added": "4.4.3"
+              "version_added": "49"
             }
           },
           "status": {
@@ -1321,10 +1321,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/SourceBuffer/videoTracks",
           "support": {
             "chrome": {
-              "version_added": "23"
+              "version_added": "49"
             },
             "chrome_android": {
-              "version_added": "33"
+              "version_added": "49"
             },
             "edge": {
               "version_added": "12"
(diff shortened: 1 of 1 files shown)


foolip (commented):
These APIs have most likely been protected by ifdefs and/or runtime flags. Testing http://mdn-bcd-collector.appspot.com/api/SourceBuffer/ on iPhones in BrowserStack the API seems to have been enabled between Safari 12 and 13. But https://caniuse.com/#feat=mediasource says MSE was added in Safari for iOS 8.

@sideshowbarker can you dig into what happened here? Perhaps it would be best split out the Safari changes, since a lot seems to be wrong with that data. @vinyldarkscratch perhaps this can be mostly filled in and confirmed by the collector, so save us all a lot of manual testing?

foolip:
Note the `[RuntimeEnabled=AudioVideoTracks]` in https://chromium.googlesource.com/chromium/src/+/92333c0e5ca6b7768e5bcf6d41ff8fc733ebbfc1%5E%21/#F19. That means it could have been enabled at a much later time. In fact, per http://mdn-bcd-collector.appspot.com/api/SourceBuffer/audioTracks/ it seems to still not be enabled...

sideshowbarker:
> @sideshowbarker can you dig into what happened here?

I haven’t been able to make time yet to look back into it. I will try to next week.

> Perhaps it would be best split out the Safari changes, since a lot seems to be wrong with that data.

OK, I opened #6644 for that (and amended the change on this branch to reduce it to just the Chrome changes)

api/SourceBuffer.json
```diff
@@ -446,10 +446,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/SourceBuffer/audioTracks",
           "support": {
             "chrome": {
-              "version_added": "23"
+              "version_added": "49"
```
foolip, line 449: Testing in https://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8419, audioTracks and videoTracks aren't shipped in Chrome yet. It's all still behind a flag: https://chromium.googlesource.com/chromium/src/+/9670bbbca886d7b15084ecec18601a23c42066b9/third_party/blink/renderer/platform/runtime_enabled_features.json5#181

I've reopened https://github.com/mdn/browser-compat-data/issues/5781.

foolip (commented):
Requesting that all the data be changed to no support instead.

sideshowbarker:
@jpmedley, this is another one (related to #5781) that it’d be great to have your insight and guidance on

foolip:
I've sent https://github.com/mdn/browser-compat-data/pull/6655 which includes the changes in this PR and other bits that go together.

ddbeck:
@sideshowbarker It looks to me that #6655 has superseded this, but a rebase would probably confirm it

foolip:
Since I caused the conflicts, I've tried merging master into this PR and resolving the conflicts. Afterwards, no differences to master are left, so there is indeed nothing left to do here. Closing.

PR #450, opened 2017-10-03
Next batch of Web Audio API BCD

chrisdavidmills:
* AudioNode
* AudioParam
* AudioProcessingEvent

Issue #4178, opened 2019-05-17
Document guidelines for opening and reviewing issues and PRs


ddbeck:
This is a follow up to #4024.

As a reviewer, I want to have a written set of PR and issue reviewing guidelines and best practices to refer to, so I can review and respond to issues and PRs more fairly and consistently.

As a contributor, I want to have a written set of PR and issue guidelines and best practices, so I can get my PRs merged and issues accepted faster.

Acceptance criteria:

- [ ] Add subsections to [`contributing.md#opening-issues-and-pull-requests`](https://github.com/mdn/browser-compat-data/blob/master/docs/contributing.md#opening-issues-and-pull-requests) for submitter and reviewer guidelines
- [ ] Include a few known guidelines and practices (e.g., welcome new contributors, wait times before pushing to a contributor's branch, CI must pass, etc.)
- [ ] Add a link to the guidelines from `.github/PULL_REQUEST_TEMPLATE` (maybe?)

ddbeck:
We should also document how to inform other reviewers that a PR is blocked on a specific reviewer, rather than any reviewer at all.

Inspired by https://github.com/mdn/browser-compat-data/pull/6638#issuecomment-686456104

PR #6638, opened 2020-09-02
Remove AudioParam.cancelValuesAndHoldAtTime which never shipped

foolip:
The name was changed to cancelAndHoldAtTime and backported to the M57
branch, so that cancelValuesAndHoldAtTime was never shipped:
https://bugs.chromium.org/p/chromium/issues/detail?id=686242#c3

Confirmed by testing http://mdn-bcd-collector.appspot.com/api/AudioParam/
in Chrome 56 and 57.

The source of this data is https://github.com/mdn/browser-compat-data/pull/450.

### DIFF
--- a/api/AudioParam.json
+++ b/api/AudioParam.json
@@ -99,26 +99,12 @@
         "__compat": {
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/AudioParam/cancelAndHoldAtTime",
           "support": {
-            "chrome": [
-              {
-                "version_added": "57"
-              },
-              {
-                "version_added": true,
-                "version_removed": "56",
-                "alternative_name": "cancelValuesAndHoldAtTime"
-              }
-            ],
-            "chrome_android": [
-              {
-                "version_added": "57"
-              },
-              {
-                "version_added": true,
-                "version_removed": "56",
-                "alternative_name": "cancelValuesAndHoldAtTime"
-              }
-            ],
+            "chrome": {
+              "version_added": "57"
+            },
+            "chrome_android": {
+              "version_added": "57"
+            },
             "edge": {
               "version_added": "≤79"
             },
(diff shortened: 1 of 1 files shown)


foolip:
@chrisdavidmills I'm quite confident I got this right now, but do you know what the source of this in https://github.com/mdn/browser-compat-data/pull/450 was? Was that a PR porting data from the MDN tables?

chrisdavidmills:
> I'm quite confident I got this right now, but do you know what the source of this in #450 was? Was that a PR porting data from the MDN tables?

Yes, that was a port of info over from the old style manual tables to BCD. We didn't check whether any of that data was correct; we just ported it ;-)

sideshowbarker:
About this PR having gotten merged despite having pending review request: One general practice that we all might want to consider following (and documenting) is that when we have a PR that we don’t merged until some particular action takes place — whether that be a review from a particular reviewer, or a kind of “needsinfo” case of being blocked waiting for response to a particular comment, or whatever — the PR submitter can set the status of the PR to Draft.

A PR with a Draft status can’t be merged until the PR submitter or an admin explicitly changes the Draft status back to normal (non-Draft). So having it at Draft status seems like an effective way to indicate to everybody that it’s not ready to be merged yet — and to prevent it from being inadvertently merged prematurely.

foolip:
@ddbeck, I asked @jpmedley for review because this changes Chrome data and he's offered to help with reviews, not because he's the only one who could review this, so no problem at all.

If I did want to ensure someone takes a look at something, I'd spell it out with more words. But even if someone misses that and merges first, one can always do more review after the fact.

ddbeck:
I like the idea of being explicit when a _specific_ reviewer is required, rather than merely requested. #4178 is already open on the topic of documenting our reviewer practices anyway, so I'll note that there.

Thanks for your understanding on this!

PR #1579, opened 2018-03-18
Corrections to Web Audio

jpmedley:
Look particularly at `AudioScheduledSourceNode` and [its children](https://developer.mozilla.org/en-US/docs/Web/API/AudioScheduledSourceNode). I hope I've fixed the weirdness around moving `onended`, `start()`, and `stop()` from concrete interfaces to the  `AudioScheduledSourceNode` parent class in Chrome 57.

PR #6642, opened 2020-09-03
Results from mdn-bcd-collector for various Multimedia APIs

vinyldarkscratch:
This PR provides results collected from [mdn-bcd-collector](https://github.com/foolip/mdn-bcd-collector) for various Multimedia APIs.  All versions have been verified manually in SauceLabs and through searching Chromium Code Search to verify the commits.  (All invalid data was removed from this PR and either corrected on the collector or separated into a new PR.)

Another PR will be coming later to cover the rest of the changes.

### DIFF
--- a/api/AudioBuffer.json
+++ b/api/AudioBuffer.json
@@ -61,7 +61,7 @@
               "notes": "The <code>context</code> parameter was supported up until version 57, but has now been removed."
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "79"
             },
             "firefox": {
               "version_added": "53"
@@ -107,10 +107,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/AudioBuffer/copyFromChannel",
           "support": {
             "chrome": {
-              "version_added": "14"
+              "version_added": "43"
             },
             "chrome_android": {
-              "version_added": "18"
+              "version_added": "43"
             },
             "edge": {
               "version_added": "13"
@@ -125,22 +125,22 @@
               "version_added": false
             },
             "opera": {
-              "version_added": "15"
+              "version_added": "30"
             },
             "opera_android": {
-              "version_added": "14"
+              "version_added": "30"
             },
             "safari": {
-              "version_added": "6"
+              "version_added": false
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": false
             },
             "samsunginternet_android": {
-              "version_added": "1.0"
+              "version_added": "4.0"
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "43"
             }
           },
           "status": {
@@ -155,10 +155,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/AudioBuffer/copyToChannel",
           "support": {
             "chrome": {
-              "version_added": "14"
+              "version_added": "43"
             },
             "chrome_android": {
-              "version_added": "18"
+              "version_added": "43"
             },
             "edge": {
               "version_added": "13"
@@ -173,10 +173,10 @@
               "version_added": false
             },
             "opera": {
-              "version_added": "15"
+              "version_added": "30"
             },
             "opera_android": {
-              "version_added": "14"
+              "version_added": "30"
             },
             "safari": {
               "version_added": false
@@ -185,10 +185,10 @@
               "version_added": false
             },
             "samsunginternet_android": {
-              "version_added": "1.0"
+              "version_added": "4.0"
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "43"
             }
           },
           "status": {
--- a/api/AudioBufferSourceNode.json
+++ b/api/AudioBufferSourceNode.json
@@ -371,10 +371,10 @@
               "version_added": "18"
             },
             "safari": {
-              "version_added": true
+              "version_added": "6.1"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "7"
             },
             "samsunginternet_android": {
               "version_added": "2.0"
--- a/api/AudioBufferSourceNode.json
+++ b/api/AudioBufferSourceNode.json
@@ -61,7 +61,7 @@
               "notes": "Before version 59, the default values were not supported."
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "79"
             },
             "firefox": {
               "version_added": "53"
@@ -203,7 +203,7 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/AudioBufferSourceNode/loop",
           "support": {
             "chrome": {
-              "version_added": "14"
+              "version_added": "15"
             },
             "chrome_android": {
               "version_added": "18"
@@ -251,10 +251,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/AudioBufferSourceNode/loopEnd",
           "support": {
             "chrome": {
-              "version_added": "14"
+              "version_added": "24"
             },
             "chrome_android": {
-              "version_added": "18"
+              "version_added": "25"
             },
             "edge": {
               "version_added": "12"
@@ -281,10 +281,10 @@
               "version_added": true
             },
             "samsunginternet_android": {
-              "version_added": "1.0"
+              "version_added": "1.5"
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "≤37"
             }
           },
           "status": {
@@ -299,10 +299,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/AudioBufferSourceNode/loopStart",
           "support": {
             "chrome": {
-              "version_added": "14"
+              "version_added": "24"
             },
             "chrome_android": {
-              "version_added": "18"
+              "version_added": "25"
             },
             "edge": {
               "version_added": "12"
@@ -329,7 +329,7 @@
               "version_added": true
             },
             "samsunginternet_android": {
-              "version_added": "1.0"
+              "version_added": "1.5"
             },
             "webview_android": {
               "version_added": true
@@ -347,16 +347,16 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/AudioBufferSourceNode/onended",
           "support": {
             "chrome": {
-              "version_added": "14"
+              "version_added": "30"
             },
             "chrome_android": {
-              "version_added": "18"
+              "version_added": "30"
             },
             "edge": {
               "version_added": "12"
             },
             "firefox": {
-              "version_added": null
+              "version_added": true
             },
             "firefox_android": {
               "version_added": null
@@ -365,24 +365,22 @@
               "version_added": false
             },
             "opera": {
-              "version_added": true,
-              "version_removed": "44"
+              "version_added": "17"
             },
             "opera_android": {
-              "version_added": true,
-              "version_removed": "43"
+              "version_added": "17"
             },
             "safari": {
-              "version_added": null
+              "version_added": true
             },
             "safari_ios": {
               "version_added": null
             },
             "samsunginternet_android": {
-              "version_added": "1.0"
+              "version_added": "2.0"
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "4.3"
             }
           },
           "status": {
@@ -445,15 +443,13 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/AudioBufferSourceNode/start",
           "support": {
             "chrome": {
-              "version_added": "14",
-              "version_removed": "57"
+              "version_added": "24"
             },
             "chrome_android": {
-              "version_added": "18"
+              "version_added": "25"
             },
             "edge": {
-              "version_added": "12",
-              "version_removed": "79"
+              "version_added": "12"
             },
             "firefox": {
               "version_added": "25"
@@ -477,10 +473,10 @@
               "version_added": true
             },
             "samsunginternet_android": {
-              "version_added": "1.0"
+              "version_added": "1.5"
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "≤37"
             }
           },
           "status": {
--- a/api/AudioBufferSourceNode.json
+++ b/api/AudioBufferSourceNode.json
@@ -368,7 +368,7 @@
               "version_added": "17"
             },
             "opera_android": {
-              "version_added": "17"
+              "version_added": "18"
             },
             "safari": {
               "version_added": true
@@ -380,7 +380,7 @@
               "version_added": "2.0"
             },
             "webview_android": {
-              "version_added": "4.3"
+              "version_added": "4.4"
             }
           },
           "status": {
--- a/api/AudioParam.json
+++ b/api/AudioParam.json
@@ -61,25 +61,25 @@
               "version_added": "≤79"
             },
             "firefox": {
-              "version_added": null
+              "version_added": false
             },
             "firefox_android": {
-              "version_added": null
+              "version_added": false
             },
             "ie": {
               "version_added": false
             },
             "opera": {
-              "version_added": true
+              "version_added": "55"
             },
             "opera_android": {
-              "version_added": true
+              "version_added": "48"
             },
             "safari": {
-              "version_added": null
+              "version_added": false
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": false
             },
             "samsunginternet_android": {
               "version_added": "10.0"
--- a/api/AudioParamMap.json
+++ b/api/AudioParamMap.json
@@ -17,7 +17,7 @@
             "version_added": "76"
           },
           "firefox_android": {
-            "version_added": null
+            "version_added": "79"
           },
           "ie": {
             "version_added": false
@@ -64,7 +64,7 @@
               "version_added": "76"
             },
             "firefox_android": {
-              "version_added": null
+              "version_added": "79"
             },
             "ie": {
               "version_added": false
@@ -112,7 +112,7 @@
               "version_added": "76"
             },
             "firefox_android": {
-              "version_added": null
+              "version_added": "79"
             },
             "ie": {
               "version_added": false
@@ -160,7 +160,7 @@
               "version_added": "76"
             },
             "firefox_android": {
-              "version_added": null
+              "version_added": "79"
             },
             "ie": {
               "version_added": false
@@ -208,7 +208,7 @@
               "version_added": "76"
             },
             "firefox_android": {
-              "version_added": null
+              "version_added": "79"
             },
             "ie": {
               "version_added": false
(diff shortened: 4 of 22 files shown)


api/AudioBufferSourceNode.json
```diff
@@ -365,24 +365,22 @@
...
+              "version_added": "17"
             },
             "opera_android": {
-              "version_added": true,
-              "version_removed": "43"
+              "version_added": "18"
             },
             "safari": {
-              "version_added": null
+              "version_added": true
```
foolip, line 374: This was added in https://trac.webkit.org/changeset/150905/webkit, at WebKit 537.44.0. That would be Safari 7 and Safari for iOS 7. Can you test if that's right?

It was added for `OscillatorNode` at the same time. Can you update the data for api.OscillatorNode.onended to match what this ends up saying?

api/AudioBufferSourceNode.json
```diff
@@ -347,16 +347,16 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/AudioBufferSourceNode/onended",
           "support": {
             "chrome": {
-              "version_added": "14"
+              "version_added": "30"
```
foolip, line 350: Everywhere one looks closely, something is a bit broken: https://github.com/mdn/browser-compat-data/issues/6656

But anyway, the `onended` attribute was added in https://chromium.googlesource.com/chromium/src/+/976bdc021a39e160f5d0d4da9b0d477eb4232e69 for AudioBufferSourceNode and OscillatorNode. That was indeed M30 per https://www.chromium.org/developers/calendar.

api/AudioBufferSourceNode.json
```diff
@@ -445,15 +443,13 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/AudioBufferSourceNode/start",
           "support": {
             "chrome": {
-              "version_added": "14",
-              "version_removed": "57"
+              "version_added": "24"
```
foolip, line 446: What was the test for this? The AudioBufferSourceNode interface wasn't always exposed, and it seems weird to have this interface at all without the start() method, so what happened in Chrome 24? https://github.com/mdn/browser-compat-data/issues/6656 also makes me suspect tests could get this wrong.

Since this is messy, splitting it out from this PR and handling separately might be good.

api/AudioBufferSourceNode.json
```diff
@@ -445,15 +443,13 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/AudioBufferSourceNode/start",
           "support": {
             "chrome": {
-              "version_added": "14",
-              "version_removed": "57"
```
foolip, line 449: The 57 here was probably added because `start` went away from `AudioBufferSourceNode.prototype`, see https://github.com/mdn/browser-compat-data/issues/6656#issuecomment-688998331.

api/AudioParamMap.json
```diff
@@ -14,7 +14,7 @@
             "version_added": "≤79"
           },
           "firefox": {
-            "version_added": null
+            "version_added": "76"
           },
           "firefox_android": {
             "version_added": null
```
foolip, line 20: Can you make a guess for firefox_android too, here and elsewhere that you're updating only firefox?

api/AudioScheduledSourceNode.json
```diff
@@ -71,10 +71,10 @@
             }
           ],
           "safari": {
-            "version_added": null
+            "version_added": false
```
foolip, line 74: The IDL file was added in https://trac.webkit.org/changeset/264538/webkit so soon this will be supported. But for now, this is correct. Per https://github.com/mdn/browser-compat-data/issues/6656 it doesn't really matter, but we can sort it out there.

api/AudioTrack.json
```diff
@@ -664,7 +664,7 @@
               "version_added": false
             },
             "ie": {
-              "version_added": "10"
+              "version_added": false
```
foolip, line 667: Can you also update api.VideoTrack.sourceBuffer? They should go together, otherwise something odd is happening.

api/AudioTrackList.json
```diff
@@ -783,7 +783,7 @@
               ]
             },
             "ie": {
-              "version_added": "10"
+              "version_added": "11"
```
foolip, line 786: This will be out of sync with the `removetrack` event entry now. Can you check if the event itself is fired in IE10 and IE11? Was it just the `onremovetrack` event handler attribute that was missing in IE10?

api/MediaDevices.json
```diff
@@ -29,10 +29,10 @@
             "version_added": "30"
           },
           "safari": {
-            "version_added": "11"
+            "version_added": "13"
```
foolip, line 32: Can you ensure that `navigator.mediaDevices` was also not in Safari 11? Or is this a case of missing interface object?

foolip (commented):
It's a big PR, I've reviewed up to api/MediaEncryptedEvent.json now :)

api/MediaDevices.json
foolip: Or better still, can you update api.Navigator.mediaDevices to match, in this PR?

api/AudioBuffer.json
```diff
@@ -125,22 +125,22 @@
...
-              "version_added": "15"
+              "version_added": "30"
             },
             "opera_android": {
-              "version_added": "14"
+              "version_added": "30"
             },
             "safari": {
-              "version_added": "6"
+              "version_added": false
```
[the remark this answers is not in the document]
vinyldarkscratch: I don't have access to Safari 14 so I cannot confirm this myself.  I'm a little bit skeptical about marking it as supported without testing it and confirming first?

api/AudioBufferSourceNode.json
```diff
@@ -203,7 +203,7 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/AudioBufferSourceNode/loop",
           "support": {
             "chrome": {
-              "version_added": "14"
+              "version_added": "15"
```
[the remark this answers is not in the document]
vinyldarkscratch: This one was confirmed manually via my own Chromium VMs -- `loop` was `false` in Chrome 15 but `undefined` in Chrome 14.  I think this may have been one of my own adjustments, as we don't have any Chrome data in the collector for anything below Chrome 26.

api/AudioBufferSourceNode.json
```diff
@@ -365,24 +365,22 @@
               "version_added": false
             },
             "opera": {
-              "version_added": true,
-              "version_removed": "44"
```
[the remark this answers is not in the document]
vinyldarkscratch: It looks like this came from #1579, however double-checking in Opera, this was __not__ removed from the browser, so I'm not sure where that data came from.

api/AudioBufferSourceNode.json
vinyldarkscratch: Looks like it was actually added in Safari 6.1.  Updating the data now!

api/AudioParamMap.json
vinyldarkscratch: I left it `null` since I was waiting on the PR to add Fenix (FxA 79), but now that it's merged I can do the mirroring!

api/AudioBufferSourceNode.json
vinyldarkscratch: The test code was as follows:

```javascript
(function() {
  var audioCtx = new (window.AudioContext || window.webkitAudioContext)();
  if (!audioCtx) {return false};
  var instance = audioCtx.createBufferSource();
  return 'start' in instance;
})();
```

Interestingly enough, running it in Chrome 15 and Chrome 23 _do_ return `false`, and Chrome 24 returns `true`.

The original data came from Confluence, it appears -- see #1346.

api/AudioBufferSourceNode.json
vinyldarkscratch: It's quite possible -- the data came from Confluence, it appears (see #1346).

api/AudioBuffer.json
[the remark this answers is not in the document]
foolip: Yeah, let's leave it as false. I also don't have the macOS 14 beta.

api/AudioBufferSourceNode.json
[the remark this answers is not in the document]
foolip: OK. Maybe it was `looping` before, but it's not worth the effort to find out really.

api/AudioBufferSourceNode.json
foolip: Thanks!

api/AudioBufferSourceNode.json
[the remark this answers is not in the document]
foolip: @jpmedley FYI that this seeming mistaken change is being undone. If there's more to it than meets the eye, please shout! :)

api/AudioBufferSourceNode.json
foolip: I checked Safari 6.1 was WebKit 537.43. That's very close. Perhaps these changes were backported to a release branch, or my method for determining the WebKit version at a change isn't entirely accurate. In any case it's best to keep verifying against the real releases when possible.

api/AudioBufferSourceNode.json
foolip: OK, so you did have an instance here, the test should be reliable. I tried it in http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8434 and also observed first support in Chrome 24.

It looks like https://trac.webkit.org/changeset/129260/webkit (WebKit 537.12.0) is when `start` and `stop` were added to `AudioBufferSourceNode` and `OscillatorNode`, then called just `Oscillator` in WebKit. That matches this being in Chrome 24.

To complete this, can you make the same change to `OscillatorNode`? (That has both `start` and `stop`, while `stop` is oddly missing here...)

api/AudioTrackList.json
foolip: Note to self: this change was backed out.

api/MediaDevices.json
foolip: @vinyldarkscratch what ended up happening here? There are still changes in MediaDevices.json, but not these ones. Is there an issue with the methodology that lead to "13" originally?

api/AudioTrack.json
foolip: Done now, thanks!

api/MediaKeyMessageEvent.json
```diff
@@ -20,20 +20,32 @@
...
           },
-          "safari": {
-            "version_added": null
-          },
-          "safari_ios": {
-            "version_added": null
-          },
+          "safari": [
+            {
+              "version_added": "13.1"
```
foolip, line 33: EME is pretty old and this doesn't seem right, http://web-confluence.appspot.com/#!/catalog?releases=%5B%22Safari_12.0_OSX_10.13.6%22,%22Safari_12.0_OSX_10.14%22,%22Safari_12.1_OSX_10.13.6%22,%22Safari_12.1_OSX_10.14.4%22,%22Safari_13.0.2_OSX_10.15%22%5D&q=%22mediakeymessageevent%22 suggests it was exposed at least from 12.1, although seemingly depending on the OS version.

Can you double check?

api/MediaKeyMessageEvent.json
```diff
@@ -68,7 +80,7 @@
               "version_added": null
             },
             "ie": {
-              "version_added": null
+              "version_added": false
```
foolip, line 83: This seems right, it's for https://developer.mozilla.org/en-US/docs/Web/API/MediaKeySession/onmessage and that whole cluster of APIs seems to not be in IE. https://caniuse.com/eme says an older event-based model is supported, but I haven't checked what exactly that is.

api/MediaKeySession.json
```diff
@@ -29,10 +29,10 @@
             "version_added": true
           },
           "safari": {
-            "version_added": null
+            "version_added": true
```
foolip, line 32: Yep, `window.MediaKeySession` is in Safari 13.1 at least. We can narrow down the range later.

api/MediaKeySession.json
```diff
@@ -29,10 +29,10 @@
...
           },
           "safari_ios": {
-            "version_added": null
+            "version_added": true
```
foolip, line 35: Yep, MediaKeySession is there at least on my iPhone with iOS 14 beta.

api/MediaKeys.json
```diff
@@ -29,10 +29,10 @@
             "version_added": true
           },
           "safari": {
-            "version_added": null
+            "version_added": true
           },
           "safari_ios": {
-            "version_added": null
+            "version_added": true
```
foolip, line 35: Confirmed that `MediaKeys` is there on an iPhone.

api/OscillatorNode.json
```diff
@@ -201,10 +201,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/OscillatorNode/onended",
           "support": {
             "chrome": {
-              "version_added": "14"
+              "version_added": "30"
```
foolip, line 204: Commented elsewhere, but can you also update `start` and `stop` for `OscillatorNode`?

foolip (commented):
OK, I have finished review! It'd be great to break out anything which isn't a quickfix so that I don't have to do a full re-review after a bunch of iteration.

api/AudioBufferSourceNode.json
vinyldarkscratch: Going to https://developer.mozilla.org/docs/Web/API/AudioBufferSourceNode/stop seems to redirect to `AudioScheduledSourceNode` instead.  To limit the scope of this PR, I will cherry-pick these changes into their own PR, removing `AudioBufferSourceNode.start` and just keeping `AudioScheduledSourceNode.start`.

api/MediaDevices.json
vinyldarkscratch: The `MediaDevices` interface wasn't directly exposed until Safari 13, so this is another case of needing to test an instance rather than the interface.  `navigator.mediaDevices` was present in Safari 11.

api/MediaEncryptedEvent.json
```diff
@@ -20,19 +20,19 @@
...
-            "version_added": true
+            "version_added": "29"
           },
           "opera_android": {
-            "version_added": true
+            "version_added": "29"
           },
           "safari": {
-            "version_added": null
+            "version_added": "13.1"
```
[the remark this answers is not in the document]
vinyldarkscratch: I don't fully trust the results for this, and there's no MDN web docs that describe this API.  Reverting.

api/MediaKeyMessageEvent.json
vinyldarkscratch: I don't fully trust the results for this, reverting the changes.

api/MediaKeySession.json
```diff
@@ -124,10 +124,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": null
+              "version_added": "13.1"
```
foolip, line 127: @vinyldarkscratch there are still a bunch of Safari EME changes you didn't revert in https://github.com/mdn/browser-compat-data/pull/6642/commits/a277c11fe16cc326d745352f012213bfae560369. This is the first. Since it's the only remaining issue with the PR, I'll push a commit reverting the remaining Safari changes in api/MediaKey*.json, and then land this.

api/AudioBuffer.json
```diff
@@ -107,10 +107,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/AudioBuffer/copyFromChannel",
           "support": {
             "chrome": {
-              "version_added": "14"
+              "version_added": "43"
```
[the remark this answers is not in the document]
jpmedley: There was an AudioBuffer in 2010 and it's part of the current interface's history, so yes, it is the same AudioBuffer. 

https://source.chromium.org/chromium/chromium/src/+/a1e775bff79314ab8068aef688451e51adb517ef

api/MediaCapabilities.json
```diff
@@ -145,10 +145,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": false
```
foolip, line 148: This is wrong `navigator.mediaCapabilities.decodingInfo` exists in Safari 13.1, and `'decodingInfo' in MediaCapabilities.prototype` is true. @vinyldarkscratch something seems to have gone wrong with the process here.

api/MediaKeySession.json
```diff
@@ -346,7 +346,7 @@
               "version_added": "55"
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "79"
```
foolip, line 349: Confluence agrees that `onkeystatuseschange` and `onmessage` was not in Edge 18, FWIW.

foolip (commented):
On final review I found one more issue. @vinyldarkscratch can you check it out?

api/MediaCapabilities.json
foolip: Actually, in order to get this huge PR landed, I'll revert it. Can you check it out separately?

api/AudioParam.json
```diff
@@ -61,25 +61,25 @@
...
-              "version_added": true
+              "version_added": "55"
             },
             "opera_android": {
-              "version_added": true
+              "version_added": "48"
             },
             "safari": {
-              "version_added": null
+              "version_added": false
```
foolip, line 79: Given the other mistake with Safari I checked this one, and `'automationRate' in AudioParam.prototype` is false in Safari 13.1.

api/AudioParamMap.json
```diff
@@ -14,25 +14,25 @@
...
-            "version_added": true
+            "version_added": "53"
           },
           "opera_android": {
-            "version_added": true
+            "version_added": "47"
           },
           "safari": {
-            "version_added": null
+            "version_added": false
```
foolip, line 32: Seems right since there's no `AudioParamMap` in WebKit source.

foolip (approved):
OK, I did one more pass, can't spot any more suspicious changes to check.

jpmedley:
@vinyldarkscratch I was in the process of commenting when this merged because I believe there to be some incorrect data in the PR. For example, there was an AudioBuffer in 2010 and it's part of the current interface's history, so yes, it is the same AudioBuffer. 

https://source.chromium.org/chromium/chromium/src/+/a1e775bff79314ab8068aef688451e51adb517ef

foolip:
@jpmedley the AudioBuffer changes were for two methods, and I linked the commit that introduced them to verify these results from black-box testing.

Nevertheless, if there are errors in anything @vinyldarkscratch and I have worked on and merged, just comment and we'll fix it.

api/AudioBuffer.json
```diff
@@ -125,22 +125,22 @@
               "version_added": false
             },
             "opera": {
-              "version_added": "15"
+              "version_added": "30"
             },
             "opera_android": {
-              "version_added": "14"
+              "version_added": "30"
```
Muhammad4599, line 131: f2bd98a7fa8abad593b09b55e6cefbb3c1396a6e

Muhammad4599 (changes requested):

PR #6643, opened 2020-09-03
CSS ::marker shipping in Chromium 86

Loirooriol:
CSS ::marker shipping in Chromium 86:
https://www.chromestatus.com/feature/6710566854852608

Also, animation and transition support only started working (behind flag) in 83:
https://storage.googleapis.com/chromium-find-releases-static/a46.html#a46d064de471a4f4705cbbffc65d69bac2a44736

I know the data because I implemented ::marker in Chromium.

### DIFF
--- a/css/selectors/marker.json
+++ b/css/selectors/marker.json
@@ -6,33 +6,48 @@
           "description": "<code>::marker</code>",
           "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/::marker",
           "support": {
-            "chrome": {
-              "version_added": "80",
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "Enable experimental Web Platform features"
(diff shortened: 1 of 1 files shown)


foolip (commented):
Thanks for the updates! If this is also enabled in WebView, can you update the "webview_android" entries as well? "opera", "opera_android" and "samsunginternet_android" will also need updating eventually, but OK to leave them for now since they don't all have releases based on M86 known yet.

Loirooriol:
@foolip Good catch, I have updated webview_android.

foolip (approved):
Thanks for updating WebView! M86 confirmed by https://storage.googleapis.com/chromium-find-releases-static/43c.html#43ce278dfdf8b1897286591564aa32e58743d6f6.

Issue #6326, opened 2020-06-24
Add "iPadOS" to label of the "Safari on iOS" column


a2sheppy:
We currently label the mobile Safari column "Safari on iOS". We should change this to either "Safari on iOS/iPadOS" or "Safari (iOS/iPadOS)". This will help to ensure clarity.

Elchi3:
The current labeling was discussed with Apple and approved by them, so I don't think we should just change this.
@apple-web-evangelist, what do you think about this suggestion?

a2sheppy:
Was that before or after iPadOS was separated from iOS? I don’t recall the timing of this decision.


Eric "Sheppy" Shepherd
Senior Technical Writer
MDN Web Docs
Mozilla

ddbeck:
See https://github.com/mdn/browser-compat-data/blob/master/docs/data-guidelines.md#safari-for-ios-versioning and #2006 for background on this.

I'm disinclined to do this without some reassurances from Apple that iOS and iPadOS data is strictly equivalent. To change the identifier for iOS to also cover iPadOS now would be a strong retroactive claim about the data. For example, we haven't been asking contributors to show some evidence for both iOS _and_ iPadOS when we're reviewing iOS data. If we changed the identifier, it might give people a false idea about the rigor of that data.

othermaciej:
Maybe it should be a separate column? There's some things that are available in iPadOS Safari but not iOS (for example, Media Source Extensions), and I think also vice versa. Another example is Fullscreen API.

vinyldarkscratch:
iOS and iPadOS are still just about the same OS at this point, and I doubt there will be much compatibility differences in the future.  As such, I don't think that creating a new column and a new browser entry are the right way to go about it.  Since there are few differences in compatibility between Safari for iOS and Safari for iPadOS, using notes has been sufficient.

Of course, there could be major compatibility changes down the line.  Let's see what Apple wants us to do going forward!

othermaciej:
We prefer to align Safari for all platforms when possible, but when there's a split, in the future it will more likely be between macOS+iPadOS on the one hand, and iOS on the other.

vinyldarkscratch:
(I just saw your profile and realize that you work at Apple, @othermaciej -- oops, sorry for not noticing before!)

othermaciej:
No worries, I didn't exactly highlight it. We'll try to come up with a proposal for the best way to handle this issue too.

a2sheppy:
[...]

PR #6644, opened 2020-09-04
Update audioTracks/videoTracks/textTracks versions for Safari

sideshowbarker:
This change corrects the Safari version data for the `audioTracks`, `videoTracks`, and `textTracks` members of the `SourceBuffer` interface to reflect, per https://trac.webkit.org/changeset/178172/webkit, the actual version of Safari which first shipped support for them.

Relates to https://github.com/mdn/browser-compat-data/issues/6102

### DIFF
--- a/api/SourceBuffer.json
+++ b/api/SourceBuffer.json
@@ -75,7 +75,9 @@
             "version_added": "8"
           },
           "safari_ios": {
-            "version_added": false
+            "version_added": "13",
+            "partial_implementation": true,
+            "notes": "Exposed in Mobile Safari on iPad but not on iPhone."
           },
           "samsunginternet_android": [
             {
@@ -515,7 +517,9 @@
               "version_added": "8"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "13",
+              "partial_implementation": true,
+              "notes": "Exposed in Mobile Safari on iPad but not on iPhone."
             },
             "samsunginternet_android": {
               "version_added": false
@@ -1179,7 +1183,9 @@
               "version_added": "8"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "13",
+              "partial_implementation": true,
+              "notes": "Exposed in Mobile Safari on iPad but not on iPhone."
             },
             "samsunginternet_android": {
               "version_added": false
@@ -1431,7 +1437,9 @@
               "version_added": "8"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "13",
+              "partial_implementation": true,
+              "notes": "Exposed in Mobile Safari on iPad but not on iPhone."
             },
             "samsunginternet_android": {
               "version_added": false
--- a/api/HTMLMediaElement.json
+++ b/api/HTMLMediaElement.json
@@ -252,10 +252,10 @@
               ]
             },
             "safari": {
-              "version_added": true
+              "version_added": "8"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "8"
             },
             "samsunginternet_android": {
               "version_added": false
@@ -3614,7 +3614,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "8"
+            },
+            "safari_ios": {
+              "version_added": "8"
             }
           },
           "status": {
@@ -3757,7 +3760,10 @@
               ]
             },
             "safari": {
-              "version_added": true
+              "version_added": "8"
+            },
+            "safari_ios": {
+              "version_added": "8"
             },
             "samsunginternet_android": {
               "version_added": false


foolip:
https://trac.webkit.org/changeset/178172/webkit does not seems related to shipping audioTracks/videoTracks/textTracks, it only moves them around in the IDL file and they're still behind `[Conditional=VIDEO_TRACK]`.

foolip:
Understanding when this really shipped is made more complicated by `[NoInterfaceObject]` used on the interface in WebKit until https://trac.webkit.org/changeset/206146/webkit, where it was removed as a drive-by in another change. @vinyldarkscratch FYI you may run into this in the next part of https://github.com/mdn/browser-compat-data/pull/6642.

api/SourceBuffer.json
```diff
@@ -75,7 +75,7 @@
             "version_added": "8"
           },
           "safari_ios": {
-            "version_added": false
+            "version_added": "8"
```
foolip, line 78: I've tested http://software.hixie.ch/utilities/js/live-dom-viewer/saved/8420 on iPhones in BrowserStack to try to confirm when MSE was added, but it's not there even on Safari 13. https://caniuse.com/mediasource suggests that it's only for iPad. If that's correct, we should have notes in the BCD data as well.

api/SourceBuffer.json
```diff
@@ -471,10 +471,10 @@
               "version_added": "14"
             },
             "safari": {
-              "version_added": "8"
+              "version_added": "11"
```
foolip, line 474: I've tested https://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8419 on Safari (macOS) in BrowserStack and see support at least back to 10.1. Before that point I'm not sure because the interface object was not exposed at one point. But at any case the methodology here seems unreliable, so another look would be good.

foolip (commented):
I'm unable to confirm the suggested versions, so more evidence would be great.

sideshowbarker:
Paging in @eric-carlson and @jernoble, who I think will be able to give us some guidance here.

Eric, Jer, we’re trying to figure out two things:

1. What were the first versions of Safari and iOS Safari that shipped with `audioTracks`/`videoTracks`/`textTracks` support?
2. What were the first versions of Safari and iOS Safari that shipped with `SourceBuffer` support?

alshamiri5 (approved):

jernoble:
> 1. What were the first versions of Safari and iOS Safari that shipped with `audioTracks`/`videoTracks`/`textTracks` support?

This first landed in Safari 8 on MacOS X Yosemite, and in Mobile Safari on iOS 8.

> 2. What were the first versions of Safari and iOS Safari that shipped with `SourceBuffer` support?

This also landed in Safari 8 on MacOS X Yosemite, and later in Mobile Safari on iPad in iOS 13.

sideshowbarker:
@jernoble Thanks much

One thing I realize now I should have been more clear about: given that both `HTMLMediaElement` and `SourceBuffer` have `audioTracks`/`videoTracks`/`textTracks` properties, we need to confirm the data for those properties for both the `HTMLMediaElement` and `SourceBuffer` cases.

So can you please either confirm the following, or let me know any further corrections needed —

* `audioTracks`/`videoTracks`/`textTracks` for `HTMLMediaElement` shipped in macOS Safari 8, and in Mobile Safari on iOS 8
* `SourceBuffer` — with `audioTracks`/`videoTracks`/`textTracks` properties implemented/exposed — shipped in macOS Safari 8, and later in Mobile Safari in iOS 13

Specifically, the particular details I’m not completely sure about are:

* When `SourceBuffer` shipped initially, did it implement/expose `audioTracks`/`videoTracks`/`textTracks` properties, or instead was support for those was added later?
* Did Mobile Safari on iPhone in iOS 13 ship with `SourceBuffer` support implemented/exposed? (I ask because in your previous comment, you specifically mention only Mobile Safari on iPad)

jernoble:
> * When `SourceBuffer` shipped initially, did it implement/expose `audioTracks`/`videoTracks`/`textTracks` properties, or instead was support for those was added later?

Yes, those properties were exposed from the start. <https://bugs.webkit.org/show_bug.cgi?id=123374>

> * Did Mobile Safari on iPhone in iOS 13 ship with `SourceBuffer` support implemented/exposed? (I ask because in your previous comment, you specifically mention only Mobile Safari on iPad)

No, Mobile Safari on iPhone does not expose any MSE APIs.

sideshowbarker:
@foolip I amended the commit to reflect the details from Jer.

To document what Jer indicated about `SourceBuffer` being exposed in Safari on iPad but not on iPhone, I marked `safari_ios` for it with `partial_implementation:true` and an *“Exposed in Mobile Safari on iPad but not on iPhone”* note. But I’m not sure if that’s actually the best way to indicate it — I don’t know if we have a precedent — so I guess it would be good if @ddbeck could also weigh in with some guidance.

vinyldarkscratch:
We've discussed separating iPadOS into its own column in #6326 a bit.  For now, I think you've got the right idea to add a note!

ddbeck:
Yep putting this in a note is the way to go right now.

foolip (approved):
Thanks @jernoble for confirming the releases for these bits, and thanks @sideshowbarker! I'll go ahead and land this now.

foolip:
@vinyldarkscratch there ought to be a whole bunch of MSE APIs that can be updated in a similar way to this PR, in case you want to collect results from iPad.

PR #6648, opened 2020-09-06
Add version numbers for Console API

vinyldarkscratch:
This PR adds version numbers for various members of the Console API, based upon manual testing (running `'memberName' in console` in various browsers).  Data is as follows:

api.Console.clear - Chrome 25, IE 8, Opera 12, Safari 6.1
api.Console.count - Chrome 1, IE 11, Opera 11, Safari 4
api.Console.countReset - Chrome 68, Opera 55, Safari 13
api.Console.debug - Chrome 1, IE 11, Opera 10, Safari 4
api.Console.dirxml - Chrome 1, Firefox 39, IE 11, Opera 11, Safari 4
api.Console.group - Opera 11
api.Console.groupCollapsed - Opera 11
api.Console.groupEnd - Opera 11
api.Console.info - Chrome 1, Opera 10.5, Safari ≤3
api.Console.profile - Chrome 4 (incorrectly marked 53), Firefox 16, IE 9, Opera 11, Safari 4
api.Console.profileEnd - Chrome 4, Firefox 16, IE 9, Opera 11, Safari 4
api.Console.timeEnd - Opera 11
api.Console.timeStamp - Chrome 14, Firefox 39, Opera Blink, Safari 6
api.Console.warn - Chrome 1, Opera 10.5, Safari ≤3

This also removes a note on `api.Console.info` for Safari that there was "no information icon", which may have been true in older versions but not in the latest Safari.

### DIFF
--- a/api/Console.json
+++ b/api/Console.json
@@ -1381,9 +1381,9 @@
           }
         }
       },
-      "timeStamp": {
+      "timestamp": {
         "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/Console/timeStamp",
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/Console/timestamp",
           "support": {
             "chrome": {
               "version_added": "14"
--- a/api/Console.json
+++ b/api/Console.json
@@ -878,8 +878,7 @@
               "version_added": "11"
             },
             "safari": {
-              "version_added": true,
-              "notes": "No information icon"
+              "version_added": true
             },
             "safari_ios": {
               "version_added": "1"
--- a/api/Console.json
+++ b/api/Console.json
@@ -114,10 +114,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Console/clear",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "25"
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": "25"
             },
             "edge": {
               "version_added": "12"
@@ -129,25 +129,28 @@
               "version_added": "48"
             },
             "ie": {
-              "version_added": true
+              "version_added": "8"
             },
             "nodejs": {
               "version_added": "8.3.0"
             },
             "opera": {
-              "version_added": true
+              "version_added": "12"
+            },
+            "opera_android": {
+              "version_added": "12"
             },
             "safari": {
-              "version_added": true
+              "version_added": "6.1"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "7"
             },
             "samsunginternet_android": {
-              "version_added": true
+              "version_added": "1.5"
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "≤37"
             }
           },
           "status": {
@@ -162,10 +165,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Console/count",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "1"
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": "18"
             },
             "edge": {
               "version_added": "12"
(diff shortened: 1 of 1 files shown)


api/Console.json
```diff
@@ -231,19 +237,19 @@
               "version_added": "8.3.0"
             },
             "opera": {
-              "version_added": true
+              "version_added": "55"
```
foolip, line 240: Can you add opera_android here as well, as you've done above? (I'll comment just "opera_android" for this in the following.)

api/Console.json
```diff
@@ -231,19 +237,19 @@
...
             },
             "safari": {
-              "version_added": null
+              "version_added": "13"
```
foolip, line 243: This change is also part of https://github.com/mdn/browser-compat-data/pull/6526 which increases my confidence that it's right.

api/Console.json
```diff
@@ -851,20 +872,23 @@
...
             },
             "opera": {
-              "version_added": true
+              "version_added": "10.5"
+            },
+            "opera_android": {
+              "version_added": "11"
             },
             "safari": {
               "version_added": true,
```
foolip, line 881:
```suggestion
              "version_added": "3",
```

This is api.Console.info which you had tested as Safari ≤3. Given that `console.log` is marked as supported from Safari 3, I suggest doing the same here. It's not likely this was introduced before `console.log`, and if they both were introduced earlier, there's very little harm in being wrong between Safari 1 and 3.

api/Console.json
```diff
@@ -1046,22 +1070,22 @@
...
             "edge": {
               "version_added": "12"
             },
             "firefox": {
-              "version_added": true
+              "version_added": "16"
             },
             "firefox_android": {
-              "version_added": "10"
+              "version_added": "16"
```
foolip, line 1085: Since this is increasing the version from 10 to 16, would you mind checking where the original data came from, to see if there might just be something surprising going on here? The same request goes for a few following changes to firefox_android data.

api/Console.json
```diff
@@ -1480,19 +1516,22 @@
...
             },
             "opera": {
-              "version_added": true
+              "version_added": "10.5"
+            },
+            "opera_android": {
+              "version_added": "11"
             },
             "safari": {
               "version_added": true
```
foolip, line 1525:
```suggestion
              "version_added": "3"
```

Same reasoning as for console.info.

api/Console.json
```diff
@@ -1353,19 +1386,19 @@
...
             "chrome_android": {
-              "version_added": true
+              "version_added": "18"
             },
             "edge": {
               "version_added": "12"
             },
             "firefox": {
-              "version_added": true
+              "version_added": "39"
```
foolip, line 1398: I've confirmed with software.hixie.ch/utilities/js/live-dom-viewer/saved/8415 that `console.timeStamp` was introduced in Firefox 39. However, there's a capitalization mismatch here. Can you file an issue to follow up on this?

api/Console.json
```diff
@@ -210,10 +216,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Console/countReset",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "68"
```
foolip, line 219: I've confirmed this with http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8416 and BrowserStack.

foolip (commented):
Great update, thank you!

I spot checked two cases which both checked out, the rest I'll trust your testing on. A few comments on a few things I think would still make sense to fix in this PR.

api/Console.json
foolip: Ah, you'd already sent https://github.com/mdn/browser-compat-data/pull/6650. Great!

api/Console.json
vinyldarkscratch: Good catch!  I should've noticed when the `console` API was implemented, haha -- certainly saves the trouble of testing in Safari 1-2!

api/Console.json
vinyldarkscratch: The original data came from the initial MDN wiki migration, so I'm betting that the data was incorrect.

foolip (approved):
LGTM!

PR #6651, opened 2020-09-07
docs: improve data quality reporting step in package publishing guide

ddbeck:
During last week's package publishing, I made a mistake in reporting data quality stats: https://github.com/mdn/browser-compat-data/issues/6369#issuecomment-688166061. This change adds some more detail to the docs, to make it more obvious what stats need to be reported into which issues. Hopefully, with this change, I won't make that mistake twice.

The actual change is c5bcb2f. 86f8f47 is a source-style change.


foolip (approved):
Thanks, these are pretty detailed docs :)

ddbeck:
> these are pretty detailed docs :)

Tech writers, it's how we roll 😄 

I'm going to deem this good enough to merge. Thank you for the review, @foolip!

PR #6655, opened 2020-09-08
Sort out AudioTrack and VideoTrack support in Chromium

foolip:
That was M37 based on the date: https://www.chromium.org/developers/calendar

Note that https://storage.googleapis.com/chromium-find-releases-static/cb9.html#cb9c513a764a62acd2da76e04c3e7ec3ab7e0dfa
claiming this was added in M45 is wrong because this change happened 
before Blink was merged into the Chromium tree.

The AudioVideoTracks has never been enabled by default.

The audioTrack and videoTrack attributes on SourceBuffer were added
later, in April/March 2016, behind the same flag:
https://chromium.googlesource.com/chromium/src/+/1b3dc5a8202a332891b6aed5d4d4521db0b26261
https://chromium.googlesource.com/chromium/src/+/92333c0e5ca6b7768e5bcf6d41ff8fc733ebbfc1

These changes were in M51:
https://storage.googleapis.com/chromium-find-releases-static/1b3.html#1b3dc5a8202a332891b6aed5d4d4521db0b26261
https://storage.googleapis.com/chromium-find-releases-static/923.html#92333c0e5ca6b7768e5bcf6d41ff8fc733ebbfc1

The data was mirrored for Edge and Opera. Samsung Internet and WebView
don't use flags.

Fixes https://github.com/mdn/browser-compat-data/issues/5781.

### DIFF
--- a/api/AudioTrack.json
+++ b/api/AudioTrack.json
@@ -5,7 +5,7 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/AudioTrack",
         "support": {
           "chrome": {
-            "version_added": "45",
+            "version_added": "37",
             "flags": [
               {
                 "type": "preference",
@@ -15,7 +15,7 @@
             ]
           },
           "chrome_android": {
-            "version_added": "45",
+            "version_added": "37",
             "flags": [
               {
                 "type": "preference",
@@ -64,7 +64,7 @@
             "version_added": "10"
           },
           "opera": {
-            "version_added": "32",
+            "version_added": "24",
             "flags": [
               {
                 "type": "preference",
@@ -74,7 +74,7 @@
             ]
           },
           "opera_android": {
-            "version_added": "32",
+            "version_added": "24",
             "flags": [
               {
                 "type": "preference",
@@ -93,7 +93,7 @@
             "version_added": false
           },
           "webview_android": {
-            "version_added": "45"
+            "version_added": false
           }
         },
         "status": {
@@ -107,7 +107,7 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/AudioTrack/enabled",
           "support": {
             "chrome": {
-              "version_added": "45",
+              "version_added": "37",
               "flags": [
                 {
                   "type": "preference",
@@ -117,7 +117,7 @@
               ]
             },
             "chrome_android": {
-              "version_added": "45",
+              "version_added": "37",
               "flags": [
                 {
                   "type": "preference",
@@ -166,7 +166,7 @@
               "version_added": "10"
             },
             "opera": {
-              "version_added": "32",
+              "version_added": "24",
               "flags": [
                 {
                   "type": "preference",
@@ -176,7 +176,7 @@
               ]
             },
             "opera_android": {
-              "version_added": "32",
+              "version_added": "24",
               "flags": [
                 {
                   "type": "preference",
@@ -195,7 +195,7 @@
               "version_added": false
             },
             "webview_android": {
-              "version_added": "45"
+              "version_added": false
             }
           },
           "status": {
@@ -210,7 +210,7 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/AudioTrack/id",
           "support": {
             "chrome": {
-              "version_added": "45",
+              "version_added": "37",
               "flags": [
                 {
                   "type": "preference",
@@ -220,7 +220,7 @@
               ]
             },
             "chrome_android": {
-              "version_added": "45",
+              "version_added": "37",
               "flags": [
                 {
                   "type": "preference",
@@ -269,7 +269,7 @@
               "version_added": "10"
             },
             "opera": {
-              "version_added": "32",
+              "version_added": "24",
               "flags": [
                 {
                   "type": "preference",
@@ -279,7 +279,7 @@
               ]
             },
             "opera_android": {
-              "version_added": "32",
+              "version_added": "24",
               "flags": [
                 {
                   "type": "preference",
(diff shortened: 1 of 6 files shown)


api/SourceBuffer.json
```diff
@@ -477,10 +518,10 @@
               "version_added": false
             },
             "samsunginternet_android": {
-              "version_added": "3.0"
+              "version_added": false
```
foolip, line 521: The data that was here was just wrong, probably copy-pasted at one point in time.

foolip:
@jpmedley can you review this? I know it's a lot, sorry :)

api/AudioTrack.json
```diff
@@ -93,7 +93,7 @@
             "version_added": false
           },
           "webview_android": {
-            "version_added": "45"
+            "version_added": false
```
jpmedley, line 96: Are you sure about this? It's unlisted in [not-webview-exposed.txt](android_webview/tools/system_webview_shell/test/data/webexposed/not-webview-exposed.txt). In six years, I would think someone would notice the false test failures.

api/HTMLMediaElement.json
```diff
@@ -206,10 +232,24 @@
...
             "opera": {
-              "version_added": true
+              "version_added": "24",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "enable-experimental-web-platform-features",
+                  "value_to_set": "enabled"
+                }
+              ]
```
jpmedley, line 242: I didn't think Opera used our flags.

api/SourceBuffer.json
```diff
@@ -446,14 +446,41 @@
...
-            "edge": {
-              "version_added": "12"
+              "version_added": "51",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "enable-experimental-web-platform-features",
+                  "value_to_set": "enabled"
+                }
+              ]
```
jpmedley, line 466: This doesn't seem like it should be here.

api/SourceBuffer.json
```diff
@@ -477,10 +518,10 @@
               "version_added": false
             },
             "samsunginternet_android": {
-              "version_added": "3.0"
+              "version_added": false
             },
             "webview_android": {
-              "version_added": "4.4.3"
+              "version_added": false
```
jpmedley, line 524: I argued against it, but Mozilla's policy is to leave pre-Chromium WebView data as is.

jpmedley (changes requested):
I have a few questions.

api/AudioTrack.json
foolip: I'm pretty sure, yes. The AudioVideoTracks feature has always been experimental, and in a grep of current Chromium I can't find anything that enables it for any platform from outside runtime_enabled_features.json5, which is what would have to happen in order to have a difference between platforms.

My understanding of not-webview-exposed.txt is that it tracks differences between WebView and some other config. I'm not sure if the test runs with experimental features enabled or not, but in any case there would be no *difference*, between the configs, the APIs would either be present in both or absent in both.

The effective status of all Chromium browsers should be false, it's just made hard to see by all the flag data.

(I edited your comment to fix the link.)

api/HTMLMediaElement.json
foolip: They do. It's under opera://flags and I've just confirmed that enabling experimental web platform features there causes the APIs touched in this PR to be exposed.

In general, I just mirror the data to Opera if the I'm familiar with the change in Chromium, so I know it's not something especially likely to be enabled or disabled by embedders. It's a hassle to maintain the flag data, but that's something to sort out in https://github.com/mdn/browser-compat-data/issues/3318.

api/SourceBuffer.json
foolip: Can you clarify what's wrong? The important bit is that these APIs aren't shipped in any Chromium based browser. Because of the presence of some of the flag data around this, I'm having to represent it has a flagged entry for consistency, rather than just setting `version_added` to false.

api/SourceBuffer.json
foolip: This was wrong, though. There cannot have been a WebView version which had this enable by default, if I'm reading the version history of WebKit and Chromium correctly.

Is the policy around WebView stated somewhere? Is it just that the pre-Chromium releases are preserved in browsers/webview_android.json aren't being removed, or actually that we should keep whatever the data is, even when incorrect?

ddbeck: This is incorrect:

> Mozilla's policy is to leave pre-Chromium WebView data as is

#4479 covers the introduction of `"≤37"` values for WebView. The summary is that we decided to allow ≤37 while continuing to permit more precise versioning (37 or earlier) if a contributor could substantiate it. We didn't freeze any data.

If the feature's unsupported now and you can't find any evidence that it was supported historically and the data conflicts with it, then the data should change.

foolip: Thanks @ddbeck. I'm quite confident this data was wrong, because WebView 4.4.3 was using Chromium M33, and the code for this feature didn't even exist until M51.

foolip:
@jpmedley can you review again?

api/SourceBuffer.json
jpmedley: Sorry. This is pre-Chromium edge, but it uses the name of a Chromium flag. This could be confusing.

foolip: The line you've commented on is in the chrome_android block, and in the Edge block below the flag is on the Chromium-based Edge, version 79.

If I'm still missing it, can you comment on the exact line you'd like to see changed?

jpmedley (approved):

PR #6657, opened 2020-09-08
fix(css): unprefix content size values for Safari and Opera

kripod:
Tested with [BrowserStack](https://www.browserstack.com/) on an iPhone 8 Plus (iOS 11) and macOS High Sierra.

- [x] Summarize your changes

### DIFF
--- a/css/properties/min-width.json
+++ b/css/properties/min-width.json
@@ -158,10 +158,15 @@
               "ie": {
                 "version_added": false
               },
-              "opera": {
-                "prefix": "-webkit-",
-                "version_added": "15"
-              },
+              "opera": [
+                {
+                  "version_added": "33"
+                },
+                {
+                  "prefix": "-webkit-",
+                  "version_added": "15"
+                }
+              ],
               "opera_android": {
                 "version_added": "33"
               },
@@ -250,10 +255,15 @@
               "ie": {
                 "version_added": false
               },
-              "opera": {
-                "prefix": "-webkit-",
-                "version_added": "15"
-              },
+              "opera": [
+                {
+                  "version_added": "33"
+                },
+                {
+                  "prefix": "-webkit-",
+                  "version_added": "15"
+                }
+              ],
               "opera_android": {
                 "version_added": "33"
               },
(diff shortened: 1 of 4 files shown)


css/properties/min-width.json
```diff
@@ -158,10 +158,15 @@
               "ie": {
                 "version_added": false
               },
-              "opera": {
-                "prefix": "-webkit-",
-                "version_added": "15"
-              },
+              "opera": [
```
ddbeck, line 161: Did you mean to include these Opera changes here? If you want, we can do a real clean up here (I think we also need to fix up the Android data while we're here) and change the PR description, or we can defer this to a separate PR. It's up to you.

ddbeck (changes requested):
This checks out and is backed up by the WebKit bug tracker:

https://bugs.webkit.org/show_bug.cgi?id=169195

(It looks like several unprefixes happened along with this one, if you're interested in checking those values too. 😄)

One question about some seemingly-unrelated Opera data here, but otherwise, this looks great. Thank you!

css/properties/min-width.json
kripod: Oh right, this was intended. I've just renamed the PR to reflect these changes, thank you for the suggestion!

kripod: As for the Android data, I'm not quite sure about which browser you're referring to. My intent was to keep data up-to-date for browsers which are [on the default Browserslist](https://github.com/browserslist/browserslist#queries).

ddbeck: Thanks for the clarification on the browsers here.

Some background on the Android stuff: we often mirror closely-related browsers within our data. In this case, the `opera` (desktop) 33 and and 15 versions corresponds to the equivalent versions for `opera_android` (just below) — they happen to share the same engine version, in this case.

Which is to say, if you wanted to copy the whole support array for `opera` to `opera_android`, that would spare the trouble of a separate PR. But this is optional. If you don't want to, I'll still merge this PR (and make sure it gets into the next release). Because it's ready to go, with or without that change.

kripod: Thank you!

Unfortunately, I currently have no time to modify this PR, sorry.

ddbeck (approved):
Thank you, @kripod! 🎉

PR #6658, opened 2020-09-08
Update when getUserMedia() shipped in Chromium

foolip:
navigator.mediaDevices.getUserMedia() and navigator.getUserMedia() were
enabled together in M53:
https://chromium.googlesource.com/chromium/src/+/d229d009b74255797c16bef6b7e398a165dbeacd

The data for navigator.getUserMedia() was already correct, except for
Opera where it incorrectly appeared to still be prefixed-only.

The change to require a secure context for these APIs was made long
before it was shipped:
https://chromium.googlesource.com/chromium/src/+/dac67a137b9c8d3be8d51bb9fe8ab761d8b054a7

The versions for api.MediaDevices.getUserMedia.secure_context_required
should thus be the same as for the parent feature.

### DIFF
--- a/api/MediaDevices.json
+++ b/api/MediaDevices.json
@@ -367,11 +367,11 @@
           "support": {
             "chrome": {
               "version_added": "53",
-              "notes": "Older versions of Chrome implement <code>navigator.webkitGetUserMedia</code>, a prefixed form of the legacy <a href='https://developer.mozilla.org/docs/Web/API/Navigator/getUserMedia'><code>navigator.getUserMedia</code></a> API."
+              "notes": "If you need this capability before version 53, refer to <code>navigator.webkitGetUserMedia</code>, a prefixed form of the deprecated <a href='https://developer.mozilla.org/docs/Web/API/Navigator/getUserMedia'><code>navigator.getUserMedia</code></a> API."
             },
             "chrome_android": {
               "version_added": "53",
-              "notes": "Older versions of Chrome implement <code>navigator.webkitGetUserMedia</code>, a prefixed form of the legacy <a href='https://developer.mozilla.org/docs/Web/API/Navigator/getUserMedia'><code>navigator.getUserMedia</code></a> API."
+              "notes": "If you need this capability before version 53, refer to <code>navigator.webkitGetUserMedia</code>, a prefixed form of the deprecated <a href='https://developer.mozilla.org/docs/Web/API/Navigator/getUserMedia'><code>navigator.getUserMedia</code></a> API."
             },
             "edge": {
               "version_added": "12"
@@ -379,7 +379,7 @@
             "firefox": {
               "version_added": "36",
               "notes": [
-                "Older versions of Firefox implement <code>navigator.mozGetUserMedia</code>, a prefixed form of the legacy <a href='https://developer.mozilla.org/docs/Web/API/Navigator/getUserMedia'><code>navigator.getUserMedia</code></a> API.",
+                "If you need this capability before version 36, refer to <code>navigator.mozGetUserMedia</code>, a prefixed form of the deprecated <a href='https://developer.mozilla.org/docs/Web/API/Navigator/getUserMedia'><code>navigator.getUserMedia</code></a> API.",
                 "Before Firefox 55, <code>getUserMedia()</code> incorrectly returns <code>NotSupportedError</code> when the list of constraints specified is empty, or has all constraints set to <code>false</code>. Starting in Firefox 55, this situation now correctly calls the failure handler with a <code>TypeError</code>.",
                 "When using the Firefox-specific <code>video</code> constraint called <code>mediaSource</code> to request display capture, Firefox 66 and later consider values of <code>screen</code> and <code>window</code> to both cause a list of screens <em>and</em> windows to be shown.",
                 "Starting in Firefox 66, <code>getUserMedia()</code> can no longer be used in sandboxed <code>&lt;iframe&gt;</code>s or <code>data</code> URLs entered in the address bar by the user."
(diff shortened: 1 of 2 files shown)


api/MediaDevices.json
```diff
@@ -367,11 +367,11 @@
...
             "chrome": [
               {
-                "version_added": "52"
+                "version_added": "53"
               },
               {
                 "version_added": "47",
-                "version_removed": "52",
+                "version_removed": "53",
                 "flags": [
```
foolip, line 375: @ddbeck a policy question for you before I can finish this change. The presence of the flag data here is inconsistent between api.MediaDevices.getUserMedia and api.Navigator.getUserMedia, and to my eye have made the data worse, as the "navigator.webkitGetUserMedia" note is on the flag entry, where it should be on the main one if it's worth having at all.

In this situation, is it OK to remove the flag data entirely, or should I make everything consistent by adding it to more places? I've also commented on https://github.com/mdn/browser-compat-data/issues/3318.

api/Navigator.json
```diff
@@ -978,8 +978,7 @@
               },
               {
                 "version_added": "21",
-                "prefix": "webkit",
-                "notes": "An outdated constraint syntax is still in use, but the syntax described here is available through the <a href='https://github.com/webrtc/adapter'>adapter.js</a> polyfill."
```
foolip, line 982: Another question for you @ddbeck. I've found that the different variants of `getUserMedia` take the same constraints dictionary since at least https://chromium.googlesource.com/chromium/src/+/8b986593441f4447834d7471440c6c1f644b0ed1, which is ~5 years ago. Is it OK to remove notes like this per https://github.com/mdn/browser-compat-data/blob/master/docs/data-guidelines.md#removal-of-irrelevant-features, or should I try to figure out when it changed and make sure that's reflected correctly for each browser? That seems like a lot of work and not particularly useful to anyone.

api/MediaDevices.json
ddbeck: The note feels easy to me: I'd drop it entirely (it seems pointless to refer people _from_ a standard feature to a non-standard, prefixed feature).

As for the flag data, I'm less sure. You said the flag data is inconsistent. The implication here is that when `api.MediaDevices.getUserMedia` was behind a flag, so too was `api.Navigator.getUserMedia`? Is that correct? In general, I'd say make the flags consistent, though I wouldn't obligate you to change to `api.Navigator.getUserMedia` in a PR for `api.MediaDevices.getUserMedia` (which is to say, if you don't want to fuss with `Navigator`, you don't have to).

But!

The details here though point to a narrow case for dropping some flag data: flags for features which are both `"deprecated": true` and `"standard_track": false` are the summit (or is that nadir?) of obscurity. There's really no plausible story for web developers to turn on that preference (there's no standard discussion to contribute to, for example). Maybe that's a partial solution to #3318; I'll open an issue for it.

api/Navigator.json
ddbeck: This note is really bad to begin with and I fully support removing it. "the syntax described here" doesn't work out of the context of an MDN page and I'm 👎 on linking to polyfills. Make it go 🔥

foolip:
Thanks @ddbeck! I'll update the PR to get rid of flags and notes.

api/MediaDevices.json
foolip: That's right, `api.MediaDevices.getUserMedia` and `api.Navigator.getUserMedia` were behind the same flag in Chromium and were shipped together. I'm not certain they were *added* behind a flag at the same time, and that's the sort of thing that doesn't feel very productive to research.

Unfortunately, the narrow case doesn't apply here, because the flags are on the `api.MediaDevices.getUserMedia` entry. I would argue there is still no plausible reason fro web developers to want to turn on that flag in an old version of Chrome, however.

As for the "Older versions of Chrome" notes, I guess those are to let people know that they might still be able to use this API under a different name. It serves the same purpose as a `prefix` or `alternative_name` entry, but in this case it was on a different interface so that's not possible.

foolip:
@ddbeck I ended up leaving the notes, but that's not the most important part of this change. This is ready for real review now, and I'll comment on https://github.com/mdn/browser-compat-data/issues/3318 about why I think it's reasonable to remove flags in a situation like this.

api/MediaDevices.json
ddbeck: Yeah, I discovered that my partial solution involving deprecations isn't all that useful (the first match seemed kind of exceptional, which wasn't very helpful).

I have a comment on the note that remains. I'll start a new line comment for that.

api/MediaDevices.json
```diff
@@ -365,40 +365,14 @@
...
-                    "name": "Experimental Web Platform features",
-                    "value_to_set": "Enabled"
-                  }
-                ],
-                "notes": "Older versions of Chrome implement <code>navigator.webkitGetUserMedia</code>, a prefixed form of the legacy <a href='https://developer.mozilla.org/docs/Web/API/Navigator/getUserMedia'><code>navigator.getUserMedia</code></a> API."
-              }
-            ],
+            "chrome": {
+              "version_added": "53",
+              "notes": "Older versions of Chrome implement <code>navigator.webkitGetUserMedia</code>, a prefixed form of the legacy <a href='https://developer.mozilla.org/docs/Web/API/Navigator/getUserMedia'><code>navigator.getUserMedia</code></a> API."
```
ddbeck, line 370: This is a tough note to get right. If this note is going to remain, then we should resolve some ambiguity of "Older versions of Chrome" and the relationship of this method to the older, prefixed one.

We could either reflect the data in the other method:

> From Chrome 21, Chrome implements a similar method, <code>navigator.webkitGetUserMedia()</code>, a prefixed form of the legacy <a href='https://developer.mozilla.org/docs/Web/API/Navigator/getUserMedia'><code>navigator.getUserMedia()</code></a> method.

Or we could be a bit more circumspect (and insulate ourselves from the other method's details):

> Some versions of Chrome before 53 may implement a similar method, <code>navigator.webkitGetUserMedia()</code>, a prefixed form of the legacy <a href='https://developer.mozilla.org/docs/Web/API/Navigator/getUserMedia'><code>navigator.getUserMedia()</code></a> method.

foolip: Oh my, when trying decide here I noticed that the data for when the prefixed variant was introduced also seems to be wrong. Will need to sort that out too...

alshamiri5 (approved):

api/MediaDevices.json
ddbeck: Can I do anything to help this along, @foolip?

foolip: No, I've just left it sitting, it's not blocked on anything. It's bumped up my list now, I'll finish it in the coming days.

foolip: OK, so I've used http://mdn-bcd-collector.appspot.com/tests/api/Navigator/webkitGetUserMedia to try and pin this down, but now find that that on both macOS and Windows, `navigator. webkitGetUserMedia` _was_ introduced in Chrome 21. I'm not sure what it was I tested 3 weeks ago.

The data for `api.Navigator.getUserMedia` looks all correct already, and going over the changes I've made here, I don't think anything is missing. It's ready for review as-is.

foolip: As for the wording of the note, since I've now confirmed this was added in Chrome 21, the "From Chrome 21, Chrome implements a similar method" wording would work. However, note that we also have the same wording for Firefox, "Older versions of Firefox ..."

We also have "Older versions of" notes for `api.EventTarget.dispatchEvent` and other `css.properties.unicode-bidi.isolate` Is the problem that a reader wouldn't be able to figure out which versions this refers to? I don't think I see the problem myself, so I can't have a preference on how to fix it.

ddbeck: The way I see it, "Older versions" could be interpreted as relative to `"version_added"` (in this case, the range of versions 1-52) or relative to the current version of the browser (the range 1-84). I suppose with the note, we're trying to say:

```
If you need this capability before version 53, refer to <code>navigator.webkitGetUserMedia</code>, a prefixed form of the deprecated <a href='https://developer.mozilla.org/docs/Web/API/Navigator/getUserMedia'><code>navigator.getUserMedia</code></a> API.
```

Maybe that's more helpful and easier to apply to the Firefox notes?

foolip: Thanks @ddbeck! Sorry it took me a while, but I've updated to use your notes, and that worked out well I think.

ddbeck (approved):
Excellent tidying up here, @foolip. Thank you! 🎉

PR #6671, opened 2020-09-11
Publish package with GitHub Actions workflow

ddbeck:
This PR moves package publish to GitHub Actions, removes publishing from the Travis config, and updates the docs to reflect the new procedure.

@peterbe could I get you to review this? (Also, I thought you'd be curious, since you were asking about inspecting npm packages before publishing.)


ddbeck:
(After a discussion with Peter, there's a better way to handle this. I'll update it next week.)

ddbeck:
Thanks for the conversation last week, @peterbe. Would you take a look at the publishing workflow again?

(You can safely ignore the docs changes—unless you want to review them—since I'm the only audience for it.)

.github/workflows/node.js.yml
```diff
@@ -16,7 +16,8 @@ jobs:
 
     strategy:
       matrix:
-        node-version: [ 10.x ]
+        # TODO: hoist to 10.x when we bump the version supported in #5931
+        node-version: [ 8.x ]
```
peterbe, line 20: Is this the action that tests across multiple versions? If so, why not list 8.x, 10.x, and 12.x?
Perhaps even [`14.x`](https://github.com/peterbe/minimalcss/blob/e432a59bb085c964dfc354f9327975f65ff78767/.github/workflows/node.js.yml#L19).

.github/workflows/release.yml
```diff
@@ -0,0 +1,26 @@
...
+    runs-on: ubuntu-latest
+    strategy:
+      matrix:
+        # TODO: hoist to 10.x when we bump the version supported in #5931
+        node-version: [ 8.x ]
+
+    steps:
+    - uses: actions/checkout@v2
+    - name: Use Node.js ${{ matrix.node-version }}
+      uses: actions/setup-node@v1
```
peterbe, line 18: You might want to consider using `v2` here instead. https://github.com/actions/setup-node#v2-beta
I'm using it on all the Yari actions and it's working fine there. 

It's your call but perhaps now that this stuff is fresh in your mind, it might worth aiming for the bleeding edge so the stuff will work longer.

.github/workflows/release.yml
```diff
@@ -0,0 +1,26 @@
+name: Release package
+
+on:
+  release:
+    types: [published]
```
peterbe, line 5: To be honest, I understand this. Do I need to?

peterbe: As in; I don't have the knowledge about these options for `on:`.

docs/publishing.md
```diff
@@ -4,28 +4,49 @@
 MDN staff [deploy the package to the MDN site](contributing.md#updating-compatibility-tables-on-mdn).
 Usually, this happens every Thursday (MDN never deploys to production on Fridays).
 
-Any owner can complete the following steps to publish a new version, but please coordinate releases with [Florian Scholz](https://github.com/Elchi3).
+## Before you begin
 
-To create and publish a new version of `mdn-browser-compat-data`:
+Any project owner (or release designee) can complete the following steps to publish a new version, but please coordinate releases with [@ddbeck](https://github.com/ddbeck).
+
+The steps in this process assume that your `origin` remote poitns to `mdn/browser-compat-data`. You may have to adjust the procedure to use a different remote name.
```
peterbe, line 11:
```suggestion
The steps in this process assume that your `origin` remote points to `mdn/browser-compat-data`. You may have to adjust the procedure to use a different remote name.
```

peterbe (approved):
I'm a bitter scatterbrained to give this a proper review but as a quick check, it looks very sane. (if it works :)

I wish there was something written somewhere about the `NPM_TOKEN` in a post-hit-by-bus era.

.github/workflows/release.yml
ddbeck: It's not essential to know, no.

This triggers the workflow on publishing a new release in https://github.com/mdn/browser-compat-data/releases. Like triggering on a release tag, except I have a chance to finish writing the release notes beforehand.

.github/workflows/release.yml
ddbeck: Nice to hear that it's working fine. I'll switch it to v2. 👍

.github/workflows/node.js.yml
ddbeck: The short version is: I'm going to fix this in a subsequent PR, but I wanted to make the Node versions consistent between workflows now. Though it occurs to me now that the _release_ workflow should never do more than one version. I'll fix that now.

I think the plan here was to have it test multiple versions of Node once we had dropped Travis (which is also testing one more than one Node, I think). I just figured I should actually test with the same version as we specify as the minimum Node version.

ddbeck:
Thank you for the review, @peterbe! I pushed a few changes based on your suggestions:

* > I wish there was something written somewhere about the NPM_TOKEN in a post-hit-by-bus era.
   
   Added a note to the publishing docs with 47b55d6.

* Got rid of superfluous `matrix` settings with 4791888. I'll do more to this problem in a subsequent PR.

* Switched to `actions/node-setup@v2` with 5a7dee6.

> (if it works :)

I plan to test this with a pre-release of `@mdn/browser-compat-data` as soon as it's merged. :-)

ddbeck:
OK, I'm going to merge this so I can do today's release.

PR #6673, opened 2020-09-12
Set IE to false for all ServiceWorker-specific APIs

vinyldarkscratch:
This PR sets Internet Explorer to `false` for all APIs exposed explicitly in service workers, which were not supported in IE.  Results confirmed via the [mdn-bcd-collector](https://github.com/foolip/mdn-bcd-collector) project.

### DIFF
--- a/api/SyncEvent.json
+++ b/api/SyncEvent.json
@@ -20,7 +20,7 @@
             "version_added": null
           },
           "ie": {
-            "version_added": null
+            "version_added": false
           },
           "opera": {
             "version_added": null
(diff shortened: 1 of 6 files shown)


api/SyncEvent.json
```diff
@@ -20,7 +20,7 @@
             "version_added": null
           },
           "ie": {
-            "version_added": null
+            "version_added": false
```
foolip, line 23: Confirmed `[Exposed=ServiceWorker]`: https://wicg.github.io/background-sync/spec/#sync-event

foolip (approved):
Confirmed that all the specs match the claim. But also, I recognize all of these APIs as much more recent than IE11, so...

PR #6675, opened 2020-09-13
mdn-bcd-collector: set null -> false for APIs for IE

vinyldarkscratch:
Using the mdn-bcd-collector project, this PR sets various APIs and API features as `false` for Internet Explorer.  Results were then confirmed for accuracy, by comparing to Edge version numbers, and then if Edge data is inconclusive, checked in the IE console.

Stats:
| browser | real values | ranged values | `true` values | `null` values |
| --- | --- | --- | --- | --- |
| total (before) | 68.26% | 3.17% | 18.89% | 9.67% |
| total (after) | 68.98% | 3.17% | 18.89% | 8.95% |
| ie (before) | 73.35% | 0.00% | 8.04% | 18.60% |
| ie (after) | 79.13% | 0.00% | 8.04% | 12.83% |

### DIFF
--- a/api/DOMConfiguration.json
+++ b/api/DOMConfiguration.json
@@ -20,7 +20,7 @@
             "version_added": false
           },
           "ie": {
-            "version_added": null
+            "version_added": false
           },
           "opera": {
             "version_added": false
--- a/api/Document.json
+++ b/api/Document.json
@@ -1272,7 +1272,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": null
+              "version_added": false
             },
             "opera": {
               "version_added": true
@@ -1932,7 +1932,7 @@
               "version_added": false
             },
             "ie": {
-              "version_added": null
+              "version_added": false
             },
             "opera": {
               "version_added": null
@@ -2029,7 +2029,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": null
+              "version_added": false
             },
             "opera": {
               "version_added": null
@@ -2125,7 +2125,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": null
+              "version_added": false
             },
             "opera": {
               "version_added": null
@@ -3186,7 +3186,7 @@
               "version_added": false
             },
             "ie": {
-              "version_added": null
+              "version_added": false
             },
             "opera": {
               "version_added": null
@@ -3687,7 +3687,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": null
+              "version_added": false
             },
             "opera": {
               "version_added": null
@@ -4311,7 +4311,7 @@
               }
             ],
             "ie": {
-              "version_added": null
+              "version_added": false
             },
             "opera": {
               "version_added": true
@@ -5994,7 +5994,7 @@
               "version_added": false
             },
             "ie": {
-              "version_added": null
+              "version_added": false
             },
             "opera": {
               "version_added": null
@@ -6517,7 +6517,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": null
+              "version_added": false
             },
             "opera": {
               "version_added": null
@@ -6921,7 +6921,7 @@
               "version_added": false
             },
             "ie": {
-              "version_added": null
+              "version_added": false
             },
             "opera": {
               "version_added": null
@@ -6969,7 +6969,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": null
+              "version_added": false
             },
             "opera": {
               "version_added": null
@@ -7017,7 +7017,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": null
+              "version_added": false
             },
             "opera": {
               "version_added": null
@@ -7065,7 +7065,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": null
+              "version_added": false
             },
             "opera": {
               "version_added": true
@@ -7113,7 +7113,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": null
+              "version_added": false
             },
             "opera": {
               "version_added": true
@@ -7415,7 +7415,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": null
+              "version_added": false
             },
             "opera": {
               "version_added": true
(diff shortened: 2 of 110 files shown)


foolip:
I trust the approach here and verifying *everything* manually would not be productive. Instead I'll check these 10 files, chosen at random:
api/Credential.json
api/EXT_blend_minmax.json
api/External.json
api/HTMLElement.json
api/HTMLImageElement.json
api/InputDeviceInfo.json
api/Performance.json
api/PublicKeyCredentialCreationOptions.json
api/SVGFEDropShadowElement.json
api/StorageEstimate.json

api/InputDeviceInfo.json
```diff
@@ -20,7 +20,7 @@
             "version_added": null
           },
           "ie": {
-            "version_added": null
+            "version_added": false
```
foolip, line 23: `InputDeviceInfo` is something returned by `navigator.mediaDevices.enumerateDevices`, which IE11 doesn't support, confirmed by http://mdn-bcd-collector.appspot.com/tests/api/MediaDevices/enumerateDevices, and already set in https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/enumerateDevices#Browser_compatibility

api/Performance.json
```diff
@@ -665,7 +665,7 @@
               "version_added": false
             },
             "ie": {
-              "version_added": null
+              "version_added": false
```
foolip, line 668: Yes, `performance.memory` is a Chrome thing.

api/Performance.json
```diff
@@ -1073,7 +1073,7 @@
               "version_added": "53"
             },
             "ie": {
-              "version_added": null
+              "version_added": false
```
foolip, line 1076: `performance.timeOrigin` confirmed not support in IE11 and Edge 15 using https://software.hixie.ch/utilities/js/live-dom-viewer/saved/8471

api/PublicKeyCredentialCreationOptions.json
```diff
@@ -20,7 +20,7 @@
             "version_added": null
           },
           "ie": {
-            "version_added": null
+            "version_added": false
```
foolip, line 23: This is from https://w3c.github.io/webauthn/, very new stuff.

api/SVGFEDropShadowElement.json
```diff
@@ -20,7 +20,7 @@
             "version_added": true
           },
           "ie": {
-            "version_added": null
+            "version_added": false
```
foolip, line 23: https://developer.mozilla.org/en-US/docs/Web/SVG/Element/feDropShadow is all red for IE, so I won't confirm this separately.

api/StorageEstimate.json
```diff
@@ -20,7 +20,7 @@
             "version_added": "51"
           },
           "ie": {
-            "version_added": null
+            "version_added": false
```
foolip, line 23: This is from https://storage.spec.whatwg.org/, a spec that didn't exist at the time of IE11.

foolip (commented):
I have checked the files listed in https://github.com/mdn/browser-compat-data/pull/6675#issuecomment-692013012, and have doubts only about External. @vinyldarkscratch can you take a look?

api/DOMConfiguration.json
```diff
@@ -20,7 +20,7 @@
             "version_added": false
           },
           "ie": {
-            "version_added": null
+            "version_added": false
```
foolip, line 23: Confirmed by http://mdn-bcd-collector.appspot.com/tests/api/DOMConfiguration.

api/Document.json
```diff
@@ -1932,7 +1932,7 @@
               "version_added": false
             },
             "ie": {
-              "version_added": null
+              "version_added": false
```
foolip, line 1935: `document.createEntityReference` confirmed absent in IE11 by https://software.hixie.ch/utilities/js/live-dom-viewer/saved/8472

api/Element.json
```diff
@@ -5789,7 +5789,7 @@
               "version_added": null
             },
             "ie": {
-              "version_added": null
+              "version_added": false
```
foolip, line 5792: I've filed https://github.com/mdn/browser-compat-data/issues/6683 about api.Element.name not being a useful table. This is correct, though.

api/External.json
```diff
@@ -20,7 +20,7 @@
             "version_added": true
           },
           "ie": {
-            "version_added": null
+            "version_added": false
```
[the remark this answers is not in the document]
vinyldarkscratch: Huh, that _is_ pretty weird...  From what I can tell, in IE11, the property is just a blank object, and isn't actually an object of the API (in other words, something that'll screw up feature detection).

Looking at the spec, an `External` API _should_ be exposed on the Window, which it is in Chrome but not in IE, so I'm leaning towards saying this isn't implemented (in fact, with the way `window.external` is, it's _worse_ than not implemented).

api/Element.json
```diff
@@ -5917,7 +5917,7 @@
               }
             ],
             "ie": {
-              "version_added": null
+              "version_added": false
```
foolip, line 5920: I thought this (api.Element. onfullscreenchange) might be wrong, but https://software.hixie.ch/utilities/js/live-dom-viewer/saved/8475 shows that actually only `document.onmsfullscreenchange` was in IE11, not `element.onmsfullscreenchange`. But I found and fixed another problem: https://github.com/mdn/browser-compat-data/pull/6684

api/Window.json
```diff
@@ -3755,7 +3755,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": null
+              "version_added": false
```
foolip, line 3758: This looked a bit implausible, but seems to be true: https://software.hixie.ch/utilities/js/live-dom-viewer/saved/8476

foolip (commented):
I reviewed some more bits that looked old to me, where it's most plausible that IE11 would actually support it.

api/External.json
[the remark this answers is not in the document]
foolip: It might be that this API was exposed in some contexts (extensions or some special kinds of scripts?) and not others, and this weird binding was left exposed to web content as well. Let's go with false.

foolip (approved):
Alright, there are no outstanding comments! It's somewhat likely that in all of these changes there is *some* error, but I'm confident it's at least 95% correct based on my spot checking, and certainly an improvement overall.

PR #6676, opened 2020-09-13
Firefox Android supports String.replaceAll since v79

atjn:
I have manually tested this with the [demo on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll).


ddbeck (approved):
Thank you, @atjn!

PR #6677, opened 2020-09-13
CSS ::marker animations and transitions not supported in WebKit

Loirooriol:
WebKit implemented and earlier version of the spec which didn't include animations and transitions.
See https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/style/PropertyCascade.cpp#L73


sideshowbarker (approved):

sideshowbarker:
Thanks!

Issue #6509, opened 2020-08-10
Choose a guideline for descendant flag data


ddbeck:
Prompted by #6465 and https://github.com/mdn/browser-compat-data/pull/6433#issuecomment-669900198, we do not have a consistent way of considering flag data for features which are descendants of a feature also behind a flag.

Or, more visually, consider a structure like this:

* `SomeObject`
   * `SomeObject()`
   * `someMethod()`

If `SomeObject` is behind a flag, should `SomeObject()` and `someMethod()` duplicate the flag data of `SomeObject`? If the data should be duplicated, by what mechanism?

ddbeck:
Some reviewers, such as myself, have been supposing that a feature behind a flag implicitly puts every descendant feature behind the flag. I described my thinking on this in more detail in this comment https://github.com/mdn/browser-compat-data/pull/6465#issuecomment-668656697. I concede this might be a source of confusion in some consumer use-cases (e.g., an MDN page for some subfeature may not show that the parent feature is gated by a flag because that's on another page), but it great simplifies maintenance. When a feature is no longer behind a flag, only the parent feature data needs to change.

Other reviewers have been doing something else, but I can't say what their thinking is.

Possible solutions that have occurred to me:

1. **Always inherit**: Set a guideline requiring that features **must** duplicate any flag data from its ancestor features (i.e., every feature must independently record any flags required to access that feature even if those features a predicate to the feature in question). Plan work to: a) add linting to catch inconsistently flagged data, b) add migration scripts to fix inconsistently flagged data and c) add a fix script to cascade new flag data.

2. **Never inherit**: Set a guideline requiring that features **must not** duplicate flag data from its ancestor features (i.e., a feature must only report the flags which apply to itself and its descendants, but not its ancestors). Plan work to a) add linting to catch inconsistently flagged data and b) add migration scripts to fix inconsistently flagged data.

3. **Dynamically inherit**: Set a guideline requiring that features **must not** duplicate flag data from its ancestor features, but dynamically inline flags into descendant features (i.e., when ingesting JSON files, cascade flags to subfeatures automatically). Plan work to a) add linting to catch inconsistently flagged data, b) add migration scripts to fix inconsistently flagged data, and c) add code to dynamically inline flags into descendant features.

   There could be variations on this approach, too, if determining whether a feature has inherited a flag becomes too complex or over-inflates the data (e.g., a sort of meta flag that says, more or less, "parent feature may X require a flag").

[...]

PR #6679, opened 2020-09-13
Update data for Payment APIs

vinyldarkscratch:
When pulling data from the `mdn-bcd-collector` project I've been working on, I noticed a few odd discrepancies with data reported by the collector and BCD's data.  It turns out that support for the Payment APIs is in more browsers than we initially thought.  This PR does the following to fix these discrepancies, plus adjust our data in a desirable way:

1) Updates the data to reflect what was found based upon testing with the collector and manual testing (Opera, Opera Android, and Samsung Internet all support the APIs, apparently -- WebView has partial API support, though this is probably not intentional)
2) Removes repeated flag information for the API's subfeatures (we don't need to mention the flag for PaymentResponse.complete, for example, when PaymentResponse already has it)

The data for these interfaces was, to put it simply, quite the mess.  This PR hopefully resolves the mess.

### DIFF
--- a/api/AbortPaymentEvent.json
+++ b/api/AbortPaymentEvent.json
@@ -4,62 +4,77 @@
       "__compat": {
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/AbortPaymentEvent",
         "support": {
-          "chrome": {
-            "version_added": "61",
-            "flags": [
-              {
-                "type": "preference",
-                "name": "#service-worker-payment-apps",
-                "value_to_set": "Enabled"
-              }
-            ]
-          },
-          "chrome_android": {
-            "version_added": "61",
-            "flags": [
-              {
-                "type": "preference",
-                "name": "#service-worker-payment-apps",
-                "value_to_set": "Enabled"
-              }
-            ]
-          },
-          "edge": {
-            "version_added": "≤79",
-            "flags": [
-              {
-                "type": "preference",
-                "name": "#service-worker-payment-apps",
-                "value_to_set": "Enabled"
-              }
-            ]
-          },
+          "chrome": [
+            {
+              "version_added": true
+            },
+            {
+              "version_added": "61",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "#service-worker-payment-apps",
+                  "value_to_set": "Enabled"
+                }
+              ]
+            }
+          ],
+          "chrome_android": [
+            {
+              "version_added": true
+            },
+            {
+              "version_added": "61",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "#service-worker-payment-apps",
+                  "value_to_set": "Enabled"
+                }
+              ]
+            }
+          ],
+          "edge": [
+            {
+              "version_added": true
+            },
+            {
+              "version_added": "79",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "#service-worker-payment-apps",
+                  "value_to_set": "Enabled"
+                }
+              ]
+            }
+          ],
           "firefox": {
-            "version_added": null
+            "version_added": false
           },
           "firefox_android": {
-            "version_added": null
+            "version_added": false
           },
           "ie": {
             "version_added": false
           },
           "opera": {
-            "version_added": false
+            "version_added": true
           },
           "opera_android": {
-            "version_added": false
+            "version_added": true
           },
           "safari": {
-            "version_added": null
+            "version_added": false
           },
           "safari_ios": {
-            "version_added": null
+            "version_added": false
           },
           "samsunginternet_android": {
-            "version_added": false
+            "version_added": true
           },
           "webview_android": {
-            "version_added": false
+            "version_added": true
           }
         },
         "status": {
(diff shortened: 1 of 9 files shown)


foolip:
> Removes repeated flag information for the API's subfeatures (we don't need to mention the flag for PaymentResponse.complete, for example, when PaymentResponse already has it)

This bit I'm not sure about. It seems to me like it will make the data incorrect, I'll comment inline.

api/PaymentResponse.json
```diff
@@ -94,22 +94,9 @@
...
-                  {
-                    "type": "preference",
-                    "name": "#web-payments",
-                    "value_to_set": "Enabled"
-                  }
-                ]
-              }
-            ],
+            "chrome_android": {
+              "version_added": "53"
```
foolip, line 98: The data now says that PaymentResponse was shipped in Chrome for Android 56, but this subfeature was shipped earlier. That should fail the consistency lint I think, but somehow doesn't. I suspect it's because that lint doesn't deal with flags correctly. IMHO, what it should do is two passes. One that considers only behind-a-flag versions and one that considers only shipping-by-default versions. That should then have failed here.

vinyldarkscratch: Hmmm...  To be honest, I don't think that this should fail the consistency check, nor should the flag information be repeated for the subfeatures, separating the subfeatures into multiple support statements just to satisfy the rendered tables.  I think this is the same problem as if an API is supported behind a prefix or with an alt. name, when the subfeatures were implemented during those prefixed versions.  (I remember we were dealing with a similar situation in regards to the Geometry APIs.)

vinyldarkscratch: It looks like there's an open issue about this, see #6509.

vinyldarkscratch: I just realized that #6670 (irrelevant flag removal) also applies, since the flags have been removed 2+ years ago.  Perhaps we should just remove the flag data for Chrome entirely?

foolip: Thanks for pointing out https://github.com/mdn/browser-compat-data/issues/6509, I didn't know there were different interpretations of flags and had just assumed everyone shared my own interpretation. Typical 👍 

Once https://github.com/mdn/browser-compat-data/pull/6670 has been merged, I'd love if we can resolve this by just removing the flag data.

For now, can we have the subfeatures represent when things were shipped by default? Even if flag data is taken to be inherited, there would then be no conflict between the parent and child features. (The problematic cases will be large features still entirely behind a flag.)

vinyldarkscratch: #6670 is merged now, so we can remove the flag data!

foolip: Oh happy day!

alshamiri5 (approved):

api/AbortPaymentEvent.json
```diff
@@ -5,61 +5,40 @@
...
           "chrome": {
-            "version_added": "61",
-            "flags": [
-              {
-                "type": "preference",
-                "name": "#service-worker-payment-apps",
-                "value_to_set": "Enabled"
-              }
-            ]
+            "version_added": "70"
```
foolip, line 8: Corroborated by https://storage.googleapis.com/chromium-find-releases-static/12a.html#12a4e4ffbbde4ffd73eeb7b475e5110578b79ca9

api/PaymentAddress.json
```diff
@@ -5,27 +5,13 @@
...
           "chrome": {
-            "version_added": "61",
-            "flags": [
-              {
-                "type": "preference",
-                "name": "#web-payments",
-                "value_to_set": "Enabled"
-              }
-            ]
+            "version_added": "60"
```
foolip, line 8: Confirmed by https://mdn-bcd-collector.appspot.com/tests/api/PaymentAddress. Finding the commit for this is a bit tricky, but https://storage.googleapis.com/chromium-find-releases-static/259.html#259bce9e315438c00d7d660b446de65b6b03c360 happened soon after it was enabled and was M60, so it checks out.

api/PaymentAddress.json
```diff
@@ -92,67 +78,37 @@
...
-              ]
+              "version_added": "56"
             },
             "edge": {
               "version_added": "15"
             },
             "firefox": {
-              "version_added": "56",
-              "notes": "Available only in nightly builds. Requires the comma-delineated list in <code>dom.payments.request.supportedRegions</code> to contain one or more of the supported 2-character ISO locales, currently US and CA.",
-              "flags": [
```
foolip, line 120: This flag removal I'm afraid isn't supported by the new guideline. Is this flag still there?

foolip (commented):
Only reviewed part of it, have to run!

api/PaymentAddress.json
vinyldarkscratch: Yeah, Firefox still requires a flag for the payment APIs.

foolip: Good, so the data was correct. Can you revert any flag removals from this PR that are not supported by https://github.com/mdn/browser-compat-data/blob/master/docs/data-guidelines.md#removal-of-irrelevant-flag-data?

vinyldarkscratch: Looking into this a little bit more, I did some research and confirmed the note about how this is only in Firefox Nightly builds.  No release build has support (even though the flags are available, they don't have any effect).  Would it perhaps make more sense to set Firefox to `false`?

foolip: Oh, that's weird, I didn't know Firefox had compile-time features, or how this can happen. @ddbeck have you come across this situation before, and how should it be represented in BCD?

ddbeck: We used to have a `compile_flag` but I removed it in #3752. #338 has some discussion about the need for a gated, non-release version option. Since we don't have anything like that yet, I'd suggest setting version_added to `false`, but keeping the note. (Better still would be to link from the note to the relevant bug so that, someday, when we have a solution to #338, updating it is easier.)

vinyldarkscratch: Fixed!

api/AbortPaymentEvent.json
```diff
@@ -5,61 +5,40 @@
...
-            "version_added": null
+            "version_added": false
           },
           "samsunginternet_android": {
-            "version_added": false
+            "version_added": true
           },
           "webview_android": {
-            "version_added": false
+            "version_added": "70"
```
foolip, line 41: Are you sure about this and the other webview_android changes? It seems a bit implausible that the event interfaces should be supported when the payment APIs that would cause the events to be fired are not.

foolip (commented):
Aaaalmost ready to merge.

api/AbortPaymentEvent.json
vinyldarkscratch: As surprising as it is, some interfaces were exposed in WebView for the Payment APIs.  I have no doubt that this was a mistake and they weren't supposed to be exposed, but I wanted to go for accuracy.  I can keep it all `false` if preferred.

foolip: That sort of mistake happens from time to time, unfortunately. Since it will look like a mistake to anyone looking at the data or MDN pages, can you leave it as false in this PR? We can make the changes, but then add notes as well. Doing that in a separate PR would be best I think.

foolip (approved):

PR #6685, opened 2020-09-14
datalist doesn't work starting Firefox for Android v67

hirusi:
* up to v80 stable as of this proposal
* also affects v81 beta (but left out of proposal as we should track only stable releases)
* bug is being tracked here - https://bugzilla.mozilla.org/show_bug.cgi?id=1535985

I have personally tested this on Firefox for Android v80 Stable and Firefox for Android v81 Beta.

- [x] Summarize your changes

### DIFF
--- a/html/elements/datalist.json
+++ b/html/elements/datalist.json
@@ -19,7 +19,7 @@
             },
             "firefox_android": {
               "version_added": "4",
-              "notes": "The dropdown menu containing available options does not appear in Firefox for Android starting version 67 - see <a href='https://bugzil.la/1535985'>bug 1535985</a>."
+              "notes": "Since Firefox for Android 79, the dropdown menu containing available options does not appear. See <a href='https://bugzil.la/1535985'>bug 1535985</a>."
             },
             "ie": {
               "version_added": "10"
--- a/html/elements/datalist.json
+++ b/html/elements/datalist.json
@@ -19,7 +19,7 @@
             },
             "firefox_android": {
               "version_added": "4",
-              "notes": "The dropdown menu containing available options does not appear in Firefox for Android starting version 67 - see <a href='https://bugzilla.mozilla.org/show_bug.cgi?id=1535985'>bug 1535985</a>."
+              "notes": "The dropdown menu containing available options does not appear in Firefox for Android starting version 67 - see <a href='https://bugzil.la/1535985'>bug 1535985</a>."
             },
             "ie": {
               "version_added": "10"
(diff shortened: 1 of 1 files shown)


html/elements/datalist.json
```diff
@@ -18,7 +18,8 @@
               "version_added": "4"
             },
             "firefox_android": {
-              "version_added": "4"
+              "version_added": "4",
+              "notes": "The dropdown menu containing available options does not appear in Firefox for Android starting version 67 - see <a href='https://bugzil.la/1535985'>bug 1535985</a>."
```
vinyldarkscratch, line 22:
```suggestion
              "notes": "Since Firefox for Android 79, the dropdown menu containing available options does not appear. See <a href='https://bugzil.la/1535985'>bug 1535985</a>."
```

vinyldarkscratch (changes requested):
Hey there, thanks for your PR and welcome to BCD!  I had noticed the same thing as well in FxA 80, so I can confirm this is true.  However, it looks like the problem is a little more recent -- rolling back to FxA 67 and 68, I see the dropdown menu, but not in the new Fenix edition (FxA 79+).

I've just got a little rewording and version number change suggestion, but otherwise looks good to me!

html/elements/datalist.json
hirusi: I have no way to test all the way back to those versions and instead, relied on the bug report to state that the problem started in v67. However, if you have tested it and are confident, that's great and I'm happy to commit this. In a second! 😊

hirusi:
I have no way to test all the way back to those specific versions and instead relied on the bug report. If you have tested them, that's good enough and I have committed your suggestions now. 😊

vinyldarkscratch (approved):
Thanks for the quick fixes, looking good to me!  👍   I'm not sure why the bug talks about those versions, since they're showing up on my end...ah well.  😛 

By the way, I use a third-party app store called Uptodown to downgrade my apps, like Firefox or Chrome, to earlier versions -- perhaps something to look into!  (I'd recommend using an emulator or a separate device when downgrading apps though, since it _does_ involve reinstalling the app every time.)

hirusi:
> By the way, I use a third-party app store called Uptodown to downgrade my apps, like Firefox or Chrome, to earlier versions -- perhaps something to look into! (I'd recommend using an emulator or a separate device when downgrading apps though, since it _does_ involve reinstalling the app every time.)

I don't have the luxury of a second device but I'll keep this in mind. 😊

> Thanks for the quick fixes, looking good to me! 👍

Thank you too for a very responsive resolution to this! 🎉

PR #6689, opened 2020-09-14
Traverse script: allow "any" argument

vinyldarkscratch:
This PR updates the traverse script to allow the word `any` as an argument for the `value` parameter.  Useful to list all the features BCD has.

### DIFF
--- a/docs/testing.md
+++ b/docs/testing.md
@@ -23,6 +23,6 @@ The browser may be any single browser defined in the [`browsers/` folder](https:
 
 The folder may be omitted or set to `all` to search through all data folders, or a comma-separated list of folders to search through.
 
-The value may be omitted to search for all non-real values (or more specifically, `true` and `null` values), or any value accepted by `version_added` and `version_removed`.
+The value may be omitted to search for all non-real values (or more specifically, `true` and `null` values), or any value accepted by `version_added` and `version_removed`. It can also be set to `any` to allow for any value (to simply list all the entries in BCD).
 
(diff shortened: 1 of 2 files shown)


ddbeck:
Hey @vinyldarkscratch, thank you for opening this PR. It got me thinking about how to improve the legibility of this script.

I'm concerned that this is taking us farther down the path of cryptic command-line interfaces. I propose we step back a little here to rework the CLI to use more traditional flags and options, instead of magic arguments. Here's what I have in mind:

* `npm run traverse` (alone, no arguments or options) ought to emit all the features, by default
* `-b/--browser` would set the browser to filter against (e.g., `-b safari` or `--browser=safari` — in a follow-up, we could even permit multiple browsers)
* `-f/--filter` (allowed multiple times) would populate the array of values to filter on (e.g., `-f ≤37` or `--filter=null`)
* `--non-real` would be a convenience flag for `-f true -f null`
* Positional arguments for would set the folders to search

Thus `npm run traverse webview_android api,javascript true` would become `npm run traverse -- --browser=webview_android --filter=true api/ javascript/`. This would get us better tab completion, less cryptic usage instructions, and emitting every feature identifier is a piece of cake.

I realize that's a slightly bigger PR, but since I had a hard time working out what this PR did on first glance, I'm reluctant to add more complexity to the current `traverse` CLI.

vinyldarkscratch:
I implemented the suggestions in #6821!  As such, I will close this PR, since this is superseded by the new one.

PR #6690, opened 2020-09-15
Update data for Acceleration/Orientation Sensor APIs

vinyldarkscratch:
This PR updates the version data for the six acceleration/orientation sensor APIs with the help of the mdn-bcd-collector project.  The collector reported that these APIs were introduced in Chrome 67 rather than Chrome 69, and I've confirmed this to be true, so this PR updates the Chromium data accordingly.  It also sets the other browsers that were null to the appropriate values (which was `false` for all of them).

Tests used:
https://mdn-bcd-collector.appspot.com/tests/api/AbsoluteOrientationSensor
https://mdn-bcd-collector.appspot.com/tests/api/Accelerometer
https://mdn-bcd-collector.appspot.com/tests/api/Gyroscope
https://mdn-bcd-collector.appspot.com/tests/api/LinearAccelerationSensor
https://mdn-bcd-collector.appspot.com/tests/api/OrientationSensor
https://mdn-bcd-collector.appspot.com/tests/api/RelativeOrientationSensor

### DIFF
--- a/api/AbsoluteOrientationSensor.json
+++ b/api/AbsoluteOrientationSensor.json
@@ -5,22 +5,25 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/AbsoluteOrientationSensor",
         "support": {
           "chrome": {
-            "version_added": "69"
+            "version_added": "67"
           },
           "chrome_android": {
-            "version_added": "69"
+            "version_added": "67"
           },
           "edge": {
-            "version_added": "≤79"
+            "version_added": "79"
           },
           "firefox": {
+            "version_added": false
+          },
+          "firefox_android": {
             "version_added": false
           },
           "ie": {
             "version_added": false
           },
           "opera": {
-            "version_added": "56"
+            "version_added": "54"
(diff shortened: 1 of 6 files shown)


jpmedley (approved):

api/AbsoluteOrientationSensor.json
```diff
@@ -5,34 +5,40 @@
...
+          },
           "ie": {
             "version_added": false
           },
           "opera": {
-            "version_added": "56"
+            "version_added": "54"
           },
           "opera_android": {
             "version_added": "48"
```
foolip, line 29: As noted in https://github.com/mdn/browser-compat-data/pull/6695, opera_android here is correct even though it can *look* like an oversight in the change.

foolip (approved):
Supported by https://storage.googleapis.com/chromium-find-releases-static/b90.html#b90312615c24b59834bfbb2f648bf1e3e9187065, and goes together with https://github.com/mdn/browser-compat-data/pull/6695.

PR #6691, opened 2020-09-15
DOMTokenList: separate whitespace trim/duplicate removal; add versions

vinyldarkscratch:
In `DOMTokenList`, we have an entry about how modifications will trim the whitespace and remove duplicates, with partial implementation notes for IE and Edge about how they trim whitespace but not remove duplicates.  However, from my testing, I've found that Chrome and Firefox also implemented whitespace trimming before duplicate removal.  This PR separates the subfeature into two subfeatures, and adds/fixes versions for both of them.

### DIFF
--- a/api/DOMTokenList.json
+++ b/api/DOMTokenList.json
@@ -831,10 +831,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "55"
+              "version_added": "51"
             },
             "firefox_android": {
-              "version_added": "55"
+              "version_added": "51"
             },
             "ie": {
               "version_added": "10"
--- a/api/DOMTokenList.json
+++ b/api/DOMTokenList.json
@@ -529,7 +529,7 @@
       },
       "remove_duplicates": {
         "__compat": {
-          "description": "Removal of duplicates",
+          "description": "Removes duplicates",
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/DOMTokenList#Trimming_of_whitespace_and_removal_of_duplicates",
           "support": {
             "chrome": {
--- a/api/DOMTokenList.json
+++ b/api/DOMTokenList.json
@@ -527,21 +527,19 @@
           }
         }
       },
-      "remove_whitespace_and_duplicates": {
+      "remove_duplicates": {
         "__compat": {
-          "description": "Trimming of whitespace and removal of duplicates",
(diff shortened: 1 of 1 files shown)


foolip:
This will remove an entry, so adding the needs-release-note label.

foolip:
@vinyldarkscratch can you link to the tests you used for this?

vinyldarkscratch:
Certainly!  I used the following pieces of JavaScript to test manually in BrowserStack:


`trim_whitespace`:
```js
var elm = document.createElement('b'); elm.className = ' foo bar foo '; elm.classList.remove('bar'); return elm.className === 'foo foo' || elm.className === 'foo';
```

`remove_duplicates`:
```js
var elm = document.createElement('b'); elm.className = ' foo bar foo '; elm.classList.remove('bar'); return elm.className === 'foo';
```

foolip:
@vinyldarkscratch the `remove_duplicates` seems to assume that if duplicates are removes then whitespace is also trimmed. That's probably true, but did you check what the `elm.className` was in the latest build where it *wasn't* 'foo'? A value like ' foo  foo ' is not impossible to imagine in some implementation where these things were done in the other order.

vinyldarkscratch:
I checked all the browser versions, and have confirmed that white space trimming came first or at the same time in all the browsers, and that duplicate removal had come later.

api/DOMTokenList.json
```diff
@@ -527,21 +527,19 @@
           }
         }
       },
-      "remove_whitespace_and_duplicates": {
+      "remove_duplicates": {
         "__compat": {
-          "description": "Trimming of whitespace and removal of duplicates",
+          "description": "Removal of duplicates",
```
ddbeck, line 532:
```suggestion
          "description": "Removes duplicates",
```

api/DOMTokenList.json
```diff
@@ -820,6 +816,55 @@
           }
         }
       },
+      "trim_whitespace": {
+        "__compat": {
+          "description": "Trimming of whitespace",
```
ddbeck, line 821:
```suggestion
          "description": "Trims whitespace",
```

ddbeck (changes requested):
I'll leave the review of the data to Philip, but I had a couple nitpicky style suggestions. Thank you!

foolip (approved):

PR #6695, opened 2020-09-15
Update data for Sensor and SensorErrorEvent APIs

vinyldarkscratch:
This PR updates the version data for the Sensor and SensorErrorEvent APIs with the help of the mdn-bcd-collector project.  The collector reported that these APIs were introduced in Chrome 67 rather than Chrome 69, and I've confirmed this to be true, so this PR updates the Chromium data accordingly.  It also sets the other browsers that were null to the appropriate values (which was `false` for all of them).

Tests used:
https://mdn-bcd-collector.appspot.com/tests/api/Sensor
https://mdn-bcd-collector.appspot.com/tests/api/SensorErrorEvent

### DIFF
--- a/api/Sensor.json
+++ b/api/Sensor.json
@@ -5,40 +5,40 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/Sensor",
         "support": {
           "chrome": {
-            "version_added": "69"
+            "version_added": "67"
           },
           "chrome_android": {
-            "version_added": "69"
+            "version_added": "67"
           },
           "edge": {
-            "version_added": "≤79"
+            "version_added": "79"
           },
           "firefox": {
-            "version_added": null
+            "version_added": false
           },
           "firefox_android": {
-            "version_added": null
+            "version_added": false
           },
           "ie": {
             "version_added": false
           },
           "opera": {
-            "version_added": "56"
+            "version_added": "54"
           },
           "opera_android": {
             "version_added": "48"
           },
           "safari": {
-            "version_added": null
+            "version_added": false
           },
           "safari_ios": {
-            "version_added": null
+            "version_added": false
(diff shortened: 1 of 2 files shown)


api/Sensor.json
```diff
@@ -5,40 +5,40 @@
...
           },
           "ie": {
             "version_added": false
           },
           "opera": {
-            "version_added": "56"
+            "version_added": "54"
           },
           "opera_android": {
             "version_added": "48"
```
foolip, line 29: Alright, so Opera for Android 48 is still correct, because release 47 was using Chromium 66.

foolip (commented):
https://storage.googleapis.com/chromium-find-releases-static/b90.html#b90312615c24b59834bfbb2f648bf1e3e9187065 is further evidence that this is correct.

That change also shipped a bunch of other APIs, where the data is also wrong:
https://developer.mozilla.org/en-US/docs/Web/API/AbsoluteOrientationSensor
https://developer.mozilla.org/en-US/docs/Web/API/Accelerometer
https://developer.mozilla.org/en-US/docs/Web/API/Gyroscope
https://developer.mozilla.org/en-US/docs/Web/API/LinearAccelerationSensor
https://developer.mozilla.org/en-US/docs/Web/API/OrientationSensor
https://developer.mozilla.org/en-US/docs/Web/API/RelativeOrientationSensor

@vinyldarkscratch since the changes to those would be incredibly similar, would you mind putting them all in this PR? That will make it easier to follow what happened in the resulting history.

foolip:
Oh wait, those changes are in https://github.com/mdn/browser-compat-data/pull/6690. I'll just merge both at the same time, that'll be clear enough to later archeologists.

foolip (approved):

PR #6668, opened 2020-09-10
Update <iframe sandbox=allow-downloads> for Firefox and Chrome

evilpie:
I manually tested in Firefox 82 and Chrome 85, but used the following
references for 81 and 83 respectively.

https://bugzilla.mozilla.org/show_bug.cgi?id=1558394
https://www.chromestatus.com/feature/5706745674465280

Fixes #6667

PR #6696, opened 2020-09-15
Add Firefox for Android 81

ddbeck:
It's [in progress](https://github.com/mozilla-mobile/fenix/releases/tag/v81.1.1-beta.5), but it's not obvious to when the Android releases drop.

This is primarily to unblock #6668.


vinyldarkscratch (approved):
👍

PR #6699, opened 2020-09-15
Fix AmbientLightSensor+Magnetometer API data

vinyldarkscratch:
This PR updates all of the data for the `AmbientLightSensor` API.  When running results from the mdn-bcd-collector project, I noticed that it was reporting that the API was not supported in Chrome.  Doing some digging around, it looks like the data was added very, _very_ incorrectly.

First off, the `AmbientLightSensor` API had always been based upon the Generic Sensor API since it was first implemented.  Second, neither API was exposed by default -- the #enable-generic-sensor flag needed to be enabled (and then later, the #enable-generic-sensor-extra-classes as well).

This data was determined by testing for `'AmbientLightSensor' in self;` and `'Magnetometer' in self;` in BrowserStack in various Chrome versions with different flags.

### DIFF
--- a/api/AmbientLightSensor.json
+++ b/api/AmbientLightSensor.json
@@ -160,7 +160,7 @@
               "notes": "In Opera 66, this method stopped returning floats and returned integers to avoid fingerprinting."
             },
             "opera_android": {
-              "version_added": "48",
+              "version_added": "43",
               "notes": "In Opera for Android 57, this method stopped returning floats and returned integers to avoid fingerprinting."
             },
             "safari": {
--- a/api/AmbientLightSensor.json
+++ b/api/AmbientLightSensor.json
@@ -6,24 +6,74 @@
         "support": {
           "chrome": [
             {
-              "version_added": "54"
+              "version_added": "62",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "#enable-generic-sensor",
+                  "value_to_set": "Enabled"
+                },
+                {
+                  "type": "preference",
+                  "name": "#enable-generic-sensor-extra-classes",
+                  "value_to_set": "Enabled"
+                }
+              ]
             },
             {
-              "version_added": "69",
-              "notes": "Based on Generic Sensor API."
+              "version_added": "56",
+              "version_removed": "62",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "#enable-generic-sensor",
+                  "value_to_set": "Enabled"
+                }
+              ]
             }
           ],
           "chrome_android": [
             {
-              "version_added": "54"
+              "version_added": "62",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "#enable-generic-sensor",
+                  "value_to_set": "Enabled"
+                },
+                {
+                  "type": "preference",
+                  "name": "#enable-generic-sensor-extra-classes",
+                  "value_to_set": "Enabled"
+                }
+              ]
             },
             {
-              "version_added": "69",
-              "notes": "Based on Generic Sensor API."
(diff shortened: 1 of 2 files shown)


api/AmbientLightSensor.json
```diff
@@ -6,24 +6,74 @@
         "support": {
           "chrome": [
             {
-              "version_added": "54"
+              "version_added": "62",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "#enable-generic-sensor",
```
foolip, line 13: This doesn't seem to exist under chrome://flags/ any longer, it was probably removed when generic sensors shipped in 67.

Looking at https://chromium.googlesource.com/chromium/src/+/73373a6ed02e4a2bbd3e42b942cf4c144e79992c%5E%21/#F13 and the current state of [runtime_enabled_features.json5](https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/platform/runtime_enabled_features.json5;l=1727;drc=13771712b0eca90b550ee3e9b187b29b83055eca), these bits could always also be enabled by the usual "Experimental Web Platform Features" flag.

Can you use that instead, and collapse these? It doesn't seem worth anyones time to track the more fine-grained flags, and they're more prone to change over time, whereas "Experimental Web Platform Features" has been with us for at least 5 years now.

api/AmbientLightSensor.json
```diff
@@ -34,36 +84,74 @@
...
-          ]
+          ],
+          "safari": {
+            "version_added": false
+          },
+          "safari_ios": {
+            "version_added": false
+          },
+          "samsunginternet_android": {
+            "version_added": false
```
foolip, line 150: I tested https://staging-dot-mdn-bcd-collector.appspot.com/tests/api/AmbientLightSensor in Samsung Internet to confirm. This seems like the kind of API that some vendor might have chosen to experiment with and ship earlier than Chrome, but not this time.

api/AmbientLightSensor.json
```diff
@@ -76,26 +164,14 @@
...
-              {
-                "version_added": "54"
-              },
-              {
-                "version_added": "69",
-                "notes": "Based on Generic Sensor API."
-              }
-            ],
+            "chrome": {
+              "version_added": "56"
```
foolip, line 168: Unfortunately this will also run into https://github.com/mdn/browser-compat-data/issues/6509. 😢

foolip (commented):
Seems blocked on https://github.com/mdn/browser-compat-data/issues/6509, or can we route around that issue somehow?

vinyldarkscratch:
Since the older flag data is within the scope of old flag data removal guidelines, I'll strip that out so we can simplify our data!

api/AmbientLightSensor.json
vinyldarkscratch: Looks like that flag wasn't removed in Chrome 67, but it was enabled by default, making it redundant to enable manually -- only the #enable-generic-sensor-extra-classes needed to be enabled.  I've double-checked the Experimental Web Platform Features flag and can confirm it enables support since Chrome 56, so I'll swap it out for that flag!

alshamiri5 (approved):

api/AmbientLightSensor.json
vinyldarkscratch: I think this is fine to merge as it is before coming to a decision in #6509?  This PR doesn't change the status of flag data inclusion in the subfeatures, which we'd need to check in every file with flags anyways.

foolip: Well, I guess so. This is a case that will be caught by any migration script, as the parent feature has flags but this does not.

foolip (approved):

PR #6706, opened 2020-09-16
Drop the "light-level" Media Queries at-rule

sideshowbarker:
This change completely drops the `light-level` Media Queries at-rule from BCD. It was never implemented in any browser engine, and has now been removed from the Media Queries spec:

* https://drafts.csswg.org/mediaqueries-5/#changes-since-2020-07-15
  > Remove the `light-level` media feature as it is redundant with `prefers-contrast` and `prefers-color-scheme`; add examples of how these media features may be automatically inferred by the user agent based on the same factors `light-level` was expected to respond to.
* https://github.com/w3c/csswg-drafts/commit/f5b663c
* https://github.com/w3c/csswg-drafts/issues/5359

I also filed https://github.com/mdn/sprints/issues/3695 to delete the https://developer.mozilla.org/en-US/docs/Web/CSS/@media/light-level MDN article.

### DIFF
--- a/css/at-rules/media.json
+++ b/css/at-rules/media.json
@@ -854,55 +854,6 @@
             }
           }
         },
-        "light-level": {
-          "__compat": {
-            "description": "<code>light-level</code> media feature",
-            "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/@media/light-level",
-            "support": {
-              "chrome": {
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Looks good. Thank you! 🎉

PR #6372, opened 2020-07-03
Add CSSTransition API data

vinyldarkscratch:
This PR adds data for the CSSTransition API based upon manual testing.  This API is a part of a [W3C draft](https://drafts.csswg.org/css-transitions-2/#the-CSSTransition-interface).

PR #6707, opened 2020-09-16
Remove constructor from CSSTransition

vinyldarkscratch:
This PR removes the constructor data for the `CSSTransition` API.  As per the spec (https://drafts.csswg.org/css-transitions-2/#the-CSSTransition-interface), this interface should not have a constructor.  It looks like I had added it in on https://github.com/mdn/browser-compat-data/pull/6372, but I don't know how I determined that Chrome supported the constructor...  (No browser has constructor support.)

### DIFF
--- a/api/CSSTransition.json
+++ b/api/CSSTransition.json
@@ -46,65 +46,17 @@
           "deprecated": false
         }
       },
-      "CSSTransition": {
-        "__compat": {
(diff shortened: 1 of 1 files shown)


jpmedley (approved):

foolip:
Since this removes an entry the release note label is needed. Added.

PR #6713, opened 2020-09-17
Implement 2.0.0 breaking changes

ddbeck:
This is one of several ugly steps to carry out #6640.

This PR does two key things:

* It proposes merging to a new branch (which I've created and I'm committing to maintaining until the old package name is fully deprecated) `master-scoped-package`.
* It implements the breaking changes proposed in #6640.
* It updates the docs and various mentions `mdn-browser-compat-data`

An additional ugly thing I'm going to do is a pre-release from _this_ branch (i.e., `ddbeck-2.0-breaking-changes`) before merging to demonstrate that, well, it works as a new package. This will also test the new packaging workflow before I release BCD 1.0.38 today.

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -1,6 +1,6 @@
 # Contributing to browser-compat-data
 
-We're really happy to accept contributions to the mdn-browser-compat-data repository!
+We're really happy to accept contributions to the mdn/browser-compat-data repository!
 
 ## Table of contents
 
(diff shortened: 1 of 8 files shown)


ddbeck:
OK, we now have a pre-release of [`@mdn/browser-compat-data`](https://www.npmjs.com/package/@mdn/browser-compat-data) so you can actually try this out. In the process, I got the new publishing workflow working and I can do the next `mdn-` release.

package.json
```diff
@@ -1,14 +1,14 @@
 {
-  "name": "mdn-browser-compat-data",
-  "version": "1.0.37",
+  "name": "@mdn/browser-compat-data",
+  "version": "2.0.0-1",
```
chrisdavidmills, line 3: Is the dash correct here?

docs/contributing.md
```diff
@@ -1,6 +1,6 @@
 # Contributing to browser-compat-data
 
-We're really happy to accept contributions to the mdn-browser-compat-data repository!
+We're really happy to accept contributions to the mdn/browser-compat-data repository!
```
chrisdavidmills, line 3: Does this need an @?

chrisdavidmills (changes requested):
Couple of comments

package.json
ddbeck: This marks the release as a semver pre-release, so yes. I could've used something like `2.0.0-alpha1` but I was lazy and used the default generated by `npm version prerelease`. If I do any more prereleases, I'll do that, but this ought to be a one-time thing.

docs/contributing.md
ddbeck: `mdn/browser-compat-data` is still the name of the repo. That part is _not_ changing.

package.json
chrisdavidmills: OK, cool.

docs/contributing.md
chrisdavidmills: We're all good then.

chrisdavidmills (approved):
LGTM to merge then.

ddbeck:
Thank you, @chrisdavidmills!

I have since updated the base branch from `master` and rebased. Merging this now.

PR #6739, opened 2020-09-21
Edge 79+ supports the Origin header without notes

mgol:
Edge 79+ uses the Chromium engine so it doesn't suffer from Edge Legacy implementation issues.

---

- [x] Summarize your changes

### DIFF
--- a/http/headers/origin.json
+++ b/http/headers/origin.json
@@ -13,7 +13,7 @@
             },
             "edge": {
               "version_added": "12",
-              "notes": "Not sent with <code>POST</code> requests, see <a href='https://developer.microsoft.com/microsoft-edge/platform/issues/10482384/'>bug 10482384</a>."
+              "notes": "Before Edge 79, this header was not sent with sent with <code>POST</code> requests."
             },
             "firefox": [
               {


vinyldarkscratch (dismissed):
Thanks for the PR!  This makes sense to me, however I think it would be better to keep a single support statement and modify the note to something like `Before Edge 79, this header was not sent with <code>POST</code> requests; see <a href='https://developer.microsoft.com/microsoft-edge/platform/issues/10482384/'>bug 10482384</a>.`?

mgol:
@vinyldarkscratch I see. What's the strategy for when multiple entries are used and when a single one?

vinyldarkscratch:
We don't have official guidelines for this, but I typically prefer to keep the support statements as few as possible to keep the end result clean.  If behavior described in a note has changed in a version, I prefer to add to the note to specify the version (as has been done for most of the notes in the repository).

ddbeck:
Whether to do one or many support statements is a bit of a judgement call. Generally, I suggest using more than one for cases which represent very different behavior (e.g., a function uses a callback then switches to a Promise) or when the behavior is substantially out-of-spec (i.e., it's broken, rather than merely buggy—this is usually paired with `partial_implementation`).

Unfortunately, the link to the Microsoft page is broken, but the Firefox bug on the same topic was instructive: it appears that POSTs weren't included in the original spec. This suggests that the behavior wasn't _broken_ at the time the new specification was adopted; rather, the spec expanded the applicable HTTP requests. This does point to using a single support statement, rather than two.

So @vinyldarkscratch's suggestion is mostly right, though we should drop the broken link: `Before Edge 79, this header was not sent with <code>POST</code> requests.`

mgol:
Thanks, I'll update the PR when I have a moment.

ddbeck:
Hi @mgol, do you plan to return to this PR? If not, let me know because we can take it over, but I'd prefer not to get in the way if you are working on this.

mgol:
@ddbeck Sorry for the delay, PR updated.

The only problem with this approach is that it taints Can I Use data as all entries for a specific browser - here Edge - will contain a note about something not being right. See how it looks now; it will essentialy look the same, only the note will be different:
https://caniuse.com/mdn-http_headers_origin
Another issue, as I understand it, is that the Edge entry here:
https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Origin
will have the star next to its version showed forever, even when Edge Legacy has stopped being relevant long ago. Compare that to the Firefox entry which shows the note but without the star.

ddbeck (approved):
Thank you, @mgol. I'll get this merged now. 🎉

About the notes: yes, it's a bit of a limitation with our schema when it comes to notes. There's several issues open about the way we handle notes. I hope this will improve with time, but right now there's nothing we can do about it directly. But one thing we can do—which we've done here—is to make sure the text of the note is explicit about when the note applies, at the beginning. For readers who don't care about Edge before 79, they know immediately that the note doesn't apply to them.

PR #6750, opened 2020-09-22
appearance is no longer experimental + other fixes

zcorpan:
- With Chrome and Firefox shipping, it's no longer experimental
- Unprefixed appearance was added in Firefox 80
- Safari's support is still a partial implementation
- Move 'button' into &lt;compat-auto> per spec change https://github.com/w3c/csswg-drafts/issues/5174

- [x] Summarize your changes
  - https://wpt.fyi/results/css/css-ui/appearance-cssom-001.html?label=experimental&label=master&aligned
  - https://github.com/mdn/browser-compat-data/pull/6505

### DIFF
--- a/css/properties/appearance.json
+++ b/css/properties/appearance.json
@@ -102,11 +102,16 @@
                 "prefix": "-webkit-"
               }
             ],
-            "opera_android": {
-              "version_added": "14",
-              "partial_implementation": true,
-              "prefix": "-webkit-"
-            },
+            "opera_android": [
+              {
+                "version_added": false
+              },
+              {
+                "version_added": "14",
+                "partial_implementation": true,
+                "prefix": "-webkit-"
+              }
+            ],
             "safari": {
               "version_added": "3",
               "partial_implementation": true,
@@ -164,7 +169,7 @@
                 "version_added": "69"
               },
               "opera_android": {
-                "version_added": false
+                "version_added": "59"
               },
               "safari": {
                 "version_added": false
@@ -247,10 +252,15 @@
                   "partial_implementation": true
                 }
               ],
-              "opera_android": {
-                "version_added": "14",
-                "partial_implementation": true
-              },
+              "opera_android": [
+                {
+                  "version_added": "59"
+                },
+                {
+                  "version_added": "14",
+                  "partial_implementation": true
+                }
+              ],
               "safari": {
                 "version_added": "3",
                 "partial_implementation": true
@@ -337,10 +347,15 @@
                   "partial_implementation": true
                 }
               ],
-              "opera_android": {
-                "version_added": "14",
-                "partial_implementation": true
-              },
+              "opera_android": [
+                {
+                  "version_added": "59"
+                },
+                {
+                  "version_added": "14",
+                  "partial_implementation": true
+                }
+              ],
               "safari": {
                 "version_added": "3",
                 "partial_implementation": true
@@ -481,10 +496,15 @@
                   "partial_implementation": true
                 }
               ],
-              "opera_android": {
-                "version_added": "14",
-                "partial_implementation": true
-              },
+              "opera_android": [
+                {
+                  "version_added": "59"
+                },
+                {
+                  "version_added": "14",
+                  "partial_implementation": true
+                }
+              ],
               "safari": {
                 "version_added": "3",
                 "partial_implementation": true
--- a/css/properties/appearance.json
+++ b/css/properties/appearance.json
@@ -190,16 +190,33 @@
           "__compat": {
             "description": "<code>&lt;compat-auto&gt;</code> (compatibility values <code>searchfield</code>, <code>textarea</code>, <code>push-button</code>, <code>slider-horizontal</code>, <code>checkbox</code>, <code>radio</code>, <code>square-button</code>, <code>menulist</code>, <code>listbox</code>, <code>meter</code>, <code>progress-bar</code>, <code>button</code>)",
             "support": {
-              "chrome": {
-                "version_added": "1"
-              },
-              "chrome_android": {
-                "version_added": "18"
-              },
-              "edge": {
-                "version_added": "12",
-                "partial_implementation": true
-              },
+              "chrome": [
+                {
+                  "version_added": "83"
+                },
+                {
+                  "version_added": "1",
+                  "partial_implementation": true
+                }
+              ],
+              "chrome_android": [
+                {
+                  "version_added": "83"
+                },
+                {
+                  "version_added": "18",
+                  "partial_implementation": true
+                }
+              ],
+              "edge": [
+                {
+                  "version_added": "83"
+                },
+                {
+                  "version_added": "12",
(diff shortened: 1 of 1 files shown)


css/properties/appearance.json
```diff
@@ -36,6 +36,9 @@
               }
             ],
             "firefox": [
+              {
+                "version_added": "80"
```
foolip, line 40: This seems right. Testing https://mdn-bcd-collector.appspot.com/tests/css/properties/appearance in Firefox 79 and 80, it starts returning true in 80.

css/properties/appearance.json
```diff
@@ -128,7 +134,7 @@
             ]
           },
           "status": {
-            "experimental": true,
+            "experimental": false,
```
foolip, line 137: I haven't reviewed a change to `experimental` before, but here's what I found:
- https://github.com/mdn/browser-compat-data/issues/1528
- https://developer.mozilla.org/en-US/docs/MDN/Contribute/Guidelines/Conventions_definitions#Experimental
- https://github.com/mdn/browser-compat-data/blob/master/schemas/compat-data-schema.md#status-information

With this shipping in two implementations now, it's no longer experimental 👍

css/properties/appearance.json
```diff
@@ -174,62 +180,15 @@
               }
             },
             "status": {
-              "experimental": true,
-              "standard_track": true,
-              "deprecated": false
-            }
-          }
-        },
-        "button": {
```
foolip, line 183: I'll add the needs-release-note label since this is removing an entry.

css/properties/appearance.json
```diff
@@ -258,26 +217,32 @@
                 "version_added": false
               },
               "opera": {
-                "version_added": "15"
+                "version_added": "15",
+                "partial_implementation": true
```
foolip, line 221: Can you say a bit about why this is considered partial? Is just because button was folded in? Notes would be helpful here, so one can tell when it should set to fully supported again, and so readers need not guess about what will work.

zcorpan: Opera 15 corresponds to an early version of Chromium which doesn't follow the spec and doesn't support `auto` etc. However, whichever version of Opera that maps to Chromium 84 should have full support. https://help.opera.com/en/opera-version-history/ only shows up to Chromium 81.

zcorpan: Fixed in 66b2ab0

I'm not sure how to explain the partial implementation in a short note. "Values can be applied to any element, rather than only widgets as per spec."?

ddbeck:
@zcorpan I don't have anything to add to the review, but I saw this is your first PR to the repo. Welcome to BCD!

css/properties/appearance.json
foolip: I usually check `browsers/opera.json` to find the matching release. But when preparing changing, you can also run something like `npm run mirror -- --modify=always opera css` and see what changes it makes.

css/properties/appearance.json
```diff
@@ -174,73 +180,43 @@
...
-            },
-            "status": {
-              "experimental": true,
+              "experimental": false,
               "standard_track": true,
               "deprecated": false
             }
           }
         },
         "compat-auto": {
```
foolip, line 189: If any other review is confused by this, the name of the "production" really is `<compat-auto>` in CSS: https://drafts.csswg.org/css-ui-4/#typedef-appearance-compat-auto

Usually "compat" in this repo is something internal :)

css/properties/appearance.json
```diff
@@ -250,34 +226,50 @@
...
+                  "version_added": "69"
+                },
+                {
+                  "version_added": "15",
+                  "partial_implementation": true
+                }
+              ],
               "opera_android": {
-                "version_added": "14"
+                "version_added": "14",
```
foolip, line 251: Can you add the non-`partial_implementation` entry here as well? If you run `npm run mirror -- --modify=always opera_android css` it will be among the resulting changes, but you may have to sift through some other changes. If they seem correct, feel free to throw them into this PR as well though.

css/properties/appearance.json
```diff
@@ -250,34 +226,50 @@
...
+                "partial_implementation": true
               },
               "safari_ios": {
-                "version_added": "1"
+                "version_added": "1",
+                "partial_implementation": true
               },
               "samsunginternet_android": {
-                "version_added": "1.0"
+                "version_added": "1.0",
```
foolip, line 263: If you run `npm run mirror -- --modify=always samsunginternet_android css` you'll get a bunch of updates to apply. One of those ought to be here.

css/properties/appearance.json
```diff
@@ -250,34 +226,50 @@
...
+                "partial_implementation": true
               },
               "webview_android": {
-                "version_added": "≤37"
+                "version_added": "≤37",
```
foolip, line 267: Same story, can you run `npm run mirror -- --modify=always webview_android css`?

css/properties/appearance.json
```diff
@@ -60,6 +63,9 @@
               }
             ],
             "firefox_android": [
+              {
+                "version_added": "80"
```
foolip, line 67: This looks good, but it looks like some Firefox Android data is missing, if you run `npm run mirror -- --modify=always firefox_android css`.

foolip (commented):
Thanks Simon, and welcome to BCD!

This looks good, and for any other first-time contributor I'd merge and follow up on the nits. But here, I might as well share the methods I'd use, if you continue contributing to BCD, which would be great :D

zcorpan:
@foolip woah, magic 🧙‍♀️

OK fixed.

css/properties/appearance.json
ddbeck: 👍 on adding notes for `partial_implementation`. That text is probably OK, but if you can point to an example or longer-form description of the problem, I can probably provide more detailed help on writing a note (it's definitely a challenging format for tech writing).

css/properties/appearance.json
```diff
@@ -102,11 +102,16 @@
...
               }
             ],
-            "opera_android": {
-              "version_added": "14",
-              "partial_implementation": true,
-              "prefix": "-webkit-"
-            },
+            "opera_android": [
+              {
+                "version_added": false
```
foolip, line 107:
```suggestion
                "version_added": "60"
```

`npm run mirror` didn't do the right thing here because ab53ef31eedfb0e36af82ebfbe8190bc694d1cf5 landed on master after you branched off.

foolip (approved):
LGTM % nit, which I'll try to apply myself.

css/properties/appearance.json
foolip: This wasn't done, but I'm not going to block the PR on that happening. @AdaRoseCannon FYI that there will be some updating needed here. Some of these changes in these PR will only be possible to mirror when the next version of Samsung Internet, probably 13.1, is in BCD.

foolip (approved):
LGTM after applying that fix. I'll merge now.

Issue #6749, opened 2020-09-22
URL constructor has deficiencies in Safari


dilyanpalauzov:
This does not work with WebKit 2.28.4 / Safari 13:
```javascript
new URL('http://x', undefined)
```
It throws an error, while in Firefox, Chromium, node it does not throw.

https://developer.mozilla.org/en-US/docs/Web/API/URL#Browser_compatibility says URL-constructor+Safari=♡.
 
See also https://bugs.webkit.org/show_bug.cgi?id=216841 and https://github.com/zloirock/core-js/pull/853 .

sideshowbarker:
Thanks for catching this.

I’ve confirmed that in Safari 14 (15610.1.28.1.9, 15610) and earlier, `new URL('http://x', undefined)` throws — but it doesn’t throw in current Safari Technology Preview (Release 113, WebKit 15610.2.3.1).

So I’ve raised #6752 to update the Safari data for the URL constructor to note this problem.

queengooborg:
Closing since this was taken care of by #6752!

PR #6752, opened 2020-09-23
Mark URL constructor support in Safari as partial

sideshowbarker:
This change marks the URL constructor with `partial_implementation:true` for Safari, due to Safari throwing for the case `new URL('http://x', undefined)` — that is, if the URL constructor is called with base URL whose value is undefined.

Relates to https://github.com/mdn/browser-compat-data/issues/6749
Relates to https://bugs.webkit.org/show_bug.cgi?id=216841

### DIFF
--- a/api/URL.json
+++ b/api/URL.json
@@ -145,10 +145,14 @@
               "version_added": "14"
             },
             "safari": {
-              "version_added": "6"
+              "version_added": "6",
+              "partial_implementation": true,
+              "notes": "In Safari 14 and earlier, calling the URL constructor with a base URL whose value is undefined causes Safari to throw a type error; see <a href='https://webkit.org/b/216841'>WebKit bug 216841</a>."
             },
             "safari_ios": {
-              "version_added": "6"
+              "version_added": "6",
+              "partial_implementation": true,
+              "notes": "In Safari 14 and earlier, calling the URL constructor with a base URL whose value is undefined causes Safari to throw a type error; see <a href='https://webkit.org/b/216841'>WebKit bug 216841</a>."
             },
             "samsunginternet_android": {
               "version_added": "1.5"
--- a/api/URL.json
+++ b/api/URL.json
@@ -152,7 +152,7 @@
             "safari_ios": {
               "version_added": "6",
               "partial_implementation": true,
-              "notes": "In Safari 14 and earlier, calling the URL constructor with a base URL whose value is undefined causes Safari to throw a type error; see <a href='https://webkit.org/b/216841'>WebKit bug 216841</a>."
+              "notes": "In Safari 14 and earlier, calling the <code>URL</code> constructor with a base URL whose value is <code>undefined</code> causes Safari to throw a <code>TypeError</code>; see <a href='https://webkit.org/b/216841'>WebKit bug 216841</a>."
             },
             "samsunginternet_android": {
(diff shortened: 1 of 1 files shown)


api/URL.json
```diff
@@ -145,10 +145,14 @@
               "version_added": "14"
             },
             "safari": {
-              "version_added": "6"
+              "version_added": "6",
+              "partial_implementation": true,
+              "notes": "In Safari 14 and earlier, calling the URL constructor with a base URL whose value is undefined causes Safari to throw a type error; see <a href='https://webkit.org/b/216841'>WebKit bug 216841</a>."
```
ddbeck, line 150: Suggestion: if this applies to all versions, let's not specify the version (so it doesn't go stale if this doesn't get fixed in the next release)

Style: more code styling! Why not?

```suggestion
              "notes": "Calling the <code>URL</code> constructor with a base URL whose value is <code>undefined</code> causes Safari to throw a <code>TypeError</code>; see <a href='https://webkit.org/b/216841'>WebKit bug 216841</a>."
```

api/URL.json
```diff
@@ -145,10 +145,14 @@
...
             },
             "safari_ios": {
-              "version_added": "6"
+              "version_added": "6",
+              "partial_implementation": true,
+              "notes": "In Safari 14 and earlier, calling the URL constructor with a base URL whose value is undefined causes Safari to throw a type error; see <a href='https://webkit.org/b/216841'>WebKit bug 216841</a>."
```
ddbeck, line 155:
```suggestion
              "notes": "Calling the <code>URL</code> constructor with a base URL whose value is <code>undefined</code> causes Safari to throw a <code>TypeError</code>; see <a href='https://webkit.org/b/216841'>WebKit bug 216841</a>."
```

ddbeck (changes requested):
Doesn't have to be exactly as I modified it, but I made some mostly stylistic suggestions on these notes. Thank you!

api/URL.json
ddbeck: Ugh, I'm sorry @sideshowbarker. I just got to the other tab I opened with #6749. I see now that it's probably fixed in whatever the next release of Safari is. That's kind of annoying—I would love to say something a bit simpler, like "Before Safari 14.1" or something. But since we don't know what the next version of Safar is going to be…

```suggestion
              "notes": "In Safari 14 and earlier, calling the <code>URL</code> constructor with a base URL whose value is <code>undefined</code> causes Safari to throw a <code>TypeError</code>; see <a href='https://webkit.org/b/216841'>WebKit bug 216841</a>."
```

api/URL.json
ddbeck:
```suggestion
              "notes": "In Safari 14 and earlier, calling the <code>URL</code> constructor with a base URL whose value is <code>undefined</code> causes Safari to throw a <code>TypeError</code>; see <a href='https://webkit.org/b/216841'>WebKit bug 216841</a>."
```

sideshowbarker:
> Doesn't have to be exactly as I modified it, but I made some mostly stylistic suggestions on these notes. Thank you!

I liked your suggestions just fine, so went ahead and committed them as-is. Thank you back~ 😋

ddbeck (approved):
Looks good! 🎉

PR #6753, opened 2020-09-23
CSS: text-underline-offset in development in chrome

manuelmeister:
Added spec url and status of chrome

https://www.w3.org/TR/css-text-decor-4/#underline-offset
https://chromestatus.com/feature/5747636764147712

- [x] Summarize your changes

### DIFF
--- a/css/properties/text-underline-offset.json
+++ b/css/properties/text-underline-offset.json
@@ -4,9 +4,11 @@
       "text-underline-offset": {
         "__compat": {
           "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/text-underline-offset",
+          "spec_url": "https://www.w3.org/TR/css-text-decor-4/#underline-offset",
           "support": {
             "chrome": {
-              "version_added": false
+              "version_added": false,
+              "notes": "Currently in development as indicated on Chrome Platform Status: <a href='https://chromestatus.com/feature/5747636764147712'>text-decoration-thickness, text-underline-offset and from-font keyword for text-underline-position</a>."
             },
             "chrome_android": {
               "version_added": false
--- a/css/properties/text-underline-offset.json
+++ b/css/properties/text-underline-offset.json
@@ -8,7 +8,7 @@
           "support": {
             "chrome": {
               "version_added": false,
-              "notes": "Currently in development as indicated on Chrome Platform Status: <a href='https://chromestatus.com/feature/5747636764147712'>text-decoration-thickness, text-underline-offset and from-font keyword for text-underline-position</a>."
+              "notes": "See Chrome Platform Status entry for <a href='https://chromestatus.com/feature/5747636764147712'>text-decoration-thickness, text-underline-offset and from-font keyword for text-underline-position</a>."
             },
             "chrome_android": {
               "version_added": false
--- a/css/properties/text-underline-offset.json
+++ b/css/properties/text-underline-offset.json
@@ -4,7 +4,6 @@
       "text-underline-offset": {
         "__compat": {
           "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/text-underline-offset",
-          "spec_url": "https://www.w3.org/TR/css-text-decor-4/#underline-offset",
           "support": {
             "chrome": {
               "version_added": false,


css/properties/text-underline-offset.json
```diff
@@ -4,9 +4,11 @@
       "text-underline-offset": {
         "__compat": {
           "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/text-underline-offset",
+          "spec_url": "https://www.w3.org/TR/css-text-decor-4/#underline-offset",
```
sideshowbarker, line 7:
```suggestion
```

sideshowbarker: We don’t yet add `spec_url` values to BCD for any features other than those in the `javascript` tree. (In https://github.com/w3c/browser-compat-data I have the spec URL data for all existing features, and I personally wish we would go ahead and upstream that to add them everywhere, but we’re not there yet…)

css/properties/text-underline-offset.json
```diff
@@ -4,9 +4,11 @@
...
           "support": {
             "chrome": {
-              "version_added": false
+              "version_added": false,
+              "notes": "Currently in development as indicated on Chrome Platform Status: <a href='https://chromestatus.com/feature/5747636764147712'>text-decoration-thickness, text-underline-offset and from-font keyword for text-underline-position</a>."
```
sideshowbarker, line 11:
```suggestion
              "notes": "Currently in development, as indicated on Chrome Platform Status: <a href='https://chromestatus.com/feature/5747636764147712'>text-decoration-thickness, text-underline-offset and from-font keyword for text-underline-position</a>."
```

sideshowbarker: I don’t think we usually add notes like this for features that are in development but which haven’t shipped in any release yet. I think we’ll need @ddbeck to make a call on this.

sideshowbarker (commented):
It’s exciting to know this feature’s in active development in Chrome, but I think we might need to consider waiting for it actually ship in a Chrome release before we update BCD for it. Let’s see what @ddbeck has to say.

css/properties/text-underline-offset.json
ddbeck: @sideshowbarker do we have an issue for `spec_url` everyhwere already? (I looked and couldn't find one.) This _seems like_ something that would be low-risk and high-reward, so please direct me to that issue or open it, so we can consider this more directly. 👍

css/properties/text-underline-offset.json
ddbeck: We don't seek out information on features that aren't yet shipping, but we do have some incidental bug links etc. I'm OK with having one here, since other browsers are shipping, but we need to make sure it can't go stale easily. Anything that says "currently" is _very_ prone to going out-of-date (development priorities change all the time), so we should go with something super minimal here, such as:

```suggestion
              "notes": "See Chrome Platform Status entry for <a href='https://chromestatus.com/feature/5747636764147712'>text-decoration-thickness, text-underline-offset and from-font keyword for text-underline-position</a>."
```

ddbeck (changes requested):

ddbeck:
Oh I just saw that this is your first BCD PR, @manuelmeister! Thank you and welcome!

css/properties/text-underline-offset.json
sideshowbarker: > @sideshowbarker do we have an issue for `spec_url` everywhere already? (I looked and couldn't find one.) This _seems like_ something that would be low-risk and high-reward, so please direct me to that issue or open it, so we can consider this more directly.

I realize now that I hadn’t yet opened a new issue for it. I had originally opened #2983 as a patch for adding spec URLs everywhere, but around https://github.com/mdn/browser-compat-data/pull/2983#issuecomment-430946543 (two years ago now…) the discussion about it with @Elchi3 resolved into deciding to just to it for JavaScript features. And so at that point, I pared the patch down to just that, and also changed the PR title and description to reflect that.

Anyway, I’ve now opened https://github.com/mdn/browser-compat-data/issues/6765 to track the broader plan.

manuelmeister: So should I revert this and only make a PR with the note?

sideshowbarker: > So should I revert this and only make a PR with the note?

There’s a **Commit suggestion** button above that you can push to apply the suggested commit (which will have the effect of reverting that part of the patch). And yes you should do that — just for consistency with the rest data in the `css` tree, none of which so far as `spec_url` values

sideshowbarker (approved):

sideshowbarker:
@manuelmeister Thanks much, and welcome to the MDN browser-compat-data project 🎉

PR #6754, opened 2020-09-23
Add missed features for Samsung Internet

AdaRoseCannon:
Last time I did an update I accidentally missed the CSS folder, this adds those changes and brings in a few changes that had been made in the meantime.

This is generated using the script.

### DIFF
--- a/css/properties/aspect-ratio.json
+++ b/css/properties/aspect-ratio.json
@@ -59,7 +59,9 @@
               "version_added": false
             },
             "samsunginternet_android": {
-              "version_added": false
+              "version_added": "12.0",
+              "partial_implementation": true,
+              "notes": "Samsung Internet 79 adds internal support only for mapped values"
             },
             "webview_android": {
               "version_added": "79",
@@ -121,7 +123,7 @@
                 "version_added": false
               },
               "samsunginternet_android": {
-                "version_added": false
+                "version_added": "12.0"
               },
               "webview_android": {
                 "version_added": "79"
--- a/css/properties/aspect-ratio.json
+++ b/css/properties/aspect-ratio.json
@@ -61,7 +61,7 @@
             "samsunginternet_android": {
               "version_added": "12.0",
               "partial_implementation": true,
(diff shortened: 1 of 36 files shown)


css/properties/aspect-ratio.json
```diff
@@ -59,7 +59,9 @@
               "version_added": false
             },
             "samsunginternet_android": {
-              "version_added": false
+              "version_added": "12.0",
+              "partial_implementation": true,
+              "notes": "Samsung Internet 79 adds internal support only for mapped values"
```
foolip, line 64: The version number here is wrong. I guess the script copies notes and replaced the browser name, but not version numbers?

css/properties/text-rendering.json
```diff
@@ -52,8 +52,11 @@
               "version_added": "4.2"
             },
             "samsunginternet_android": {
-              "version_added": false,
-              "notes": "This property is only supported on Windows and Linux. Samsung Internet is not on Windows or Linux."
+              "version_added": "1.0",
+              "notes": [
+                "This property is only supported on Windows and Linux.",
```
foolip, line 57: The old notes suggest that this shouldn't be listed as supported. That chrome_android is listed as supported is pretty weird too. Can you test whether it's actually supported, and update notes if so?

foolip (commented):
I skimmed the changes and they mostly seem plausible, but two notes seem wrong.

css/properties/text-rendering.json
AdaRoseCannon: Whoops that was a git mistake on my part that should've been reverted, thanks for the spot!

foolip (approved):
Looks good, thanks @AdaRoseCannon!

lolopedigo:
Hi there. My name is lauren.  I am being bombarded with these emails
and I have 2 phones that someone has somehow got into.  They have
destroyed my sanity and my qiulity of life and now they have taken
thousands of dollars from me.  I am getting encrypted messages from
somewhere and I just need someone to help me.  Can you help me decode
or or lead me to someone who can help me, please :(

PR #6763, opened 2020-09-25
Add visibilitychange event; Safari support partial

sideshowbarker:
This patch in this PR adds data for the `visibilitychange` event — basically just by copying over the existing data for the `onvisibilitychange` event handler, but with Safari support marked as partial (due the fact that Safari apparently doesn’t fire the event as expected when `visibilityState` transitions to `hidden`).

The patch in this PR also includes a change which marks Safari support for the `onvisibilitychange` event handler and `document.visibilityState` property as partial (for the same reason described above).

See also: https://bugs.webkit.org/show_bug.cgi?id=151234 (+ related bugs)
Marginally related: https://github.com/mdn/sprints/issues/3722

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -10801,6 +10801,67 @@
           }
         }
       },
+      "visibilitychange_event": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/Document/visibilitychange_event",
+          "description": "<code>visibilitychange</code> event",
+          "support": {
+            "chrome": [
+              {
+                "version_added": "33"
+              },
+              {
+                "version_added": "13",
+                "prefix": "webkit"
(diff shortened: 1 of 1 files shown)


sideshowbarker:
To indicate clearly to developers that when `visibilityState` transitions to `hidden`, the `visibilitychange` event doesn’t fire as expected in Safari, I added warnings to the following MDN articles:

* https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event
* https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilityState
* https://developer.mozilla.org/en-US/docs/Web/API/Document/onvisibilitychange

ddbeck (approved):
Looks good. Thank you! 🎉

PR #6767, opened 2020-09-27
Update intrinsicsize attribute support in Firefox and Chrome

LegionMammal978:
Neither Firefox nor Chrome currently support the `intrinsicsize` attribute on `` or `<video>` elements.

**Firefox:** The `intrinsicsize` attribute was initally described on April 24, 2018. I did an automated search of all versions of `xpcom/ds/nsGkAtomList.h` and `xpcom/ds/StaticAtoms.py` from that date, and I didn't find `intrinsicsize` in any of them. Since all attribute names are contained in this atom list, `intrinsicsize` can never have been an attribute of any HTML element.

**Chrome:** All functionality was removed in version 78; see https://bugs.chromium.org/p/chromium/issues/detail?id=997286 and the [relevant commit](https://chromium.googlesource.com/chromium/src.git/+/d1a6a7200ef9195b6cd701758eddb7a660790942).

It might be a good idea to mark the feature as deprecated, since WICG has since dropped it.

### DIFF
--- a/html/elements/img.json
+++ b/html/elements/img.json
@@ -382,6 +382,7 @@
             "support": {
               "chrome": {
                 "version_added": "71",
+                "version_removed": "78",
                 "flags": [
                   {
                     "type": "preference",
@@ -391,6 +392,7 @@
               },
               "chrome_android": {
                 "version_added": "71",
+                "version_removed": "78",
                 "flags": [
                   {
                     "type": "preference",
@@ -408,10 +410,10 @@
                 ]
               },
               "firefox": {
-                "version_added": null
+                "version_added": false
               },
               "firefox_android": {
-                "version_added": null
+                "version_added": false
               },
               "ie": {
                 "version_added": false
--- a/html/elements/img.json
+++ b/html/elements/img.json
@@ -401,13 +401,7 @@
                 ]
               },
               "edge": {
-                "version_added": "≤79",
-                "flags": [
-                  {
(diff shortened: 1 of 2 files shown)


jpmedley:
Consider this approval for the Chrome material. I'm not qualified to comment on Firefox.

ddbeck (changes requested):
Hi @LegionMammal978 and welcome to BCD! This is a really good clean up—thank you for investigating it. You're right that we should update the status information. Before we merge, I suggest changing both features' statuses to be:

```
{
 "experimental": false,
 "standard_track": false,
 "deprecated": true
}
```

Since we're here, there are two other changes that follow from the Chromium data. These suggestions are optional, but would make a great addition to the PR:

* Set Edge to `"version_added": false` — since Edge switched to Chromium _after_ this feature was removed.
* Set Opera and Opera for Android to `"version_added": false`. The existing data is wrong. For extra credit, you could map the data from Chrome to the equivalent versions of  Opera and Opera for Android (using the engine data in `browsers/opera*.json`). Either way though, we ought to show that the feature is not currently shipping in Opera either.

Thank you again!

LegionMammal978:
To be sure, I installed Opera 60 on a VM and verified that the attribute only worked when the flag was enabled.

ddbeck (approved):
Excellent. Thank you, @LegionMammal978! 🎉

PR #6769, opened 2020-09-28
adds support for overflow: clip in Firefox

rachelandrew:
Adding support in Firefox 81 for `overflow: clip`, and adding `-moz-hidden-unscrollable` as an old name.

Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1531609

Relates to: https://github.com/mdn/sprints/issues/3713

### DIFF
--- a/css/properties/overflow.json
+++ b/css/properties/overflow.json
@@ -69,7 +69,7 @@
                 },
                 {
                   "alternative_name": "-moz-hidden-unscrollable",
-                  "version_added": true,
+                  "version_added": null,
                   "version_removed": "81"
                 }
               ],
@@ -79,7 +79,7 @@
                 },
                 {
                   "alternative_name": "-moz-hidden-unscrollable",
-                  "version_added": true,
+                  "version_added": null,
                   "version_removed": "81"
                 }
               ],
--- a/css/properties/overflow.json
+++ b/css/properties/overflow.json
@@ -68,8 +68,9 @@
                   "version_added": "81"
                 },
                 {
-                  "version_added": "≤80",
-                  "alternative_name": "-moz-hidden-unscrollable"
+                  "alternative_name": "-moz-hidden-unscrollable",
+                  "version_added": true,
+                  "version_removed": "81"
                 }
               ],
               "firefox_android": [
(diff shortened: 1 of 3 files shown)


rachelandrew:
I can't find when overflow:-moz-hidden-unscrollable was implemented, but the linter won't let me use a range value, true, or null for `version_added` in order to add that as an old name.

chrisdavidmills:
This sounds OK to me, but the data structure here is a bit weird. @ddbeck can you give me a second pair of eyes on this?

ddbeck:
The short version: let's set this to `"version_added": "1.5"`.

The reason CI is rejecting `true` and `null` is that we now require real values for all the data in `css/`. When we achieved 100% coverage in CSS, we decided to lock that in by making it a requirement for all CSS data going forward. We've also never permitted a ranged value for any version Firefox (there's [an allow list](https://github.com/mdn/browser-compat-data/blob/16a00ff6cb865bd2d7ced073751febaaec83aaae/schemas/compat-data-schema.md#ranged-versions) for ranged versions).

I tried to find the actual version, but I couldn't narrow it down to a single version. I tested it as far back as Firefox 3 (I couldn't test before that) and I found a few bugs mentioning problems with `-moz-hidden-unscrollable` dating back to at least 2007 (e.g., [bug 404140](https://bugzilla.mozilla.org/show_bug.cgi?id=404140)). Since the Firefox 2 release notes don't mention any CSS changes, we have a likely release range from Firefox 1 to 3, and the 1.5 release notes mention a bunch of CSS changes. We're talking about something pretty ancient, so I'm happy to go with an educated guess.

css/properties/overflow.json
```diff
@@ -63,12 +63,26 @@
...
+                  "version_removed": "81"
+                }
+              ],
+              "firefox_android": [
+                {
+                  "version_added": "81"
+                },
+                {
+                  "alternative_name": "-moz-hidden-unscrollable",
+                  "version_added": null,
```
ddbeck, line 82: Firefox for Android's first release:

```suggestion
                  "version_added": "4",
```

ddbeck (changes requested):

rachelandrew:
thanks @ddbeck

ddbeck:
Looks great. Thank you, @rachelandrew! 🎉

PR #6774, opened 2020-09-30
forgiving selector list support for is and where pseudos

rachelandrew:
Firefox 82 is adding support for the forgiving selector list value for `is()` and `where()`. I've added this as an entry rather than just a note as it's actually quite a different behaviour.

I also noted that Safari 14 implements `is()` and `where()` so have updated the general section to add this: https://developer.apple.com/documentation/safari-release-notes/safari-14-release-notes

I also checked that these pseudos are still behind a flag in Chrome and confirmed they are.

This relates to: https://bugzilla.mozilla.org/show_bug.cgi?id=1664718

### DIFF
--- a/css/selectors/is.json
+++ b/css/selectors/is.json
@@ -2,227 +2,225 @@
   "css": {
     "selectors": {
       "is": {
-        "basic_support": {
-          "__compat": {
-            "description": "<code>:is()</code>",
-            "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/:is",
-            "support": {
-              "chrome": [
-                {
-                  "version_added": "68",
-                  "flags": [
-                    {
-                      "type": "preference",
-                      "name": "#enable-experimental-web-platform-features",
-                      "value_to_set": "Enabled"
-                    }
-                  ],
-                  "notes": "Combinators in the selector list argument may not match correctly (see <a href='https://crbug.com/842157'>bug 842157</a>)."
-                },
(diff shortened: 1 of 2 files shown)


css/selectors/is.json
```diff
@@ -2,219 +2,276 @@
...
-                  {
-                    "type": "preference",
-                    "name": "layout.css.is-where-selectors.enabled",
-                    "value_to_set": "enabled"
-                  }
-                ],
-                "notes": "Enabled by default in Firefox Nightly."
-              },
-              {
+        "basic_support": {
```
ddbeck, line 5: You're right to create a new feature for `forgiving_selector_list`, but the existing `"__compat"` object doesn't need to be wrapped in this new `"basic_support"` object. Instead, put the original `"__compat"` key and the new `"forgiving_select_list"` key at the same level (with `"forgiving_select_list"` after the original `"__compat"`).

ddbeck (changes requested):
Hi @rachelandrew and thank you for this PR. I've got a request to change the structure a little—see my line comment for a complete explanation. I think that will make the rest of the diff easier to understand, though I don't expect to have any comments on the substance, based on what I've seen so far.

rachelandrew:
@ddbeck I'm hoping this is right - let me know if not!

ddbeck (approved):
Looks great. Thank you, @rachelandrew! 🎉

PR #6775, opened 2020-09-30
Fix two mistakes about AudioWorklet interfaces availability

padenot:
This has been available by default in Firefox starting in 76, and is not experimental. Additionally, I'd like to mark it as available on Fenix, should we use `firefox_android` ?

### DIFF
--- a/api/AudioWorklet.json
+++ b/api/AudioWorklet.json
@@ -17,7 +17,7 @@
             "version_added": "76"
           },
           "firefox_android": {
-            "version_added": false
+            "version_added": "79"
           },
           "ie": {
             "version_added": false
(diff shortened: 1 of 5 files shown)


padenot:
Per @chrisdavidmills, the version for Firefox for Android should be the first publicly released version that has this. It's the first version of Fenix, which runs Gecko 79.

ddbeck (approved):
Looks good, @padenot! I'm assuming you worked on these features directly and that this PR is authoritative.

> I'd like to mark it as available on Fenix, should we use `firefox_android` ?

It looks like you already sorted this out but, for the record, this is correct. We generally follow the public name of browsers, rather than engine or project codenames, as these names are what web developers are most familiar with.

Thank you!

padenot:
> Looks good, @padenot! I'm assuming you worked on these features directly and that this PR is authoritative.

Yes exactly, that's why I didn't really provide any other source. Thanks for the quick turnaround!

PR #6777, opened 2020-09-30
Add mdn- and Node.js 8 deprecation warnings for 1.1 release

ddbeck:
Another entry in the #6640 saga and the v1.1 counterpart to the v2.0.0 changes in #6713.

## What this PR attempts to do
- Adds a notice to the top of the README (which will appear on the npm package page)
- Adds docs for upgrading 1.1.x → 2.0.x
- Adds a deprecation notice that the package name is changing and to use the scoped package instead. It links to the aforementioned upgrade doc
- Adds a deprecation notice for Node.js 8. This should only fire if Node.js 8 is detected in the `process.version` string

## What needs a close look
* Do the deprecation warnings make sense? Are they too wordy or not wordy enough?
* Do the upgrade instructions make sense? Are there any missing steps?
* Would you be comfortable seeing this ship in the next release of BCD?

## Additional notes
* I've tested the deprecation notices on my computer (using [`nodenv`](https://github.com/nodenv/nodenv) to switch between Node.js versions), but I'd welcome tests from others. You can install it from my branch directly: `npm install https://github.com/ddbeck/browser-compat-data#add-deprecation-warnings`
* The link to the `UPGRADE` doc from the deprecation notice won't work until there's a tag for `v1.1.0`. This is expected; the tag will be created by `npm version` when I do the 1.1 release
* The pattern for emitting the warnings once and only once comes from [Node.js's `process.emitWarning()` docs](https://nodejs.org/api/process.html#process_avoiding_duplicate_warnings). It's a bit verbose, but it seems to be preferred

## Known follow-up tasks

Once the `master-scoped-package` branch merges back into `master`, we'll want to:

- Make sure the notice at the top of the README goes away
- Remove the upgrade doc (I assume linking the file with the tag ref will be adequate)

### DIFF
--- a/README.md
+++ b/README.md
@@ -1,4 +1,4 @@
-**Warning:** This package has soon to be renamed. Use [`@mdn/browser-compat-data`](https://www.npmjs.com/package/@mdn/browser-compat-data) instead. If you're already using `mdn-browser-compat-data`, read [the upgrade guide](https://github.com/mdn/browser-compat-data/blob/v1.1.0/UPGRADE-2.0.md).
+**Warning:** This package is soon to be renamed. Use [`@mdn/browser-compat-data`](https://www.npmjs.com/package/@mdn/browser-compat-data) instead. If you're already using `mdn-browser-compat-data`, read [the upgrade guide](https://github.com/mdn/browser-compat-data/blob/v1.1.0/UPGRADE-2.0.md).
 
 # mdn-browser-compat-data
 
--- a/README.md
+++ b/README.md
@@ -1,3 +1,5 @@
+**Warning:** This package is soon-to-be renamed. Use [`@mdn/browser-compat-data`](https://www.npmjs.com/package/@mdn/browser-compat-data) instead. If you're already using `mdn-browser-compat-data`, read [the upgrade guide](https://github.com/mdn/browser-compat-data/blob/v1.1.0/UPGRADE-2.0.md).
+
 # mdn-browser-compat-data
 
 [https://github.com/mdn/browser-compat-data](https://github.com/mdn/browser-compat-data)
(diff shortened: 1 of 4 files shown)


ddbeck:
Oh good, we got the correct deprecation notices in tests:

- The GitHub Actions build (running Node.js 8) emits both warnings: https://github.com/mdn/browser-compat-data/runs/1188596653?check_suite_focus=true#step:5:2272
- The Travis CI build (running Node.js 10) emits only the package warning: https://travis-ci.org/github/mdn/browser-compat-data/jobs/731640511#L2487

ddbeck:
Reviewing the diff here, I noticed that the Node version-checking code runs whether the warning has been emitted already or not. I moved it into the warning function and confirmed that the tests still show the correct warnings.

README.md
```diff
@@ -1,3 +1,5 @@
+**Warning:** This package is soon-to-be renamed. Use [`@mdn/browser-compat-data`](https://www.npmjs.com/package/@mdn/browser-compat-data) instead. If you're already using `mdn-browser-compat-data`, read [the upgrade guide](https://github.com/mdn/browser-compat-data/blob/v1.1.0/UPGRADE-2.0.md).
```
chrisdavidmills, line 1: `soon-to-be` doesn't need hyphens in this context, imo. If it was in a context like "The soon-to-be-renamed browser-compat-data package", then it would have hyphens.

UPGRADE-2.0.x.md
```diff
@@ -0,0 +1,30 @@
...
+
+3. Add `@mdn/browser-compat-data` to your package dependencies.
+
+   In the same directory as your `package.json` file, run `npm install @mdn/browser-compat-data`.
+
+4. In your code, replace any `require("mdn-browser-compat-data")` calls with `require("@mdn/browser-compat-data")`.
+
+   If possible, run your test suite to make sure this worked.
+
+If you encountered any undocumented breaking changes as a result of this upgrade, please [open an issue](https://github.com/mdn/browser-compat-data/issues/new).
```
chrisdavidmills, line 30: Link "breaking changes" to the section named "Before you start", above, as this is where the breaking changes are documented?

chrisdavidmills (changes requested):
A couple of small bits, not nothing major. Nice work @ddbeck !

README.md
ddbeck: Thank you, fixed with 638d3fcf017f50fd5cd370e78fe77963fcca3d1f. I wasn't sure if I was modifying _is_ or _renamed_.

UPGRADE-2.0.x.md
ddbeck: 👍 Fixed with 11551cd97dad6c41af942afa3f03cfed6e113faf.

README.md
ddbeck: And fixed again with 2aba539b605e30950089e17c2d555bab9ac705ba.

chrisdavidmills (approved):
Looks great, nice one @ddbeck .

PR #6778, opened 2020-09-30
Update CODEOWNERS

ddbeck:
@Elchi3 @wbamberg I'm swapping in my username for yours for automated PR review requests. I hope that you'll return to active participation here soon, but until that day comes, I figured it was only fair I were spammed with notifications and not you.

If either of you object, please chime in, otherwise I'll assume this is OK and get Chris to merge this in the next few days. Thank you!

### DIFF
--- a/.github/CODEOWNERS
+++ b/.github/CODEOWNERS
@@ -3,9 +3,9 @@
 
 # Schema, linter and infrastructure changes must be reviewed by an owner:
 # ======================================================================
-/test/    @Elchi3
-/schemas/ @Elchi3
-/.github/ @Elchi3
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
Meh, but yes, it makes sense :) 
You're doing really great work here by the way, Daniel! :100: :tada:

ddbeck:
> Meh

@Elchi3 That's been my feeling about it too. Thank you for the kind words though. 😃

ddbeck:
Got a thumbs up from both of you, so I'm merging. Thanks again! ❤️

PR #6782, opened 2020-09-30
inputmode is supported on Firefox 79 for Android

makotokato:
Shipped by https://bugzilla.mozilla.org/show_bug.cgi?id=1631681.

Also, `inputmode` preference should be same version of Firefox Desktop.

- [x] Summarize your changes

### DIFF
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -1323,7 +1323,7 @@
               ]
             },
             "firefox_android": {
-              "version_added": false
+              "version_added": "79"
             },
             "ie": {
               "version_added": false
--- a/html/global_attributes.json
+++ b/html/global_attributes.json
@@ -1107,16 +1107,21 @@
                 "version_removed": "23"
               }
             ],
-            "firefox_android": {
-              "version_added": "68",
-              "flags": [
-                {
-                  "type": "preference",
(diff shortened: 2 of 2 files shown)


ddbeck:
Thanks for this PR!

> Also, `inputmode` preference should be same version of Firefox Desktop.

Do you want to include this change in this PR, @makotokato? I'd welcome that addition, or a follow-up. If you don't want to do it at all, we could open an issue to track that suggestion separately.

foolip:
I think only remaining question here is missing flag data. Since we're definitely missing flag data all over the place, that seems quite OK to me. It will become consistent in 2 years when the existing flag data can be removed :)

makotokato:
> Do you want to include this change in this PR, @makotokato? I'd welcome that addition, or a follow-up. If you don't want to do it at all, we could open an issue to track that suggestion separately.

Hmm, OK, this PR is Firefox Android 79 only. So I will update this PR.

Also, I will file another PR that is https://github.com/mdn/browser-compat-data/pull/6468 is incorrect.

ddbeck (approved):
Thank you, @makotokato! 

Also, I wasn't entirely clear on what the situation was for Firefox on the desktop, so I opened #6921 to track that independently.

PR #6784, opened 2020-09-30
aspect-ratio internal mapping supported Safari 14

JoshTumath:
`aspect-ratio` mapping of `width` and `height` attributes to `img` element shipped in Safari 14. ~Not yet released on iOS.~

See: https://developer.apple.com/documentation/safari-release-notes/safari-14-release-notes#Media

### DIFF
--- a/css/properties/aspect-ratio.json
+++ b/css/properties/aspect-ratio.json
@@ -53,7 +53,9 @@
               "version_added": false
             },
             "safari": {
-              "version_added": false
+              "version_added": "14",
+              "partial_implementation": true,
+              "notes": "Safari 14 adds internal support only for mapped values"
             },
             "safari_ios": {
               "version_added": false
--- a/css/properties/aspect-ratio.json
+++ b/css/properties/aspect-ratio.json
@@ -58,7 +58,9 @@
               "notes": "Safari 14 adds internal support only for mapped values"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "14",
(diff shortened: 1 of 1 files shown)


jensimmons:
> Not yet released on iOS.

Safari for iO14 and iPad OS 14 do both support this feature. And both have shipped.

jensimmons:
Might I also suggest the title for this data on MDN be changed. Currently it's 
"CSS property: aspect-ratio: Internal mapping of width and height", but this actually has nothing to do with CSS or the `aspect-ratio` property. This is a change to the HTML specification, and how the height and width attributes are used in providing information to the browser about an image's intrinsic aspect ratio.

JoshTumath:
> Safari for iO14 and iPad OS 14 do both support this feature. And both have shipped.

Sorry, my source for checking if Safari 14 was released on iOS was er... Wikipedia, so not the most reliable source. 🙈 I'll update the PR to include iOS.

ddbeck:
@jensimmons

> this actually has nothing to do with CSS or the `aspect-ratio` property. This is a change to the HTML specification, and how the height and width attributes are used in providing information to the browser about an image's intrinsic aspect ratio.

I noticed this as well and opened #6786 to address it. I have written a more complete proposal for representing that data in that issue and I'd appreciate your feedback on it there. Thank you!

ddbeck (approved):
Looks good. We'll get this merged now and clean up the structure of this data in a follow up. Thank you, @JoshTumath! 🎉

PR #6789, opened 2020-10-01
Update AudioContext + BaseAudioContext data

vinyldarkscratch:
This PR uses the mdn-bcd-collector project to update the data for the `AudioContext` and `BaseAudioContext` APIs, including updating the version information for features, as well as correcting version numbers for the AudioContext constructor.  This also updates the `BaseAudioContext` version numbers to be the same as `AudioContext`'s (since it doesn't make sense for it to be added before or after the AudioContext).

### DIFF
--- a/api/AudioContext.json
+++ b/api/AudioContext.json
@@ -90,14 +90,14 @@
           "description": "<code>AudioContext()</code> constructor",
           "support": {
             "chrome": {
-              "version_added": "55",
+              "version_added": "35",
               "notes": [
                 "Each tab is limited to 6 audio contexts in Chrome; attempting to create more will throw a <code>DOMException</code>. For details see <a href='https://developer.mozilla.org/docs/Web/API/AudioContext/AudioContext#Per-tab_audio_context_limitation_in_Chrome'>Per-tab audio context limitation in Chrome</a>.",
                 "If <code>latencyHint</code> isn't valid, Chrome throws a <code>TypeError</code> exception. See <a href='https://developer.mozilla.org/docs/Web/API/AudioContext/AudioContext#Non-standard_exceptions_in_Chrome'>Non-standard exceptions in Chrome</a> for details."
               ]
             },
             "chrome_android": {
-              "version_added": "55",
+              "version_added": "35",
               "notes": [
                 "Each tab is limited to 6 audio contexts in Chrome; attempting to create more will throw a <code>DOMException</code>. For details see <a href='https://developer.mozilla.org/docs/Web/API/AudioContext/AudioContext#Per-tab_audio_context_limitation_in_Chrome'>Per-tab audio context limitation in Chrome</a>.",
                 "If <code>latencyHint</code> isn't valid, Chrome throws a <code>TypeError</code> exception. See <a href='https://developer.mozilla.org/docs/Web/API/AudioContext/AudioContext#Non-standard_exceptions_in_Chrome'>Non-standard exceptions in Chrome</a> for details."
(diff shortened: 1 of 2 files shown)


foolip:
> This also updates the `BaseAudioContext` version numbers to be the same as `AudioContext`'s (since it doesn't make sense for it to be added before or after the AudioContext).

The history of this interface is that first there was just `AudioContext`, and later `BaseAudioContext` was added as a common parent interface for both `AudioContext` and the new `OfflineAudioContext`. So while supporting `AudioContext` without `BaseAudioContext` is impossible per the spec as it looks today, I think every browser actually should have `BaseAudioContext` introduced after `AudioContext`.

api/AudioContext.json
```diff
@@ -90,21 +90,21 @@
           "description": "<code>AudioContext()</code> constructor",
           "support": {
             "chrome": {
-              "version_added": "55",
+              "version_added": "35",
```
foolip, line 93: I tested https://software.hixie.ch/utilities/js/live-dom-viewer/saved/8529 in Chrome 34 and 35, and indeed the unprefixed constructor came in 35. But the prefixed constructor was there earlier, since Chrome 10 if https://caniuse.com/audio-api is correct. Or 14 if the api.AudioContext entry is correct. Can you test old Chrome and update?

BCD is not consistent about whether prefix data should be on the interface, constructor or both. Here, the prefix is repeated for Safari but not for Chrome. I'd prefer for it to be repeated for Chrome as well, since otherwise in the context of https://developer.mozilla.org/en-US/docs/Web/API/AudioContext/AudioContext it will look like the unprefixed constructor is supported going back before Chrome 35, which will be wrong.

Also, this is a good case for adding prefixed constructor tests in mdn-bcd-collector. That there's a prefixed BCD entry for the interface should be enough to infer that the test is needed.

api/BaseAudioContext.json
```diff
@@ -5,54 +5,40 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/BaseAudioContext",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "14"
```
foolip, line 8: This isn't from mdn-bcd-collector, right? Can you update the data to match the collector results, and not any manual changes from that? `BaseAudioContext` was introduced much later than this.

api/BaseAudioContext.json
```diff
@@ -113,15 +99,20 @@
...
-              "version_added": "10",
-              "prefix": "webkit"
-            },
+            "chrome": [
+              {
+                "version_added": "56"
+              },
+              {
+                "version_added": "14",
+                "prefix": "webkit"
```
foolip, line 108: I'm pretty sure there's never been a `webkitCreateAnalyser` method anywhere as this would imply. The prefixed things have been `webkitAudioContext`, `webkitAudioPannerNode` and `webkitOfflineAudioContext`.

I presume this change wasn't suggested by mdn-bcd-collector?

api/BaseAudioContext.json
```diff
@@ -178,15 +169,20 @@
...
```
foolip, line 178: I think this is also wrong, that there's been no `webkitCreateBiquadFilter`. There are a bunch more of these that follow, so I won't comment on each one.

api/BaseAudioContext.json
```diff
@@ -113,15 +99,20 @@
         "__compat": {
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/BaseAudioContext/createAnalyser",
           "support": {
-            "chrome": {
-              "version_added": "10",
-              "prefix": "webkit"
-            },
+            "chrome": [
+              {
+                "version_added": "56"
```
foolip, line 104: This runs into https://github.com/mdn/browser-compat-data/issues/3463. `createAnalyser` existed unprefixed much earlier, but on the `AudioContext` interface.

I think what we'll have to do is an entry with a note for the versions of Chrome before `BaseAudioContext` was added that instead had `createAnalyser` on `AudioContext`.

vinyldarkscratch:
This one has a little bit more complexity that I'd like to look into before I say this PR is ready.  Marking as a draft while I look into it more.

api/BaseAudioContext.json
vinyldarkscratch: Unfortunately, this will cause consistency issues if we set it to what comes from the collector, which is "56".  We've got entries within here that were initially implemented on just AudioContext.

vinyldarkscratch:
Due to the complexity of BaseAudioContext, and the cherry-picking of AudioContext's changes, I will be closing this PR.

PR #6790, opened 2020-10-01
Use mdn-bcd-collector to update version numbers for a few Web Audio APIs

vinyldarkscratch:
This PR uses the mdn-bcd-collector project to update version numbers in five Web Audio APIs.  Results have been double-checked for accuracy.

### DIFF
--- a/api/AudioProcessingEvent.json
+++ b/api/AudioProcessingEvent.json
@@ -59,7 +59,7 @@
               "version_added": "57"
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "79"
             },
             "firefox": {
               "version_added": false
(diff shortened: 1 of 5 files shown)


api/AudioProcessingEvent.json
```diff
@@ -71,16 +71,16 @@
...
-              "version_added": null
+              "version_added": "44"
             },
             "opera_android": {
-              "version_added": null
+              "version_added": "43"
             },
             "safari": {
-              "version_added": null
+              "version_added": false
```
foolip, line 80: Seems right, this was just landed in WebKit: https://bugs.webkit.org/show_bug.cgi?id=215237

api/AudioTrack.json
```diff
@@ -664,7 +664,7 @@
               "version_added": false
             },
             "ie": {
-              "version_added": false
+              "version_added": "11"
```
foolip, line 667: api.VideoTrack.sourceBuffer should also change, but that need not happen in this PR, as it's not an inconsistency that jumps out at you when looking at any single MDN page.

foolip (approved):

PR #6791, opened 2020-10-01
Use mdn-bcd-collector to update Web Audio API nodes

vinyldarkscratch:
This PR uses the mdn-bcd-collector to update various web audio nodes, specifically their constructors in Edge and Safari.

### DIFF
--- a/api/AnalyserNode.json
+++ b/api/AnalyserNode.json
@@ -59,7 +59,7 @@
               "version_added": "55"
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "79"
             },
             "firefox": {
               "version_added": "53"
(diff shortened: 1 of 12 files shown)


api/AnalyserNode.json
```diff
@@ -77,10 +77,10 @@
               "version_added": "42"
             },
             "safari": {
-              "version_added": "6"
+              "version_added": false
```
foolip, line 80: The constructor was just recently implemented: https://trac.webkit.org/changeset/265196/webkit

api/PannerNode.json
```diff
@@ -61,7 +61,7 @@
               "notes": "Before Chrome 59, the default values were not supported."
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "79"
```
foolip, line 64: This is the `PannerNode` constructor. The Safari data is null here, and that's OK, but I'll just note that `webkitAudioPannerNode exists in Safari. It doesn't work as a constructor, though.

As a follow-up, it would be good to add `"alternative_name": "webkitAudioPannerNode"` for Safari. I don't know why just that interface is exposed, when many others are not, but that's the situation.

foolip (approved):

PR #6793, opened 2020-10-01
Add note about HTTP Content-Disposition inline directive FF82

hamishwillee:
This adds a note about a change in behaviour for FF82 in the From FF82 the download directive for an anchor is respected rather than the HTTP Content-Disposition header, inline directive. 
- The defect fix is https://bugzilla.mozilla.org/show_bug.cgi?id=1658877
- Related documentation updates are here: https://github.com/mdn/sprints/issues/3733

Behaviour was verified by the engineer in the bug report.

### DIFF
--- a/http/headers/content-disposition.json
+++ b/http/headers/content-disposition.json
@@ -15,7 +15,11 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": true
+              "version_added": true,
+              "notes": [
+                "From version 82 the inline directive is ignored if the HTML5 'a download' attribute is specified (to match the specification)",
+                "Earlier versions respected the header inline directive over the HTML, see <a href='https://bugzil.la/1658877 '>bug 1658877</a>."
+              ] 
             },
             "firefox_android": {
               "version_added": true
--- a/http/headers/content-disposition.json
+++ b/http/headers/content-disposition.json
@@ -16,13 +16,11 @@
             },
             "firefox": {
               "version_added": true,
-              "notes": [
-                "From version 82 the inline directive is ignored if the HTML5 'a download' attribute is specified (to match the specification)",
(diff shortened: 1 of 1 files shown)


hamishwillee:
I think this is a linter problem though I can't see obviously what I have wrong. I can probably fix it, but before I do want to check the content seems reasonable (this is my first submit to BCD)

http/headers/content-disposition.json
```diff
@@ -15,7 +15,11 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": true
+              "version_added": true,
+              "notes": [
+                "From version 82 the inline directive is ignored if the HTML5 'a download' attribute is specified (to match the specification)",
```
ddbeck, line 20: A suggestion and a question here.

Suggestion: I think these two notes tell the same story, so we could probably combine them. Also, I had a little difficulty following the `a download` story—I had to read the bug to figure out. What do you think of this rewrite that combines both?

```
"notes": "From version 82, if an <code>&lt;a&gt;</code> element's <code>download</code> attribute is set then the <code>inline</code> directive is ignored. Earlier versions did not match the specification and respected the header directive over the attribute. See <a href='https://bugzil.la/1658877'>bug 1658877</a>."
```

Question: Do you know if this applies to Firefox for Android as well? (If so, I'd copy whatever final note we get to the "firefox_android" entry below.)

ddbeck (changes requested):
Hi @hamishwillee and welcome to BCD! This is a great first PR. You've got the right idea, but I have some suggestions for improving the note(s). Thank you!

hamishwillee:
@ddbeck Thanks very much - that is a much cleaner wording. I have made minor modification to the note to clarify that download only works in same-origin requests and added to both FF and FF android (which it should apply to given this is part of the core).

I believe it should be good to merge as it passes the test.

ddbeck (approved):
Looks good. Thank you again, @hamishwillee! 🎉

PR #6794, opened 2020-10-02
Update BlobBuilder (removed everywhere)

foolip:
Updated primarily because of the incorrect prefix. It was
"WebKitBlobBuilder" from its introduction until its removal:
https://trac.webkit.org/changeset/83884/webkit
https://trac.webkit.org/changeset/130343/webkit

Chrome's `version_removed` was determined using this test:
http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8531

The `version_added` was not verified, and the removal version was
mirrored for *_android.

Edge 15 and 18 were tested to confirm that there was only
"MSBlobBuilder".

### DIFF
--- a/api/BlobBuilder.json
+++ b/api/BlobBuilder.json
@@ -11,7 +11,7 @@
           },
           "chrome_android": {
             "version_added": "18",
-            "version_removed": "24",
+            "version_removed": "25",
             "prefix": "WebKit"
           },
           "edge": {
--- a/api/BlobBuilder.json
+++ b/api/BlobBuilder.json
@@ -6,27 +6,19 @@
         "support": {
           "chrome": {
             "version_added": "8",
-            "prefix": "Webkit"
+            "version_removed": "24",
+            "prefix": "WebKit"
           },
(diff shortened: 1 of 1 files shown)


api/BlobBuilder.json
```diff
@@ -6,27 +6,19 @@
...
           "chrome": {
             "version_added": "8",
-            "prefix": "Webkit"
+            "version_removed": "24",
+            "prefix": "WebKit"
           },
           "chrome_android": {
             "version_added": "18",
-            "prefix": "Webkit"
+            "version_removed": "24",
```
ddbeck, line 14: I think this needs to be bumped up to `"25"` to pass the tests

```suggestion
            "version_removed": "25",
```

ddbeck (changes requested):
Looks good. Just needs a small fix to clear the tests. Thank you!

api/BlobBuilder.json
ddbeck: (feel free to merge once this is fixed, if this is the only change)

PR #6797, opened 2020-10-02
Use Node.js ≥10 for CI and drop Travis

ddbeck:
This upgrades the GitHub Actions workflows to use Node.js 10 and later (10 for publishing releases, 10 and 14 for tests). It also removes the config for Travis CI.

Note that this PR applies to the `master-scoped-package` branch; it won't affect work on `master` until after we end BCD 1.x-series releases (and consequently merge `master-scoped-branch` into `master`).

This is a step toward #5931, but it doesn't completely resolve it.

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -16,8 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        # TODO: hoist to 10.x when we bump the version supported in #5931
-        node-version: [ 8.x ]
+        node-version: [ 10.x, 14.x ]
 
     steps:
     - uses: actions/checkout@v2
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -12,8 +12,7 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-node@v2-beta
       with:
-        # TODO: hoist to 10.x when we bump the version supported in #5931
-        node-version: 8.x
+        node-version: 10.x
         registry-url: "https://registry.npmjs.org/"
     - run: npm install
     - run: npm test
--- a/.travis.yml
+++ b/.travis.yml
@@ -1,7 +1,0 @@
-language: node_js
-node_js:
-  - 8
-  - 'lts/dubnium'
-sudo: false
-notifications:
-  email: false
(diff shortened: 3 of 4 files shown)


.github/workflows/node.js.yml
```diff
@@ -16,8 +16,7 @@ jobs:
 
     strategy:
       matrix:
-        # TODO: hoist to 10.x when we bump the version supported in #5931
-        node-version: [ 8.x ]
+        node-version: [ 10.x, 14.x ]
```
foolip, line 19: Should we not keep 8.x here until ">=8.0.0" in package.json changes?

ddbeck: This has already happened for the base branch (`master-scoped-package`), but not `master`: https://github.com/mdn/browser-compat-data/blob/3331bed505ca9115150079eaa27a2d27d49dc301/package.json#L11 In other words, we'll still have Node 8.x on `master` until we're done with #6640.

foolip: Oh, I didn't notice the target branch.

foolip (approved):

vinyldarkscratch:
We'll also need to update the readme to include a new build status badge, since it's currently connected to Travis CI.

foolip:
If there's no first party badge support for GitHub Actions then just not having one seems fine too. Maybe it's just me, but I've never found a badge like that useful when looking at someone else's repository or for my own.

sideshowbarker:
> If there's no first party badge support for GitHub Actions then just not having one seems fine too. Maybe it's just me, but I've never found a badge like that useful when looking at someone else's repository or for my own.

I have personally also never found build-status badges useful at all in repo READMEs, so I’d also vote for not considering it a loss if we don’t have one.

ddbeck:
I went ahead and dropped all the badges. I'm also a member of team doesn't-care-about-badges, but I thought there was a case for removing each one individually:

- The Travis badge needed to go for aforementioned reasons
- The npm badge duplicates the new GitHub releases sidebar
- The Twitter badge looked pretty sad all by itself, plus it's kinda cryptic (not actually saying what social network it's for)

peterbe (approved):

PR #6799, opened 2020-10-03
Remove duplicate api.AudioContext.createConstantSource

vinyldarkscratch:
This PR removes `api.AudioContext.createConstantSource` from BCD.  The linked MDN URL simply redirects to BaseAudioContext.  Furthermore, the data is invalid; the method was not present with the prefix in Chrome, and not present in Firefox/Edge/etc. in old versions.


sideshowbarker (approved):

vinyldarkscratch:
Marking as needing a release note since it removes a feature in BCD.

PR #6803, opened 2020-10-03
Opera Android 60 supports String.replaceAll

atjn:
This is in accordance with the addition in Chromium 85.

I have manually tested and verified support with the example on [MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/String/replaceAll), the same way i did in  #6676


sideshowbarker (approved):

sideshowbarker:
Thanks!

PR #6808, opened 2020-10-03
Force LF line ending for all text files

saschanaz:
- [x] Summarize your changes

This prevents prettier failure in environments where CRLF checkout is the default.

### DIFF
--- a/.gitattributes
+++ b/.gitattributes
@@ -1,1 +1,1 @@
-*.json text eol=lf
+* text=auto eol=lf


.gitattributes
```diff
@@ -1 +1 @@
-*.json text eol=lf
+* text=auto eol=lf
```
ddbeck, line 1: If I understand this change correctly, then if we ever add a binary file to the repository (e.g., `something.png`), then we'll have to update this config to _not_ wreck the line endings for that file. That's probably OK, but I'm mentioning the word "binary" so it'll turn up when I search the issues trying to fixing the problem when it happens. But we should probably comment just to be safe

```suggestion
# note: this may affect line endings for binary files and may need to be amended if we ever add one
* text=auto eol=lf
```

ddbeck (changes requested):
Thanks for this, @saschanaz. One suggestion (in a line a comment) and a question, to better document this change. The change itself seems fine to me.

The question: what does it look like when Prettier fails as a result of not having this configuration change? I trust that line endings are provoking problems, but I'd like to have some more detail about what we're fixing in the PR, to fully document what we're doing.

.gitattributes
saschanaz: I think `text=auto` automatically excludes such binaries, so maybe not needed?

ddbeck: You're right and I was mistaken. [This Stack Overflow question](https://stackoverflow.com/questions/29435156/what-will-text-auto-eol-lf-in-gitattributes-do) indicates that I was thinking of an older Git behavior. Reading https://github.com/prettier/prettier/issues/5528 was also instructive, which suggests that we should be using the config you propose in this PR. Perhaps not coincidentally, this is also what Prettier's own `.gitattributes` file looks like.

ddbeck (approved):
Thanks to https://github.com/mdn/browser-compat-data/pull/6808#discussion_r502673437, I'm withdrawing my earlier question and suggestion and merging as is. This appears to be a best practice for using Prettier. Thank you, @saschanaz! 🎉

Destop-Trang1801:
saschanaz:force-lf

Issue #6812, opened 2020-10-03
Remove status.experimental when nonstandard/deprecated


saschanaz:
I think nonstandard or deprecated items should not be considered as experimental. This should be an easier step for #1528.

PR #6813, opened 2020-10-03
Add linter for experimental+deprecated statuses

saschanaz:
- [x] Summarize your changes
- [x] Link to related issues or pull requests, if any: #6812


test/linter/test-experimental.js
```diff
@@ -0,0 +1,62 @@
...
+  const logger = {
+    /** @param {...unknown} message */
+    error: (...message) => {
+      errors.push(message.join(' '));
+    },
+  };
+
+  checkStatus(data, logger);
+
+  if (errors.length) {
```
saschanaz, line 48: This part is quite duplicated, there should be a helper function for this 🤔

test/linter/test-experimental.js
```diff
@@ -0,0 +1,62 @@
...
+ * @typedef {import('../../types').Identifier} Identifier
+ * @typedef {import('../utils').Logger} Logger
+ */
+
+/**
+ * @param {Identifier} data
+ * @param {Logger} logger
+ */
+function checkStatus(data, logger, path = []) {
+  const status = data.__compat?.status;
```
vinyldarkscratch, line 13: This is invalid syntax.  I think you were trying to do something like the following?

```suggestion
  if (!data.__compat) return;
  const status = data.__compat.status;
```

test/linter/test-experimental.js
```diff
@@ -0,0 +1,62 @@
...
+  if (status?.experimental) {
```
vinyldarkscratch, line 14:
```suggestion
  if (status && status.experimental) {
```

vinyldarkscratch (changes requested):

test/linter/test-experimental.js
saschanaz: Oops, please Node 14+ everywhere 😂😂

vinyldarkscratch: I just looked into this syntax more -- wow, that is so useful, I never knew this was a thing!

vinyldarkscratch:
Marking as "not ready" since we'll need to do some data updates before this linter passes.

saschanaz:
webextensions should probably be excluded as it has no spec.

saschanaz:
@ddbeck could you remove the rebase needed flag? 👀

foolip:
@saschanaz I have updated https://github.com/mdn/browser-compat-data/pull/11506 to fix experimental values, starting with the simplest cases, but it could be expanded to fix more and generally match the lint here.

Are you interested in picking this up again, given a willing reviewer? I could be that reviewer if so :)

saschanaz:
Will try 👍

foolip:
Oh, sorry, I was a bit confused. https://github.com/mdn/browser-compat-data/pull/11506 does fix some of the things that fail this lint by setting experimental to false, but I thought (until I looked just now) that this lint was also for incorrectly experimental status based on support. It's not, but that's fine, happy to review anyway.

I would like to review the fixes for the lint rule separately from the lint/fixer though. The changes will just keep conflicting as changes happen all the time :)

saschanaz:
> The changes will just keep conflicting as changes happen all the time :)
For this specific item, the only big conflicts were WebVR changes, and I think such change won't happen frequently 😉

queengooborg:
Even if the big conflicts involved data that isn't expected to change frequently, it is still easier for reviewers if the linter addition and data changes were kept in separate PRs.  By combining the two together, a reviewer now must review _both_ the data and the linter at once, which causes additional delays with correspondence and re-reviewing (not to mention, merge conflicts that form along the way since the PR sits around) before either part can be merged.  Even though the CI tests will fail until the data is updated, it is much better for it to fail to show that the linter is working in the first place.  🙂 

May you please separate the data changes into a separate PR?  They're mostly LGTM to me and I would love to merge them!

saschanaz:
>Even though the CI tests will fail until the data is updated, it is much better for it to fail to show that the linter is working in the first place.

Makes sense, opened #15913.

saschanaz:
I love #15931.

saschanaz:
#15913 made this green 🟩

test/linter/test-status-contradiction.js
```diff
@@ -0,0 +1,45 @@
...
+ * @param {Identifier} data
+ * @param {Logger} logger
+ */
+function checkStatus(data, logger, path = []) {
+  const status = data.__compat?.status;
+  if (status && status.experimental && status.deprecated) {
+    logger.error(
+      chalk`{red Unexpected simultaneous experimental and deprecated status in ${path.join(
+        '.',
+      )}}`,
```
queengooborg, line 21:
```suggestion
      )}}`,
      chalk`Run {bold npm run fix} to fix this issue automatically`,
```

test/linter/test-status-contradiction.js
```diff
@@ -0,0 +1,45 @@
...
+    checkStatus(data[member], logger, [...path, member]);
+  }
+}
+
+/**
+ * @param {Identifier} data The contents of the file to test
+ * @returns {boolean} If the file contains errors
+ */
+function testStatusContradiction(data) {
+  const logger = new Logger('Flag consistency');
```
queengooborg, line 37:
```suggestion
  const logger = new Logger('Status');
```

test/linter/test-status-contradiction.js
```diff
@@ -0,0 +1,45 @@
...
+    }
+    checkStatus(data[member], logger, [...path, member]);
+  }
+}
+
+/**
+ * @param {Identifier} data The contents of the file to test
+ * @returns {boolean} If the file contains errors
+ */
+function testStatusContradiction(data) {
```
queengooborg, line 36: Let's actually name this `testStatus` rather than `testStatusContradiction` (this applies to the filenames as well).  This will cause merge conflicts with other PRs that introduce status linters, but IMO they should all be merged together in the end anyways!

```suggestion
function testStatus(data) {
```

queengooborg (changes requested):

queengooborg (approved):
LGTM, thank you!

PR #5605, opened 2020-01-29
Add tips/hints to version test

queengooborg:
This PR introduces tips to the version linter for common mistakes contributors may make when creating a PR.  This was inspired by a [comment Daniel posted](https://github.com/mdn/browser-compat-data/issues/4833#issuecomment-533158735) on a suggestion to rename "safari_ios" to "ios":

> It seems reasonable to me that if you try to use a valid desktop version number for iOS, then the linter could remind you of the possibility that you're following the wrong versioning scheme.

Well, here you go, @ddbeck -- one fresh addition of code hot off the grill!  😄

PR #6814, opened 2020-10-03
Add Logger class for linters

saschanaz:
- [x] Summarize your changes

Linters have duplications in error logging logic, and this patch tries to reduce it.

### DIFF
--- a/test/linter/test-browsers.js
+++ b/test/linter/test-browsers.js
@@ -1,10 +1,10 @@
 'use strict';
 const path = require('path');
 const chalk = require('chalk');
+const { Logger } = require('./utils.js');
 
 /**
  * @typedef {import('../../types').Identifier} Identifier
- * @typedef {import('../utils').Logger} Logger
  */
 
 /** @type {Record<string, string[]>} */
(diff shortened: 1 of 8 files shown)


vinyldarkscratch (commented):
This is a great way to de-dupe the code, thank you!  I'm thinking that we can take this even further to consolidate the code.  Here's what I'm thinking:

1) We apply this new Logger class to all of the test scripts (that is, throw this class into `test/utils.js` and use it in `test-format`, `test-migrations`, etc.).  As a part of this, we could also throw in a `tip` argument to the `error` function to include the blue tip like in `test-format` and proposed by #5605. (Written as #6852)
2) We move the `emit()` functionality to `lint.js`, so that we no longer perform our `console.error` hook/replacement.  I'm thinking of a function like `getErrors()` that returns the errors array (or a dictionary like `{name: '', errors: []}`).
3) We remove the `hasABCErrors` variables and simplify them into a dictionary, `errors`, structured like `{[name]: [error1, error2, ...]}` -- we'd only include keys/values for tests which returned errors, so that if there's no errors, the dictionary is empty.  `fileHasErrors` would simply be `Object.keys(errors).length`, and then we can perform the `emit()` console logging right there.

Of course, this can all be saved for a follow-up PR ~~you may have inspired me to want to write~~!

saschanaz:
Oh wow, didn't know `lint.js` is overriding `console.error`. I guess that's a good idea and maybe as a followup 👀

ddbeck (approved):
I love this refactor. It improves one thing really clearly and paves the way for more. Thank you, @saschanaz! 🎉

PR #6818, opened 2020-10-04
Add/update Firefox data for Document API

vinyldarkscratch:
This PR updates the Firefox data for the Document API based upon manual testing to achieve more real data for the Document API.  This also includes a little bit of irrelevant flag removal to simplify the data.

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -11259,11 +11259,11 @@
             },
             "firefox": {
               "version_added": "1",
-              "version_removed": true
+              "version_removed": "10"
             },
             "firefox_android": {
               "version_added": "4",
-              "version_removed": true
+              "version_removed": "10"
             },
             "ie": {
               "version_added": null
@@ -11309,11 +11309,11 @@
             },
             "firefox": {
               "version_added": "1",
-              "version_removed": true
+              "version_removed": "10"
             },
             "firefox_android": {
               "version_added": "4",
-              "version_removed": true
+              "version_removed": "10"
             },
             "ie": {
               "version_added": null
@@ -11359,11 +11359,11 @@
             },
             "firefox": {
               "version_added": "1",
-              "version_removed": true
+              "version_removed": "10"
             },
             "firefox_android": {
               "version_added": "4",
-              "version_removed": true
+              "version_removed": "10"
             },
             "ie": {
               "version_added": null
--- a/api/Document.json
+++ b/api/Document.json
@@ -5933,11 +5933,11 @@
             },
             "firefox": {
               "version_added": "1",
-              "version_removed": true
+              "version_removed": "6"
             },
             "firefox_android": {
               "version_added": "4",
-              "version_removed": true
+              "version_removed": "6"
             },
             "ie": {
               "version_added": false
@@ -11109,11 +11109,11 @@
             },
             "firefox": {
               "version_added": "1",
-              "version_removed": true
+              "version_removed": "6"
             },
             "firefox_android": {
               "version_added": "4",
-              "version_removed": true
+              "version_removed": "6"
             },
             "ie": {
               "version_added": false
--- a/api/Document.json
+++ b/api/Document.json
@@ -8711,7 +8711,7 @@
               "version_added": false
             },
             "firefox": {
-              "version_added": false,
+              "version_added": true,
               "notes": "Available only to <a href='https://developer.mozilla.org/docs/Mozilla/Tech/XUL'>XUL documents</a>."
             },
             "firefox_android": {
@@ -10186,7 +10186,7 @@
               "version_added": false
             },
             "firefox": {
-              "version_added": false,
+              "version_added": true,
               "notes": "Available only to <a href='https://developer.mozilla.org/docs/Mozilla/Tech/XUL'>XUL documents</a>."
             },
             "firefox_android": {
--- a/api/Document.json
+++ b/api/Document.json
@@ -159,10 +159,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": true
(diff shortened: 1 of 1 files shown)


api/Document.json
```diff
@@ -5988,10 +6006,12 @@
               "version_added": false
             },
             "firefox": {
-              "version_added": false
+              "version_added": "1",
+              "version_removed": true
```
foolip, line 6010: Can you find out and fill in the removal version? I've checked that `document.height` wasn't in Firefox 40, which was 5 years ago, so the original state of this entry was more useful in practice than this, even if it failed to note that this API was once supported.

More generally it would be good to not introduce new trues, unless finding the real version is too hard.

api/Document.json
```diff
@@ -6819,10 +6839,10 @@
               "version_added": false
             },
             "firefox": {
-              "version_added": false
+              "version_added": "4"
```
foolip, line 6842: Yep, `document.mozSetImageElement` is still a thing.

api/Document.json
```diff
@@ -6963,10 +6983,10 @@
               "version_added": false
             },
             "firefox": {
-              "version_added": true
+              "version_added": "10"
```
foolip, line 6986: Some event handler attributes added starting here. The form sensible groups, and the versions are way in the past but still plausible, so I haven't tried to verify further.

api/Document.json
```diff
@@ -8765,7 +8785,7 @@
               "version_added": false
             },
             "firefox": {
-              "version_added": true,
+              "version_added": false,
```
foolip, line 8788: Can you revert this change? (This is `document.popupNode`.) The note makes it clear this isn't web exposed. I don't think we should have this in BCD, but it's a different kind of change to the rest in this PR.

api/Document.json
```diff
@@ -10240,7 +10260,7 @@
...
```
foolip, line 10263: This (`document.tooltipNode`) is also a XUL-only API, like `document.popupNode`. Can you revert this change?

api/Document.json
```diff
@@ -11162,10 +11182,12 @@
               "version_added": false
             },
             "firefox": {
-              "version_added": false
+              "version_added": "1",
+              "version_removed": true
```
foolip, line 11186: The same comment as for `document.height` applies here.

api/Document.json
```diff
@@ -11310,10 +11332,12 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": false
+              "version_added": "1",
+              "version_removed": true
```
foolip, line 11336: This seems right per https://bugzilla.mozilla.org/show_bug.cgi?id=687426, but can you find the removal version and add here?

The same goes for the follow document.xml* entries, and they might not have been removed together.

foolip (commented):
Wohoo, a big change, and just a few small things to fix!

api/Document.json
```diff
@@ -314,10 +314,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": true
+              "version_added": "24"
```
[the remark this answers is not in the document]
vinyldarkscratch: In Firefox 23, `document.all` is indeed just `undefined` and not an object.  It also doesn't show up in the console's autocomplete suggestions.

api/Document.json
```diff
@@ -5319,6 +5334,9 @@
               }
             ],
             "firefox_android": [
+              {
+                "version_added": "79"
```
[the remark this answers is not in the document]
vinyldarkscratch: Indeed -- unfortunately we will not be able to remove most of this flagged data, but we can strip some of it out!

api/Document.json
vinyldarkscratch: I planned to fill those versions out but I guess I forgot to, whoops!

api/Document.json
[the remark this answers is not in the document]
foolip: Thanks for confirming!

foolip (approved):

api/Document.json
```diff
@@ -4629,24 +4631,11 @@
...
-                    "type": "preference",
-                    "name": "layout.css.font-loading-api.enabled",
-                    "value_to_set": "true"
-                  }
-                ]
-              }
-            ],
+            "firefox": {
+              "version_added": "41"
+            },
```
ddbeck, line 4636: Hey @foolip and @vinyldarkscratch — I'm looking back through some recently merged PRs, just to get a sense of how things are going. Since the flag removal guideline is new, I'm keeping on eye on that sort of thing, to make sure it's working out as we'd hoped.

It looks like this was removed to be consistent with Android. I'm guessing this is a reverse mirroring, where Firefox for Android data overwrote the desktop data. Is that the whole story?

More broadly, it'd be helpful to me if PRs would at least note that some data was removed and what the motivation was for it, at least in the short term. That might spare me some having to inquire about already-merged PRs. 😄

foolip: This came from me pushing back against adding more old flag data in https://github.com/mdn/browser-compat-data/pull/6818#discussion_r499510959 and suggesting removing the existing data instead.

@vinyldarkscratch when we remove flag data going forward, can you update the PR description? I've recently been using the PR description as the commit message, so it would then show up there too.

ddbeck: Thank you for the explanation on this. And thanks for filling me on how this happened. Seeing the comment in context, it does strike me as… uninteresting to replicate old flag data. This is fine. I do want to continue to follow our use of the new guideline though—thank you for any effort to keep me in the loop on future removals. 👍

(My inbox is in a dire state apparently. Sorry for only just getting to this now.)

PR #6819, opened 2020-10-04
Add/update IE/EdgeHTML data for Document API

vinyldarkscratch:
This PR updates the Internet Explorer and pre-Blink Edge data for the Document API based upon manual testing to achieve more real data for the Document API.

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -165,7 +165,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": true
+              "version_added": "4"
             },
             "opera": {
               "version_added": true
@@ -263,7 +263,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": null
+              "version_added": "4"
             },
             "opera": {
               "version_added": true
@@ -320,7 +320,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": null
+              "version_added": "4"
             },
             "opera": {
               "version_added": true
@@ -376,7 +376,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": null
+              "version_added": "4"
             },
             "opera": {
               "version_added": true
@@ -650,7 +650,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": null
+              "version_added": "4"
             },
             "opera": {
               "version_added": true
@@ -700,7 +700,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": null
+              "version_added": "4"
             },
             "opera": {
               "version_added": true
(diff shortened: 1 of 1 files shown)


api/Document.json
```diff
@@ -1528,7 +1528,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": null
+              "version_added": "9"
```
foolip, line 1531: This surprised me, but indeed it seems like `document.createAttributeNS` isn't in IE8. The documents we're testing in are HTMLDocument instances, and it's possible `createAttributeNS` was on XMLDocument, but IE8 actually doesn't seem to have XMLDocuemnt.

https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/platform-apis/ff975561(v=vs.85) seems to support this change.

api/Document.json
```diff
@@ -2988,7 +2988,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": true
+              "version_added": false
```
foolip, line 2991: This is consistent with `document.documentURI` being added in Edge 17.

api/Document.json
```diff
@@ -7660,7 +7660,7 @@
               "version_added": "56"
             },
             "ie": {
-              "version_added": "10"
+              "version_added": false
```
foolip, line 7663: This is `document.onvisibilitychange`, where the data claimed that it was in IE 10.

https://developer.mozilla.org/en-US/docs/Web/API/Document/visibilitychange_event#Browser_compatibility (just edited) says the event itself was in IE 10. I'm not sure which is correct, but can you back out this change and update both separately? The event can't be tested by mdn-bcd-collector, and it's just as well to separate out such changes.

api/Document.json
```diff
@@ -7769,7 +7769,7 @@
               "version_added": false
             },
             "ie": {
-              "version_added": true
+              "version_added": false
```
foolip, line 7772: I was a bit skeptical of this, so tested it separate in https://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8540 using IE11. But `document.origin` seems to not have been in IE, so this change is correct.

api/Document.json
```diff
@@ -10150,7 +10150,7 @@
               }
             ],
             "ie": {
-              "version_added": true
+              "version_added": false
```
foolip, line 10153: Yeah, the true there wasn't plausible, given that Edge didn't have `document.timeline`.

api/Document.json
vinyldarkscratch: It looks like IE 10 didn't support `onvisibilitychange`, however it supported the event through `addEventListener`.  Reverting these changes and cherry-picking in a separate PR.

vinyldarkscratch: Cherry-picked as #6833.

foolip (approved):

PR #6820, opened 2020-10-04
Add/update Safari data for Document API

vinyldarkscratch:
This PR updates the Safari data for the Document API based upon a combination of manual testing and collector results to achieve more real data for the Document API.

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -7681,10 +7681,10 @@
               "notes": "Doesn't fire the <code>visibilitychange</code> event when the browser window is minimized, nor when <code>hidden</code> is set to true."
             },
             "safari": {
-              "version_added": "10.1"
+              "version_added": "7"
             },
             "safari_ios": {
-              "version_added": "10.3"
+              "version_added": "7"
             },
             "samsunginternet_android": {
               "version_added": "2.0"
--- a/api/Document.json
+++ b/api/Document.json
@@ -5037,11 +5037,11 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true,
+              "version_added": "6",
               "prefix": "webkit"
             },
             "safari_ios": {
-              "version_added": true,
+              "version_added": "6",
               "prefix": "webkit"
             },
             "samsunginternet_android": [
@@ -7273,11 +7273,11 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true,
+              "version_added": "5.1",
               "alternative_name": "onwebkitfullscreenchange"
             },
             "safari_ios": {
-              "version_added": true,
+              "version_added": "5.1",
               "alternative_name": "onwebkitfullscreenchange"
             },
             "samsunginternet_android": {
@@ -7378,11 +7378,11 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true,
+              "version_added": "6",
               "alternative_name": "onwebkitfullscreenerror"
             },
             "safari_ios": {
-              "version_added": true,
+              "version_added": "6",
               "alternative_name": "onwebkitfullscreenerror"
             },
             "samsunginternet_android": {
--- a/api/Document.json
+++ b/api/Document.json
@@ -272,10 +272,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "1.2"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "1"
             },
             "samsunginternet_android": {
               "version_added": "9.0",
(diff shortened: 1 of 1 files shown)


foolip:
@vinyldarkscratch this is a very big PR, can you separate out the Safari changes? Then I need not worry about overlooking them when focusing mostly on Chromium in this PR.

api/Document.json
```diff
@@ -1233,10 +1233,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": "6"
```
foolip, line 1236: This is `document.contains` and doesn't match https://developer.mozilla.org/en-US/docs/Web/API/Node/contains. Is that entry wrong?

api/Document.json
```diff
@@ -4320,10 +4322,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "10.1"
```
foolip, line 4325: This is `document.exitPointerLock()`, added much later than most of the things in this PR, but seems right in terms of when that API was introduced.

api/Document.json
```diff
@@ -5035,10 +5037,12 @@
               "version_added": true
             },
             "safari": {
-              "version_added": null
+              "version_added": true,
```
foolip, line 5040: Can you update this with when `document.webkitFullscreenEnabled` was introduced? I've just done https://github.com/foolip/mdn-bcd-collector/pull/589 to make sure we generate the test for it.

api/Document.json
```diff
@@ -6978,10 +6984,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": false
```
foolip, line 6987: This is `onafterscriptexecute`. To confirm, I also grepped a WebKit checkout and its history for "afterscriptexecute" to ensure the event itself hasn't been supported. No matches 👍

api/Document.json
```diff
@@ -7026,10 +7032,10 @@
...
```
foolip, line 7035: Also no matches in WebKit for "beforescriptexecute" 👍

api/Document.json
```diff
@@ -7267,10 +7273,12 @@
               "version_added": true
             },
             "safari": {
-              "version_added": null
+              "version_added": true,
```
foolip, line 7276: Can you determine the version for this? I've sent https://github.com/foolip/mdn-bcd-collector/pull/590 to create the test for it.

api/Document.json
```diff
@@ -7370,10 +7378,12 @@
...
```
foolip, line 7381: Also covered by https://github.com/foolip/mdn-bcd-collector/pull/590.

api/Document.json
```diff
@@ -7671,14 +7681,10 @@
               "notes": "Doesn't fire the <code>visibilitychange</code> event when the browser window is minimized, nor when <code>hidden</code> is set to true."
             },
             "safari": {
-              "version_added": "7",
```
foolip, line 7674: Why is the old entry being removed? I guess it's wrong, that `onvisibilitychange` wasn't exposed before Safari 10.1, but perhaps the "visibilitychange" event was still fired?

If sorting this out will also involve updating the `visibilitychange_event` entry, can you break the change out of this PR?

foolip (commented):
Great stuff, just a few small nits before this can be merged!

api/Document.json
vinyldarkscratch: The MDN URL for this feature is a 404, so I'm willing to bet this is a duplicate entry.  Although, the data should match, since Document inherits from Node...

vinyldarkscratch: Opened #6841 for this!

api/Document.json
vinyldarkscratch: Yeah, this is a similar situation to #6833.  I can split this out and add the note like the other PR?

foolip: Yes, that would be great! It's the only remaining blocker for this PR to land.

vinyldarkscratch: Cherry-picked as #6849!

foolip:
@vinyldarkscratch oops, this conflicts now, I left it sitting for too long. Can you resolve the conflicts?

api/Document.json
```diff
@@ -7268,10 +7276,12 @@
               "version_added": true
             },
             "safari": {
-              "version_added": null
+              "version_added": "6",
+              "alternative_name": "onwebkitfullscreenerror"
```
foolip, line 7280: Huh, so `onwebkitfullscreenerror` was added after `onwebkitfullscreenchange`: https://trac.webkit.org/changeset/104838/webkit

api/Document.json
foolip: Can you also revert these changes?

foolip (approved):
Just need to revert the onvisibibilitychange entry. I'll see if I can do that myself and push to the branch to get this merged.

api/Document.json
foolip: I've reverted it now.

PR #6821, opened 2020-10-04
Refactor traverse script

queengooborg:
This PR refactors the traverse script to improve legibility and use more traditional flags and options, based upon @ddbeck's [suggestions](https://github.com/mdn/browser-compat-data/pull/6689#issuecomment-700785154).  This also includes functionality presented by #6689 (no `--filter` argument will traverse all features regardless of their value).


ddbeck:
This is really cool, @vinyldarkscratch. Thank you for getting started on this!

I tried checking out your branch locally and playing with the new CLI. It seems really promising, but I encountered some odd behavior. One part I was able to figure out: the `--browser`, `--filter`, and `--depth` options need `nargs: 1` (so they don't consume additional arguments).

But the other part I haven't been able to figure out and I was hoping you might do some more testing and investigation. I'm not sure if it's from the other refactoring that happens here* or if it's something with the way yargs is setup, but this seems to struggle with the variadic folder argument. If I pass it one folder, it seems to work fine, but I get weird results with two. Here's an example:

On `master` (which I trust and spot-checked to be correct):
```shell
$ npm run traverse -- chrome css,api
> mdn-browser-compat-data@1.0.40 traverse /Users/ddbeck/Code/Mozilla/browser-compat-data
> node scripts/traverse "chrome" "css,api"

api.BaseAudioContext
… a bunch of other features …
api.XSLTProcessor.transformToFragment
1588
```

On your branch:
```
$ npm run traverse -- --nonreal --browser=chrome css api
> mdn-browser-compat-data@1.0.40 traverse /Users/ddbeck/Code/Mozilla/browser-compat-data/.worktrees/6821-scripts-traverse-refactor
> node scripts/traverse "--nonreal" "--browser=chrome" "css" "api"

0
```

A couple of assumptions here:

* `master`'s `npm run traverse -- chrome css,api` is equivalent to your branch's `npm run traverse -- --nonreal --browser=chrome css api`
* I patched `traverse.js` to use `nargs: 1` for the `browser` option

You might have a better idea of what's going on here. 😄

\* in the future, it might help to break refactors like this into smaller commits or separate PRs—then I could step through changes to figure out what was going on a little better

vinyldarkscratch:
It seems that `yargs` doesn't actually allow for a positional `array`, so it refuses to take more than just the first one.  Unfortunately, I'm not finding much that tells me how I can effectively catch all non-option arguments...

vinyldarkscratch:
Issues are now resolved!  After digging in some more, I found out that you're not supposed to include the optional flags in the usage, because then, it turns them into positional arguments as well.

docs/testing.md
```diff
@@ -17,12 +17,12 @@ To see how changes will affect the statistics of real (either `false` or a versi
 
 ## Traverse
 
-To find all the entries that are non-real, or of a specified value, you can run `npm run traverse <browser> [folder] [value]`.
+To find all the entries that are non-real, or of a specified value, you can run `npm run traverse [folder] -- [-b/--browser browsername, ...] [-f/--filter value, ...] [--nonreal] [-d/--depth depth, ...]`.
 
-The browser may be any single browser defined in the [`browsers/` folder](https://github.com/mdn/browser-compat-data/blob/master/browsers/).
+The folder argument may be a comma-separated list of folder to check entries in. By default, the script will traverse all folders.
```
ddbeck, line 22: I propose we trim this a bit and let `--help` do the talking.

What do you think of this?

```suggestion
By default, the script will traverse and print the dotted path to every feature. One or more positional arguments can limit the traversal to a specific folder (for example, `api`). Additional options can limit the features listed to features with data matching specific browser and version values (for example, `--non-real`).

Run `npm run traverse -- --help` for a complete list of options and examples.
```

Then delete everything after this.

docs/testing.md
```diff
@@ -17,12 +17,12 @@ To see how changes will affect the statistics of real (either `false` or a versi
...
+The `-f` or `--filter` argument may be any value accepted by `version_added` or `version_removed`. This argument may be repeated to test multiple values. By default, the script will traverse all features regardless of their value. The `--nonreal` argument may be included as a convenience alias for `-f null -f true`.
 
-For example, to search for all Safari entries that are non-real, run `npm run traverse safari`. To search for all WebView entries that are marked as `true` in `api` and `javascript`, run `npm run traverse webview_android api,javascript true`. To search for all Firefox entries supported since `10` across all folders, run `npm run traverse firefox all 10`.
+Examples: to search for all Safari entries that are non-real, run `npm run traverse -- -b safari --noreal`. To search for all WebView entries that are marked as `true` in `api` and `javascript`, run `npm run traverse api,javascript -- -b webview_android -f true`. To search for all Firefox entries supported since `10` across all folders, run `npm run traverse -- -b firefox -f 10`.
```
ddbeck, line 28: This part is still really useful, but we could put it on the command line with `yargs`:

```js
      .example(
        'npm run traverse -- --browser=safari --non-real',
        'Find all features containing non-real Safari entries',
      )
      .example(
        'npm run traverse -- -b webview_android -f true`,
        'Find all features marked as true for WebVeiw',
      )
      .example(
        'npm run traverse -- -b firefox -f 10`,
        'Find all features marked as supported since Firefox 10',
      )
```

We should probably double-check this is both what's supposed to happen and what actually happens though…

scripts/traverse.js
```diff
@@ -2,89 +2,143 @@
...
+      )
+      .example(
+        'npm run traverse -- -b firefox -f 10',
+        'Find all features marked as supported since Firefox 10',
+      );
   },
 );
 
-function traverseFeatures(obj, depth, identifier) {
+function traverseFeatures(obj, browsers, values, depth, identifier) {
```
foolip, line 57: What do you think about exposing `traverseFeatures` for other scripts to use when not invoking this script directly? That would also make it easy to write tests for this.

scripts/traverse.js
```diff
@@ -2,89 +2,145 @@
...
```
foolip, line 57: How about a `iterateFeatures` using https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Statements/function* that yields one entry a a time, and this `traverseFeatures` just uses `Array.from()` to return an array? The code that I have in mind will usually just want to iterate entries, having them all in an array would just be extra lines of code.

ddbeck: I like this `iterateFeatures` idea too! 👍

vinyldarkscratch: I'm not all that familiar with generator functions (this would be my first time writing one), and I'm not too certain on how to implement it into this code.  Having a function like that sounds like a good idea, though!  I'll see what I can whip up (and I'll talk about it a bit during my next meeting with foolip).

foolip: Oops, we didn't end up talking about this when we met. It's basically a structure like this:
```js
function* iterateFeatures () {
  for (let i = 0; i < 10; i++) {
    yield i;
  }
}
```

ddbeck: Hey @vinyldarkscratch, do you have questions about this? Or is this just waiting on you having time to complete it? If it's the former, let me know how I can help on this—I can write up a fuller generator example, if you'd like. If it's the latter, there's no rush here. Thank you!

vinyldarkscratch: Slip of the mind on this one, actually -- I've been focusing my efforts on getting the real values up (along with the mdn-bcd-collector project, getting its coverage up more).  Spent a little bit of time on this again, and I think I got it!  I'm pushing a commit that should get the desired result, let me know how it looks?

queengooborg:
I'm going to merge this since the main reviewer for BCD infra is out of office and nobody else has reviewed this in two years.

scripts/traverse.js
```diff
@@ -15,124 +15,169 @@ import bcd from '../index.js';
...
+        )
+        .example(
+          'npm run traverse -- -b firefox -f 10',
+          'Find all features marked as supported since Firefox 10',
+        );
     },
   );
 
-  main(argv.folder, argv.value, argv.browser, argv.depth);
+  const features = main(argv.folder, argv.value, argv.browser, argv.depth);
```
foolip, line 178: The order of `argv.value` and `argv.browser` here are flipped, and that's one cause of https://github.com/mdn/browser-compat-data/issues/16505. But even fixing that none of the filtering seems to work.

Issue #6738, opened 2020-09-21
Interim project goals


ddbeck:
## Background

Right now, MDN doesn't have full-time staff focused on managing BCD ([read more about this on Hacks](https://hacks.mozilla.org/2020/08/an-update-on-mdn-web-docs/)). As a contractor, this has led to me being a mostly-solo caretaker for BCD, while MDN revises its goals and establishes long-term project plans.

In the interest of transparency and to calibrate expectations, I've decided to lay out my near-term goals for BCD and what I hope that means for the project.

## Current goals

In rough order of priority, this is where **my** focus is going to be:

1. Provide project continuity. This means maintaining high expectations for data quality, issuing releases on most Thursdays, and preserving a welcoming, convivial project culture.
2. Share the work. Answer questions and document processes that help others contribute to BCD and work on reducing the "lottery factor" for continuing efforts on BCD.
3. Review routine data pull requests and provide timely feedback to contributors.

I expect other contributors to have their own goals and aspirations—to scratch their own itch, as it were—and I want to support that. In short, I want to help you to help keep the project going.

## A principle: don't make any difficult-to-revert decisions

Points one and two are slightly in tension with each other: sometimes making decisions means breaking with the past. To the extent that I'm able to, I plan to make decisions that permit us to experiment, try things, and change our mind.

The principle that's going to guide me is this: favor decisions that are easy to revert. This may lead to some untidiness in data structure, or writing more verbose notes, or asking contributors for more issues or documentation, but I hope it will ensure that BCD's future can fit into whatever's next for MDN and the needs of BCD’s other consumers.

That said, I don't expect this exact posture to remain forever, so please expect refinements and changes as we go.

## Thanks

I know several people have stepped up their participation here recently, or continued your ongoing efforts to contribute to BCD. Thank you, thank you, thank you. This is essential to keeping BCD healthy.

ddbeck:
CCing @vinyldarkscratch, @jpmedley, @sideshowbarker, @foolip. You're the most active Peers right now, so I wanted to first emphasize my thanks to you. Plus I wanted to make sure you saw what my focus was going to be in the near term. Thank you!

foolip:
Thank you @ddbeck, it's been a pleasure working in the project so far! The principle of avoiding difficult-to-revert decisions is one that I can easily get behind right now, and is actually one that might help resolve certain long-standing issues, at least tentatively.

For my own involvement in the project, I would say my goals are twofold:
[...]

PR #6822, opened 2020-10-04
Migration 008: add script for removing redundant flags

vinyldarkscratch:
This PR adds a migration script to go along with the new irrelevant flag removal guidelines added in #6670 by doing the following:

- Removes flag data in features added by default over two years ago
- Removes flag data for flags removed over two years ago

Note: since the migration script is time-based, the test script will probably fail in the future.  I didn't want to introduce a new dependency just to mock the time, however.

### DIFF
--- a/scripts/migrations/008-remove-redundant-flags.js
+++ b/scripts/migrations/008-remove-redundant-flags.js
@@ -1,0 +1,148 @@
+#!/usr/bin/env node
+/* Any copyright is dedicated to the Public Domain.
+ * http://creativecommons.org/publicdomain/zero/1.0/ */
+
+'use strict';
+const fs = require('fs');
+const path = require('path');
+const { platform } = require('os');
+
+const bcd = require('../..');
+
+/** Determines if the OS is Windows */
+const IS_WINDOWS = platform() === 'win32';
+
+let twoYearsAgo = new Date(Date.now());
+twoYearsAgo.setFullYear(twoYearsAgo.getFullYear() - 2);
+
+const removeRedundantFlags = (key, value) => {
+  if (key === '__compat') {
+    for (const [browser, supportData] of Object.entries(value.support)) {
+      if (supportData !== undefined && Array.isArray(supportData)) {
+        const result = [];
+
+        const simpleStatement = supportData.find(statement => {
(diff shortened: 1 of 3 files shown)


ddbeck:
I've got some bad news about this PR, @vinyldarkscratch. 😬 

I don't see our guideline, as written, sustaining a change like this. The guideline has two qualifiers that I think preclude doing a migration for this kind of data:

> The removal of the support statement must not create an ambiguous gap or void in the data for that browser

> These conditions represent minimum requirements for the removal of valid flag data; other considerations may result in flag data continuing to be relevant, even after the guideline conditions are met.

The first might be solvable (though I'm not sure we have a robust definition of "ambiguous" data), but the second definitely isn't—and that's on purpose. I might've been more explicit about my intent when initially proposing the guideline in #6670, but I didn't actually want to sanction automated removal just yet. I had a few motives:

- I didn't want to do any large-scale changes to the data that would be difficult to undo (see #6738). A migration like this would become quickly unrevertable.
- Removing old flags may lead to cascading removals (e.g., if a flag removal resulted in an all-`false` feature, then we'd also need to remove that feature, depending on its development status). En masse, this might be disruptive to consumers (necessitating a version increment). It would also be a lot of work that I didn't want to take on while also managing the `mdn-` to `@mdn/` migration.
- I wanted to facilitate the removal of dubious data or data that would be difficult to reconcile, not scrub our data set of old flags generally. _High-quality_ historic flag data may still be of interest to spec authors and the like and I didn't want to foreclose on that audience.

These considerations, taken together, point toward closing the PR. I'm annoyed at myself for letting you put in all this work for little gain. To avoid this in the future, we should probably begin to follow [our own advice](https://github.com/mdn/browser-compat-data/blob/master/docs/migrations.md#step-1-announce-your-intent) more closely and use issues to declare intent on this sort of thing before implementing it (I've been an offender too, on this count).

I'm sorry to have to close this, especially knowing that you've put some work into this. I am very grateful that you've put in the effort here and that you've given me an opportunity to be more clear about the aims of the guideline. Thank you.

vinyldarkscratch:
I totally understand, no worries!  I'll keep this code around if we ever want to look at it again in the future.  😉

PR #6825, opened 2020-10-04
Add/update Safari data for Window API

vinyldarkscratch:
This PR updates the Safari data for the Window API based upon a combination of manual testing and collector results to achieve more real data for the Window API.

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -1585,10 +1585,10 @@
               "version_added": "24"
             },
             "safari": {
-              "version_added": "3"
+              "version_added": "6.1"
             },
             "safari_ios": {
-              "version_added": "1"
+              "version_added": "6.1"
             },
             "samsunginternet_android": {
               "version_added": "3.0"
--- a/api/Window.json
+++ b/api/Window.json
@@ -892,10 +892,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "1"
             },
             "safari_ios": {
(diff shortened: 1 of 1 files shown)


api/Window.json
```diff
@@ -1585,10 +1585,10 @@
               "version_added": "24"
             },
             "safari": {
-              "version_added": "6.1"
+              "version_added": "3"
```
foolip, line 1588: This does not seem plausible, https://w3c.github.io/webcrypto/ is too new for this. http://mdn-bcd-collector.appspot.com/tests/api/Window/crypto does show true in Safari 5, but this seems like a flaw in our tooling. Can you check it out?

I'll hold off reviewing the rest, since there are more 3s added in this PR, and now I distrust them all until this is explained :)

foolip (changes requested):
Something has gone wrong :)

api/Window.json
vinyldarkscratch: Alright, so I dug deeper into this and found out what's going on.  Apparently, the `crypto` variable _is_ in the `window` instance in Safari 3, but it is set to `undefined`.  I guess we'll also need to configure our tests to check if the variable is not `undefined`...

Screen Shot 2020-10-08 at 09 16 13

foolip: Also discussed in https://github.com/mdn/browser-compat-data/pull/6838#discussion_r501844503

foolip (approved):

PR #6827, opened 2020-10-05
Updates to allow-downloads

rachelandrew:
The `sandbox="allow-downloads"` seems to be shipping in 82 not 81. https://bugzilla.mozilla.org/show_bug.cgi?id=1656212 so I have updated that while working on https://github.com/mdn/sprints/issues/3727


ddbeck (approved):
Looks good. Thank you! 🎉

PR #6828, opened 2020-10-05
Add/update IE/EdgeHTML data for Navigator API

vinyldarkscratch:
This PR updates the IE and EdgeHTML data for the Navigator API using the mdn-bcd-collector project, aided with a little manual testing, to achieve more real data for the Navigator API.

Notes:
- `productSub` had a note for IE that said it "always returns `undefined`".  Since that's basically the same behavior of an unimplemented feature, it makes no sense to mark it as `true`.
- `productSub` __does__ return the same value in Edge 12 as it does in Chrome/Safari.

Tests used: https://mdn-bcd-collector.appspot.com/tests/api/Navigator

### DIFF
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -441,7 +441,7 @@
               "version_added": "66"
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "79"
             },
             "firefox": {
               "version_added": "63"
@@ -489,7 +489,7 @@
               "version_added": "38"
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "79"
             },
             "firefox": {
               "version_added": true,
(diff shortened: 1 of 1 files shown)


api/Navigator.json
```diff
@@ -1713,8 +1714,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": true,
-              "notes": "Always returns <code>undefined</code>."
+              "version_added": false
```
foolip, line 1717: Confirmed with http://mdn-bcd-collector.appspot.com/tests/api/Navigator/productSub, which amounts to `'productSub' in navigator`.

If there had been a getter there, but it always returned undefined, that would have been worth a note, since 'productSub' in navigator` would then have been true, which would mess with feature detection.

foolip (approved):

PR #6830, opened 2020-10-05
Add/update Firefox data for Navigator API

vinyldarkscratch:
This PR updates the Firefox data for the Navigator API using the mdn-bcd-collector project, aided with a little manual testing, to achieve more real data for the Navigator API.

Tests used: https://mdn-bcd-collector.appspot.com/tests/api/Navigator

### DIFF
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -1550,7 +1550,7 @@
               "notes": "The preference <code>\"general.oscpu.override\"</code> can be used to set a value to be returned instead of the true CPU description. The preference setting is ignored for calls made by privileged code, which continue to get the actual CPU description."
             },
             "firefox_android": {
-              "version_added": true
+              "version_added": "4",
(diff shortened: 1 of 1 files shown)


api/Navigator.json
```diff
@@ -1448,10 +1448,10 @@
               "version_added": false
             },
             "firefox": {
-              "version_added": "71"
+              "version_added": false
```
foolip, line 1451: This is `navigator.mediaSession`, and in my testing Firefox 81 on macOS it's not there. It is in Firefox Developer Edition, so probably behind some flag. But the existing data is wrong, glad to fix some false positives too!

api/Navigator.json
```diff
@@ -2201,7 +2202,7 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": true
+              "version_added": "1"
             },
             "firefox_android": {
               "version_added": true
```
foolip, line 2208: I see firefox_android wasn't updated here. That's OK, we'll get to it later, but might point to a problem with some script or method.

foolip (approved):

PR #6831, opened 2020-10-05
Fix data in Console API using mdn-bcd-collector

vinyldarkscratch:
This PR uses the data collected from the mdn-bcd-collector project to fix a few data inaccuracies in the Console API.  Results were manually verified for accuracy.

Tests used: https://mdn-bcd-collector.appspot.com/tests/api/Console

### DIFF
--- a/api/Console.json
+++ b/api/Console.json
@@ -1396,7 +1396,7 @@
               "version_added": "18"
             },
             "edge": {
-              "version_added": "16"
+              "version_added": "12"
             },
             "firefox": {
               "version_added": "39"
@@ -1405,7 +1405,7 @@
               "version_added": "39"
             },
             "ie": {
-              "version_added": false
+              "version_added": "11"
             },
             "nodejs": {
               "version_added": "8.0.0",
--- a/api/Console.json
+++ b/api/Console.json
@@ -123,10 +123,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "48"
+              "version_added": "39"
             },
             "firefox_android": {
-              "version_added": "48"
+              "version_added": "39"
             },
             "ie": {
               "version_added": "8"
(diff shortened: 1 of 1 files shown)


api/Console.json
```diff
@@ -1403,7 +1405,7 @@
               "version_added": "39"
             },
             "ie": {
-              "version_added": "11"
+              "version_added": false
```
saschanaz, line 1408: I think this is supported.

image

foolip: @vinyldarkscratch did something go wrong with the methodology here? Can the other changes be trusted?

saschanaz: AFAIK IE only defines console functions when DevTools is opened, so maybe that's the cause.

vinyldarkscratch: If I'm correct, this was an issue of testing IE on Windows 7 vs. Windows 10 -- in BrowserStack and SauceLabs, the IE versions are ever so slightly behind on updates.  I've made sure that in the future, the Selenium scripts will always prefer Windows 10 first, and then Windows 7 if Windows 10 isn't possible.

saschanaz: Does that mean IE has different behaviors in Win7 and Win10? Should that be documented here?

vinyldarkscratch: It's less that IE has different behaviors, and more that the test cloud (SauceLabs) does.

foolip: > AFAIK IE only defines console functions when DevTools is opened, so maybe that's the cause.

It sounds like maybe this wasn't the cause here, but I'll keep this in mind when reviewing any IE changes. Where would this be documented, though?

api/Console.json
```diff
@@ -1362,18 +1362,20 @@
             "opera": {
               "version_added": "60"
             },
+            "opera_android": {
+              "version_added": "50"
+            },
             "safari": {
-              "version_added": false,
-              "notes": "See <a href='https://webkit.org/b/186833'>bug 186833</a>."
```
foolip, line 1367: https://bugs.webkit.org/show_bug.cgi?id=199184 RESOLVED FIXED, nice!

foolip (approved):

PR #6833, opened 2020-10-05
Fix visibilitychange event data for IE/Edge

vinyldarkscratch:
This PR fixes data for the `visibilitychange` event and handler for IE and Edge.  As determined by manual testing (after reviewing results from the mdn-bcd-collector), Internet Explorer and Edge had no support for `onvisibilitychange` until Edge 18.  However, it supported the event since IE 10, through `addEventListener`.

Test code used:
```js
document.addEventListener("visibilitychange", function() {
  if (document.visibilityState === 'visible') {
    console.log('addEventListener: visible');
  } else {
    console.log('addEventListener: hidden');
  }
});

  if (document.visibilityState === 'visible') {
    console.log('onvisibilitychange: visible');
  } else {
    console.log('onvisibilitychange: hidden');
  }
};
```

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -7651,7 +7651,8 @@
               "version_added": "33"
             },
             "edge": {
-              "version_added": "18"
+              "version_added": "18",
+              "notes": "Before Edge 18, this event handler was not supported, however the event itself was supported since Edge 12. The event can be listened to via <code>document.addEventListener('visibilitychange', function() {});</code>."
             },
             "firefox": {
               "version_added": "56"
@@ -7660,7 +7661,8 @@
               "version_added": "56"
             },
             "ie": {
-              "version_added": false
+              "version_added": false,
+              "notes": "This event handler is not supported, however the event itself is supported since IE 10. The event can be listened to via <code>document.addEventListener('visibilitychange', function() {});</code>."
             },
             "opera": {
               "version_added": "12.1",
--- a/api/Document.json
+++ b/api/Document.json
@@ -10825,7 +10825,8 @@
               "version_added": "33"
             },
             "edge": {
-              "version_added": "12"
+              "version_added": "12",
+              "notes": "The <code>onvisibilitychange</code> property was not supported until Edge 18. To listen to this event in earlier versions of Edge, use <code>document.addEventListener('visibilitychange', function() {});</code>."
             },
             "firefox": {
               "version_added": "56"
@@ -10834,7 +10835,8 @@
               "version_added": "56"
             },
             "ie": {
-              "version_added": "10"
+              "version_added": "10",
+              "notes": "The <code>onvisibilitychange</code> property is not supported in IE. To listen to this event, use <code>document.addEventListener('visibilitychange', function() {});</code>."
             },
             "opera": {
               "version_added": "12.1",
--- a/api/Document.json
+++ b/api/Document.json
@@ -7652,7 +7652,7 @@
             },
             "edge": {
               "version_added": "18",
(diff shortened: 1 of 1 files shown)


api/Document.json
```diff
@@ -7660,7 +7660,7 @@
               "version_added": "56"
             },
             "ie": {
-              "version_added": "10"
+              "version_added": false
```
foolip, line 7663: Can you add a note to say that while the event handler attribute wasn't supported, the event itself was?

api/Document.json
```diff
@@ -10823,7 +10823,7 @@
               "version_added": "33"
             },
             "edge": {
-              "version_added": "18"
+              "version_added": "12"
```
foolip, line 10826: Similarly, here can you point out that the event handler attribute won't work? It would be natural to assume it does.

foolip:
@ddbeck here's a good example of a event handler vs event disconnect discussed in https://github.com/mdn/browser-compat-data/pull/6710#issuecomment-702592082. You might want to review this PR to make sure you like how the notes end up.

api/Document.json
vinyldarkscratch: Done!  Does the note look good?

api/Document.json
```diff
@@ -7651,7 +7651,8 @@
               "version_added": "33"
             },
             "edge": {
-              "version_added": "18"
+              "version_added": "18",
+              "notes": "Before Edge 18, this event handler was not supported, however the event itself was supported since Edge 12. The event can be listened to via <code>document.addEventListener('visibilitychange', function() {});</code>."
```
foolip, line 7655:
```suggestion
              "notes": "Before Edge 18, this event handler attribute was not supported, however the event itself was supported since Edge 12. The event can be listened to via <code>document.addEventListener('visibilitychange', function() {});</code>."
```

I would call it an "event handler attribute" to be as specific as possible. There's also a note for api.Element.wheel_event for a very similar situation &ndash; it would be good to use similar language for both, and for the likely many future cases we'll find.

foolip (commented):
I think this looks about right, but requesting review from @ddbeck.

api/Document.json
```diff
@@ -7651,7 +7651,8 @@
               "version_added": "33"
             },
             "edge": {
-              "version_added": "18"
+              "version_added": "18",
+              "notes": "Before Edge 18, this event handler attribute was not supported, however the event itself was supported since Edge 12. The event can be listened to via <code>document.addEventListener('visibilitychange', function() {});</code>."
```
ddbeck, line 7655: nitpick: punctuation near "however" is so particular
```suggestion
              "notes": "Before Edge 18, this event handler attribute was not supported; however, the event itself was supported since Edge 12. The event can be listened to via <code>document.addEventListener('visibilitychange', function() {});</code>."
```

api/Document.json
```diff
@@ -7660,7 +7661,8 @@
               "version_added": "56"
             },
             "ie": {
-              "version_added": "10"
+              "version_added": false,
+              "notes": "This event handler attribute is not supported, however the event itself is supported since IE 10. The event can be listened to via <code>document.addEventListener('visibilitychange', function() {});</code>."
```
ddbeck, line 7665:
```suggestion
              "notes": "This event handler attribute is not supported; however, the event itself is supported since IE 10. The event can be listened to via <code>document.addEventListener('visibilitychange', function() {});</code>."
```

ddbeck (changes requested):
I have the tiniest nitpick about a couple notes, but otherwise, this looks great. Thank you!

api/Document.json
ddbeck: Might be relevant over on #6849, too

foolip (approved):
Great notes, thanks for the guidance, @ddbeck!

PR #6834, opened 2020-10-05
text-combine-upright note

rachelandrew:
Added a note to the `text-combine-upright` property as it was made non-animatable in Firefox 81: https://bugzilla.mozilla.org/show_bug.cgi?id=1654195

### DIFF
--- a/css/properties/text-combine-upright.json
+++ b/css/properties/text-combine-upright.json
@@ -33,6 +33,10 @@
             },
             "firefox": [
               {
+                "version_added": "81",
+                "notes": "Before version 81, Firefox implemented the property as animatable. This was corrected to spec in 81."
+              },
+              {
                 "version_added": "48",
                 "notes": "Before version 48, Firefox did not implement layout support for tate-chū-yoko."
               },
@@ -71,6 +75,10 @@
               }
             ],
             "firefox_android": [
+              {
+                "version_added": "81",
+                "notes": "Before version 81, Firefox implemented the property as animatable. This was corrected to spec in 81."
+              },
               {
                 "version_added": "48",
                 "notes": "Before version 48, Firefox did not implement layout support for tate-chū-yoko."


jpmedley:
@rachelandrew, thank you for submitting this. There is an [`alternative_name` member](https://github.com/mdn/browser-compat-data/blob/master/schemas/compat-data-schema.md#alternative_name) for cases like this. Can you please update this?

vinyldarkscratch:
@jpmedley, I'm confused: this has PR nothing to do with alternative names.

rachelandrew:
@jpmedley I'm not sure what change you are expecting? This is just a note about the removal of animation to match the spec. Maybe there is a better way to indicate that, but I don't think it's an alternative name.

ddbeck (approved):
Looks good. Thank you! 🎉

jpmedley:
>I'm confused: this has PR nothing to do with alternative names.

Doesn't it? The property was called `animatable` before 81. Haven't we used alternative names for this use case before?

vinyldarkscratch:
> >I'm confused: this has PR nothing to do with alternative names.
> 
> Doesn't it? The property was called `animatable` before 81. Haven't we used alternative names for this use case before? 

For the `text-combine-upright` property?  I highly doubt it.  If we _are_ using alt. names to represent behavioral changes rather than what the property itself is called, then that is a problem in our data.

ddbeck:
@jpmedley:

> The property was called `animatable` before 81. Haven't we used alternative names for this use case before?

Can you link to some evidence that the property previously had a different name? I understood Rachel's PR (and the linked bug) as addressing the property's [_Animation type_ in the specification](https://drafts.csswg.org/css-writing-modes-4/#text-combine-upright), not the name of the property. If I've misunderstood something and merged this erroneously, I'd be happy to set it right—but I need some more information to understand it first.

PR #6836, opened 2020-10-05
Add/update Chromium data for Document API

vinyldarkscratch:
This PR updates the Chromium data for the Document API based upon a combination of manual testing and collector results to achieve more real data for the Document API.  (Note: most Opera updates have been deferred to #6835.)

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -153,7 +153,7 @@
               "version_added": "1"
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": "18"
             },
             "edge": {
               "version_added": "12"
@@ -174,16 +174,16 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "1"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "1"
             },
             "samsunginternet_android": {
-              "version_added": true
+              "version_added": "1.0"
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "1"
             }
           },
           "status": {
@@ -201,7 +201,7 @@
               "version_added": "1"
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": "18"
             },
             "edge": {
               "version_added": "12"
@@ -222,16 +222,16 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "3"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "1"
             },
             "samsunginternet_android": {
-              "version_added": true
+              "version_added": "1.0"
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "1"
             }
           },
           "status": {
@@ -246,11 +246,11 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Document/alinkColor",
           "support": {
             "chrome": {
-              "version_added": "64",
+              "version_added": "1",
               "notes": "Before Chrome 64, this property was accessed through the <code>HTMLDocument</code> alias."
             },
             "chrome_android": {
-              "version_added": "64",
+              "version_added": "18",
               "notes": "Before Chrome 64, this property was accessed through the <code>HTMLDocument</code> alias."
             },
             "edge": {
@@ -278,11 +278,11 @@
               "version_added": true
             },
             "samsunginternet_android": {
-              "version_added": "9.0",
+              "version_added": "1.0",
               "notes": "Before Samsung Internet 9.0, this property was accessed through the <code>HTMLDocument</code> alias."
             },
             "webview_android": {
-              "version_added": "64",
+              "version_added": "1",
               "notes": "Before Chrome 64, this property was accessed through the <code>HTMLDocument</code> alias."
             }
           },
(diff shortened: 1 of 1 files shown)


api/Document.json
```diff
@@ -174,16 +174,16 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "1"
```
foolip, line 177: Based on the PR description I didn't expect to see Safari changes, and there are only a few. I'll push a commit to revert those to make it Chromium-only.

api/Document.json
```diff
@@ -246,11 +246,11 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Document/alinkColor",
           "support": {
             "chrome": {
-              "version_added": "64",
+              "version_added": "1",
```
foolip, line 249: Can we figure this one out as part of https://github.com/mdn/browser-compat-data/issues/6839, along with anything that has notes for `HTMLDocument`?

Since we don't have an api.HTMLDocument.alinkColor entry, I think the best we can do here is to have two entries, the first being `partial_implementation`. The difference is not entirely academic, since `document.alinkColor` won't work in SVG and XML documents before the change.

foolip (commented):
I reviewed this for a bit, but it's too big, changing something like 285 things. Can you split this alphabetically or something like that into at least 3 pieces, so it's more manageable to review?

api/Document.json
vinyldarkscratch: I've separated these changes out as #6865!

foolip:
Thanks for trimming this, now I think it's small enough for me to review. Diving in!

api/Document.json
```diff
@@ -5879,10 +5921,12 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Document/height",
           "support": {
             "chrome": {
-              "version_added": false
+              "version_added": "1",
+              "version_removed": true
```
foolip, line 5925: It was probably removed long ago, this is introducing uncertainty for what's in current Chrome. Can you either add the exact removal version or defer this change until later when it is known?

api/Document.json
```diff
@@ -7164,12 +7210,24 @@
...
-            },
-            "opera_android": {
-              "version_added": true
-            },
+            "opera": [
+              {
+                "version_added": "58"
+              },
+              {
+                "version_added": "32",
```
foolip, line 7218: The data that this is mirroring is wrong. `document.onwebkitfullscreenchange` was in Chrome 15 per http://mdn-bcd-collector.appspot.com/tests/api/Document/onwebkitfullscreenchange, probably earlier. Can you update the Chrome data for that, or just not mirror the incorrect data until it's fixed?

api/Document.json
```diff
@@ -7269,12 +7333,24 @@
...
```
foolip, line 7341: This is also wrong per http://mdn-bcd-collector.appspot.com/tests/api/Document/onwebkitfullscreenerror. It wasn't added at the same time as `onwebkitfullscreenchange`, but before Chrome 45 at least.

api/Document.json
```diff
@@ -7546,17 +7628,11 @@
...
-              {
-                "version_added": "33"
-              },
-              {
-                "version_added": "13",
-                "prefix": "webkit"
-              }
-            ],
+            "chrome": {
+              "version_added": "62"
```
foolip, line 7632: This seems much too new. I suspect a similar situation as in https://github.com/mdn/browser-compat-data/pull/6849, can you break out this change?

api/Document.json
```diff
@@ -8846,7 +8922,7 @@
               "version_added": true
             },
             "opera_android": {
-              "version_added": true
+              "version_added": "14"
```
foolip, line 8925: This is a peculiar edit. chrome_android is true and wasn't changed, plus opera above was not mirrored. Did you intend to make this change?

api/Document.json
```diff
@@ -9442,7 +9520,7 @@
               "version_added": null
             },
             "opera_android": {
-              "version_added": null
+              "version_added": false
```
foolip, line 9523: Here's another stray opera_android edit. Looks correct, but an accident nonetheless?

api/Document.json
```diff
@@ -9521,13 +9599,13 @@
...
-              "version_added": false
+              "version_added": "85"
             },
             "chrome_android": {
-              "version_added": false
+              "version_added": "85"
             },
             "edge": {
-              "version_added": false
+              "version_added": "85"
```
foolip, line 9608: I thought Edge had shipped `document.requestStorageAccess` before Chrome, but maybe not. They were directly involved though: https://blogs.windows.com/msedgedev/2020/07/08/introducing-storage-access-api/

api/Document.json
```diff
@@ -11063,10 +11139,12 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Document/width",
           "support": {
             "chrome": {
-              "version_added": false
+              "version_added": "1",
+              "version_removed": true
```
foolip, line 11143: Same comment as for `document.height` applies here.

foolip (commented):
A few changes requested here. I'll not revert bits and land directly in this case, since I'm not certain about every suggestion.

foolip:
Actually, since this ends up blocking https://github.com/mdn/browser-compat-data/pull/6864 I *will* revert the bits I had comments on myself, and summarize needed follow-up in a comment.

foolip:
Alright, merged. The required followup is simply the stuff reverted in https://github.com/mdn/browser-compat-data/commit/8dcd73f9974f76865c63876c43df98d045c1dc0d:
- document.height/width
- document.onwebkitfullscreenchange/error
- document.onvisibilitychange
- document.queryCommandIndeterm

api/Document.json
andybons: MS is driving the implementation in chromium: https://bugs.chromium.org/p/chromium/issues/detail?id=989663

They enabled it to be on by default in Edge before Chrome but I can’t find any publicly source that indicates which Edge version it was enabled in 😿

PR #6837, opened 2020-10-05
Add/update Opera data for Window API

vinyldarkscratch:
This PR updates the Opera (Presto) data for the Window API based upon manual testing to achieve more real data for the Window API. Some data is mirrored from Chrome.

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -886,7 +886,7 @@
               "version_added": true
             },
             "opera": {
-              "version_added": true
+              "version_added": "≤12.1"
             },
             "opera_android": {
               "version_added": "≤12.1"
--- a/api/Window.json
+++ b/api/Window.json
@@ -889,7 +889,7 @@
               "version_added": true
             },
             "opera_android": {
-              "version_added": true
+              "version_added": "≤12.1"
             },
             "safari": {
               "version_added": true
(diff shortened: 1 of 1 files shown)


api/Window.json
```diff
@@ -2199,7 +2199,10 @@
               "version_added": null
             },
             "opera": {
-              "version_added": true
+              "version_added": "15"
```
foolip, line 2202: This is post-Presto. There's a bunch of that, but the PR isn't too big, so I'll just change the description instead.

api/Window.json
```diff
@@ -6243,10 +6246,10 @@
               "version_added": "9"
             },
             "opera": {
-              "version_added": true
+              "version_added": "3"
```
foolip, line 6249: Is this based on testing Opera 3? It's the only so early version in this PR... If there's any doubt, then I'd prefer "≤12.1".

foolip (commented):
Good stuff, just a few minor things to fix or back out.

api/Window.json
vinyldarkscratch: I copied this from `pageYOffset`, and then confirmed the results!

foolip (approved):

PR #6838, opened 2020-10-05
Add/update Chromium data for Window API

vinyldarkscratch:
This PR updates the Chromium data for the Window API based upon a combination of manual testing and collector results to achieve more real data for the Window API. (Note: most Opera updates have been deferred to #6837.)

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -1560,10 +1560,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Window/crypto",
           "support": {
             "chrome": {
-              "version_added": "1"
+              "version_added": "37"
             },
             "chrome_android": {
-              "version_added": "18"
+              "version_added": "37"
             },
             "edge": {
               "version_added": "12"
@@ -1579,10 +1579,10 @@
               "prefix": "ms"
             },
             "opera": {
-              "version_added": "15"
+              "version_added": "24"
             },
             "opera_android": {
-              "version_added": "14"
+              "version_added": "24"
             },
             "safari": {
               "version_added": "6.1"
@@ -1591,10 +1591,10 @@
               "version_added": "6.1"
             },
             "samsunginternet_android": {
-              "version_added": "1.0"
+              "version_added": "3.0"
             },
             "webview_android": {
-              "version_added": "1"
+              "version_added": "37"
             }
           },
           "status": {
@@ -2180,10 +2180,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Window/external",
           "support": {
             "chrome": {
-              "version_added": "1"
+              "version_added": true
             },
             "chrome_android": {
-              "version_added": "18"
+              "version_added": true
             },
             "edge": {
               "version_added": "12"
@@ -2205,10 +2205,10 @@
               "version_added": null
             },
             "samsunginternet_android": {
-              "version_added": "1.0"
+              "version_added": true
             },
             "webview_android": {
-              "version_added": "1"
+              "version_added": true
             }
           },
           "status": {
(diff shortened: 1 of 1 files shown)


api/Window.json
```diff
@@ -972,10 +972,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Window/cancelAnimationFrame",
```
foolip, line 972: Can you test for when the prefixed variants were added? They're included for `requestAnimationFrame`: https://developer.mozilla.org/en-US/docs/Web/API/Window/requestAnimationFrame

Tests:
http://mdn-bcd-collector.appspot.com/tests/api/Window/webkitCancelAnimationFrame
http://mdn-bcd-collector.appspot.com/tests/api/Window/webkitCancelRequestAnimationFrame (after https://github.com/foolip/mdn-bcd-collector/pull/601)

foolip (commented):
I didn't review, just spotted `cancelAnimationFrame` and left that comment.

api/Window.json
```diff
@@ -1560,10 +1560,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Window/crypto",
           "support": {
             "chrome": {
-              "version_added": "37"
+              "version_added": "1"
```
foolip, line 1563: This can't be right, `window.crypto` is newer than this. https://trac.webkit.org/changeset/78321/webkit is when it was added to WebKit, although there was something before that returning undefined with a "FIXME: Implement me" comment. I didn't check how that that was around for.

This also threw off results for Safari, so maybe we should do a custom test for `window.crypto` that ensures it exists *and* is an object at least?

api/Window.json
```diff
@@ -2180,10 +2180,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Window/external",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "1"
```
foolip, line 2183: `window.external` is a bit special, I think that for a long time it wasn't defined using Web IDL a bit like `console`. Have you checked what kind of thing it is in Chrome 1, was it an object with methods on it? I'm worried it might be a similar situation to `window.crypto` above.

api/Window.json
```diff
@@ -8441,19 +8441,19 @@
           "support": {
             "chrome": [
               {
-                "version_added": true
+                "version_added": "1"
               },
               {
-                "version_added": true,
+                "version_added": "1",
                 "alternative_name": "pageXOffset"
```
foolip, line 8448: This is api.Window.scrollX. I only noticed this now, but we also have a api.Window.pageXOffset entry, so these a `alternative_name` entries are weird, especially when both go back to ancient times. Can you clean that up in a separate PR?

foolip (commented):
I've finished reviewing now. This is a big improvement, so I'll go ahead and back out the bits I'm not so sure about and merge what remains, to not risk getting conflicts and multiple roundtrips.

api/Window.json
foolip: https://github.com/mdn/browser-compat-data/pull/6825#discussion_r501629768 the other place we talked about this.

foolip (approved):

foolip:
@vinyldarkscratch I'll merge this now, but to avoid ambiguity these are the things I'm asking for post-merging:
- redoing the `window.crypto` and `window.external` changes that I reverted in a separate PR (or 2 if you like)
- adding prefixed statements for `webkitCancelAnimationFrame` and `webkitCancelRequestAnimationFrame`. might also check if the `webkitRequestAnimationFrame` data is correct too.
- getting rid of `"alternative_name": "pageXOffset"` and `"alternative_name": "pageYOffset"` statements under `scrollX`/`scrollY` if they're not adding beyond the `pageXOffset`/`pageYOffset` entries.

PR #6840, opened 2020-10-05
mdn-bcd-collector: update data for WebGL EXT_ extensions

vinyldarkscratch:
This PR uses the mdn-bcd-collector project to update the compatibility data for all of the `EXT_` prefixed WebGL extensions, along with some mirroring and manual testing.  All results were confirmed for accuracy.  (`EXT_color_buffer_float` was omitted due to a slight issue in the collector tests, which will be followed-up later.)

Tests used:
https://mdn-bcd-collector.appspot.com/tests/api/EXT_blend_minmax
https://mdn-bcd-collector.appspot.com/tests/api/EXT_color_buffer_half_float
https://mdn-bcd-collector.appspot.com/tests/api/EXT_disjoint_timer_query
https://mdn-bcd-collector.appspot.com/tests/api/EXT_frag_depth
https://mdn-bcd-collector.appspot.com/tests/api/EXT_sRGB
https://mdn-bcd-collector.appspot.com/tests/api/EXT_shader_texture_lod
https://mdn-bcd-collector.appspot.com/tests/api/EXT_texture_compression_bptc
https://mdn-bcd-collector.appspot.com/tests/api/EXT_texture_compression_rgtc
https://mdn-bcd-collector.appspot.com/tests/api/EXT_texture_filter_anisotropic

### DIFF
--- a/api/EXT_blend_minmax.json
+++ b/api/EXT_blend_minmax.json
@@ -5,40 +5,40 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/EXT_blend_minmax",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "38"
           },
           "chrome_android": {
-            "version_added": true
+            "version_added": "38"
           },
           "edge": {
             "version_added": "17"
           },
           "firefox": {
-            "version_added": "33"
+            "version_added": "47"
           },
           "firefox_android": {
-            "version_added": null
+            "version_added": "47"
           },
           "ie": {
             "version_added": false
           },
           "opera": {
-            "version_added": true
+            "version_added": "25"
           },
           "opera_android": {
-            "version_added": true
+            "version_added": "25"
           },
           "safari": {
-            "version_added": null
+            "version_added": "7"
           },
           "safari_ios": {
-            "version_added": null
+            "version_added": "7"
           },
           "samsunginternet_android": {
-            "version_added": true
+            "version_added": "3.0"
           },
           "webview_android": {
-            "version_added": true
+            "version_added": "38"
           }
         },
         "status": {
--- a/api/EXT_blend_minmax.json
+++ b/api/EXT_blend_minmax.json
@@ -13,10 +13,17 @@
           "edge": {
             "version_added": "17"
           },
-          "firefox": {
-            "version_added": "35",
-            "notes": "Before Firefox 47, this extension was not available on Windows."
-          },
+          "firefox": [
+            {
+              "version_added": "47"
+            },
+            {
+              "version_added": "35",
+              "version_removed": "47",
+              "partial_implementation": true,
+              "notes": "Before Firefox 47, this extension was not available on Windows."
+            }
+          ],
           "firefox_android": {
             "version_added": null
           },
--- a/api/EXT_blend_minmax.json
+++ b/api/EXT_blend_minmax.json
@@ -21,7 +21,7 @@
               "version_added": "35",
               "version_removed": "47",
               "partial_implementation": true,
-              "notes": "Before Firefox 47, this extension was not available on Windows."
+              "notes": "Not supported on Windows."
             }
           ],
           "firefox_android": {
--- a/api/EXT_blend_minmax.json
+++ b/api/EXT_blend_minmax.json
@@ -14,10 +14,11 @@
             "version_added": "17"
           },
           "firefox": {
-            "version_added": "47"
+            "version_added": "35",
+            "notes": "Before Firefox 47, this extension was not available on Windows."
           },
           "firefox_android": {
-            "version_added": "47"
+            "version_added": null
           },
           "ie": {
             "version_added": false
--- a/api/EXT_color_buffer_float.json
+++ b/api/EXT_color_buffer_float.json
@@ -5,40 +5,40 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/EXT_color_buffer_float",
         "support": {
           "chrome": {
-            "version_added": "63"
+            "version_added": false
           },
           "chrome_android": {
-            "version_added": "63"
+            "version_added": false
           },
           "edge": {
-            "version_added": "79"
+            "version_added": false
           },
           "firefox": {
-            "version_added": "49"
+            "version_added": false
           },
           "firefox_android": {
-            "version_added": null
+            "version_added": false
           },
           "ie": {
             "version_added": false
           },
           "opera": {
-            "version_added": null
+            "version_added": false
           },
           "opera_android": {
-            "version_added": null
+            "version_added": false
           },
           "safari": {
-            "version_added": null
+            "version_added": false
           },
           "safari_ios": {
-            "version_added": null
+            "version_added": false
           },
           "samsunginternet_android": {
-            "version_added": "8.0"
+            "version_added": false
           },
           "webview_android": {
-            "version_added": "63"
+            "version_added": false
           }
         },
         "status": {
--- a/api/EXT_color_buffer_float.json
+++ b/api/EXT_color_buffer_float.json
@@ -5,40 +5,40 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/EXT_color_buffer_float",
         "support": {
           "chrome": {
-            "version_added": false
+            "version_added": "63"
           },
           "chrome_android": {
-            "version_added": false
+            "version_added": "63"
           },
           "edge": {
-            "version_added": false
+            "version_added": "79"
           },
           "firefox": {
-            "version_added": false
+            "version_added": "49"
           },
           "firefox_android": {
-            "version_added": false
+            "version_added": null
           },
           "ie": {
             "version_added": false
           },
           "opera": {
-            "version_added": false
+            "version_added": null
           },
           "opera_android": {
-            "version_added": false
+            "version_added": null
           },
           "safari": {
-            "version_added": false
+            "version_added": null
           },
           "safari_ios": {
-            "version_added": false
+            "version_added": null
           },
           "samsunginternet_android": {
-            "version_added": false
+            "version_added": "8.0"
           },
           "webview_android": {
-            "version_added": false
+            "version_added": "63"
           }
         },
         "status": {
--- a/api/EXT_color_buffer_half_float.json
+++ b/api/EXT_color_buffer_half_float.json
@@ -14,19 +14,19 @@
             "version_added": "17"
           },
           "firefox": {
-            "version_added": "30"
+            "version_added": "47"
           },
           "firefox_android": {
-            "version_added": null
+            "version_added": "47"
           },
           "ie": {
             "version_added": false
           },
           "opera": {
-            "version_added": null
+            "version_added": "50"
           },
           "opera_android": {
-            "version_added": null
+            "version_added": "46"
           },
           "safari": {
             "version_added": "14"
--- a/api/EXT_color_buffer_half_float.json
+++ b/api/EXT_color_buffer_half_float.json
@@ -21,7 +21,7 @@
               "version_added": "36",
               "version_removed": "47",
               "partial_implementation": true,
-              "notes": "Before Firefox 47, this extension was not available on Windows."
+              "notes": "Not supported on Windows."
             }
           ],
           "firefox_android": {
--- a/api/EXT_color_buffer_half_float.json
+++ b/api/EXT_color_buffer_half_float.json
@@ -13,10 +13,17 @@
           "edge": {
             "version_added": "17"
           },
-          "firefox": {
-            "version_added": "36",
-            "notes": "Before Firefox 47, this extension was not available on Windows."
-          },
+          "firefox": [
+            {
+              "version_added": "47"
+            },
+            {
+              "version_added": "36",
+              "version_removed": "47",
+              "partial_implementation": true,
+              "notes": "Before Firefox 47, this extension was not available on Windows."
+            }
+          ],
           "firefox_android": {
             "version_added": null
           },
--- a/api/EXT_color_buffer_half_float.json
+++ b/api/EXT_color_buffer_half_float.json
@@ -14,10 +14,11 @@
             "version_added": "17"
           },
           "firefox": {
-            "version_added": "47"
+            "version_added": "36",
+            "notes": "Before Firefox 47, this extension was not available on Windows."
           },
           "firefox_android": {
-            "version_added": "47"
+            "version_added": null
           },
           "ie": {
             "version_added": false
--- a/api/EXT_disjoint_timer_query.json
+++ b/api/EXT_disjoint_timer_query.json
@@ -15,12 +15,11 @@
             "notes": "Removed due to the <a href='https://www.vusec.net/projects/glitch/'>GLitch exploit</a>."
           },
           "edge": {
-            "version_added": false,
-            "notes": "Removed due to the <a href='https://www.vusec.net/projects/glitch/'>GLitch exploit</a>."
+            "version_added": false
           },
           "firefox": {
             "version_added": "51",
-            "version_removed": "63",
+            "version_removed": "59",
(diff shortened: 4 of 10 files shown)


api/EXT_blend_minmax.json
```diff
@@ -5,40 +5,40 @@
...
           },
           "chrome_android": {
-            "version_added": true
+            "version_added": "38"
           },
           "edge": {
             "version_added": "17"
           },
           "firefox": {
-            "version_added": "33"
```
foolip, line 17: This seems to have come from the initial migration in https://github.com/mdn/browser-compat-data/pull/70, so probably an error from the wiki.

api/EXT_color_buffer_float.json
```diff
@@ -5,40 +5,40 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/EXT_color_buffer_float",
         "support": {
           "chrome": {
-            "version_added": "63"
+            "version_added": false
```
foolip, line 8: Is the "EXT_color_buffer_float" extension not supported anywhere?

Seems like this is a problem with our tests. We create a WebGL1 context, and this extension is only supported for WebGL2.

I don't know if there are any WebGL1-only extensions, but if there aren't, then we should try to create a WebGL2 context and fall back to WebGL1. Or we create one of each and try both in these extension tests.

I won't review beyond this point because the results could all be wrong due to this issue.

foolip (changes requested):
Oops, seems like we have a collector issue here.

api/EXT_color_buffer_float.json
vinyldarkscratch: Ah, my bad, I thought this was a WebGL1 extension.  I'll get the collector fixed to also test WebGL2!

vinyldarkscratch: I reverted the changes to this file and will defer them to a new PR when collector v0.5.3 is released.  I did a quick check of the other files, and have confirmed that the extensions missing in Edge are not exposed through WebGL2 (otherwise, all of the data is for browser versions before WebGL2 was introduced).

api/EXT_frag_depth.json
```diff
@@ -5,40 +5,41 @@
...
+            "version_added": "38"
           },
           "chrome_android": {
-            "version_added": true
+            "version_added": "38"
           },
           "edge": {
-            "version_added": "≤18"
+            "version_added": "12",
+            "version_removed": "79"
```
foolip, line 15: What's going on here? Does current Chrome support this extension, but current Edge does not?

api/EXT_frag_depth.json
```diff
@@ -5,40 +5,41 @@
...
           },
           "firefox": {
-            "version_added": "30"
```
foolip, line 17: This is also from https://github.com/mdn/browser-compat-data/pull/70.

api/EXT_sRGB.json
```diff
@@ -5,40 +5,40 @@
...
-            "version_added": true
+            "version_added": "40"
           },
           "chrome_android": {
-            "version_added": true
+            "version_added": "40"
           },
           "edge": {
-            "version_added": "79"
+            "version_added": false
```
foolip, line 14: This also seems suspect. A testing problem?

api/EXT_sRGB.json
```diff
@@ -5,40 +5,40 @@
...
           },
           "firefox": {
-            "version_added": "28"
+            "version_added": "58"
```
foolip, line 17: This is EXT_sRGB, and https://bugzilla.mozilla.org/show_bug.cgi?id=843668#c56 suggests that Firefox 28 was correct, but https://bugzilla.mozilla.org/show_bug.cgi?id=1403798#c5 suggests that it was fixed on Windows in Firefox 58.

WebGL is more OS- and hardware-dependent than most web APIs. This and the WebGL 1 vs 2 issues does make me doubt the validity of the results overall, not just where a real version number has been changed.

What do you think about collecting results from macOS as well to get some more confidence in the results?

api/EXT_blend_minmax.json
```diff
@@ -5,40 +5,40 @@
...
+            "version_added": "47"
```
foolip, line 17: Oops, no, 33 seems to have been correct: https://bugzilla.mozilla.org/show_bug.cgi?id=973815

api/EXT_color_buffer_half_float.json
```diff
@@ -14,19 +14,19 @@
             "version_added": "17"
           },
           "firefox": {
-            "version_added": "30"
+            "version_added": "47"
```
foolip, line 17: Firefox 30 or 39 seems to be correct here:
https://bugzilla.mozilla.org/show_bug.cgi?id=843666
https://bugzilla.mozilla.org/show_bug.cgi?id=1130616

api/EXT_blend_minmax.json
```diff
@@ -5,40 +5,40 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/EXT_blend_minmax",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "38"
```
foolip, line 8: Shipped by https://bugs.chromium.org/p/chromium/issues/detail?id=389728, and the date supports that this was Chrome 38.

api/EXT_frag_depth.json
```diff
@@ -5,40 +5,41 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/EXT_frag_depth",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "38"
```
foolip, line 8: Also confirmed by https://bugs.chromium.org/p/chromium/issues/detail?id=389728.

api/EXT_shader_texture_lod.json
```diff
@@ -5,40 +5,41 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/EXT_shader_texture_lod",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "38"
```
foolip, line 8: Also confirmed by https://bugs.chromium.org/p/chromium/issues/detail?id=389728.

api/EXT_shader_texture_lod.json
```diff
@@ -5,40 +5,41 @@
...
           },
           "chrome_android": {
-            "version_added": true
+            "version_added": "38"
           },
           "edge": {
-            "version_added": "17"
+            "version_added": "17",
+            "version_removed": "79"
```
foolip, line 15: No Edge support?

api/EXT_texture_compression_bptc.json
```diff
@@ -14,7 +14,7 @@
             "version_added": false
           },
           "firefox": {
-            "version_added": "65"
+            "version_added": "68"
```
foolip, line 17: I can't find a Firefox bug for this, but do wonder if "65" might actually be correct, as with a few of the other changes.

api/EXT_texture_compression_rgtc.json
```diff
@@ -14,7 +14,7 @@
             "version_added": false
           },
           "firefox": {
-            "version_added": "65"
+            "version_added": false
```
foolip, line 17: I'd like to see more evidence of "65" being wrong.

api/EXT_texture_filter_anisotropic.json
```diff
@@ -6,73 +6,73 @@
         "support": {
           "chrome": [
             {
-              "version_added": true
+              "version_added": "34"
             },
             {
-              "version_added": true,
+              "version_added": "35",
```
foolip, line 12: This seems a bit implausible, that the prefixed variant would be shipped *after* the unprefixed variant. Can you confirm these results? I found https://bugs.chromium.org/p/chromium/issues/detail?id=118245 but would have to dig deeper to be sure about the versions from source code alone.

api/EXT_texture_filter_anisotropic.json
```diff
@@ -6,73 +6,73 @@
...
               "prefix": "WEBKIT_"
             }
           ],
           "edge": {
-            "version_added": "≤18"
+            "version_added": "12"
```
foolip, line 26: If Chrome still has the prefixed variant, shouldn't Edge also get it? (I half suspect it's not in Chrome any longer.)

foolip (changes requested):
OK, I took a much closer look at the individual changes once I started to notice some issues, and as a result have many questions.

There are parts of this that are confirmed and could land first, if you'd like to break out the tricky stuff into their own PRs.

api/EXT_frag_depth.json
vinyldarkscratch: Yeah, Edge has four less extensions than Chrome does.

api/EXT_disjoint_timer_query.json
```diff
@@ -15,12 +15,11 @@
...
           },
           "edge": {
-            "version_added": false,
-            "notes": "Removed due to the <a href='https://www.vusec.net/projects/glitch/'>GLitch exploit</a>."
+            "version_added": false
           },
           "firefox": {
             "version_added": "51",
-            "version_removed": "63",
+            "version_removed": "59",
```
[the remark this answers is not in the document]
vinyldarkscratch: That's correct, the actual removal version for this API was Firefox 59, confirmed manually.  I also double-checked to make sure that it wasn't made only available in WebGL2 or WebGL1.

api/EXT_sRGB.json
vinyldarkscratch: This is one of the extensions Edge doesn't support.

api/EXT_shader_texture_lod.json
vinyldarkscratch: Unfortunately no, this is one of the extensions Edge doesn't support.

api/EXT_texture_filter_anisotropic.json
vinyldarkscratch: I thought it was weird as well, but alas, based upon manual testing, the prefixed variant did come a version later.

api/EXT_texture_filter_anisotropic.json
vinyldarkscratch: I forgot to test the prefix in Edge, oops!  It's still around in both Chrome and Edge.

api/EXT_sRGB.json
vinyldarkscratch: It looks like it _is_ an OS-specific thing.  It isn't available in Windows until Firefox 58, but on macOS, 28 returns `true`.  I can certainly update the collector's Selenium script to obtain macOS results as well!

api/EXT_blend_minmax.json
vinyldarkscratch: Actually, 33 is also incorrect.  The extension is not available in either macOS or Windows in Firefox 33.  However, there is a case of OS imparity: Windows was supported since 47, and macOS since 35.

api/EXT_color_buffer_half_float.json
vinyldarkscratch: It appears that this extension was supported in macOS since Firefox 36, and then Windows in 47.

api/EXT_texture_compression_bptc.json
vinyldarkscratch: Performing testing in both Windows and macOS, this extension wasn't present in either Firefox 65 or 67, but was present in both OSes for 68.

api/EXT_texture_compression_rgtc.json
vinyldarkscratch: Looks like this is only implemented on macOS.

api/EXT_disjoint_timer_query.json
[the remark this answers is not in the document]
foolip: Excellent, thank you!

api/EXT_sRGB.json
foolip: That would be great, so that we don't set some versions numbers here to earlier than the real cross-platform support.

api/EXT_color_buffer_half_float.json
foolip: Oh my, so much OS-dependence. I'll bet some of this also depends on the graphics card, but how to test for that I don't know. I know that for Chromium there are a bunch of different real machines running WebGL tests, but let's not go there unless we discover such a dependency...

api/EXT_texture_compression_rgtc.json
foolip: Aha, I guess we go with partial_implementation then?

api/EXT_texture_filter_anisotropic.json
foolip: This web platform thing is weird...

api/EXT_texture_filter_anisotropic.json
foolip: Aww, too bad, prefixes are lame.

api/EXT_sRGB.json
```diff
@@ -5,16 +5,17 @@
...
+            "version_added": "40"
           },
           "edge": {
-            "version_added": "79"
+            "version_added": false
           },
           "firefox": {
-            "version_added": "28"
+            "version_added": "28",
+            "notes": "Before Firefox 58, this extension was not available on Windows."
```
foolip, line 18: Can we use `partial_implementation` to represent this?

api/EXT_texture_compression_rgtc.json
```diff
@@ -14,7 +14,8 @@
             "version_added": false
           },
           "firefox": {
-            "version_added": "65"
+            "version_added": "65",
+            "notes": "Only available on macOS."
```
foolip, line 18: Can we use `partial_implementation` to represent this?

foolip (commented):
I'd like to use `partial_implementation` to reflect limited support, like was done for https://github.com/mdn/browser-compat-data/pull/6644. This way, after it is supported on all platforms, there's no note that developers have to look at, a note which will be less and less relevant over time.

api/EXT_blend_minmax.json
```diff
@@ -13,10 +13,17 @@
...
-          },
+          "firefox": [
+            {
+              "version_added": "47"
+            },
+            {
+              "version_added": "35",
+              "version_removed": "47",
+              "partial_implementation": true,
+              "notes": "Before Firefox 47, this extension was not available on Windows."
```
foolip, line 24:
```suggestion
              "notes": "Not supported on Windows."
```

Since the note applies to the Firefox 35-46 entry, specifying the version in the note doesn't seem that useful, and in fact would cause me to look harder to see if there's a distinction between "Firefox 35-46" and "Before Firefox 47", which there isn't here.

api/EXT_color_buffer_half_float.json
```diff
@@ -13,20 +13,28 @@
...
-          },
+          "firefox": [
+            {
+              "version_added": "47"
+            },
+            {
+              "version_added": "36",
+              "version_removed": "47",
+              "partial_implementation": true,
+              "notes": "Before Firefox 47, this extension was not available on Windows."
```
foolip, line 24:
```suggestion
              "notes": "Not supported on Windows."
```

api/EXT_sRGB.json
```diff
@@ -5,40 +5,48 @@
...
           },
+          "firefox": [
+            {
+              "version_added": "58"
+            },
+            {
+              "version_added": "28",
+              "version_removed": "58",
+              "partial_implementation": true,
+              "notes": "Before Firefox 58, this extension was not available on Windows."
```
foolip, line 24:
```suggestion
              "notes": "Not supported on Windows."
```

api/EXT_texture_compression_rgtc.json
```diff
@@ -14,7 +14,9 @@
             "version_added": false
           },
           "firefox": {
-            "version_added": "65"
+            "version_added": "65",
+            "partial_implementation": true,
+            "notes": "This extension is currently only available on macOS."
```
foolip, line 19:
```suggestion
            "notes": "Only supported on macOS."
```

foolip (approved):

PR #6842, opened 2020-10-05
Document.execCommand - note that calls cannot be nested

hamishwillee:
This adds a note that nested `Document.execCommand` calls are not supported in FF82 and later to both FireFox and FF Android. The information originates from https://bugzilla.mozilla.org/show_bug.cgi?id=1634262

Passes linter locally.

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -3779,10 +3779,12 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "1"
+              "version_added": "1",
+              "notes": "Nested calls are not supported from Firefox 82 (return <code>false</code>). See <a href='https://bugzil.la/1634262'>bug 1634262</a>."
             },
             "firefox_android": {
-              "version_added": "4"
+              "version_added": "4",
+              "notes": "Nested calls are not supported from Firefox 82 (return <code>false</code>). See <a href='https://bugzil.la/1634262'>bug 1634262</a>."
             },
             "ie": {
               "version_added": "4"
--- a/api/Document.json
+++ b/api/Document.json
@@ -3784,7 +3784,7 @@
             },
             "firefox_android": {
               "version_added": "4",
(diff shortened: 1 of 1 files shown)


chrisdavidmills (approved):
Looks pretty decent to me @hamishwillee .

@ddbeck does the data structure look OK to you?

api/Document.json
```diff
@@ -3779,10 +3779,12 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "1"
+              "version_added": "1",
+              "notes": "Nested calls are not supported from Firefox 82 (return <code>false</code>). See <a href='https://bugzil.la/1634262'>bug 1634262</a>."
```
ddbeck, line 3783: style: if there's some condition for the note (e.g., it applies only to certain versions or operating system), I try to lead with it

```suggestion
              "notes": "From Firefox 82, nested calls are not supported (return <code>false</code>). See <a href='https://bugzil.la/1634262'>bug 1634262</a>."
```

api/Document.json
```diff
@@ -3779,10 +3779,12 @@
...
             },
             "firefox_android": {
-              "version_added": "4"
+              "version_added": "4",
+              "notes": "Nested calls are not supported from Firefox 82 (return <code>false</code>). See <a href='https://bugzil.la/1634262'>bug 1634262</a>."
```
ddbeck, line 3787:
```suggestion
              "notes": "From Firefox 82, nested calls are not supported (return <code>false</code>). See <a href='https://bugzil.la/1634262'>bug 1634262</a>."
```

ddbeck (changes requested):
This looks pretty good. I do have one style suggestion in line comments, though. Thank you!

hamishwillee:
@ddbeck Thanks for the style tip - your suggestions merged and I'll take note for future changes.

PR #6843, opened 2020-10-06
Element.setPointerCapture throws NotFoundError

hamishwillee:
This adds a note to `Element.setPointerCapture()` that it throws `NotFoundError` instead of `InvalidPointerId` from FF82 (to both android and desktop versions). This originates from defect fix: https://bugzilla.mozilla.org/show_bug.cgi?id=1662124

I wasn't 100% sure of right place to put this information - see notes inline.

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -8138,7 +8138,8 @@
             ],
             "firefox_android": [
               {
-                "version_added": false
+                "version_added": true,
+                "notes": "Before Firefox 82, <code>setPointerCapture()</code> throws <code>InvalidPointerId</code> for an invalid <code>pointerId</code> argument. From Firefox 82, it throws <a href='https://w3c.github.io/pointerevents/#setting-pointer-capture'>the specified</a> <code>NotFoundError</code> exception. See <a href='https://bugzil.la/1662124'>bug 1662124</a>."
               },
               {
                 "version_added": "41",
@@ -8148,8 +8149,7 @@
                     "name": "dom.w3c_pointer_events.enabled",
                     "value_to_set": "true"
                   }
-                ],
-                "notes": "From Firefox 82 throws <code>NotFoundError</code> exception if pointer id is invalid (as per <a href='https://w3c.github.io/pointerevents/#setting-pointer-capture'>WC3 PointerEvents specification</a>). Earlier Firefox versions throw <code>InvalidPointerId</code>. See <a href='https://bugzil.la/1662124'>bug 1662124</a>."
+                ]
               }
             ],
             "ie": [
--- a/api/Element.json
+++ b/api/Element.json
@@ -8138,7 +8138,7 @@
             ],
             "firefox_android": [
               {
-                "version_added": true,
+                "version_added": "79",
                 "notes": "Before Firefox 82, <code>setPointerCapture()</code> throws <code>InvalidPointerId</code> for an invalid <code>pointerId</code> argument. From Firefox 82, it throws <a href='https://w3c.github.io/pointerevents/#setting-pointer-capture'>the specified</a> <code>NotFoundError</code> exception. See <a href='https://bugzil.la/1662124'>bug 1662124</a>."
               },
               {
--- a/api/Element.json
+++ b/api/Element.json
@@ -8122,7 +8122,8 @@
             },
             "firefox": [
               {
-                "version_added": "59"
+                "version_added": "59",
+                "notes": "From Firefox 82 throws <code>NotFoundError</code> exception if pointer id is invalid (as per <a href='https://w3c.github.io/pointerevents/#setting-pointer-capture'>WC3 PointerEvents specification</a>). Earlier Firefox versions throw <code>InvalidPointerId</code>. See <a href='https://bugzil.la/1662124'>bug 1662124</a>."
               },
               {
                 "version_added": "41",
@@ -8147,7 +8148,8 @@
                     "name": "dom.w3c_pointer_events.enabled",
                     "value_to_set": "true"
                   }
-                ]
+                ],
+                "notes": "From Firefox 82 throws <code>NotFoundError</code> exception if pointer id is invalid (as per <a href='https://w3c.github.io/pointerevents/#setting-pointer-capture'>WC3 PointerEvents specification</a>). Earlier Firefox versions throw <code>InvalidPointerId</code>. See <a href='https://bugzil.la/1662124'>bug 1662124</a>."
               }
             ],
             "ie": [
--- a/api/Element.json
+++ b/api/Element.json
@@ -8123,7 +8123,7 @@
             "firefox": [
               {
                 "version_added": "59",
-                "notes": "From Firefox 82 throws <code>NotFoundError</code> exception if pointer id is invalid (as per <a href='https://w3c.github.io/pointerevents/#setting-pointer-capture'>WC3 PointerEvents specification</a>). Earlier Firefox versions throw <code>InvalidPointerId</code>. See <a href='https://bugzil.la/1662124'>bug 1662124</a>."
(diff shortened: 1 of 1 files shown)


api/Element.json
```diff
@@ -8122,7 +8122,8 @@
             },
             "firefox": [
               {
-                "version_added": "59"
+                "version_added": "59",
+                "notes": "From Firefox 82 throws <code>NotFoundError</code> exception if pointer id is invalid (as per <a href='https://w3c.github.io/pointerevents/#setting-pointer-capture'>WC3 PointerEvents specification</a>). Earlier Firefox versions throw <code>InvalidPointerId</code>. See <a href='https://bugzil.la/1662124'>bug 1662124</a>."
```
hamishwillee, line 8126: So I added this to the latest full `version_added` section. Should I have added a new version  for FF82 in order to add the note to that? I think not, but can you please confirm.

Further, for firefox android, what does it mean that three is both `"version_added": false` and `"version_added": "41",`? I added the same note to the one that wasn't false.

ddbeck: If this is the specified behavior, then I think we describe the old behavior as exceptional. What do you think of inverting things like this?

```suggestion
                "notes": "Before Firefox 82, <code>setPointerCapture()</code> throws <code>InvalidPointerId</code> for an invalid <code>pointerId</code> argument. From Firefox 82, it throws <a href='https://w3c.github.io/pointerevents/#setting-pointer-capture'>the specified</a> <code>NotFoundError</code> exception. See <a href='https://bugzil.la/1662124'>bug 1662124</a>."
```

api/Element.json
```diff
@@ -8147,7 +8148,8 @@
                     "name": "dom.w3c_pointer_events.enabled",
                     "value_to_set": "true"
                   }
-                ]
+                ],
+                "notes": "From Firefox 82 throws <code>NotFoundError</code> exception if pointer id is invalid (as per <a href='https://w3c.github.io/pointerevents/#setting-pointer-capture'>WC3 PointerEvents specification</a>). Earlier Firefox versions throw <code>InvalidPointerId</code>. See <a href='https://bugzil.la/1662124'>bug 1662124</a>."
```
ddbeck, line 8152: We could do the same note here. Though I'm pretty suspicious of the existing data. Seems unlikely to me that the API is actually unsupported on Android for such a long time. Might be worth double checking whether we should mirror all the data from the desktop version.

```suggestion
                "notes": "Before Firefox 82, <code>setPointerCapture()</code> throws <code>InvalidPointerId</code> for an invalid <code>pointerId</code> argument. From Firefox 82, it throws <a href='https://w3c.github.io/pointerevents/#setting-pointer-capture'>the specified</a> <code>NotFoundError</code> exception. See <a href='https://bugzil.la/1662124'>bug 1662124</a>."
```

ddbeck (changes requested):
A couple questions for you, @hamishwillee. Thank you!

api/Element.json
hamishwillee: @ddbeck Three are arguments either way, but on balance I think you're right.

hamishwillee: @ddbeck Further, is adding notes to the version_added the right thing to do? You might argue this note could belong to all versions, or be added to a new version_added entry with value 82. [ie. just confirming that doing it the way I have is "the normal way"]

api/Element.json
hamishwillee: Good catch. I have raised the question on https://bugzilla.mozilla.org/show_bug.cgi?id=1662124#c9. Will commit both of these when that is confirmed.

api/Element.json
```diff
@@ -8137,7 +8138,8 @@
             ],
             "firefox_android": [
               {
-                "version_added": false
+                "version_added": true,
```
hamishwillee, line 8141: @ddbeck @chrisdavidmills  I've had [an answer](https://bugzilla.mozilla.org/show_bug.cgi?id=1662124#c10)  from engineer about pointer capture support on FF for Android:

> The story for FF Android is a bit different, we didn't ship pointer event on FF Android along with FF Desktop from 59.
We shipped it while we move FF Android to GeckoView, see https://bugzilla.mozilla.org/show_bug.cgi?id=1507495. But since it depend on runtime status on FF Android, so we need to update the pref dynamically.

I tested the latest version of FF for Android on Google Play and it pointer events works (which is why it is added as true) here.
- Do you understand what the engineer means by " we need to update the pref dynamically."?
- In terms of the version it went in as supported we know that it wasn't 59. bug 1507495](https://bugzilla.mozilla.org/show_bug.cgi?id=1507495), indicates it comes in with Gecko in Gecko 70. Do you know how I can find out the mapping between gecko versions and firefox for Android versions. Otherwise I guess we could merge as true?

hamishwillee: @ddbeck If not, then I propose we accept this as is with "true". 

Though *personally*. I'd find it more useful to have the current version than just true, since going forward people would know that it worked in version 82, while if you put in "true" then all they know is that it works in the current version.

ddbeck: @hamishwillee Thank you for your patience while I came back to this. Some answers, finally!

> Do you understand what the engineer means by " we need to update the pref dynamically."?

No, that's mysterious to me. That said, the issue you mentioned after this seems to settle the question of "when did this ship to all users?"

> …[bug 1507495](https://bugzilla.mozilla.org/show_bug.cgi?id=1507495), indicates it comes in with Gecko in Gecko 70. Do you know how I can find out the mapping between gecko versions and firefox for Android versions.

We actually include this in BCD's browser data! For example, [`firefox_android.json`](https://github.com/mdn/browser-compat-data/blob/master/browsers/firefox_android.json). This is very helpful for syncing up, for example, Opera desktop and Android, which have some slippage between them. Though, helpfully, Gecko versions are equivalent to Firefox versions from version 5 onward.

> Though personally. I'd find it more useful to have the current version than just true, since going forward people would know that it worked in version 82, while if you put in "true" then all they know is that it works in the current version.

Yeah, I get where you're coming from. In terms of BCD's schema, however, `true` always means "unsupported, but initial version unknown". `true` (and `null`) versions are incomplete, but recognizably so (we even track this sort of thing, see https://github.com/mdn/browser-compat-data/issues/3555#issuecomment-698485184 for an example); subbing in the current version for `true` is silently wrong, which I actively avoid.

But that's academic in this case, since it sounds like 70 is a safe value to set here. 🏁

hamishwillee:
@ddbeck Updated and ready for review.

Thanks for info on FF > FF android > gecko mapping. Slowly building up an idea of how you find out various versions. 

> "subbing in the current version for true is silently wrong, which I actively avoid.". 

The problem with "true" is that it is unhelpful and never gets less unhelpful. i.e. you use this to tell you that it is supported in the current version, but because you don't know what versions it works in you'll still have to code as though it isn't supported. And everyone will have to do that forever, because in a year it is still just "true" and you still don't know what versions it works in.

By contrast, while it might be "wrong", if I put in version 83 now, then in a year people can say "hey, I don't care about version 83" so I'll just code as if this works. 

Anyway, I get your point. But what I was thinking is whether we might be able to indicate this with some new syntax. i.e. a prefix.

ddbeck:
A quick comment before I approve and merge:

> The problem with "true" is that it is unhelpful and never gets less unhelpful.

This is true about the data as a snapshot, but for the project over time this isn't really the case. There's an active effort to chase down `true` and `null` values and replace them with accurate version numbers. With each release, there's been a decline in what we've called "non-real" values; see https://github.com/mdn/browser-compat-data/issues/3555 for a detailed breakdown. I know that `true`/`null` values aren't ideal, but they give us a map to where we need to improve.

ddbeck (approved):
Looks good. Thank you! 🎉

PR #6844, opened 2020-10-06
Mark navigator.canShare() experimental and nonstandard

sideshowbarker:
This change marks the `navigator.canShare()` method as `experimental:true` and `standard_track:false,` per https://github.com/w3c/web-share/pull/134 and https://github.com/w3c/web-share/pull/134#issuecomment-703653048 in particular. https://github.com/w3c/web-share/commit/2bbde9c dropped it as a specified feature (along with the entire Web Share Level 2 spec.)

### DIFF
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -424,8 +424,8 @@
             }
           },
           "status": {
-            "experimental": false,
-            "standard_track": true,
+            "experimental": true,
+            "standard_track": false,
             "deprecated": false
           }
         }


saschanaz:
I think nonstandard things are just nonstandard, they shouldn't be marked as experimental: #6812

sideshowbarker:
> I think nonstandard things are just nonstandard, they shouldn't be marked as experimental: #6812

I don’t really have a strong opinion about that either way. But as far as the MDN side of this goes, it seems like there’s value to web developers in knowing both whether something’s experimental or not, and also knowing whether something’s nonstandard. And the difference it’s just hypothetical; I believe we have cases of features that are nonstandard which aren’t experimental.

ddbeck:
Relatedly, some questions on experimental: #6905. I kinda wish we had originally named experimental something like "stable." Supposing my guideline gets some support, I guess the question here would be, as a Web developer, should I expect `navigator.canShare()` to change in the future?

sideshowbarker:
> Relatedly, some questions on experimental: #6905. I kinda wish we had originally named experimental something like "stable." Supposing my guideline gets some support, I guess the question here would be, as a Web developer, should I expect `navigator.canShare()` to change in the future?

I think the answer to that question is Yes — based on https://github.com/w3c/web-share/pull/134#issuecomment-554841908, which says:

> we don't have consensus on `canShare()` or the shape of that API

…and on https://github.com/w3c/web-share/issues/108:

> I wonder if we can simplify canShare() a bit... right now, canShare() conflates type checking behavior by intertwining the method call with share(). This is problematic because share() returns a promise, not TypeErrors.

…and from reading the rest of the discussion in that issue, it seems clear one possibility is that the `canShare()` API could end up being changed in way to address the problems surfaced in that issue. (Another possibility is that `canShare()` will be replaced by an API with a different name — but that’s a different matter.)

In other words the gist of it is that the `canShare()` API is not stable.

ddbeck:
OK, thank you, @sideshowbarker. I think this will end up being experimental no matter what we ultimately decide in #6905, so I'm going to approve and merge this.

ddbeck (approved):

PR #6845, opened 2020-10-06
MediaSession enabled all builds from FF82

hamishwillee:
MediaSession (and friends) are enabled on all builds from FF82 - this adds the version. See bug [Enable MediaSession API](https://bugzilla.mozilla.org/show_bug.cgi?id=1665496)

- The fix enables everything that is conditional on `dom.media.mediasession.enabled` so set covered is everything that was previously tagged with that preference.
- I have not added this for FF Android, as this was feature not previously supported for FF Android even via preference (am querying to make sure this is correct).
- The [spec is still marked as draft](https://w3c.github.io/mediasession/#the-mediasession-interface) so I did not change the experimental flag.

### DIFF
--- a/api/MediaMetadata.json
+++ b/api/MediaMetadata.json
@@ -29,7 +29,8 @@
             }
           ],
           "firefox_android": {
-            "version_added": null
+            "version_added": "82",
+            "notes": "The particular Media Session used for Android media control depends on the user agent. The Android browser implementation does not allow any session to be selected as active."
           },
           "ie": {
             "version_added": false
--- a/api/MediaMetadata.json
+++ b/api/MediaMetadata.json
@@ -31,7 +31,7 @@
           "firefox_android": {
             "version_added": "82",
             "partial_implementation": true,
-            "notes": "Firefox does not yet integrate this API into Android's media control interfaces. The API is exposed, but the corresponding user-facing behavior is not available."
+            "notes": "Firefox exposes the API, but does not provide a corresponding user-facing media control interface."
           },
           "ie": {
             "version_added": false
--- a/api/MediaMetadata.json
+++ b/api/MediaMetadata.json
@@ -30,7 +30,8 @@
           ],
           "firefox_android": {
             "version_added": "82",
-            "notes": "The particular Media Session used for Android media control depends on the user agent. The Android browser implementation does not allow any session to be selected as active."
+            "partial_implementation": true,
+            "notes": "Firefox does not yet integrate this API into Android's media control interfaces. The API is exposed, but the corresponding user-facing behavior is not available."
           },
           "ie": {
             "version_added": false
--- a/api/MediaSession.json
+++ b/api/MediaSession.json
@@ -29,7 +29,8 @@
             }
           ],
           "firefox_android": {
-            "version_added": false
+            "version_added": "82",
+            "notes": "The particular Media Session used for Android media control depends on the user agent. The Android browser implementation does not allow any session to be selected as active."
           },
           "ie": {
             "version_added": false
@@ -88,7 +89,8 @@
               }
             ],
             "firefox_android": {
-              "version_added": false
+              "version_added": "82",
+              "notes": "The particular Media Session used for Android media control depends on the user agent. The Android browser implementation does not allow any session to be selected as active."
             },
             "ie": {
               "version_added": false
@@ -148,7 +150,8 @@
               }
             ],
             "firefox_android": {
-              "version_added": false
+              "version_added": "82",
+              "notes": "The particular Media Session used for Android media control depends on the user agent. The Android browser implementation does not allow any session to be selected as active."
             },
             "ie": {
               "version_added": false
@@ -209,7 +212,8 @@
               }
             ],
             "firefox_android": {
-              "version_added": false
+              "version_added": "82",
+              "notes": "The particular Media Session used for Android media control depends on the user agent. The Android browser implementation does not allow any session to be selected as active."
             },
             "ie": {
               "version_added": false
@@ -270,7 +274,8 @@
               }
             ],
             "firefox_android": {
-              "version_added": false
+              "version_added": "82",
+              "notes": "The particular Media Session used for Android media control depends on the user agent. The Android browser implementation does not allow any session to be selected as active."
             },
             "ie": {
               "version_added": false
--- a/api/MediaSession.json
+++ b/api/MediaSession.json
@@ -309,68 +309,6 @@
             "standard_track": true,
             "deprecated": false
           }
-        },
-        "MediaPositionState": {
-          "__compat": {
-            "description": "Dictionary passed to <code>setPositionState</code> to specify play position within a track (<code>{duration, playbackRate, position}</code>).",
-            "support": {
-              "chrome": {
-                "version_added": "73"
-              },
-              "chrome_android": {
-                "version_added": "57"
-              },
-              "edge": {
-                "version_added": "≤79"
-              },
-              "firefox": [
-                {
-                  "version_added": "82"
-                },
-                {
-                  "version_added": "71",
-                  "flags": [
-                    {
-                      "type": "preference",
-                      "name": "dom.media.mediasession.enabled",
-                      "value_to_set": "true"
-                    }
-                  ]
-                }
-              ],
-              "firefox_android": {
-                "version_added": "82",
-                "partial_implementation": true,
-                "notes": "Firefox exposes the API, but does not provide a corresponding user-facing media control interface."
-              },
-              "ie": {
-                "version_added": false
-              },
-              "opera": {
-                "version_added": true
-              },
-              "opera_android": {
-                "version_added": false
-              },
-              "safari": {
-                "version_added": false
-              },
-              "safari_ios": {
-                "version_added": false
-              },
-              "samsunginternet_android": {
-                "version_added": "7.0"
-              },
-              "webview_android": {
-                "version_added": false
-              }
-            },
-            "status": {
-              "experimental": true,
-              "standard_track": true,
-              "deprecated": false
-            }
-          }
         }
       }
     }
--- a/api/MediaSession.json
+++ b/api/MediaSession.json
@@ -30,7 +30,8 @@
           ],
           "firefox_android": {
             "version_added": "82",
-            "notes": "The particular Media Session used for Android media control depends on the user agent. The Android browser implementation does not allow any session to be selected as active."
+            "partial_implementation": true,
+            "notes": "Firefox does not yet integrate this API into Android's media control interfaces. The API is exposed, but the corresponding user-facing behavior is not available."
           },
           "ie": {
             "version_added": false
@@ -90,7 +91,8 @@
             ],
             "firefox_android": {
               "version_added": "82",
-              "notes": "The particular Media Session used for Android media control depends on the user agent. The Android browser implementation does not allow any session to be selected as active."
+              "partial_implementation": true,
+              "notes": "Firefox does not yet integrate this API into Android's media control interfaces. The API is exposed, but the corresponding user-facing behavior is not available."
             },
             "ie": {
               "version_added": false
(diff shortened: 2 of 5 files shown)


ddbeck:
> I have not added this for FF Android, as this was feature not previously supported for FF Android even via preference (am querying to make sure this is correct).

@hamishwillee I've marked this as info-needed and not-ready, pending an answer on this bit. Or we can move on to reviewing without it, if you want to open an issue to track that question separately. Thank you!

hamishwillee:
@ddbeck Thanks very much. I have [asked the engineer](https://bugzilla.mozilla.org/show_bug.cgi?id=1665496).

hamishwillee:
@ddbeck On a closely related matter, would appreciate your advice. [Navigator.mediaSession](https://wiki.developer.mozilla.org/en-US/docs/Web/API/Navigator/mediaSession) is how the spec says you get hold of a [MediaSession](https://wiki.developer.mozilla.org/en-US/docs/Web/API/MediaSession#Browser_compatibility). So to my mind there really is no way the compatibility for these two objects should not match. You wouldn't invent a non-spec way to access everything else 'according to spec'.

I think that probably `Navigator.mediaSession` just wasn't updated last time the mediaSession was.

So I propose to also make these match. Reasonable?

image

hamishwillee:
@ddbeck I [asked the engineer](https://bugzilla.mozilla.org/show_bug.cgi?id=1665496) about this change on FF. 

He says that it is enabled but it won't affect the media control on Android, because Fenix is still using their own android component. What that means is that you should be able to call the API without error, but the device UI won't reflect the changes you request. i.e. a kind of silent fail. 

What is the right thing to do in this case - where it is in platform but won't work? I THINK probably best to add it as version-supported in FF82 on Android but add a note like the above.

chrisdavidmills:
@hamishwillee this looks OK to me; I'd move forward with that change.

chrisdavidmills:
@hamishwillee in the list of files updated, I don't see `Navigator` listed. Surely `Navigator.mediaSession` should also be updated to show support from Fx version 82?

chrisdavidmills:
@hamishwillee I also noticed that this PR does not include data for the [`MediaPositionState`](https://developer.mozilla.org/en-US/docs/Web/API/MediaPositionState) and [`MediaSessionActionDetails`](https://developer.mozilla.org/en-US/docs/Web/API/MediaSessionActionDetails) — those pages don't seem to have any browser-compat-data at all. Can you add it as part of this PR?

hamishwillee:
> @hamishwillee in the list of files updated, I don't see Navigator listed. Surely Navigator.mediaSession should also be updated to show support from Fx version 82?

@chrisdavidmills That was my question in https://github.com/mdn/browser-compat-data/pull/6845#issuecomment-705855398 - I think it was missed last time. I'll just make them match and the reviewer can argue it.

hamishwillee:
> @hamishwillee I also noticed that this PR does not include data for the MediaPositionState and MediaSessionActionDetails — those pages don't seem to have any browser-compat-data at all. Can you add it as part of this PR?

@chrisdavidmills @ddbeck I added `MediaPositionState` by pulling in #6002 and updating with the new information (and a fix or 2).

`MediaSessionActionDetails` was already documented [here](https://github.com/mdn/browser-compat-data/pull/6845/files#diff-0896cc10ab676a9871cafeb8debd21a95196fd23dfcb76168127065390b8f013) and table is in the docs:
image

@ddbeck This should be ready to review

hamishwillee:
@ddbeck This is just a reminder. I am not nagging. Honest :-)

ddbeck:
@hamishwillee Thanks for the nudge. I was more or less out last week, so it might take me a little time to catch up on things. I started on reviewing this before, but the backstory (e.g., #6002) is so complicated and angsty that it's slowing going. Thanks for your patience!

hamishwillee:
Still not nagging :-).

api/MediaMetadata.json
```diff
@@ -13,18 +13,24 @@
...
+                  "name": "dom.media.mediasession.enabled",
+                  "value_to_set": "true"
+                }
+              ]
+            }
+          ],
           "firefox_android": {
-            "version_added": null
+            "version_added": "82",
+            "notes": "The particular Media Session used for Android media control depends on the user agent. The Android browser implementation does not allow any session to be selected as active."
```
ddbeck, line 33: I'm having a hard time understanding this note. I ended up with a bunch of questions.

What does "depends on the user agent" mean here? Isn't Firefox the user agent? What about "does not allow any session to be selected as active"? This feature is the `MediaMetadata` interface itself. How does one use the interface to "select" something?

I need some help here to figure out how to translate this into something that a web developer can act on. `MediaMetadata` apparently has some limitation on Android. As developer, what can I do with it? What can't I do with it? How is it like other implementations? How is it different?

I see now that this note appears in several places. When is it most important to developers? When are they most likely to notice the behavior on Android?

api/MediaPositionState.json
```diff
@@ -0,0 +1,248 @@
+{
+  "api": {
+    "MediaPositionState": {
```
ddbeck, line 3: OK, this steps into a complex question about whether and how to represent IDL dictionaries in BCD. In this case, since, as far as I can tell, `MediaPositionState` only matters for one method, we should move the subfeatures of `MediaPositionState` to `api.MediaSession.setPositionState` (e.g., described as `"<code>state.duration</code> argument"`) and get rid of this file.

Now, _why_ we should do this pretty complex. I'm sorry about this. If you're really curious, you can read #6002 and #6265, but it's a long story. Here's my highly-opinionated summary:

The Media Session spec names several abstract dictionaries as a convenience to spec authors and browser developers. However, these dictionary names don't have a concrete manifestation for web developers; they indirectly appear as patterns for concrete return values and method arguments. Consequently, it's an open question what it means to "support" an abstract dictionary.

That said, we don't have a good answer in BCD or on MDN for how to represent these things yet. There's way more discussion on this point in #7287. In the past (e.g., #6265), I've compromised when such dictionaries have touched many APIs, but I've resisted introducing them in cases like this one.

As far as I can tell, `MediaPositionState` represents one thing only: the structure of the object argument to `MediaSession.setPositionState()`. I don't know what it means to support `MediaPositionState`, but I do have a sense for what it means for `setPositionState()` to support `state.duration`. We can represent this information as subfeatures of the method and call it a day.

ddbeck (changes requested):
Thanks for your reminders, @hamishwillee. The actual version data looks unobjectionable. 👍 There are a couple of structural hurdles here, however (not your fault—just a complexity we have to face here). I've got a couple big line comments to get started on those. Thanks again!

api/MediaMetadata.json
hamishwillee: @ddbeck What this says is that the API has been implemented but really can't be used. However it code running in a page that uses this API will never know that. Explanation:

The theory is that pages use this API to be able to expose media data and control to a user agent. In theory a user should be able to access an interface to select from pages using the API (and other media sources) and set one as being used by standard mechanisms for media display/control on the device. In other words, if your device supports playing media from the lockscreen (and displaying the current song say) then you should be able to select one of the pages as the source for media.

But what has happened is that the API was implemented, but the components required to make that UI selection to assign a page to the lockscreen (say) do not work/exist. So the API "works" - you can write a page that uses it. But then you still can't select the page.

An subtle point here though is that many pages might use the API, and most of them will not be selected by the user agent. No page actually knows if it is being used.  There is nothing a developer can do about this. 

The text is what the developer thought made sense to cover this. It is strictly accurate. But perhaps we should be blunt and say - there is no point using this?

Maybe something more like (?):

> At time of writing the Android browser implementation does not allow web pages to be set as a media source. This API can therefore be used, but not tested.

api/MediaSession.json
```diff
@@ -270,7 +274,8 @@
               }
             ],
             "firefox_android": {
-              "version_added": false
+              "version_added": "82",
```
hamishwillee, line 277: @ddbeck With respect to your comment on `MediaPositionState` here: https://github.com/mdn/browser-compat-data/pull/6845#discussion_r521497837

I have deleted the specific doc for the `MediaPositionState`. There is no code change here, so arguably we could just ignore adding anything for this - I added it because I though it was an omission. If we do need subfeature documentation, could you point me at a good example of how this would be done in BCD that I can copy.

api/MediaMetadata.json
ddbeck: Ah, OK. Thank you, @hamishwillee! That's really helpful. I think this helps me understand what's going on here a bit better. I think we need to revise the note and tweak the data a bit.

I think your draft note is pretty good start. To help this note standalone better in multiple contexts (MDN's not the only BCD consumer), does something like this revision work?

> Firefox does not implement a user interface for choosing a web page as a media source. This API is exposed, but has no corresponding user-facing behavior.

That seems like a big enough caveat that we'd also want to set `"partial_implementation": true`.

api/MediaSession.json
ddbeck: A quick search turned up something like `api.HTMLOrForeignElement.focus.preventScroll_option` ([MDN](https://developer.mozilla.org/en-US/docs/Web/API/HTMLOrForeignElement/focus)). The equivalent here would be something like `api.MediaSession.setPositionState.{duration,playbackRate,position}`, or, to preserve the top-level `MediaPositionState`, `api.MediaSession.setPositionState.MediaPositionState.{duration,playbackRate,position}`.

Also an unrelated heads up: you don't need to rebase PRs on BCD. We squash most merges anyway.

api/MediaMetadata.json
hamishwillee: > Firefox does not implement a user interface for choosing a web page as a media source. This API is exposed, but has no corresponding user-facing behavior.

I will go back and check with engineering because it may be that the missing part of the picture is outside of the browser/in the OS; in which case this feels wrong. That then would also make `"partial_implementation": true` feel wrong.

api/MediaSession.json
hamishwillee: Thanks @ddbeck . I think I've done this bit "as expected" now. Note again that I've copied in the same note as we discussed above - and I'm not sure this is appropriate within this level of nesting.

api/MediaMetadata.json
hamishwillee: @ddbeck An excellent description of the [integration added here](https://bugzilla.mozilla.org/show_bug.cgi?id=1665496#c13). In summary, it is as we thought - this API is integrated into GeckoView, but anything that uses GeckoView must further integrate with the underlying platform via the embedding application - in this case Firefox Android.

I have therefore taken your advice and changed to partial implementation and added note as below in all locations.

> Firefox does not yet integrate this API into Android's media control interfaces. Hence the API is exposed, but the corresponding user-facing behavior is not available.

Should be good to review again.

api/MediaSession.json
```diff
@@ -299,6 +309,68 @@
             "standard_track": true,
             "deprecated": false
           }
+        },
+        "MediaPositionState": {
+          "__compat": {
+            "description": "Dictionary passed to <code>setPositionState</code> to specify play position within a track (<code>{duration, playbackRate, position}</code>).",
```
ddbeck, line 315: Yeah, I see what you mean here. IIRC, you had subfeatures for duration, playbackRate, position in your original PR (that data seems to have disappeared with a rebase, so I don't know this for a fact). Seeing this singular `MediaPositionState` feature though, without anything underneath, I don't think it adds anything to the parent feature—it's effectively saying, "this method has an argument, which is an object".

My inclination here is to drop this feature and move on.

ddbeck (changes requested):
OK, I think we're getting pretty close here. Some hopefully straight-forward changes to make to get this in. Thank you, @hamishwillee for helping to fix up this (sometimes agitating) set of features.

api/MediaSession.json
hamishwillee: Completely agree. Done!

hamishwillee:
@ddbeck Thanks very much. Deleted the dictionary and accepted your note (yes, much better).

ddbeck (approved):
Finally! This was a hard one, but you stuck with it. Thank you, @hamishwillee! 🎉

hamishwillee:
Thanks for your patience. Possibly the most important skill for a tech writer - even more than "writing good"  :-)

Issue #6846, opened 2020-10-06
place-self supported in Safari 14.0 (or earlier?)


janosh:
According to https://caniuse.com/mdn-css_properties_place-self_flex_context which sources its data here, `place-self` is not supported in any version of Safari.

However, testing with Safari 14.0 on macOS 10.15.7, `place-self` is working just fine as shown here:

![place-self](https://user-images.githubusercontent.com/30958850/94992820-bed32100-058c-11eb-95da-43e4c7c8ecd5.png)

`place-items` and `place-content` have been supported since Safari 11.0 so either there's been a mistake here and `place-self` has been supported for a while as well or support for it was added in the update to 14.0 on Sep 17, 2020.

## HTML for testing

```html
<!DOCTYPE html>
<html lang="en">
  <head>
    <meta charset="UTF-8" />
    <meta name="viewport" content="width=device-width, initial-scale=1.0" />
    <title>Document</title>
  </head>
  <body style="display: grid; height: 100vh">
    <div
      style="width: 100px; height: 100px; background: black; place-self: center"
    ></div>
  </body>
</html>
```

janosh:
Looks like this data was added in #711.

janosh:
Related issue over at Fyrd/caniuse#5638.

sideshowbarker:
Yeah — thanks for catching this and taking time to report it. I think `place-self` has been in Safari for relatively quite a while now — several releases back. Anyway, I’ll investigate and raise a PR with an update (after I figure out the exact release).

sideshowbarker:
Raised https://github.com/mdn/browser-compat-data/pull/6847

PR #6847, opened 2020-10-06
Mark CSS place-self supported in Safari 11+

sideshowbarker:
This change marks the CSS `place-self` property as supported in Safari 11, per https://developer.apple.com/safari/technology-preview/release-notes/#r31 (TP 31) and https://trac.webkit.org/changeset/216829/webkit (2017-05-13).

Fixes https://github.com/mdn/browser-compat-data/issues/6846

### DIFF
--- a/css/properties/place-self.json
+++ b/css/properties/place-self.json
@@ -32,10 +32,10 @@
                 "version_added": "43"
               },
               "safari": {
-                "version_added": false
+                "version_added": "11"
               },
               "safari_ios": {
-                "version_added": false
+                "version_added": "11"
               },
               "samsunginternet_android": {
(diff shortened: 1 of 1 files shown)


css/properties/place-self.json
```diff
@@ -32,10 +32,10 @@
                 "version_added": "43"
               },
               "safari": {
-                "version_added": false
+                "version_added": "11"
```
foolip, line 35: Using http://mdn-bcd-collector.appspot.com/tests/css/properties/place-self I see it's not supported in Safari 10.1 and is supported in 11.1, so this seems right 👍

foolip (approved):

PR #6848, opened 2020-10-06
Refactor statistics script for importability

vinyldarkscratch:
This PR refactors the statistics script to allow it to be imported as a module as well, moving the stats collection into a `getStats()` function (which is exported) and making `printStats()` just a console output function.  This will simplify the development of a little GitHub Actions bot @foolip will be working on.

### DIFF
--- a/scripts/statistics.js
+++ b/scripts/statistics.js
@@ -18,8 +18,9 @@ const { argv } = require('yargs').command(
         default: '',
       })
       .option('all', {
+        alias: 'a',
         describe: 'Show statistics for all browsers within BCD',
-        type: 'flags',
+        type: 'boolean',
         nargs: 0,
       });
   },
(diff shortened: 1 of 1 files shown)


ddbeck (commented):
> This will simplify the development of a little GitHub Actions bot @foolip will be working on.

Can I get some elaboration on how the stats are going to be used? If this is being consumed elsewhere, it would be interesting to know the use case in more detail. In the past, we've made changes like this to decompose BCD's own tools for use internal to BCD. This is a bit different—it might even merit exposing this is a proper external API, if it's being consumed externally.

When it comes to the actual changes, I'm finding this monolithic commit difficult to follow.

Explain it to me like it's late on a Friday: why does this do more than export the `stats` object? Why does this rename `printTable` to `printStats`? Why does the `processData` signature change? A lot happens in one commit, so I need your help to understand the story behind each of the changes you've made here.

vinyldarkscratch:
Sure thing!  Sorry, I'm not the greatest at explaining changes, haha... 😅   Okay, let's big this a shot:

This PR refactors the script to allow for importability into other code.  First, it moves all statistics generation into a `getStats` function, which returns a stats object based upon the folder and browsers specified.  The `getStats` function is exposed as the module export, allowing other scripts to obtain the raw JSON data.  All of the console logging to display the stats is moved into `printStats` (combining it and `printTable`), which takes the statistics as an argument (rather than a global variable) generated by `getStats`.

Does that help?

ddbeck:
> Does that help?

Yes, it does. Thank you!

(At least in my experience, I get a lot out of recording lots of little commits along the way—it's like a logbook so I can remember _how_ to write the description. It's a lifesaver when I have to switch between tasks. 😄)

ddbeck (approved):
Looks good. Thank you, @vinyldarkscratch.

Also, just to make sure we didn't introduce anything unintended, I checked out this branch in a worktree, merged in `master`, then ran the branch stats script and `master` script, and diffed the two outputs. They both do the same thing. 👍

PR #6849, opened 2020-10-06
Update Safari onvisibilitychange support

vinyldarkscratch:
This PR fixes data for the `visibilitychange` event and handler for Safari (similar to #6833 for IE/Edge).  As determined by manual testing (after reviewing results from the mdn-bcd-collector), Safari had no support for `onvisibilitychange` until Safari 10.1.  However, it supported the event since Safari 7, through `addEventListener`.  Furthermore, the notes that were there are not accurate -- the event fired when transitioning to `hidden` no problem.

Test code used:
```js
document.addEventListener("visibilitychange", function() {
  if (document.visibilityState === 'visible') {
    console.log('addEventListener: visible');
  } else {
    console.log('addEventListener: hidden');
  }
});

  if (document.visibilityState === 'visible') {
    console.log('onvisibilitychange: visible');
  } else {
    console.log('onvisibilitychange: hidden');
  }
};
```

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -10788,13 +10788,11 @@
             },
             "safari": {
               "version_added": "7",
-              "partial_implementation": true,
-              "notes": "Doesn't fire the <code>visibilitychange</code> event when <code>document.visibilityState</code> transitions to <code>hidden</code>, so also include code to check for the <code>pagehide</code> event (which does fire for that case in all current browsers). See WebKit bugs <a href='https://webkit.org/b/116769'>116769</a>, <a href='https://webkit.org/b/151234'>151234</a>, <a href='https://webkit.org/b/151610'>151610</a>, and <a href='https://webkit.org/b/194897'>194897</a>."
+              "notes": "The <code>onvisibilitychange</code> attribute was not supported until Safari 10.1. To listen to this event in earlier versions of Edge, use <code>document.addEventListener('visibilitychange', function() {});</code>."
             },
             "safari_ios": {
               "version_added": "7",
-              "partial_implementation": true,
-              "notes": "Doesn't fire the <code>visibilitychange</code> event when <code>document.visibilityState</code> transitions to <code>hidden</code>, so also include code to check for the <code>pagehide</code> event (which does fire for that case in all current browsers). See WebKit bugs <a href='https://webkit.org/b/116769'>116769</a>, <a href='https://webkit.org/b/151234'>151234</a>, <a href='https://webkit.org/b/151610'>151610</a>, and <a href='https://webkit.org/b/194897'>194897</a>."
+              "notes": "The <code>onvisibilitychange</code> attribute was not supported until Safari iOS 10.3. To listen to this event in earlier versions of Edge, use <code>document.addEventListener('visibilitychange', function() {});</code>."
             },
             "samsunginternet_android": {
               "version_added": "2.0"
--- a/api/Document.json
+++ b/api/Document.json
@@ -7618,11 +7618,11 @@
             },
             "safari": {
               "version_added": "10.1",
-              "notes": "Before Safari 10.1, this event handler attribute was not supported, however the event itself was supported since Safari 7. The event can be listened to via <code>document.addEventListener('visibilitychange', function() {});</code>."
+              "notes": "Before Safari 10.1, this event handler attribute was not supported; however, the event itself was supported since Safari 7. The event can be listened to via <code>document.addEventListener('visibilitychange', function() {});</code>."
             },
             "safari_ios": {
               "version_added": "10.3",
-              "notes": "Before Safari iOS 10.3, this event handler attribute was not supported, however the event itself was supported since Safari iOS 7. The event can be listened to via <code>document.addEventListener('visibilitychange', function() {});</code>."
+              "notes": "Before Safari iOS 10.3, this event handler attribute was not supported; however, the event itself was supported since Safari iOS 7. The event can be listened to via <code>document.addEventListener('visibilitychange', function() {});</code>."
             },
             "samsunginternet_android": {
               "version_added": "2.0"
--- a/api/Document.json
+++ b/api/Document.json
@@ -7618,11 +7618,17 @@
             },
             "safari": {
               "version_added": "10.1",
-              "notes": "Before Safari 10.1, this event handler attribute was not supported; however, the event itself was supported since Safari 7. The event can be listened to via <code>document.addEventListener('visibilitychange', function() {});</code>."
+              "notes": [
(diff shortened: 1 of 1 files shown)


sideshowbarker:
> Furthermore, the notes that were there are not accurate -- the event fired when transitioning to `hidden` no problem.

In which versions of Safari?

@cdumez, can you comment? The note I added has this:

> Doesn't fire the visibilitychange event when document.visibilityState transitions to hidden, so also include code to check for the pagehide event (which does fire for that case in all current browsers). See WebKit bugs <a href='https://webkit.org/b/116769'>116769</a>, <a href='https://webkit.org/b/151234'>151234</a>, <a href='https://webkit.org/b/151610'>151610</a>, and <a href='https://webkit.org/b/194897'>194897</a>.

vinyldarkscratch:
This has been the case since Safari 7, which is when the event was first supported.

sideshowbarker:
> This has been the case since Safari 7, which is when the event was first supported.

I’m confused by hearing that, given the patch in https://trac.webkit.org/changeset/267614/webkit didn’t land until September 25 (11 days ago), and that part of the description of that patch was this:

>  Fire a visibilitychange during document unload, as per the specification

… and examining patch chunk at https://trac.webkit.org/changeset/267614/webkit#file37 shows that prior to the patch, the part of the source that handles transitioning visibilityState to hidden clearly didn’t seem to have been dispatching the event:
```diff
Index: trunk/Source/WebCore/dom/Document.cpp
===================================================================
--- a/trunk/Source/WebCore/dom/Document.cpp
+++ b/trunk/Source/WebCore/dom/Document.cpp
@@ -1773,5 +1773,5 @@
     // that the page is hidden, as specified by the spec:
     // https://w3c.github.io/page-visibility/#visibilitystate-attribute
-    if (!m_frame || !m_frame->page())
+    if (!m_frame || !m_frame->page() || m_visibilityHiddenDueToDismissal)
         return VisibilityState::Hidden;
     return m_frame->page()->visibilityState();
@@ -3258,4 +3258,13 @@
 }
 
+void Document::setVisibilityHiddenDueToDismissal(bool hiddenDueToDismissal)
+{
+    if (m_visibilityHiddenDueToDismissal == hiddenDueToDismissal)
+        return;
+
+    m_visibilityHiddenDueToDismissal = hiddenDueToDismissal;
+    dispatchEvent(Event::create(eventNames().visibilitychangeEvent, Event::CanBubble::Yes, Event::IsCancelable::No));
+}
```

cdumez:
Until very recently, Safari/WebKit was only firing visibilitychange events when the view's state would become visible/hidden. As of very recently, we also fire the visibilitychange event upon navigating away (which means either unload or entering the back/forward cache).

cdumez:
Also WebKit has been supporting Document.onvisibilitychange for 4 years:
https://trac.webkit.org/changeset/209750/webkit

Unlike what is stated in the description.

sideshowbarker:
> Until very recently, Safari/WebKit was only firing visibilitychange events when the view's state would become visible/hidden. As of very recently, we also fire the visibilitychange event upon navigating away (which means either unload or entering the back/forward cache).

Thanks @cdumez — So, given all that, as far as our note to indicate to developers what the Safari behavior is prior to that change, it seems like we could make i accurate by updating it to this:

> [Safari] Doesn't fire the `visibilitychange` event when navigating way from a document, so also include code to check for the `pagehide` event (which does fire for that case in all current browsers). See WebKit bugs 116769, 151234, 151610, and 194897.

Or if the above isn’t accurate, maybe you can suggest alternate wording that is?

cdumez:
> > Until very recently, Safari/WebKit was only firing visibilitychange events when the view's state would become visible/hidden. As of very recently, we also fire the visibilitychange event upon navigating away (which means either unload or entering the back/forward cache).
> 
> Thanks @cdumez — So, given all that, as far as our note to indicate to developers what the Safari behavior is prior to that change, it seems like we could make i accurate by updating it to this:
> 
> > [Safari] Doesn't fire the `visibilitychange` event when navigating way from a document, so also include code to check for the `pagehide` event (which does fire for that case in all current browsers). See WebKit bugs 116769, 151234, 151610, and 194897.
> 
> Or if the above isn’t accurate, maybe you can suggest alternate wording that is?

Your proposal sounds good to me.

sideshowbarker (approved):

sideshowbarker:
I now notice something I missed when reviewing this patch: The patch removed the `partial_implementation` keys. Was that intentional? If not, I raised https://github.com/mdn/browser-compat-data/pull/6890 to restore them.

vinyldarkscratch:
Thanks for asking!  Yes, that was intentional, since the behavior is different than the original note.  Originally, it talked about any change to `hidden` rather than just page navigation, which would qualify as `partial_implementation` since it's such a critical part of the event.  Since it's on page navigation, however, I think that it's minimal enough that a note alone works.

sideshowbarker:
From https://github.com/mdn/browser-compat-data/blob/master/schemas/compat-data-schema.md#partial_implementation and for the record here:

> partial_implementation: A boolean value indicating whether or not the implementation of the sub-feature deviates from the specification in a way that may cause significant compatibility problems.

The Safari behavior certainly deviates from the specification — it’s non-conforming and isn’t interoperable with other browsers — otherwise the https://trac.webkit.org/changeset/267614/webkit code change wouldn’t have been necessary.

So I guess the question is whether the way in which it deviates may cause significant compatibility problems.

Given that, can you explain in more detail why you assess it to be minimal enough to not merit being marked partial?

I could understand the conclusion if we didn’t have any evidence that the deviation was causing real problems for developers in practice. But we do actually seem to have such evidence:

* https://www.ctrl.blog/entry/safari-beacon-issues.html
> The Page Visibility API gives us a visibilitychange event after the pagehide event. However, neither macOS Safari or iOS Safari fires this event when navigating away from a page. It’s only fired as intended when pages go in and out of view (minimizing, off-screen, etc.). Chromium has the same behavior, and only Firefox implements the full specification here.

Your only option is to use the pagehide event from Safari 12.2 and later, or to send your beacon earlier in the page lifecycle. This makes it difficult to track reading-depth and time-on-page, unfortunately. This is probably the event you want to use to capture page transitions in all browsers and on all platforms regardless.
* https://developers.google.com/web/updates/2018/07/page-lifecycle-api#developer-recommendations-for-each-state
> Safari does not reliably fire the <code translate="no" dir="ltr">pagehide</code> or <code translate="no" dir="ltr">visibilitychange</code> events when closing a tab (webkit bugs: <a href="https://bugs.webkit.org/show_bug.cgi?id=151610">151610</a> and <a href="https://bugs.webkit.org/show_bug.cgi?id=151234">151234</a>), so in Safari you may need to <em>also</em> listen to the <code translate="no" dir="ltr">beforeunload</code> event in order to detect a change to the hidden state. But since the <code translate="no" dir="ltr">beforeunload</code> event can be canceled, you need <a href="https://github.com/GoogleChromeLabs/page-lifecycle/blob/0.1.1/src/Lifecycle.mjs#L156-L172">to wait until after the event has finished propagating</a> to know if the state has changed to hidden. <strong>Important</strong>: using the <code translate="no" dir="ltr">beforeunload</code> event this way should <em>only</em> be done in Safari, as using this event in other browsers can hurt performance. See the <a href="#legacy-lifecycle-apis-to-avoid">legacy APIs</a> section for details.

In other words, the spec deviation seems to mean that developers are required to write and maintain a separate code path to get their applications to work as needed in Safari.

Is that not a significant compatibility problem that merits the `partial_implementation` flag?

Issue #6855, opened 2020-10-07
Use a test framework


foolip:
The test/ directory is a mix of tests that run on the BCD data itself to ensure style/consistency/etc., and at least one test for code itself, in test-utils.js. That file as this comment:
https://github.com/mdn/browser-compat-data/blob/b3cc680b9ebbb2ab4b56f71378f80e660abd2ff6/test/test-utils.js#L60

The test defines its own `it` helper, and would be easy to convert to any JS test frameworks.

The main candidates that come to mind are https://mochajs.org/ and https://jestjs.io/. Preferences, anyone?

I'm filing this issue because I'd like to write tests for some new code I'm writing in scripts/.

foolip:
To make this concrete, I've sent PRs showing what it would look like with each framework, https://github.com/mdn/browser-compat-data/pull/6856 for Jest and https://github.com/mdn/browser-compat-data/pull/6857 for Mocha. Both are failing because both require Node.JS 10, but they work locally.

My preference would be for Mocha, because of the fewer dependencies and that the output is already similar (green checkmarks) to what our lint script does.

Note that with Mocha we may end up wanting to pull in chai for more powerful asserts, and nyc for coverage, while Jest has a lot more built in.

vinyldarkscratch:
+1 for this.  I tried to get a discussion going in #5829, but sadly, nobody responded... :(

foolip:
Oh, I didn't know about that issue, I searched a bit but didn't find an existing one.

I'm not going to close either, I think that if this issue is fixed, https://github.com/mdn/browser-compat-data/issues/5829 would still remain open to actually add tests for the linter.

foolip:
@ddbeck @sideshowbarker your thoughts on this one would be appreciated.

sideshowbarker:
> My preference would be for Mocha, because of the fewer dependencies and that the output is already similar (green checkmarks) to what our lint script does.

I agree, for those same reasons.

That said, I know Jest is pretty widely used too — maybe even more widely used now than Mocha — and I wouldn’t object if others felt more strongly that Jest was the better choice.

> Note that with Mocha we may end up wanting to pull in chai for more powerful asserts, and nyc for coverage

In general, I don’t think adding additional dependencies is a problem for us in our case for BCD

> while Jest has a lot more built in.

I think the flip side of that is, Mocha arguably gives you more choices

ddbeck:
Thank you so much for opening these comparison PRs, @foolip!

[...]

PR #6856, opened 2020-10-07
Introduce Jest to test test/utils.js

foolip:
This increases the number of dependencies from 86 to 643.

Fixes https://github.com/mdn/browser-compat-data/issues/6855.


ddbeck:
I decided we should go with #6857 over this PR. See https://github.com/mdn/browser-compat-data/issues/6855#issuecomment-707904361 for details. Thank you for making the comparison easy to grasp, @foolip!

foolip:
Thanks @ddbeck, that's a choice I agree with 👍

PR #1742, opened 2018-04-05
Add Web Authentication API.

jpmedley:
I rearranged Navigator.json while I was in there because the file is still pretty small.

PR #6860, opened 2020-10-07
Removing outdated / duplicate data on WebAuthn implementation

dontcallmedom:
Actual data is available through PublicKeyCredential et al


foolip:
This was added in https://github.com/mdn/browser-compat-data/pull/1742 so I'm requesting review from @jpmedley.

ddbeck:
Hi @dontcallmedom, sorry for the wait on this. I don't think we need to wait longer on this. But #6897 seems to have (dubiously) updated this data. Would you mind rebasing, so we can see this one through?

(I think a bunch of the APIs touched in #6897 might be historic, actually—we should probably have a look at the WebAuthn and Crypto APIs more broadly.)

dontcallmedom:
rebased. I think the approach I used (but didn't document anywhere yet :) for https://github.com/mdn/content/issues/80 would likely also help identify historical BCD files.

ddbeck (approved):
Thank you!

> rebased. I think the approach I used (but didn't document anywhere yet :) for [mdn/content#80](https://github.com/mdn/content/issues/80) would likely also help identify historical BCD files.

Oh nice. If you're going to be making PRs against the post-launch Yari content, then I'd love to be @-ed when that happens (or get corresponding PRs to BCD), so we can update the BCD at the same time.

PR #6862, opened 2020-10-08
Prototype: npm run diff

foolip:


saschanaz:
It seems this currently requires a separate physical clone of BCD. How about using some git command to load previous state, if it's possible?

foolip:
This in fact requires two additional copies of BCD, or at least that's how I've been using it. In a CI setup, my thinking was to first check out the target branch and save the state of BCD there into a JSON file, and then check out the head branch merged with the target branch, treating that as the RHS. This will be easy enough to make work.

For local use, however, it would be nice to be able to just run `npm run diff` and show a summary of the local changes. The problem with this is that if any JavaScript has changed, then the interpretation of JSON files might have changed, and it might not make sense to interpret the JSON files from another checkout using the JavaScript of this checkout. So far that mode, I think it would have to fail or print a warning if anything other than JSON files has changed.

Accepting that limitation, then a bunch of `git` commands can indeed get the state to compare to.

ddbeck:
> The problem with this is that if any JavaScript has changed, then the interpretation of JSON files might have changed, and it might not make sense to interpret the JSON files from another checkout using the JavaScript of this checkout. So far that mode, I think it would have to fail or print a warning if anything other than JSON files has changed.

Perhaps if we were careful, going forward, we could use the package version to guard against this sort of thing? e.g., bake in an assumption that `npm run diff` can only act on the same major or major.minor version?

foolip:
@ddbeck yep, I think that would work.

For transparency, I don't know when I can make time to work on this, so if someone has all the pieces in their head and wants to go implement it, don't let this WIP PR stand in your way.

I think that https://github.com/mdn/yari/tree/master/client/src/document/ingredients/browser-compatibility-table is the new BCD table rendering for Yari, but I don't know if it can be used standalone.

saschanaz:
Thanks for the update, I implemented the git way in #7486. Still text-only.

foolip:
Thanks @saschanaz! I'll keep this open for the time being since I use it every day myself, but once the scaffolding is there I might rework some of this on top of your script.

queengooborg:
#7486 has now been merged -- thanks for getting the ball rolling on this script!

PR #6864, opened 2020-10-08
Mirror Chrome -> Chromium Android for Document API

vinyldarkscratch:
This PR mirrors Chrome to Chrome Android, Samsung Internet, and WebView Android for various members of the Document API.  This PR was created as a cherry-pick from #6836 to reduce the diff.

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -153,7 +153,7 @@
               "version_added": "1"
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": "18"
             },
             "edge": {
               "version_added": "12"
@@ -180,10 +180,10 @@
               "version_added": true
             },
             "samsunginternet_android": {
-              "version_added": true
+              "version_added": "1.0"
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "1"
             }
           },
           "status": {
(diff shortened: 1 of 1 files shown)


api/Document.json
```diff
@@ -867,10 +867,10 @@
               "version_added": "5"
             },
             "samsunginternet_android": {
-              "version_added": true
+              "version_added": "1.0"
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "≤37"
```
foolip, line 873: This is the only irregular change in this PR. It probably was in earlier WebView, but let's not chase down that version number.

foolip (approved):
All changes except the one commented on change Chrome Android to "18", Samsung Internet to "1.0" and WebView Android to "1". I've confirmed that all of the APIs touched are ones I recognize to be ancient or certainly not new in recent years, so all of this looks very plausible. I'm thus not taking a very close look at any individual edit.

foolip:
I'd like to merge this after https://github.com/mdn/browser-compat-data/pull/6836, in case it turns out that any of the data being mirrored here was wrong there.

foolip (approved):
Good to go now after https://github.com/mdn/browser-compat-data/pull/6836 was merged and I've confirmed none of the bits mentioned in https://github.com/mdn/browser-compat-data/pull/6836#issuecomment-706235380 were mirrored in this PR.

Issue #6839, opened 2020-10-05
Many notes on api.Document.* refer to the spec-only HTMLDocument alias


foolip:
Many entries in api.Document.* have notes like this one for api.Document.alinkColor:
> Before Chrome 64, this property was accessed through the <code>HTMLDocument</code> alias.

This refers to the `HTMLDocument` property being defined as an alias in https://html.spec.whatwg.org/multipage/window-object.html:
> For historical reasons, `Window` objects must also have a writable, configurable, non-enumerable property named `HTMLDocument` whose value is the `Document` interface object.

However, this aliasing has happened only in the spec, no implementation has followed, see https://github.com/whatwg/html/issues/4792.

It would be better if these notes refer to what is actually in implementations, and refer to the `HTMLDocument` interface.

@vinyldarkscratch I spotted these notes while reviewing https://github.com/mdn/browser-compat-data/pull/6820.

ddbeck:
@foolip, do you know if this is resolved? Were these all fixed by #6865?

foolip:
There is one note about an issue like this which still remains:
https://github.com/mdn/browser-compat-data/blob/b79d65e3abdc2d9b1986cae0fa0467561dc8ee5c/api/Document.json#L933-L940

It seems to me this can just be made consistent with the other notes.

foolip:
I've sent https://github.com/mdn/browser-compat-data/pull/10683 to fix that last case and filed https://github.com/mdn/browser-compat-data/issues/10682 about the inconsistency of it all :)

PR #6865, opened 2020-10-08
Update Chrome data for Document features that were on HTMLDocument alias

vinyldarkscratch:
This PR updates the Chromium data for various Document API features that were initially implemented via the `HTMLDocument` alias, first by obtaining the version numbers via the mdn-bcd-collector project, then separating them into sections with `partial_implementation`.  This is a cherry-pick from #6836, applying the review suggestions from https://github.com/mdn/browser-compat-data/pull/6836#discussion_r501736375.


api/Document.json
```diff
@@ -245,14 +245,28 @@
...
-            },
+            "chrome": [
+              {
+                "version_added": "64"
+              },
+              {
+                "version_added": "1",
+                "version_removed": "64",
+                "partial_implementation": true,
+                "notes": "Before Chrome 64, this property was accessed through the <code>HTMLDocument</code> alias."
```
foolip, line 256:
```suggestion
                "notes": "Only supported for <a href='https://developer.mozilla.org/docs/Web/API/HTMLDocument'><code>HTMLDocument</code></a>, not all <code>Document</code> objects."
```

This is in order to resolve https://github.com/mdn/browser-compat-data/issues/6839, and I've used the wording for api.Node.contains from https://github.com/mdn/browser-compat-data/pull/6850.

I'd like @ddbeck to review this situation though, since it'll be a lot more visible than that single entry.

foolip:
@ddbeck I'd like your review on the note wording for this.

api/Document.json
ddbeck: Yes, I prefer this new note text. I'm happy to see it applied elsewhere. 👍

vinyldarkscratch: Sounds good, I'll go ahead and apply the note text then!

api/Document.json
```diff
@@ -296,20 +348,30 @@
...
-              "version_added": "64",
-              "notes": [
-                "Starting in Chrome 65, this property is readonly.",
-                "Before Chrome 64, this property was accessed through the <code>HTMLDocument</code> alias."
-              ]
-            },
+            "chrome": [
+              {
+                "version_added": "64",
+                "notes": "Starting in Chrome 65, this property is readonly."
```
foolip, line 354: I don't think this is a useful note to have, but I'll remove it in a separate PR.

api/Document.json
```diff
@@ -265,26 +279,64 @@
...
-            },
-            "opera_android": {
-              "version_added": true
-            },
+            "opera": [
+              {
+                "version_added": "51"
+              },
+              {
+                "version_added": "15",
```
foolip, line 287:
```suggestion
                "version_added": "≤12.1"
```

Attributes aren't on the prototype in Presto, but you can tell that `document.all` was only supported for `HTMLDocument` by creating different documents using https://developer.mozilla.org/en-US/docs/Web/API/DOMImplementation/createDocument

As a future improvement to the collector it would be good to test all `Document` members using both an `HTMLDocument` and an `XMLDocument`, and also a plain `Document` if the `new Document()` constructor is supported.

foolip (commented):
The notes look good to me now, the only issue I can spot now is the Opera Presto ranges, which can probably all be collapsed.

api/Document.json
```diff
@@ -322,32 +376,68 @@
...
+                "notes": "Starting in Opera 52, this property is readonly."
+              },
+              {
+                "version_added": "15",
+                "version_removed": "51",
+                "partial_implementation": true,
+                "notes": "Only supported for <a href='https://developer.mozilla.org/docs/Web/API/HTMLDocument'><code>HTMLDocument</code></a>, not all <code>Document</code> objects."
+              },
+              {
+                "version_added": "≤12.1",
```
foolip, line 391: This and more Opera ranges that need to be merged into the `partial_implementation` range remain. Or are there some of these things that were actually on both `HTMLDocument` and `XMLDocument` in Presto?

foolip:
Ping @vinyldarkscratch. If there's something tricky to resolve here, maybe we can split out those changes?

api/Document.json
vinyldarkscratch: Ranges are now merged together!

foolip (approved):
I've done another lighter review, all looks good to me now!

PR #6868, opened 2020-10-09
Add real values for Chromium for WebGL2RenderingContext

vinyldarkscratch:
This PR updates the data for two features in WebGL2RenderingContext based upon results from the mdn-bcd-collector (and mirroring).

Tests used:
https://mdn-bcd-collector.appspot.com/tests/api/WebGL2RenderingContext/bufferSubData
https://mdn-bcd-collector.appspot.com/tests/api/WebGL2RenderingContext/getBufferSubData

### DIFF
--- a/api/WebGL2RenderingContext.json
+++ b/api/WebGL2RenderingContext.json
@@ -436,10 +436,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/WebGL2RenderingContext/bufferSubData",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "56"
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": "58"
             },
             "edge": {
               "version_added": "79"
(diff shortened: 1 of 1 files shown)


api/WebGL2RenderingContext.json
```diff
@@ -436,10 +436,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/WebGL2RenderingContext/bufferSubData",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "56"
```
foolip, line 439: OK, so the whole feature was shipped in Chrome 56 and Chrome for Android 58, and almost all of the rows in https://developer.mozilla.org/en-US/docs/Web/API/WebGL2RenderingContext#Browser_compatibility already match this. Makes sense.

api/WebGL2RenderingContext.json
```diff
@@ -2134,10 +2134,10 @@
               "version_added": false
             },
             "opera": {
-              "version_added": true
+              "version_added": "43"
```
foolip, line 2137: Also matches the other existing data.

foolip (approved):

PR #6870, opened 2020-10-09
Fix Console API version for Firefox

vinyldarkscratch:
This PR fixes the Firefox version for the Console API.  It was incorrectly marked as "2", however based upon manual testing, the `console` variable was undefined until Firefox 4.


foolip (approved):
This is also consistent with the subfeatures, the earliest of which like `console.log` are in Firefox 4.

PR #3141, opened 2018-11-27
Update Element based on confluence.

jpmedley:

PR #6874, opened 2020-10-09
Update compat data for insertAdjacent* methods

cvn:
Hello,

I've updated the Safari compatibility data for the Element.insertAdjacent* methods.

The existing data says that these methods were added in Safari 10 and iOS Safari 10, which were both released in September 2016. This disagreed with the caniuse data, so I did a little research.

- the existing data was committed in PR #3141
- https://caniuse.com/?search=insertadjacent says:
  - insertAdjacentElement and insertAdjacentText are present in Safari 3.1 (Mar 2008) and iOS Safari 3.2 (Apr 2010)
  - insertAdjacentHTML are present in Safari 4 (Jun 2009) and iOS Safari 4 (Sep 2010)
- https://bugs.webkit.org/show_bug.cgi?id=6520 says insertAdjacentElement was committed to WebKit in March 2006
- https://bugs.webkit.org/show_bug.cgi?id=17622 says insertAdjacentHTML and insertAdjacentText were committed in June 2008

The caniuse data seems more in line with the webkit bug tracker dates, with the exception of insertAdjacentText, which it says is present in a March 2008 version of Safari, even though the commit didn't land until June 2008. I attribute this to the fact that caniuse groups insertAdjacentElement and insertAdjacentText into [one entry](https://github.com/Fyrd/caniuse/blob/master/features-json/insert-adjacent.json). In this case, I used the same compat data as [insertAdjacentHTML](https://github.com/Fyrd/caniuse/blob/master/features-json/insertadjacenthtml.json), whose dates make more sense.

I don't have a way to test these old versions, so these changes are based entirely on what I've presented above.

Thanks for your time.

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -4789,10 +4789,10 @@
               "version_added": "10.1"
             },
             "safari": {
-              "version_added": "3.1"
+              "version_added": "3"
             },
             "safari_ios": {
-              "version_added": "3.2"
+              "version_added": "1"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
--- a/api/Element.json
+++ b/api/Element.json
@@ -4789,10 +4789,10 @@
               "version_added": "10.1"
             },
             "safari": {
-              "version_added": "10"
+              "version_added": "3.1"
             },
             "safari_ios": {
-              "version_added": "10"
+              "version_added": "3.2"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
@@ -4848,7 +4848,7 @@
               "version_added": "10.1"
             },
             "safari": {
-              "version_added": "10"
+              "version_added": "4"
             },
             "safari_ios": {
               "version_added": "4"
@@ -4904,10 +4904,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": "10"
+              "version_added": "4"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "4"
             },
             "samsunginternet_android": {
               "version_added": "1.0"


sideshowbarker (approved):

sideshowbarker:
I reviewed all the information at the links in the PR description, and it all looks correct — but for definitive confirmation, I believe @vinyldarkscratch does have a way to actually test older Safari versions.

cvn:
I remembered that I do have an old iPhone 1. If I'm reading the UA correctly, it's running iOS 3.1.3, with Mobile Safari 4.0.

I made a test page http://files.vonnau.com/fiddle/insert-adjacent/

All the methods appear to work, which confirms some of the data.

IMG_0424

api/Element.json
```diff
@@ -4789,10 +4789,10 @@
               "version_added": "10.1"
             },
             "safari": {
-              "version_added": "10"
+              "version_added": "3.1"
```
vinyldarkscratch, line 4792:
```suggestion
              "version_added": "3"
```

api/Element.json
```diff
@@ -4789,10 +4789,10 @@
...
             },
             "safari_ios": {
-              "version_added": "10"
+              "version_added": "3.2"
```
vinyldarkscratch, line 4795:
```suggestion
              "version_added": "1"
```

vinyldarkscratch (changes requested):
Looks good to me, thanks for the PR!  Just one little thing: I did some testing in older Safari versions and found it was implemented in Safari 3 rather than 3.1.  Can we adjust the version real quick?

vinyldarkscratch (approved):
Thanks, looks good to me!

cvn:
Thank you, @vinyldarkscratch and @sideshowbarker! MDN has been so valuable to me as a developer, it feels good to give something back, however small.

sideshowbarker:
@cvn thanks much and welcome to the MDN browser-compat-data project 🎉

PR #6883, opened 2020-10-11
Update Chromium data for api.MediaDevices.getUserMedia

vinyldarkscratch:
This PR was created based upon the mdn-bcd-collector. It was determined that this function was implemented in Chome 53 rather than 52 -- Chrome 52 implemented it behind a prefix. This also removes the note that talks about the prefixed support (which seems odd that it was added as a note), and removes old flag data.

### DIFF
--- a/api/MediaDevices.json
+++ b/api/MediaDevices.json
@@ -367,36 +367,22 @@
           "support": {
             "chrome": [
               {
-                "version_added": "52"
+                "version_added": "53"
               },
               {
-                "version_added": "47",
-                "version_removed": "52",
-                "flags": [
-                  {
-                    "type": "preference",
-                    "name": "Experimental Web Platform features",
-                    "value_to_set": "Enabled"
-                  }
-                ],
-                "notes": "Older versions of Chrome implement <code>navigator.webkitGetUserMedia</code>, a prefixed form of the legacy <a href='https://developer.mozilla.org/docs/Web/API/Navigator/getUserMedia'><code>navigator.getUserMedia</code></a> API."
+                "version_added": "52",
+                "version_removed": "53",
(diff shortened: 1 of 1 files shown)


foolip:
This is a dupe of https://github.com/mdn/browser-compat-data/pull/6658 which I've unfortunately left sitting for too long. I updated it just today before I got to this PR.

api/MediaDevices.json
```diff
@@ -367,36 +367,22 @@
...
-                  {
-                    "type": "preference",
-                    "name": "Experimental Web Platform features",
-                    "value_to_set": "Enabled"
-                  }
-                ],
-                "notes": "Older versions of Chrome implement <code>navigator.webkitGetUserMedia</code>, a prefixed form of the legacy <a href='https://developer.mozilla.org/docs/Web/API/Navigator/getUserMedia'><code>navigator.getUserMedia</code></a> API."
+                "version_added": "52",
+                "version_removed": "53",
+                "prefix": "webkit"
```
foolip, line 375: Hmm, so you're saying there was `navigator.mediaDevices.webkitGetUserMedia` for one release? I tested https://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8570 in Chrome 52 and couldn't confirm. The prefixed variant was `navigator.webkitGetUserMedia`, going back to much further in time.

foolip (commented):
I spotted a difference to https://github.com/mdn/browser-compat-data/pull/6658, but let's update that PR since it also has updated notes.

ddbeck:
I think this was resolved by #6658. I'm going to close this, but I've missed something, please @ me if I've made a mistake and we'll fix it.

PR #6886, opened 2020-10-12
window.name is reset on cross domain load - FF82

hamishwillee:
This adds a note to api.window.name that the `window.name` is reset (set to empty string) if a new page is loaded from a different domain - from FF82.

There is a very long and old bug on this https://bugzilla.mozilla.org/show_bug.cgi?id=444222 . A bit of context on the change can be found in https://github.com/mdn/sprints/issues/3731#issuecomment-704017095, but in summary
- About 12 years ago some people were storing data in windows.name because it has a lot of space and some people were advising this as a way for cross-domain sharing. They should not be.
- This is also used by some _frameworks_ as a method for cross domain messaging. There are now better ways to do this.
- A bad actor loaded into same tab might previously have accessed information from a page previously loaded into the same tab.
- A bad actor might also write to that data, and if old page is reloaded it could be impacted.

People really should not be adding data to window.name. I decided primarily that this was useful to add here because it might affect some of those frameworks. i.e for visibility. Hope that makes sense!

FYI, the corresponding new note [in the docs here](https://developer.mozilla.org/en-US/docs/Web/API/Window/name#Notes) is:

image

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -4396,11 +4396,11 @@
             },
             "firefox": {
               "version_added": "1",
-              "notes": "From Firefox 82, if a new page from another domain is loaded into a tab, then <code>window.name</code> is set to the empty string. This change prevents some cross-site attacks. See <a href='https://bugzil.la/444222'>bug 444222</a>."
+              "notes": "From Firefox 82, if a new page from another domain is loaded into a tab, then <code>window.name</code> is set to the empty string (the original string is restored if the tab reverts back to the original page). This change prevents some cross-site attacks. See <a href='https://bugzil.la/444222'>bug 444222</a>."
             },
             "firefox_android": {
               "version_added": "4",
-              "notes": "From Firefox 82, <code>window.name</code> is reset if a new page from another domain is loaded into a tab. See <a href='https://bugzil.la/444222'>bug 444222</a>."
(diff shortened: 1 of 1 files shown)


chrisdavidmills (approved):
LGTM, thanks @hamishwillee!

@ddbeck , does the data structure here look OK to you?

api/Window.json
```diff
@@ -4395,10 +4395,12 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "1"
+              "version_added": "1",
+              "notes": "From Firefox 82, <code>window.name</code> is reset if a new page from another domain is loaded into a tab. See <a href='https://bugzil.la/444222'>bug 444222</a>."
```
ddbeck, line 4399: I'm a little concerned that this note reads as if this is unintended behavior, plus I'd like to be more explicit about what "reset" means. I'm open to further edits, but what to do you think of this?

```suggestion
              "notes": "From Firefox 82, if a new page from another domain is loaded into a tab, then <code>window.name</code> is set to the empty string. This change prevents some cross-site attacks. See <a href='https://bugzil.la/444222'>bug 444222</a>."
```

hamishwillee:
> I'm a little concerned that this note reads as if this is unintended behavior, plus I'd like to be more explicit about what "reset" means. I'm open to further edits, but what to do you think of this?

@ddbeck Certainly more explicit. I added to this with a comment that the string is reverted back to the original if the original content is reverted. By which I mean that the users selects the back button to reverse history, not that they load the original page as a new page.

I also pushed this to both instances.

Better now?

ddbeck (approved):
Looks great. Thank you! 🎉

PR #6887, opened 2020-10-12
Update strict mime type checks on workers/sharedworkers to FF81

hamishwillee:
Changes the version of FF in which strict mime-type checks for workers and shared workers from 70 to 81. 

The support was added/re-enabled in https://bugzilla.mozilla.org/show_bug.cgi?id=1569123

The history is a little messy/complicated, but it looks like FF70 added support and then immediately made is only present in beta/nightly builds. There is a bit more information on that captured [here](https://www.fxsitecompat.dev/en-CA/docs/2020/worker-scripts-with-wrong-mime-type-will-be-blocked-from-loading-with-worker-or-sharedworker/).

It isn't clear whether this was also enabled in FF for Android. I am querying and will either update this PR if it doesn't get merged, or create a new PR if it does. IMO not worth holding this up if you have bandwidth to review now.


ddbeck (approved):
Looks good. Thank you! 🎉

PR #6889, opened 2020-10-12
Deprecate MerchantValidationEvent

sideshowbarker:
This change marks `MerchantValidationEvent` and all its subfeatures `standard_track:false` and `deprecated:true,` due to the fact it was all removed from the Payment Request spec:

* https://github.com/w3c/payment-request/pull/929
* https://github.com/w3c/payment-request/commit/8337fee

https://developer.mozilla.org/en-US/docs/Web/API/MerchantValidationEvent and its sub-articles already updated.

cc @marcoscaceres

### DIFF
--- a/api/MerchantValidationEvent.json
+++ b/api/MerchantValidationEvent.json
@@ -59,8 +59,8 @@
         },
         "status": {
           "experimental": false,
-          "standard_track": true,
-          "deprecated": false
+          "standard_track": false,
+          "deprecated": true
         }
       },
       "MerchantValidationEvent": {
@@ -109,8 +109,8 @@
           },
           "status": {
             "experimental": false,
-            "standard_track": true,
-            "deprecated": false
+            "standard_track": false,
+            "deprecated": true
           }
         }
       },
@@ -160,8 +160,8 @@
           },
           "status": {
             "experimental": false,
-            "standard_track": true,
-            "deprecated": false
+            "standard_track": false,
+            "deprecated": true
           }
         }
       },
@@ -210,8 +210,8 @@
           },
           "status": {
             "experimental": false,
-            "standard_track": true,
-            "deprecated": false
+            "standard_track": false,
+            "deprecated": true
           }
         }
       },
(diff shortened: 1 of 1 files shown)


marcoscaceres (approved):

foolip:
This is implemented behind a flag in both Chrome and Firefox. Calling it deprecated when it was never even shipped seems a bit weird to me. The definition in https://github.com/mdn/browser-compat-data/blob/master/schemas/compat-data-schema.md#status-information talks about "might be removed in the future or might only be kept for compatibility purposes" and neither really make sense.

I'd argue it's more experimental than deprecated, although the fact that it has two implementations would make it non-experimental per the suggested guideline https://github.com/mdn/browser-compat-data/issues/6905.

sideshowbarker:
> I'd argue it's more experimental than deprecated, although the fact that it has two implementations would make it non-experimental per the suggested guideline #6905.

I’d be fine with marking it experimental instead. And given this case, maybe the refinements to our description of experimental being developed #6905 need be broadened to include the sense *“an experiment that failed”* (or maybe *“failed”* is too strong a word but instead more like *“a historical experiment”*, or *“an experiment the never ended up reaching maturity”*.

Anyway, I don’t personally have a strong opinion about experimental vs deprecated for this case. But I do think we need to mark it with one or other — since not doing so leaves the implication that it’s stable, or possibly has a future.

marcoscaceres:
> I'd argue it's more experimental than deprecated, although the fact that it has two implementations would make it non-experimental per the suggested guideline #6905.

It's a hard one - but on reflection, I would agree it's not "deprecated". The reality is that it's a Safari-only API for now (and for the foreseeable future).

sideshowbarker:
> on reflection, I would agree it's not "deprecated". The reality is that it's a Safari-only API for now (and for the foreseeable future)

I guess that puts this into the category of what we’ve ended up (re)discussing on https://github.com/mdn/browser-compat-data/issues/6905

To me, the gist of it is, we need to add a new flag to BCD for cases like this, to flag *“not an agreed-upon part of the web platform”* features. A new flag named simply “single-vendor” would be accurate for doing that in the vast majority of cases.

ddbeck:
Just to making something explicit here: I'm treating making a decision in #6905 as blocking this PR.

PR #6890, opened 2020-10-12
(Re)Mark Safari visibilitychange support partial

sideshowbarker:
This change adds `partial_implementation:true` to the data for the `visibilitychange` event and `onvisibilitychange` event handler. In https://github.com/mdn/browser-compat-data/pull/6849 those flags got dropped (maybe inadvertently).

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -7624,6 +7624,7 @@
             },
             "safari": {
               "version_added": "10.1",
+              "partial_implementation": true,
               "notes": [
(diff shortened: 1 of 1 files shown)


vinyldarkscratch (approved):
Based upon the comment in the PR that removed this, you strike a valid point.  Sure thing, let's re-add them!

sideshowbarker:
Thanks much — I also noticed a separate issue with the data for `visibilityState` in `api/Document.json`, and not sure what the best way is to fix it this point, but I guess I’ll raise a separate PR for that

sideshowbarker:
>  I also noticed a separate issue with the data for `visibilityState` in `api/Document.json`, and not sure what the best way is to fix it this point, but I guess I’ll raise a separate PR for that

Raised https://github.com/mdn/browser-compat-data/pull/6898

PR #6891, opened 2020-10-12
Add HTMLElement.enterKeyHint

makotokato:
Actually  `HTMLElement.enterKeyHint` is missing. We should add this and update supported version.

- [x] Summarize your changes

### DIFF
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -868,6 +868,67 @@
           }
         }
       },
+      "enterKeyHint": {
+        "__compat": {
+          "support": {
+            "chrome": {
+              "version_added": "77"
+            },
+            "chrome_android": {
+              "version_added": "77"
+            },
+            "edge": {
+              "version_added": false
+            },
+            "firefox": {
+              "version_added": "79",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "dom.forms.enterkeyhint",
+                  "value_to_set": "true"
+                }
+              ]
+            },
+            "firefox_android": {
+              "version_added": "79",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "dom.forms.enterkeyhint",
+                  "value_to_set": "true"
+                }
+              ]
+            },
+            "ie": {
+              "version_added": false
+            },
+            "opera": {
+              "version_added": false
+            },
+            "opera_android": {
+              "version_added": false
+            },
+            "safari": {
+              "version_added": "13.1"
+            },
+            "safari_ios": {
+              "version_added": "13.4"
+            },
+            "samsunginternet_android": {
+              "version_added": false
+            },
+            "webview_android": {
+              "version_added": "77"
+            }
+          },
+          "status": {
+            "experimental": false,
+            "standard_track": true,
+            "deprecated": false
+          }
+        }
+      },
       "focus": {
         "__compat": {
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLOrForeignElement/focus",
--- a/html/global_attributes.json
+++ b/html/global_attributes.json
@@ -883,10 +883,24 @@
               "version_added": false
             },
             "firefox": {
-              "version_added": false
+              "version_added": "79",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "dom.forms.enterkeyhint",
+                  "value_to_set": "true"
+                }
+              ]
             },
             "firefox_android": {
-              "version_added": false
+              "version_added": "79",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "dom.forms.enterkeyhint",
+                  "value_to_set": "true"
+                }
+              ]
             },
             "ie": {
               "version_added": false


api/HTMLElement.json
```diff
@@ -868,6 +868,67 @@
...
+              "version_added": false
+            },
+            "opera": {
+              "version_added": false
+            },
+            "opera_android": {
+              "version_added": false
+            },
+            "safari": {
+              "version_added": "13.1"
```
sideshowbarker, line 913: I notice that https://webkit.org/blog/10247/new-webkit-features-in-safari-13-1/ says (emphasis added):

> **On iOS**, WebKit supports the `enterkeyhint` attribute

That seems to imply it’s *not* supported in desktop Safari?

api/HTMLElement.json
```diff
@@ -868,6 +868,67 @@
           }
         }
       },
+      "enterKeyHint": {
+        "__compat": {
+          "support": {
+            "chrome": {
+              "version_added": "77"
```
sideshowbarker, line 875: Version 77 confirmed by https://www.chromestatus.com/feature/5654529808793600

api/HTMLElement.json
makotokato: DOM binding supports on both macOS and iOS (https://bugs.webkit.org/show_bug.cgi?id=203440).  I tested on wpt.live on Safari/macOS.
UX support for software keyboard is https://bugs.webkit.org/show_bug.cgi?id=189546. This is iOS only.

sideshowbarker: > DOM binding supports on both macOS and iOS (https://bugs.webkit.org/show_bug.cgi?id=203440). I tested on wpt.live on Safari/macOS.

OK, so then I think it’s clear that marking it as macOS-supported (as this PR does) is accurate.

> UX support for software keyboard is https://bugs.webkit.org/show_bug.cgi?id=189546. This is iOS only.

So should we add a note field for the `safari` data that says something like:

> "note": "Doesn’t expose UX support (software keyboard is exposed only on iOS)."

…or something like that?

makotokato: > So should we add a note field for the safari data that says something like:
>> "note": "Doesn’t expose UX support (software keyboard is exposed only on iOS)."
> …or something like that?

This issue is same as Chrome desktop since Android and iOS has the OS API to change `[enterkey]`, but no API for others such as Windows.  Should we add same note to chrome desktop too?

api/HTMLElement.json
```diff
@@ -868,6 +868,67 @@
...
+            },
+            "chrome_android": {
+              "version_added": "77"
+            },
+            "edge": {
+              "version_added": false
```
foolip, line 881:
```suggestion
              "version_added": "79"
```

Based on testing https://mdn-bcd-collector.appspot.com/tests/api/HTMLElement/enterKeyHint in Edge 18 and 79.

api/HTMLElement.json
foolip: Also confirmed by:
- testing https://mdn-bcd-collector.appspot.com/tests/api/HTMLElement/enterKeyHint in Chrome 76 and 77
- https://storage.googleapis.com/chromium-find-releases-static/ae7.html#ae770f5e4a155964130ac56abff9438e488aca50

api/HTMLElement.json
```diff
@@ -868,6 +868,67 @@
...
+                  "name": "dom.forms.enterkeyhint",
+                  "value_to_set": "true"
+                }
+              ]
+            },
+            "ie": {
+              "version_added": false
+            },
+            "opera": {
+              "version_added": false
```
foolip, line 907:
```suggestion
              "version_added": "64"
```

Based on running `npm run mirror -- --modify=always opera api.HTMLElement.enterKeyHint`

api/HTMLElement.json
```diff
@@ -868,6 +868,67 @@
...
+            },
+            "opera_android": {
+              "version_added": false
```
foolip, line 910:
```suggestion
              "version_added": "55"
```

(Also from `npm run mirror`.)

api/HTMLElement.json
```diff
@@ -868,6 +868,67 @@
...
+            },
+            "safari": {
+              "version_added": "13.1"
+            },
+            "safari_ios": {
+              "version_added": "13.4"
+            },
+            "samsunginternet_android": {
+              "version_added": false
```
foolip, line 919:
```suggestion
              "version_added": "12.0"
```

(Also from `npm run mirror`.)

api/HTMLElement.json
foolip: I would expect this to be true of all desktop platforms, not just Safari. This is a similar situation to some media query never matching on some platforms. Do we have notes for other such situations? If we can't foresee that this attribute would ever do anything on desktop platforms, then such notes don't seem useful.

sideshowbarker: > This issue is same as Chrome desktop since Android and iOS has the OS API to change `[enterkey]`, but no API for others such as Windows. Should we add same note to chrome desktop too?

I’m not sure what the BCD policy is for these kinds of cases. I am sure we have some precedents, but I’m not sure what we did on those cases. I guess that in general we have a number of features that we all known aren’t really meant to be used on desktop — because no UX is exposed for them on desktop, even though the underlying APIs are exposed. So in those cases, we shouldn’t mark the engine support as missing.

Anyway, I’ve opened https://github.com/mdn/browser-compat-data/issues/6914 to track that issue for this case. But I don’t think we need to block this PR on waiting for resolution of that — instead we can just do a follow-up PR later.

So I’ll go ahead and merge this PR.

sideshowbarker (approved):

api/HTMLElement.json
```diff
@@ -868,6 +868,67 @@
...
+            },
+            "opera_android": {
+              "version_added": false
+            },
+            "safari": {
+              "version_added": "13.1"
+            },
+            "safari_ios": {
+              "version_added": "13.4"
```
foolip, line 916: I can't find an iOS 13.4 device on BrowserStack, but could at least verify that https://mdn-bcd-collector.appspot.com/tests/api/HTMLElement/enterKeyHint returns false on 13.3 and true on 14. So this seems right.

html/global_attributes.json
```diff
@@ -883,10 +883,24 @@
               "version_added": false
```
foolip, line 883:
```suggestion
              "version_added": "79"
```

Can you also update the rest of this data to match the IDL attribute after applying those suggestions?

sideshowbarker:
@foolip Sorry! I guess I merged this just before you submitted these review changes.

@makotokato You can either raise a follow-up PR to include these changes, or I can (since it was my fault that this got merged prematurely).

foolip:
@sideshowbarker no worries, race conditions are unavoidable :)

makotokato:
> You can either raise a follow-up PR to include these changes, or I can (since it was my fault that this got merged prematurely).

I will send new PR to update notes and etc

PR #6893, opened 2020-10-12
Mark Accelerometer experimental

sideshowbarker:
This change marks the `Accelerometer` interface and all its subfeatures with `experimental:true`.

---

I’ve also marked https://wiki.developer.mozilla.org/en-US/docs/Web/API/Accelerometer and all its sub-articles with `{{seecompattable}}`.

See related discussion at https://github.com/mdn/browser-compat-data/pull/6873#issuecomment-707180136

### DIFF
--- a/api/Accelerometer.json
+++ b/api/Accelerometer.json
@@ -42,7 +42,7 @@
           }
         },
         "status": {
-          "experimental": false,
+          "experimental": true,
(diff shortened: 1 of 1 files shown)


ddbeck:
Like a couple other PRs: I'm treating making a decision in #6905 as blocking this PR.

PR #6895, opened 2020-10-12
Mark LargestContentfulPaint experimental

sideshowbarker:
This change marks the `LargestContentfulPaint` interface and all is subfeatures with `experimental:true`.

---

Also marked https://wiki.developer.mozilla.org/en-US/docs/Web/API/LargestContentfulPaint with `{{seecompattable}}`

See related discussion at https://github.com/mdn/browser-compat-data/pull/6873#issuecomment-707180136

### DIFF
--- a/api/LargestContentfulPaint.json
+++ b/api/LargestContentfulPaint.json
@@ -42,7 +42,7 @@
           }
         },
         "status": {
-          "experimental": false,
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Looks good. Thank you! 🎉

PR #6898, opened 2020-10-13
Update Safari data for visibilityState

sideshowbarker:
This change updates the Safari data for `visibilityState` in light of related changes made to the `visibilitychange` data in https://github.com/mdn/browser-compat-data/pull/6849.

Specifically, testing indicates the following note isn’t accurate:

> Doesn't fire the `visibilitychange` event when `document.visibilityState` transitions to `hidden`

But what’s instead accurate is the following:

> Doesn't fire the `visibilitychange` event when navigating way from a document

However, that information applies only to `visibilitychange` and not to `visibilityState` — so it’s also not accurate to keep the `visibilityState` support marked as partial.

Therefore, this change drops the `partial_implementation` flag, as well as dropping the note entirely.

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -10891,14 +10891,10 @@
               "notes": "Doesn't fire the <code>visibilitychange</code> event when the browser window is minimized, nor when <code>hidden</code> is set to true."
             },
             "safari": {
-              "version_added": "7",
-              "partial_implementation": true,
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
That's a really clear explanation, @sideshowbarker. Thank you! 🎉

PR #6900, opened 2020-10-13
Set Firefox data for CSSAnimation API

vinyldarkscratch:
This PR sets Firefox data for the CSSAnimation API based upon data from the mdn-bcd-collector project.  Test used: https://mdn-bcd-collector.appspot.com/tests/api/CSSAnimation

### DIFF
--- a/api/CSSAnimation.json
+++ b/api/CSSAnimation.json
@@ -14,10 +14,10 @@
             "version_added": "84"
           },
           "firefox": {
-            "version_added": false
+            "version_added": "75"
           },
           "firefox_android": {
-            "version_added": false
+            "version_added": "79"
           },
           "ie": {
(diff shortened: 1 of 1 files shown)


api/CSSAnimation.json
```diff
@@ -14,10 +14,10 @@
             "version_added": "84"
           },
           "firefox": {
-            "version_added": false
+            "version_added": "75"
```
foolip, line 17: Checks out, I think this was included in https://bugzilla.mozilla.org/show_bug.cgi?id=1619821#c12, part of https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/75#Web_animations_API.

foolip (approved):

PR #6901, opened 2020-10-13
Set Safari data for CSSAnimation API

vinyldarkscratch:
This PR sets Safari data for the CSSAnimation API based upon data from the mdn-bcd-collector project.  Test used: https://mdn-bcd-collector.appspot.com/tests/api/CSSAnimation

### DIFF
--- a/api/CSSAnimation.json
+++ b/api/CSSAnimation.json
@@ -29,10 +29,10 @@
             "version_added": null
           },
           "safari": {
-            "version_added": false
+            "version_added": "13.1"
           },
           "safari_ios": {
-            "version_added": false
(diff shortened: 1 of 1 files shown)


api/CSSAnimation.json
```diff
@@ -29,10 +29,10 @@
             "version_added": null
           },
           "safari": {
-            "version_added": false
+            "version_added": "13.1"
```
foolip, line 32: Mentioned in https://developer.apple.com/documentation/safari-release-notes/safari-13_1-release_notes so this seems right.

foolip (approved):

Issue #6786, opened 2020-10-01
Consider moving or copying internal aspect-ratio data


ddbeck:
We have data for the CSS property [`aspect-ratio`](https://github.com/mdn/browser-compat-data/blob/master/css/properties/aspect-ratio.json) ([MDN docs](https://developer.mozilla.org/docs/Web/CSS/aspect-ratio)), but it's rather unusual data. We should consider moving or duplicating the data elsewhere.

All of the implementations to date are browser internals, for the purpose of computing the aspect ratio of replaced elements. No browser appears to have actually shipped the `aspect-ratio` property to web developers (though Firefox and Chrome appear to be developing the feature). We currently indicate this through the `partial_implementation` flag; this seems misleading. As far as web developers are concerned, there is no implementation.

The genuinely useful bit for web developers is the subfeature `internal-mapping`, which captures how browsers are using the `width` and `height` attributes of replaced elements to compute the aspect ratio for rendering a page. To me, this feels like the essential thing—but it's more of a feature of _elements_, rather than the `aspect-ratio` property. For example, [the Firefox UA stylesheet _uses_ the property to do this](https://developer.mozilla.org/en-US/docs/Web/CSS/aspect-ratio#Examples); it doesn't appear to be a feature _of_ the property. It seems to me that the story here is that browsers have changed the behavior of certain HTML elements' attributes.

(This structure of the existing data was introduced in #5114, which is useful additional reading.)

I propose something that's a bit fussy, but might more accurately reflect the role of the `aspect-ratio` property today:

1. Copy the internal mapping data to new subfeatures of the applicable individual HTML elements. For example, there would be a new subfeature on `html.elements.img` which communicates the aspect ratio set by the element's attributes.
2. Drop the internal mapping data from `aspect-ratio`.
3. Set the `aspect-ratio` data to `"version_added": false` and link to the relevant implementation bugs in the notes.

I'd welcome feedback on this proposal.

ddbeck:
@rachelandrew You were the originator of the `aspect-ratio` property data, so you might be especially interested in this suggestion.

rachelandrew:
That seems a reasonable approach, it was difficult to figure out what to do with it in the first place. 

BTW It's in Canary already as a property. (Here's a pen: https://codepen.io/rachelandrew/pen/WNrRZaV)

jensimmons:
[...]

PR #6903, opened 2020-10-13
update aspect ratio

rachelandrew:
Removing details of internal mapping and adding support only for property.

As discussed in #6786 this PR removes details of the internal mapping to HTML width and height, and instead adds the support for the property. Partial support behind a flag in Firefox, and support behind a flag in Chrome.


ddbeck:
The `internal-value` feature ought to be replaced in the same PR as one that fixes #6786 (to ensure it's all fixed at once and reflected in the release notes). I'll open a complementary PR for this soon.

ddbeck:
@rachelandrew I've opened #6918 to relocate the HTML data. If you restore the `internal-value` data in this PR, then I can track the move for the release notes a little more gracefully. Alternatively, I've requested your review on that PR—if we merge that one first, you can rebase this PR on top of it.

rachelandrew:
@ddbeck sorry I don't understand what the end result is that you want. If you just want it put back then we can close this PR and start over?

css/properties/aspect-ratio.json
```diff
@@ -53,92 +46,23 @@
...
+              "version_added": false
             }
           },
           "status": {
-            "experimental": true,
+            "experimental": false,
             "standard_track": true,
             "deprecated": false
           }
-        },
```
ddbeck, line 81: Sorry, Rachel. I should've been more clear about what I was asking for. I'd like you to undo the changes from this line down. In other words, leave the `internal-value` feature in the file (which will be removed in my PR), but keep the changes from above this line.

ddbeck (changes requested):

rachelandrew:
@ddbeck hopefully what I've done is right, also added the note about the implementation of support for flex items in Firefox 83.

ddbeck (changes requested):
One consistency suggestion here, but otherwise this looks really good. Thank you, @rachelandrew!

ddbeck (approved):
Looks good. Thank you!

To get the tests to pass (and to fix my earlier mistake of asking you to remove the subfeature removal), I'm going to push a merge from master into this branch, so it can appear in today's release. 🚀

ddbeck:
Travis is stuck again. Merging without it.

PR #6904, opened 2020-10-13
add conic-gradient support data for Firefox

myfonj:
According
- https://github.com/Fyrd/caniuse/blob/master/features-json/css-conic-gradients.json
- https://bugzilla.mozilla.org/show_bug.cgi?id=1175958

### DIFF
--- a/css/types/image.json
+++ b/css/types/image.json
@@ -370,15 +370,20 @@
                     ]
                   }
                 ],
-                "firefox_android": {
-                  "version_added": "79",
-                  "flags": [
-                    {
-                      "type": "preference",
-                      "name": "layout.css.conic-gradient.enabled"
-                    }
-                  ]
-                },
+                "firefox_android": [
+                  {
+                    "version_added": "83"
+                  },
+                  {
+                    "version_added": "79",
+                    "flags": [
+                      {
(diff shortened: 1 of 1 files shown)


css/types/image.json
```diff
@@ -356,11 +356,28 @@
...
+                      {
+                        "type": "preference",
+                        "name": "layout.css.conic-gradient.enabled"
+                      }
+                    ]
+                  }
+                ],
                 "firefox_android": {
-                  "version_added": false
+                  "version_added": "79",
```
ddbeck, line 374: Reading https://bugzilla.mozilla.org/show_bug.cgi?id=1632351, it looks like this will also be in Firefox for Android 83. Can you copy the entry from desktop here too?

ddbeck (changes requested):
Thanks for this PR, @myfonj. I have one suggestion in a line comment, otherwise this looks good!

css/types/image.json
myfonj: Good catch, added. (Will mention this PR in that bug, just in case.)

ddbeck (approved):
Looks good. Thank you, @myfonj! 🎉

PR #6907, opened 2020-10-13
Edge 16+ supports ...rest in array destructuring

developit:
Currently the compat data makes it seem like anything that was behind the "Experimental JavaScript Features" flag in pre-Chromium Edge is still behind a flag, which I don't believe is true for any ES features. There may be a few of these cutoffs missing (`AsyncIterator` is another one).

Here's a test in Edge 16:
Screen Shot 2020-10-13 at 2 44 22 PM

Here's a test in Edge 79:
Screen Shot 2020-10-13 at 2 33 35 PM

### DIFF
--- a/javascript/operators/destructuring.json
+++ b/javascript/operators/destructuring.json
@@ -116,15 +116,20 @@
               "chrome_android": {
                 "version_added": "49"
               },
-              "edge": {
-                "version_added": "14",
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Looks good. Thank you! 🎉

PR #6915, opened 2020-10-14
Add Safari 4 as ranged version

### LINKED ISSUE #6807: Add range for Safari ≤3?
Considering how challenging it is to test Safari 1-2, I'm thinking we should add a range for Safari 3 and lower?

vinyldarkscratch:
This PR fixes #6807.  Due to the challenges of testing Safari 1 through 2, and the lack of Safari 3 in BrowserStack, this PR adds a range for Safari 3 and lower, as well as Safari 4 and lower.

Explanation of challenges:
Safari 1 and 2 were introduced for Mac OS X 10.2 and 10.3, both of which are built upon PowerPC architecture.  As a result, modern emulators do not support running such old editions of Mac OS X.  I was able to find a program that emulates PowerPC architecture, though it was been in beta for over a decade.  There is no sensible way to send files to the PowerPC VMs (you need a Mac or Mac image reader software like Paragon HFS+), and although I was able to set up an internet connection, it's spotty at best, and requires disabling and re-enabling the network connection for the host VM every time the host VM boots up, then resetting the PowerPC VM's connection a few times.

Bottom line is...it's inefficient, borderline impossible, to test Safari 1-2.  While I do plan on adding exact data for Safari 1-2, it would be much easier to run through all of those features at once at a later date, and simply set a ranged value for the time being.  (This will also give me some extra time to try configuring another emulator while we achieve our goals.)

Edit: Since Safari 3 is not available on BrowserStack or SauceLabs, this PR includes a range for Safari 4 instead.


foolip:
I think we should do this! @vinyldarkscratch for Safari 3, is that available for macOS running on x86, or are only the Windows versions convenient to test for some range?

vinyldarkscratch:
Unfortunately Safari 3.x for macOS is also for versions that run on PowerPC architecture, so at the moment, it's only sensible to test in Windows.

foolip:
What is the earliest version of Safari that's practical to test on macOS, available on either BrowserStack or Sauce? That would be a useful cutoff point for more contributors, and if it's also very far back in time, the difference doesn't matter much.

sideshowbarker:
+1 👍 I also think we should do this.

foolip:
To answer my own question, Safari 4.0.5 on Mac OS X 10.6.8 is the oldest available on BrowserStack. Sauce's oldest is Safari 8.

So I'd suggest we add ≤4, as that's something more people could verify.

vinyldarkscratch:
I added a range for Safari 4 as well, good point!  Since I'm able to test as far back as Safari 3 efficiently, I thought it would good to add them both, since we can get closer to real values that way?  (Happy to keep it as just ≤4 though!)

vinyldarkscratch:
Verbal discussion with @foolip led to the conclusion to simply keep Safari 4 as the only ranged version.

foolip (approved):
This looks good to me, but I shouldn't merge myself, deferring to @ddbeck.

foolip:
Friendly ping, @ddbeck, this is blocking @vinyldarkscratch and me making progress on Safari PRs.

ddbeck:
@foolip Thanks for the ping (and your patience on this).

I'm OK with this, but https://github.com/mdn/browser-compat-data/blob/master/schemas/compat-data-schema.md#ranged-versions needs to be updated as well.

foolip:
@vinyldarkscratch over to you :)

schemas/compat-data-schema.md
```diff
@@ -251,7 +251,7 @@ Examples:
 
 ### Ranged versions
 
-For certain browsers, ranged versions are allowed as it is sometimes impossible to find out in which early version of a browser a feature shipped. Ranged versions should be used sparingly and only when it is impossible to find out the version number a feature initially shipped in. The following ranged version values are allowed:
+For certain browsers, ranged versions are allowed as it is sometimes impossible, or simply unreasonable, to find out in which early version of a browser a feature shipped. Ranged versions should be used sparingly and only when it is impossible to find out the version number a feature initially shipped in. The following ranged version values are allowed:
```
ddbeck, line 254:
```suggestion
For certain browsers, ranged versions are allowed as it is sometimes impractical to find out in which early version of a browser a feature shipped. Ranged versions should be used sparingly and only when it is impossible to find out the version number a feature initially shipped in. The following ranged version values are allowed:
```

ddbeck (approved):
This looks good. I do have one suggestion for the docs but, in the interest of unblocking this, I'm going to directly apply and merge it. Thank you!

foolip:
Wohoo, thanks @ddbeck! @vinyldarkscratch looking forward to turning Safari nulls and trues into something else! :)

PR #6916, opened 2020-10-14
UC browser supports Object.create

developit:
It is (was?) based on Android Browser 4.x, which supported it. This likely means UC supports most of the other Object.* methods prior to ES2015.

### DIFF
--- a/javascript/builtins/Object.json
+++ b/javascript/builtins/Object.json
@@ -252,6 +252,9 @@
               },
               "webview_android": {
                 "version_added": "1"
+              },
+              "uc_android": {
+                "version_added": true
               }
             },
             "status": {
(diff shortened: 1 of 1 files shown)


jpmedley (approved):

jpmedley (changes requested):
Browsers are listed alphabetically. Can you please move this?

developit:
Updated - thanks @jpmedley

ddbeck:
Thanks for this update, @developit, but unfortunately I'm going to close this PR. This is not a judgment on your PR, however. Due to pervasive issues with our UC and QQ data, we're getting ready to drop them from BCD's data set (at least until we have a more sustainable maintenance story). Take a look at #7240 and #7238 for details. Thank you for your understanding on this.

PR #6918, opened 2020-10-14
Move width and height mapping to aspect ratio to HTML data

### LINKED ISSUE #6786: Consider moving or copying internal aspect-ratio data

ddbeck:
The data about some HTML elements' `width` and `height` attributes mapping to that element's aspect ratio is in in the wrong place. This PR fixes this problem (and fixes #6786) by:

- Changing the description of the misplaced `aspect-ratio` subfeature to describe the attributes' relationship
- Renames the subfeature to `aspect_ratio_computed_from_attributes`
- Mirrors data to complete the Opera and Firefox data
- Moves the data to the HTML elements

Related to #6903.

### DIFF
--- a/css/properties/aspect-ratio.json
+++ b/css/properties/aspect-ratio.json
@@ -78,67 +78,6 @@
             "standard_track": true,
             "deprecated": false
           }
-        },
-        "aspect_ratio_computed_from_attributes": {
-          "__compat": {
-            "description": "Aspect ratio computed from <code>width</code> and <code>height</code> attributes",
-            "support": {
-              "chrome": {
-                "version_added": "79"
-              },
-              "chrome_android": {
-                "version_added": "79"
-              },
-              "edge": {
-                "version_added": "79"
-              },
-              "firefox": [
-                {
-                  "version_added": "71"
-                },
-                {
-                  "version_added": "69",
-                  "version_removed": "71",
-                  "flags": [
-                    {
-                      "type": "preference",
-                      "name": "layout.css.width-and-height-map-to-aspect-ratio.enabled",
-                      "value_to_set": "true"
-                    }
-                  ]
-                }
-              ],
-              "firefox_android": {
-                "version_added": "79"
-              },
-              "ie": {
-                "version_added": false
-              },
-              "opera": {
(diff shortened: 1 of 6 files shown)


ddbeck:
@rachelandrew could I get your review on this? If it's OK, then merging this will create the fix needed to make the tests pass for for your PR #6903.

rachelandrew (approved):
Looks good to me :)

PR #6919, opened 2020-10-15
Update page-margin-boxes mdn_url

sideshowbarker:
This change updates the `mdn_url` value for the page-margin-boxes feature to include the anchor for the specific Syntax section part for `page-margin-box`.


ddbeck (approved):
Thank you! 🎉

Issue #6768, opened 2020-09-27
Remove totally unimplemented things


saschanaz:
Below are interfaces that have never been implemented per the data, or just no data.

Marked as deprecated/obsolete:

- [x] [BluetoothAdvertisingData](https://developer.mozilla.org/docs/Web/API/BluetoothAdvertisingData)
- [x] [DOMConfiguration](https://developer.mozilla.org/docs/Web/API/DOMConfiguration)
- [x] [DOMImplementationList](https://developer.mozilla.org/docs/Web/API/DOMImplementationList)
- [x] [HTMLIsIndexElement](https://developer.mozilla.org/docs/Web/API/HTMLIsIndexElement)
- [x] [Notation](https://developer.mozilla.org/docs/Web/API/Notation)
- [x] [PushRegistrationManager](https://developer.mozilla.org/docs/Web/API/PushRegistrationManager)
- [x] [SVGColorProfileElement](https://developer.mozilla.org/docs/Web/API/SVGColorProfileElement)
- [x] [SVGSolidcolorElement](https://developer.mozilla.org/docs/Web/API/SVGSolidcolorElement)
- [x] [SVGAltGlyphDefElement](https://developer.mozilla.org/docs/Web/API/SVGAltGlyphDefElement)
- [x] [SVGAltGlyphItemElement](https://developer.mozilla.org/docs/Web/API/SVGAltGlyphItemElement)
- [x] [SVGFontElement](https://developer.mozilla.org/docs/Web/API/SVGFontElement)
- [x] [SVGFontFaceElement](https://developer.mozilla.org/docs/Web/API/SVGFontFaceElement)
- [x] [SVGFontFaceFormatElement](https://developer.mozilla.org/docs/Web/API/SVGFontFaceFormatElement)
- [x] [SVGFontFaceNameElement](https://developer.mozilla.org/docs/Web/API/SVGFontFaceNameElement)
- [x] [SVGFontFaceSrcElement](https://developer.mozilla.org/docs/Web/API/SVGFontFaceSrcElement)
- [x] [SVGFontFaceUriElement](https://developer.mozilla.org/docs/Web/API/SVGFontFaceUriElement)
- [x] [SVGGlyphElement](https://developer.mozilla.org/docs/Web/API/SVGGlyphElement)
- [x] [SVGGlyphRefElement](https://developer.mozilla.org/docs/Web/API/SVGGlyphRefElement)
[...]

PR #6929, opened 2020-10-15
Remove never-supported nodeFromPoint

saschanaz:
- [x] Summarize your changes


api/DocumentOrShadowRoot.json
```diff
@@ -467,106 +467,6 @@
...
-            },
-            "chrome_android": {
-              "version_added": false
-            },
-            "edge": {
-              "version_added": false
-            },
-            "firefox": {
-              "version_added": false,
-              "notes": "Implemented in Firefox but currently only works in chrome code."
```
foolip, line 485: https://github.com/mozilla/gecko-dev/blob/f98dad153b59a985efd4505912588d4651033395/dom/webidl/DocumentOrShadowRoot.webidl#L18-L23 confirms this. There's a reference to https://github.com/w3c/csswg-drafts/issues/556 which is still open, there isn't anything to implement yet.

foolip (approved):

foolip:
https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot/nodeFromPoint and https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot/nodesFromPoint would have to be archived. @sideshowbarker do you have the power to do this? (Not merging this yet.)

sideshowbarker:
> https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot/nodeFromPoint and https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot/nodesFromPoint would have to be archived. @sideshowbarker do you have the power to do this?

No, I don’t — and not sure who else among BCD contributors might, except for @chrisdavidmills

chrisdavidmills:
Yup, I can do this. Do you think archiving is the best idea? I could just delete them, or perhaps redirect them to the main interface page, to avoid the chance of broken links.

foolip:
I don't know what's usually done with these pages, but I think @Elchi3 or @ddbeck helped me archive one in a previous instance.

We will need to archive or remove a lot of pages to land the [many PRs @saschanaz has sent](https://github.com/mdn/browser-compat-data/pulls/saschanaz) so it'd be great to pick an approach and coordinate a bit. Maybe like this:
- Someone reviews to confirm historical non-implementation of the removed entries
- That someone approves and ping a second reviewer who can archive/delete MDN pages
- That second reviewer does the MDN thing and then merges the PR

ddbeck:
> it'd be great to pick an approach and coordinate a bit. Maybe like this:
> 
> * Someone reviews to confirm historical non-implementation of the removed entries
> * That someone approves and ping a second reviewer who can archive/delete MDN pages
> * That second reviewer does the MDN thing and then merges the PR

I agree with this and I like this as proposed. To fill in one detail: I like the idea of redirecting to the parent interface (if it exists). If there's no parent, then we'll need to take it on a case-by-case basis, perhaps with a bias toward deletion unless someone can come up with an affirmative case for preserving the page.

For pages that need redirects, I'm happy to be the default second reviewer for the MDN redirect + merge. For deletes/archives, I think that needs extra permissions. I guess we pull in @chrisdavidmills for those?

Any suggestions or objections before we just go for it?

foolip:
Sounds great to me, if @chrisdavidmills is up for some page deletion duty :)

chrisdavidmills:
> Sounds great to me, if @chrisdavidmills is up for some page deletion duty :)

Yup, definitely! 

I'd like to go with @ddbeck 's suggestion of redirecting to parent page if there is a sensible parent to redirect to, or deleting if not.

Aside from that, send me the list of pages and I can get to work.

foolip:
I've summarized our process in https://github.com/mdn/browser-compat-data/pull/6854#issuecomment-718672722 and will start with this very PR.

foolip (approved):
These are Gecko APIs never exposed to web content. These pages need redirecting to the parent page:
https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot/nodeFromPoint
https://developer.mozilla.org/en-US/docs/Web/API/DocumentOrShadowRoot/nodesFromPoint

Over to @ddbeck

ddbeck (approved):
Redirects done and merging. Thanks everyone for coming together for a good way to handle these. 👍

PR #6930, opened 2020-10-15
Remove never-supported DOMConfiguration/DOMImplementationList

saschanaz:
- [x] Summarize your changes


api/DOMConfiguration.json
```diff
@@ -1,244 +0,0 @@
-{
-  "api": {
-    "DOMConfiguration": {
-      "__compat": {
-        "mdn_url": "https://developer.mozilla.org/docs/Web/API/DOMConfiguration",
```
foolip, line 5: This went together with `document.domConfig`, which was once in Firefox:
https://github.com/mdn/browser-compat-data/blob/c1f99a1ee3166e343cd509d7e205a06f890fdc46/api/Document.json#L3525-L3532

But it was so long ago now that the entry can still be removed. Can you roll that into this change a well?

api/DOMImplementationList.json
```diff
@@ -1,148 +0,0 @@
-{
-  "api": {
-    "DOMImplementationList": {
```
foolip, line 3: This went together with https://www.w3.org/TR/DOM-Level-3-Core/core.html#DOMImplementationSource, which isn't in BCD. This removal seems appropriate.

foolip (approved):

foolip:
Over to @chrisdavidmills . This pages should be deleted:
https://developer.mozilla.org/en-US/docs/Web/API/DOMConfiguration
https://developer.mozilla.org/en-US/docs/Web/API/DOMImplementationList

These pages are 404s and no action needed:
https://developer.mozilla.org/en-US/docs/Web/API/DOMConfiguration/canSetParameter
https://developer.mozilla.org/en-US/docs/Web/API/DOMConfiguration/getParameter
https://developer.mozilla.org/en-US/docs/Web/API/DOMConfiguration/parameterNames
https://developer.mozilla.org/en-US/docs/Web/API/DOMConfiguration/setParameter
https://developer.mozilla.org/en-US/docs/Web/API/DOMImplementationList/item
https://developer.mozilla.org/en-US/docs/Web/API/DOMImplementationList/length

chrisdavidmills:
@foolip 

* https://wiki.developer.mozilla.org/en-US/docs/Web/API/Document/domConfig redirected to its parent element.
* https://developer.mozilla.org/en-US/docs/Web/API/DOMConfiguration deleted.
* https://developer.mozilla.org/en-US/docs/Web/API/DOMImplementationList deleted.

PR #6932, opened 2020-10-15
Add flags for NDEF APIs

saschanaz:
- [x] Summarize your changes


ddbeck:
Are all of these actually unimplemented? This bug is _super_ long but it looks like Web NFC might be at least partly implemented in Chromium behind a flag? https://bugs.chromium.org/p/chromium/issues/detail?id=520391

saschanaz:
Yes, I see [some interfaces implemented](https://github.com/chromium/chromium/blob/fa8fc338994b772ad68b673f2a5108dc7eeb9ff4/third_party/blink/renderer/modules/nfc/ndef_message.cc). We should figure out which flag controls them since which version.

vinyldarkscratch:
I was doing some research into flags with the mdn-bcd-collector, however we figured that it wasn't within the scope of the project.  While there _is_ a way to enable flags in a Selenium context, it's another to obtain the flag values in a reasonable, cross-platform method (and that's just for Chrome).  Additionally, we're not sure whether browser vendors want to include flag data in BCD (Joe mentioned a possible "no more flags" but hadn't followed up on it).

Bottom line: it would be better to add the flag data manually.

saschanaz:
Makes sense, maybe we can just add notes that it's being implemented.

vinyldarkscratch:
Okay, so I did a little bit more research into this to get to the bottom of this.  As it turns out, Web NFC was implemented not behind a flag, but an Origin Trial (which is, for those that aren't familiar with Origin Trials, a trial feature that can be enabled via https://developers.chrome.com/origintrials, and ends on a specific _date_ rather than a Chrome release).  Joe Medley has requested that Origin Trial data is not included in BCD, so for the purposes of BCD, we should mark the APIs as unimplemented.

See https://developers.chrome.com/origintrials/#/view_trial/236438980436951041 for details on the Origin Trial.

saschanaz:
Not applicable anymore, closing.

PR #6934, opened 2020-10-16
Remove never-supported IDBVersionChangeEvent#setVersion 

saschanaz:
- [x] Summarize your changes

Need to check the webkit-prefixed version had it.

### DIFF
--- a/api/IDBVersionChangeRequest.json
+++ b/api/IDBVersionChangeRequest.json
@@ -48,54 +48,6 @@
           "standard_track": false,
           "deprecated": true
         }
-      },
-      "setVersion": {
-        "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/IDBVersionChangeRequest/setVersion",
-          "support": {
-            "chrome": {
-              "version_added": false
-            },
-            "chrome_android": {
-              "version_added": false
-            },
-            "edge": {
-              "version_added": false
-            },
-            "firefox": {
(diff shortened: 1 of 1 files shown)


ddbeck:
Need to check the webkit-prefixed version had it.

Hey @saschanaz, did you ever check on this?

saschanaz:
No, and maybe we can ignore it as the interface itself was removed long ago, per the current BCD.

vinyldarkscratch (approved):
I did some testing in Chrome 22 to see if this was present in the webkit-prefixed version.  Doesn't look like it was!

ddbeck:
@vinyldarkscratch We need to fix (e.g., delete or archive) pages which lose BCD as a consequence of feature removals before merging. See https://github.com/mdn/browser-compat-data/pull/6929#issuecomment-717981716). I'll fix this one now, but let's avoid this going forward. Thank you!

ddbeck:
Wait, no. I can't fix this one; redirecting to the parent is pointless, since this feature is the only functional bit of the parent interface. We probably ought to have removed this feature and it's parent and both https://wiki.developer.mozilla.org/en-US/docs/Web/API/IDBVersionChangeRequest and https://wiki.developer.mozilla.org/en-US/docs/Web/API/IDBVersionChangeRequest/setVersion at the same time.

ddbeck:
OK, I've opened #7411 to try to fix this.

PR #6941, opened 2020-10-16
Remove never-supported SVGMeshElement

saschanaz:
- [x] Summarize your changes


vinyldarkscratch (approved):
Confirmed with results from the mdn-bcd-collector, thanks!

ddbeck:
A couple post hoc review notes here

- Seems to me that if this is true, then the [`<mesh>`](https://developer.mozilla.org/docs/Web/SVG/Element/mesh) element should also have been removed with it. I raised #7413.
- https://developer.mozilla.org/docs/Web/SVG/Element/mesh hasn't been deleted yet (I'm making a list of these for Chris to fix)

PR #6946, opened 2020-10-16
Remove never-supported CSS font-variant subfeatures

saschanaz:
- [x] Summarize your changes

### DIFF
--- a/css/properties/font-variant.json
+++ b/css/properties/font-variant.json
@@ -122,63 +122,6 @@
             }
           }
         },
-        "greek_accented_characters": {
-          "__compat": {
-            "description": "Greek accented characters",
-            "support": {
-              "chrome": {
-                "version_added": false,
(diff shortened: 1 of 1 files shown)


rachelandrew (commented):
Uppercase_eszett should stay, it's in a spec at CR.

The Greek accented chars isn't a value. This was added by @ddbeck so maybe he has the history https://github.com/mdn/browser-compat-data/pull/4608

rachelandrew:
Actually, I think I'm wrong there, I think these two things should be treated in the same way @ddbeck?

ddbeck:
Yeah, let's leave these intact for now. #4548 is probably interesting reading too. There's a compatability consequence for web developers to know that some systems may do the correct behavior, even if the browser hasn't implemented it in a consistent way. This was super weird to research in the first place and I feel kinda weird keeping it, but I don't have a better idea at the moment.

Thank you @rachelandrew and @saschanaz!

PR #6949, opened 2020-10-16
Add notes for unimplemented CSS mask-origin keywords

saschanaz:
- [x] Summarize your changes


rachelandrew (commented):
The spec containing these (CSS Masking) is a Candidate Recommendation so it is unlikely they will be removed, masking has fairly patchy implementation I think.

ddbeck (changes requested):
I have no idea why I let this sit around so long. Sorry about that, @saschanaz.

If they're going to stick around in the spec, then my only suggestion here is: can we flag these features as experimental? No implementations implies that, I think.

ddbeck (approved):
Thank you! 🎉

PR #6960, opened 2020-10-16
Remove never-supported WebExtensions features

saschanaz:
- [x] Summarize your changes

### DIFF
--- a/webextensions/match_patterns.json
+++ b/webextensions/match_patterns.json
@@ -195,32 +195,6 @@
             }
           }
         },
-        "ftps": {
-          "__compat": {
-            "support": {
-              "chrome": {
-                "version_added": false
-              },
-              "edge": {
-                "version_added": false
-              },
-              "firefox": {
-                "version_added": false,
-                "notes": "See <a href='https://bugzil.la/1463440'>bug 1463440</a>"
-              },
-              "firefox_android": {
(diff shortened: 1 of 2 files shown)


webextensions/match_patterns.json
```diff
@@ -195,32 +195,6 @@
...
-            "support": {
-              "chrome": {
-                "version_added": false
-              },
-              "edge": {
-                "version_added": false
-              },
-              "firefox": {
-                "version_added": false,
-                "notes": "See <a href='https://bugzil.la/1463440'>bug 1463440</a>"
```
saschanaz, line 209: (This bug is closed as wontfix.)

webextensions/manifest/manifest_version.json
```diff
@@ -85,36 +85,6 @@
               "deprecated": false
             }
           }
-        },
-        "v3": {
```
ddbeck, line 89: It looks like this feature will survive, after rebasing.

ddbeck (changes requested):
Hi @saschanaz, thanks for your patience while I got to this. I think this is ready to go, since there are no redirects needed on the content side. But it needs a rebase and I think the consequence of that will be that one of these features survives. Thanks again!

ddbeck (approved):
Thank you! 🎉

PR #6961, opened 2020-10-16
Remove never-supported CSS break-* features

saschanaz:
- [x] Summarize your changes


rachelandrew:
So the `always`, `recto` and `verso` values should stay. Probably worth noting that fragmentation is something I believe is being improved in Blink with LayoutNG so maybe we will see support of these at some point.

Regions is an interesting one, as the Regions spec as is, will probably never be implemented. I think at some point Regions will be brought back, and will definitely rely on fragmentation just like multicol. However given we have no active implementations of the old spec now that Edge is based on Chromium there probably isn't a reason to keep this in our data right now.

So I think we could lose the Regions stuff, but keep the values as mentioned above.

saschanaz:
@rachelandrew Is `always` really a thing? It seems [CSS Fragmentation Module](https://drafts.csswg.org/css-break-3/#break-between) does not define it. Somehow [Gecko supports it](https://phabricator.services.mozilla.com/D12211#inline-53116) saying that CSSWG resolved to add `always` but I can't confirm whether it is true.

rachelandrew:
Moved to Fragmentation 4: https://drafts.csswg.org/css-break-4/#valdef-break-before-always

ddbeck (approved):
Thank you! 🎉

PR #6964, opened 2020-10-16
set place-self version_added for safari(ios) from false to 11 (#1)

sideshowbarker:
This change is a follow up to #6847, which only updated the `place-self` Safari version for the `flex_context` case, but missed updating it for the `grid_context` case also. This change fixes that.


ddbeck (approved):
Thank you! 🎉

PR #6965, opened 2020-10-16
Update data for CSSKeyframeRule API

vinyldarkscratch:
This PR updates the data for the `CSSKeyframeRule` API.  I recently updated the `CSSKeyframesRule` API data (note the plural word form) in a few PRs, however it seems that the `CSSKeyframeRule` API (singular form) was also implemented at the same time.  This PR updates the `CSSKeyframeRule` API accordingly, as well as its subfeatures, based upon manual testing.

### DIFF
--- a/api/CSSKeyframeRule.json
+++ b/api/CSSKeyframeRule.json
@@ -19,7 +19,7 @@
               "version_added": "31"
             },
             {
-              "version_added": "1",
+              "version_added": "18",
               "version_removed": "31",
               "prefix": "WebKit"
             }
--- a/api/CSSKeyframeRule.json
+++ b/api/CSSKeyframeRule.json
@@ -80,21 +80,21 @@
           ],
           "safari": [
             {
-              "version_added": "9.1"
+              "version_added": "9"
             },
             {
               "version_added": "4",
-              "version_removed": "9.1",
+              "version_removed": "9",
               "prefix": "WebKit"
             }
           ],
           "safari_ios": [
             {
-              "version_added": "9.3"
+              "version_added": "9"
             },
             {
               "version_added": "3.2",
-              "version_removed": "9.3",
+              "version_removed": "9",
               "prefix": "WebKit"
             }
           ],
--- a/api/CSSKeyframeRule.json
+++ b/api/CSSKeyframeRule.json
@@ -4,12 +4,26 @@
       "__compat": {
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/CSSKeyframeRule",
         "support": {
-          "chrome": {
-            "version_added": true
-          },
-          "chrome_android": {
-            "version_added": "45"
-          },
+          "chrome": [
+            {
+              "version_added": "31"
+            },
+            {
+              "version_added": "1",
+              "version_removed": "31",
+              "prefix": "WebKit"
+            }
+          ],
+          "chrome_android": [
+            {
+              "version_added": "31"
+            },
+            {
+              "version_added": "1",
+              "version_removed": "31",
+              "prefix": "WebKit"
+            }
+          ],
           "edge": {
             "version_added": "12"
           },
@@ -19,7 +33,8 @@
             },
             {
               "version_added": "5",
-              "prefix": "moz"
+              "version_removed": "48",
+              "prefix": "Moz"
             }
           ],
           "firefox_android": [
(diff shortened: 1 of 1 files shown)


api/CSSKeyframeRule.json
```diff
@@ -28,31 +43,81 @@
...
+              "version_removed": "48",
+              "prefix": "Moz"
             }
           ],
           "ie": {
             "version_added": "10"
           },
-          "opera": {
-            "version_added": "12",
-            "prefix": "o"
```
foolip, line 39: This did seem to be wrong, there's no `OCSSKeyframeRule` or `oOCSSKeyframeRule` in Opera 12.16.

api/CSSKeyframeRule.json
```diff
@@ -28,31 +43,81 @@
...
+            {
+              "version_added": "18"
+            },
+            {
+              "version_added": "15",
+              "version_removed": "18",
+              "prefix": "WebKit"
+            },
+            {
+              "version_added": "12",
```
foolip, line 63: This doesn't seem right, there's no `CSSKeyframeRule` in Opera 12.1 or 12.16 in my testing on BrowserStack.

api/CSSKeyframeRule.json
```diff
@@ -28,31 +43,81 @@
...
+            {
+              "version_added": "18"
+            },
+            {
+              "version_added": "14",
+              "version_removed": "18",
+              "prefix": "WebKit"
+            },
+            {
+              "version_added": "12",
```
foolip, line 77: I haven't tested this, but I'd bet it's also wrong.

api/CSSKeyframeRule.json
```diff
@@ -4,12 +4,26 @@
...
         "support": {
-          "chrome": {
-            "version_added": true
-          },
-          "chrome_android": {
-            "version_added": "45"
-          },
+          "chrome": [
+            {
+              "version_added": "31"
```
foolip, line 9: Confirmed by looking up the date of https://chromium.googlesource.com/chromium/src/+/5a98a4be279589ff8b93bf32a34d38f380de4946 in https://www.chromium.org/developers/calendar.

api/CSSKeyframeRule.json
```diff
@@ -28,31 +43,81 @@
...
+              "prefix": "WebKit"
+            },
+            {
+              "version_added": "12",
+              "version_removed": "14"
+            }
+          ],
+          "safari": [
+            {
+              "version_added": "9.1"
```
foolip, line 83: The change was in https://trac.webkit.org/changeset/176157/webkit, at WebKit trunk version 601.1.10. Based on that I would have expected it to be in Safari 9. It looks like we have test results for 9.1 but not 9 in mdn-bcd-results. Does the script incorrectly infer 9.1, or was this a range version where you removed the range?

api/CSSKeyframeRule.json
```diff
@@ -86,19 +151,19 @@
               "version_added": "12"
             },
             "opera_android": {
-              "version_added": true
+              "version_added": "12"
```
foolip, line 154: Is this supported by test results, or just mirrored from Opera desktop? I suspect it's wrong, since the interface object isn't exposed.

api/CSSKeyframeRule.json
vinyldarkscratch: It isn't exposed on its own in that Opera version; one would access it through `CSSKeyframesRule.cssRules`.  I'll add a test into the collector's custom tests!

foolip: Ah, so just a problem of a missing interface object.

api/CSSKeyframeRule.json
vinyldarkscratch: In BrowserStack and SauceLabs, we unfortunately don't have a way to differentiate 9 and 9.1 in testing.  (This is probably a good reason to get more testing VMs.) I think I determined the version number from reading through the documentation, but I can bump it down to Safari 9!

foolip: I see. As a reviewer, when this uncertainty exists due to a lack of browser versions to test, it would be great if it's made explicit that it's a guess. Where it's recent enough or there are other entries I'd expect it to be in sync with, I might then hunt down the WebKit commit to confirm which of the two versions seems more likely.

Let's go with "9" here.

foolip (approved):

PR #6968, opened 2020-10-16
Update data for CSSMediaRule API

vinyldarkscratch:
This PR updates the compat data for the `CSSMediaRule` API based upon a combination of results from the mdn-bcd-collector, as well as manual testing.

### DIFF
--- a/api/CSSMediaRule.json
+++ b/api/CSSMediaRule.json
@@ -8,7 +8,7 @@
             "version_added": "1"
           },
           "chrome_android": {
-            "version_added": "1"
+            "version_added": "18"
           },
           "edge": {
             "version_added": "12"
--- a/api/CSSMediaRule.json
+++ b/api/CSSMediaRule.json
@@ -5,42 +5,42 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/CSSMediaRule",
         "support": {
           "chrome": {
-            "version_added": "45"
+            "version_added": "1"
           },
           "chrome_android": {
-            "version_added": "45"
+            "version_added": "1"
           },
           "edge": {
             "version_added": "12"
           },
           "firefox": {
-            "version_added": "17",
+            "version_added": "1",
             "notes": "Before Firefox 20, <code>conditionText</code> could not be set."
           },
           "firefox_android": {
-            "version_added": "17",
+            "version_added": "4",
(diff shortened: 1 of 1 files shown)


sideshowbarker (approved):

api/CSSMediaRule.json
```diff
@@ -54,40 +54,40 @@
...
-              "version_added": true
+              "version_added": "≤12.1"
             },
             "opera_android": {
-              "version_added": true
+              "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": true
+              "version_added": "≤3"
```
sideshowbarker, line 81:
```
              "version_added": "≤3"
```

Wasn’t the decision to do with `≤4` rather than `≤3`? (I can’t remember of that was the resolution or not…)

api/CSSMediaRule.json
```diff
@@ -5,42 +5,42 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/CSSMediaRule",
         "support": {
           "chrome": {
-            "version_added": "45"
+            "version_added": "1"
           },
           "chrome_android": {
-            "version_added": "45"
+            "version_added": "1"
```
sideshowbarker, line 11: I see the linter is saying, *“version_added: "1" is NOT a valid version number for chrome_android”*

api/CSSMediaRule.json
vinyldarkscratch: In that PR, I added both ranged versions.  Since the VMs I have can go as far back as Safari 3 in an efficient manner, I figured it would make sense to keep them both for more real data sooner.

sideshowbarker: > In that PR, I added both ranged versions. Since the VMs I have can go as far back as Safari 3 in an efficient manner, I figured it would make sense to keep them both for more real data sooner.

Got it — thanks for clarifying

sideshowbarker (approved):

api/CSSMediaRule.json
vinyldarkscratch: I was also able to get a new set of Safari 1-2 emulators working that function much, much better than the last ones did, which means more testing opportunities, and clearing the ranged versions in this PR, so this is now ready to go!

sideshowbarker (approved):

Issue #6905, opened 2020-10-13
Improve documentation of experimental status


ddbeck:
[The schema docs describe `status.experimental` as](https://github.com/mdn/browser-compat-data/blob/master/schemas/compat-data-schema.md#status-information):

> a `boolean` value that indicates this functionality is intended to be an addition to the Web platform. Some features are added to conduct tests. Set to `false`, it means the functionality is mature, and no significant incompatible changes are expected in the future.

This text doesn't quite capture the meaning of the data as we've come to update it. For consumers, we should revise the wording to more clearly articulate the meaning of the `true` and `false` states; for contributors, we should write contribution guidelines for choosing between the two values.

Once these docs and the guideline are written, then we should start follow-up issues to clean up any data accordingly.

ddbeck:
For the schema docs, I propose something like:

> a `boolean` value that indicates whether the feature is a stable part of the Web platform. Set to `false`, Web developers should consider the feature subject to ongoing development and at high risk of compatibility-breaking changes; set to `false`, Web developers should expect the feature's behavior to remain the same over time.

For the data guidelines, I propose something like:

> ## Choosing an experimental status (`status.experimental`)
>
> If a feature is standard track and only one browser engine supports the feature without prefixes or flags, always choose `true` for `status.experimental`.
>
> If a feature is standard track and more than one browser engine supports the feature without prefixes or flags, always choose `false` for `status.experimental`.

This feels rather definitive (in fact, we could probably automate this, as written). But it leaves out non-standard features. I can imagine scenarios were features are non-standard and experimental (e.g., some vendor decides to go it alone on something), or where features are non-standard non-experimental (e.g., a non-standard feature has been explicitly deprecated or in the process of removal). I'm not sure how exactly to phrase that and have run out of time to think about it this afternoon.

ddbeck:
Also, #5392 is important reading for this issue.

foolip:
The most important effect of the `experimental` status in BCD is the flask icon with the title text "Experimental. Expect behavior to change in the future." on MDN. ([example](https://developer.mozilla.org/en-US/docs/Web/API/ShadowRoot#Browser_compatibility)) Assuming that isn't changing on the MDN side, then I think the meaning of `experimental` in BCD should be close to "behavior is expected to change in the future".

[...]

PR #6979, opened 2020-10-19
Add Feature-Policy:web-share

hamishwillee:
This confirms Feature-Policy web-share is supported in FF81 but makes the note that the protected feature (Web Share API) is itself not implemented. The bug that should track the implementation is included in that note.

I have set the other browsers as `null`. It is possible that some of them do support this, but there is no obvious information and no easy way to test (i.e. to test would take more effort than knowing would be worth).

### DIFF
--- a/http/headers/feature-policy.json
+++ b/http/headers/feature-policy.json
@@ -1853,6 +1853,56 @@
             }
           }
         },
+        "web-share": {
+          "__compat": {
+            "mdn_url": "https://developer.mozilla.org/docs/Web/HTTP/Headers/Feature-Policy/web-share",
+            "support": {
+              "chrome": {
+                "version_added": false
+              },
+              "chrome_android": {
+                "version_added": false
+              },
+              "edge": {
+                "version_added": false
+              },
+              "firefox": {
+                "version_added": "81",
+                "notes": "Firefox recognizes the <code>web-share</code> permissions policy, but this has no effect in versions of Firefox that do not support the <a href='https://developer.mozilla.org/docs/Web/API/Navigator/share'><code>share()</code></a> method."
+              },
+              "firefox_android": {
+                "version_added": "81",
+                "notes": "Firefox recognizes the <code>web-share</code> permissions policy, but this has no effect in versions of Firefox that do not support the <a href='https://developer.mozilla.org/docs/Web/API/Navigator/share'><code>share()</code></a> method."
+              },
+              "ie": {
+                "version_added": false
+              },
+              "opera": {
+                "version_added": false
+              },
+              "opera_android": {
+                "version_added": false
+              },
+              "safari": {
+                "version_added": false
+              },
+              "safari_ios": {
+                "version_added": false
+              },
+              "samsunginternet_android": {
+                "version_added": false
+              },
+              "webview_android": {
+                "version_added": false
+              }
+            },
+            "status": {
+              "experimental": true,
+              "standard_track": true,
+              "deprecated": false
+            }
+          }
+        },
         "xr-spatial-tracking": {
           "__compat": {
             "mdn_url": "https://developer.mozilla.org/docs/Web/HTTP/Headers/Feature-Policy/xr-spatial-tracking",


http/headers/feature-policy.json
```diff
@@ -1853,6 +1853,56 @@
...
+              },
+              "samsunginternet_android": {
+                "version_added": null
+              },
+              "webview_android": {
+                "version_added": null
+              }
+            },
+            "status": {
+              "experimental": true,
```
hamishwillee, line 1900: FYI this spec is draft still.  This is no longer marked as experimental in Mozilla code though.

hamishwillee:
@ddbeck Is this something you might look at?

http/headers/feature-policy.json
```diff
@@ -1853,6 +1853,56 @@
             }
           }
         },
+        "web-share": {
+          "__compat": {
+            "mdn_url": "https://developer.mozilla.org/docs/Web/HTTP/Headers/Feature-Policy/web-share",
+            "support": {
+              "chrome": {
+                "version_added": null
```
ddbeck, line 1861: This appears to have not yet shipped in Chromium: https://www.chromestatus.com/feature/6362499966304256. I'd apply this change to all the Chromium browsers (Chrome, Edge, Opera, Samsung Internet, and WebView)

```suggestion
                "version_added": false
```

http/headers/feature-policy.json
```diff
@@ -1853,6 +1853,56 @@
...
+                "version_added": false
+              },
+              "opera": {
+                "version_added": null
+              },
+              "opera_android": {
+                "version_added": null
+              },
+              "safari": {
+                "version_added": null
```
ddbeck, line 1887: I couldn't find evidence that WebKit has implemented this: https://bugs.webkit.org/show_bug.cgi?id=214448. I'd set both Safaris to `false` too.

http/headers/feature-policy.json
```diff
@@ -1853,6 +1853,56 @@
...
+              },
+              "chrome_android": {
+                "version_added": null
+              },
+              "edge": {
+                "version_added": null
+              },
+              "firefox": {
+                "version_added": "81",
+                "notes": "Firefox 81 enables support for the Feature-Policy web-share, but does not support the WebShare API. See <a href='https://bugzil.la/1653481'>bug 1653481</a>."
```
ddbeck, line 1871: If Firefox doesn't implement `navigator.share()`, then it seems to me that the header doesn't really have the specified effect of the policy, which is to define the conditions under which `navigator.share()` throws a `NotAllowedError`. This is _technically_ a case where `"partial_implementation"` should be `true`, but that's a bit misleading too. Taken together, I think it means we should be a bit more circumspect about "support" here. Perhaps we could say something like:

```suggestion
                "notes": "Firefox recognizes the <code>web-share</code> permissions policy, but this has no effect in versions of Firefox that do not support the <a href='https://developer.mozilla.org/docs/Web/API/Navigator/share'><code>share()</code></a> method."
```

I've dropped the bug link because, ideally, the `share()` method itself is going to be more interesting and actionable for web developers.

ddbeck (changes requested):
Thanks for this PR, @hamishwillee. A few comments in line comments. Also, could you link to a bug or test or something to substantiate the Firefox implementation claim? It's nice to have a record of why set values to what we did. Thank you!

http/headers/feature-policy.json
ddbeck: #6905 is open to cover this better, but "experimental" in BCD is likely to mean "only has one implementation" (or near enough to it) in the future. This is value is good for now! 👍

http/headers/feature-policy.json
hamishwillee: Fixed (and rebased). Thank you!

hamishwillee:
Hi @ddbeck 
Thanks for all the help. I have marked up all the non-Firefox cases as false, and accepted your note for both the firefox versions.

> Thanks for this PR, @hamishwillee. A few comments in line comments. Also, could you link to a bug or test or something to substantiate the Firefox implementation claim? It's nice to have a record of why set values to what we did. Thank you!

Isn't that the bug report you thought we should remove? https://bugzilla.mozilla.org/show_bug.cgi?id=1653199
Either way, it went in in FF81 and that is the bug that says so! Should I add this back into the text?

ddbeck:
Sorry for being unclear, @hamishwillee. I meant to ask for you to link to the bug in the PR description—not the notes.

> Isn't that the bug report you thought we should remove? https://bugzilla.mozilla.org/show_bug.cgi?id=1653199
> Either way, it went in in FF81 and that is the bug that says so! Should I add this back into the text?

Ah, this is the key thing. Thank you! Your original note linked to [bug 1653481](https://bugzilla.mozilla.org/show_bug.cgi?id=1653481), ("[Meta] Implement and maintain WebShare API"). This new [bug 1653199](https://bugzilla.mozilla.org/show_bug.cgi?id=1653199) ("Web Share should be controlled by permission policy"), is what I was thinking of for the PR record keeping. I just needed something, apart from the diff, which claimed that the stuff _in_ the diff was true. 😄

ddbeck (approved):
This is ready to go now. Thank you! 🎉

PR #6981, opened 2020-10-19
Add CSS math-style property

rachelandrew:
I'm working on the CSS updates for Firefox 83 in https://github.com/mdn/sprints/issues/3800

This PR adds BCD for the `math-style` property (I haven't created the page yet). It is behind a flag in Firefox and Chrome.

- Firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1665975
- Chromium: https://chromium-review.googlesource.com/c/chromium/src/+/2100787

(Recently added to WebKit but not in Safari yet https://bugs.webkit.org/show_bug.cgi?id=216702)

### DIFF
--- a/css/properties/math-style.json
+++ b/css/properties/math-style.json
@@ -15,7 +15,13 @@
               ]
             },
             "chrome_android": {
-              "version_added": false
+              "version_added": "83",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "Enable experimental Web Platform features"
+                }
+              ]
             },
             "edge": {
               "version_added": false
@@ -31,7 +37,14 @@
               ]
             },
             "firefox_android": {
-              "version_added": false
+              "version_added": "83",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "layout.css.math-style.enabled",
+                  "value_to_set": "true"
+                }
+              ]
             },
             "ie": {
               "version_added": false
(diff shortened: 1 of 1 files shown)


css/properties/math-style.json
```diff
@@ -0,0 +1,67 @@
...
+              "version_added": "83",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "Enable experimental Web Platform features"
+                }
+              ]
+            },
+            "chrome_android": {
+              "version_added": false
```
ddbeck, line 18: I don't see anything in the diff which suggests that this has been switched off for Android. Can you copy the data from chrome to chrome_android? Or show me how wrong I am?

css/properties/math-style.json
```diff
@@ -0,0 +1,67 @@
...
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "layout.css.math-style.enabled",
+                  "value_to_set": "true"
+                }
+              ]
+            },
+            "firefox_android": {
+              "version_added": false
```
ddbeck, line 34: Likewise for firefox→firefox_android

ddbeck (changes requested):
A couple minor suggestions here, to mirror the data for the mobile browsers. Otherwise, this is good. Thank you, @rachelandrew!

ddbeck (approved):
Looks good. Thank you! 🎉

Siilwyn:
Shouldn't `status.experimental be `true`?

PR #6982, opened 2020-10-19
Drop the HTMLInputElement.p.moz* methods

sideshowbarker:
This change drops the `mozGetFileNameArray()`, `mozSetFileNameArray()`, and `mozSetFileArray()` methods of `HTMLInputElement.` They’re not exposed to web content in any UAs — not even Gecko. See https://github.com/mdn/browser-compat-data/pull/6982#issuecomment-715895179


evilpie:
These methods are marked as `[ChromeOnly]` so they should probably be marked as removed for Firefox as well.  https://searchfox.org/mozilla-central/source/dom/webidl/HTMLInputElement.webidl#157

foolip:
@saschanaz should these be removed in https://github.com/mdn/browser-compat-data/pull/6931? Is the reason they're not because BCD says it's supported?

saschanaz:
> @saschanaz should these be removed in #6931? Is the reason they're not because BCD says it's supported?

#6931 only covers things that only filled with explicit `false`. In this case there is a support from Firefox and there are some `null`s, so these are not covered there.

foolip:
I see, thanks @saschanaz!

foolip:
@sideshowbarker do you want to just remove these entries instead?

sideshowbarker:
> @sideshowbarker do you want to just remove these entries instead?

Yes — I’ve now amended the commit here to do that

foolip:
Passing this along to @chrisdavidmills. These two pages need to be redirected to the parent page:
https://developer.mozilla.org/docs/Web/API/HTMLInputElement/mozSetFileNameArray
https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/mozSetFileNameArray

https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/mozSetFileArray didn't exist.

chrisdavidmills:
@foolip redirects put in place.

foolip (approved):

PR #6983, opened 2020-10-19
Mark InstallEvent and subfeatures non-standard

sideshowbarker:
This change marks the `InstallEvent` interface and all its subfeatures `standard_track:false` and `deprecated:true`.

https://github.com/w3c/ServiceWorker/commit/4d17228 (in 2015) completely dropped `InstallEvent` from the spec.


ddbeck (approved):
Looks good. Thank you! 🎉

PR #6987, opened 2020-10-19
Drop window.restore (unspecified & unimplemented)

sideshowbarker:
This drops the `window.restore()` method. The method isn’t implemented in any browser engines, and isn’t part of any spec.

---

I think this one was not included yet in any of the *Remove never-supported* PRs that @saschanaz raised so far.

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -7525,54 +7525,6 @@
           }
         }
       },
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Thank you! 🎉

PR #6989, opened 2020-10-19
Mark CSSPositionValue & subfeatures nonstandard

sideshowbarker:
This change marks the `CSSPositionValue` interface non-standard. https://github.com/w3c/css-houdini-drafts/commit/5261c1a dropped it, so it’s no longer part of any spec.


ddbeck (approved):
Looks good. Thank you! 🎉

PR #6990, opened 2020-10-19
Mark CanvasRenderingContext2D.p.currentTransform non-standard

sideshowbarker:
This change marks `CanvasRenderingContext2D.p.currentTransform` non-standard. https://github.com/whatwg/html/commit/bb00977 dropped it from the specification.


ddbeck (approved):
Thank you! 🎉

PR #6991, opened 2020-10-19
Update data for CSSSupportsRule API

vinyldarkscratch:
This PR updates the compat data for the `CSSSupportsRule` API based upon a combination of results from the mdn-bcd-collector, as well as manual testing.

Notice: this removes deprecated flag data for Firefox.

### DIFF
--- a/api/CSSSupportsRule.json
+++ b/api/CSSSupportsRule.json
@@ -35,10 +35,10 @@
             }
           ],
           "safari": {
-            "version_added": "10.1"
+            "version_added": "9"
           },
           "safari_ios": {
-            "version_added": "10.3"
+            "version_added": "9"
           },
           "samsunginternet_android": {
             "version_added": "1.5"
--- a/api/CSSSupportsRule.json
+++ b/api/CSSSupportsRule.json
@@ -5,57 +5,54 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/CSSSupportsRule",
         "support": {
           "chrome": {
-            "version_added": false
+            "version_added": "41"
           },
           "chrome_android": {
-            "version_added": false
+            "version_added": "41"
           },
           "edge": {
-            "version_added": "≤18",
-            "version_removed": "79"
+            "version_added": "12"
           },
           "firefox": {
-            "version_added": "17",
-            "notes": "From Firefox 17 to 19, methods and properties were defined on <code>CSSSupportsRule</code>. From version 20, they were on <code>CSSConditionRule</code>.",
-            "flags": [
-              {
-                "type": "preference",
-                "name": "layout.css.supports-rule.enable",
-                "value_to_set": "true"
-              }
-            ]
+            "version_added": "22",
+            "notes": "Methods and properties are defined on <code>CSSConditionRule</code>."
           },
           "firefox_android": {
-            "version_added": "17",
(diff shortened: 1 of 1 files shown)


api/CSSSupportsRule.json
```diff
@@ -5,57 +5,54 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/CSSSupportsRule",
         "support": {
           "chrome": {
-            "version_added": false
+            "version_added": "41"
```
foolip, line 8: This interface was originally not exposed, and I fixed that in https://chromium.googlesource.com/chromium/src/+/3e3b38c7fdf9a0753b369576713578aa0895262c, which was M41 per https://www.chromium.org/developers/calendar.

Testing http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8599 in Chrome 40 will confirm earlier support. Can you add a custom test for this? We should mark the initial range as `partial_implementation`.

api/CSSSupportsRule.json
```diff
@@ -5,57 +5,54 @@
...
           },
           "edge": {
-            "version_added": "≤18",
-            "version_removed": "79"
+            "version_added": "12"
           },
           "firefox": {
-            "version_added": "17",
-            "notes": "From Firefox 17 to 19, methods and properties were defined on <code>CSSSupportsRule</code>. From version 20, they were on <code>CSSConditionRule</code>.",
-            "flags": [
```
foolip, line 20: Can you mention this removal in the description, as requested by @ddbeck in another PR?

api/CSSSupportsRule.json
```diff
@@ -5,57 +5,54 @@
...
-              {
-                "type": "preference",
-                "name": "layout.css.supports-rule.enable",
-                "value_to_set": "true"
-              }
-            ]
+            "version_added": "22",
+            "notes": "Methods and properties are defined on <code>CSSConditionRule</code>."
```
foolip, line 18: This doesn't need a note, it's the way it should be per https://drafts.csswg.org/css-conditional/#the-csssupportsrule-interface.

However, if the interface was exposed in Firefox 17, then I think we should keep that. Whether to include a note about the members is debatable, but I'd probably say no, a note about more things being supported than is represented in the data isn't really useful.

api/CSSSupportsRule.json
```diff
@@ -5,57 +5,54 @@
...
           },
-          "opera": {
-            "version_added": "12.1"
-          },
-          "opera_android": {
-            "version_added": false
-          },
+          "opera": [
+            {
+              "version_added": "28"
```
foolip, line 29: This may also need updating after fixing the Chrome version.

api/CSSSupportsRule.json
```diff
@@ -5,57 +5,54 @@
...
+            },
+            {
+              "version_added": "12.1",
+              "version_removed": "14"
+            }
+          ],
           "safari": {
-            "version_added": false
+            "version_added": "10.1"
```
foolip, line 46: Can you check with http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8599 if it wasn't actually supported earlier but missing the interface object?

foolip:
Ping @vinyldarkscratch

vinyldarkscratch:
This PR was created before the custom test for this API was implemented.  I updated the data using the newest results.

foolip (approved):

Issue #6875, opened 2020-10-09
Public static fields from Public class fields compatibility table is incorrect for iOS Safari


shant24k:
- This's about: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Public_class_fields

Here Public static fields not working on iOS 14 Safari. It's encountering syntax error: `Uncaught SyntaxError: Unexpected token '='. Expected an opening '(' before a method's parameter list`. (Below code ending with this error)

 

```

class ClassWithStaticField {

  static staticField = 'static field'

}

 

console.log(ClassWithStaticField.staticField)

```

But other parts of same API i.e. Public instance fields & Public Methods do work in iOS 14. Looks like, there's no issue with public static methods but there's issue with Public static fields.

 

- Also, there's problem with `#` prefix in private class field & methods in iOS 14. It's throwing syntax error: `Uncaught SyntaxError: Invalid character: '#'`

sideshowbarker:
As far as the case of public *static* fields specifically, I think the fact they’re not supported is a known limitation, with work in progress at https://bugs.webkit.org/show_bug.cgi?id=194095

But public *instance* fields are supported. See the test at https://repl.it/@jonasjelonek/Safari-14-public-class-fields#script.js and see https://trac.webkit.org/changeset/254653/webkit and #6413.

So I think as far as BCD, what we’ll need to do here is, add a note for now which explicitly states that public static fields aren’t yet supported in Safari (and then update that later after https://bugs.webkit.org/show_bug.cgi?id=194095 lands). For that, I’ll raise a separate PR with a patch that adds such a note.

I hope @caitp and @xanlpz can please correct me if I got anything wrong above about the current state of Safari support. Heads-up also to @jonasjelonek (who contributed the BCD patch in #6413).

sideshowbarker:
I’ve raised https://github.com/mdn/browser-compat-data/pull/6994 for this.

xanlpz:
> I hope @caitp and @xanlpz can please correct me if I got anything wrong above about the current state of Safari support. Heads-up also to @jonasjelonek (who contributed the BCD patch in #6413).

This is correct. Also keep in mind that private static fields are in a separate patch, so they might not land at the same time than public static fields.

sideshowbarker:
> This is correct. Also keep in mind that private static fields are in a separate patch, so they might not land at the same time than public static fields.

OK — thanks much for the clarificaion

PR #6994, opened 2020-10-20
Mark Safari support for JS public class fields partial, with note

sideshowbarker:
This change marks the support in Safari for public class fields `partial_implementation:true` — due to the fact that while Safari does support public *instance* fields, it does not yet support public *static* fields.
See https://bugs.webkit.org/show_bug.cgi?id=194095

Fixes https://github.com/mdn/browser-compat-data/issues/6875

### DIFF
--- a/javascript/classes.json
+++ b/javascript/classes.json
@@ -560,10 +560,14 @@
               "version_added": "51"
             },
             "safari": {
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Thanks! 🎉

Issue #6826, opened 2020-10-05
Decide what to do about the data for the dragexit event (drop completely? maintain the data indefinitely?) and ondragexit event handler


sideshowbarker:
See discussion in https://github.com/mdn/browser-compat-data/pull/6710

Unless we have evidence that developers are actually relying on `dragexit` and `ondragexit` (despite them apparently only having been implemented in Gecko, and not in any other engines), I would think that developers aren’t going miss them being gone from BCD, if were to drop them.

The alternative is to keep entries for them in BCD. But if we do that, it’s not clear to me at least what the plan is for retaining them. We keep them in BCD forever?

ddbeck:
I agree that it's unlikely that developers are going to miss them if we drop the data. However, I do think BCD's posture to date has been to record things as they exist, even if they're not advisable to use. This is somewhat reflected in the [guideline for irrelevant features](https://github.com/mdn/browser-compat-data/blob/master/docs/data-guidelines.md#removal-of-irrelevant-features), which suggests dropping data only when it stops shipping in any browser.

> The alternative is to keep entries for them in BCD. But if we do that, it’s not clear to me at least what the plan is for retaining them. We keep them in BCD forever?

Ideally, Gecko drops the feature someday, which starts the clock on removing it from BCD under the irrelevance guideline. [Bug 1635018](https://bugzilla.mozilla.org/show_bug.cgi?id=1635018) proposes removing it. I'm partly inclined to comment there that tracking this feature is a drag (pun intended) on BCD. But if it's never removed from the engine and Firefox keeps shipping it, then I suppose it stays in BCD indefinitely.

All that said, if you still think it's a good idea to drop this data while it's still in an actively developer browser, then I'd like to hear your ideas for a guideline that would allow for it gracefully.

sideshowbarker:
> … BCD's posture to date has been to record things as they exist, even if they're not advisable to use. This is somewhat reflected in the [guideline for irrelevant features](https://github.com/mdn/browser-compat-data/blob/master/docs/data-guidelines.md#removal-of-irrelevant-features), which suggests dropping data only when it stops shipping in any browser.
> …
> Ideally, Gecko drops the feature someday, which starts the clock on removing it from BCD under the irrelevance guideline. [Bug 1635018](https://bugzilla.mozilla.org/show_bug.cgi?id=1635018) proposes removing it. 

OK, that’s all clear to me now. I honestly had just not be aware yet of what the existing guidelines has to say for this.

[...]

PR #6995, opened 2020-10-20
Update dragexit data to match ondragexit data

sideshowbarker:
This change updates the incomplete version info we had in the support data for the `dragexit` event; the change just copies over the more-complete version info from the support data for the `ondragexit` event handler.

Relates to https://github.com/mdn/browser-compat-data/issues/6826

---

Since in #6826 we decided to keep `dragexit` in BCD, we should at least make the support data for it accurate — by syncing it up with what we already have for the `ondragexit` event handler.

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -3354,25 +3354,26 @@
               "version_added": false
             },
             "edge": {
-              "version_added": false
+              "version_added": "12",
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Thank you! 🎉

PR #6996, opened 2020-10-20
Update the data for the management API

Dexterp37:
The various notifications (e.g. `onInstalled`) were implemented as part of [bug 1336908](https://bugzilla.mozilla.org/show_bug.cgi?id=1336908), which landed in Firefox 55.

- [ ] Summarize your changes

### DIFF
--- a/webextensions/api/management.json
+++ b/webextensions/api/management.json
@@ -288,10 +288,12 @@
                 "version_added": "79"
               },
               "firefox": {
-                "version_added": false
+                "version_added": "55",
+                "notes": "Before version 56, only extensions whose <code>type</code> is <code>'theme'</code> are supported."
               },
               "firefox_android": {
-                "version_added": false
+                "version_added": "55",
+                "notes": "Before version 56, only extensions whose <code>'type'</code> is <code>'theme'<code> are supported."
               },
               "opera": {
                 "version_added": true
@@ -313,10 +315,12 @@
                 "version_added": "79"
               },
               "firefox": {
-                "version_added": false
+                "version_added": "55",
+                "notes": "Before version 56, only extensions whose <code>type</code> is <code>'theme'</code> are supported."
               },
               "firefox_android": {
-                "version_added": false
+                "version_added": "55",
+                "notes": "Before version 56, only extensions whose <code>type</code> is <code>'theme'</code> are supported."
               },
               "opera": {
                 "version_added": true
@@ -338,10 +342,12 @@
                 "version_added": "79"
               },
               "firefox": {
-                "version_added": false
+                "version_added": "55",
+                "notes": "Before version 56, only extensions whose <code>type</code> is <code>'theme'</code> are supported."
               },
               "firefox_android": {
-                "version_added": false
+                "version_added": "55",
+                "notes": "Before version 56, only extensions whose <code>type</code> is <code>'theme'</code> are supported."
               },
               "opera": {
                 "version_added": true
@@ -363,10 +369,18 @@
                 "version_added": "79"
               },
               "firefox": {
-                "version_added": false
+                "version_added": "55",
+                "notes": [
+                  "Before version 56, only extensions whose <code>type</code> is <code>'theme'</code> are supported.",
+                  "This event is not emitted when the extension is in the \"pending uninstall\" state. The event is emitted as expected once the extension is completely removed (for example, when the <code>about:addons</code> tab is closed)."
+                ]
               },
               "firefox_android": {
-                "version_added": false
+                "version_added": "55",
+                "notes": [
+                  "Before version 56, only extensions whose <code>type</code> is <code>'theme'</code> are supported.",
+                  "This event is not emitted when the extension is in the \"pending uninstall\" state. The event is emitted as expected once the extension is completely removed (for example, when the <code>about:addons</code> tab is closed)."
+                ]
               },
               "opera": {
                 "version_added": true


webextensions/api/management.json
```diff
@@ -338,10 +338,10 @@
                 "version_added": "79"
               },
               "firefox": {
-                "version_added": false
+                "version_added": "55"
               },
               "firefox_android": {
-                "version_added": false
+                "version_added": "55"
               },
```
rpl, line 345: Nit, technically all the management events have been added in Firefox 55 but:
- in Firefox 55, all the events does only support themes (as a confirmation here is `allowedTypes` as landed from [Bug 1336908](https://bugzilla.mozilla.org/show_bug.cgi?id=1336908): https://hg.mozilla.org/mozilla-central/rev/0df833bf2b9f#l3.81)
- in Firefox 56, the read-only part of the API (basically management.get/management.getAll and these events) have been extended to also support extensions (see here the `allowedTypes` as landed from [Bug 1282981](https://bugzilla.mozilla.org/show_bug.cgi?id=1282981): https://hg.mozilla.org/mozilla-central/rev/fe3ed659f568#l2.13)

If we want to capture that in the compatibility table, we could add a `notes` property that points that the API supports the "extension" add-on type starting from Firefox 56.

webextensions/api/management.json
```diff
@@ -363,10 +363,10 @@
...
```
rpl, line 370: For onUninstalled we may want to point out in a note that the onUninstalled events is not emitted when the extension is in the "pending uninstall" state, once it is completely removed (e.g. when the about:addons tab is closed or refreshed) the onUninstalled event is emitted as expected.

rpl (commented):
Thanks @Dexterp37! This looks good, but there are some notes (described below in the inline review comments) that it may be reasonable to add into the compatibility table.

In my opinion the note on the `onUninstalled` event is likely the most important one to include, because I doubt that anyone would really be noticing the differences between Firefox 55 and 56, given that are both old and none of them is an ESR release).

Let me know what you think.

webextensions/api/management.json
Dexterp37: Thanks! I updated this to add a note :)

webextensions/api/management.json
Dexterp37: Thanks! I updated this to add a note :)

webextensions/api/management.json
```diff
@@ -338,10 +338,12 @@
                 "version_added": "79"
               },
               "firefox": {
-                "version_added": false
+                "version_added": "55",
+                "notes": "Before version 56, only extensions whose 'type' is 'theme' are supported."
```
rpl, line 342: @Dexterp37 the same note should also be added to `onEnabled` and `onDisabled`

rpl (changes requested):
Thanks for adding the notes, we should add the same note we added on the `onInstall` event also to onEnabled and onDisabled..

rpl (approved):
lgtm \o/

thanks again for updating this metadata file

rpl:
@ddbeck the updates in this pull request looks good from my perspective (from a "Firefox WebExtensions module peer" perspective)

would you mind to give it a look and see if there is anything to change? (e.g. from an editorial perspective, and for consistency with the conventions used in this repo and/or the other WebExtensions API).

Feel free to merge once this is approved from your side too.

webextensions/api/management.json
```diff
@@ -288,10 +288,12 @@
...
```
ddbeck, line 292: style: If I understand correctly, this refers to an `ExtensionInfo` property. If so, I'd probably mark up the note (and its siblings/cousins below) like this:

```suggestion
                "notes": "Before version 56, only extensions whose <code>type</code> is <code>'theme'</code> are supported."
```

webextensions/api/management.json
```diff
@@ -363,10 +369,12 @@
                 "version_added": "79"
               },
               "firefox": {
-                "version_added": false
+                "version_added": "55",
+                "notes": "Before version 56, only extensions whose 'type' is 'theme' are supported. Also note that this event is not emitted when the extension is in the \"pending uninstall\" state: once it is completely removed (e.g. when the `about:addons` tab is closed or refreshed) the event is emitted as expected."
```
ddbeck, line 373: suggestion: BCD accepts an array of notes, when the notes refer to different issues. So I'd recommend breaking this up into something like this:



```
[
  "Before version 56, only extensions whose <code>type</code> is <code>'theme'</code> are supported".,
  "This event is not emitted when the extension is in the \"pending uninstall\" state. The event is emitted as expected once the extension is completely removed (for example, when the <code>about:addons</code> tab is closed)."
]
```

(OK, I also did some wordsmithing and markup changes. I'm a technical writer and I cannot help myself. 😄)

This kind of change would need to be applied to the other below note as well.

ddbeck (changes requested):
Hi @Dexterp37 and welcome to BCD! This looks really good (and thanks to @rpl for the prior review). I have a couple of convention/style suggestions in line comments, but nothing major. Thank you!

webextensions/api/management.json
Dexterp37: Totally makes sense! Thanks for the suggestion!

webextensions/api/management.json
Dexterp37: > 
> 
> suggestion: BCD accepts an array of notes, when the notes refer to different issues. 

TIL!

> (OK, I also did some wordsmithing and markup changes. I'm a technical writer and I cannot help myself. 😄)
> 
> This kind of change would need to be applied to the other below note as well.

This is great! Thanks for this, I'm pushing an update now :)

ddbeck (approved):
Looks great, @Dexterp37! Thank you! 🎉

PR #7001, opened 2020-10-21
margin, border, padding, inset in Chrome

rachelandrew:
Working on https://github.com/mdn/sprints/issues/3793

Chrome are shipping the logical margin, border, padding shorthands along with the inset property and shorthands in 87.

This PR updates the BCD.

https://www.chromestatus.com/feature/6618771051511808

### DIFF
--- a/css/properties/border-block-color.json
+++ b/css/properties/border-block-color.json
@@ -54,16 +54,21 @@
             "ie": {
               "version_added": false
             },
-            "opera": {
-              "version_added": "56",
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "Experimental Web Platform Features",
-                  "value_to_set": "enabled"
-                }
-              ]
-            },
+            "opera": [
+              {
+                "version_added": "73"
+              },
+              {
+                "version_added": "56",
+                "flags": [
+                  {
+                    "type": "preference",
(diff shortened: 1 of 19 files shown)


ddbeck (commented):
Hey @rachelandrew, the changes already in the PR look great. There's a couple of optional things we could do here to greatly improve that data though, if you're willing. I'd be very appreciative!

First, the Chrome status entry says that this shipped in WebView 87 as well. WebView doesn't support prefs, so only the unpreffed 87 would need to be reflected in the WebView data.

I saw that you corrected some Edge data. This made me notice that the other Chromium derived browsers have similar issues (they were probably mirrored from each other). If you'd like to fix up those too that would be a big help. The mirroring script may be of use here (e.g., `npm run mirror -- --modify=always opera css.properties.border-block` syncs Opera to the Chrome data), though it's not flawless (it may mirror flags to browsers that don't support flags).

If you're not interested in this, that's OK too. I can open an issue upon merging. Thank you!

rachelandrew:
@ddbeck hopefully the commit today fixed both of those up, let me know if there is anything else.

ddbeck (approved):
Great. Thank you! 🎉

ddbeck:
Oh, I may have got head of myself just a little. I opened a follow-up: #7355.

PR #7003, opened 2020-10-21
autocapitalize attribute

rachelandrew:
Adding support for the HTML global `autocapitalize` attribute:

- Firefox is shipping in 83: https://bugzilla.mozilla.org/show_bug.cgi?id=1425291
- Safari has had since 4/iOS5: https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariHTMLRef/Articles/Attributes.html#//apple_ref/doc/uid/TP40008058-autocapitalize
- Chrome since 43: https://www.chromestatus.com/feature/4529989986811904

MDN page: https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autocapitalize

Part of: https://github.com/mdn/sprints/issues/3802

### DIFF
--- a/html/global_attributes.json
+++ b/html/global_attributes.json
@@ -63,10 +63,24 @@
               "version_added": "79"
             },
             "firefox": {
-              "version_added": "83"
+              "version_added": "83",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "dom.forms.autocapitalize",
+                  "value_to_set": "true"
+                }
+              ]
             },
             "firefox_android": {
-              "version_added": "83"
+              "version_added": "83",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "dom.forms.autocapitalize",
+                  "value_to_set": "true"
+                }
+              ]
             },
             "ie": {
               "version_added": null
--- a/html/global_attributes.json
+++ b/html/global_attributes.json
@@ -54,19 +54,19 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/HTML/Global_attributes/autocapitalize",
           "support": {
             "chrome": {
-              "version_added": false
+              "version_added": "43"
             },
             "chrome_android": {
-              "version_added": "66"
+              "version_added": "43"
             },
             "edge": {
-              "version_added": null
+              "version_added": "79"
             },
             "firefox": {
-              "version_added": false
+              "version_added": "83"
             },
             "firefox_android": {
-              "version_added": false
+              "version_added": "83"
             },
             "ie": {
               "version_added": null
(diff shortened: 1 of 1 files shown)


makotokato:
> Firefox is shipping in 83: https://bugzilla.mozilla.org/show_bug.cgi?id=142529

No, this is Nightly build only, so we need preference (`dom.forms.autocapitalize`) for beta and release.

html/global_attributes.json
```diff
@@ -81,7 +81,7 @@
               "version_added": null
             },
             "safari": {
-              "version_added": false
+              "version_added": "4"
```
makotokato, line 84: Do you test this?  My understand is Apple disable this by build flag.

html/global_attributes.json
```diff
@@ -54,19 +54,19 @@
...
-              "version_added": null
+              "version_added": "79"
             },
             "firefox": {
-              "version_added": false
+              "version_added": "83"
             },
             "firefox_android": {
-              "version_added": false
+              "version_added": "83"
```
makotokato, line 69: I don't ship this to release channel. Add `dom.forms.autocapitalize` prefs.

html/global_attributes.json
```diff
@@ -54,19 +54,19 @@
...
-              "version_added": "66"
+              "version_added": "43"
             },
             "edge": {
-              "version_added": null
+              "version_added": "79"
             },
             "firefox": {
-              "version_added": false
+              "version_added": "83"
```
makotokato, line 66: I don't ship this to release channel. Add `dom.forms.autocapitalize` prefs.

html/global_attributes.json
rachelandrew: It appears to work in current iOS, and the information was taken from here as to initial support. https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariHTMLRef/Articles/Attributes.html#//apple_ref/doc/uid/TP40008058-autocapitalize

Do you have better information? If so please point me to it and I'll take a look - thanks!

makotokato: > It appears to work in current iOS, and the information was taken from here as to initial support. https://developer.apple.com/library/archive/documentation/AppleApplications/Reference/SafariHTMLRef/Articles/Attributes.html#//apple_ref/doc/uid/TP40008058-autocapitalize
> 
> Do you have better information? If so please point me to it and I'll take a look - thanks!

This value is Safari for macOS, not iOS.

As long as I test this by wpt.live on Safari TP, macOS version isn't still supported. Also https://bugs.webkit.org/show_bug.cgi?id=180821 is for macOS support and source code is https://github.com/WebKit/webkit/blob/950143da027e80924b4bb86defa8a3f21fd3fb1e/Source/WTF/wtf/PlatformEnableCocoa.h#L86-L88, so it means macOS version is "not support". So you should keep "false" for macOS version.

rachelandrew: thanks! Updated.

ddbeck (approved):
Looks good. Thank you! 🎉

PR #6882, opened 2020-10-11
Set Chromium version for api.MediaDeviceInfo.toJSON

vinyldarkscratch:
This PR uses the mdn-bcd-collector to determine the version number for `api.MediaDeviceInfo.toJSON`.

PR #7005, opened 2020-10-21
Remove remaining polyfill notes from MediaDeviceInfo API for Opera

foolip:
These notes don't make sense when the toJSON method was never shipped.

Similar to https://github.com/mdn/browser-compat-data/pull/7002.


vinyldarkscratch (approved):
Looks good!  Should we merge #6882 first?

foolip:
Merged https://github.com/mdn/browser-compat-data/pull/6882 and resolved conflicts, but the commit message of this doesn't make sense now. I'll edit it when merging.

PR #7010, opened 2020-10-22
updating Firefox support for private class fields

chrisdavidmills:
As per https://bugzilla.mozilla.org/show_bug.cgi?id=1435826 and https://bugzilla.mozilla.org/show_bug.cgi?id=1659134, private class fields are now supported in Firefox 81.

### DIFF
--- a/javascript/classes.json
+++ b/javascript/classes.json
@@ -504,7 +504,19 @@
               ]
             },
             "firefox_android": {
-              "version_added": "81"
+              "version_added": "81",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "javascript.options.experimental.private_fields",
+                  "value_to_set": "true"
+                },
+                {
+                  "type": "preference",
+                  "name": "javascript.options.experimental.private_methods",
+                  "value_to_set": "true"
+                }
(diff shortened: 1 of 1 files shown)


littledan:
Is this actually unflagged? This doesn't work for me in Firefox 81.

chrisdavidmills:
> Is this actually unflagged? This doesn't work for me in Firefox 81.

Thanks for catching that @littledan ! I'd forgotten to add the pref information. I've added it in.

javascript/classes.json
```diff
@@ -489,10 +489,22 @@
...
+                {
+                  "type": "preference",
+                  "name": "javascript.options.experimental.private_methods",
+                  "value_to_set": "true"
+                }
+              ]
             },
             "firefox_android": {
-              "version_added": false
+              "version_added": "81"
```
ddbeck, line 507: @chrisdavidmills I'm guessing this is also behind a preference on Android?

chrisdavidmills: Yup, added!

ddbeck (approved):
Thank you, Chris! 🎉

Issue #7009, opened 2020-10-22
focus-visible landed in Chrome 86


jakub-g:
Source: https://www.chromestatus.com/feature/5823526732824576

Relevant file:
https://github.com/mdn/browser-compat-data/blob/master/css/selectors/focus-visible.json#L10-L27

(I don't know how to edit the JSON to tell "67 with flag, 86 without flag)

sideshowbarker:
> (I don't know how to edit the JSON to tell "67 with flag, 86 without flag)

Yeah, I guess the JSON syntax used for that’s a bit unintuitive/quirky:

```json
"chrome": [
  {
    "version_added": "86"
  },
  {
    "version_added": "67",
    "flags": [
      {
        "name": "#enable-experimental-web-platform-features",
        "type": "preference",
        "value_to_set": "enabled"
      }
    ]
  }
]
```
…that is, you need to change the object into an array of objects.

But at https://github.com/mdn/browser-compat-data/blob/master/schemas/compat-data-schema.md#the-support_statement-object it is actually documented with a good example of exactly that.

Anyway, now you know for next time :) For this I went ahead and raised #7011.

sideshowbarker:
Quick work! — change merged to the main branch just now. Thanks much for raising this @jakub-g

jakub-g:
Thx folks, that was fast! ⚡

PR #7011, opened 2020-10-22
Mark CSS focus-visible with enabled-by-default support in Chrome (86+)

sideshowbarker:
This change marks the CSS `:focus-visible` pseudo-class as enabled-by-default supported in Chrome (+Android, +WebView) 86+, per https://www.chromestatus.com/feature/5823526732824576

Fixes https://github.com/mdn/browser-compat-data/issues/7009

### DIFF
--- a/css/selectors/focus-visible.json
+++ b/css/selectors/focus-visible.json
@@ -6,26 +6,36 @@
           "description": "<code>:focus-visible</code>",
           "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/:focus-visible",
           "support": {
-            "chrome": {
-              "version_added": "67",
-              "flags": [
-                {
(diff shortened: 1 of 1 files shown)


css/selectors/focus-visible.json
```diff
@@ -6,26 +6,36 @@
...
-                {
-                  "name": "#enable-experimental-web-platform-features",
-                  "type": "preference",
-                  "value_to_set": "enabled"
-                }
-              ]
-            },
+            "chrome": [
+              {
+                "version_added": "86"
```
foolip, line 11: Confirmed by https://storage.googleapis.com/chromium-find-releases-static/edb.html#edbfef569f251aab76cb9c4774be82dbd4879d74

foolip (approved):

PR #7013, opened 2020-10-22
Fixing private class fields compat table data of safari on mac and iOS

### LINKED ISSUE #6773: Private class fields compatibility table is not accurate
The problem is that https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/Private_class_fields is showing that Safari 14 supports private class fields, but this is not very accurate. Private class fields were added behind a runtime flag on Safari 14, and this flag is disabled by default. This can cause some confusion because it is not well clarified on article and majority of Safari users won't enable such flag.

It's possible to verify it by running the following program on Safari 14:

```
class C {
  #field ;
}

console.log("test");
```

The result is going to be a `SyntaxError` due to unexpected token `#`.

caiolima:
It fixes #6773.

### DIFF
--- a/javascript/classes.json
+++ b/javascript/classes.json
@@ -507,10 +507,10 @@
               "version_added": "53"
             },
             "safari": {
-              "version_added": "14"
+              "version_added": false
             },
             "safari_ios": {
-              "version_added": "14"
+              "version_added": false
             },
             "samsunginternet_android": {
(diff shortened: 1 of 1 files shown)


sideshowbarker (approved):

sideshowbarker:

PR #7016, opened 2020-10-22
CSS "content-visibility" property

ramiy:
- [x] Summarize your changes

New CSS property.

**Required for:**

https://developer.mozilla.org/en-US/docs/Web/CSS/content-visibility

**Resources:**

https://www.chromestatus.com/feature/4613920211861504

https://developers.google.com/web/updates/2020/08/nic85#content-visibility

https://web.dev/content-visibility/

https://caniuse.com/css-content-visibility

### DIFF
--- a/css/properties/content-visibility.json
+++ b/css/properties/content-visibility.json
@@ -27,7 +27,7 @@
               "version_added": "71"
             },
             "opera_android": {
-              "version_added": false
+              "version_added": "60"
             },
             "safari": {
               "version_added": false
--- a/css/properties/content-visibility.json
+++ b/css/properties/content-visibility.json
@@ -43,7 +43,7 @@
             }
           },
           "status": {
-            "experimental": false,
+            "experimental": true,
             "standard_track": true,
             "deprecated": false
           }
--- a/css/properties/content-visibility.json
+++ b/css/properties/content-visibility.json
@@ -39,7 +39,7 @@
               "version_added": false
             },
             "webview_android": {
-              "version_added": false
+              "version_added": "85"
             }
           },
           "status": {
--- a/css/properties/content-visibility.json
+++ b/css/properties/content-visibility.json
@@ -1,0 +1,54 @@
+{
+  "css": {
+    "properties": {
+      "inset": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/content-visibility",
+          "support": {
+            "chrome": {
+              "version_added": "85"
+            },
+            "chrome_android": {
+              "version_added": "85"
+            },
+            "edge": {
+              "version_added": "85"
+            },
+            "firefox": {
+              "version_added": false
+            },
+            "firefox_android": {
+              "version_added": false
+            },
+            "ie": {
+              "version_added": false
+            },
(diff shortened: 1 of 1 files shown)


css/properties/content-visibility.json
```diff
@@ -0,0 +1,54 @@
...
+              "version_added": false
+            },
+            "ie": {
+              "version_added": false
+            },
+            "opera": {
+              "version_added": "71"
+            },
+            "opera_android": {
+              "version_added": false
```
ddbeck, line 30: question: How'd you decide to set the value for Opera for Andorid?

It seems notable to me that this would be in all the other Chromiums @ 85, but not in Opera for Android 60. We should probably note an explanation on this PR, if it's correct.

css/properties/content-visibility.json
```diff
@@ -0,0 +1,54 @@
...
+            },
+            "safari_ios": {
+              "version_added": false
+            },
+            "samsunginternet_android": {
+              "version_added": false
+            },
+            "webview_android": {
+              "version_added": false
```
ddbeck, line 42: suggestion: the Chrome status entry says WebView shipped this in 85 too

```suggestion
              "version_added": "85"
```

css/properties/content-visibility.json
```diff
@@ -0,0 +1,54 @@
...
+            }
+          },
+          "status": {
+            "experimental": false,
```
ddbeck, line 46: There's only one implementation for this so far, so this needs to be `true`.

```suggestion
            "experimental": true,
```

ddbeck (changes requested):
Thanks for this PR, @ramiy. A few comments and suggestions in line comments, but otherwise looks good.

(Sorry for the noise of individual comments. Not caffeinated enough today, apparently.)

css/properties/content-visibility.json
ramiy: @ddbeck check https://caniuse.com/css-content-visibility

css/properties/content-visibility.json
ramiy: **Opera 71** (released on Sep 15, 2020) and **Chrome 85** (released on Aug 25, 2020).

css/properties/content-visibility.json
ddbeck: The caniuse table says that Opera for Android 59 doesn't support the property. But if Opera for Android did support the feature, then I would have expected it to first appear in Opera for Android 60 (which, [by browser our data](https://github.com/mdn/browser-compat-data/blob/master/browsers/opera_android.json#L330), uses the same engine as Chrome 85). If you don't want to test it, then I think we should assume that it works rather than assuming that it doesn't.

```suggestion
              "version_added": "60"
```

css/properties/content-visibility.json
ddbeck: Opera and Chrome are both Chromium/Blink, so that's only one engine implementing the feature. I stand by my original suggestion.

ddbeck (changes requested):

ddbeck (approved):
Thank you, @ramiy! 🎉

Issue #6620, opened 2020-08-31
Update api.Body.formData for Safari 14.1


foolip:
A tracking bug for a few things that will need updating for future releases of Safari and Safari for iOS. This might be guesses based on what's currently in Safari Technology Preview, or recent commits to WebKit.

Safari 14.1:
- [ ] api.Body.formData (https://github.com/mdn/browser-compat-data/pull/6616, https://github.com/mdn/browser-compat-data/pull/6619, https://webkit.org/b/215671, https://trac.webkit.org/r266087)

@apple-web-evangelist FYI. If there's anything that could be done to make it easier to update Safari data for each release, that would be great :D

foolip:
This idea of a tracking issue didn't work out.

foolip:
However, there was one TODO which is still in the future, so renaming and reopening to track just that one thing.

jculver050:
Close

Issue #6601, opened 2020-08-30
TODO for future Opera releases


foolip:
A tracking bug for a few things that will need updating in the future for Opera and Opera for Android, based on what is in Chromium.

Chromium M84:
  - [ ] [ContentIndex and ContentIndexEvent](https://github.com/mdn/browser-compat-data/pull/6547) (Android only)
  - [ ] [Element.getAnimations()](https://github.com/mdn/browser-compat-data/pull/6592)

foolip:
@ddbeck FYI that I've filed this and https://github.com/mdn/browser-compat-data/issues/6602. Can you keep an eye out for updates in browsers/opera.json? We can try it once and if it works, maybe formalize it all with labels and whatnot?

foolip:
I closed https://github.com/mdn/browser-compat-data/issues/6602 as not useful. I guess this too wouldn't be useful if someone updates all Opera data for each release. I guess nobody is doing that?

ddbeck:
Hey @foolip, I opened #6561 which sorta covers how we might do this. How do you feel about scoping this to a specific release (M84)? I kinda like the idea of making sure our issues have some sort of criteria to help us know when to close them, rather than letting them live on perpetuity.

foolip:
Sure, I'd be happy for you to rename/reformat this issue, or close it and create a new one.

foolip:
I didn't keep updating this issue, and I think the approach is completely unworkable, but nevertheless I've sent https://github.com/mdn/browser-compat-data/pull/7300 to mirror the two things I did note to close this issue.

ddbeck:
@foolip Yeah, I agree this isn't really working. https://github.com/mdn/browser-compat-data/pull/7020#issuecomment-722739785 also suggests this is not going to happen properly. I think your diff tool might be a better route for us, actually. I'll update in #6561.

In the mean time, we should not continue to expand on this issue.

Issue #6561, opened 2020-08-25
Tracking issues for to-be-mirrored features


ddbeck:
When a feature lands in one browser engine, it often lands in browsers that use that engine at various times. While we can accommodate future releases in BCD's schema (with beta and nightly annotations), we don't have a way of tracking an _unversioned_ next release. To help deal with this better, we should document a process to track features which are impacted by this.

## Process proposal

I am sure I can write better instructions for a docs PR, but here's my imagined process for your consideration.

To borrow from #6551, suppose a feature is landing in Chromium M86. It'll show up in Chrome 86, Chrome for Android 86, and WebView 86. It will also show up in Opera and Samsung Internet in some future version, but we don't have a version numbers for those browsers yet. We need to make a log of features affected by this. Here's what we should to do keep and clear that log:

1. When a new engine version is first added to BCD, open a new issue with a title of the form "Mirror _Project_ _Number_ features", where _Project_ is the name of the originating project and _Number_ is the version number we'd mirror from. For example, when the PR is opened to add Chrome 86, then at the same time an issue should be opened called `Mirror Chromium 86 features`.

   The issue itself needs two key things:

   - an `## Applicable features` section for a checklist of dotted feature identifiers (e.g., `api.SomeAPI.someMethod`)
   - an assignee or a call for an assignee (i.e., someone who's committing to following up when browsers get applicable releases)

2. For each version 86 feature added or modified, add the feature to the list of dotted feature identifiers and annotate with a link to the PR(s) that created those changes.

3. When a browser has a corresponding release (e.g., Opera has a release which matches or subsumes Chromium 86), the assignee opens a PR that runs the mirroring script for the features listed in the issue. Reviewers or PR authors would be advised to update the tracking issue when a `/browsers` PR is merged (we could change the PR template to note this or add perhaps add some GitHub automation to remind us).

4. This process continues until there are no more browsers with unrecorded Chromium 86 releases. When all of the features in the list have been mirrored to all of the browsers, the issue is closed.

---

I'm sure I'm missing something here or there are some details we can bikeshed ("Mirror Blink 86 features"?). Let me know what you think!

sideshowbarker:
Thanks for writing this up. The proposal here all sounds reasonable to me — and I have no improvements to suggest. So 👍

jpmedley:
Sounds good.

foolip:
[...]

PR #7020, opened 2020-10-23
Add Intl.toString

hamishwillee:
This adds BCD info for `Intl.prototype[@@toStringTag]` which was added in FF83. See bug report https://bugzilla.mozilla.org/show_bug.cgi?id=1670053

Docs are here: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/@@toStringTag

I tried this out on other browsers including latest chrome and opera and it works. However not sure how to record that because when I tried setting values to `true` the test tells me that the I'm not allowed. Similarly `null`. I suspect I could add specific versions, but I don't know when this went in.

### DIFF
--- a/javascript/builtins/intl/Intl.json
+++ b/javascript/builtins/intl/Intl.json
@@ -104,6 +104,58 @@
               "deprecated": false
             }
           }
+        },
+        "@@toStringTag": {
+          "__compat": {
+            "mdn_url": "https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Intl/@@toStringTag",
+            "spec_url": "https://tc39.es/ecma402/#sec-Intl-toStringTag",
+            "support": {
+              "chrome": {
+                "version_added": "86"
+              },
+              "chrome_android": {
+                "version_added": "86"
+              },
+              "edge": {
+                "version_added": "86"
+              },
+              "firefox": {
+                "version_added": "83"
+              },
+              "firefox_android": {
+                "version_added": "83"
+              },
+              "ie": {
+                "version_added": false
+              },
+              "nodejs": {
+                "version_added": "15.0.0"
+              },
+              "opera": {
+                "version_added": "72"
+              },
+              "opera_android": {
+                "version_added": false
+              },
+              "safari": {
+                "version_added": false
+              },
+              "safari_ios": {
+                "version_added": false
+              },
+              "samsunginternet_android": {
+                "version_added": false
+              },
+              "webview_android": {
+                "version_added": "86"
+              }
+            },
+            "status": {
+              "experimental": false,
+              "standard_track": true,
+              "deprecated": false
+            }
+          }
         }
       }
     }


javascript/builtins/intl/Intl.json
```diff
@@ -104,6 +104,58 @@
...
+                "version_added": false
+              },
+              "opera": {
+                "version_added": false
+              },
+              "opera_android": {
+                "version_added": false
+              },
+              "safari": {
+                "version_added": false
```
Constellation, line 141: Safari 14 supports it
https://trac.webkit.org/changeset/262568/webkit

javascript/builtins/intl/Intl.json
```diff
@@ -104,6 +104,58 @@
...
+              },
+              "safari_ios": {
+                "version_added": false
```
Constellation, line 144: Ditto.

javascript/builtins/intl/Intl.json
Constellation: Ah, no. `Intl[@@toStringTag]` is https://trac.webkit.org/changeset/266015/webkit, and it is not included in Safari 14 yet.

hamishwillee:
@Constellation So what needs to happen for this to merge/progress?

@chrisdavidmills FYI, I tested this on Opera and the interactive example worked. But when I tried to set supported as `true` I got the following error:
image
Ditto for setting the ones I don't know to `null`.

chrisdavidmills:
> FYI, I tested this on Opera and the interactive example worked. But when I tried to set supported as `true` I got the following error ; Ditto for setting the ones I don't know to null.

Hrm, I'm not sure why that is. @ddbeck, does JS BCD not longer accept `true` or `null` values? Sounds like possibly some weirdness in @hamishwillee 's local testing, as the CI run has passed for this one.

ddbeck:
@chrisdavidmills @hamishwillee That is weird, but I can't explain why it's passing CI. We do require real values in JavaScript though, so we _should be_ getting an error in CI: https://github.com/mdn/browser-compat-data/blob/master/test/linter/test-real-values.js#L13

hamishwillee:
> That is weird, but I can't explain why it's passing CI. We do require real values in JavaScript though, so we should be getting an error in CI: https://github.com/mdn/browser-compat-data/blob/master/test/linter/test-real-values.js#L13

@chrisdavidmills @ddbeck It passes CI because I've set the values to `false` as "the only passing choice I had". That means "not supported". I would like to set to `true` (supported but I don't know the version) and `null` (I don't know if it is supported). If I changed to null or false then I'd get the error.

Could be wrong, but 
- that bit you're pointing at is the name string for the browser not version_added. The relevant code for error is here: https://github.com/mdn/browser-compat-data/blob/master/test/linter/test-real-values.js#L60-L79
- I interpret that as saying that once you've set that a particular browser is supported you can't then go saying that you don't know whether a particular feature is supported. I would guess the theory is that you should in this case have partial support in the parent level. That seems both strictly correct and daft - because it means that if you add a new feature to an API then all the parents that say "supported" would now become partial support - lots of churn.

This check was added October 4 in https://github.com/mdn/browser-compat-data/pull/4122.

chrisdavidmills:
@foolip @jpmedley Hi folks, could you help @hamishwillee figure out Chromium support for this? Thanks ;-)

gilmoreorless:
Looks like `Intl[@@toStringTag]` was added in V8 8.6.

* [Issue tracker](https://bugs.chromium.org/p/v8/issues/detail?id=10744)
* [Version 8.6 commit list](https://chromium.googlesource.com/v8/v8.git/+log/branch-heads/8.6/?s=f94e772441dd7a6644dd0a54b22e2b5fe2b6090d)

That should mean it's in Chrome 86 and equivalents. Also Node.js 15.0.0 (which I've checked manually).

foolip:
I've used https://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8636 to confirm Chrome 86 is when this changed.

But I do wonder why this behavior matters to web developers, why does it matter what `Intl.toString()` will return?

hamishwillee:
@foolip Thanks. My understanding is that we document everything and let developers work out what interests them. @gilmoreorless Thanks!

@chrisdavidmills @ddbeck 
- I was able to to track down that opera 72 corresponds to Chrome 86, and I presume Chrome 86 == Chromium 86 for this purpose. Is there a table mapping chromium builds to the browsers/libraries that use them? I.e. how do I find out what version edge, webview, opera for android etc, uses of chromium? 
- @ddbeck This update fails locally due to the node value of 15.0.0 - however that is listed as a valid version. Thoughts?
  image

gilmoreorless:
>  This update fails locally due to the node value of 15.0.0 - however that is listed as a valid version. Thoughts?

It looks like you unfortunately created your branch in the small time window where the data for nodejs releases had a typo. It was fixed in #7018, so rebasing your branch off the latest `master` _should_ fix that error.

chrisdavidmills:
>     * I was able to to track down that opera 72 corresponds to Chrome 86, and I presume Chrome 86 == Chromium 86 for this purpose. Is there a table mapping chromium builds to the browsers/libraries that use them? I.e. how do I find out what version edge, webview, opera for android etc, uses of chromium?

@ddbeck or @foolip would be better placed to answer that. Do we have anything like this? If not, we really should have; I spend ages looking up which version of Chrome corresponds to Chromium a, Chromium b, etc., when I do BCD work ;-)

Constellation:
@hamishwillee Safari 14.2 includes this support. So the version is 14.2.

hamishwillee:
Thanks @Constellation - will this also be on the iOS Safari version 14.2? 
@ddbeck We currently only have [Safari 14](https://github.com/mdn/browser-compat-data/blob/master/browsers/safari.json#L170) (not 14.2). When/who does updates of this JSON? I plan to just leave as false for now.

hamishwillee:
>> I was able to to track down that opera 72 corresponds to Chrome 86, and I presume Chrome 86 == Chromium 86 for this purpose. Is there a table mapping chromium builds to the browsers/libraries that use them? I.e. how do I find out what version edge, webview, opera for android etc, uses of chromium?

> @ddbeck or @foolip would be better placed to answer that. Do we have anything like this? If not, we really should have; I spend ages looking up which version of Chrome corresponds to Chromium a, Chromium b, etc., when I do BCD work ;-)

What Chris said. 

We can't simultaneously require that accurate values for versions be listed without also making it possible to work out what those values might be. For this case by a table, but alternatively by a test suite or similar.

ddbeck:
@hamishwillee @chrisdavidmills BCD itself has largely complete data about which browsers correspond to each other: take a peek at the files in [the `browsers` directory](https://github.com/mdn/browser-compat-data/tree/master/browsers). You'll find that most entries have an `"engine"` and `"engine_version"` field. So you can check that Chrome X is Blink X is Opera Y and so on. This can also be done, for example, between historic versions of Safari (by comparing WebKit version numbers).

I'll grant that reading plain JSON isn't particularly ergonomic. I've started on a little web tool that makes these comparisons easier, but I haven't had time to finish it or share it. If it's a major hurdle, I can prioritize it, but it so far hasn't been a big enough hurdle for me alone to justify it.

For Chromium, it's typical that are not gated to specific browsers, though (rarely) there are discrepancies between desktop and Android releases. But in most cases, knowing one version means that the data in `browsers` can complete the other Chromium-derivatives.

To be clear, I don't expect perfection on these things, but I do expect a decent guess based on the data already in BCD, or an explanation when diverging from it. If the data is incomplete or doesn't give you a clue, then please flag that for my attention. There's a wealth of knowledge stuck in the heads of the people who work on this project; you don't have to go it alone.

Constellation:
@hamishwillee Yes, Safari 14.2 iOS
Can you put Safari 14.2 entry to browsers? Since we know that this is supported, `false` is incorrect data.

hamishwillee:
@Constellation Sorry, no can do. But we can update once @ddbeck et al have added it.

hamishwillee:
@ddbeck This should now be ready to merge - it is up to date for current browser versions (as stored in https://github.com/mdn/browser-compat-data/tree/master/browsers).

Browsers like Opera Android and Safari have not got a corresponding release version yet in BCD so I can't update those. How do we track (make sure we don't forget) those? Add a separate issue as a reminder?

PS Thanks for your info above. I now appreciate why you see this as relatively trivial - hadn't realised that the browser JSON had this level of info. Provided you know a blink, webkit, v8 version, gecko version, you have everything. I might add a PR "how to work out versions" note to the docs at some point soon.

foolip:
I see that nobody has mentioned the mirror script yet. For features that one can reasonably assume will be the same across all browser and platforms using a particular engine, one can get away with updating just the `chrome`, `firefox` and `safari` data and using `npm run mirror` to update everything else. It does take some work to figure out how to invoke `npm run mirror` in the right way, but it saves time after the first few times.

Constellation:
@hamishwillee So, can you put `null` instead of `false`? `null` means "not sure". `false` means "not supported".

hamishwillee:
> So, can you put null instead of false? null means "not sure". false means "not supported".

@Constellation Again, no. Not because I don't want to but because the toolchain no longer allows `null`. In fact `false` is correct for the listed browsers. We just need someone to add those new ones. I'm sure that it will be done, it is just not something that I consider I *should* do as I am new to this. There is nothing stopping you creating a PR for a new version if you wish.

Constellation:
@hamishwillee OK, I see. Added in https://github.com/mdn/browser-compat-data/pull/7248

ddbeck:
OK, a few things to cover here. First the meta stuff and then I'll actually review this PR.

## Docs and matching up browsers

> Thanks for your info above. I now appreciate why you see this as relatively trivial - hadn't realised that the browser JSON had this level of info. Provided you know a blink, webkit, v8 version, gecko version, you have everything. I might add a PR "how to work out versions" note to the docs at some point soon.

Thank you for your patience with me on this, @hamishwillee! It took me a while to understand where you were coming from and what the sources of confusion were. It's absolutely not trivial to get your head around BCD and I shouldn't have suggested so.

I'd welcome a PR with more docs! If you prefer, you can also open an issue and log any first impressions, initial hurdles, half-baked ideas, and such. It's been so long since I was new to any of this that I have a hard time figuring out what new contributors need. Anything you have to share for this would be greatly appreciated.

## Upcoming releases

> Browsers like Opera Android and Safari have not got a corresponding release version yet in BCD so I can't update those. How do we track (make sure we don't forget) those? Add a separate issue as a reminder?

I think an issue is the way to go. #6601 and #6620 are existing tracking issues, or you can open new issues and link between them. This is an area that haven't got a good process about yet. So if you have any commentary on the process, that'd be interesting too.

@Constellation I'm going to review this PR (and probably merge it) shortly. A quick look at #7248 leaves me with questions, so I don't want to block this on that. If you'd like to open a follow-up PR with your suggestions, that would probably be best.

ddbeck (approved):
Looks good, merging. Thank you! 🎉

hamishwillee:
Thanks @ddbeck. FYI only

> Regarding a process for managing features in upcoming releases https://github.com/mdn/browser-compat-data/pull/7020#issuecomment-721044583.

The approach used in  [TODO for future Opera releases](https://github.com/mdn/browser-compat-data/issues/6601) is a bit messy as it already has a "conversation". For this approach to work it would have to be formalised:
- Issue for each browser
- Description: checklist of open items sorted by probable release. Each item links to earlier items if known and other relevant background. Also trigger for it to come active if not obvious from release.
- Don't allow comments - these make a "forever lasting" issue like this hard to read/use
- Periodically sweep out the completed items so it is only open actions.
- Someone has to sweep this regularly to check if ready to do.

Alternative is what I did - just create open items either by API to update, or by item and browser. I created [TODO Intl.toString for SamsungAd 14+, OperaAd 16+, Safari14.2+](https://github.com/mdn/browser-compat-data/issues/7294) (i.e. item API to update and list. 

What I like about this is that each item is specific. It could be assigned to one person, title includes relevant item and browser version at which I should look at it.  This is my preference. The downside is that you end up with lots of these.

hamishwillee:
> Re "Docs and matching up browsers": 

Almost everything about the process is covered in https://github.com/mdn/browser-compat-data/blob/master/docs/contributing.md and its linked doc https://wiki.developer.mozilla.org/en-US/docs/MDN/Contribute/Processes/Matching_features_to_browser_version. There is just a LOT to take in, and even now I know this, I still think the job is by its nature not easy. 

I think perhaps that it might be worth moving  https://wiki.developer.mozilla.org/en-US/docs/MDN/Contribute/Processes/Matching_features_to_browser_version into the readme alongside "contributing". Then updating that doc with an overview/summary which notes:
- Many browsers share the same engine, so provided you can find a release for one you can take a reasonable guess that it will be in all (the exceptions being listed in the matching features doc above.)
- https://github.com/mdn/browser-compat-data/blob/master/browsers/ has information about the versions for all release browsers along with their engine. 
- Maybe a summary list of the best resource finding info with link down docs for more detail.

ddbeck:
@hamishwillee 

Thanks for the assessment of the docs situation. I opened #7304 based on your comment. 👍

> The approach used in [TODO for future Opera releases](https://github.com/mdn/browser-compat-data/issues/6601) is a bit messy as it already has a "conversation".

You're right. We've got #6561 open, but that's a mess too. I've commented there that we've got to find a better (likely, more automated) approach to that particular problem. In the mean time, I would suggest opening individual issues, as you've done.

PR #7025, opened 2020-10-23
adding support for displaystyle

rachelandrew:
I am documenting the MathML `displaystyle` attribute, which is being added to all MathML elements in the latest spec and implemented in Firefox 83.

https://bugzilla.mozilla.org/show_bug.cgi?id=1666075

This PR adds the support data for the attribute.


ddbeck (approved):
Thank you! 🎉

Issue #7037, opened 2020-10-25
revert - Browser support update


hakantr:
[revert](https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/revert)

Hi, the browser support needs to be updated for the `revert` value.

sideshowbarker:
Thanks much for reporting this. I’ve raised https://github.com/mdn/browser-compat-data/pull/7038 with an update for the data.

karelbilek:
Still shows as unsupported in MDN one month later. Is there some other issue?

karelbilek:
Oh. I made an empty newline edit and it re-rendered the compat table.

Strange. Anyway it helped.

sideshowbarker:
> Oh. I made an empty newline edit and it re-rendered the compat table.

Yeah, sorry the details about that aren’t super-discoverable, but do have them documented at https://github.com/mdn/browser-compat-data/blob/master/docs/contributing.md#updating-compatibility-tables-on-mdn.

As explained there, if you’re logged into your MDN account and do a force-reload when viewing any article, that’ll cause the CMS to regenerate the article contents — which includes pulling in any updated browser-compat data.

So in other words, you don’t actually even need to do this empty-edit thing — just a force-reload will do the trick.

karelbilek:
Oh OK.

Sorry about commenting on both the PR and the issue.

sideshowbarker:
> Sorry about commenting on both the PR and the issue.

 No worries. It was right for you to ask — it’s a good reminder that we should maybe try to find a way to make it more clear in general to all contributors and people who take time to raise issues

PR #7038, opened 2020-10-26
Mark CSS "revert" supported in Chrome

sideshowbarker:
This change marks the CSS `revert` keyword as supported in Chrome (+Android, +WebView) from version 84 (per https://www.chromestatus.com/feature/5644990145363968), and in the corresponding Edge and Opera versions.

Fixes https://github.com/mdn/browser-compat-data/issues/7037

### DIFF
--- a/css/types/global_keywords.json
+++ b/css/types/global_keywords.json
@@ -163,16 +163,13 @@
             "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/revert",
             "support": {
               "chrome": {
-                "version_added": false,
-                "notes": "See <a href='https://crbug.com/579788'>bug 579788</a>."
+                "version_added": "84"
               },
               "chrome_android": {
-                "version_added": false,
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Looks good. Thank you! 🎉

karelbilek:
Why is MDN still showing revert as unsupported?

https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/revert

sideshowbarker:
> Why is MDN still showing revert as unsupported in Chrome?

Yeah, for the record here too, I’ll just paste in my same comment from https://github.com/mdn/browser-compat-data/issues/7037#issuecomment-736224373 —

Sorry the details about that aren’t super-discoverable, but we do have them documented at https://github.com/mdn/browser-compat-data/blob/master/docs/contributing.md#updating-compatibility-tables-on-mdn.

As explained there, if you’re logged into your MDN account and do a force-reload when viewing any article, that’ll cause the CMS to regenerate the article contents — which includes pulling in any updated browser-compat data.

PR #7040, opened 2020-10-26
Mark AudioContextOptions Safari-unsupported

sideshowbarker:
This change marks the `AudioContextOptions` dictionary unsupported in Safari; confirmed by manual testing that even with the prefixed `webkitAudioContext()` constructor, any `latencyHint` or `sampleRate` values given in the object argument to the constructor don’t end up affecting any properties for the object that gets created.

### DIFF
--- a/api/AudioContextOptions.json
+++ b/api/AudioContextOptions.json
@@ -29,10 +29,10 @@
             "version_added": null
           },
           "safari": {
-            "version_added": null
+            "version_added": false
           },
           "safari_ios": {
-            "version_added": null
+            "version_added": false
           },
           "samsunginternet_android": {
             "version_added": "8.0"
@@ -76,10 +76,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": false
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": false
             },
             "samsunginternet_android": {
               "version_added": "8.0"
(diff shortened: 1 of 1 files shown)


foolip:
Support was added in https://trac.webkit.org/changeset/264342/webkit, which was WebKit trunk 610.1.20. Safari 14 was 610.1.28, so are you entirely sure this didn't make it into Safari 14?

sideshowbarker:
> Support was added in https://trac.webkit.org/changeset/264342/webkit, which was WebKit trunk 610.1.20. Safari 14 was 610.1.28, so are you entirely sure this didn't make it into Safari 14?

I’ll re-test.

(And I’ll also try to figure out when Safari 14 was branched. (I am pretty sure it was some time before August 12 — because the WritableStream code was merged on that date but did not end up in Safari 14 — but https://trac.webkit.org/changeset/264342/webkit shows the AudioContext code went in on July 14 so it seems like that should have been early enough to make it into the Safari 14 branch.)

sideshowbarker:
OK yes https://trac.webkit.org/browser/webkit/branches/safari-610.1.28.0-branch/Source/WebCore/Modules/webaudio/AudioContextOptions.idl shows that the code is in fact in the Safari 14 release branch — but https://wpt.live/webaudio/the-audio-api/the-audiocontext-interface/audiocontextoptions.html, which passes in current Safari TP, fails in Safari 14

foolip:
OK, so I think the explanation is that AudioContext.idl and WebKitAudioContext.idl have been split in WebKit:
https://trac.webkit.org/changeset/263198/webkit

These changes were for the unprefixed AudioContext, and that didn't ship in Safari 14.

foolip (approved):

PR #7041, opened 2020-10-26
fix repeat for Firefox

rachelandrew:
The BCD for `grid-template-columns` and `grid-template-rows` still shows partial support for Firefox, due to multiple values (a tracklisting) not being used.

This was fixed in 76: https://bugzilla.mozilla.org/show_bug.cgi?id=1341507

### DIFF
--- a/css/properties/grid-template-columns.json
+++ b/css/properties/grid-template-columns.json
@@ -381,6 +381,9 @@
               },
               "firefox": [
                 {
+                  "version_added": "76"
+                },
+                {
                   "version_added": "57",
                   "version_removed": "76",
                   "notes": "<code>repeat(auto-fill, ...)</code> and <code>repeat(auto-fit, ...)</code> only support one repeated column (see <a href='https://bugzil.la/1341507'>bug 1341507</a>).",
--- a/css/properties/grid-template-columns.json
+++ b/css/properties/grid-template-columns.json
@@ -382,6 +382,7 @@
               "firefox": [
                 {
                   "version_added": "57",
+                  "version_removed": "76",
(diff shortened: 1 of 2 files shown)


css/properties/grid-template-columns.json
```diff
@@ -382,6 +382,7 @@
               "firefox": [
                 {
                   "version_added": "57",
+                  "version_removed": "76",
```
ddbeck, line 385: Hmm. This makes it appear as if support for `repeat()` has been dropped entirely. Unless I've misunderstood something about the bug you linked to, I think this PR is incomplete without an additional support statement above this one, with `"version_added: "76"`. I think the same goes for the other file as well.

ddbeck (changes requested):

css/properties/grid-template-columns.json
rachelandrew: I've got that listed against the section with the notes, I'm not sure otherwise how to do it, do you mean to add a full support section above?

```
{
  "version_added": "76",
 },
{
  "version_added": "57",
  "version_removed": "76",
   "notes": "<code>repeat(auto-fill, ...)</code> and <code>repeat(auto-fit, ...)</code> only support one repeated column (see <a href='https://bugzil.la/1341507'>bug 1341507</a>).",
 }
```

ddbeck: Yep, a full object above, just like that 👍

rachelandrew: @ddbeck I've made it so - thanks :)

ddbeck (approved):
Great. Thank you, @rachelandrew! 🎉

PR #7042, opened 2020-10-26
adds font-variant

rachelandrew:
Adding a page for `font-variant`: https://wiki.developer.mozilla.org/en-US/docs/Web/CSS/@font-face/font-variant

This adds the data which is missing among the other values as used in `@font-face`.

### DIFF
--- a/css/at-rules/font-face.json
+++ b/css/at-rules/font-face.json
@@ -479,6 +479,54 @@
             }
           }
         },
+        "font-variant": {
+          "__compat": {
+            "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/@font-face/font-variant",
+            "support": {
+              "chrome": {
+                "version_added": "4"
+              },
+              "chrome_android": {
+                "version_added": "18"
+              },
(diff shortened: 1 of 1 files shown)


ddbeck:
@rachelandrew since this is a shorthand, should we also have features for all of the constituent descriptors?

Actually, this is an area where I'm just not well-acquainted enough with the specifications. Is there a distinction to be made between `font-variant` as a property (and its constituents) and as a descriptor of `@font-face`?

rachelandrew:
> @rachelandrew since this is a shorthand, should we also have features for all of the constituent descriptors?
> 
> Actually, this is an area where I'm just not well-acquainted enough with the specifications. Is there a distinction to be made between `font-variant` as a property (and its constituents) and as a descriptor of `@font-face`?

yeah I think we would defer to the individual pages for the constituent properties. This is just regarding the fact it is usable for `@font-face`. Really for consistency as we have everything else already.

ddbeck (approved):
Thank you! 🎉

PR #7053, opened 2020-10-27
Set Chromium to 1 for HTMLEmbedElement

vinyldarkscratch:
This PR uses the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com/) project to set `true` to `1` for Chrome for the `HTMLEmbedElement` API and many of its subfeatures.  Data is also mirrored to Chrome Android, Samsung Internet, and WebView Android.


foolip (approved):
`<embed>` is a bit special like `<object>`, but I don't recall exactly what to look out for. In any case, if http://mdn-bcd-collector.appspot.com/tests/api/HTMLEmbedElement returns true then I'm confident it's supported, it's false negatives that seems more likely.

PR #7055, opened 2020-10-27
Set Chromium to 1 for HTMLFontElement

vinyldarkscratch:
This PR uses the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com/) project to set `true` to `1` for Chrome for the `HTMLFontElement` API and many of its subfeatures.  Data is also mirrored to Chrome Android, Samsung Internet, and WebView Android.


foolip (approved):
Updates the same entries as https://github.com/mdn/browser-compat-data/pull/7106, will merge together with that.

PR #7062, opened 2020-10-27
Set Chromium to 1 for HTMLHtmlElement

vinyldarkscratch:
This PR uses the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com/) project to set `true` to `1` for Chrome for the `HTMLHtmlElement` API and many of its subfeatures.  Data is also mirrored to Chrome Android, Samsung Internet, and WebView Android.

### DIFF
--- a/api/HTMLHtmlElement.json
+++ b/api/HTMLHtmlElement.json
@@ -5,10 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLHtmlElement",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "1"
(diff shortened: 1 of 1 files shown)


foolip (approved):
Goes with https://github.com/mdn/browser-compat-data/pull/7113

foolip:
@vinyldarkscratch note that the "many of its subfeatures" part of the description doesn't apply here. If you're able to generate the PR description, then listing the touched entries would be useful, although it overlaps with the PR preview tooling we've discussed so I wouldn't spend more than a few minutes on it if it's not trivial.

PR #7096, opened 2020-10-27
Set Opera to ≤12.1 for HTMLBRElement

vinyldarkscratch:
This PR uses the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com/) project to set `true` to `≤12.1` for Opera for the `HTMLBRElement` API and many of its subfeatures.  Data is also mirrored to Opera Android.

### DIFF
--- a/api/HTMLBRElement.json
+++ b/api/HTMLBRElement.json
@@ -23,10 +23,10 @@
             "version_added": true
           },
           "opera": {
-            "version_added": true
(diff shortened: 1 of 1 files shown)


foolip (approved):

foolip:
@vinyldarkscratch I see the chrome_android data is missing for the touched entries. It think they can be mirrored, just like opera_android.

PR #7098, opened 2020-10-27
Set Opera to ≤12.1 for HTMLBodyElement

vinyldarkscratch:
This PR uses the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com/) project to set `true` to `≤12.1` for Opera for the `HTMLBodyElement` API and many of its subfeatures.  Data is also mirrored to Opera Android.

### DIFF
--- a/api/HTMLBodyElement.json
+++ b/api/HTMLBodyElement.json
@@ -23,10 +23,10 @@
             "version_added": "4"
           },
           "opera": {
-            "version_added": true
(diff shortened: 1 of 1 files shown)


foolip (approved):

foolip:
@vinyldarkscratch I think it would be safe to mirror chrome_android for HTMLBodyElement too.

PR #7099, opened 2020-10-27
Set Opera to ≤12.1 for HTMLButtonElement

vinyldarkscratch:
This PR uses the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com/) project to set `true` to `≤12.1` for Opera for the `HTMLButtonElement` API and many of its subfeatures.  Data is also mirrored to Opera Android.

### DIFF
--- a/api/HTMLButtonElement.json
+++ b/api/HTMLButtonElement.json
@@ -23,10 +23,10 @@
             "version_added": true
           },
           "opera": {
-            "version_added": true
+            "version_added": "≤12.1"
           },
           "opera_android": {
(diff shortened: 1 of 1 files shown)


foolip (approved):

foolip:
@vinyldarkscratch while this matches https://github.com/mdn/browser-compat-data/pull/7050, it looks like there's still a fair bit of data missing for `HTMLButtonElement`. Updating all of it in a single PR would probably be the most effective way to review what's left here.

PR #7101, opened 2020-10-27
Set Opera to ≤12.1 for HTMLDataListElement

vinyldarkscratch:
This PR uses the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com/) project to set `true` to `≤12.1` for Opera for the `HTMLDataListElement` API and many of its subfeatures.  Data is also mirrored to Opera Android.

### DIFF
--- a/api/HTMLDataListElement.json
+++ b/api/HTMLDataListElement.json
@@ -23,10 +23,10 @@
             "version_added": false
           },
           "opera": {
-            "version_added": true
+            "version_added": "≤12.1"
           },
           "opera_android": {
-            "version_added": true
+            "version_added": "≤12.1"
           },
           "safari": {
             "version_added": false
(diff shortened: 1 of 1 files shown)


foolip (approved):

foolip:
@vinyldarkscratch I see there's no matching Chromium PR for this, and it seems likely to me that the Chrome versions (62) is wrong. Can you investigate?

vinyldarkscratch:
I plan to follow-up with more PRs to correct existing data, including Chrome data for `HTMLDataListElement`!  Figured it'd be better for review to keep non-real to `1` separated from version number changes.  :P

foolip:
@vinyldarkscratch indeed, keeping each PR to a single type of change like true/null &rarr; "1" is great. But grouping together changes to a single entry as much as possible is also helpful. For ease of review, what I'd suggest is to bundle together trivial changes across engines/browsers when all of the changes will be of the same type, like true/null &rarr; earliest version, and to separately fix entries that have messier existing data, even if that means holding back *some* of the trivial changes.

But if grouping changes like that is in itself a bunch of work, then let's just keep going like this :)

PR #7103, opened 2020-10-27
Set Opera to ≤12.1 for HTMLElement

vinyldarkscratch:
This PR uses the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com/) project to set `true` to `≤12.1` for Opera for the `HTMLElement` API and many of its subfeatures.  Data is also mirrored to Opera Android.

### DIFF
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -70,10 +70,10 @@
               "version_added": null
             },
             "opera": {
-              "version_added": true
+              "version_added": "≤12.1"
             },
             "opera_android": {
-              "version_added": true
+              "version_added": "≤12.1"
             },
             "safari": {
               "version_added": "6"
@@ -742,10 +742,10 @@
               "version_added": null
             },
             "opera": {
-              "version_added": true
+              "version_added": "≤12.1"
             },
             "opera_android": {
-              "version_added": true
+              "version_added": "≤12.1"
             },
             "safari": {
               "version_added": "6"
@@ -1452,10 +1452,10 @@
               "version_added": null
             },
             "opera": {
-              "version_added": true
+              "version_added": "≤12.1"
             },
             "opera_android": {
-              "version_added": true
+              "version_added": "≤12.1"
             },
             "safari": {
               "version_added": "6"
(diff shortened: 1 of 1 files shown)


foolip:
I see that https://mdn-bcd-collector.appspot.com/tests/api/HTMLElement uses `document.createElement('p')` as the instance to test, so that should be OK. The only member that `HTMLParagraphElement` itself has is [`align`](https://html.spec.whatwg.org/multipage/obsolete.html#HTMLParagraphElement-partial), and there's no api. HTMLElement.align entry that one would have to take a closer look at.

api/HTMLElement.json
```diff
@@ -2245,10 +2245,10 @@
               "version_added": null
             },
             "opera": {
-              "version_added": true
+              "version_added": "≤12.1"
```
foolip, line 2248: The Chrome version for this (`oncopy`) is 71, which if it were true would mean this probably not in Opera 15. But it was, I tested http://mdn-bcd-collector.appspot.com/tests/api/HTMLElement/oncopy. The Chrome data is wrong.

But anyway, for all of the PRs setting the version to ≤12.1, are they things that you've confirmed to be in Opera 12.1, as well as Opera 15 and later

It would generally be useful for PR descriptions to name which browser versions were tested.

foolip:
I'll hold off reviewing the Opera PRs until https://github.com/mdn/browser-compat-data/pull/7103#discussion_r513063766 is resolved.

foolip:
Or actually, I'll hold off reviewing changes that are only for Opera. Where there's a pair of Chromium 1 + Opera ≤12.1 PRs changing the same entries, I'm confident in those.

vinyldarkscratch:
The clipboard event handlers seem to be supported in Chrome 1, yes -- I can cherry-pick those changes and put them in a new PR/ together with Chrome changes?

foolip:
@vinyldarkscratch yeah, or since this is the PR where it was discovered, just roll more HTMLElement changes into this one and rename it. With lots of small PRs, it's OK if *some* of them take a few rounds and grow in size.

foolip:
Per https://github.com/mdn/browser-compat-data/pull/7107#issuecomment-718588929 I have confidence in the methodology behind these changes. I'm no longer asking for more changes to be rolled into this PR, we will update Chromium separately.

foolip (approved):

PR #7054, opened 2020-10-27
Set Chromium to 1 for HTMLFieldSetElement

vinyldarkscratch:
This PR uses the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com/) project to set `true` to `1` for Chrome for the `HTMLFieldSetElement` API and many of its subfeatures.  Data is also mirrored to Chrome Android, Samsung Internet, and WebView Android.

PR #7105, opened 2020-10-27
Set Opera to ≤12.1 for HTMLFieldSetElement

vinyldarkscratch:
This PR uses the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com/) project to set `true` to `≤12.1` for Opera for the `HTMLFieldSetElement` API and many of its subfeatures.  Data is also mirrored to Opera Android.

### DIFF
--- a/api/HTMLFieldSetElement.json
+++ b/api/HTMLFieldSetElement.json
@@ -23,10 +23,10 @@
             "version_added": true
           },
           "opera": {
-            "version_added": true
+            "version_added": "≤12.1"
           },
           "opera_android": {
-            "version_added": true
+            "version_added": "≤12.1"
           },
           "safari": {
             "version_added": true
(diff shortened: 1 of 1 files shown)


api/HTMLFieldSetElement.json
```diff
@@ -70,10 +70,10 @@
               "version_added": true
             },
             "opera": {
-              "version_added": true
+              "version_added": "≤12.1"
```
foolip, line 73: https://github.com/mdn/browser-compat-data/pull/7054 doesn't update `disabled` (this entry) or the following entries for `elements`, `validationMessage` or `validity` updated in this PR.

Is that because they *weren't* introduced in Chrome 1? If they were not, then when reviewing I'd check if they were introduced so late that they were actually removed in Opera 15.

vinyldarkscratch: The collector data that went into this PR _did_ include Opera 15 in its results; if it wasn't present in Opera 15, the script would generate a `version_removed` statement for us.

Most likely this is another case of a feature being in Chrome 1 but the existing data says otherwise -- I promise to follow up with some more PRs to fix those issues!

foolip: Super, thank you!

foolip (approved):

PR #7056, opened 2020-10-27
Set Chromium to 1 for HTMLFormElement

vinyldarkscratch:
This PR uses the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com/) project to set `true` to `1` for Chrome for the `HTMLFormElement` API and many of its subfeatures.  Data is also mirrored to Chrome Android, Samsung Internet, and WebView Android.

PR #7107, opened 2020-10-27
Set Opera to ≤12.1 for HTMLFormElement

vinyldarkscratch:
This PR uses the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com/) project to set `true` to `≤12.1` for Opera for the `HTMLFormElement` API and many of its subfeatures.  Data is also mirrored to Opera Android.

### DIFF
--- a/api/HTMLFormElement.json
+++ b/api/HTMLFormElement.json
@@ -70,10 +70,10 @@
               "version_added": null
             },
             "opera": {
-              "version_added": true
+              "version_added": "≤12.1"
             },
             "opera_android": {
-              "version_added": true
+              "version_added": "≤12.1"
             },
             "safari": {
               "version_added": true
@@ -118,10 +118,10 @@
               "version_added": null
             },
             "opera": {
-              "version_added": true
+              "version_added": "≤12.1"
             },
             "opera_android": {
-              "version_added": true
+              "version_added": "≤12.1"
             },
             "safari": {
               "version_added": true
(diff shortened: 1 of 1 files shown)


api/HTMLFormElement.json
```diff
@@ -214,10 +214,10 @@
               "version_added": null
             },
             "opera": {
-              "version_added": true
+              "version_added": "≤12.1"
```
foolip, line 217: This is the `checkValidity` entry, not updated by https://github.com/mdn/browser-compat-data/pull/7056. Is that an oversight, or when was it added to Chrome?

api/HTMLFormElement.json
```diff
@@ -599,10 +599,10 @@
...
```
foolip, line 602: This entry, `noValidate`, was also not updated in https://github.com/mdn/browser-compat-data/pull/7056.

foolip (commented):
I see two entries not updated in https://github.com/mdn/browser-compat-data/pull/7056 that I'd like to understand before merging either PR.

foolip:
https://github.com/mdn/browser-compat-data/pull/7105#discussion_r513688280 and current results in https://github.com/foolip/mdn-bcd-results show that the Opera versions tested are 12.14, 15 and 70. That makes me confident in the results, and I'm not very worried about mismatches what's touched in Chromium and Opera PRs as a result.

I have been and will keep using https://github.com/mdn/browser-compat-data/pull/6862 to summarize the differences to myself while reviewing, however.

@ddbeck FYI, this explains most of how I've approached these HTML* PRs.

foolip (approved):
Goes with https://github.com/mdn/browser-compat-data/pull/7056

ddbeck:
> @ddbeck FYI, this explains most of how I've approached these HTML* PRs.

Makes sense. Thank you, @foolip! 👍

PR #7114, opened 2020-10-27
Set Opera to ≤12.1 for HTMLIFrameElement

vinyldarkscratch:
This PR uses the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com/) project to set `true` to `≤12.1` for Opera for the `HTMLIFrameElement` API and many of its subfeatures.  Data is also mirrored to Opera Android.

### DIFF
--- a/api/HTMLIFrameElement.json
+++ b/api/HTMLIFrameElement.json
@@ -23,10 +23,10 @@
             "version_added": true
           },
           "opera": {
-            "version_added": true
+            "version_added": "≤12.1"
           },
           "opera_android": {
-            "version_added": true
+            "version_added": "≤12.1"
           },
           "safari": {
             "version_added": true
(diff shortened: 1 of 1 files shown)


api/HTMLIFrameElement.json
```diff
@@ -70,10 +70,10 @@
               "version_added": null
             },
             "opera": {
-              "version_added": true
+              "version_added": "≤12.1"
```
foolip, line 73: This is not consistent with it being added in Chrome 43, so that data must be wrong. That's the case for a lot of changes in this PR which aren't in https://github.com/mdn/browser-compat-data/pull/7063, but I trust these changes based on them being based on testing in Opera 12.1, 15 and the latest version of Opera.

foolip (approved):
Merging together with https://github.com/mdn/browser-compat-data/pull/7063.

PR #7115, opened 2020-10-27
Set Opera to ≤12.1 for HTMLImageElement

vinyldarkscratch:
This PR uses the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com/) project to set `true` to `≤12.1` for Opera for the `HTMLImageElement` API and many of its subfeatures.  Data is also mirrored to Opera Android.

### DIFF
--- a/api/HTMLImageElement.json
+++ b/api/HTMLImageElement.json
@@ -116,7 +116,10 @@
               "version_added": null
             },
             "opera": {
-              "version_added": true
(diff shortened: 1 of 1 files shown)


foolip (approved):
This doesn't have a companion "Set Chromium to 1" PR, but mostly `chrome` is already set to "1" and `chrome_android` is set to true. So `chrome_android` mirroring as a follow-up would be good.

PR #7116, opened 2020-10-27
Set Opera to ≤12.1 for HTMLInputElement

vinyldarkscratch:
This PR uses the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com/) project to set `true` to `≤12.1` for Opera for the `HTMLInputElement` API and many of its subfeatures.  Data is also mirrored to Opera Android.

### DIFF
--- a/api/HTMLInputElement.json
+++ b/api/HTMLInputElement.json
@@ -70,10 +70,10 @@
               "version_added": true
             },
             "opera": {
-              "version_added": true
+              "version_added": "≤12.1"
             },
             "opera_android": {
-              "version_added": true
+              "version_added": "≤12.1"
             },
             "safari": {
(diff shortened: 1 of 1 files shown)


api/HTMLInputElement.json
```diff
@@ -504,10 +504,10 @@
               "version_added": null
             },
             "opera": {
-              "version_added": true
+              "version_added": "≤12.1"
```
foolip, line 507: This is `height` and I found it suspect that there's no corresponding change for `width`. Filed https://github.com/mdn/browser-compat-data/issues/7165 about that.

foolip (approved):
Will merge together with https://github.com/mdn/browser-compat-data/pull/7064.

PR #7118, opened 2020-10-27
Set Opera to ≤12.1 for HTMLLabelElement

vinyldarkscratch:
This PR uses the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com/) project to set `true` to `≤12.1` for Opera for the `HTMLLabelElement` API and many of its subfeatures.  Data is also mirrored to Opera Android.

### DIFF
--- a/api/HTMLLabelElement.json
+++ b/api/HTMLLabelElement.json
@@ -23,10 +23,10 @@
             "version_added": true
           },
           "opera": {
-            "version_added": true
+            "version_added": "≤12.1"
           },
           "opera_android": {
-            "version_added": true
+            "version_added": "≤12.1"
           },
           "safari": {
             "version_added": true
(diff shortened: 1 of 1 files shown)


api/HTMLLabelElement.json
```diff
@@ -70,10 +70,10 @@
               "version_added": true
             },
             "opera": {
-              "version_added": true
+              "version_added": "≤12.1"
```
foolip, line 73: This (`control`) isn't included in https://github.com/mdn/browser-compat-data/pull/7066 but I take it that's because it really wasn't in Chrome 1. Looking forward to it in a later PR.

foolip (approved):
Will merge together with https://github.com/mdn/browser-compat-data/pull/7066.

PR #7122, opened 2020-10-27
Set Opera to ≤12.1 for HTMLMarqueeElement

vinyldarkscratch:
This PR uses the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com/) project to set `true` to `≤12.1` for Opera for the `HTMLMarqueeElement` API and many of its subfeatures.  Data is also mirrored to Opera Android.

### DIFF
--- a/api/HTMLMarqueeElement.json
+++ b/api/HTMLMarqueeElement.json
@@ -259,10 +259,10 @@
               "version_added": null
             },
             "opera": {
-              "version_added": true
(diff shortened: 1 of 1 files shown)


foolip (approved):
There's no "Chromium to 1" companion for this PR, but it looks good, `<marquee>` is ancient :) I confirmed support in Chrome 15, but don't know if it went back to Chrome 1.

PR #7123, opened 2020-10-27
Set Opera to ≤12.1 for HTMLMediaElement

vinyldarkscratch:
This PR uses the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com/) project to set `true` to `≤12.1` for Opera for the `HTMLMediaElement` API and many of its subfeatures.  Data is also mirrored to Opera Android.

### DIFF
--- a/api/HTMLMediaElement.json
+++ b/api/HTMLMediaElement.json
@@ -131,10 +131,10 @@
               "version_added": null
             },
             "opera": {
-              "version_added": true
+              "version_added": "≤12.1"
             },
             "opera_android": {
-              "version_added": true
+              "version_added": "≤12.1"
             },
             "safari": {
               "version_added": true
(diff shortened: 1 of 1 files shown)


api/HTMLMediaElement.json
```diff
@@ -131,10 +131,10 @@
               "version_added": null
             },
             "opera": {
-              "version_added": true
+              "version_added": "≤12.1"
```
foolip, line 134: There's no "Chromium to 1" buddy PR for this, so I notice that the Chrome data for `addTextTrack` is wrong. It was in Opera 15 and Chrome 28, so this bit is correct. I trust there will be more PRs to fix the incorrect Chrome data.

As an aside, I don't believe there was ever a prefixed variant of this, so the data is wrong in multiple ways. There was an alternative name though, see https://bugs.webkit.org/show_bug.cgi?id=77381.

foolip (approved):
Deja vu! I shipped most of these in Opera 10.5 :D https://dev.opera.com/blog/re-introducing-video/

Looking forward to future PRs setting the exact versions for this.

PR #7153, opened 2020-10-27
Notes about Safari around sandbox-allow-same-origin

aleen42:

- [x] Summarize your changes

    Make a note about sandbox-allow-same-origin in Safari

     I have tested under Safari 13.1.2 (14609.3.5.1.5) and found that when setting with `sandbox="allow-same-origin"` without `allow-scripts`, we can't bind any event handlers upon nodes inside an iframe via accessing `iframe.contentDocument` because Safari may throw an error:

    ```
   [Error] Blocked script execution in 'xxx' because the document's frame is sandboxed and the 'allow-scripts' permission is not set.
    ``` 

    Besides, the same error can be caught when trying to post messages toward such an iframe.

- [ ] Link to related issues or pull requests if any

### DIFF
--- a/html/elements/iframe.json
+++ b/html/elements/iframe.json
@@ -1027,6 +1027,55 @@
             }
           }
         },
+        "sandbox-allow-same-origin": {
+          "__compat": {
+            "description": "<code>sandbox=\"allow-same-origin\"</code>",
+            "support": {
+              "chrome": {
+                "version_added": true
+              },
+              "chrome_android": {
+                "version_added": true
+              },
+              "edge": {
+                "version_added": true
+              },
+              "firefox": {
+                "version_added": true
+              },
+              "firefox_android": {
+                "version_added": true
+              },
+              "ie": {
+                "version_added": true
+              },
+              "opera": {
+                "version_added": true
+              },
+              "opera_android": {
+                "version_added": true
+              },
+              "safari": {
+                "version_added": true,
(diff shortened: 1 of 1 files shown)


aleen42:
Any progress?

html/elements/iframe.json
```diff
@@ -1027,6 +1027,55 @@
...
+              },
+              "opera": {
+                "version_added": true
+              },
+              "opera_android": {
+                "version_added": true
+              },
+              "safari": {
+                "version_added": true,
+                "notes": "Safari will block script execution without 'allow-scripts' even if 'allow-same-origin' is set. For example, any bound handlers for click events of nodes inside an iframe will throw an Error of blocking script execution."
```
ddbeck, line 1060: suggestion: writing style and markup

Nothing major here. For consistency with other notes, we'll use the present tense and mark up keywords with code.

```suggestion
                "notes": "Safari blocks script execution without <code>allow-scripts</code> even if <code>allow-same-origin</code> is set. For example, any bound handlers for click events of nodes inside an iframe throw an error for blocked script execution."
```

ddbeck (changes requested):
Thank you for this PR, @aleen42. Some little style suggestions for the note in a line comment, but otherwise this look good!

html/elements/iframe.json
aleen42: Well done!

ddbeck (approved):
Thank you! 🎉

aleen42:
Recently, I found that the problem also arises before Chrome 71, and I need to search for some detailed changelog of Chrome to check.

PR #7156, opened 2020-10-27
Reduce README to upgrade doc

ddbeck:
This reduces the README in `master` down to a notice to update to `@mdn/browser-compat-data`. This is the closest we can get to a redirect on npm.

I plan to merge this PR shortly before releasing the next and final version of `mdn-browser-compat-data`, this Thursday.

The changes won't be long for this world, however. The content of the README will be restored by #7155, which will clobber the changes here.


ddbeck:
Merging this now and starting the release process; the README will look funny for a short time today.

PR #7172, opened 2020-10-29
Navigator/vibrate - Firefox for Android doesn't vibrate

VoxelPrismatic:
- [X] Summarize your changes
- [X] Data: link to resources that verify support information (such as browser's docs, changelogs, source control, bug trackers, and tests)
- [X] Data: if you tested something, describe how you tested with details like browser and version

**Summary:** Firefox for Android [Daylight, Nightly, Beta] doesn't support vibration, although it always returns true when the window is active
**Data:** Tested on Firefox for Android Daylight, Nightly and Beta on a Galaxy S8 running Android 9. Test it out with [this tool](https://voxelprismatic.github.io/prizm.dev/vibration_api_test)
**Notes:** I tested that same tool out in Samsung Internet, vibration works fine. I also have `mozVibrate()` on that tool just in case.

**Possible help:** Is there a separate note for API/Vibration_API or is it linked? Otherwise all is good

### DIFF
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -2276,8 +2276,7 @@
                 "notes": [
                   "Until Firefox 26 included, when the vibration pattern was too long or any of its elements too large, Firefox threw an exception instead of returning <code>false</code> (<a href='https://bugzil.la/884935'>bug 884935</a>).",
                   "From Firefox 32 onwards, when the vibration pattern is too long or any of its elements too large, it returns <code>true</code> but truncates the pattern (<a href='https://bugzil.la/1014581'>bug 1014581</a>).",
-                  "Beginning in Firefox 72, this is not supported in cross-origin iframes.",
-                  "Firefox for Android doesn't support vibration, and always returns `true` when the window is visible."
+                  "Beginning in Firefox 72, this is not supported in cross-origin iframes."
                 ]
               },
               {
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -2287,15 +2287,17 @@
             ],
             "firefox_android": [
               {
-                "version_added": "16",
-                "notes": [
-                  "Until Firefox 26 included, when the vibration pattern was too long or any of its elements too large, Firefox threw an exception instead of returning <code>false</code> (<a href='https://bugzil.la/884935'>bug 884935</a>).",
-                  "From Firefox 32 onwards, when the vibration pattern is too long or any of its elements too large, it returns <code>true</code> but truncates the pattern (<a href='https://bugzil.la/1014581'>bug 1014581</a>)."
-                ]
+                "version_added": "72",
+                "partial_implementation": true,
+                "notes": "Vibration is disabled. If the window is visible, then <code>navigator.vibrate()</code> returns <code>true</code>, but no vibration takes place (regardless of hardware support). See <a href='https://bugzil.la/1591113'>bug 1591113</code>."
               },
               {
-                "version_added": "14",
-                "prefix": "moz"
+                "version_added": "16",
+                "version_removed": "72",
+                "notes": [
+                  "Until Firefox 26 included, when the vibration pattern was too long or any of its elements too large, Firefox threw an exception instead of returning <code>false</code> (<a href='https://bugzil.la/884935'>bug 884935</a>).",
+                  "From Firefox 32 onwards, when the vibration pattern is too long or any of its elements too large, it returns <code>true</code> but truncates the pattern (<a href='https://bugzil.la/1014581'>bug 1014581</a>).",
+                ]
               }
             ],
             "ie": {
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -2293,7 +2293,7 @@
               },
               {
                 "version_added": "16",
-                "version_removed": "72",
+                "version_removed": "79",
                 "notes": [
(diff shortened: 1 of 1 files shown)


saschanaz:
There is a bug for this: https://github.com/mozilla-mobile/fenix/issues/2371

VoxelPrismatic:
I didn't catch that one, thank you.

saschanaz:
Oops, we need this PR because it really is a known issue! If this has been broken for a while we should document it here.

VoxelPrismatic:
well then

ddbeck:
Wow, the existing data here is kinda messy! I think the gist here is that vibration was (inadvertently?) removed in Firefox 72. Check out https://bugzilla.mozilla.org/show_bug.cgi?id=1591113 for details. I think this means we need to restructure the data for both the "firefox" entry and "firefox_android".

It should look a little something like this (ignoring the prefixed entry after these):

```json
  {
    "version_added": "72",
    "notes": "Vibration is disabled. If the window is visible, then <code>navigator.vibrate()</code> returns <code>true</code>, but no vibration takes place (regardless of hardware support). See <a href='https://bugzil.la/1591113'>bug 1591113</code>."
  },
  {
    "version_added": "16",
    "version_removed": "72",
    "notes": [
      "Until Firefox 26 included, when the vibration pattern was too long or any of its elements too large, Firefox threw an exception instead of returning <code>false</code> (<a href='https://bugzil.la/884935'>bug 884935</a>).",
      "From Firefox 32 onwards, when the vibration pattern is too long or any of its elements too large, it returns <code>true</code> but truncates the pattern (<a href='https://bugzil.la/1014581'>bug 1014581</a>).",
    ]
  }
```

VoxelPrismatic:
I dont think we should put it on `firefox` because as far as i know, no desktop or laptop actually has a vibration motor. Adding that info would be redundant.

VoxelPrismatic:
strange, the build is failing because firefox version 72 is invalid even though the latest firefox daylight is 82.1.3

ddbeck:
Oops, that was an oversight on my part. Firefox for Android had a discontinuity—it jumped from 68 to 79. So that value should be bumped to 79. Sorry about that!

api/Navigator.json
```diff
@@ -2285,16 +2286,18 @@
               }
             ],
             "firefox_android": [
+              {
+                "version_added": "72",
```
vinyldarkscratch, line 2290:
```suggestion
                "version_added": "79",
```

api/Navigator.json
```diff
@@ -2285,16 +2286,18 @@
...
+                "partial_implementation": true,
+                "notes": "Vibration is disabled. If the window is visible, then <code>navigator.vibrate()</code> returns <code>true</code>, but no vibration takes place (regardless of hardware support). See <a href='https://bugzil.la/1591113'>bug 1591113</code>."
+              },
               {
                 "version_added": "16",
+                "version_removed": "72",
```
vinyldarkscratch, line 2296:
```suggestion
                "version_removed": "79",
```

vinyldarkscratch (dismissed):

ddbeck:
@VoxelPrismatic do you plan to return this PR? Or should we open a follow up issue instead? Either way, thank you!

VoxelPrismatic:
Oh sorry I didn't receive any emails on this thread.

VoxelPrismatic:
Just tested on Nightly 210125 17:01 [latest as of writing], still doesn't vibrate

api/Navigator.json
```diff
@@ -2285,16 +2286,18 @@
...
+                "notes": "Vibration is disabled. If the window is visible, then <code>navigator.vibrate()</code> returns <code>true</code>, but no vibration takes place (regardless of hardware support). See <a href='https://bugzil.la/1591113'>bug 1591113</code>."
+              },
               {
                 "version_added": "16",
+                "version_removed": "79",
                 "notes": [
                   "Until Firefox 26 included, when the vibration pattern was too long or any of its elements too large, Firefox threw an exception instead of returning <code>false</code> (<a href='https://bugzil.la/884935'>bug 884935</a>).",
                   "From Firefox 32 onwards, when the vibration pattern is too long or any of its elements too large, it returns <code>true</code> but truncates the pattern (<a href='https://bugzil.la/1014581'>bug 1014581</a>)."
                 ]
-              },
```
ddbeck, line 2294: suggestion: instead of removing the `moz` prefix support object here, I propose we add `"version_removed": true`" instead.

`mozVibrate` definitely existed historically, but I don't know when it was removed. We could test to find out, but I think it's out of scope to for this PR. I'd like to see the really big thing (vibration not doing anything on Android) get merged before investigating a historical footnote. 🙂

ddbeck (changes requested):
Thank you for persisting at this, @VoxelPrismatic! A couple of suggestions here. We're quite close, but it's unusual data.

api/Navigator.json
ddbeck: This is unresolved still. We need something like this in the final array:

```
              {
                "version_added": "11",
                "version_removed": true,
                "prefix": "moz"
              }
```

VoxelPrismatic: easy enough

ddbeck (approved):
Thank you! 🎉

PR #7177, opened 2020-10-30
Update Opera data for HTML element APIs

vinyldarkscratch:
This PR updates the Opera data for the HTML element APIs based upon results from the mdn-bcd-collector project (using results from Opera 12.14, 15, and 70), followed by additional mirroring from Chrome data.

### DIFF
--- a/api/HTMLAreaElement.json
+++ b/api/HTMLAreaElement.json
@@ -214,10 +214,10 @@
               "version_added": true
             },
             "opera": {
-              "version_added": true
+              "version_added": "41"
             },
             "opera_android": {
-              "version_added": true
+              "version_added": "41"
             },
             "safari": {
               "version_added": true
@@ -454,10 +454,10 @@
               "version_added": true
             },
             "opera": {
-              "version_added": true
+              "version_added": "41"
             },
             "opera_android": {
-              "version_added": true
+              "version_added": "41"
             },
             "safari": {
               "version_added": true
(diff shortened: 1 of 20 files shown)


foolip:
Was mirroring done with https://github.com/mdn/browser-compat-data/pull/7179 applied, or might this be mirroring some data known to be incorrect?

vinyldarkscratch:
I made sure to cherry-pick the mirroring so that it is only performed on data that doesn't conflict with #7179; there may be some crossover with that PR, if so, my bad!

api/HTMLAreaElement.json
```diff
@@ -214,10 +214,10 @@
               "version_added": true
             },
             "opera": {
-              "version_added": true
+              "version_added": "41"
```
foolip, line 217: Confirmed with Opera 40+41 on BrowserStack using https://mdn-bcd-collector.appspot.com/tests/api/HTMLAreaElement. I thought this seems too recent, but it's correct.

api/HTMLElement.json
```diff
@@ -1506,7 +1506,8 @@
               "version_removed": "15"
             },
             "opera_android": {
-              "version_added": null
+              "version_added": "11",
+              "version_removed": "14"
```
foolip, line 1510: Ah, yes, the Microdata API. This has been removed for long enough that it can be removed from BCD now. (Not in this PR though.)

api/HTMLSpanElement.json
```diff
@@ -23,10 +23,10 @@
             "version_added": true
           },
           "opera": {
-            "version_added": true
+            "version_added": "15"
```
foolip, line 26: Interesting. `<span>` is ancient, but in Presto it was a plain `HTMLElement`, not `HTMLSpanElement`. So this is correct.

foolip (approved):

PR #7178, opened 2020-10-30
Add Chromium data for HTML element APIs

vinyldarkscratch:
This PR updates the Chromium data for the HTML element APIs based upon results from the mdn-bcd-collector project (using results from Chrome 1-86), including mirroring to other Chromium browsers.

### DIFF
--- a/api/HTMLMediaElement.json
+++ b/api/HTMLMediaElement.json
@@ -471,10 +471,10 @@
           "description": "<code>canPlayType()</code>",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "3"
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": "18"
             },
             "edge": {
               "version_added": "12"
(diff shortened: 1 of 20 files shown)


api/HTMLMediaElement.json
```diff
@@ -1706,10 +1709,19 @@
               "version_added": false
             },
             "opera": {
-              "version_added": true
+              "version_added": "29"
```
foolip, line 1712: Confirmed with https://mdn-bcd-collector.appspot.com/tests/api/HTMLMediaElement/mediaKeys in Opera 29. This was an API I expected shipped a few versions later in Opera, but no.

api/HTMLUnknownElement.json
```diff
@@ -5,10 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLUnknownElement",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "15"
```
foolip, line 8: I don't have Chrome 14 to test with, but I tested Safari 4, which is a WebKit version soon before this, and can see that `document.createElement('bla')` returns a `HTMLElement`, not `HTMLUnknownElement`. So this seems correct. I had to go forward to Safari 6 to see it supported, while I expected 5, but nevertheless I trust these results based on how http://mdn-bcd-collector.appspot.com/tests/api/HTMLUnknownElement works.

foolip (approved):

PR #7179, opened 2020-10-30
Correct Chromium data for HTML element APIs

vinyldarkscratch:
This PR corrects the Chrome data for the HTML element APIs based upon results from the mdn-bcd-collector project (using results from Chrome 1-86 and Opera 12.14, 15, and 70).

Helpful note: I noticed that a number of data was set to "Chrome 43"; however, from my testing, I've found that Chrome 43 was a magic number of when all properties are initialized on the object prototype.  Thus, it's best to trust results that indicate Chrome 43 less.

### DIFF
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -2227,10 +2227,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLElement/oncopy",
           "support": {
             "chrome": {
-              "version_added": "1"
+              "version_added": "71"
             },
             "chrome_android": {
-              "version_added": "18"
+              "version_added": "71"
             },
             "edge": {
               "version_added": "12"
@@ -2257,10 +2257,10 @@
               "version_added": true
             },
             "samsunginternet_android": {
-              "version_added": "1.0"
+              "version_added": "10.0"
             },
             "webview_android": {
-              "version_added": "1"
+              "version_added": "71"
             }
           },
           "status": {
@@ -2275,10 +2275,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLElement/oncut",
           "support": {
             "chrome": {
-              "version_added": "1"
+              "version_added": "71"
             },
             "chrome_android": {
-              "version_added": "18"
+              "version_added": "71"
             },
             "edge": {
               "version_added": "12"
@@ -2305,10 +2305,10 @@
               "version_added": true
             },
             "samsunginternet_android": {
-              "version_added": "1.0"
+              "version_added": "10.0"
             },
             "webview_android": {
-              "version_added": "1"
+              "version_added": "71"
             }
           },
           "status": {
@@ -2371,10 +2371,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLElement/onpaste",
           "support": {
             "chrome": {
-              "version_added": "1"
+              "version_added": "71"
             },
             "chrome_android": {
-              "version_added": "18"
+              "version_added": "71"
             },
             "edge": {
               "version_added": "12"
@@ -2401,10 +2401,10 @@
               "version_added": true
             },
             "samsunginternet_android": {
-              "version_added": "1.0"
+              "version_added": "10.0"
             },
             "webview_android": {
-              "version_added": "1"
+              "version_added": "71"
             }
           },
           "status": {
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -3115,10 +3115,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLElement/spellcheck",
           "support": {
             "chrome": {
-              "version_added": "9"
+              "version_added": "43"
             },
             "chrome_android": {
-              "version_added": "18"
+              "version_added": "43"
             },
             "edge": {
               "version_added": "12"
@@ -3145,10 +3145,10 @@
               "version_added": true
             },
             "samsunginternet_android": {
-              "version_added": "1.0"
+              "version_added": "4.0"
             },
             "webview_android": {
-              "version_added": "≤37"
+              "version_added": "43"
             }
           },
           "status": {
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -772,7 +772,7 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLElement/draggable",
           "support": {
             "chrome": {
-              "version_added": "7"
+              "version_added": "3"
             },
             "chrome_android": {
               "version_added": "18"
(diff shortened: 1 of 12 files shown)


api/HTMLElement.json
```diff
@@ -2227,10 +2227,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLElement/oncopy",
           "support": {
             "chrome": {
-              "version_added": "71"
+              "version_added": "1"
```
foolip, line 2230: The original data (Chrome 71) came from https://github.com/mdn/browser-compat-data/pull/3502 which in turn used Confluence:
http://web-confluence.appspot.com/#!/catalog?releases=%5B%22Chrome_70.0.3538.67_OSX_10.13.6%22,%22Chrome_70.0.3538.67_Windows_10.0%22,%22Chrome_71.0.3578.80_OSX_10.14.0%22,%22Chrome_71.0.3578.80_Windows_10.0%22%5D&q=%22oncopy%22

What happened was that `oncopy` was moved from `Element` to `HTMLElement` here:
https://chromium.googlesource.com/chromium/src/+/bce553b6022a43cc4f9744de69ef85a753f47624

This is a similar situation to https://github.com/mdn/browser-compat-data/issues/6545, and I'd like to handle it as suggested in https://github.com/mdn/browser-compat-data/issues/6545#issuecomment-679147588, by treating it as supported on `HTMLElement` with no note even when that support was inherited from `Element`.

api/HTMLElement.json
```diff
@@ -3115,10 +3115,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLElement/spellcheck",
           "support": {
             "chrome": {
-              "version_added": "43"
```
foolip, line 3118: This came from https://github.com/mdn/browser-compat-data/pull/3502, also seemingly sourced from Confluence:
http://web-confluence.appspot.com/#!/catalog?releases=%5B%22Chrome_42.0.2311.135_OSX_10.11.6%22,%22Chrome_42.0.2311.135_Windows_10.0%22,%22Chrome_43.0.2357.65_OSX_10.11.6%22,%22Chrome_43.0.2357.81_Windows_10.0%22%5D&q=%22spellcheck%22

Here it seems like the `spellcheck` attribute has moved from a number of specific interfaces up to `HTMLElement`, which means that before Chrome 43, `document.createElement('div').spellcheck` would have returned something different.

It's not a particularly interesting difference, but nonetheless I think we'll need to treat it as `partial_implementation` before Chrome 43.

api/HTMLElement.json
```diff
@@ -3163,10 +3163,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/ElementCSSInlineStyle/style",
           "support": {
             "chrome": {
-              "version_added": "45"
+              "version_added": "1"
```
foolip, line 3166: Ugh, a mixin. Since `MathMLElement` now includes this mixin, we'll need to resolve https://github.com/mdn/browser-compat-data/issues/472. Setting it as supported from Chrome 1 is taking the *any* option from https://github.com/mdn/browser-compat-data/issues/472#issuecomment-696628301.

foolip (commented):
This is too big and has too many tricky cases to review as a single PR. I reviewed up to api/HTMLElement.json, but given the kinds of prototype moves that are going on here, each change needs careful review.

@vinyldarkscratch would you mind splitting this up into smaller parts? And it would really help if for each original Chrome version, the PR that originally added it was found, and if you could check if there was ever an attribute by the same name on some other interface. If there was, then there was likely a move on the prototype chain that might need to be represented.

api/HTMLElement.json
vinyldarkscratch: I just did a quick test, and it seems that `spellcheck` is actually present on _every_ HTML element, even unknown/made-up elements, so it's unclear to me why Confluence would report otherwise.

I wrote the following to test support:

```javascript
var tags = ['a', 'area', 'audio', 'base', 'basefont', 'body', 'br', 'button', 'canvas', 'content', 'data', 'datalist', 'details', 'dialog', 'div', 'dl', 'embed', 'fieldset', 'font', 'form', 'frame', 'frameset', 'head', 'h1', 'hr', 'html', 'iframe', 'img', 'input', 'isindex', 'keygen', 'label', 'legend', 'li', 'link', 'map', 'marquee', 'menu', 'menuitem', 'meta', 'meter', 'del', 'object', 'ol', 'optgroup', 'option', 'output', 'p', 'param', 'picture', 'pre', 'progress', 'blockquote', 'script', 'select', 'shadow', 'slot', 'source', 'span', 'style', 'caption', 'td', 'col', 'table', 'tr', 'tbody', 'template', 'textarea', 'time', 'title', 'track', 'ul', 'unknown', 'video', 'foo'];

var result = "";

for (var i = 0; i < tags.length; i++) {
	var tag = tags[i];

	var el = document.createElement(tag);
	var sc = el.spellcheck;

	if (!sc) {
		sc = "<span style='color: red;'>FALSE</span>";
	}

	result += tag + ": " + sc + "<br />";
}

var r = document.createElement('p');
r.innerHTML = result;
document.body.appendChild(r);
```

vinyldarkscratch: I cherry-picked this as #7239 in efforts to reduce the size of this PR.

api/HTMLElement.json
vinyldarkscratch: To consolidate the size of this PR, I cherry-picked these changes as #7242.

api/HTMLElement.json
vinyldarkscratch: This is the data for HTMLElement in particular, but it's the only place where the data for this mixin is.  Do you think we should keep the changes?

api/HTMLElement.json
foolip: Thanks!

api/HTMLElement.json
foolip: Thanks!

api/HTMLElement.json
foolip: Aha. So in BCD we have only api.HTMLElement.style, and https://developer.mozilla.org/en-US/docs/Web/API/ElementCSSInlineStyle/style includes that. That's a weird mismatch, but in any case this BCD entry should be for `HTMLElement` only based on the path, so this is good!

api/HTMLInputElement.json
```diff
@@ -1683,10 +1683,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLInputElement/webkitEntries",
           "support": {
             "chrome": {
-              "version_added": "13"
+              "version_added": "22"
```
foolip, line 1686: This is the biggest change in the PR in terms of removing claimed support. Can you check where the Chrome 13 claim came from?

(There are a few other changes that bump `version_added` by one version.)

foolip (commented):
Just one remaining question about `webkitEntries`.

api/HTMLInputElement.json
vinyldarkscratch: Looks like the original data comes from the wiki migration (in fact, it happens to be my PR, haha!), so I have a feeling that it was inaccurate.

foolip: Alright, thanks for checking!

PR #7180, opened 2020-10-30
Update Chalk to v4.1.0

vinyldarkscratch:
This PR bumps `chalk` to v4.1.0.  The only difference that affects our project is that `chalk` v4 requires NodeJS v10; otherwise, it's basically the same as `chalk` v3.0.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -138,9 +138,9 @@
       "dev": true
     },
     "chalk": {
-      "version": "3.0.0",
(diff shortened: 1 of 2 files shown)


ddbeck:
Thanks, @vinyldarkscratch, but I'm going to close these dependency PRs. I'd like to use these upgrades to confirm that the new dependabot config works (and, historically, the bot provided a pretty good paper trail for change logs, compat stats, and the like—I'd like to keep that stuff around).

PR #7192, opened 2020-10-30
Add "bcd" to package keyword list

### LINKED ISSUE #7187: Add "bcd" as a package keyword?
Now BCD is frequently used as an abbreviation of browser-compat-data but the npm search with "bcd" does not show our package. Maybe we should add one.

saschanaz:
Fixes #7187


ddbeck (approved):
Good idea! 👍

PR #7193, opened 2020-10-30
Remove notes about document.all being readonly

foolip:
These notes were added in https://github.com/mdn/browser-compat-data/pull/3391,
after the change in https://chromium-review.googlesource.com/c/chromium/src/+/823513/.

Previously it was possible to assign document.all, and now it is not.
The uses of document.all that currently work also work with older
browsers, so there isn't anything that web developers need to be aware
of here. The only time when it might have mattered was at the time of
the change, if someone depending on being able to use document.all to
store their own values, which in itself is an unlikely scenario.

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -342,8 +342,7 @@
           "support": {
             "chrome": [
               {
-                "version_added": "64",
-                "notes": "Starting in Chrome 65, this property is readonly."
+                "version_added": "64"
               },
               {
                 "version_added": "3",
@@ -354,8 +353,7 @@
             ],
             "chrome_android": [
               {
-                "version_added": "64",
-                "notes": "Starting in Chrome 65, this property is readonly."
+                "version_added": "64"
               },
               {
                 "version_added": "18",
@@ -378,8 +376,7 @@
             },
             "opera": [
               {
-                "version_added": "51",
-                "notes": "Starting in Opera 52, this property is readonly."
+                "version_added": "51"
               },
               {
                 "version_added": "≤12.1",
@@ -390,8 +387,7 @@
             ],
             "opera_android": [
               {
-                "version_added": "47",
-                "notes": "This property is readonly."
+                "version_added": "47"
               },
               {
                 "version_added": "≤12.1",
@@ -408,8 +404,7 @@
             },
             "samsunginternet_android": [
               {
-                "version_added": "9.0",
-                "notes": "This property is readonly."
+                "version_added": "9.0"
               },
               {
                 "version_added": "1.0",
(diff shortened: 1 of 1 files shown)


foolip:
@jpmedley @ddbeck I'd like both of your feedback on this one, if possible.

@vinyldarkscratch this is the followup I promised in #6865.

jpmedley:
It's not clear to me why you're removing this information. We tend not to remove information that is correct even after it ceases to be enabled. (I'm excluding my position on flags here because flags aren't part of the platform.)

vinyldarkscratch:
I'd say that the reasoning behind the removal is because this property is defined as a read-only property per the spec.  The chances of someone wanting to override a property like `document.all` are very slim, so a note like this isn't relevant to developers.

ddbeck:
What's the story for the specification and the implementation? From #3391, it kinda looks like there was some sort of ambiguity in the spec. That seems interesting enough to note, though I'd probably frame it in historical terms ("Before version 65, this property is assignable… From version 65, the property is readonly, conforming to the updated specification" or something along those lines).

I'm trying to imagine a scenario where dropping the note makes sense and I guess it would be if the assignability was undefined in a specification that otherwise defined the behavior of `document.all`. Strictly speaking, there'd be no expectation of compatibility. But even that feels a tenuous.

foolip:
`document.all` was added to HTML in https://github.com/whatwg/html/commit/906460ab271640fb7a0c6cba9d9513a32f0c0ffe, and was a readonly attribute from the start. There were a bunch of changes after that, for example https://github.com/whatwg/html/commit/7e2400ba8c2df64cc6313237c84121438547ae6e + https://github.com/whatwg/html/commit/6ba5e6df8a6dff2c5a4d43b774342ea752f64ebd introduced+used `HTMLAllCollection` for it.

Before https://chromium-review.googlesource.com/c/chromium/src/+/823513/, it wasn't writable in any interesting way, it was merely possible to replace the object. Compare this do [setting `document.body`](https://html.spec.whatwg.org/#dom-document-body) which actually does something.

The reason that `document.all` was replaceable to begin with in Chromium is that it started out as having custom bindings in WebKit, and then was migrated to something more standard in a number of steps. These are two of them:
https://codereview.chromium.org/13674015
https://codereview.chromium.org/14589010

Being replaceable was never an interesting part of the API, and I don't think it's plausible that these notes will be helpful, even if they say something true.

ddbeck (approved):
OK, I'm sold on this. It's more of an interesting story for spec authors and implementers, but not so much for web developers. Thank you for filling in all the details, @foolip! 👍

PR #7195, opened 2020-10-31
Update HTML element API data for IE/Edge using mdn-bcd-collector

vinyldarkscratch:
This PR updates the data for the HTML element APIs for Internet Explorer and Microsoft Edge using the mdn-bcd-collector project (using results from IE 5.5 to 11, and Edge 13 to 85), aided with some manual adjustments.  Because there is no real way to differentiate whether an element is initialized using the HTMLElement API or its own API until IE 9, the version numbers for the APIs themselves are guesstimated based upon the values from the subfeatures (or when not possible, when the elements themselves were supported).  Additionally, this removes a little useless and inaccurate data on `api.HTMLScriptElement.defer` where it was marked as non-standard implementation since IE 4 -- the HTMLElement API wasn't supported until IE 5.5, and having a note that just reads "Non-standard implementation" is simply unhelpful.

### DIFF
--- a/api/HTMLAnchorElement.json
+++ b/api/HTMLAnchorElement.json
@@ -20,7 +20,7 @@
             "version_added": "4"
           },
           "ie": {
-            "version_added": true
+            "version_added": "5.5"
           },
           "opera": {
             "version_added": "≤12.1"
@@ -115,7 +115,7 @@
               "version_added": "4"
             },
             "ie": {
-              "version_added": true
+              "version_added": "6"
             },
             "opera": {
               "version_added": "≤12.1"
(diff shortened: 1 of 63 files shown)


foolip:
This PR is too big for me, and there are changes in here that warrant more attention than others. For example, the `msIsStereo3D` change seems suspect and I'd want to take a closer look.

@vinyldarkscratch can you split this a bit by what the original values were? I think one PR for each of true, false and null would make sense, and then finally a PR for where there was a real or ranged version already. That last one is in need of closer review than the rest.

vinyldarkscratch:
I've separated this into #7246 and #7247.  If desired, I can break them down even further!

foolip:
Thanks!

PR #7196, opened 2020-10-31
Set Chromium 1 + Opera ≤12.1/15 for many SVG Element APIs

vinyldarkscratch:
This PR sets Chromium to 1 and Opera to ≤12.1/15 for various SVG element APIs based upon results from the mdn-bcd-collector project, followed by mirroring to Chrome Android, Opera Android, Samsung Internet, and WebView.  Data comes from the results from Chrome 1-85, as well as Opera 12.14, 15, and 70, on both macOS and Windows.

Notes:
- This PR removes a `version_removed` from `api.SVGPathElement.pathLength`.  Collector results indicate that this feature was never removed from Chrome.
- There are several results that change WebView from `false` to `1` (and one entry from `3` to `1`).  This is because anything that is in Chrome 1 is _guaranteed_ to be in WebView for Android 1 (as stated by Joe Medley).

### DIFF
--- a/api/SVGAElement.json
+++ b/api/SVGAElement.json
@@ -5,10 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/SVGAElement",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "1"
           },
           "chrome_android": {
-            "version_added": true
+            "version_added": "18"
           },
           "edge": {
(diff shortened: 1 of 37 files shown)


foolip (approved):
This is a huge change, but a very regular one, setting all versions to the same values. I've looked especially at the cases where the original version was false, and those make sense to me.

PR #7197, opened 2020-10-31
Update Chrome/Opera data for SVG element APIs using mdn-bcd-collector

vinyldarkscratch:
This PR sets results (from true/null to a real value) for Chrome and Opera for various SVG element APIs based upon results from the mdn-bcd-collector project, followed by mirroring to Chrome Android, Opera Android, Samsung Internet, and WebView.  Data comes from the results from Chrome 1-85, as well as Opera 12.14, 15, and 70, on both macOS and Windows.

Notes:
- There are several results that change WebView from `false` to `1` (and one entry from `3` to `1`).  This is because anything that is in Chrome 1 is _guaranteed_ to be in WebView for Android 1 (as stated by Joe Medley).

### DIFF
--- a/api/SVGComponentTransferFunctionElement.json
+++ b/api/SVGComponentTransferFunctionElement.json
@@ -5,10 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/SVGComponentTransferFunctionElement",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "5"
           },
           "chrome_android": {
-            "version_added": true
+            "version_added": "18"
           },
           "edge": {
             "version_added": "≤18"
@@ -23,10 +23,10 @@
             "version_added": null
           },
           "opera": {
-            "version_added": true
+            "version_added": "15"
           },
           "opera_android": {
-            "version_added": true
+            "version_added": "14"
           },
           "safari": {
             "version_added": true
@@ -35,10 +35,10 @@
             "version_added": true
           },
           "samsunginternet_android": {
-            "version_added": true
+            "version_added": "1.0"
           },
           "webview_android": {
-            "version_added": true
+            "version_added": "≤37"
           }
         },
         "status": {
--- a/api/SVGGraphicsElement.json
+++ b/api/SVGGraphicsElement.json
@@ -5,10 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/SVGGraphicsElement",
         "support": {
           "chrome": {
-            "version_added": "30"
+            "version_added": true
           },
           "chrome_android": {
-            "version_added": "30"
+            "version_added": true
           },
           "edge": {
             "version_added": "12"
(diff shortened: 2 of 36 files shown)


api/SVGGraphicsElement.json
```diff
@@ -5,10 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/SVGGraphicsElement",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "30"
```
foolip, line 8: http://mdn-bcd-collector.appspot.com/tests/api/SVGGraphicsElement doesn't have an instance because it's a base class for other things. It was introduced in https://chromium.googlesource.com/chromium/src/+/1c0121db5bdec1c757a78a9eb7b2888985809cd8 and probably moved some attributes around on the prototype chain. As a result the version numbers for subfeatures are nuanced. We may need `partial_implementation`.

Can you check if there are other interfaces updated in this PR for which we also don't test any instance? It might be good to break those out into its own PR. This PR is big, but until I got to this point seemed pretty straightforward.

vinyldarkscratch: It looks like we don't have a custom test implemented for this API, oops.  I've reverted the changes to this and will add a custom test so we can test this properly.

foolip: Great, thank you!

It was I who removed the custom test for `SVGComponentTransferFunctionElement` + `SVGGraphicsElement` + `SVGTextContentElement` + `SVGTextPositioningElement ` in https://github.com/foolip/mdn-bcd-collector/pull/643.

`SVGTextPositioningElement` isn't updated in this PR, but the remaining two are, and those tests also don't use any instances:
http://mdn-bcd-collector.appspot.com/tests/api/SVGComponentTransferFunctionElement
http://mdn-bcd-collector.appspot.com/tests/api/SVGTextContentElement

api/SVGComponentTransferFunctionElement.json
```diff
@@ -5,10 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/SVGComponentTransferFunctionElement",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "5"
```
foolip, line 8: This is one of the interfaces without instances in the tests mentioned in https://github.com/mdn/browser-compat-data/pull/7197#discussion_r516655475, but this data looks plenty plausible anyway, and based on https://caniuse.com/svg-filters it couldn't really have been earlier than this.

api/SVGTextContentElement.json
```diff
@@ -427,10 +427,10 @@
         "__compat": {
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "43"
```
foolip: This seems wrong. http://mdn-bcd-collector.appspot.com/tests/api/SVGTextContentElement doesn't test any instances and I think this is just a matter of the interface object not being exposed until Chrome 43, or something like that. The IDL file for this goes back to 2005: https://trac.webkit.org/changeset/10577/webkit

I'll revert the changes to this file to get this PR landed.

foolip (approved):

PR #7198, opened 2020-10-31
Remove redundant Safari note from api.Location.origin

vinyldarkscratch:
This PR removes a redundant note from `api.Location.origin` from Safari.  The note talks about how https://developer.apple.com/documentation/webkitjs does not have an accurate version number, so the real version cannot be determined reliably.  However, **https://developer.apple.com/documentation/webkitjs is known bad resource as documented on the [Matching web features to browser release version numbers](https://developer.mozilla.org/docs/MDN/Contribute/Processes/Matching_features_to_browser_version) page.**  Furthermore, especially with the mdn-bcd-collector project, we have other ways than reading documentation to obtain the version number.

### DIFF
--- a/api/Location.json
+++ b/api/Location.json
@@ -367,8 +367,7 @@
               "version_added": null
             },
             "safari": {
-              "version_added": true,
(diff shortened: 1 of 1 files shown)


vinyldarkscratch:
Actually, I'll combine this with another PR to update this data.

PR #4209, opened 2019-05-24
Update compatibility table for location.origin

jurca:
This PR adds compatibility information for `location.origin` for the following browsers:
- Opera desktop
- Safari desktop
- Safari iOS

I was able to track down these implementation issues for Safari (but I am not sure how to determine the actual version of Safari that shipped the feature from these):
- https://bugs.webkit.org/show_bug.cgi?id=46558
- https://bugs.webkit.org/show_bug.cgi?id=151614

I verified the compatibility by hand using browserstack.com. Since the feature is supported by the oldest Safari desktop and Opera desktop versions that browserstack offers, I am unable to determine the exact versions from which this feature has been shipped.

PR #7202, opened 2020-10-31
Update Opera data for Location API

vinyldarkscratch:
This PR updates the Opera data for the Location API using results from the mdn-bcd-collector project.

### DIFF
--- a/api/Location.json
+++ b/api/Location.json
@@ -166,10 +166,10 @@
               "version_added": true
             },
             "opera": {
-              "version_added": true
+              "version_added": "≤12.1"
             },
             "opera_android": {
-              "version_added": true
+              "version_added": "≤12.1"
             },
             "safari": {
               "version_added": true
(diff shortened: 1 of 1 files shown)


api/Location.json
```diff
@@ -361,10 +361,10 @@
               "notes": "Intranet sites are set to Compatibility View, which will emulate IE7 and omit <code>window.location.origin</code>."
             },
             "opera": {
-              "version_added": "10"
```
foolip, line 364: This came from https://github.com/mdn/browser-compat-data/pull/4209 and looks like it was based on real testing. http://mdn-bcd-collector.appspot.com/tests/api/Location/origin does look good, but can you dig into this and see if there's something to the Opera 10 claim?

vinyldarkscratch: I'm not sure how they obtained Opera 10 in their results; it is `undefined` in both Opera 10.6 and 12.1.  My guess is that they must have performed a test while on a page that implemented a polyfill.

foolip: That might be. I'll leave a comment on that PR just in case, but thanks for testing further!

foolip (approved):

PR #7204, opened 2020-10-31
Update Chrome data for Element API

vinyldarkscratch:
This PR updates and/or corrects the Chrome data for the Element API based upon results from the mdn-bcd-collector project (using results from Chrome 1-86; Opera 12.14, 15, and 70; and Edge 13-85).  The updates include mirroring of data and notes to Chrome-derived browsers.

Helpful note: I noticed that a number of data was set to "Chrome 43"; however, from my testing, I've found that Chrome 43 was a magic number of when all properties are initialized on the object prototype. Thus, it's best to trust results that indicate Chrome 43 less.


api/Element.json
```diff
@@ -1167,10 +1167,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Element/attributes",
           "support": {
             "chrome": {
-              "version_added": "26"
+              "version_added": "1"
```
foolip, line 1170: For anything on `Element` where the supported version is being lowered, can you confirm it was always on `Element` and not just `HTMLElement`? I would expect a few cases of moves between `HTMLElement` and `Element`, some of which will require `partial_implementation` and notes like the ones added in https://github.com/mdn/browser-compat-data/pull/6865

vinyldarkscratch: Good call -- it looks like there were only two adjusted features that were initially implemented on the `HTMLElement` API.  https://github.com/foolip/mdn-bcd-collector/pull/699 will fix this so we won't get conflicting results.

api/Element.json
```diff
@@ -1674,12 +1674,28 @@
...
-            },
+            "chrome": [
+              {
+                "version_added": "22"
+              },
+              {
+                "version_added": "1",
+                "version_removed": "22",
+                "partial_implementation": true,
+                "notes": "This function is implemented in the <code>HTMLElement</code> API, meaning non-HTML elements (like SVG elements) cannot use this function."
```
foolip, line 1685:
```suggestion
                "notes": "Only supported for <a href='https://developer.mozilla.org/docs/Web/API/HTMLElement'><code>HTMLElement</code></a>, not all <code>Element</code> objects, such as <a href='https://developer.mozilla.org/docs/Web/API/SVGElement'><code>SVGElement</code></a>."
```

Suggesting something closer to the wording used in https://github.com/mdn/browser-compat-data/pull/6865. (Also note that this is an attribute, not a function.)

Having written that out, I'd also suggest testing if this was ever added to `SVGElement` specifically. A bunch of things were at one point on `HTMLElement` + `SVGElement` and then moved up to `Element`, which would make the note wrong

api/Element.json
```diff
@@ -1674,12 +1674,28 @@
...
+            ],
+            "chrome_android": [
+              {
+                "version_added": "25"
+              },
+              {
+                "version_added": "18",
+                "version_removed": "25",
+                "partial_implementation": true,
+                "notes": "This function is implemented in the <code>HTMLElement</code> API, meaning non-HTML elements (like SVG elements) cannot use this function."
```
foolip, line 1696: This and many following notes still need to be updated to match the above note: "Only supported for <a href='https://developer.mozilla.org/docs/Web/API/HTMLElement'><code>HTMLElement</code></a>, not all <code>Element</code> objects, such as <a href='https://developer.mozilla.org/docs/Web/API/SVGElement'><code>SVGElement</code></a>."

api/Element.json
```diff
@@ -6877,16 +6945,22 @@
...
             ],
-            "edge": {
-              "version_added": "13"
-            },
+            "edge": [
+              {
+                "version_added": "13"
+              },
+              {
+                "version_added": "79",
```
foolip, line 6960: Are you sure about this? If tests `webkitRequestPointerLock` are being used to infer this, can the other prefix support statements also be updated?

api/Element.json
```diff
@@ -7005,10 +7091,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Element/scroll",
           "support": {
             "chrome": {
-              "version_added": "45"
+              "version_added": "61"
```
foolip, line 7094: Based on http://mdn-bcd-collector.appspot.com/tests/api/Element/scroll this change could be because `scroll` was initially on `HTMLElement`. But I confirmed with https://software.hixie.ch/utilities/js/live-dom-viewer/saved/8714 that it was added in Chrome 61.

api/Element.json
```diff
@@ -7311,14 +7397,18 @@
...
-              "version_added": "29"
+              "version_added": "18"
             },
             "edge": [
+              {
+                "version_added": "79"
+              },
               {
                 "version_added": "17",
+                "version_removed": "79",
```
foolip, line 7411: Should we add `partial_implemenation` here?

api/Element.json
```diff
@@ -8375,13 +8466,13 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Element/tabStop",
           "support": {
             "chrome": {
-              "version_added": "43"
+              "version_added": false
```
foolip, line 8469: Looks like https://developer.mozilla.org/en-US/docs/Web/API/Element/tabStop was never shipped. We should remove this entry, but not in this PR.

api/Element.json
vinyldarkscratch: This one was a manual update, since the collector doesn't have prefix support enabled.

foolip: Ah, thanks!

foolip:
@vinyldarkscratch there are conflicts that need to be resolved now, can you take a look?

api/Element.json
```diff
@@ -5133,7 +5200,7 @@
               "notes": "This API was previously available on the <code>Node</code> API."
             },
             "webview_android": {
-              "version_added": true,
+              "version_added": "4.4.3",
```
foolip, line 5203: 4.4.3 is being set for `localName`, `namespaceURI`, and `prefix`, the 3 attributes in https://github.com/mdn/browser-compat-data/issues/6545. This is the result of mirroring, and looks correct. We'll change these entries again later, but that's OK.

api/Element.json
```diff
@@ -4194,7 +4226,7 @@
               {
                 "version_added": "12",
                 "version_removed": "16",
-                "notes": "This function is implemented in the <code>HTMLElement</code> API, meaning non-HTML elements (like SVG elements) cannot use this function."
+                "notes": "Only supported for <a href='https://developer.mozilla.org/docs/Web/API/HTMLElement'><code>HTMLElement</code></a>, not all <code>Element</code> objects, such as <a href='https://developer.mozilla.org/docs/Web/API/SVGElement'><code>SVGElement</code></a>."
```
foolip, line 4229: Good, more consistent notes. The old notes weren't wrong because this (`getElementsByClassName`) really is a function, but the new one works for both functions and properties.

foolip (approved):

PR #7205, opened 2020-10-31
Update Firefox data for Element API

vinyldarkscratch:
This PR updates and/or corrects the Firefox data for the Element API based upon results from the mdn-bcd-collector project (using results from Firefox 1-82). The updates include mirroring of data and notes to Firefox Android, as well as updating a few notes for accuracy.

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -6894,15 +6894,9 @@
                 "prefix": "moz"
               }
             ],
-            "firefox_android": [
-              {
-                "version_added": "50"
-              },
-              {
-                "version_added": true,
-                "prefix": "moz"
-              }
-            ],
+            "firefox_android": {
+              "version_added": true
+            },
             "ie": {
               "version_added": false
             },
--- a/api/Element.json
+++ b/api/Element.json
@@ -1176,10 +1176,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "22"
+              "version_added": "1"
             },
             "firefox_android": {
-              "version_added": "22"
+              "version_added": "4"
             },
             "ie": {
               "version_added": "5.5"
@@ -1224,10 +1224,10 @@
               "version_added": "≤79"
             },
             "firefox": {
-              "version_added": null
+              "version_added": false
             },
             "firefox_android": {
-              "version_added": null
+              "version_added": false
             },
             "ie": {
               "version_added": false
@@ -3647,10 +3647,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "23"
+              "version_added": "1"
             },
             "firefox_android": {
-              "version_added": "23"
+              "version_added": "4"
             },
             "ie": {
               "version_added": "8"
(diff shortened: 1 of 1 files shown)


api/Element.json
```diff
@@ -4339,18 +4340,15 @@
...
                 "Prior to Firefox 19, this method was returning a <code>NodeList</code>; it was then changed to reflects the spec change."
               ]
             },
             "firefox_android": {
-              "version_added": true,
-              "notes": [
-                "The behavior of <code>element.getElementsByTagNameNS</code> changed between Firefox 3.5 and Firefox 3.6. In Firefox 3.5 and before, this function would automatically case-fold any queries so that a search for \"foo\" would match \"Foo\" or \"foo\". In Firefox 3.6 and later this function is now case-sensitive so that a query for \"foo\" will only match \"foo\" and not \"Foo\". For more background on this, please see the <a href='https://bugzil.la/542185#c5'>comment from Henri Sivonen about the change</a>. You can also look at the <a href='https://developer.mozilla.org/docs/Case_Sensitivity_in_class_and_id_Names'>relevant part of the standard, which states which parts of the API are case-sensitive and which parts aren't.</a>",
```
foolip, line 4351: Yep, this note isn't needed for Firefox for Android.

api/Element.json
```diff
@@ -4545,11 +4539,12 @@
               "version_added": "16"
             },
             "firefox": {
-              "version_added": true,
-              "notes": "[1] Before Firefox 35, it was implemented on the <code>Node</code> interface."
+              "version_added": "1",
+              "notes": "Before Firefox 35, it was implemented on the <code>Node</code> interface."
```
foolip, line 4543: Since `Element` inherits from `Node`, this note strikes me as not useful. But I'll not suggest changing it here. There are lots of notes I'd like to see changed all over, better keep the projects separate :)

api/Element.json
```diff
@@ -5734,7 +5729,8 @@
               "notes": "This API was previously available on the <code>Node</code> API."
             },
             "firefox_android": {
-              "version_added": true
+              "version_added": "48",
+              "notes": "This API was previously available on the <code>Node</code> API."
```
foolip, line 5733: This is a mirrored note, so OK :)

api/Element.json
```diff
@@ -6893,9 +6894,15 @@
...
-            "firefox_android": {
-              "version_added": true
-            },
+            "firefox_android": [
+              {
+                "version_added": "50"
+              },
+              {
+                "version_added": true,
+                "prefix": "moz"
```
foolip, line 6903: The data being mirrored is incorrect, this API has already been removed:
https://github.com/mozilla/gecko-dev/commit/6a9f5ff102c0f9b72dae14490608c22305a276ca

Can you add these APIs to the collector to find out when it exactly it was removed?

vinyldarkscratch: Not sure why I mirrored this one when the original desktop data wasn't being updated.  I'll undo these changes and add the missing features for a follow-up PR later.

foolip: Thanks!

foolip (approved):

PR #7206, opened 2020-10-31
Update IE and EdgeHTML data for Element API

vinyldarkscratch:
This PR updates and/or corrects the Internet Explorer and EdgeHTML data for the Element API based upon results from the mdn-bcd-collector project (using results from IE 5.5 to 11 with manual testing in IE 5, and Edge 13-79).

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -1690,7 +1690,7 @@
               "version_added": "4"
             },
             "ie": {
-              "version_added": "8"
+              "version_added": "5"
             },
             "opera": {
               "version_added": "8"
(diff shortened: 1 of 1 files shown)


api/Element.json
```diff
@@ -4353,7 +4353,7 @@
               ]
             },
             "ie": {
-              "version_added": "5.5"
+              "version_added": "9"
```
foolip, line 4356: Makes sense that various *NS methods would be added together in IE9, when they were needed for SVG. Previously the `getElementsByTagNameNS` data matched `getElementsByTagName`, and was probably just copied at some point in the past.

foolip (approved):

PR #7210, opened 2020-10-31
Drop the “MediaQueryListListener” feature

sideshowbarker:
This change drops the “MediaQueryListListener” feature. It’s not the name of an actual API but instead apparently was just a kind of shorthand for something that’s no longer part of any spec anyway. See https://wiki.developer.mozilla.org/docs/Web/API/MediaQueryListListener


foolip:
@ddbeck can you redirect https://wiki.developer.mozilla.org/en-US/docs/Web/API/MediaQueryListListener before we merge this?

sideshowbarker:
> @ddbeck can you redirect https://wiki.developer.mozilla.org/en-US/docs/Web/API/MediaQueryListListener before we merge this?

I can redirect it — so can you actually, I think: In the top editing bar of the MDN editing UI, you should see that the furthest-right button is a curved-arrow icon with **Create a Redirect** tooltip text. When you push push that it pops up a dialog that lets you choose the article name to redirect to.

foolip:
Aha, I thought special powers were needed. I've redirected it now, but https://developer.mozilla.org/docs/Web/API/MediaQueryListListener hasn't updated yet.

foolip:
Going through some of the docs for this, I think the interesting thing to communicate to web developers is whether `MediaQueryList` supports `addListener`, `addEventListener`, or both. How the callback function for `addListener` was specified is not in itself observable to web developers.

What gives me doubt is what to do with these notes still left in MDN:
https://wiki.developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/addListener#Syntax
https://wiki.developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/removeListener#Syntax

Is there anything about those that should be preserved?

sideshowbarker:
> Going through some of the docs for this, I think the interesting thing to communicate to web developers is whether `MediaQueryList` supports `addListener`, `addEventListener`, or both.

Yeah, agreed

> How the callback function for `addListener` was specified is not in itself observable to web developers.
> 
> What gives me doubt is what to do with these notes still left in MDN:
> https://wiki.developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/addListener#Syntax
> https://wiki.developer.mozilla.org/en-US/docs/Web/API/MediaQueryList/removeListener#Syntax
> 
> Is there anything about those that should be preserved?

No, I think — nothing there that needs to be preserved (given as you noted that how the callback function is specified isn’t observable).

ddbeck:
So what needs to happen here? Remove the mentions of `MediaQueryListListener` on MDN and merge this?

sideshowbarker:
> So what needs to happen here? Remove the mentions of `MediaQueryListListener` on MDN and merge this?

I think so — but I guess we need confirmation from @foolip

vinyldarkscratch:
Foolip's been away from BCD for a while -- reading through this thread, I'd say that this is ready to go.  If those notes don't need to be preserved, then we can push go and remove this!

sideshowbarker:
> I'd say that this is ready to go. If those notes don't need to be preserved, then we can push go and remove this!

OK, thanks — then I’m happy if you (or anybody other than me, since I’m the author for the patch) can go ahead and push the button to merge this

sideshowbarker:
Thanks for the merge!

ddbeck:
So it's not exactly explicit from the process [we discussed previously](https://github.com/mdn/browser-compat-data/pull/6929#issuecomment-717981716), but I'd like to make sure that we clean up any content messes that follow from our feature removals before merging. By redirecting the `MediaQueryListListener` page back to `MediaQueryList`, we created oddities on the pages for `MediaQueryList.addListener()` and `.removeListener()`. I'll fix those pages now, but, in the future, I'd like open questions about content problems to be answered (with content changes or issues against mdn/sprints) before merging.

PR #7211, opened 2020-10-31
Fix release statuses for NodeJS

vinyldarkscratch:
This PR updates the release status information for the NodeJS releases based upon the calendar on https://nodejs.org/en/about/releases/ by doing the following:
- Marks the older of two v12 versions labeled ESR as "retired"
- Marks v13 as "retired" (previously "current")
- Marks v14 as "esr" (previously "retired")

### DIFF
--- a/browsers/nodejs.json
+++ b/browsers/nodejs.json
@@ -223,7 +223,7 @@
         "12.11.0": {
           "release_date": "2019-09-25",
           "release_notes": "https://nodejs.org/en/blog/release/v12.11.0/",
-          "status": "esr",
+          "status": "retired",
           "engine": "V8",
           "engine_version": "7.7"
         },
@@ -244,7 +244,7 @@
         "13.2.0": {
           "release_date": "2019-11-21",
           "release_notes": "https://nodejs.org/en/blog/release/v13.2.0/",
-          "status": "current",
+          "status": "retired",
           "engine": "V8",
           "engine_version": "7.9"
         },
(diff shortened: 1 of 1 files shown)


sideshowbarker (approved):

vinyldarkscratch:
Did you mean to merge this as well, @sideshowbarker?

sideshowbarker:
> Did you mean to merge this as well, @sideshowbarker?

Yeah but then I recalled that https://github.com/mdn/browser-compat-data/blob/master/GOVERNANCE.md#owners says:

> Manage and merge non-data pull requests such as schema, linter, or infrastructure changes.

…is an Owners task, and https://github.com/mdn/browser-compat-data/blob/master/GOVERNANCE.md#peers says Peers:

> May only merge other people's pull requests that relate to compat data updates.

So I guess strictly speaking, the policy is that all changes to the browser schemas can only be merged by Owners?

(I know we have merged such changes ourselves in the past, but if that’s actually OK, I guess we should take time to update the governance doc to make that clear.)

vinyldarkscratch:
Ah, I see where the confusion is (it's something that got me too).  PRs like this are considered browser _data_, not schema -- schema would be the data structure, in other words all the files in `schemas/`.  So, we should probably update the governance doc a little bit to help clarify?

vinyldarkscratch:
Closing this as it's superseded by #7212.

sideshowbarker:
> Ah, I see where the confusion is (it's something that got me too). PRs like this are considered browser _data_, not schema -- schema would be the data structure, in other words all the files in `schemas/`

Aha — yeah, of course, that makes sense to me now ⋯ Thanks for clarifying!



> So, we should probably update the governance doc a little bit to help clarify?

Yeah I think so — so I’ve raised https://github.com/mdn/browser-compat-data/pull/7215

PR #7212, opened 2020-10-31
Add Node.js 14.3/14.8 and fix Node release statuses

saschanaz:
- [x] Summarize your changes
   * https://nodejs.org/en/blog/release/v14.3.0/
   * https://nodejs.org/en/blog/release/v14.8.0/
   * `node -p process.versions.v8`
- [x] Link to related issues or pull requests, if any: #7209

### DIFF
--- a/browsers/nodejs.json
+++ b/browsers/nodejs.json
@@ -272,7 +272,7 @@
         "14.8.0": {
           "release_date": "2020-08-11",
           "release_notes": "https://nodejs.org/en/blog/release/v14.8.0/",
-          "status": "retired",
+          "status": "esr",
           "engine": "V8",
           "engine_version": "8.4"
         },
--- a/browsers/nodejs.json
+++ b/browsers/nodejs.json
@@ -223,7 +223,7 @@
         "12.11.0": {
           "release_date": "2019-09-25",
           "release_notes": "https://nodejs.org/en/blog/release/v12.11.0/",
-          "status": "esr",
+          "status": "retired",
           "engine": "V8",
           "engine_version": "7.7"
         },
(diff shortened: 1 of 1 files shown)


sideshowbarker (approved):

browsers/nodejs.json
```diff
@@ -255,13 +255,27 @@
...
           "release_date": "2020-07-21",
           "release_notes": "https://nodejs.org/en/blog/release/v14.6.0/",
           "status": "retired",
           "engine": "V8",
           "engine_version": "8.4"
         },
+        "14.8.0": {
+          "release_date": "2020-08-11",
+          "release_notes": "https://nodejs.org/en/blog/release/v14.8.0/",
+          "status": "retired",
```
vinyldarkscratch, line 275:
```suggestion
          "status": "esr",
```

vinyldarkscratch (changes requested):
Thanks for cherry-picking these changes!  Since NodeJS v14 is still in Active LTS, can we mark the most recent v14 version as "esr"?

(Note: I updated the title of this PR since it includes fixes for other NodeJS version statuses.)

saschanaz:
14.8 is not most recent, 14.15 is. But then we probably don't want to add every version that we'll never use? 🤔

browsers/nodejs.json
saschanaz: 14.8 is not most recent, 14.15 is. But then we probably don't want to add every version that we'll never use? What do you think?

vinyldarkscratch: Yeah, I've generally stuck with the lowest versions for each column on https://node.green/ with some exceptions (such as major versions or support differences undocumented on node.green).

vinyldarkscratch (approved):
Thanks, looking good to me!

PR #7214, opened 2020-11-01
Update Chrome + Opera data for Document API

vinyldarkscratch:
This PR updates and/or corrects the Chrome and Opera data for the Element API based upon results from the mdn-bcd-collector project (using results from Chrome 1-86; Opera 12.14, 15, and 70; and Edge 13-85). The updates include mirroring of data and notes to Chrome-derived browsers.

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -8177,24 +8177,12 @@
         "__compat": {
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Document/onvisibilitychange",
           "support": {
-            "chrome": [
-              {
-                "version_added": "62"
-              },
-              {
-                "version_added": "13",
-                "prefix": "webkit"
-              }
-            ],
-            "chrome_android": [
-              {
-                "version_added": "62"
-              },
-              {
-                "version_added": "18",
-                "prefix": "webkit"
-              }
-            ],
+            "chrome": {
+              "version_added": "62"
+            },
+            "chrome_android": {
+              "version_added": "62"
+            },
             "edge": {
               "version_added": "18",
               "notes": "Before Edge 18, this event handler attribute was not supported; however, the event itself was supported since Edge 12. The event can be listened to via <code>document.addEventListener('visibilitychange', function() {});</code>."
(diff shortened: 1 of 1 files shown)


api/Document.json
```diff
@@ -1200,7 +1200,7 @@
             ],
             "opera": [
               {
-                "version_added": true
+                "version_added": "≤12.1"
```
foolip, line 1203: This is the only Presto change in this PR. Accidentally included?

api/Document.json
```diff
@@ -8171,16 +8179,22 @@
           "support": {
             "chrome": [
               {
-                "version_added": "33"
+                "version_added": "62"
```
foolip, line 8182: 33 vs. 62 is a big difference. Can you check where 33 came from?

api/Document.json
vinyldarkscratch: This is the only Presto change for this API from the collector, I figured it made sense to include it in this PR.  I've updated the title and description to assist.

api/Document.json
vinyldarkscratch: The "33" had come from the initial wiki migration.  I did a quick test to confirm the results:

```js
	console.log("Page visibility changed!");
}
```

foolip: Thanks for double checking!

api/Document.json
```diff
@@ -8171,16 +8179,22 @@
...
-            "chrome_android": {
-              "version_added": "33"
-            },
+            "chrome_android": [
+              {
+                "version_added": "62"
+              },
+              {
+                "version_added": "18",
+                "prefix": "webkit"
```
foolip, line 8195: I now see the data being mirrored here is wrong. `onwebkitvisibilitychange` isn't in current Chrome. Can you add a custom test for it in the collector and confirm `version_added` + add right `version_removed`?

I also note that it's not a prefix, since the name of this isn't `webkitonvisibilitychange`, so switching to `alternative_name` would be good. But maybe as a part of a mass change PR for all `onfoo` event handler attributes involving prefixes.

vinyldarkscratch: It turns out that `onwebkitvisibilitychange` wasn't supported in Chrome; testing in Chrome 15 revealed no support.  The original data came from wiki mirroring, so I don't trust the data much.  Reverting the mirroring and removing the data.

foolip: Yay, thanks!

api/Document.json
```diff
@@ -8169,17 +8177,11 @@
...
-              {
-                "version_added": "33"
-              },
-              {
-                "version_added": "13",
-                "prefix": "webkit"
-              }
-            ],
+            "chrome": {
+              "version_added": "62"
```
foolip, line 8181: OK, so one more thing to sort out with these `onfoo` event handler attributes. We also have `foo_event` entries, and `visibilitychange_event` still says 33, which was changed here. Can you update that entry too? It's fully possible that the event itself was supported earlier than the event handler attribute, and that there was a prefixed event.

There are also `afterscriptexecute_event` or `beforecriptexecute_event` entries in `Element`.

Might I suggest actually just handling these `onfoo` things separately, backing them out from this PR?

foolip (commented):
More complications :)

api/Document.json
vinyldarkscratch: It looks like the event itself was indeed supported earlier; the data  for `visibilitychange_event` is correct.  Although, it does need some mirroring -- I'll cherry-pick these changes and do it altogether in its own PR.

foolip: Thanks!

foolip (approved):

PR #7221, opened 2020-11-01
Mark DeviceMotionEventAcceleration and DeviceMotionEventRotationRate unsupported in Safari

sideshowbarker:
This change marks the `DeviceMotionEventAcceleration` and `DeviceMotionEventRotationRate` interfaces unsupported in Safari and iOS Safari. Test: https://wpt.live/orientation-event/idlharness.https.window.html

### DIFF
--- a/api/DeviceMotionEventAcceleration.json
+++ b/api/DeviceMotionEventAcceleration.json
@@ -37,10 +37,14 @@
             "version_added": null
           },
           "safari": {
-            "version_added": null
+            "version_added": "5.1",
+            "partial_implementation": true,
+            "notes": "Doesn't expose the <code>DeviceMotionEventAcceleration</code> interface itself."
           },
           "safari_ios": {
-            "version_added": null
+            "version_added": "5",
+            "partial_implementation": true,
+            "notes": "Doesn't expose the <code>DeviceMotionEventAcceleration</code> interface itself."
           },
           "samsunginternet_android": [
             {
@@ -98,10 +102,14 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": "5.1",
+              "partial_implementation": true,
+              "notes": "Doesn't expose the <code>DeviceMotionEventAcceleration</code> interface itself."
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "5",
+              "partial_implementation": true,
+              "notes": "Doesn't expose the <code>DeviceMotionEventAcceleration</code> interface itself."
             },
             "samsunginternet_android": {
               "version_added": "1.5"
@@ -147,10 +155,14 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": "5.1",
+              "partial_implementation": true,
+              "notes": "Doesn't expose the <code>DeviceMotionEventAcceleration</code> interface itself."
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "5",
+              "partial_implementation": true,
+              "notes": "Doesn't expose the <code>DeviceMotionEventAcceleration</code> interface itself."
             },
             "samsunginternet_android": {
               "version_added": "1.5"
@@ -196,10 +208,14 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": "5.1",
+              "partial_implementation": true,
+              "notes": "Doesn't expose the <code>DeviceMotionEventAcceleration</code> interface itself."
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "5",
+              "partial_implementation": true,
+              "notes": "Doesn't expose the <code>DeviceMotionEventAcceleration</code> interface itself."
             },
             "samsunginternet_android": {
               "version_added": "1.5"
@@ -245,10 +261,14 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": "5.1",
+              "partial_implementation": true,
+              "notes": "Doesn't expose the <code>DeviceMotionEventAcceleration</code> interface itself."
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "5",
+              "partial_implementation": true,
+              "notes": "Doesn't expose the <code>DeviceMotionEventAcceleration</code> interface itself."
             },
             "samsunginternet_android": {
               "version_added": "1.5"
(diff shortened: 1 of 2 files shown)


foolip:
Running http://mdn-bcd-collector.appspot.com/tests/api/DeviceMotionEvent in Safari iOS it looks like both `acceleration` and `rotationRate` attributes are supported.

foolip:
I guess this is a case of the interface objects not being exposed?

sideshowbarker:
> I guess this is a case of the interface objects not being exposed?

The interface definitions are not even in the IDL in the WebKit sources, as far as can see.

foolip:
Looks like dictionaries are used instead: https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/dom/DeviceMotionEvent.idl?rev=267813

The difference will rarely matter to a web developer, but I'd probably call this partial support. In practice it's similar to if it was an interface but with the interface object not exposed.

sideshowbarker:
> Looks like dictionaries are used instead: https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/dom/DeviceMotionEvent.idl?rev=267813

Yeah

> The difference will rarely matter to a web developer, but I'd probably call this partial support.

So mark the DeviceMotionEventAcceleration and DeviceMotionEventRotationRate interfaces a partially supported?

> In practice it's similar to if it was an interface but with the interface object not exposed.

Yeah I don’t understand why they were specified as exposed objects to begin with — but given that the spec as currently written explicitly requires them to be exposed as objects, it’d seem odd to me for us to consider the implementation in conformance to the spec.

And anyway it seems like that’s all completely separate from how we do the data for `acceleration` and `rotationRate` attributes for `DeviceMotionEvent`.

foolip:
> Yeah I don’t understand why they were specified as exposed objects to begin with

Two things put together result in this somewhat silly situation:
- https://heycam.github.io/webidl/#dfn-dictionary says "Dictionaries must not be used as the type of an attribute or constant" (Because a new object would be returned every time, meaning `event.rotation === event.rotation` doesn't hold.)
- https://heycam.github.io/webidl/#LegacyNoInterfaceObject is legacy

> And anyway it seems like that’s all completely separate from how we do the data for the `acceleration` and `rotationRate` attributes for `DeviceMotionEvent`, right?

I don't think so, because how does a developer or tooling figure out if `window.addEventListener('devicemotion', event => doStuffWith(event.acceleration.x))` is problematic or not? It will work just fine, it really doesn't matter what kind of object `event.acceleration` is, and we don't have anything other than the standard interfaces in BCD to express this.

sideshowbarker:
> I don't think so, because how does a developer or tooling figure out if `window.addEventListener('devicemotion', event => doStuffWith(event.acceleration.x))` is problematic or not? It will work just fine, it really doesn't matter what kind of object `event.acceleration` is, and we don't have anything other than the standard interfaces in BCD to express this.

OK, yeah, that’s a good point.

So what concrete change should we make here? Mark the interfaces as partially supported with a note that says what?

foolip:
Yeah, I think the `api.DeviceMotionEventAccelerationExperimental` entry itself (but not its children) can get `partial_implementation` with a note that says "The &lt;code>DeviceMotionEventAcceleration&lt;/code> interface itself is not exposed."

Searching for "exposed" and "interface" I don't find any analogous situation to copy from, which surprises me, so I'd like to hear from @ddbeck if there's a better path. It feels to me like we're forced into giving web developers almost useless information in order to be able to record that the non-exposure of DeviceMotionEventAcceleration doesn't mean that we maintainers should mark its subfeatures as unsupported.

saschanaz:
FYI, filed https://github.com/w3c/deviceorientation/issues/89 as dictionaries look like a valid way to implement them.

ddbeck:
> I think the `api.DeviceMotionEventAcceleration` entry itself (but not its children) can get `partial_implementation` with a note that says "The <code>DeviceMotionEventAcceleration</code> interface itself is not exposed."

This sounds like a reasonable way to go to me. I can't think of anything else like it to compare to.

> Searching for "exposed" and "interface" I don't find any analogous situation to copy from, which surprises me, so I'd like to hear from @ddbeck if there's a better path. It feels to me like we're forced into giving web developers almost useless information in order to be able to record that the non-exposure of DeviceMotionEventAcceleration doesn't mean that we maintainers should mark its subfeatures as unsupported.

Almost, but not quite completely useless in this case, right? If I read the spec, I'd expect the interface to be exposed (and maybe I could use `instanceof DeviceMotionEventAcceleration` somewhere?). As long as we have a thin justification, I think we're OK. With so many specs and so many browsers, there are bound to be some oddities. I don't think we have to hide that fact from web devs.

But if that's dubious, then maybe we should consider adding some sort of internal-only-to-BCD comments field, so we can record this sort of thing without cluttering the public notes.

foolip:
I think `x instanceof DeviceMotionEventAcceleration` is a very thin justification, because what you'd have to do is something like this:
```js
window.addEventListener('devicemotion', (event) => {
  objList.push(event.acceleration);
  objList.push(event.rotationRate);
});

function somewhereElseLater() {
  for (const obj of objList) {
    if (obj instanceof DeviceMotionEventAcceleration) {
      doCoolStuff(obj);
    }
  }
}
```

In other words, you'd have to throw the object together with other objects and later have a need to distinguish them based on type testing.

It's not impossible, though, so a note about the interface object not being exposed would be OK I think.

ddbeck:
OK, great. @sideshowbarker if you'd add the note, then I think we can merge this.

sideshowbarker:
> OK, great. @sideshowbarker if you'd add the note, then I think we can merge this.

Notes added now, along with marking as `partial_implementation`.

Also, because marking these as `partial_implementation` rather than false means we need to specify actual version numbers too, I’ve also added those.

Since https://trac.webkit.org/changeset/68197/webkit is where the support was added for the members of these interfaces, per https://trac.webkit.org/browser/webkit/releases/Apple/Safari%205.1/WebCore/dom/DeviceMotionEvent.idl that support shipped in Safari 5.1 and in iOS Safari 5 per https://trac.webkit.org/browser/webkit/tags/Safari-534.46/Source/WebCore/dom/DeviceMotionEvent.idl (and the fact the iOS Safari version number is lower than the desktop Safari version number is not a mistake; instead that the order in which those versions ended up being released, and I vaguely recall running into this before, and doublechecking then also to make sure it was right).

vinyldarkscratch (approved):
Looks good to me, thanks!

PR #7222, opened 2020-11-01
Add linter for release status

saschanaz:
#7212 shows statuses can be messed up, so this adds a linter for that.


vinyldarkscratch:
This PR's a small subset of what's in #6167, so if it's alright, I'd like to close this and get the other PR merged.

saschanaz:
Ah cool, let's do this there.

PR #7224, opened 2020-11-02
Mark Basic Card features unsupported in Safari

sideshowbarker:
This change marks the `BasicCardRequest` and `BasicCardResponse` dictionaries as unsupported in Safari and iOS Safari. Confirmed through manual testing and inspection of WebKit source code.


foolip (approved):
Confirmed no "BasicCard" in WebKit source.

PR #7225, opened 2020-11-02
Mark FontFaceSetLoadEvent unsupported in Safari

sideshowbarker:
This change marks the `FontFaceSetLoadEvent` interface unsupported in Safari and iOS Safari.
Test: https://wpt.live/css/css-font-loading/idlharness.https.html

### DIFF
--- a/api/FontFaceSetLoadEvent.json
+++ b/api/FontFaceSetLoadEvent.json
@@ -29,10 +29,10 @@
             "version_added": "22"
           },
           "safari": {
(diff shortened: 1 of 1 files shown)


foolip (approved):
This is accurate, there is no `FontFaceSetLoadEvent` in WebKit. However, other related bits are exposed. I'll file an issue.

foolip:
I've filed https://github.com/mdn/browser-compat-data/issues/7231.

PR #7226, opened 2020-11-02
Mark Presentation API features unsupported in Safari

sideshowbarker:
This change marks the `Presentation`, `PresentationAvailability`, `PresentationConnection`, `PresentationConnectionAvailableEvent`, `PresentationConnectionCloseEvent`, `PresentationConnectionList`, `PresentationReceiver`, and `PresentationRequest` interfaces as unsupported in Safari and iOS Safari.

Test: https://wpt.live/presentation-api/controlling-ua/idlharness.https.html

### DIFF
--- a/api/Presentation.json
+++ b/api/Presentation.json
@@ -43,10 +43,10 @@
             "version_added": true
           },
           "safari": {
-            "version_added": null
+            "version_added": false
           },
           "safari_ios": {
-            "version_added": null
+            "version_added": false
           },
           "samsunginternet_android": {
             "version_added": "5.0"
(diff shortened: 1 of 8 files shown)


foolip (approved):
I've also confirmed no relevant matches for `find * -iname '*presentation*'` in WebKit source.

foolip:
However, there is a prefixed API available in Safari:
https://developer.apple.com/documentation/webkitjs/htmlvideoelement

AFAICT that has to do with how the video is presented on the local device, not about presenting on other screens.

foolip:
I've sent https://github.com/foolip/mdn-bcd-collector/pull/697 to cover those APIs and make sure we eventually make a decision about whether to include in BCD or not.

PR #7233, opened 2020-11-02
Drop reference to wpt.fyi results for api.Animation.commitStyles

foolip:
The test results are for Safari on macOS. But more importantly, the
existence of failing tests doesn't in itself indicate that there's a
problem that matters to web developers. To make that claim, we would
have to understand what's causing the failures, and whether the thing
they're testing is something that's likely to matter.

This came from https://github.com/mdn/browser-compat-data/pull/5963.

### DIFF
--- a/api/Animation.json
+++ b/api/Animation.json
@@ -277,9 +277,7 @@
               "version_added": "13.1"
             },
             "safari_ios": {
-              "version_added": "13.4",
-              "partial_implementation": true,
-              "notes": "Passes 13/27 <a href='https://wpt.fyi/results/web-animations/interfaces/Animation/commitStyles.html?label=master&label=stable&product=safari'>web platform tests</a>."
+              "version_added": "13.4"
             },
             "samsunginternet_android": {
               "version_added": false


foolip:
@Hexcles @stephenmcgruer FYI that wpt.fyi has been linked (just once) from BCD. We have talked about ways of connecting the two, and this removal isn't a vote against doing that.

lpd-au (commented):
Hi foolip, it appears Safari 14 has made some progress on this implementation, now passing 16 of 27 tests, up from only 13. However, to my eye it unfortunately still appears to have significant shortcomings that could easily trip up a web developer (like myself). Rather than simply removing any warning, perhaps you could contribute a better note to summarise them? If I'm not mistaken, aside from incorrectly triggering mutation observers, the remaining test failures all relate to "nicely" committing "readable" rather than calculated values as well as how logical properties are handled.

sideshowbarker (approved):

foolip:
@lpd-au, yes, let's try to get some bugs filed on WebKit, and then link to those from both wpt.fyi and BCD.

@graouts, I'm guessing based on https://bugs.webkit.org/show_bug.cgi?id=186495 that you might care about these test results in WebKit. Might you be able to make sense of the Safari failures in https://wpt.fyi/results/web-animations/interfaces/Animation/commitStyles.html to see what could plausibly matter to web developers and not?

Since I don't know Web Animations much at all, I can only really make sense of the "Does NOT trigger mutation observers when the change to style is redundant" subtest, and that's one that to me seems pretty low risk, since no-change animations aren't something one does deliberately, and delivering a redundant mutation record would likely end up doing the right thing anyway.

ddbeck:
@chrisdavidmills @Elchi3 do either of you know anything about Web Animations or know someone who might ask about it? This has stalled and I think getting some help framing this note for web developers could unblock things.

chrisdavidmills:
I would've asked @graouts, but I see that he's already been asked....maybe @birtles ?

To make this as easy to answer as possible, the question is basically this — the [compat data for `Animation.commitStyles` in iOS](https://github.com/mdn/browser-compat-data/blob/master/api/Animation.json#L246) is defined as partial support in 13.4 onwards, and includes the following note:

"Passes 13/27 <a href='https://wpt.fyi/results/web-animations/interfaces/Animation/commitStyles.html?label=master&label=stable&product=safari'>web platform tests</a>."

@foolip was suggesting to remove it and the partial implementation status because it doesn't really mean anything to web developers, but @lpd-au is making the point that support is still incomplete, and we should provide a better note that explains the situation.

Can you help us understand what the support situation is, and suggest a better note wording?

Many thanks in advance.

birtles:
At a glance it looks like the failures are:

* Incorrect handling of logical properties -- fairly minor?
* Incorrect handling of custom variables -- probably due to not implementing registered properties -- Firefox also fails this for that reason
* Incorrect unit conversion / serialization for em units, line-height, transforms -- only an issue if you are actually reading back the value from style
* Incorrect handling of animations in a stack of composited animations -- pretty minor case
* Incorrect handling of redundant changes with regards to mutation observers -- edge case

So I'd suggest it's pretty useable but with the main caveat that some values might be serialized incorrectly.

For what it's worth, we plan to change the behavior of this API anyway in the near future so I suspect there will be a chance to update this again then. For now, merging this with or without a note seems fine.

chrisdavidmills:
Thanks @birtles , your time here is much appreciated.

foolip:
Thanks @birtles! I've filed https://bugs.webkit.org/show_bug.cgi?id=222096 so that your analysis can be useful for WebKit test failure triage too.

For this PR, I'm happy with my original suggestion or removing the note.

ddbeck (approved):
Great! Thank you to everyone who helped here (or at least tolerated being @ mentioned). It was a small diff with a big effort behind it. 😄

PR #7240, opened 2020-11-03
Drop all data for UC & QQ browsers, and remove them the JSON schema/docs, and from other infra bits

sideshowbarker:
This change drops all data for the UC and QQ browsers, as well as removing them from the project JSON schema files and schema docs, and from the `test/linter/test-browsers.js` and `types.d.ts` files. See https://github.com/mdn/browser-compat-data/pull/7232. Related https://github.com/mdn/browser-compat-data/issues/7238


vinyldarkscratch (approved):
I'm all for dropping these two browsers.  The hope was to add data for them since they have such a [large marketshare in China](https://gs.statcounter.com/browser-market-share/all/china/#monthly-201909-202010), but since we haven't had many contributions to the data, it's not worth maintaining.  (Besides, we can mirror data for them later.)

ddbeck:
I asked for Chris's input on this before going ahead https://github.com/mdn/browser-compat-data/issues/7238#issuecomment-721032559, so this will be blocked on that until he replies.

Also, making a note to myself here: this includes a schema change, so this will probably require a **major semver release**.

vinyldarkscratch:
@ddbeck, although this PR is altering the schema files, I'd imagine this would fall under the same level as removals of features.  It's not an adjustment of data structure, just removal of some browsers.  Perhaps a minor version makes more sense?

ddbeck (approved):
I've reviewed this change and it looks good. We're also unblocked now with the adoption of engine add/remove guidelines in #7244.

Partly a consequence of this change, I've decided that this will be a semver major release, following the new semver policy I intend to merge today (#7615).

Thank you, @sideshowbarker!

PR #7242, opened 2020-11-03
Update clipboard event data

vinyldarkscratch:
This PR updates the Chromium data for the `oncopy`, `oncut`, and `onpaste` events based upon data from the mdn-bcd-collector project.  The original data came from [Confluence](http://web-confluence.appspot.com/#!/catalog?releases=%5B%22Chrome_70.0.3538.67_OSX_10.13.6%22,%22Chrome_70.0.3538.67_Windows_10.0%22,%22Chrome_71.0.3578.80_OSX_10.14.0%22,%22Chrome_71.0.3578.80_Windows_10.0%22%5D&q=%22oncopy%22).  What had happened is that these event handlers were initially implemented on the `Element` API, and then moved down to just the `HTMLElement` API in Chrome 71.

Edit: the copy/cut/paste events in the `Element`, `Document`, and `Window` APIs also had been set to Chrome 71.  This PR corrects the data for it based upon manual testing.

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -1658,10 +1658,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Document/copy_event",
           "support": {
             "chrome": {
-              "version_added": "58"
+              "version_added": "1"
             },
             "chrome_android": {
-              "version_added": "58"
+              "version_added": "18"
             },
             "edge": {
               "version_added": "≤18"
@@ -1676,10 +1676,10 @@
               "version_added": false
             },
             "opera": {
-              "version_added": "45"
+              "version_added": "15"
             },
             "opera_android": {
-              "version_added": "43"
+              "version_added": "14"
             },
             "safari": {
               "version_added": true
@@ -1688,10 +1688,10 @@
               "version_added": null
             },
             "samsunginternet_android": {
-              "version_added": "7.0"
+              "version_added": "1.0"
             },
             "webview_android": {
-              "version_added": "58"
+              "version_added": "1"
             }
           },
           "status": {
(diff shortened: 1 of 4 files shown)


foolip:
I see we also have an `oncopy` entry on `Document` and `copy_event` entries on `Document` + `Element` + `Window`. Same for cut and paste. Can you update all of these together, so that they continue to make sense as a whole, if they in fact current do? :)

api/Document.json
```diff
@@ -1658,10 +1658,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Document/copy_event",
           "support": {
             "chrome": {
-              "version_added": "58"
+              "version_added": "1"
```
foolip, line 1661: Didn't know these events were this old, but I tested http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8654 back in Chrome 16 on macOS and the event does fire, so this seems right. The `ClipboardEvent` interface is much newer though, the event was originally a plain `Event`.

foolip (approved):

foolip:
@vinyldarkscratch thanks for the great PR descriptions you're writing/updating. Since I like a readable commit history, I've been copying them into the commit message with s/PR/commit/ and rewrapping for a while, and you can see the result for this PR in https://github.com/mdn/browser-compat-data/commit/c1f99a1ee3166e343cd509d7e205a06f890fdc46.

It would be actually be great if you wrote these commit messages directly, since the PR description is based on the commit message. Then, most of the time, just merging the PR would get the right commit message. (If there's more than one commit, I usually edit the unsightly commit message that GitHub suggests, to make it look like just the first commit.)

PR #7243, opened 2020-11-03
Correct Chrome data for HTMLIFrameElement API

vinyldarkscratch:
This PR corrects the Chrome data for the HTML `<iframe>` element API based upon results from the mdn-bcd-collector project (using results from Chrome 1-86 and Opera 12.14, 15, and 70).  Most of the data was set to "Chrome 43"; however, from my testing, I've found that Chrome 43 was a magic number of when all properties are initialized on the object prototype. (Thus, it's best not to trust results that indicate Chrome 43 as much.)

### DIFF
--- a/api/HTMLIFrameElement.json
+++ b/api/HTMLIFrameElement.json
@@ -52,10 +52,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLIFrameElement/align",
           "support": {
             "chrome": {
-              "version_added": "43"
+              "version_added": "1"
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": "18"
             },
             "edge": {
               "version_added": "12"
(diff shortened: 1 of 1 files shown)


api/HTMLIFrameElement.json
```diff
@@ -1113,10 +1113,10 @@
         "__compat": {
           "support": {
             "chrome": {
-              "version_added": "43"
+              "version_added": "20"
             },
             "chrome_android": {
-              "version_added": "18"
+              "version_added": "25"
```
foolip, line 1119: This is `srcdoc`, and this (and the mirrored data) is the only change in the PR that increases the `version_added`. This came from https://github.com/mdn/browser-compat-data/pull/1356 which just bumped the version number from "4", and "4" came from wiki migration in https://github.com/mdn/browser-compat-data/pull/2017, so I see no reason to especially trust it. `srcdoc` is not a feature that I can see being per-platform, it's deep inside HTML.

foolip (approved):

PR #7245, opened 2020-11-03
Add HTTP headers for accept-patch and accept-post

hamishwillee:
This adds _placeholder_ BCD entries for two missing HTTP headers - `Accept-Patch` and `Accept-Post`. At the moment Accept-Patch just adds `true` for items listed [here](https://www.geeksforgeeks.org/http-headers-accept-patch/) while `Accept-Post` is set to null for all browsers - but has the correct MDN link for docs.

I have searched mdn Bugzilla - no mention. FF source has a mention of Accept-Patch but not accept post. Is there anyone you could suggest I talk to for finding out support on this, or is there test code that might be used for verification? Note that even the Accept header docs just use `true`, not versions.

### DIFF
--- a/http/headers/accept-patch.json
+++ b/http/headers/accept-patch.json
@@ -1,0 +1,54 @@
+{
+  "http": {
+    "headers": {
+      "Accept-Patch": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/HTTP/Headers/Accept-Patch",
+          "support": {
+            "chrome": {
+              "version_added": true
+            },
+            "chrome_android": {
+              "version_added": true
+            },
+            "edge": {
+              "version_added": true
+            },
+            "firefox": {
+              "version_added": true
+            },
+            "firefox_android": {
+              "version_added": true
+            },
+            "ie": {
+              "version_added": true
+            },
+            "opera": {
+              "version_added": true
+            },
+            "opera_android": {
+              "version_added": true
+            },
+            "safari": {
+              "version_added": true
+            },
+            "safari_ios": {
+              "version_added": true
+            },
+            "samsunginternet_android": {
+              "version_added": true
+            },
+            "webview_android": {
+              "version_added": true
+            }
+          },
+          "status": {
+            "experimental": false,
+            "standard_track": true,
+            "deprecated": false
+          }
+        }
+      }
+    }
+  }
+}
--- a/http/headers/accept-post.json
+++ b/http/headers/accept-post.json
@@ -1,0 +1,54 @@
+{
+  "http": {
+    "headers": {
+      "Accept-Patch": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/HTTP/Headers/Accept-Post",
+          "support": {
+            "chrome": {
+              "version_added": null
+            },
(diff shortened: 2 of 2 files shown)


ddbeck:
I'm not sure this in scope for BCD. Maybe it's just because it's late in the day, but I'm having a hard time figuring out what it means for a browser to support a response header like this. If a browser makes a request and gets back an `Accept-Patch` header, what does a supporting browser do that an unsupporting browser doesn't?

hamishwillee:
@ddbeck I wasn't certain either. I posted this because the docs have a [placeholder for BCD](https://wiki.developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Post#Browser_compatibility) and you always seem to know what is relevant :-)

This is part of [content negotiation](https://wiki.developer.mozilla.org/en-US/docs/Web/HTTP/Content_negotiation); the server can't handle something in a patch so it sends back an "unsupported media type" or "not acceptable" response that includes this header. The docs don't say what a browser _should_ do next - but they could presumably send a new request with the appropriate types -right?  So my GUESS is that supporting these headers would mean "automatically sending back a response that has the right types, if available", while not supporting browsers would just fail the request. Of course it might not work this way, or be intended to work this way. Can you recommend an expert to ask, or dismiss this out of hand?

ddbeck:
Thanks, @hamishwillee. I guess I'm inclined to defer adding this data for now. I know of a few other HTTP pages (e.g., [PATCH](https://developer.mozilla.org/en-US/docs/Web/HTTP/Methods/PATCH)) which omit BCD and this seems like the appropriate thing to do here too.

Thinking this through, I suspect the existence of the header is interesting to Web developers writing code that makes HTTP requests directly (e.g., if you're using `fetch()`, then you'll need to know about this to write code that handles `Accept-Patch` properly), but there's not anything meaningful to say about the browser's own behavior at this point.

@chrisdavidmills, do you have a hot take on this? Feels like I'm making a bit of an editorial decision here and I'd appreciate your input.

hamishwillee:
@ddbeck On further thought I think you're probably right that this is not relevant. The spec talks about when a _server_ might optionally return these headers - not returning them is pretty much "explicitly" compliant. The spec appears to  say nothing at all about what a client could/should do if it gets this - which implies that anything you do or do not do with it is compliant.

@chrisdavidmills If you agree then we can close this. But what happens in the browser compatibility section in the docs? e.g. https://wiki.developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Post#Browser_compatibility - how do we mark "compatibility not relevant"?

chrisdavidmills:
@hamishwillee  @ddbeck I agree that we can probably close this, but with the caveat that we should probably write something in the Browser compatibility section to explain why we are not including any data, rather than just omitting it entirely. As people may ask, y'know?

"No compat data is included here because..."

hamishwillee:
Great. I have updated both docs with a note "Browser compatibility is not relevant for this header (header is sent by server, and the specification does not define client behaviour)."
- https://wiki.developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Post#Browser_compatibility
- https://wiki.developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Accept-Patch#Browser_compatibility

PR #7246, opened 2020-11-03
Set real values for HTML element APIs for IE/Edge

vinyldarkscratch:
This PR updates the data for the HTML element APIs for Internet Explorer and Microsoft Edge using the mdn-bcd-collector project (using results from IE 5.5 to 11, and Edge 13 to 85), aided with some manual adjustments.  There are a few changes to existing data for consistency reasons.

Notes:
- Because there is no real way to differentiate whether an element is initialized using the HTMLElement API or its own API until IE 9, the version numbers for the APIs themselves are guesstimated based upon the values from the subfeatures (or when not possible, when the elements themselves were supported).
- This removes a little useless and inaccurate data on `api.HTMLScriptElement.defer` where it was marked as non-standard implementation since IE 4 -- the HTMLElement API wasn't supported until IE 5.5, and having a note that just reads "Non-standard implementation" is simply unhelpful.

### DIFF
--- a/api/HTMLScriptElement.json
+++ b/api/HTMLScriptElement.json
@@ -206,16 +206,9 @@
             "firefox_android": {
               "version_added": "4"
             },
-            "ie": [
-              {
-                "version_added": "10"
-              },
-              {
-                "version_added": "4",
-                "version_removed": "10",
-                "notes": "Non-standard implementation"
-              }
-            ],
+            "ie": {
+              "version_added": "5.5"
+            },
             "opera": {
               "version_added": "≤12.1"
             },
--- a/api/HTMLScriptElement.json
+++ b/api/HTMLScriptElement.json
@@ -20,7 +20,7 @@
             "version_added": "4"
           },
           "ie": {
-            "version_added": true
+            "version_added": "5.5"
           },
           "opera": {
             "version_added": "≤12.1"
@@ -113,7 +113,7 @@
               "version_added": "4"
             },
             "ie": {
-              "version_added": true
+              "version_added": "6"
             },
             "opera": {
               "version_added": "≤12.1"
@@ -261,7 +261,7 @@
               "version_added": "4"
             },
             "ie": {
-              "version_added": true
+              "version_added": "5.5"
             },
             "opera": {
               "version_added": "15"
@@ -308,7 +308,7 @@
               "version_added": "4"
             },
             "ie": {
-              "version_added": true
+              "version_added": "5.5"
             },
             "opera": {
               "version_added": "15"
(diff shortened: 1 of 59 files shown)


foolip:
> This PR updates the data for the HTML element APIs for Internet Explorer and Microsoft Edge using the mdn-bcd-collector project (using results from IE 5.5 to 11, and Edge 13 to 85), aided with some manual adjustments.

@vinyldarkscratch does this mean that for everything set to "5.5", it may in fact have been supported even earlier, but it's too much work to find out? Do we need a range version for IE similar to https://github.com/mdn/browser-compat-data/pull/6915?

vinyldarkscratch:
The `HTMLElement` API isn't supported until IE 5.5, so it's not possible for any support before then.

Although, it would be helpful to create ranged values for all of the browser testing gaps.  I'll create a PR for that.

foolip:
> The `HTMLElement` API isn't supported until IE 5.5, so it's not possible for any support before then.

I see. Many HTML elements themselves were of course supported before that point, but that makes sense for this PR then. We'll need to deal with this situation in https://github.com/mdn/browser-compat-data/issues/6571 if we try to match up HTML elements + attributes with interfaces + reflected attributes.

foolip (approved):
I've done another review pass, and this all seems very plausible. With a change this big (340 version changes) there's bound to be *something* that's not quite right, but I'm confident it would be minor things and this is a huge improvement in the data.

api/HTMLScriptElement.json
```diff
@@ -206,16 +206,9 @@
...
               "version_added": "4"
             },
-            "ie": [
-              {
-                "version_added": "10"
-              },
-              {
-                "version_added": "4",
-                "version_removed": "10",
-                "notes": "Non-standard implementation"
```
foolip, line 216: Looking at this again now as I was about to merge with the "useless and inaccurate" note in the commit message, I note that this actually lines up with a comment in https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script#Browser_compatibility saying "In versions prior to Internet Explorer 10, it implemented defer by a proprietary specification. Since version 10 it conforms to the W3C specification." Neither are great notes, but I think it'd be best to make them consistent rather than removing this note. Something was different about `defer` before IE10, apparently.

I'll back out this change to get the rest landed.

foolip: I've filed https://github.com/mdn/browser-compat-data/issues/7258 to follow up on that.

vinyldarkscratch:
Oof, reverting those changes causes a linter failure...

PR #7247, opened 2020-11-03
Fix data for HTML element APIs for IE/Edge

vinyldarkscratch:
This PR updates the data for the HTML element APIs for Internet Explorer and Microsoft Edge using the mdn-bcd-collector project (using results from IE 5.5 to 11, and Edge 13 to 85), aided with some manual adjustments.

### DIFF
--- a/api/HTMLMediaElement.json
+++ b/api/HTMLMediaElement.json
@@ -20,7 +20,7 @@
             "version_added": "4"
           },
           "ie": {
-            "version_added": "9"
+            "version_added": "5.5"
           },
           "opera": {
             "version_added": "10.5"
@@ -732,7 +732,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": "9"
+              "version_added": false
             },
             "opera": {
               "version_added": true
@@ -960,7 +960,7 @@
               "version_added": "49"
             },
             "edge": {
-              "version_added": "12"
+              "version_added": "79"
             },
             "firefox": {
               "version_added": "20"
@@ -969,7 +969,7 @@
               "version_added": "15"
             },
             "ie": {
-              "version_added": "9"
+              "version_added": false
             },
             "opera": {
               "version_added": "36"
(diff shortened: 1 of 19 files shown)


api/HTMLMediaElement.json
```diff
@@ -960,7 +960,7 @@
               "version_added": "49"
             },
             "edge": {
-              "version_added": "12"
+              "version_added": "79"
```
foolip, line 963: The "12" came from mirroring of IE data in https://github.com/mdn/browser-compat-data/pull/5566, which in turn came from https://github.com/mdn/browser-compat-data/pull/1166.

I've confirmed that http://mdn-bcd-collector.appspot.com/tests/api/HTMLMediaElement/disableRemotePlayback returns false in IE11. https://w3c.github.io/remote-playback/ is newer than IE11, so I think the data was just wrong.

api/HTMLVideoElement.json
```diff
@@ -513,7 +513,7 @@
               "version_added": false
             },
             "ie": {
-              "version_added": "10"
+              "version_added": false
```
foolip, line 516: I've confirmed https://mdn-bcd-collector.appspot.com/tests/api/HTMLVideoElement/msIsStereo3D returns false in IE 10 and 11.

And yet, https://docs.microsoft.com/en-us/previous-versions//hh772572(v=vs.85) says this was added in IE 10, but going by https://docs.microsoft.com/en-us/previous-versions/windows/apps/hh452749(v=win.10)?redirectedfrom=MSDN maybe it was only for some platforms, and there was a runtime flag for exposing the API.

I'm going to back out this change, because even if ultimately correct, it's a trickier case that deserves the visibility of its own PR, and maybe asking from friends from Microsoft.

api/HTMLMeterElement.json
```diff
@@ -17,7 +17,7 @@
             "version_added": "16"
           },
           "ie": {
-            "version_added": "10"
+            "version_added": false
```
foolip, line 20: https://caniuse.com/meter agrees with these changes, and a quick test of `<meter>` parsing/rendering in IE11 suggests no support indeed.

foolip (approved):
LGTM % `msIsStereo3D` which I will back out.

Issue #6896, opened 2020-10-12
(Re)Consider adding “pre-release” browser-version value for cases where we don’t yet have a BCD-known version number


sideshowbarker:
## Concrete proposal

Here’s a concrete proposal for marking up data for pre-release browser versions; let’s start doing one of the following:

1. Start using Safari Technology Preview release numbers in addition to Safari major-release numbers; or
2. Start using some generic value “pre-release version” value — for example, `dev` or `pre-release` or whatever; or
3. Start using values that correspond to how the browser projects actually label there particular pre-release versions; for example, `nightly`, `canary`, and `TP` (technology preview).

## Problem description

_The concrete proposal above is intended to address a problem that we’ve had discussion about before. So maybe this is a basically duplicate of an existing issue. If so, I can just take all of the following and add it as comment to that other issue later._

Here’s a detailed explanation of the actual problem I am hoping we can find a way to solve together:

We’re aware that we have a significant number of cases when we know support for a certain feature has been implemented in a particular browser, but only in a pre-release version. And in those cases, if we do not yet have that version number in any https://github.com/mdn/browser-compat-data/tree/master/browsers file, then we have no way of updating the related BCD data to indicate the browser in question has support for that feature.

I assert that’s a deficiency which we should try to find some way to correct — because the data point *“this browser has implemented support for this feature”* is a useful fact, regardless of what the actual browser version is.

In particular, I think that *“this browser has implemented support for this feature”* data is potentially useful to downstream consumers of the BCD data.

And more particularly, I can state that it’s absolutely useful to the downstream-consumer tooling that I maintain myself at https://github.com/w3c/mdn-spec-links and that’s used for adding MDN annotations to all W3C and WHATWG specs.

So I’d like to request that we re-consider coming up with some way to mark our data for such cases.

### The special problem with Safari data

It’s worth noting here that this is a bigger problem for the Safari data than it is for the Chrome or Firefox data. The reason is that while major numbered versions of Firefox and Chrome are released every six or seven *weeks* (that is, 7 or 8 releases each year) — major numbered versions of Safari are released only every six *months* (that is, only twice a year; typically with one major release in March and then another in September).

What that ends up meaning for us is that we have to wait a very long time between Safari releases before we can update our data — even in cases where we know that something has already shipped in a Safari Technology Preview release.

### Safari gets actually get updates often; let’s stop ignoring the existing feature data already available

[...]

PR #7248, opened 2020-11-03
Add beta Safari from iOS 14.2 and Big Sur 11.0.1

Constellation:
- iOS 14.2 Release Candidate (18B91)
- macOS Big Sur 11.0.1 beta (20B5012d)

https://developer.apple.com/news/releases/

to list Safari data for `Intl[@@toStringTag]` in #7020.

### DIFF
--- a/browsers/safari.json
+++ b/browsers/safari.json
@@ -169,10 +169,15 @@
         },
         "14": {
           "release_date": "2020-09-16",
-          "release_notes": "https://developer.apple.com/documentation/safari-release-notes/safari-14-beta-release-notes",
+          "release_notes": "https://developer.apple.com/documentation/safari-release-notes/safari-14-release-notes",
           "status": "current",
           "engine": "WebKit",
           "engine_version": "610.1.28"
+        },
+        "14.0.1": {
+          "status": "beta",
+          "engine": "WebKit",
+          "engine_version": "610.2.11"
         }
       }
     }
--- a/browsers/safari_ios.json
+++ b/browsers/safari_ios.json
@@ -160,10 +160,15 @@
         },
         "14": {
           "release_date": "2020-09-16",
-          "release_notes": "https://developer.apple.com/documentation/safari-release-notes/safari-14-beta-release-notes",
+          "release_notes": "https://developer.apple.com/documentation/safari-release-notes/safari-14-release-notes",
           "status": "current",
           "engine": "WebKit",
           "engine_version": "610.1.28"
+        },
+        "14.2": {
+          "status": "beta",
+          "engine": "WebKit",
+          "engine_version": "610.2.11"
         }
       }
     }


Constellation:
@foolip Could you review this PR? Thanks!

ddbeck:
Thanks for starting this PR, @Constellation. For posterity, could you inline additional details from the Apple site in the PR description? The seemingly-relevant pages linked from https://developer.apple.com/news/releases/ require a sign-in and I find that Apple developer links tend to break rather often anyway. Some additional detail (such as how you found the webkit version numbers) would be helpful to reviewing this and to understanding this change if it ever needs to be revisited in the future. Thank you again!

Constellation:
Thanks!

> Thanks for starting this PR, @Constellation. For posterity, could you inline additional details from the Apple site in the PR description? The seemingly-relevant pages linked from https://developer.apple.com/news/releases/ require a sign-in and I find that Apple developer links tend to break rather often anyway.

I think the release notes will become ready after the beta software is released. So at this point, there is no URL for document. I think this is the same to chrome.json's beta versions.

> Some additional detail (such as how you found the webkit version numbers) would be helpful to reviewing this and to understanding this change if it ever needs to be revisited in the future. Thank you again!

I installed these softwares and checked version by opening "Safari">"About Safari" menu. It shows Safari version (in this case, `14.0.1`, and WebKit build `610.2.11`).

Constellation:
The screenshot from installed `macOS Big Sur 11.0.1 beta (20B5012d)`.

Screen Shot 2020-11-03 at 5 25 09 AM

foolip:
Do we have confirmation that iOS 14.2 will use a different version of WebKit from a phone with the beta installed? In the past, more often than not it's been the x.3 release that bumps WebKit, and this usually happens ~6 months about the x.0 release.

Constellation:
> Do we have confirmation that iOS 14.2 will use a different version of WebKit from a phone with the beta installed?

Yes, iOS 14 and iOS 14.2 are shipping different versions of WebKit, and its WebKit revision is `610.2.11`, this is the same to Big Sur's latest beta's WebKit. I installed and checked WebKit's plist information.

It seems that browser-compat-data is using iOS version for iOS Safari while Safari has different version number (in this case, iOS Safari version is 14.0.1 too), I used 14.2 for iOS and 14.0.1 for macOS (since in macOS, Safari 14 can be installed into Catalina etc.).

Constellation:
Ping review :)

foolip:
I'm going to have to defer to @ddbeck on this, I know BCD skips some Safari releases which are too similar, but I don't know if this is going to be one of those. Since Safari 10, each release in BCD as included a bump to the WebKit major version, and that's not the case here.

I've tried to work out from https://trac.webkit.org/log/webkit/tags/Safari-610.1.28 + https://trac.webkit.org/log/webkit/tags/Safari-610.2.11 how much change from WebKit trunk (if any) is included but don't know how, since all my VCS skills are with Git and WebKit tags/branches aren't included in the Git mirror.

cc @vinyldarkscratch given https://github.com/mdn/browser-compat-data/issues/5846.

cc @sideshowbarker given https://github.com/mdn/browser-compat-data/issues/6896.

vinyldarkscratch:
For browser versions, we don't really record patch versions as separate versions.  This decision was made under the expectation that users are likely to run the latest patch of the minor version of a browser.  Thus, if there are any new features added to 14.0.1, we'd record it as 14.  (See https://github.com/mdn/browser-compat-data/blob/master/docs/data-guidelines.md#release-lines-and-backported-features)

Constellation:
So, for features that is enabled on Safari 14.0.1, should we just record it as 14 feature?

Constellation:
@vinyldarkscratch OK, I'll open a PR that records `Intl[@@toStringTag]` as Safari 14.

Another question: so, next time, when patched version of Safari is in beta, and the feature is implemented in that beta (in this case, `Intl[@@toStringTag]`), can we just mark it as a major released feature (since we do not have an entry to this beta, and this will be marked as a major released one anyway)?

vinyldarkscratch:
> So, for features that is enabled on Safari 14.0.1, should we just record it as 14 feature?

That's correct!  Anything in Safari 14.0.x should be recorded as Safari 14 -- once we reach Safari 14.1, then we'll add that new version.

> Another question: so, next time, when patched version of Safari is in beta, and the feature is implemented in that beta (in this case, `Intl[@@toStringTag]`), can we just mark it as a major released feature (since we do not have an entry to this beta, and this will be marked as a major released one anyway)?

Overall, I recommend avoiding adding results for browsers while they're still in beta (and _especially_ any releases under the canary channel, including Safari Technology Previews).  Sometimes, a browser may release a feature that is available in that beta, but will back out of it or place it behind a disabled flag due to instability or otherwise.

Constellation:
Thanks!

> Overall, I recommend avoiding adding results for browsers while they're still in beta (and especially any releases under the canary channel, including Safari Technology Previews). Sometimes, a browser may release a feature that is available in that beta, but will back out of it or place it behind a disabled flag due to instability or otherwise.

OK, so, should we avoid including beta releases of Firefox and Chrome too? Chrome and Firefox also have `beta` and `nightly` entries, and #7020's Intl feature is only available in beta in Firefox.
https://github.com/mdn/browser-compat-data/blob/master/browsers/chrome.json#L602-L611
https://github.com/mdn/browser-compat-data/blob/master/browsers/firefox.json#L609-L622

vinyldarkscratch:
We do keep data about beta and nightly releases around, but generally, I recommend adding data for final releases only.

Constellation:
OK, thanks!!!

PR #7252, opened 2020-11-03
Document.all is specified as obsolete

ddbeck:
This flips `api.Document.all`'s status from `standard_track` to `true` and `deprecated` to `true`, reflecting its weird state [as specified but explicitly obsolete](https://html.spec.whatwg.org/multipage/obsolete.html#other-elements,-attributes-and-apis).

Prompted by #7193.

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -459,8 +459,8 @@
           },
           "status": {
             "experimental": false,
-            "standard_track": false,
-            "deprecated": false
+            "standard_track": true,
+            "deprecated": true
           }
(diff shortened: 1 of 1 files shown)


foolip (approved):

foolip:
Yep, we might want to do this for everything in https://html.spec.whatwg.org/multipage/obsolete.html.

foolip:
I've filed https://github.com/mdn/browser-compat-data/issues/7255.

PR #7254, opened 2020-11-03
' my string '.trim() is not supported in IE9 (but in IE10).

apepper:
`String.prototype.trim()` is not working on IE9:

trim on IE9

On IE10 and later it does work:
trim on IE10


ddbeck (approved):
Thank you! 🎉

Issue #7171, opened 2020-10-29
Turn on branch protection rules for main


ddbeck:
On #6615, there was discussion about what [`GOVERNANCE.md` says about pushing to branches on this repo](https://github.com/mdn/browser-compat-data/blob/master/GOVERNANCE.md#peers) and also my fear of accidentally pushing to `main`.

@sideshowbarker suggested turning on branch protection rules for `main` in https://github.com/mdn/browser-compat-data/pull/6615#issuecomment-718039025. I think this is a great idea. In addition to curing my fear, it will also encourage some better bookkeeping around the small number of changes pushed to `main` directly, such as incrementing the package version number.

Tasks:

- [x] Review and, if applicable, revise names of required checks
- [x] Enable branch protection rules for `main`
- [ ] Optional: see if we can protect any any branch with some sort of user or issue prefix. That is, branches _not_ in the form `[\w_-]+\/.+` (e.g., `someuser/topic` or `1234/fixes-issue`).
- [x] Optional: retire any unprefixed branches
- [x] Revise package publish doc to use PRs instead of pushing to `main`
- [x] Revise `GOVERNANCE.md` to reflect our actual use of branches on the repo, rather than a never-practiced prohibition against using branches

foolip:
For what it's worth, I found the prohibition against pushing branches to this repo to be refreshing, even if unusual. The branches in https://github.com/web-platform-tests/wpt just keep piling up, and that just can't happen if you never create them in the first place. There's also less noise when doing `git fetch`, which I appreciate.

sideshowbarker:
> For what it's worth, I found the prohibition against pushing branches to this repo to be refreshing, even if unusual. The branches in https://github.com/web-platform-tests/wpt just keep piling up, and that just can't happen if you never create them in the first place. There's also less noise when doing `git fetch`, which I appreciate.

As a counter-example, for WHATWG specs, we push branches to the canonical repos, and the branches don’t pile up.

But anyway, I have no objections to us enforcing a policy of prohibiting pushing of branches to this repo.

However, ideally we should be able to have GitHub actually prevent that. Does GitHub provide an option for it?

The only thing I am aware of that would maybe work to enforce it is if we configured the repo to restrict pushing to *any* branches — by specifying a wildcard in the pattern field of the branch-protection UI. But I’ve never tried that, so I’m not certain it actually would have the effect we need.

[...]

PR #7259, opened 2020-11-03
Fix api.HTMLScriptElement.defer data for IE

### LINKED ISSUE #7258: Sort out notes for the defer attribute in IE
Filing an issue to follow up on https://github.com/mdn/browser-compat-data/pull/7246#discussion_r516931091.

We have two bits of data about `defer` with notes that aren't super specific about what works and not, which should be aligned:

https://github.com/mdn/browser-compat-data/blob/82f3394ae95f1021302bf962c1e9662303fcbbfe/api/HTMLScriptElement.json#L209-L218

https://github.com/mdn/browser-compat-data/blob/82f3394ae95f1021302bf962c1e9662303fcbbfe/html/elements/script.json#L168-L171

vinyldarkscratch:
This PR fixes the data for the `defer` property of the `HTMLScriptElement` API.  In #7246, the data was simplified as a note that simply states "Non-standard implementation" is simply unhelpful.  Before it was merged, it was reverted to its original state before changes were made, which is causing a linter failure.  This is a quick fix to make the data match what's in `html/elements/script.json`.

Fixes #7258.

### DIFF
--- a/api/HTMLScriptElement.json
+++ b/api/HTMLScriptElement.json
@@ -206,16 +206,10 @@
             "firefox_android": {
               "version_added": "4"
             },
-            "ie": [
-              {
-                "version_added": "10"
-              },
-              {
-                "version_added": "4",
-                "version_removed": "10",
-                "notes": "Non-standard implementation"
-              }
-            ],
+            "ie": {
+              "version_added": "10",
+              "notes": "In versions prior to Internet Explorer 10, it implemented <code>defer</code> by a proprietary specification. Since version 10 it conforms to the W3C specification."
+            },
             "opera": {
               "version_added": "≤12.1"
             },


vinyldarkscratch:
This is a fix for data that is failing linter in the current `master`.  To get this resolved as quickly as possible, I'm going to merge this without review for now, so that it's not causing issues with new/updated PRs.  (I know that I'm not exactly supposed to merge changes without review, but I feel this scenario qualifies as an exception.  @ddbeck, please let me know if I should have waited for a review instead.)

foolip:
Oh my, I merged https://github.com/mdn/browser-compat-data/pull/7246 with linter errors, I wasn't looking out for that because I'm so used to not being able to merge before checks pass. Sorry about this! https://github.com/mdn/browser-compat-data/issues/7171 would be great!

PR #7260, opened 2020-11-03
Update Chromium data for visibilityState and visibilitychange event

vinyldarkscratch:
This PR updates the Chrome and Opera data for the document `visibilityState` property and `visibilitychange` event + handler.  Based upon results from the mdn-bcd-collector project, the `onvisibilitychange` event handler was implemented in a much later version of Chrome, and never had a prefix.  Both the event and the `visibilityState` property were introduced in Chrome 13 with a prefix, then unprefixed in Chrome 33.


api/Document.json
```diff
@@ -8146,17 +8146,11 @@
...
-              {
-                "version_added": "33"
-              },
-              {
-                "version_added": "13",
-                "prefix": "webkit"
-              }
-            ],
+            "chrome": {
+              "version_added": "62"
```
foolip, line 8150: Can you add a note for this similar to what's already there for Edge and Safari, and maybe make all the notes consistent?

Also this should be mirrored to Opera, which would look like removal in Opera 15 and then later adding back.

api/Document.json
```diff
@@ -11519,9 +11513,15 @@
...
-            "chrome_android": {
-              "version_added": "33"
-            },
+            "chrome_android": [
+              {
+                "version_added": "33"
+              },
+              {
+                "version_added": "18",
+                "prefix": "webkit"
```
foolip, line 11522: Yep, seems like the prefixed "webkitvisibilitychange" event is still being fired:
https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/core/dom/document.cc;l=1932;drc=df872ce8fcce25af51aa6b0f9fe8b1135b687524

api/Document.json
```diff
@@ -11536,14 +11536,24 @@
               "version_added": "10",
               "notes": "The <code>onvisibilitychange</code> attribute is not supported in IE. To listen to this event, use <code>document.addEventListener('visibilitychange', function() {});</code>."
             },
-            "opera": {
-              "version_added": "12.1",
-              "notes": "Doesn't fire the <code>visibilitychange</code> event when the browser window is minimized, nor when <code>hidden</code> is set to true."
```
foolip, line 11541: Why remove these notes?

api/Document.json
```diff
@@ -11622,14 +11644,24 @@
             "ie": {
               "version_added": "10"
             },
-            "opera": {
-              "version_added": "12.1",
-              "notes": "Doesn't fire the <code>visibilitychange</code> event when the browser window is minimized, nor when <code>hidden</code> is set to true."
```
foolip, line 11627: Keep these notes?

api/Document.json
```diff
@@ -11645,9 +11677,15 @@
                 "version_added": "1.0"
               }
             ],
-            "webview_android": {
-              "version_added": "4.4.3"
-            }
+            "webview_android": [
+              {
+                "version_added": "4.4.3"
```
foolip, line 11682: This doesn't quiet square with the data for Chrome. The unprefixed variant was added after the prefixed one, but the order is reversed here, presumably because we're trusting the original "4.4.3" here.

If we knew when prefixed and unprefixed variants were added in Safari we could probably update this data with confidence. My best guess from current data is that we should set both to "4.4.3".

foolip:
Ping @vinyldarkscratch

api/Document.json
vinyldarkscratch: I could not confirm this behavior in either Opera 12.1 or 15, nor in more recent versions of Opera (such as Opera 60).

api/Document.json
vinyldarkscratch: Well, the `≤37` is to say that it's in some version between 1 and 37, we just don't know when exactly.  Honestly, though, it would be more accurate if the version were `≤4.4`, since we can determine versions of 4.4 and 4.4.3.

api/Document.json
foolip: OK, if you can see the event firing when minimizing the browser, then dropping this note is fine. (The bit about `hidden` is misleading, it changing and the event firing is connected, both being caused by something else, like the window being hidden.)

api/Document.json
foolip: Ah, I see. This is fine then, could possibly be resolved later when fixing ≤37 data.

foolip:
@vinyldarkscratch I see you've pushed changes, but there's one comment still open, I've resolved the others to make it clear which.

api/Document.json
vinyldarkscratch: I have not been able to confirm that the behavior described in Safari and Edge is also in Chromium, though?

api/Document.json
```diff
@@ -11592,7 +11612,7 @@
                 "version_added": "33"
               },
               {
-                "version_added": true,
+                "version_added": "18",
```
foolip, line 11615: Confirmed `document.webkitVisibilityState` is still around, never removed.

api/Document.json
foolip: The (unprefixed) visibilitychange event was introduced in Chrome 33, so I mean a note similar to this one: `"Before Edge 18, this event handler attribute was not supported; however, the event itself was supported since Edge 12. The event can be listened to via <code>document.addEventListener('visibilitychange', function() {});</code>."`

The note would also apply to Chrome for Android and all other Chromium-based browsers here.

api/Document.json
```diff
@@ -8173,12 +8167,10 @@
               "notes": "This event handler attribute is not supported; however, the event itself is supported since IE 10. The event can be listened to via <code>document.addEventListener('visibilitychange', function() {});</code>."
             },
             "opera": {
-              "version_added": "12.1",
-              "notes": "Doesn't fire the <code>visibilitychange</code> event when the browser window is minimized, nor when <code>hidden</code> is set to true."
+              "version_added": "12.1"
```
foolip, line 8170: If the Chrome data is correct, this should be removed in 15, and then added back in 49.

It might be worth re-testing some Opera versions though, I wonder if this shouldn't be a ranged version and this was added earlier than 12.1.

api/Document.json
```diff
@@ -11536,14 +11534,24 @@
...
-              "version_added": "12.1",
-              "notes": "Doesn't fire the <code>visibilitychange</code> event when the browser window is minimized, nor when <code>hidden</code> is set to true."
-            },
-            "opera_android": {
-              "version_added": "12.1",
-              "notes": "Doesn't fire the <code>visibilitychange</code> event when the browser window is minimized, nor when <code>hidden</code> is set to true."
-            },
+            "opera": [
+              {
+                "version_added": "12.1"
```
foolip, line 11539: If the Chrome data is correct, this should be removed in 15 and added back in 20.

api/Document.json
```diff
@@ -11622,14 +11642,24 @@
...
```
foolip, line 11647: If the Chrome data is correct, this should be removed in 15 and added back in 20.

foolip (commented):
Aw this one is complicated. Prefix + unprefixed, event vs. event handler attribute, and then add Opera's engine switch. I think I caught all the remaining things to fix, I hope so :)

api/Document.json
foolip: I've checked this out in Opera Presto. I confirmed with http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8816 that support for the event *was* added in 12.1 (event logged in 12.1, but not in 11.6). But the event handler attribute was never supported in Opera Presto. So notes will be needed here too.

api/Document.json
foolip: I've confirmed using http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8817 that the unprefixed `visibilityState` was gone in Opera 15, and then back in 20.

api/Document.json
```diff
@@ -8173,12 +8169,12 @@
...
-              "version_added": "12.1",
-              "notes": "Doesn't fire the <code>visibilitychange</code> event when the browser window is minimized, nor when <code>hidden</code> is set to true."
+              "version_added": "49",
+              "notes": "Before Opera 49, this event handler attribute was not supported; however, the event itself was supported since Opera 20. The event can be listened to via <code>document.addEventListener('visibilitychange', function() {});</code>."
             },
             "opera_android": {
-              "version_added": "12.1",
-              "notes": "Doesn't fire the <code>visibilitychange</code> event when the browser window is minimized, nor when <code>hidden</code> is set to true."
+              "version_added": "46",
+              "notes": "Before Opera Android 46, this event handler attribute was not supported; however, the event itself was supported since Opera 20. The event can be listened to via <code>document.addEventListener('visibilitychange', function() {});</code>."
```
foolip, line 8177:
```suggestion
              "notes": "Before Opera Android 46, this event handler attribute was not supported; however, the event itself was supported since Opera Android 20. The event can be listened to via <code>document.addEventListener('visibilitychange', function() {});</code>."
```

foolip (approved):
I have now pedantically fixed everything that I commented on and added lots of notes. I'm happy with this PR, but since I authored non-trivials parts of it now, I'd like to ask @ddbeck or @Elchi3 for review as well.

ddbeck:
I will give this a proper review, but I don't think it will make today's release. Regretfully, I too have some pedantic suggestions to make (about note text).

foolip:
Thanks @ddbeck! There's no rush on this so it's fine it missed today's release.

api/Document.json
```diff
@@ -8257,12 +8253,12 @@
               "notes": "This event handler attribute is not supported; however, the event itself is supported since IE 10. The event can be listened to via <code>document.addEventListener('visibilitychange', function() {});</code>."
             },
             "opera": {
-              "version_added": "12.1",
-              "notes": "Doesn't fire the <code>visibilitychange</code> event when the browser window is minimized, nor when <code>hidden</code> is set to true."
+              "version_added": "49",
+              "notes": "Before Opera 49, this event handler attribute was not supported; however, the event itself was supported since Opera 20. The event can be listened to via <code>document.addEventListener('visibilitychange', function() {});</code>."
```
ddbeck, line 8257:
```suggestion
              "notes": "Before Opera 49, this event handler attribute is not supported; however, the event itself is supported since Opera 20. The event can be listened to via <code>document.addEventListener('visibilitychange', function() {});</code>."
```

api/Document.json
```diff
@@ -8230,17 +8230,13 @@
...
-                "version_added": "33"
-              },
-              {
-                "version_added": "13",
-                "prefix": "webkit"
-              }
-            ],
+            "chrome": {
+              "version_added": "62",
+              "notes": "Before Chrome 62, this event handler attribute was not supported; however, the event itself was supported since Chrome 33. The event can be listened to via <code>document.addEventListener('visibilitychange', function() {});</code>."
```
ddbeck, line 8235:
```suggestion
              "notes": "Before Chrome 62, this event handler attribute is not supported; however, the event itself is supported since Chrome 33. The event can be listened to via <code>document.addEventListener('visibilitychange', function() {});</code>."
```

api/Document.json
```diff
@@ -8230,17 +8230,13 @@
...
+            },
             "chrome_android": {
-              "version_added": "33"
+              "version_added": "62",
+              "notes": "Before Chrome 62, this event handler attribute was not supported; however, the event itself was supported since Chrome 33. The event can be listened to via <code>document.addEventListener('visibilitychange', function() {});</code>."
```
ddbeck, line 8239:
```suggestion
              "notes": "Before Chrome 62, this event handler attribute is not supported; however, the event itself is supported since Chrome 33. The event can be listened to via <code>document.addEventListener('visibilitychange', function() {});</code>."
```

api/Document.json
```diff
@@ -8257,12 +8253,12 @@
...
-              "version_added": "12.1",
-              "notes": "Doesn't fire the <code>visibilitychange</code> event when the browser window is minimized, nor when <code>hidden</code> is set to true."
+              "version_added": "49",
+              "notes": "Before Opera 49, this event handler attribute was not supported; however, the event itself was supported since Opera 20. The event can be listened to via <code>document.addEventListener('visibilitychange', function() {});</code>."
             },
             "opera_android": {
-              "version_added": "12.1",
-              "notes": "Doesn't fire the <code>visibilitychange</code> event when the browser window is minimized, nor when <code>hidden</code> is set to true."
+              "version_added": "46",
+              "notes": "Before Opera Android 46, this event handler attribute was not supported; however, the event itself was supported since Opera Android 20. The event can be listened to via <code>document.addEventListener('visibilitychange', function() {});</code>."
```
ddbeck, line 8261:
```suggestion
              "notes": "Before Opera Android 46, this event handler attribute is not supported; however, the event itself is supported since Opera Android 20. The event can be listened to via <code>document.addEventListener('visibilitychange', function() {});</code>."
```

api/Document.json
```diff
@@ -8281,10 +8277,12 @@
               ]
             },
             "samsunginternet_android": {
-              "version_added": "2.0"
+              "version_added": "8.0",
+              "notes": "Before Samsung Internet 8.0, this event handler attribute was not supported; however, the event itself was supported since Samsung Internet 2.0. The event can be listened to via <code>document.addEventListener('visibilitychange', function() {});</code>."
```
ddbeck, line 8281:
```suggestion
              "notes": "Before Samsung Internet 8.0, this event handler attribute is not supported; however, the event itself is supported since Samsung Internet 2.0. The event can be listened to via <code>document.addEventListener('visibilitychange', function() {});</code>."
```

api/Document.json
```diff
@@ -8281,10 +8277,12 @@
...
             },
             "webview_android": {
-              "version_added": "4.4.3"
+              "version_added": "62",
+              "notes": "Before WebView 62, this event handler attribute was not supported; however, the event itself was supported since WebView 4.4.3. The event can be listened to via <code>document.addEventListener('visibilitychange', function() {});</code>."
```
ddbeck, line 8285:
```suggestion
              "notes": "Before WebView 62, this event handler attribute is not supported; however, the event itself is supported since WebView 4.4.3. The event can be listened to via <code>document.addEventListener('visibilitychange', function() {});</code>."
```

api/Document.json
```diff
@@ -11679,16 +11677,24 @@
           "support": {
             "chrome": [
               {
-                "version_added": "33"
+                "version_added": "33",
+                "notes": "The <code>onvisibilitychange</code> attribute was not supported until Chrome 62. To listen to this event in earlier versions of Chrome, use <code>document.addEventListener('visibilitychange', function() {});</code>."
```
ddbeck, line 11681:
```suggestion
                "notes": "Before Chrome 62, the <code>onvisibilitychange</code> attribute is not supported. To listen to this event in earlier versions of Chrome, use <code>document.addEventListener('visibilitychange', function() {});</code>."
```

api/Document.json
```diff
@@ -11679,16 +11677,24 @@
...
                 "prefix": "webkit"
               }
             ],
-            "chrome_android": {
-              "version_added": "33"
-            },
+            "chrome_android": [
+              {
+                "version_added": "33",
+                "notes": "The <code>onvisibilitychange</code> attribute was not supported until Chrome 62. To listen to this event in earlier versions of Chrome, use <code>document.addEventListener('visibilitychange', function() {});</code>."
```
ddbeck, line 11691:
```suggestion
                "notes": "Before Chrome 62, the <code>onvisibilitychange</code> attribute is not supported. To listen to this event in earlier versions of Chrome, use <code>document.addEventListener('visibilitychange', function() {});</code>."
```

api/Document.json
```diff
@@ -11703,36 +11709,72 @@
...
-              "notes": "Doesn't fire the <code>visibilitychange</code> event when the browser window is minimized, nor when <code>hidden</code> is set to true."
-            },
-            "opera_android": {
-              "version_added": "12.1",
-              "notes": "Doesn't fire the <code>visibilitychange</code> event when the browser window is minimized, nor when <code>hidden</code> is set to true."
-            },
+            "opera": [
+              {
+                "version_added": "20",
+                "notes": "The <code>onvisibilitychange</code> attribute was not supported until Opera 49. To listen to this event in earlier versions of Opera, use <code>document.addEventListener('visibilitychange', function() {});</code>."
```
ddbeck, line 11715:
```suggestion
                "notes": "Before Opera 49, the <code>onvisibilitychange</code> attribute is not supported. To listen to this event in earlier versions of Opera, use <code>document.addEventListener('visibilitychange', function() {});</code>."
```

api/Document.json
```diff
@@ -11703,36 +11709,72 @@
...
+              },
+              {
+                "version_added": "12.1",
+                "version_removed": "15"
+              }
+            ],
+            "opera_android": [
+              {
+                "version_added": "20",
+                "notes": "The <code>onvisibilitychange</code> attribute was not supported until Opera Android 46. To listen to this event in earlier versions of Opera Android, use <code>document.addEventListener('visibilitychange', function() {});</code>."
```
ddbeck, line 11729:
```suggestion
                "notes": "Before Opera  Android 46, the <code>onvisibilitychange</code> attribute is not supported. To listen to this event in earlier versions of Opera Android, use <code>document.addEventListener('visibilitychange', function() {});</code>."
```

api/Document.json
```diff
@@ -11703,36 +11709,72 @@
...
               "partial_implementation": true,
               "notes": [
                 "Doesn't fire the <code>visibilitychange</code> event when navigating away from a document, so also include code to check for the <code>pagehide</code> event (which does fire for that case in all current browsers). See WebKit bugs <a href='https://webkit.org/b/116769'>116769</a>, <a href='https://webkit.org/b/151234'>151234</a>, <a href='https://webkit.org/b/151610'>151610</a>, and <a href='https://webkit.org/b/194897'>194897</a>.",
-                "Before Safari 14, the event does not bubble, so <code>document.addEventListener('visibilitychange', ...)</code> works, but <code>window.addEventListener('visibilitychange', ...)</code> does not."
+                "Before Safari 14, the event does not bubble, so <code>document.addEventListener('visibilitychange', ...)</code> works, but <code>window.addEventListener('visibilitychange', ...)</code> does not.",
+                "The <code>onvisibilitychange</code> attribute was not supported until Safari 10.1. To listen to this event in earlier versions of Safari, use <code>document.addEventListener('visibilitychange', function() {});</code>."
```
ddbeck, line 11746:
```suggestion
                "Before Safari 10.1, the <code>onvisibilitychange</code> attribute is not supported. To listen to this event in earlier versions of Safari, use <code>document.addEventListener('visibilitychange', function() {});</code>."
```

api/Document.json
```diff
@@ -11703,36 +11709,72 @@
...
               "notes": [
                 "Doesn't fire the <code>visibilitychange</code> event when navigating away from a document, so also include code to check for the <code>pagehide</code> event (which does fire for that case in all current browsers). See WebKit bugs <a href='https://webkit.org/b/116769'>116769</a>, <a href='https://webkit.org/b/151234'>151234</a>, <a href='https://webkit.org/b/151610'>151610</a>, and <a href='https://webkit.org/b/194897'>194897</a>.",
-                "Before Safari 14, the event does not bubble, so <code>document.addEventListener('visibilitychange', ...)</code> works, but <code>window.addEventListener('visibilitychange', ...)</code> does not."
+                "Before Safari iOS 14, the event does not bubble, so <code>document.addEventListener('visibilitychange', ...)</code> works, but <code>window.addEventListener('visibilitychange', ...)</code> does not.",
+                "The <code>onvisibilitychange</code> attribute was not supported until Safari iOS 10.3. To listen to this event in earlier versions of Safari iOS, use <code>document.addEventListener('visibilitychange', function() {});</code>."
```
ddbeck, line 11755:
```suggestion
                "Before Safari iOS 10.3, the <code>onvisibilitychange</code> attribute is not supported. To listen to this event in earlier versions of Safari iOS, use <code>document.addEventListener('visibilitychange', function() {});</code>."
```

api/Document.json
```diff
@@ -11703,36 +11709,72 @@
...
-            "samsunginternet_android": {
-              "version_added": "2.0"
-            },
-            "webview_android": {
-              "version_added": "4.4.3"
-            }
+            "samsunginternet_android": [
+              {
+                "version_added": "2.0",
+                "notes": "The <code>onvisibilitychange</code> attribute was not supported until Samsung Internet 8.0. To listen to this event in earlier versions of Samsung Internet, use <code>document.addEventListener('visibilitychange', function() {});</code>."
```
ddbeck, line 11761:
```suggestion
                "notes": "Before Samsung Internet 8.0, the <code>onvisibilitychange</code> attribute is not supported. To listen to this event in earlier versions of Samsung Internet, use <code>document.addEventListener('visibilitychange', function() {});</code>."
```

api/Document.json
```diff
@@ -11703,36 +11709,72 @@
...
+              },
+              {
+                "version_added": "1.0",
+                "prefix": "webkit"
+              }
+            ],
+            "webview_android": [
+              {
+                "version_added": "4.4.3",
+                "notes": "The <code>onvisibilitychange</code> attribute was not supported until WebView 62. To listen to this event in earlier versions of WebView, use <code>document.addEventListener('visibilitychange', function() {});</code>."
```
ddbeck, line 11771:
```suggestion
                "notes": "Before WebView 62, the <code>onvisibilitychange</code> attribute is not supported. To listen to this event in earlier versions of WebView, use <code>document.addEventListener('visibilitychange', function() {});</code>."
```

ddbeck (changes requested):
I made an effort here to make the notes more consistent. Not perfect, but it's better, I think.

api/Document.json
```diff
@@ -11679,16 +11677,24 @@
...
+                "notes": "Before Chrome 62, the <code>onvisibilitychange</code> attribute is not supported. To listen to this event in earlier versions of Chrome, use <code>document.addEventListener('visibilitychange', function() {});</code>."
+              },
+              {
+                "version_added": "18",
+                "prefix": "webkit"
+              }
+            ],
             "edge": {
               "version_added": "12",
               "notes": "The <code>onvisibilitychange</code> attribute was not supported until Edge 18. To listen to this event in earlier versions of Edge, use <code>document.addEventListener('visibilitychange', function() {});</code>."
```
foolip, line 11700:
```suggestion
              "notes": "Before Edge 18, the <code>onvisibilitychange</code> attribute is not supported. To listen to this event in earlier versions of Edge, use <code>document.addEventListener('visibilitychange', function() {});</code>."
```

foolip: Right, @ddbeck?

ddbeck: Yes! I didn't touch the ones outside the +/- diff bounds—thanks for catching this one. If you want to change this (and others) and merge at your discretion, I'm fine with that. 👍

ddbeck (approved):

api/Document.json
```diff
@@ -11703,36 +11709,72 @@
...
+              },
+              {
+                "version_added": "12.1",
+                "version_removed": "15"
+              }
+            ],
+            "opera_android": [
+              {
+                "version_added": "20",
+                "notes": "Before Opera  Android 46, the <code>onvisibilitychange</code> attribute is not supported. To listen to this event in earlier versions of Opera Android, use <code>document.addEventListener('visibilitychange', function() {});</code>."
```
ddbeck, line 11729:
```suggestion
                "notes": "Before Opera Android 46, the <code>onvisibilitychange</code> attribute is not supported. To listen to this event in earlier versions of Opera Android, use <code>document.addEventListener('visibilitychange', function() {});</code>."
```

oh just caught this oddity too!

api/Document.json
ddbeck: @foolip Are you planning to come back to this? I was worried it slipped through the cracks, since you're not the original author

foolip (approved):

PR #7263, opened 2020-11-04
Deprecate img element obsolete attributes (and corresponding HTMLImageElement IDL attributes)

sideshowbarker:
This PR includes a patch that marks the `longdesc` and `usemap` attributes of the `img` element as deprecated, because at https://html.spec.whatwg.org/multipage/obsolete.html, the HTML spec defines those attributes as obsolete.

Other than those, we’ve already marked as deprecated all other `img` attributes that the HTML spec defines as obsolete.

This PR also includes a patch that marks as deprecated all IDL attributes of the `HTMLImageElement` interface for which the HTML spec defines the corresponding content/markup attributes as obsolete.

The affected IDL attributes are `align`, `border`, `hspace`, `longdesc`, `name`, `useMap`, and `vspace`.

Other than those attributes, we’ve already marked as deprecated all other `HTMLImageElement` IDL attributes for which the HTML spec defines the corresponding content/markup attributes as obsolete.

Relates to https://github.com/mdn/browser-compat-data/issues/7255

### DIFF
--- a/api/HTMLImageElement.json
+++ b/api/HTMLImageElement.json
@@ -137,7 +137,7 @@
           "status": {
             "experimental": false,
             "standard_track": true,
-            "deprecated": false
+            "deprecated": true
           }
         }
       },
@@ -233,7 +233,7 @@
           "status": {
             "experimental": false,
             "standard_track": true,
-            "deprecated": false
+            "deprecated": true
           }
         }
       },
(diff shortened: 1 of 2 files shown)


foolip (approved):

foolip:
There are two commits here with good commit message, so I'll rebase this rather than squash it into a single commit.

foolip:
Oh, but rebase merge isn't supported. I'll try to make the single commit message reasonable instead...

PR #7265, opened 2020-11-04
Compat data for scrollbar-gutter

felipeerias:
This PR adds compat data for the scrollbar-gutter CSS property, defined in the CSS4 Overflow spec.

This feature will be available in Chrome from version 88.0.4289.0 onwards (it is already available in Chrome Canary) by enabling the "Experimental Web Platform features" flag.

* Spec: https://drafts.csswg.org/css-overflow-4/#scrollbar-gutter-property
* MDN entry: https://developer.mozilla.org/en-US/docs/Web/CSS/scrollbar-gutter
  * Related issue: https://github.com/mdn/sprints/issues/3849
* Explainer: https://github.com/felipeerias/scrollbar-gutter-explainer
* Chromium bug: https://crbug.com/710214
* Change making the feature available behind a flag: https://crrev.com/a8c0c42e8340b922689b6b802585fafc196933b7

Thank you very much for your review. This is my first contribution, I hope I haven't missed anything 🙂

### DIFF
--- a/css/properties/scrollbar-gutter.json
+++ b/css/properties/scrollbar-gutter.json
@@ -1,0 +1,66 @@
+{
+  "css": {
+    "properties": {
+      "scrollbar-width": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/scrollbar-gutter",
+          "support": {
+            "chrome": {
+              "version_added": "88.0.4289.0",
+                "flags": [
+                  {
+                    "type": "preference",
+                    "name": "Enable experimental Web Platform features"
+                  }
+                ]
+            },
+            "chrome_android": {
+              "version_added": "88.0.4289.0",
+                "flags": [
+                  {
+                    "type": "preference",
+                    "name": "Enable experimental Web Platform features"
+                  }
+                ]
+            },
+            "edge": {
+              "version_added": false
+            },
+            "firefox": {
+              "version_added": false
+            },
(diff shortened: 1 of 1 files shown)


css/properties/scrollbar-gutter.json
```diff
@@ -0,0 +1,66 @@
...
+            },
+            "samsunginternet_android": {
+              "version_added": false
+            },
+            "webview_android": {
+              "version_added": false
+            }
+          },
+          "status": {
+            "experimental": false,
```
ddbeck, line 58: Since there's only one implementation (so far) and it's not shipping generally, let's set this to experimental for now:

```suggestion
            "experimental": true,
```

css/properties/scrollbar-gutter.json
```diff
@@ -0,0 +1,66 @@
+{
+  "css": {
+    "properties": {
+      "scrollbar-width": {
```
ddbeck, line 4: Looks like a little copy-and-paste mishap here:

```suggestion
      "scrollbar-gutter": {
```

ddbeck (changes requested):
Welcome to BCD, @felipeerias, and thank you for this PR! There's two small issues to address, but otherwise this looks good. Thank you again!

felipeerias:
I have applied your suggestions to the PR, thank you very much for the review @ddbeck

ddbeck (approved):
Looks good. Thank you! 🎉

PR #7266, opened 2020-11-04
Wrap utils.tests.js test in describe()

foolip:
This way, the output will look much more sensible when additional test
files are added. With no describe(), there's no grouping.


ddbeck (approved):
Thank you! 🎉

PR #7271, opened 2020-11-04
Deprecate the link element rev attribute

sideshowbarker:
This change marks the `link` element `rev` attribute as deprecated, because at https://html.spec.whatwg.org/multipage/obsolete.html the HTML spec defines that attribute as obsolete.


ddbeck (approved):
Thank you!

PR #7272, opened 2020-11-04
Correct Safari data for HTML element APIs

vinyldarkscratch:
This PR corrects the Safari data for the HTML element APIs based upon results from the mdn-bcd-collector project (using results from Safari 3-14), including mirroring to Safari iOS.

### DIFF
--- a/api/HTMLAnchorElement.json
+++ b/api/HTMLAnchorElement.json
@@ -412,10 +412,10 @@
               "version_added": "41"
             },
             "safari": {
-              "version_added": "11.1"
+              "version_added": "14"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "14"
             },
             "samsunginternet_android": {
               "version_added": "7.2"
--- a/api/HTMLAreaElement.json
+++ b/api/HTMLAreaElement.json
@@ -412,7 +412,7 @@
               "version_added": "41"
             },
             "safari": {
-              "version_added": "11.1"
+              "version_added": false
             },
             "safari_ios": {
               "version_added": false
--- a/api/HTMLButtonElement.json
+++ b/api/HTMLButtonElement.json
@@ -654,10 +654,10 @@
               "version_added": "27"
             },
             "safari": {
-              "version_added": "10"
+              "version_added": "10.1"
             },
             "safari_ios": {
-              "version_added": "10"
+              "version_added": "10.3"
             },
             "samsunginternet_android": {
               "version_added": "4.0"
@@ -942,10 +942,10 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "3.2"
+              "version_added": "4"
             },
             "safari_ios": {
-              "version_added": "3"
+              "version_added": "3.2"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -79,7 +79,7 @@
               "version_added": "6"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "6"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
@@ -599,10 +599,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": "6"
+              "version_added": "3"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "1"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
@@ -748,10 +748,10 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "6"
+              "version_added": "3"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "1"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
@@ -796,10 +796,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": "6"
+              "version_added": "5"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "5"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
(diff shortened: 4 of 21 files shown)


api/HTMLAreaElement.json
```diff
@@ -412,7 +412,7 @@
               "version_added": "41"
             },
             "safari": {
-              "version_added": "11.1"
+              "version_added": false
```
foolip, line 415: Yep, no `referrerPolicy` in WebKit's HTMLAreaElement.idl.

api/HTMLAnchorElement.json
```diff
@@ -412,10 +412,10 @@
               "version_added": "41"
             },
             "safari": {
-              "version_added": "11.1"
+              "version_added": "14"
```
foolip, line 415: Added in https://trac.webkit.org/changeset/257707/webkit, 11.1 can't have been right.

api/HTMLButtonElement.json
```diff
@@ -654,10 +654,10 @@
               "version_added": "27"
             },
             "safari": {
-              "version_added": "10"
+              "version_added": "10.1"
```
foolip, line 657: I don't trust these changes unless we've tested both Safari 10 and 10.1. Does the updater script set a ≤ 10.1 range here, or might we be setting the wrong version in lots of places without any indication that the underlying data is incomplete?

api/HTMLButtonElement.json
```diff
@@ -942,10 +942,10 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "3.2"
+              "version_added": "4"
```
foolip, line 945: How confident are you that this wasn't supported in Safari 3? Which platform was Safari 3 tested on?

https://trac.webkit.org/changeset/35381/webkit added the `willValidate` attribute to a bunch of interfaces, and the timing of it is consistent with is being in Safari 3.2 and Safari for iOS 3.

api/HTMLElement.json
```diff
@@ -1969,10 +1969,10 @@
               "version_added": "45"
             },
             "safari": {
-              "version_added": "10"
+              "version_added": false
```
foolip, line 1972: Where did the "10" come from. WebKit source has `nonce` in some IDL files, so this is a more complicated story that needs more digging.

api/HTMLFieldSetElement.json
```diff
@@ -316,10 +316,10 @@
               "version_added": "27"
             },
             "safari": {
-              "version_added": "10"
+              "version_added": "10.1"
```
foolip, line 319: This is a minor version bump, did we test both?

api/HTMLIFrameElement.json
```diff
@@ -184,10 +184,10 @@
               "version_added": "25"
             },
             "safari": {
-              "version_added": "10"
+              "version_added": "10.1"
```
foolip, line 187: This is a minor version bump, did we test both?

api/HTMLInputElement.json
```diff
@@ -943,10 +943,10 @@
               "version_added": "27"
             },
             "safari": {
-              "version_added": "10"
+              "version_added": "10.1"
```
foolip, line 946: This is a minor version bump, did we test both?

api/HTMLElement.json
```diff
@@ -599,10 +599,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": "6"
+              "version_added": "3"
```
foolip, line 602: Since Safari 3 was the earliest version tested for this PR, can we really say this? Let's use &le;4 where it's not clear what the status of Safari 1-2 were.

api/HTMLMeterElement.json
```diff
@@ -26,7 +26,10 @@
             "version_added": "11"
           },
           "safari": {
-            "version_added": "5"
+            "version_added": "6"
```
foolip, line 29: Can you test if `<meter>` itself was also added in Safari 6 and if that entry needs updating? (Maybe split into a separate PR if it's complicated.)

api/HTMLScriptElement.json
```diff
@@ -169,10 +169,10 @@
               "version_added": "14"
             },
             "safari": {
-              "version_added": "4"
+              "version_added": "6"
```
foolip, line 172: Can you check where the original "4" came from? Perhaps the content attribute was supported earlier?

foolip (commented):
So far I've looked only at the cases where the claimed support is being shrunk, since false negatives from our tests are much more likely than false positives.

A bunch of comments, but I stopped asking about 10 vs 10.x after a while so not all of those are covered.

This is fairly big, so I'd suggest splitting off anything that cannot be easily resolved, to leave us only with simple cases.

foolip:
It would be great to separate out a "supported since earliest tested version of Safari" subset of this PR, since that's a lot of it and easy to review.

api/HTMLElement.json
vinyldarkscratch: The `HTMLElement` API wasn't supported until Safari 3, so "1-2" is not possible -- I wouldn't set it so confidently otherwise.

api/HTMLButtonElement.json
vinyldarkscratch: The collector tests both 10 and 10.1; otherwise, it would set a range to ≤10.1, which would cause no modifications to the data.

api/HTMLButtonElement.json
vinyldarkscratch: I tested these in Mac OS X on Safari 3.2.3 (which is the latest 3.2 release).

api/HTMLButtonElement.json
foolip: Sweet!

api/HTMLButtonElement.json
foolip: Continued in https://github.com/mdn/browser-compat-data/pull/.

PR #7273, opened 2020-11-04
Deprecate the area element hreflang/noHref attrs

sideshowbarker:
This change marks the `hreflang` content attribute of the `area` element and `noHref` IDL attribute of the `HTMLAreaElement` interface as deprecated, because at https://html.spec.whatwg.org/multipage/obsolete.html, the HTML spec defines both of those as obsolete.


ddbeck (approved):
Looks good! 🎉

PR #7274, opened 2020-11-04
Deprecate the menu element label & type attributes

sideshowbarker:
This change marks the `menu` element `label` and `type` attributes deprecated, because at https://html.spec.whatwg.org/multipage/obsolete.html, the HTML spec defines those attributes as obsolete.


ddbeck (approved):
Thank you!

(I wonder if it isn't time to remove that type_toolbar feature.)

PR #7276, opened 2020-11-04
Deprecate object element typemustmatch attribute

sideshowbarker:
This change marks the `object` element `typemustmatch` attribute deprecated, because at https://html.spec.whatwg.org/multipage/obsolete.html, the HTML spec defines that attribute as obsolete.


ddbeck (approved):
Looks good. Thank you! 🎉

PR #7277, opened 2020-11-04
Automatically include correct version in statistics script

vinyldarkscratch:
This PR updates the statistics script to automatically include the current version set in `package.json`.

### DIFF
--- a/scripts/statistics.js
+++ b/scripts/statistics.js
@@ -176,7 +176,7 @@ const printStats = (stats, folder) => {
   }
 
   console.log(
-    chalk`{bold Status as of version ${version} (released on 2020-MM-DD) for ${
+    chalk`{bold Status as of version ${process.env.npm_package_version} (released on 2020-MM-DD) for ${
       folder ? `${folder}/ directory` : 'web platform features'
     }}: \n`,
   );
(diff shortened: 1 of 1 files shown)


scripts/statistics.js
```diff
@@ -6,6 +6,7 @@
 const chalk = require('chalk');
 
 const bcd = require('..');
+const version = require('../package.json').version;
```
ddbeck, line 9: We can skip a require with `process.env.npm_package_version`:

```suggestion
```

scripts/statistics.js
```diff
@@ -175,7 +176,7 @@ const printStats = (stats, folder) => {
   }
 
   console.log(
-    chalk`{bold Status as of version 1.0.xx (released on 2020-MM-DD) for ${
+    chalk`{bold Status as of version ${version} (released on 2020-MM-DD) for ${
```
ddbeck, line 179:
```suggestion
    chalk`{bold Status as of version ${process.env.npm_package_version} (released on 2020-MM-DD) for ${
```

ddbeck (changes requested):
One small suggestion in line comments. Thanks, @vinyldarkscratch!

scripts/statistics.js
vinyldarkscratch: Ooh, I didn't know about that, I'll start using that in my other projects too!

ddbeck (approved):
Thank you! 🎉

PR #7288, opened 2020-11-05
Correct Safari data for HTML element APIs (set to Safari 3)

vinyldarkscratch:
This PR corrects the Safari data for the HTML element APIs based upon results from the mdn-bcd-collector project (using results from Safari 3-14), including mirroring to Safari iOS.  This particular PR is a cherry-pick for all the changes that set it to Safari 3.

Helpful tip: anything set to "Chrome 1" is guaranteed to be in Safari 4 or below (based upon WebKit versions).

### DIFF
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -599,10 +599,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": "6"
+              "version_added": "3"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "1"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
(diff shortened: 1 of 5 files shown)


foolip:
> Helpful tip: anything set to "Chrome 1" is guaranteed to be in Safari 3 or below (based upon WebKit versions).

Chrome 1 was based on WebKit 528, so that should only imply that it was *likely* in Safari 4 (WebKit 530.17) and that we don't know about Safari 3.2 (WebKit 525.26) and earlier.

If we haven't tested Safari 1 and 2, should we not use &le;4 for all of these changes?

vinyldarkscratch:
Oops, my bad -- I meant to type "4" for that hint.  Alas, the HTMLElement API itself was supported since Safari 3, so I'm confident in setting the results to "3".

foolip:
@vinyldarkscratch I see, so the whole `HTMLElement` hierarchy of APIs were introduced in Safari 3. This makes sense then. I've filed https://github.com/mdn/browser-compat-data/issues/7305 about some outliers that would be good to fix.

foolip (approved):
I've gone over the changes with https://github.com/mdn/browser-compat-data/pull/6862 since they're very uniform. The list of touched entries make sense to me, I can't see anything anachronistic that gives me concern.

PR #7289, opened 2020-11-05
Correct Safari data for HTML element APIs (for Safari 3.1 to 4)

vinyldarkscratch:
This PR corrects the Safari data for the HTML element APIs based upon results from the mdn-bcd-collector project (using results from Safari 3-14), including mirroring to Safari iOS.  This particular PR is a cherry-pick for all the changes that set it to Safari 3.1, 3.2, or 4.

### DIFF
--- a/api/HTMLButtonElement.json
+++ b/api/HTMLButtonElement.json
@@ -942,10 +942,10 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "3.2"
+              "version_added": "4"
             },
             "safari_ios": {
-              "version_added": "3"
+              "version_added": "3.2"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
--- a/api/HTMLMediaElement.json
+++ b/api/HTMLMediaElement.json
@@ -499,7 +499,7 @@
               "version_added": "4"
             },
             "safari_ios": {
-              "version_added": "3.2"
+              "version_added": "3"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
--- a/api/HTMLMediaElement.json
+++ b/api/HTMLMediaElement.json
@@ -300,10 +300,10 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "6"
+              "version_added": "3.2"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "3"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
@@ -348,10 +348,10 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "6"
+              "version_added": "3.2"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "3"
             },
             "samsunginternet_android": {
               "version_added": "4.0"
@@ -496,10 +496,10 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "6"
+              "version_added": "4"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "3.2"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
@@ -644,10 +644,10 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "6"
+              "version_added": "3.2"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "3"
             },
             "samsunginternet_android": {
               "version_added": "4.0"
@@ -789,10 +789,10 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "6"
+              "version_added": "3.2"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "3"
             },
             "samsunginternet_android": {
               "version_added": "4.0"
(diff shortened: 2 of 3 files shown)


api/HTMLButtonElement.json
```diff
@@ -942,10 +942,10 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "3.2"
+              "version_added": "4"
```
foolip, line 945: Quoting from https://github.com/mdn/browser-compat-data/pull/7272/files#r517971169:

@foolip:

> How confident are you that this wasn't supported in Safari 3? Which platform was Safari 3 tested on?
>
> https://trac.webkit.org/changeset/35381/webkit added the `willValidate` attribute to a bunch of interfaces, and the timing of it is consistent with is being in Safari 3.2 and Safari for iOS 3.

@vinyldarkscratch:

> I tested these in Mac OS X on Safari 3.2.3 (which is the latest 3.2 release).

http://mdn-bcd-collector.appspot.com/tests/api/HTMLButtonElement/willValidate looks good, so I don't think the test is at fault.

Timing of the change aside, it looks like this was simply after the Safari 3.2 branch, it shows up between Safari 3.2.1 and Safari 4.0:
https://trac.webkit.org/browser/webkit/releases/Apple/Safari%203.2.1/WebCore/html/HTMLButtonElement.idl
https://trac.webkit.org/browser/webkit/releases/Apple/Safari%204.0/WebCore/html/HTMLButtonElement.idl

It's excellent that our tests are turning out to be correct instead of our inferences from source code! :D

api/HTMLMediaElement.json
```diff
@@ -300,10 +300,10 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "6"
+              "version_added": "3.2"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "3"
```
foolip, line 306: Media playback is a very platform specific thing, so I don't trust mirroring from Safari macOS when it comes to `HTMLMediaElement`, especially its initial launch.

https://caniuse.com/video claims it's since Safari iOS 3.2, but I tried http://mdn-bcd-collector.appspot.com/tests/api/HTMLMediaElement on the oldest iOS simulator BrowserStack has, which is iOS 3.0, and the interfaces are there. The WebKit version in the UA string was 528.18. So this seems correct!

api/HTMLMediaElement.json
```diff
@@ -496,10 +496,10 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "6"
+              "version_added": "4"
```
foolip, line 499: Are you sure it wasn't present earlier? See below comment for iOS.

api/HTMLMediaElement.json
```diff
@@ -496,10 +496,10 @@
...
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "3.2"
```
foolip, line 502:
```suggestion
              "version_added": "3"
```

Based on http://mdn-bcd-collector.appspot.com/tests/api/HTMLMediaElement in iOS 3 simulator.

foolip:
Were you able to test all of Safari 3, 3.1 and 3.2 for this?

vinyldarkscratch:
I was, yes!

api/HTMLMediaElement.json
```diff
@@ -3538,10 +3538,10 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "6"
+              "version_added": "3.2"
```
foolip, line 3541: That the `src` attribute should have shipped one release after the rest of the API seems fishy, but not impossible.

https://trac.webkit.org/changeset/27277/webkit shows it all landing at once, but that was at WebKit trunk 523.11.0, so I think the Safari support was released from a branch first and the WebKit source history shouldn't be trusted here.

Still, how confident are about the distinctions between Safari 3.1 and 3.1 in this PR? Were both tested, and what's the UA string of both?

foolip (commented):
It's my old friend `HTMLMediaElement`. Super extra attention to this one :)

api/HTMLMediaElement.json
vinyldarkscratch: I'm positive on this one, yeah -- the Chrome version this feature was first supported in provides additional confirmation (by matching WebKit versions).

api/HTMLMediaElement.json
vinyldarkscratch: Both were indeed tested, yeah.  Admittedly, due to internet connection issues in the Mac-based VMs for Safari 3.x, I'm not able to _actually_ submit reports to the collector; in other words, they were manually written up, though I used the mdn-bcd-collector to see the results.  I'm testing in Safari 3.1.2 and 3.2.3, however (which are the latest patch versions for each, as per [Wikipedia](https://en.wikipedia.org/wiki/Safari_version_history#Safari_3)).

foolip: Alright, then I suppose they might have forgotten to add the reflected `src` attribute initially, which is an easy enough mistake to make, and then fixed it before anything was merged into WebKit.

api/HTMLMediaElement.json
foolip: Given the nature of these APIs I wouldn't trust anything based on what Chrome supported when, this was all behind an ifdef in WebKit back then, and as we've seen there were differences on release branches compared to what later showed up on WebKit trunk.

But, since you did test Safari 3.2 and 4 for this PR, I'm confident in the results. It makes some sense, `canPlayType` wasn't really necessary in a first version, it's there to detect differing codec support between browsers. (I don't know exactly how it came about, it was before my time.)

foolip (approved):

PR #7290, opened 2020-11-05
Correct Safari data for HTML element APIs (for Safari 5 to 6.1)

vinyldarkscratch:
This PR corrects the Safari data for the HTML element APIs based upon results from the mdn-bcd-collector project (using results from Safari 3-14), including mirroring to Safari iOS. This particular PR is a cherry-pick for all the changes that set it to Safari 5 to 6.1.

### DIFF
--- a/api/HTMLMeterElement.json
+++ b/api/HTMLMeterElement.json
@@ -26,7 +26,10 @@
             "version_added": "11"
           },
           "safari": {
-            "version_added": "5"
+            "version_added": "6"
+          },
+          "safari_ios": {
+            "version_added": "6"
           }
         },
         "status": {
@@ -61,7 +64,10 @@
               "version_added": "11"
             },
             "safari": {
-              "version_added": "5"
+              "version_added": "6"
+            },
+            "safari_ios": {
+              "version_added": "6"
             }
           },
           "status": {
@@ -136,7 +142,10 @@
               "version_added": "11"
             },
             "safari": {
-              "version_added": "5"
+              "version_added": "6"
+            },
+            "safari_ios": {
+              "version_added": "6"
             }
           },
           "status": {
@@ -172,7 +181,10 @@
               "version_added": "11"
             },
             "safari": {
-              "version_added": "5"
+              "version_added": "6"
+            },
+            "safari_ios": {
+              "version_added": "6"
             }
           },
           "status": {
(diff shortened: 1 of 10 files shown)


api/HTMLMeterElement.json
```diff
@@ -26,7 +26,10 @@
             "version_added": "11"
           },
           "safari": {
-            "version_added": "5"
+            "version_added": "6"
```
foolip, line 29: Where did the original "5" come from, and can you check when `<meter>` itself was supported?

api/HTMLScriptElement.json
```diff
@@ -169,10 +169,10 @@
               "version_added": "14"
             },
             "safari": {
-              "version_added": "4"
+              "version_added": "6"
```
foolip, line 172: Where did the original "4" come from?

api/HTMLOListElement.json
```diff
@@ -124,10 +124,10 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "5"
+              "version_added": "6"
```
foolip, line 127: Where did the original "5" come from?

foolip (commented):
First pass I've focused on what I see with `npm run diff ../bcd-lhs/ ../bcd-rhs/ -- --filter`.

api/HTMLMeterElement.json
vinyldarkscratch: The original data comes from the wiki migration.  The element itself was introduced in Safari 6, which matches this new data.

api/HTMLScriptElement.json
vinyldarkscratch: Looks like it came from the wiki migration.

api/HTMLOListElement.json
vinyldarkscratch: Another entry from the wiki migration, looks like.

api/HTMLMeterElement.json
foolip: Thanks for checking!

api/HTMLOListElement.json
foolip: Thanks for checking!

api/HTMLScriptElement.json
foolip: Thanks for checking!

foolip (approved):
I've reviewed using my `npm run diff` prototype, the set of changes together seem sensible, nothing about it raises red flags for me.

PR #7291, opened 2020-11-05
Correct Safari data for HTML element APIs (for Safari 9 to 14)

vinyldarkscratch:
This PR corrects the Safari data for the HTML element APIs based upon results from the mdn-bcd-collector project (using results from Safari 3-14), including mirroring to Safari iOS. This particular PR is a cherry-pick for all the changes that set it to Safari 9 to 14.

### DIFF
--- a/api/HTMLAnchorElement.json
+++ b/api/HTMLAnchorElement.json
@@ -412,10 +412,10 @@
               "version_added": "41"
             },
             "safari": {
-              "version_added": "11.1"
+              "version_added": "14"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "14"
             },
             "samsunginternet_android": {
               "version_added": "7.2"
--- a/api/HTMLIFrameElement.json
+++ b/api/HTMLIFrameElement.json
@@ -184,10 +184,10 @@
               "version_added": "25"
             },
             "safari": {
-              "version_added": "10"
+              "version_added": "10.1"
             },
             "safari_ios": {
-              "version_added": "10"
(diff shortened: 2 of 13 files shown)


api/HTMLAnchorElement.json
```diff
@@ -412,10 +412,10 @@
               "version_added": "41"
             },
             "safari": {
-              "version_added": "11.1"
+              "version_added": "14"
```
foolip, line 415: Yep, this was recent: https://trac.webkit.org/changeset/257707/webkit

api/HTMLImageElement.json
```diff
@@ -1063,10 +1063,10 @@
               "version_added": "41"
             },
             "safari": {
-              "version_added": "11.1"
+              "version_added": "14"
```
foolip, line 1066: Checks out, this was also recent: https://trac.webkit.org/changeset/257572/webkit

api/HTMLIFrameElement.json
```diff
@@ -184,10 +184,10 @@
               "version_added": "25"
             },
             "safari": {
-              "version_added": "10"
+              "version_added": "10.1"
```
foolip, line 187: I wanted to check one of the 10&rarr;10.1 changes and picked this, `allowFullscreen`. The change was in https://trac.webkit.org/changeset/205686/webkit, at WebKit trunk 603.1.6. That points to Safari 10.1 indeed, Safari 10 was branched before this point. Confirmed!

foolip (approved):

PR #7292, opened 2020-11-05
Correct Safari data for HTML element APIs (for Safari unsupported)

vinyldarkscratch:
This PR corrects the Safari data for the HTML element APIs based upon results from the mdn-bcd-collector project (using results from Safari 3-14), including mirroring to Safari iOS. This particular PR is a cherry-pick for all the changes that set it to `false`.

### DIFF
--- a/api/HTMLAreaElement.json
+++ b/api/HTMLAreaElement.json
@@ -412,7 +412,7 @@
               "version_added": "41"
             },
             "safari": {
-              "version_added": "11.1"
+              "version_added": false
             },
             "safari_ios": {
               "version_added": false
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -1969,10 +1969,10 @@
               "version_added": "45"
             },
             "safari": {
-              "version_added": false
+              "version_added": "10"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "10"
             },
             "samsunginternet_android": {
               "version_added": "8.0"
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -1969,10 +1969,10 @@
               "version_added": "45"
             },
             "safari": {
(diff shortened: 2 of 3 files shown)


api/HTMLAreaElement.json
```diff
@@ -412,7 +412,7 @@
               "version_added": "41"
             },
             "safari": {
-              "version_added": "11.1"
+              "version_added": false
```
foolip, line 415: Yep, no `referrerPolicy` in WebKit's HTMLAreaElement.idl.

api/HTMLLinkElement.json
```diff
@@ -172,7 +172,7 @@
               "version_added": "43"
             },
             "safari": {
-              "version_added": "11.1"
+              "version_added": false
```
foolip, line 175: Confirmed no `referrerPolicy` in WebKit's HTMLLinkElement.idl.

api/HTMLElement.json
```diff
@@ -1969,10 +1969,10 @@
               "version_added": "45"
             },
             "safari": {
-              "version_added": "10"
+              "version_added": false
```
foolip, line 1972: This (`nonce`) should be `partial_implementation` for some range of versions, as WebKit as `nonce` on `HTMLLinkElement`, `HTMLScriptElement` and `HTMLStyleElement`, with a FIXME to add it to their `HTMLOrForeignElement` mixin.

Can you add custom IDL for `nonce` on those three interfaces and test how far back it goes? Maybe it's Safari 10, just claimed here :)

vinyldarkscratch: Can confirm that they're on the individual elements; reverting and cherry-picking.

vinyldarkscratch: Defered to #7303.

foolip (approved):

PR #7295, opened 2020-11-06
Update Credential API data for Edge

vinyldarkscratch:
This PR updates the Edge data for four Credential APIs based upon results from the mdn-bcd-collector project (tested in Edge 13-85).


foolip (approved):
Excellent! Seems like quite a useful clarification of the data in this case, that a bit of `CredentialsContainer` was supported in EdgeHTML, but not all.

PR #7296, opened 2020-11-06
Update Credential API data for Safari

vinyldarkscratch:
This PR updates the Safar data for four Credential APIs based upon results from the mdn-bcd-collector project (tested in Safari 6-14).  Additionally, this corrects the data for the CredentialsContainer API for Safari iOS based upon manual testing (simulator for iPadOS 13.0).

### DIFF
--- a/api/CredentialsContainer.json
+++ b/api/CredentialsContainer.json
@@ -32,7 +32,7 @@
             "version_added": "13"
           },
           "safari_ios": {
-            "version_added": "13.3"
+            "version_added": "13"
           },
           "samsunginternet_android": {
             "version_added": "5.0"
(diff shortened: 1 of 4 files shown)


api/CredentialsContainer.json
```diff
@@ -32,7 +32,7 @@
             "version_added": "13"
           },
           "safari_ios": {
-            "version_added": "13.3"
+            "version_added": "13"
```
foolip, line 35: OK, so `CredentialsContainer` was supported earlier in Safari iOS than the previous recorded, and that matches the Safari macOS data so is very plausible.

foolip (approved):

PR #7300, opened 2020-11-06
Mirror Opera data for ContentIndex(Event) and Element.getAnimations()

foolip:
Fixes https://github.com/mdn/browser-compat-data/issues/6601.


ddbeck (approved):
Thank you! 🎉

PR #7308, opened 2020-11-06
Update DOMParser compat data for iOS Safari

camelid:
Update DOMParser compat data for iOS Safari. I tested it and the constructor works, as does parsing from HTML.

I used Safari on iOS version 14.1.

### DIFF
--- a/api/DOMParser.json
+++ b/api/DOMParser.json
@@ -80,7 +80,7 @@
               "version_added": "3.2"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": true
             },
             "samsunginternet_android": {
               "version_added": "1.0"
@@ -175,7 +175,7 @@
                 "version_added": "9.1"
               },
               "safari_ios": {
-                "version_added": null
+                "version_added": true
               },
               "samsunginternet_android": {
                 "version_added": "3.0"
--- a/api/DOMParser.json
+++ b/api/DOMParser.json
@@ -80,7 +80,7 @@
               "version_added": "3.2"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "2"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
@@ -175,7 +175,7 @@
                 "version_added": "9.1"
               },
               "safari_ios": {
-                "version_added": true
+                "version_added": "9.3"
               },
               "samsunginternet_android": {
                 "version_added": "3.0"


api/DOMParser.json
```diff
@@ -80,7 +80,7 @@
               "version_added": "3.2"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": true
```
sideshowbarker, line 83:
```suggestion
              "version_added": "2"
```

api/DOMParser.json
```diff
@@ -175,7 +175,7 @@
                 "version_added": "9.1"
               },
               "safari_ios": {
-                "version_added": null
+                "version_added": true
```
sideshowbarker, line 178:
```suggestion
                "version_added": "9.3"
```

sideshowbarker (changes requested):
We can set actual version numbers, based on the (desktop) Safari version numbers

camelid:
I know we can set actual version numbers, I just don't know what version support was introduced in. And iOS Safari is different from desktop Safari, so I'm not sure how using desktop Safari numbers is correct.

sideshowbarker:
> I know we can set actual version numbers,

It’s not just that we can — we need to. See #3555; we have an ongoing effort to replace all `true` values in the existing data with real version numbers — and so we’re trying not to add any new `true` values unless we really can’t identify a likely version number. 


> I just don't know what version support was introduced in.

Based on the existing data we have, it appears that the constructor support went into WebKit version 525.26 or before, and the text/html support in WebKit version 601.5.17 — and that’s why we have the version numbers for desktop Safari marked as 3.2 and 9.1.

And because iOS Safari and Safari use the same engine, that would mean the constructor support went into iOS Safari in whatever version shipped with WebKit version 525.26, and the text/html support in whatever version shipped with WebKit version 601.5.17 — which per https://github.com/mdn/browser-compat-data/blob/master/browsers/safari_ios.json are iOS Safari 2 and iOS Safari 9.3

> And iOS Safari is different from desktop Safari, so I'm not sure how using desktop Safari numbers is correct.

As detailed above, it’s not about using the desktop Safari numbers — it’s about using the iOS Safari versions which correspond to the same WebKit engine versions as we have for existing desktop Safari data.

api/DOMParser.json
saschanaz: Does it mean `safari` value should also be fixed?

camelid:
Okay, thanks for the detailed explanation! I'll apply your suggestions.

camelid:
I can squash too if you want.

sideshowbarker:
> I can squash too if you want.

Thanks, but no need — in this repo, all PR branches always get squashed before merging (though I’d personally prefer we at least had the option of doing rebase-without-squashing merges sometimes…)

sideshowbarker (approved):

sideshowbarker:

ddbeck:
> though I’d personally prefer we at least had the option of doing rebase-without-squashing merges sometimes…

@sideshowbarker open an issue, please! It wouldn't hurt to hear about what's lacking from our current approach.

sideshowbarker:
> @sideshowbarker open an issue, please! It wouldn't hurt to hear about what's lacking from our current approach.

OK, I’ve raised https://github.com/mdn/browser-compat-data/issues/7368 for it.

PR #7313, opened 2020-11-07
WebExtension API browsingData.remove cookieStoreId

evilpie:
This Firefox only API was introduced in https://bugzilla.mozilla.org/show_bug.cgi?id=1670811


ddbeck (approved):
Thank you! 🎉

PR #7314, opened 2020-11-08
Remove "x", "y", and "z" from api.LinearAccelerationSensor

vinyldarkscratch:
This PR removes the `x`, `y`, and `z` properties from the `LinearAccelerationSensor` API.  These properties are available via the `Accelerometer` API, of which this API inherits.


foolip:
This will require sorting out of the MDN pages too. For some reason we have both:
https://developer.mozilla.org/en-US/docs/Web/API/Accelerometer/x
https://developer.mozilla.org/en-US/docs/Web/API/LinearAccelerationSensor/x

Looking at their history they were added by @jpmedley. Joe, what should we do here?

jpmedley (approved):

ddbeck:
@jpmedley What's the story with the content here, mentioned by @foolip in https://github.com/mdn/browser-compat-data/pull/7314#issuecomment-724585346? If we're merging a removal of a feature, I need to have a plan to deal with any pages left stranding.

vinyldarkscratch:
I suggest that we remove the pages for `LinearAccelerationSensor.x/y/z` and mark them as a redirect to `Accelerometer.x/y/z`.  Comparing the pages, it looks like they're basically the same page, just with a few slight variations in wording to refer to each individual interface, plus since `LinearAccelerationSensor` is a child of `Accelerometer`, it makes sense to redirect to the parent's page.

foolip:
@ddbeck does that sound good to you? If so I can try to clean it up.

ddbeck:
Yes, this sounds good to me. Thank you, @vinyldarkscratch and @foolip 👍

jpmedley:
The redirects aren't needed. On the `LinearAccelerationSensor` interface page, the list of properties needs to be replaced with "Inherits methods from its ancestor, LinearAccelerationSensor." I'll fix the text.

(Yikes! It looks like I was the one who did it the wrong way.)

jpmedley:
And I guess the redirects are needed because these duplicate pages might have been indexed by Google already. I don't remember how to do that.

jpmedley:
Fixed. https://wiki.developer.mozilla.org/en-US/docs/Web/API/LinearAccelerationSensor

foolip:
I have created the redirects:
https://wiki.developer.mozilla.org/en-US/docs/Web/API/LinearAccelerationSensor/x
https://wiki.developer.mozilla.org/en-US/docs/Web/API/LinearAccelerationSensor/y
https://wiki.developer.mozilla.org/en-US/docs/Web/API/LinearAccelerationSensor/z

PR #7315, opened 2020-11-08
Rename api.MediaKeyStatusMap.iterator to @@iterator

vinyldarkscratch:
This PR renames the "iterator" feature of the MediaKeyStatusMap API to the proper "@@iterator".

### DIFF
--- a/api/MediaKeyStatusMap.json
+++ b/api/MediaKeyStatusMap.json
@@ -239,53 +239,6 @@
           }
         }
       },
-      "iterator": {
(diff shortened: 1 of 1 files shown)


foolip (approved):
It looks like there are lots of other changes here, but that's just because the iterator entry had to be moved, and in the resulting diff that's expressed as a bunch of changes throughout. I've confirmed locally that undoing the ordering, only the "iterator" &rarr; "@@iterator" change remains.

PR #7316, opened 2020-11-08
Remove api.Node.outerText

vinyldarkscratch:
This PR removes `api.Node.outerText`.  This property is set to all false (or null), and the MDN document just says "See HTMLElement.outerText".

### DIFF
--- a/api/Node.json
+++ b/api/Node.json
@@ -1619,54 +1619,6 @@
           }
         }
       },
-      "outerText": {
-        "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/Node/outerText",
-          "support": {
-            "chrome": {
-              "version_added": false
-            },
-            "chrome_android": {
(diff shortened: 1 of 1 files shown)


foolip (approved):
I've turned https://wiki.developer.mozilla.org/en-US/docs/Web/API/Node/outerText into a redirect.

foolip:
And to be clear, I do not believe `outerText` has ever been implemented on `Node` in any browser. In Confluence, which goes back roughly to 2014, it's not ever been detected outside of HTML elements:
http://web-confluence.appspot.com/#!/catalog?releases=%5B%22Edge_18.17763_Windows_10.0%22,%22Chrome_85.0.4183.83_Windows_10.0%22,%22Firefox_80.0_Windows_10.0%22,%22Safari_14.0_OSX_10.15.6%22%5D&q=%22outertext%22

PR #7320, opened 2020-11-09
Update Safari data for CSSValue + CSSValueList

vinyldarkscratch:
This PR uses the mdn-bcd-collector project to update the Safari data for the CSSValue and CSSValueList APIs, including mirroring to Safari iOS.  (Note: Ranges for ≤4 were actually determined as ≤3, but I haven't tested any further back.)

### DIFF
--- a/api/CSSValue.json
+++ b/api/CSSValue.json
@@ -31,10 +31,10 @@
             "version_added": null
           },
           "safari": {
-            "version_added": null
+            "version_added": "≤4"
           },
           "safari_ios": {
-            "version_added": null
+            "version_added": "1"
           },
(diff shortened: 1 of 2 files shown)


api/CSSValue.json
```diff
@@ -31,10 +31,10 @@
             "version_added": null
           },
           "safari": {
-            "version_added": null
+            "version_added": "≤4"
```
foolip, line 34: Confirmed with http://mdn-bcd-collector.appspot.com/tests/api/CSSValue that it's still there in STP. A Safari-only API at this point then, interesting!

foolip (approved):

PR #7321, opened 2020-11-09
fit content function

rachelandrew:
I am working on splitting out the `fit-content` keyword for width from the `fit-content()` function for width in https://github.com/mdn/sprints/issues/3723

This PR adds the data for the `fit-content()` function for width, which no-one yet supports, however given it is confusing people I think it worth showing that lack of support. As then on the function page I can show that it is well supported for track sizing in grid layout, and not at all for use as a general length, which would have helped the poster of that issue.

### DIFF
--- a/css/properties/width.json
+++ b/css/properties/width.json
@@ -286,6 +286,54 @@
             }
           }
         },
+        "fit-content-function": {
+          "__compat": {
+            "description": "<code>fit-content()</code>",
+            "support": {
+              "chrome": {
+                "version_added": false
+              },
+              "chrome_android": {
+                "version_added": false
+              },
+              "edge": {
+                "version_added": false
+              },
+              "firefox": {
+                "version_added": false
+              },
+              "firefox_android": {
+                "version_added": false
+              },
+              "ie": {
+                "version_added": false
+              },
+              "opera": {
+                "version_added": false
+              },
+              "opera_android": {
+                "version_added": false
+              },
+              "safari": {
+                "version_added": false
+              },
+              "safari_ios": {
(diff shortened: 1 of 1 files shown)


css/properties/width.json
```diff
@@ -286,6 +286,54 @@
             }
           }
         },
+        "fit-content-function": {
```
ddbeck, line 289: I don't think we actually document this anywhere, but we typically use underscores for word spaces and dashes literally.

```suggestion
        "fit-content_function": {
```

ddbeck (changes requested):
This is a weird one, though I see the logic in it. Is this something that might exist in the future? It might be nice to link to implementation bugs or something, mostly to prevent me from removing the whole feature mistakenly.

One little style nit in a line comment. Thank you!

rachelandrew:
@ddbeck It's in the level 3 specification, I can't find anyone with an implementation bug. However given it does seem to cause confusion (and the other intrinsic sizes are implemented) I think it worth adding.

css/properties/width.json
```diff
@@ -286,6 +286,54 @@
...
+                "version_added": false
+              },
+              "chrome_android": {
+                "version_added": false
+              },
+              "edge": {
+                "version_added": false
+              },
+              "firefox": {
+                "version_added": false
```
saschanaz, line 303: Found a tracking bug:

```suggestion
                "version_added": false,
                "notes": "See <a href='https://bugzil.la/1312588'>bug 1312588</a>."
```

ddbeck (approved):
Thanks, y'all! 🎉

PR #7322, opened 2020-11-09
Add Firefox data for HTML element APIs

vinyldarkscratch:
This PR adds real data for the HTML element APIs for Firefox using results from the mdn-bcd-collector project (Firefox 4-82), as well as manual testing (Firefox 1-3.6.9).

Additionally, this also sets `api.HTMLSelectElement` from `2` to `1` to fix consistency issues.

### DIFF
--- a/api/HTMLIFrameElement.json
+++ b/api/HTMLIFrameElement.json
@@ -61,10 +61,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": true
+              "version_added": "1"
             },
             "firefox_android": {
-              "version_added": true
+              "version_added": "4"
             },
             "ie": {
               "version_added": "5.5"
@@ -279,7 +279,7 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": true
+              "version_added": "1"
             },
             "firefox_android": {
               "version_added": "4"
@@ -375,10 +375,10 @@
               "version_added": "79"
             },
             "firefox": {
-              "version_added": null
+              "version_added": false
             },
             "firefox_android": {
-              "version_added": null
+              "version_added": false
             },
             "ie": {
               "version_added": false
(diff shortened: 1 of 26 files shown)


api/HTMLIFrameElement.json
```diff
@@ -375,10 +375,10 @@
               "version_added": "79"
             },
             "firefox": {
-              "version_added": null
+              "version_added": false
```
foolip, line 378: OK, so https://developer.mozilla.org/en-US/docs/Web/API/HTMLIFrameElement/csp / https://w3c.github.io/webappsec-cspee/#dom-htmliframeelement-csp seems to not be in Firefox or Safari. https://wpt.fyi/results/content-security-policy/embedded-enforcement/iframe-csp-attribute.html agrees.

Oh well :(

api/HTMLMediaElement.json
```diff
@@ -1757,7 +1760,10 @@
               "version_added": false
             },
             "firefox": {
-              "version_added": null
+              "version_added": "15"
```
foolip, line 1763: There we go, this `mozAudioCaptured` no longer looks like something that should be removed.

api/HTMLMediaElement.json
```diff
@@ -3481,7 +3502,10 @@
               "version_added": "17"
             },
             "firefox": {
-              "version_added": null
+              "version_added": false
```
foolip, line 3505: Yep: https://bugzilla.mozilla.org/show_bug.cgi?id=1498512

api/HTMLUnknownElement.json
```diff
@@ -14,10 +14,10 @@
             "version_added": "12"
           },
           "firefox": {
-            "version_added": true
+            "version_added": "1"
```
foolip, line 17: I took another look at http://mdn-bcd-collector.appspot.com/tests/api/HTMLUnknownElement in Firefox 3 to convince myself the results can be trusted. I still think the test is good.

foolip (approved):

PR #7323, opened 2020-11-09
Correct Firefox data for HTML element APIs

vinyldarkscratch:
This PR corrects the HTML element API data for Firefox using results from the mdn-bcd-collector project (Firefox 4-82), aided with a little manual testing (Firefox 1-3.6.9).

### DIFF
--- a/api/HTMLAnchorElement.json
+++ b/api/HTMLAnchorElement.json
@@ -445,10 +445,10 @@
               "version_added": "18"
             },
             "firefox": {
-              "version_added": "1"
+              "version_added": "30"
             },
             "firefox_android": {
-              "version_added": "4"
+              "version_added": "30"
             },
             "ie": {
               "version_added": false
--- a/api/HTMLAudioElement.json
+++ b/api/HTMLAudioElement.json
@@ -112,11 +112,11 @@
             },
             "firefox": {
               "version_added": "4",
-              "version_removed": "31"
+              "version_removed": "28"
             },
             "firefox_android": {
-              "version_added": true,
-              "version_removed": "31"
+              "version_added": "4",
+              "version_removed": "28"
             },
             "ie": {
               "version_added": false
@@ -163,11 +163,11 @@
             },
             "firefox": {
               "version_added": "4",
-              "version_removed": "31"
+              "version_removed": "28"
             },
             "firefox_android": {
-              "version_added": true,
-              "version_removed": "31"
+              "version_added": "4",
+              "version_removed": "28"
             },
             "ie": {
               "version_added": false
@@ -213,10 +213,12 @@
               "version_added": false
             },
             "firefox": {
-              "version_added": "4"
+              "version_added": "4",
+              "version_removed": "28"
             },
             "firefox_android": {
-              "version_added": true
+              "version_added": "4",
+              "version_removed": "28"
             },
             "ie": {
               "version_added": false
--- a/api/HTMLCanvasElement.json
+++ b/api/HTMLCanvasElement.json
@@ -613,10 +613,11 @@
             },
             "firefox": {
               "version_added": "13",
-              "version_removed": "43"
+              "version_removed": "23"
             },
             "firefox_android": {
-              "version_added": false
+              "version_added": "14",
+              "version_removed": "23"
             },
             "ie": {
               "version_added": false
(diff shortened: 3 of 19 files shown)


api/HTMLAnchorElement.json
```diff
@@ -445,10 +445,10 @@
               "version_added": "18"
             },
             "firefox": {
-              "version_added": "1"
+              "version_added": "30"
```
foolip, line 448: This is removing claimed support, but http://mdn-bcd-collector.appspot.com/tests/api/HTMLAnchorElement/relList is as simple as a test could be for this, and I've tried it manually on Firefox 29 and 30, this is correct.

api/HTMLAudioElement.json
```diff
@@ -112,11 +112,11 @@
             },
             "firefox": {
               "version_added": "4",
-              "version_removed": "31"
+              "version_removed": "28"
```
foolip, line 115: This has been gone for long enough now that it can be removed from BCD. https://developer.mozilla.org/docs/Web/API/HTMLAudioElement/mozCurrentSampleOffset is 404 too. But fixing the data what we're up to here, and that's OK.

api/HTMLCanvasElement.json
```diff
@@ -766,10 +767,10 @@
               }
             ],
             "firefox": {
-              "version_added": "19"
+              "version_added": "18"
             },
             "firefox_android": {
-              "version_added": "4"
+              "version_added": "18"
```
foolip, line 773: Yep, I'll trust http://mdn-bcd-collector.appspot.com/tests/api/HTMLCanvasElement/toBlob over the existing data here, false negatives aren't plausible from how the test works.

api/HTMLElement.json
```diff
@@ -1493,10 +1493,12 @@
               "version_added": false
             },
             "firefox": {
-              "version_added": "6"
+              "version_added": "16",
+              "version_removed": "49"
```
foolip, line 1497: Alright, with this change it's clear that we can remove the Microdata API from BCD, it's been gone from all browsers for a long enough. Not in this PR though.

api/HTMLElement.json
```diff
@@ -2242,10 +2254,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "3"
+              "version_added": "9"
```
foolip, line 2257: Can you split out the `onfoo` changes and update them together with the `foo_event` entries which are sprinkled around?

api/HTMLMediaElement.json
```diff
@@ -966,10 +966,10 @@
               "version_added": "79"
             },
             "firefox": {
-              "version_added": "20"
+              "version_added": false
```
foolip, line 969: Where did the original "20" come from? This is the only thing being set to false in this PR. I suspect it's correct, but warrants a closer look.

api/HTMLMediaElement.json
```diff
@@ -1841,10 +1841,12 @@
               "version_added": false
             },
             "firefox": {
-              "version_added": "4"
+              "version_added": "4",
+              "version_removed": "28"
```
foolip, line 1845: api.HTMLMediaElement.moz* are also ready for cleanup, then.

api/HTMLSourceElement.json
```diff
@@ -117,10 +117,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "3.5"
+              "version_added": "15"
```
foolip, line 120: This and https://developer.mozilla.org/en-US/docs/Web/HTML/Element/source might warrant a note, because at different times the `media` attribute on `<source>` has been used for `<video>` and `<picture>`, and I believe the latter is the only thing left supported in any engine. But `<picture>` wasn't added in Firefox 15, if https://developer.mozilla.org/en-US/docs/Web/HTML/Element/picture#Browser_compatibility is to be trusted.

It might be that for some time the IDL attribute was there and merely reflecting the content attribute, even though it didn't do anything. If so, then a `partial_implementation` range seems appropriate.

Probably split this out?

foolip (commented):
All looks good except the entries I've commented out. Splitting those out would allow us to merge all the other changes fast.

api/HTMLMediaElement.json
vinyldarkscratch: The original data had come from the initial wiki migration, it looks like.

foolip: OK, thanks!

foolip (approved):
OK, I also reverted api.HTMLElement.on* and took another pass, all looks good now.

PR #7326, opened 2020-11-10
Add real data for Safari for RTC APIs

vinyldarkscratch:
This PR adds real values for the RTC APIs for Safari and Safari iOS/iPadOS based upon results from the mdn-bcd-collector project (Safari 3-14 and Safari iOS 3-14).

### DIFF
--- a/api/RTCErrorEvent.json
+++ b/api/RTCErrorEvent.json
@@ -29,10 +29,10 @@
             "version_added": null
           },
           "safari": {
-            "version_added": null
+            "version_added": false
           },
           "safari_ios": {
-            "version_added": "6"
+            "version_added": false
           },
           "samsunginternet_android": {
             "version_added": null
@@ -76,10 +76,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": false
             },
             "safari_ios": {
-              "version_added": "6"
+              "version_added": false
             },
             "samsunginternet_android": {
               "version_added": null
--- a/api/RTCIceCandidate.json
+++ b/api/RTCIceCandidate.json
@@ -29,10 +29,10 @@
             "version_added": true
           },
           "safari": {
-            "version_added": true
+            "version_added": "11"
           },
           "safari_ios": {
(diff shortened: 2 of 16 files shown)


api/RTCErrorEvent.json
```diff
@@ -29,10 +29,10 @@
             "version_added": null
           },
           "safari": {
-            "version_added": null
+            "version_added": false
           },
           "safari_ios": {
-            "version_added": "6"
+            "version_added": false
```
foolip, line 35: Yep, no RTCErrorEvent in WebKit source. The "6" there was definitely wrong, Safari iOS back in 2012 did not have anything around WebRTC :)

api/RTCIceCandidate.json
```diff
@@ -134,10 +134,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": false
```
foolip, line 137: Yep, https://trac.webkit.org/changeset/266698/webkit added `address` and other attributes recently, they haven't shipped yet.

api/RTCPeerConnection.json
```diff
@@ -274,10 +274,12 @@
               }
             ],
             "safari": {
-              "version_added": null
+              "version_added": "11",
```
foolip, line 277: Yeah, a bunch of things seem to have been shipped in Safari 11 but then removed in 12, because there was no agreement on the spec side that these things should be defined.

api/RTCPeerConnectionIceErrorEvent.json
```diff
@@ -29,10 +29,10 @@
             "version_added": null
           },
           "safari": {
-            "version_added": null
+            "version_added": false
```
foolip, line 32: Yep, no sign of `RTCPeerConnectionIceErrorEvent` in WebKit source, and I did some digging and failed to find it under another name either.

foolip (approved):
A big change, but I feel I've spot checked enough to have confidence it's mostly correct and definitely a net improvement to the data.

PR #7327, opened 2020-11-10
Correct Safari data for RTC APIs

vinyldarkscratch:
This PR corrects the Safari and Safari iOS data for three RTC APIs based upon results from the mdn-bcd-collector project (Safari 3-14 and Safari iOS 3-14).


foolip (approved):
Yay! (This was pretty simple to review.)

PR #7329, opened 2020-11-10
Add real data for Safari for Node API

vinyldarkscratch:
This PR adds real values for the Node API for Safari and Safari iOS/iPadOS based upon results from the mdn-bcd-collector project (Safari 3-14 and Safari iOS 3-14).

### DIFF
--- a/api/Node.json
+++ b/api/Node.json
@@ -132,10 +132,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "7"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "7"
             },
             "samsunginternet_android": {
               "version_added": true
@@ -397,7 +397,7 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "≤4"
             },
             "safari_ios": {
               "version_added": true
@@ -495,10 +495,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "7"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "7"
             },
             "samsunginternet_android": {
               "version_added": true
@@ -543,10 +543,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": false
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": false
             },
             "samsunginternet_android": {
               "version_added": false
@@ -738,10 +738,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "≤4"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "1"
             },
             "samsunginternet_android": {
               "version_added": true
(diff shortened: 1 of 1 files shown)


api/Node.json
```diff
@@ -543,10 +543,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": false
```
foolip, line 546: This is `getFeature`, and the only null&rarr;false change in this PR. I trust that it's right, and we should remove this entry since it's not implemented anywhere.

api/Node.json
```diff
@@ -397,7 +397,7 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "≤4"
```
foolip, line 400: https://trac.webkit.org/browser/webkit/trunk?rev=4552 is where Safari 1 branched from WebKit trunk, so searching the repo at that time can help determine if something was in Safari 1. Unfortunately `compareDocumentPosition` wasn't in the tree then, so I can't suggest an improvement here :)

api/Node.json
```diff
@@ -397,7 +397,7 @@
...
             },
             "safari_ios": {
               "version_added": true
```
foolip, line 403: I guess we're missing a range for Safari iOS in order to fix this?

api/Node.json
```diff
@@ -738,10 +738,10 @@
...
             },
             "safari": {
-              "version_added": true
+              "version_added": "≤4"
```
foolip, line 741:
```suggestion
              "version_added": "1"
```

Based on https://trac.webkit.org/browser/webkit/trunk/WebCore/khtml/ecma/kjs_dom.lut.h?rev=4552

api/Node.json
```diff
@@ -882,10 +882,10 @@
...
```
foolip, line 885: No suggestion here, no `isDefaultNamespace` in https://trac.webkit.org/browser/webkit/trunk?rev=4552

api/Node.json
```diff
@@ -1040,10 +1040,12 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": "≤4",
```
foolip, line 1043:
```suggestion
              "version_added": "1",
```

Based on https://trac.webkit.org/browser/webkit/trunk/WebCore/khtml/ecma/kjs_dom.lut.h?rev=4552

api/Node.json
```diff
@@ -1600,10 +1602,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "≤4"
```
foolip, line 1605:
```suggestion
              "version_added": "1"
```

Based on https://trac.webkit.org/browser/webkit/trunk/WebCore/khtml/ecma/kjs_dom.lut.h?rev=4552

foolip (approved):
I've made some suggestions, but after digging deeper, I think it'd be better to deal with all the WebKit source archeology to fix ≤4 together, a single time, once all the Safari trues and nulls are gone.

PR #7330, opened 2020-11-10
Add real data for Safari for NodeList API

vinyldarkscratch:
This PR adds real values for the NodeList API for Safari and Safari iOS/iPadOS based upon results from the mdn-bcd-collector project (Safari 3-14 and Safari iOS 3-14).

### DIFF
--- a/api/NodeList.json
+++ b/api/NodeList.json
@@ -172,10 +172,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "≤4"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "1"
             },
             "samsunginternet_android": {
               "version_added": true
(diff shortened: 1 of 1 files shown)


foolip (approved):

api/NodeList.json
```diff
@@ -268,10 +268,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "10"
```
foolip, line 271: Wait, no, this is not plausible, the `length` attribute must have been around much earlier than this.

http://mdn-bcd-collector.appspot.com/tests/api/NodeList tests the prototype only, is Safari 10 when attributes were moved to prototypes?

foolip (changes requested):

api/NodeList.json
vinyldarkscratch: Ack, my bad, I didn't mean to include this change...  Reverted!

foolip (approved):

PR #7332, opened 2020-11-10
Add data guideline: partial_implementation requires a note

ddbeck:
This makes something explicit we've largely adhered to already. A step toward completing #4162.


foolip (approved):

sideshowbarker (approved):

ddbeck:
Thank you, @foolip and @sideshowbarker!

PR #7334, opened 2020-11-10
document.visibilityState: Safari 7+ supports "prerender"

jakub-g:
I tested on BrowserStack, every version of desktop Safari from 7 to 14 and each of them actually starts with `visibilityState: prerender`

Test page: https://jg-testpage.netlify.app/document.hidden/
image

Scenario to reproduce:

1. Copy `https://jg-testpage.netlify.app/document.hidden/` to clipboard
2. Open new tab
3. Paste URL to the URL bar
4. Navigate to the URL (press enter)

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -11642,7 +11642,7 @@
                 "version_added": null
               },
               "safari": {
-                "version_added": null
+                "version_added": "7"
               },
               "safari_ios": {
                 "version_added": null


ddbeck (approved):
Looks good. Thank you! 🎉

sideshowbarker:
After looking through the WebKit change history, I think this actually shipped in Safari 6.1 — so I reckon I’ll raise a follow-up PR to get us further updated

sideshowbarker:
> After looking through the WebKit change history, I think this actually shipped in Safari 6.1 — so I reckon I’ll raise a follow-up PR to get us further updated

OK, raised #7349

PR #7337, opened 2020-11-11
Add IE 6 as ranged version

vinyldarkscratch:
This PR adds Internet Explorer 6 as a ranged version.  IE 6 was released all the way back in 2001, and is the oldest version of IE supported by BrowserStack.  While I own VMs that are able to test IE 1-5.5, these versions of IE are so old that it is irrelevant to developers whether it was implemented in 2001, or all the way back in 1996.  (Furthermore, the mdn-bcd-collector project is not compatible with IE 5 or lower, thus making testing inefficient -- much like the case of Safari 1-2 when we added Safari 4 as a range.)


foolip (approved):
I think we should do this, but over to @ddbeck

ddbeck (commented):
Yeah, seems reasonable. IE6 was _the_ browser for a long time. Seems like a nice enough boundary.

This needs a rebase though—happy to review and merge when that's done.

vinyldarkscratch:
What rebase?  >:3

(Thinking about it by the way, should we perhaps sort that dictionary alphabetically?)

ddbeck:
> (Thinking about it by the way, should we perhaps sort that dictionary alphabetically?)

Yeah, that'd be good. Want to include that now?

vinyldarkscratch:
Sure thing -- done!

ddbeck (approved):
Thank you! 🎉

PR #7343, opened 2020-11-12
Update Firefox Android support for tracklisting in grid-auto-columns and grid-auto-rows

rachelandrew:
While working on https://github.com/mdn/sprints/issues/3835 spotted that the support for a tracklisting in `grid-auto-columns` and `grid-auto-rows` for Firefox Android hadn't been updated.

Have tested this using Browserstack also and the value works as expected now.


ddbeck (approved):
Thank you! 🎉

PR #7344, opened 2020-11-12
Update MediaQueryList support for Safari. (inheriting EventTarget)

paulirish:
Well, it shipped in Safari 14.

I suppose #6745 barely missed making this change, since it updated the `onchange` support.

https://bugs.webkit.org/show_bug.cgi?id=203288#c21 commit landing in april 2020. 

saf 14 shipped in june 2020.

I tested saf 14 and indeed it has it all there:

image

context: i started looking at this because of https://github.com/paulirish/matchMedia.js/issues/86

### DIFF
--- a/api/MediaQueryList.json
+++ b/api/MediaQueryList.json
@@ -126,10 +126,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": false
+              "version_added": "14"
             },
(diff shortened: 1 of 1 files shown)


sideshowbarker (approved):

sideshowbarker:
Nice catch

PR #7345, opened 2020-11-12
Add Safari iOS 3 as ranged version

vinyldarkscratch:
This PR is a sibling to #6915.  Due to the challenges of testing Safari Desktop 1 through 2, and the lack of Safari Desktop 3 in BrowserStack (plus iOS 1-2), there are quite a few Safari iOS versions that we will not be able to easily mirror data, or otherwise obtain real data, for.  Since iOS 3 is the earliest iOS version available on BrowserStack, I figured it makes sense for us to add it as a ranged version.


sideshowbarker (approved):

sideshowbarker:
(Reviewed but not merging since per https://github.com/mdn/browser-compat-data/blob/master/GOVERNANCE.md#peers because this is a change to project meta-docs and the linter, it requires Owner review.)

ddbeck (approved):
Yeah, offering version ranges for both Safaris makes sense to me. Thank you!

Issue #7350, opened 2020-11-12
Remove UIEvent.cancelBubble


saschanaz:
[It's redundant](https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/cancelBubble) as there is [Event.cancelBubble](https://developer.mozilla.org/en-US/docs/Web/API/Event/cancelBubble).

PR #7351, opened 2020-11-12
Add linter to make sure statuses have consistent inheritance

### LINKED ISSUE #7348: Add a linter to make sure deprecated flag is also applied to children
See https://github.com/mdn/browser-compat-data/pull/7311/files#r522266719

saschanaz:
- [x] Summarize your changes
- [x] Link to related issues or pull requests, if any: closes #7348, https://github.com/mdn/browser-compat-data/pull/15932


html/elements/script.json
```diff
@@ -551,7 +551,7 @@
             "status": {
               "experimental": false,
               "standard_track": true,
-              "deprecated": true
+              "deprecated": false
```
saschanaz, line 554: This is done in #7264. I think we should instead create `type.javascript` and mark it deprecated, because `type="module"` is a totally modern and valid thing.

api/WorkerGlobalScope.json
```diff
@@ -91,7 +91,7 @@
           "status": {
             "experimental": false,
             "standard_track": false,
-            "deprecated": true
+            "deprecated": false
```
saschanaz, line 94: We should probably remove this item?

api/UIEvent.json
```diff
@@ -140,7 +140,7 @@
           "status": {
             "experimental": false,
             "standard_track": false,
-            "deprecated": true
+            "deprecated": false
```
saschanaz, line 143: Filed #7350

foolip (commented):
This looks pretty straightforward, happy to review this to completion. In this cases the changes to the data are few enough that I think it's fine to do them in the same PR as the lint.

Do you think we should also lint for experimental status, that a non-experimental feature must have a non-experimental parent?

foolip:
Right, also it shouldn't be possible to have a standard feature as a child of a non-standard feature. So all 3 statuses should be possible to have similar lints for.

saschanaz:
Got a lot more after including `experimental` status check, so splitting this too.

test/linter/test-status-inheritance.js
```diff
@@ -0,0 +1,60 @@
...
+ * @typedef {import('../../types.js').StatusBlock} StatusBlock
+ */
+
+/**
+ * @param {Identifier} data
+ * @param {Logger} logger
+ * @param {StatusBlock?} parentStatus,
+ * @param {string[]} path
+ */
+function checkStatusInheritance(data, logger, parentStatus = {}, path = []) {
```
foolip, line 15: @saschanaz do you want to add a "Unexpected standards track status while the parent is non-standards track"?

queengooborg:
Looks like this needs to be formatted, could you run a formatter real quick to see what needs to be done to make this linter pass?

saschanaz:
(https://github.com/mdn/browser-compat-data/pull/15932 exists for this)

saschanaz:
I don't think I'll merge repeatedly only to be replaced by others, closing.

PR #7352, opened 2020-11-12
Add missing URLs.

jpmedley:


vinyldarkscratch:
I thought there was discussion that we shouldn't include MDN URLs that we're expecting articles for (AKA, no 404 URLs), but I can't seem to find the discussion for it...

sideshowbarker:
> I thought there was discussion that we shouldn't include MDN URLs that we're expecting articles for (AKA, no 404 URLs), but I can't seem to find the discussion for it...

https://github.com/mdn/browser-compat-data/pull/5297 is a related PR — which, if/when we ever agree to land it, will bulk-remove all MDN URLs in BCD that are 404s.

jpmedley:
You do realize that this doubles the number of PRs that need to be submitted or all new entries. I always do the BCD first because otherwise the call to action by the missing BCD on the result page could lead multiple contributors submitting PRs for the same material. If I can't submit URLs with the version numbers, then I need to come back after the pages are created and submit another PR.

vinyldarkscratch:
I think that the effort of creating follow-up PRs to add MDN URLs later, when the docs are created, is minimal enough to not be a concern.

The PR that @sideshowbarker linked contains the conversation I was referring to earlier, about disallowing 404 URLs for `mdn_url`.  Both Florian and Chris agreed that we should remove the existing 404 URLs (and technically, that extends to not adding new ones) for both SEO purposes and to give ourselves a way to determine what articles still need to be written.  If we include broken links, data consumers can't assume that the existence of the `mdn_url` property means that the documentation exists, and will have to implement 404 checking in their code.

ddbeck:
> If I can't submit URLs with the version numbers, then I need to come back after the pages are created and submit another PR.

I'm optimistic we can do some automation around this, since adding pages will soon go through GitHub as well. I imagine we could have Yari trigger a workflow that updates BCD for new pages that reference BCD identifiers that lack and MDN URL.

In the meantime, we haven't laid out a data guideline forbidding 404 URLs—which now that I think of it, would be a really nice on-ramp to #5297—so I'll merge this.

ddbeck (approved):

jpmedley:
Thank you.

PR #7353, opened 2020-11-13
Add note about partial srcset/sizes support in Safari before version 9

sideshowbarker:
This change adds notes fields to explain why `html.elements.source.srcset` `html.elements.source.sizes` are marked as `partial_implementation` for Safari 7. Relates to https://github.com/mdn/browser-compat-data/issues/4162.

### DIFF
--- a/html/elements/source.json
+++ b/html/elements/source.json
@@ -154,7 +154,8 @@
                 },
                 {
                   "version_added": "7",
-                  "partial_implementation": true
+                  "partial_implementation": true,
+                  "notes": "Supports a subset of the syntax for resolution switching (using the <code>x</code> descriptor), but not the full syntax that can be used with <code>sizes</code> (using the <code>w</code> descriptor)."
                 }
               ],
               "safari_ios": [
@@ -163,7 +164,8 @@
                 },
                 {
                   "version_added": "8",
-                  "partial_implementation": true
+                  "partial_implementation": true,
+                  "notes": "Supports a subset of the syntax for resolution switching (using the <code>x</code> descriptor), but not the full syntax that can be used with <code>sizes</code> (using the <code>w</code> descriptor)."
                 }
               ],
               "samsunginternet_android": {
@@ -284,7 +286,8 @@
                 },
                 {
                   "version_added": "7",
-                  "partial_implementation": true
+                  "partial_implementation": true,
+                  "notes": "Supports a subset of the syntax for resolution switching (using the <code>x</code> descriptor), but not the full syntax that can be used with <code>sizes</code> (using the <code>w</code> descriptor)."
                 }
               ],
               "safari_ios": [
@@ -293,7 +296,8 @@
                 },
                 {
                   "version_added": "8",
-                  "partial_implementation": true
+                  "partial_implementation": true,
+                  "notes": "Supports a subset of the syntax for resolution switching (using the <code>x</code> descriptor), but not the full syntax that can be used with <code>sizes</code> (using the <code>w</code> descriptor)."
                 }
               ],
               "samsunginternet_android": {


ddbeck (changes requested):
Thank you, @sideshowbarker. I fell down an interesting rabbit hole thanks to this. 😄

Since caniuse is one of BCD's consumers and this text would ultimately appear on caniuse, I thought it would be weirdly recursive to cite caniuse. So I tried to independently verify this. I succeeded and also turned up a bunch of information on how this feature shipped in Safari and Chrome.

I think dropping the citation is the only thing that needs to happen here, though if you want to act on anything I've shared below, that would be welcome, but not required (I'll open a follow up if needed).

--- 

* This Chrome [intent to ship](https://groups.google.com/a/chromium.org/g/blink-dev/c/SVQxmsJRa9c/m/Qk7xgUOXI8kJ) indicates that Chrome's original implementation was [a port from WebKit](https://bugs.webkit.org/show_bug.cgi?id=110252). It seems like the implementation was intentionally incomplete (or perhaps the spec wasn't complete at that time—either way, the described limitation definitely existed).
* The Chrome [status entry](https://chromestatus.com/features/4644337115725824) showed when that originally shipped (Chrome 34).
* Chrome appears to have got the complete syntax in Chrome 38 ([intent to ship](https://groups.google.com/a/chromium.org/g/blink-dev/c/vuLeAM9fJww/m/UHnMz7QLRWgJ), [bug](https://bugs.chromium.org/p/chromium/issues/detail?id=357586), [status entry](https://chromestatus.com/features/5910974510923776)).
* WebKit got the complete implementation circa June 2014 ([bug](https://bugs.webkit.org/show_bug.cgi?id=133620), [changeset](https://trac.webkit.org/changeset/170576/webkit)); the versions for Safari for iOS might be suspect (it's possible Safari for iOS never had the partial implementation) as might be the desktop Safari version for the full implementation (it looks like the currently listed version is much later than the changeset implies).

sideshowbarker:
@ddbeck Nice research on this — it’s great to have the record here

Focusing for now just on the Safari side of this —

> * WebKit got the complete implementation circa June 2014 ([bug](https://bugs.webkit.org/show_bug.cgi?id=133620), [changeset](https://trac.webkit.org/changeset/170576/webkit)); the versions for Safari for iOS might be suspect (it's possible Safari for iOS never had the partial implementation) as might be the desktop Safari version for the full implementation (it looks like the currently listed version is much later than the changeset implies).

https://trac.webkit.org/browser/webkit/tags/Safari-538.35/Source/WebCore/html/HTMLImageElement.idl would seem to indicate that https://trac.webkit.org/changeset/170576/webkit didn’t ship in Safari 8 (2014-10-16) but per https://trac.webkit.org/browser/webkit/tags/Safari-601.1.56/Source/WebCore/html/HTMLImageElement.idl did ship in Safari 9.

So I think that means what we already have for the first Safari and iOS versions with full support is correct, right?

> I think dropping the citation is the only thing that needs to happen here

OK, I’ve pushed that change

> though if you want to act on anything I've shared below, that would be welcome, but not required (I'll open a follow up if needed)

I’m happy to follow up on it myself too. Do you mean we should update the Chrome version info? Or we still need to make refinements to the Safari data?

ddbeck:
> So I think that means what we already have for the first Safari and iOS versions with full support is correct, right?

Ah, yes. I didn't actually follow the changes to see when they actually shipped. Thank you for taking that the rest of the way.

> Do you mean we should update the Chrome version info? Or we still need to make refinements to the Safari data?

Yeah, I was thinking that if we updated the Chrome version info, we could reuse the Safari note text for the support statement covering Chrome 34-37 (and without the note, from 38). And we'd have some really complete data here, too.

ddbeck (approved):
OK, I'm going to merge this and put the Chrome-related data into a separate PR. Thank you! 🎉

ddbeck:
Opened #8889 based on this discussion.

PR #7356, opened 2020-11-13
Update Firefox for android video autoplay support

rsepierre:
**Summary :** Adding note stating the current firefox for android state regarding video autoplay. Currently, the feature is most likely not working for most if not all users.

**Justification :**
It is safe to say that a developper using "caniuse.com" expects to know if the feature will "most likely work for a percentage of users" and on which platform (thus trying to answer to the question "can I use it ?").
In the case of Firefox for android the answer is "it will most likely not work" as
- the setting is disabled by default. 
- At no time during installation are you prompted to opt-in or out of the feature. 
- Web developpers can not ask for user consent to use the feature
- The setting is hidden
Adding this note is the bare minimum and I would even consider changing the status to "partial support" to make it evident in the graphs and stats that you must expect that the feature is not working instead of the opposite.

**Sources:**
The last patch note signaling the an update to the "autoplay" feature dates back to firefox for android version 66.
https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/66
This update states that any media element **with audio** will be blocked. But this is not the current state of the feature as all media will be blocked.
Every piece of mozilla's documentation states that "muted media element" should autoplay just fine. Which might signify that it is a oversight (or a bug ?). But it is the current state of the browser and is presented as a feature.

**Testing:**
Version tested on : 82.1.3 (Build #2015774643)
Using the Recommended (default) settings : "Block audio and video" loading a page with a video element with autoplay but also without audio track, and muted, and volume at 0 : the video is frozen.
Using the setting "Block only audio" and restarting the browser : the video plays.
Video plays by default on other browsers

**Related:**
Google article on consequences of disabling autoplay and why is it important for developpers to have accurate data on if they can use the feature or not : https://developers.google.com/web/updates/2016/07/autoplay
Dev.com's article on important benefits of the feature over older (but supported) technologies : https://web.dev/replace-gifs-with-videos/

### DIFF
--- a/html/elements/video.json
+++ b/html/elements/video.json
@@ -125,7 +125,8 @@
                 "version_added": "3.5"
               },
               "firefox_android": {
-                "version_added": "4"
+                "version_added": "4",
+                "notes": "Will most likely not work. Disabled by default for all users whether the video is silent or not. Users to have manually opt-in for all websites. Developpers can not ask users to opt-in for the feature on their website."
               },
               "ie": {
                 "version_added": "9"


ddbeck (changes requested):
Hi @rsepierre. Thank you for opening this PR and welcome to BCD!

You raise a number of interesting issues here and I think you're right to want a note on this feature. It's tougher to get this right than I would've guessed. Here's my overall impression:

The [behavior of `autoplay` in the specification](https://html.spec.whatwg.org/multipage/media.html#attr-media-autoplay) ([see also](https://html.spec.whatwg.org/multipage/media.html#eligible-for-autoplay)) is rather permissive. It's a "hint" to user agents, but user agents don't appear to be required to do much. For that reason, I don't think `"partial_implementation"` is a good fit here.

That said, the behavior of Firefox for Android is rather different from desktop Firefox and the other major browsers. For that reason, you're right to suggest a note, though I think we actually have two distinct notes to make:

- Firefox for Android's default behavior is to block all autoplaying video and audio (compared to desktop's default to audible media only).
- Firefox for Android never prompts users to grant autoplay permission to a site.

Moreover, both of these behaviors appear to be new from when Fenix shipped as Firefox for Android 79.

There's also an option to capture the historic behavior (from 66 until 79), of blocking autoplay for audible media only.

So we'll probably need an array of notes, something like this, to move forward:

```js
[
  "From version 79, Firefox for Android blocks autoplay for all audio and video by default.",
  "From version 79, Firefox for Android never prompts users to grant permission to autoplay media.",
  "From version 66 to 68, Firefox for Android blocks autoplay for audible media."
]
```

What do you think? Do you think this sort of detail speaks to the issues you've raised?

rsepierre:
I think it is important insight for developpers !

SmartSelect_20201119-182722_Firefox

For now, there is only one note about iOS Safari blocking autoplay on audible elements. 

Actually chrome does the same and as you stated maybe firefox for Android from 66 to 68. I'm not sure how many other mobile browsers do the same.

But this follows most modern recommandations and documentation on autoplaying content and makes particularly sense on mobile browsers.

So ideally, we could use the one same or less ideally mutiple identical notes accross the different browsers and versions implementing the "block audible autoplay" strategy.

Now where firefox for android (fenix) is truly unique is when blocking all media (muted of not). This without asking for preference during the installation, or without being able to prompt the user for permission on the spot.

It is the combination of those three parameters that makes that as a developper, you should expect every fenix users's autoplay feature to not work.

Chrome doesn't prompt you for permission if you willingly block autoplay in their parameters and they dont get a note for that.

It could be one, two or three notes, be because it's only when those conditions are combined that it is a worst case / users are not even aware problem. Hence why I was trying to make it into one note.

Now I posted on fenix's github shortly after I posted here.
This might be useful insight on how we could choose to implement notes :

- They are working on a "per-website" autoplay permissions feature and (i'm not sure but) you might be able to prompt users for permission.
- And mainly, they should (hopefully) be going back to blocking audio only by default.

https://github.com/mozilla-mobile/fenix/issues/16557#issuecomment-729820241

(Also : if fenix get's an update following my github pull request, I'm not joke going to tell my future children about how I saved the internet)

ddbeck:
Thanks for the update, @rsepierre (and sorry I took so long to come back to this). Having re-read this thread, I think this is the key point, where you said:

> So ideally, we could use the one same or less ideally mutiple identical notes accross the different browsers and versions implementing the "block audible autoplay" strategy.

Can you propose some note text? I understand that you think this is important, but apparently my suggested text wasn't really working for you. Since you're more directly affected by this, could you give a me a sense of the text you would've had liked to see, when you first looked up this feature?

ddbeck:
Due to inactivity and neither of my proposals having been taken up, I'm closing this PR. I'm not opposed to more detailed notes for this feature to be added in the future though, so please don't consider this a rejection of the idea in general. Thank you!

PR #7360, opened 2020-11-14
Remove UIEvent.cancelBubble

### LINKED ISSUE #7350: Remove UIEvent.cancelBubble

saschanaz:
- [x] Summarize your changes
   * `data:text/html,<script>alert("cancelBubble" in UIEvent.prototype)</script>` via mozregression
- [x] Link to related issues or pull requests, if any: closes #7350


api/Event.json
```diff
@@ -226,14 +226,24 @@
...
-              "version_added": "53",
-              "notes": "Prior to Firefox 53, this property was defined on the <a href='https://developer.mozilla.org/docs/Web/API/UIEvent'><code>UIEvent</code></a> interface. See <a href='https://bugzil.la/1298970'>bug 1298970</a> for more details."
-            },
+            "firefox": [
+              {
+                "version_added": "53"
+              },
+              {
+                "version_added": "1",
+                "notes": "This property was defined on the <a href='https://developer.mozilla.org/docs/Web/API/UIEvent'><code>UIEvent</code></a> interface."
```
saschanaz, line 235: Could we add a dedicated field to reduce this kind of prose?

ddbeck: Please open an issue for this. I'm not saying no, but we need to talk to consumers about whether they would actually use such a thing (or whether we need to consider something else, like generated notes).

saschanaz: Thanks, #7400

foolip: That was duped into https://github.com/mdn/browser-compat-data/issues/3463, good old prototype chain moves. Do we need to resolve that before finishing this PR though?

ddbeck: Thanks for the nudge on this. I don't think we need to resolve that first. This makes some sense as-is and I'm going to merge it.

ddbeck (approved):
Thank you! 🎉

foolip (commented):
Oops, I had a pending comment/suggestion.

ddbeck:
@foolip Sorry for moving ahead prematurely on that. I opened #8040.

foolip:
@ddbeck I hasn't submitted the review, so it was my mistake. Thanks for fixing nevertheless!

PR #7363, opened 2020-11-14
Remove WorkerGlobalScope.close()

### LINKED ISSUE #7362: Remove WorkerGlobalScope.close()
It has been moved to DedicatedWorkerGlobalScope and SharedWorkerGlobalScope as the current BCD item says. to remove it from ServiceWorkerGlobalScope.

An open question is that whether we need to mention that it once also existed on ServiceWorkerGlobalScope. I guess probably not?

saschanaz:
- [x] Summarize your changes
- [x] Link to related issues or pull requests, if any: closes #7362 

This needs double check for the version numbers. I thought I could at least peek at mdn-bcd-results but sadly I couldn't as the results from 0.5.9 didn't include data for workers.


ddbeck:
> This needs double check for the version numbers.

Hi @saschanaz (and happy new year). Is this something you were planning to do or is this something you were looking for help on?

saschanaz:
I can confirm IE/Firefox/Chrome/Safari desktop versions via mdn-bcd-results. Can't confirm mobile versions because of no sufficient data.

saschanaz:
>you were looking for help on?

Yes, please help 🙏

api/DedicatedWorkerGlobalScope.json
```diff
@@ -52,31 +52,31 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/DedicatedWorkerGlobalScope/close",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "4"
             },
             "chrome_android": {
               "version_added": true
```
ddbeck, line 58: OK, I did some testing and research and I can't find any evidence this was introduced later on Android

```suggestion
              "version_added": "18"
```

api/DedicatedWorkerGlobalScope.json
```diff
@@ -85,7 +85,7 @@
...
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "37"
```
ddbeck, line 88: I couldn't find out if this existed earlier or not

```suggestion
              "version_added": "≤37"
```

api/DedicatedWorkerGlobalScope.json
```diff
@@ -52,31 +52,31 @@
...
-              "version_added": null
+              "version_added": "10"
             },
             "opera": {
-              "version_added": true
+              "version_added": "11.5"
             },
             "opera_android": {
-              "version_added": true
+              "version_added": null
```
ddbeck, line 76: Given, Opera's history, I think it's safe to mirror from desktop:

```suggestion
              "version_added": "11.5"
```

api/DedicatedWorkerGlobalScope.json
```diff
@@ -52,31 +52,31 @@
...
             },
             "safari": {
-              "version_added": null
+              "version_added": "4"
             },
             "safari_ios": {
               "version_added": null
```
ddbeck, line 82: Another one that looks safe to mirror:

```suggestion
              "version_added": "3.2"
```

api/DedicatedWorkerGlobalScope.json
```diff
@@ -85,7 +85,7 @@
               "version_added": true
```
ddbeck, line 85: Another mirrorable entry

```suggestion
              "version_added": "1.0"
```

api/SharedWorkerGlobalScope.json
```diff
@@ -100,40 +100,40 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/SharedWorkerGlobalScope/close",
           "support": {
             "chrome": {
-              "version_added": false
+              "version_added": "4"
             },
             "chrome_android": {
-              "version_added": false
+              "version_added": true
```
ddbeck, line 106:
```suggestion
              "version_added": "18"
```

api/SharedWorkerGlobalScope.json
```diff
@@ -100,40 +100,40 @@
...
-              "version_added": null
+              "version_added": "5"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "7"
             },
             "samsunginternet_android": {
-              "version_added": false
+              "version_added": true
```
ddbeck, line 133:
```suggestion
              "version_added": "1.0"
```

api/SharedWorkerGlobalScope.json
```diff
@@ -100,40 +100,40 @@
...
             },
             "webview_android": {
-              "version_added": false
+              "version_added": "37"
```
ddbeck, line 136:
```suggestion
              "version_added": "≤37"
```

ddbeck (changes requested):
OK, I did some digging and I think it's mostly safe to mirror here.

One questionable thing I did however, was to suggest changing webview 37s to ≤37. If you actually managed to test that directly, then feel free to ignore, but my guess is that it's an upper bound.

api/DedicatedWorkerGlobalScope.json
```diff
@@ -52,40 +52,40 @@
...
-              "version_added": true
+              "version_added": "11.5"
             },
             "safari": {
-              "version_added": null
+              "version_added": "4"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "3.2"
```
saschanaz, line 82: This fails because the base interface is marked as 5.1, needs to check that one 🤔

ddbeck: Hmm. I don't trust that version much, though I'm less confident about mirroring it. It looks like the base interface data was added to BCD during the migration from the wiki. It doesn't look like there was any attempt to verify it at the time. As it stands, it suggests that this API existed on desktop for 2.5 years before making it to iOS. That seems pretty unlikely to me. I guess we could resolve this by setting the parent interface to `"3.2"` for iOS as well.

saschanaz: Worker support needs some process/thread work so I wouldn't be surprised if that happened.

[This blog post](https://www.davidbcalhoun.com/2011/new-mobile-safari-stuff-in-ios5-positionfixed-overflowscroll-new-input-type-support-web-workers-ecmascript-5/) says Apple indeed tried enabling it on iOS 3 but then deferred it to 5. Several other posts also report same, so I guess 5.1 sounds right 👀

ddbeck: Oh, nice find! Should we go the other direction to resolve the conflict then?

```suggestion
              "version_added": "5.1"
```

ddbeck (approved):
Thank you! 🎉

PR #7364, opened 2020-11-15
Update scroll-behavior.json

break-stuff:
After updating my Mac OS and testing this feature in Safari 14.0.1, this feature still does not work. I am setting the `version_added` for Safari and iOS to `false`. This is validated [here](https://caniuse.com/css-scroll-behavior).

- [x] Summarize your changes

### DIFF
--- a/css/properties/scroll-behavior.json
+++ b/css/properties/scroll-behavior.json
@@ -30,10 +30,22 @@
               "version_added": "45"
             },
             "safari": {
-              "version_added": "14"
+              "version_added": "14",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "CSSOM View Smooth Scrolling"
+                }
+              ]
             },
             "safari_ios": {
-              "version_added": "14"
+              "version_added": "14",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "CSSOM View Smooth Scrolling"
+                }
+              ]
             },
             "samsunginternet_android": {
               "version_added": "8.0"


saschanaz:
[Per this post](https://blog.amp.dev/2020/07/06/contributions-to-web-platform-interoperability-first-half-of-2020/) it's behind a flag:

>In WebKit, we began with scroll-behavior, which provides the ability to do smooth scrolling. Based on our previous patch, it has landed and is guarded by an experimental flag “CSSOM View Smooth Scrolling” which is disabled by default.

But I'm not familiar about how to document this here, both Firefox and Chrome has machine-readable flag names but not sure about Safari.

sideshowbarker:
> [Per this post](https://blog.amp.dev/2020/07/06/contributions-to-web-platform-interoperability-first-half-of-2020/) it's behind a flag:
> 
> > In WebKit, we began with scroll-behavior, which provides the ability to do smooth scrolling. Based on our previous patch, it has landed and is guarded by an experimental flag “CSSOM View Smooth Scrolling” which is disabled by default.

Yeah — and the only way to enable it is from **Develop** > **Experimental Features** menu in the Safari UI:

Screen Shot 2020-11-16 at 16 19 27


> But I'm not familiar about how to document this here, both Firefox and Chrome has machine-readable flag names but not sure about Safari.
> 
> Edit: This is the commit https://webkit-search.igalia.com/webkit/commit/f51558c0534a07c1a160e540848e7847db9279c4

https://trac.webkit.org/browser/webkit/trunk/Source/WTF/Scripts/Preferences/WebPreferencesExperimental.yaml#L130 shows the relevant feature’s internal machine-readable flag — which is `CSSOMViewSmoothScrollingEnabled`.

But as far as I know, there’s no way for users to actually set it using that `CSSOMViewSmoothScrollingEnabled` name — instead, the only way that users can toggle it is from the **Develop** > **Experimental Features** menu in the Safari UI.

css/properties/scroll-behavior.json
```diff
@@ -30,10 +30,10 @@
               "version_added": "45"
             },
             "safari": {
-              "version_added": "14"
+              "version_added": false
```
sideshowbarker, line 33:
```suggestion
              "version_added": "14"
              "flags": [
                {
                  "type": "preference",
                  "name": "CSSOM View Smooth Scrolling"
                }
              ],
```

css/properties/scroll-behavior.json
```diff
@@ -30,10 +30,10 @@
...
             },
             "safari_ios": {
-              "version_added": "14"
+              "version_added": false
```
sideshowbarker, line 36:
```suggestion
              "version_added": "14"
              "flags": [
                {
                  "type": "preference",
                  "name": "CSSOM View Smooth Scrolling"
                }
              ],
```

sideshowbarker (changes requested):
We should keep `"14"` as the version number, but add a `flags` array, with `CSSOM View Smooth Scrolling` as the relevant flag name (and which, like all Safari experimental flags, is a boolean — so we don’t need to also specify a `value_to_set` for it).

sideshowbarker (approved):

sideshowbarker:

break-stuff:
I'm glad I could contribute a little. 😄

rik:
https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior still shows a green full support box. Has the data not been synced with the website yet or is there an issue with how those flags are displayed?

sideshowbarker:
> https://developer.mozilla.org/en-US/docs/Web/CSS/scroll-behavior still shows a green full support box. Has the data not been synced with the website yet or is there an issue with how those flags are displayed?

Just not been synced yet. I think typically it gets synced once a week, at the end of the week.

So at the beginning of next week, it should be showing the correct data.

ddbeck:
And for even more detail on the publishing process: https://github.com/mdn/browser-compat-data/blob/master/docs/contributing.md#updating-compatibility-tables-on-mdn. I'll put this on my to do list to refresh after today's release though. 😄

PR #7370, opened 2020-11-16
Update Safari 14 release notes URL

gsouquet:
The PR changes the `release_notes` for Safari 14. The old value was referring to the beta release notes.

I would be inclined to change the `release_date` to be aligned with the macOS Big Sur release but I'm not 100% sure this is the official value. Anyone able to assist here?

### DIFF
--- a/browsers/safari.json
+++ b/browsers/safari.json
@@ -169,7 +169,7 @@
         },
         "14": {
           "release_date": "2020-09-16",
-          "release_notes": "https://developer.apple.com/documentation/safari-release-notes/safari-14-beta-release-notes",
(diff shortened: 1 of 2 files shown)


sideshowbarker (approved):

sideshowbarker:
> I would be inclined to change the `release_date` to be aligned with the macOS Big Sur release but I'm not 100% sure this is the official value. Anyone able to assist here?

I don’t personally have any insight to add on what date we should choose there. But to ge a decision about whether to change it or not, I guess raising a separate PR with a change seems like the he right way to go.

PR #7372, opened 2020-11-16
Document how to note operating system limitations with "partial_implementation"

### LINKED ISSUE #6906: Guideline proposal: operating system limitations imply partial_implementation
It's not always clear when to use `partial_implementation`. One case that feels pretty clear-cut however, is when a feature that otherwise has no reason to be limited by operating system only works on a specific OS. I'm proposing a wording along these lines:

> If a browser or engine is available on more than one operating system and a feature only works on a subset of those operating systems, then the support statement should set `"partial_implementation": true`. For example, if a browser supports both Windows and Linux, but only implements a feature on Windows, then a support statement for that feature should should set `"partial_implementation": true`.
>
> However, this guideline does not apply to features where the browser's expected behavior is conditional on the behavior of the operating system. For example, a browser can fully implement a CSS media query even if an underlying operating system can never satisfy the media query's condition because it does not support the requisite hardware.

This seems kinda clear to me, but I'd love some early input before moving toward a PR.

@foolip and @vinyldarkscratch this was inspired by a few comments in #6840. I'd appreciate your take on this.

ddbeck:
I've received some good feedback on #6906. This PR fixes #6906, with a very slight revision of the wording from the original issue.

### DIFF
--- a/docs/data-guidelines.md
+++ b/docs/data-guidelines.md
@@ -12,6 +12,7 @@ This file contains recommendations to help you record data in a consistent and u
   - [Permissions API permissions (`permissionname_permission`)](#permissions-api-permissions-permissionname_permission)
   - [`"partial_implementation"` requires a note](#partial_implementation-requires-a-note)
   - [Non-functional defined names imply `partial_implementation`](#non-functional-defined-names-imply-partial_implementation)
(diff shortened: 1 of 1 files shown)


sideshowbarker (approved):

ddbeck:
Since there's been no additional feedback on this one, I'm merging it. 🚀  (And thanks for the review, @sideshowbarker.)

PR #7375, opened 2020-11-17
is and where chrome

rachelandrew:
While looking up support info for `:not()` saw that Chrome are shipping support for `is()` and `where()`.

https://groups.google.com/a/chromium.org/g/blink-dev/c/Wodp9dSSdL8/m/OZo2vJqTAwAJ

I also tested the examples on our pages. The intent to ship above also confirms that Safari does not implement the forgiving selector list support (though it is fixed in webkit).

### DIFF
--- a/css/selectors/is.json
+++ b/css/selectors/is.json
@@ -8,6 +8,10 @@
           "support": {
             "chrome": [
               {
+                "version_added": "88"
+              },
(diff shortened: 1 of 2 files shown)


ddbeck (approved):
Thank you! 🎉

PR #7376, opened 2020-11-17
complex selector support

rachelandrew:
I am working on documenting https://github.com/mdn/sprints/issues/3905

This PR adds details of complex selector support as a value for `:not()`.

- in FIrefox 84: https://bugzilla.mozilla.org/show_bug.cgi?id=933562
- Chrome implementing [bug](https://bugs.chromium.org/p/chromium/issues/detail?id=1139865), [intent to ship](https://groups.google.com/a/chromium.org/g/blink-dev/c/0alTkXvDCL8/m/qWIdYMpNBQAJ)
- Safari has had implementation since 9, which I tested.


css/selectors/not.json
```diff
@@ -49,6 +49,55 @@
             "deprecated": false
           }
         },
+        "complex_selectors": {
```
ddbeck, line 52: How is this different from the `selector_list` feature below? I'm not sure I understand the nuance between selector lists, compound selectors, and complex selectors. Can we do anything to make the distinction clearer (because it's not clear to me)?

ddbeck: @rachelandrew Do you plan to return to this PR at some point?

rachelandrew: @ddbeck sorry I'd missed this request, I'll have a look tomorrow.

rachelandrew: @ddbeck there are examples on the documentation page at: https://developer.mozilla.org/en-US/docs/Web/CSS/:not the selector list being the penultimate, and a complex selector the last item.

I'm unclear as to what more you expect to be part of the BCD though.

ddbeck: OK, I think I get the selector list distinction, but not the compound and complex distinction.

[The spec](https://www.w3.org/TR/selectors-4/#complex) defines a complex selector as "one or more compound selectors separated by combinators". By that definition, the last item in the linked examples demonstrates a compound (not complex) selector inside the `:not`. Have I badly misunderstood this PR to mean that the new feature describes whether `:not` itself—not its contents—can be part of a complex selector?

If this is, in fact, about the contents of `:not` then I'd like to know what not supporting this feature means. For example, if support is `false`, does that mean _neither_ compound nor complex selectors work (e.g. `:not(span)` works, but neither `:not(span.a)` nor `:not(span .b)` work)? Or Is it possible for a browser to support compound but not complex selectors (e.g. `:not(span)` and `:not(span.a)` work, but `:not(span .b)` doesn't)?

ddbeck: (sorry, forgot to tag you on this @rachelandrew)

rachelandrew: @ddbeck I keep coming back to try and reply to this, but I'm struggling to understand the problem you are having.

Do you think the MDN page needs updating? As far as support goes I think complex selectors are all pretty much lumped together. 

I don't really understand what you are driving at with,

> Have I badly misunderstood this PR to mean that the new feature describes whether :not itself—not its contents—can be part of a complex selector?

We're talking about what matches for `:not`. Essentially what was added in Selectors 4: https://drafts.csswg.org/selectors-4/#negation (see the note)

ddbeck: I think I misunderstood one part from the outset, then figured it out, and muddied things a little. Let me try to reboot things here a bit.

Here's the problem I have right now: I think the description of this feature creates ambiguities. I don't know your intent, so I can't tell you how to fix those ambiguities. Once I know more about what you mean by this data, then I can suggest some paths forward.

Here are the ambiguities I'm trying to fix:

1. The other subfeature of `:not()` is described as "Selector list argument". This feature doesn't mention the word argument. Is this feature correctly understood to also be about the argument to `:not()`? That is, is it correct to read this subfeature as "Complex and compound selector argument"?

   (I'm pretty sure it is, but I want to double check to be sure.)
   
1. This subfeature refers to selectors plural, while the other subfeature refers to selector list support (that is, whether `:not` accepts more than one selector as an argument). Is it correct to understand this subfeature to mean "`:not` accepts one complex or compound selector as an argument"? Or am I supposed to understand this as subsuming the selector list question too?

   (Again, I'm pretty sure it means that it accepts one complex/compound selector, not a selector list, but the plural makes me want to double check.)
   
2. "Complex and compound selectors" sounds like two things (e.g., "Complex selector argument" and "Compound selector argument"). As I see it, a `false` value for this feature, as it stands, could be plausibly read by web developers as:

   - "Supports neither a complex nor a compound selector as an argument to `:not()`"
   - "Supports a compound selector but not a complex selector as an argument to `:not()`"
   - "Supports a compound or complex selector as an argument to `:not()`, but I haven't told you which one"
   - "Supports a complex but not a compound selector as an argument to `:not()`"

   (I happen to know the definition of complex better now than I did at the start, so I know the last two are definitionally impossible. But I thought this might be the case when we started! And I'm still not sure about the first two bullets.)

I hope this helps explain what I'm getting at here. 😅

rachelandrew:
My intent is to document what was added to Firefox in https://bugzilla.mozilla.org/show_bug.cgi?id=933562 that's my intent, I'm completely confused by the problem you seem to be having with it, so I am obviously missing something.

rachelandrew:
also if this needs a rebase I'll close this PR and redo it, I have a 100% track record of rebases just ending up in a mess.

ddbeck:
@rachelandrew Are none of my questions answerable? I'll pull them out here, as best I can:

1. Does a `true` value in this feature mean "`:not()` accepts one (and only one) complex or one (and only one) compound selector as an argument"?
2. Does a `false` value in this feature mean "Supports neither a complex nor a compound selector as an argument to `:not()`", "Supports a compound selector but not a complex selector as an argument to `:not()`"? If yes, which one is correct? If neither is correct, why?

ddbeck:
Also, there's no rebase needed here (I'm not sure why that label was added).

ddbeck:
Hi @rachelandrew. I spent a lot of today time reviewing this carefully and I've pushed some changes to resolve this. Here's what I've done:

- I've combined the existing selector list feature with these compound-complex details. My rationale:
  
  The spec for `:not()` includes this:

  > The negation pseudo-class, `:not()`, is a functional pseudo-class taking a <a href="https://drafts.csswg.org/selectors-4/#selector-list">selector list</a> as an argument.

  The linked _selector list_ is defined like this:

  > A list of simple/compound/complex selectors is a comma-separated list of simple, compound, or complex selectors. This is also called just a selector list when the type is either unimportant or specified in the surrounding prose; if the type is important and unspecified, it defaults to meaning a list of complex selectors.

  I took this to mean that the two features referred to the same kind of potential arguments.
  
  Finally, I tested to confirm that all browsers that support the selector list also supported simple, compound, and complex selectors as arguments to `:not` in the same release.

- I replaced the link to the Chromium bug with version numbers, where applicable, since this sat long enough that the feature has shipped in Chrome 88.

- I merged `master` in this branch, to pick up the new GitHub Actions workflows.

I'm merging this so that the data can be brought up to date with today's release. If you have any questions or want to make to changes to this, I'd welcome @ replies, issues, or PRs. Thanks!

PR #7378, opened 2020-11-17
Add real data for Safari for HTML element APIs

vinyldarkscratch:
This PR uses Safari (4-14) and Safari iOS (3-14) results from the mdn-bcd-collector project to add real values for the HTML element APIs.


sideshowbarker (approved):
❤️ ❤️ ❤️  really great to see these nulls and trues getting upgraded

PR #7384, opened 2020-11-17
mask border properties

rachelandrew:
I'm looking at the missing BCD reported in https://github.com/mdn/sprints/issues/3436

This PR is for the `mask-border-` properties. These have no support but most are implemented with a different name, prefixed in Safari and Chrome. I haven't been able to find out exactly when these versions went into Webkit, but Safari 6 seems to have them and Chrome 30, as I tested them.


rachelandrew:
ah I can't use true for Chrome anymore. Well, my best guess is these went into Safari 3.1, based on this 2008 blog post https://webkit.org/blog/181/css-masks/ which would put it into Chrome 1 I think? I've updated the others based on that, unless anyone has better knowledge.

css/properties/mask-border-outset.json
```diff
@@ -0,0 +1,109 @@
...
+          "support": {
+            "chrome": [
+              {
+                "version_added": false
+              },
+              {
+                "alternative_name": "-webkit-mask-box-image-outset",
+                "version_added": "1"
+              }
+            ],
```
ddbeck, line 16: When there's a support statement with `alternative_name`, then the `{ "version_added": false }` support statements isn't needed. Can you remove all such cases in this PR?

css/properties/mask-border-mode.json
```diff
@@ -0,0 +1,54 @@
+{
```
ddbeck, line 1: I'd drop this file. If it's unimplemented and there's no sign of implementation, then I don't think it should be in BCD.

(I know I need to resolve the whole content linter problem with missing BCD. Actually, there's a few problems with it now—thank you for your patience while I get to it.)

css/properties/mask-border.json
```diff
@@ -0,0 +1,109 @@
+{
+  "css": {
+    "properties": {
+      "mask-border": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/mask-bordere",
```
ddbeck, line 6:
```suggestion
          "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/mask-border",
```

ddbeck (changes requested):
A few small issues to resolve here before merging. Thank you, @rachelandrew!

rachelandrew:
@ddbeck thanks - I think I've covered those requests.

css/properties/mask-border-outset.json
ddbeck: The other files are fixed, but this one still has the unneeded `{ "version_added": false }` statements. We can merge as soon as that's done. Thank you!

rachelandrew: sorry about that - missed one - should be fixed now.

ddbeck (approved):
Thank you! 🎉

Issue #6772, opened 2020-09-29
Chrome Android shows compatibility for speechSynthesis.onboundary however doesnt work


clarodanesh:
On Chrome Android regardless of TTS engine selected, the onboundary event does not fire see: [bug](https://bugs.chromium.org/p/chromium/issues/detail?id=1122143&q=1122143&can=2)

joaoeudes7:
Any notice?

clarodanesh:
> Any notice?

no notice so far

sideshowbarker:
Raised https://github.com/mdn/browser-compat-data/pull/7385

PR #7385, opened 2020-11-17
Mark SpeechSynthesisUtterance.p.onboundary partial

sideshowbarker:
This change marks `SpeechSynthesisUtterance.p.onboundary` and the `boundary` event partially supported in Chrome Android, due to the fact the event doesn’t fire expected. See https://crbug.com/1122143. Fixes https://github.com/mdn/browser-compat-data/issues/6772.

### DIFF
--- a/api/SpeechSynthesisUtterance.json
+++ b/api/SpeechSynthesisUtterance.json
@@ -131,7 +131,9 @@
               "version_added": "33"
             },
             "chrome_android": {
-              "version_added": "33"
+              "version_added": "33",
+              "partial_implementation": true,
+              "notes": "Doesn't fire the <code>boundary</code> event as expected. See <a href='https://crbug.com/1122143'>bug 1122143</a>."
             },
             "edge": {
               "version_added": "14"
@@ -439,7 +441,9 @@
               "version_added": "33"
             },
             "chrome_android": {
-              "version_added": "33"
+              "version_added": "33",
+              "partial_implementation": true,
+              "notes": "Doesn't fire the <code>boundary</code> event as expected. See <a href='https://crbug.com/1122143'>bug 1122143</a>."
             },
             "edge": {
               "version_added": "14"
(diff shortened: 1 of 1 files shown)


api/SpeechSynthesisUtterance.json
```diff
@@ -131,7 +131,9 @@
               "version_added": "33"
             },
             "chrome_android": {
-              "version_added": "33"
+              "version_added": "33",
+              "partial_implementation": true,
+              "notes": "Doesn't fire the <code>boundary</code> event as expected. See <a href='https://crbug.com/1122143'>bug 1122143</a>."
```
ddbeck, line 136: I don't understand this note. What does it mean for the `boundary` event to not fire the `boundary` event?

sideshowbarker: > I don't understand this note. What does it mean for the `boundary` event to not fire the `boundary` event?

The engine has support for the event, but due to some yet-unresolved bug in the implementation, it doesn’t fire when expected on mobile — somewhat similar to the case for https://github.com/mdn/browser-compat-data/pull/6890 (the `visibilitychange` event in Safari).

foolip (approved):
Looks like a useful note to me, but I see that @ddbeck has self-assigned for review, so not merging.

api/SpeechSynthesisUtterance.json
ddbeck: Ah, OK. I misunderstood this in the first place—something about it struck me as recursive. What do you think of rephrasing this slightly?

```suggestion
              "notes": "The <code>boundary</code> event does not fire as expected. See <a href='https://crbug.com/1122143'>bug 1122143</a>."
```

api/SpeechSynthesisUtterance.json
```diff
@@ -439,7 +441,9 @@
...
```
ddbeck, line 446:
```suggestion
              "notes": "The <code>boundary</code> event does not fire as expected. See <a href='https://crbug.com/1122143'>bug 1122143</a>."
```

ddbeck (dismissed):

sideshowbarker (commented):
> What do you think of rephrasing this slightly?

Yep — thanks 👍

foolip (approved):
Changes look good, thanks!

PR #7387, opened 2020-11-17
Remove css.properties.word-wrap

bershanskiy:
Remove `css.properties.word-wrap` because it is a duplicate of `css.properties.overflow-wrap`.

As per [MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap):
> The property was originally a nonstandard and unprefixed Microsoft extension called word-wrap, and was implemented by most browsers with the same name. It has since been renamed to overflow-wrap, with word-wrap being an alias.

This data is tracked under the standard name [`overflow-wrap`](https://github.com/mdn/browser-compat-data/blob/master/css/properties/overflow-wrap.json), so `word-wrap.json` can be deleted.

Please note that there were some differences in two files on when either name was implemented, but `overflow-wrap` always reports an earlier version. I did not verify the data myself because I could not find release notes that old.

- [x] Summarize your changes


Barwick54 (approved):

saschanaz:
Should there be a new subitem in overflow-wrap to indicate support for the alias?

bershanskiy:
> Should there be a new subitem in overflow-wrap to indicate support for the alias?

I'm not sure what you mean. The support for `word-wrap` alias is indicated with `alternative_name` under `overflow-wrap` already. E.g., for Firefox it is stored as as [`css.properties.overflow-wrap.__compat.support.firefox[1]`](https://github.com/mdn/browser-compat-data/blob/3bcefe9ecbe7a05a329eb23fb2670b05e7fa0305/css/properties/overflow-wrap.json#L40):
```
{
  "alternative_name": "word-wrap",
  "version_added": "3.5"
}
```

This is displayed as "Uses the non-standard name: `word-wrap`" on the [documentation page](https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-wrap#Browser_compatibility):
image

Every browser has such note, reflecting the alias (aka `word-wrap`) support.

ddbeck (approved):
Nice find. Thank you! 🎉

PR #7394, opened 2020-11-18
Combine test-real-values into test-versions

vinyldarkscratch:
When I was looking at the code for `test-real-values` and `test-versions`, I had noticed that their code is nearly identical, with the same `for` loops iterating through the data, not to mention that they are both testing the versions within the data.  It seemed redundant for these to be two separate tests.  This PR merges the functionality from `test-real-values` into `test-versions`, so that the version checker accounts for our block list (the map of categories and browsers to require real data for).

Some additional code cleanup has been performed within this PR alongside these updates.


test/linter/test-versions.js
```diff
@@ -49,16 +49,50 @@ for (const browser of Object.keys(browsers)) {
...
+
 /**
  * Test to see if the browser allows for the specified version
  *
  * @param {string} browser The browser to check
  * @param {VersionValue} version The version to test
  * @returns {boolean} Whether the browser allows that version
  */
-function isValidVersion(browserIdentifier, version) {
+function isValidVersion(browser, category, version) {
```
caugner, line 91: JSDoc needs update

test/linter/test-versions.js
```diff
@@ -111,59 +145,60 @@ function addedBeforeRemoved(statement) {
...
+              logger.error(
+                chalk`{red → {bold ${relPath}} - {bold ${property}: "${
+                  statement[property]
+                }"} is {bold NOT} a valid version number for {bold ${browser}}\n    Valid {bold ${browser}} versions are: ${validBrowserVersions[
+                  browser
+                ].join(', ')}}${
+                  browserTips[browser]
+                    ? chalk`\n    {blue {bold Tip:} ${browserTips[browser]}}`
+                    : ''
+                }`,
```
caugner, line 186: Would it make sense to pass the tip as the 2nd parameter, like you did in other PRs?

```suggestion
                ].join(', ')}}`,
                browserTips[browser],
```

caugner (changes requested):

test/linter/test-versions.js
```diff
@@ -49,16 +49,48 @@ for (const browser of Object.keys(browsers)) {
...
+  'opera',
+  'opera_android',
+  'safari',
+  'safari_ios',
+  'samsunginternet_android',
+  'webview_android',
+];
+
+/** @type {object.<string, string[]>} */
+const blockList = {
```
caugner, line 69: What do `blockMany` and `blockList` actually represent? I'm not really sure.

test/linter/test-versions.js
```diff
@@ -111,33 +143,45 @@ function addedBeforeRemoved(statement) {
...
+            );
+          }
+
+          continue;
+        }
         const statementKeys = Object.keys(statement);
 
         for (const property of ['version_added', 'version_removed']) {
-          if (!isValidVersion(browser, statement[property])) {
+          if (
```
caugner, line 171: This block would profit from extracting a local variable `version`, which could also be used instead of `statement.version_removed`:

```
const version = statement[property];
```

caugner (approved):
LGTM, just one question and one refactoring opportunity.

test/linter/test-versions.js
queengooborg: `blockMany` is the list of all the browsers that we wish to have real values for, and `blockList` is the list of the browsers we require real values for per category.  I'm down for renaming them in a future PR to better clarify this!

caugner: Thanks! Renaming is a good idea, and maybe a JSDoc could be helpful.

queengooborg: Opened #16350 for that!

PR #7395, opened 2020-11-18
Chrome 87 added support for text-underline-offset

stapelberg:
Changelog:
https://chromium.googlesource.com/chromium/src/+log/86.0.4240.198..87.0.4280.66?pretty=fuller&n=10000
Commit:
https://chromium.googlesource.com/chromium/src/+/6ee93f1d7e463530d9423c79b5a58ca2bf881dcb

Tested manually in Chrome 87’s inspector.

- [x] Summarize your changes

### DIFF
--- a/css/properties/text-underline-offset.json
+++ b/css/properties/text-underline-offset.json
@@ -6,11 +6,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/text-underline-offset",
           "support": {
             "chrome": {
-              "version_added": false,
-              "notes": "See Chrome Platform Status entry for <a href='https://chromestatus.com/feature/5747636764147712'>text-decoration-thickness, text-underline-offset and from-font keyword for text-underline-position</a>."
+              "version_added": "87"
             },
             "chrome_android": {
-              "version_added": false
+              "version_added": "87"
             },
             "edge": {
               "version_added": false
@@ -52,7 +51,7 @@
               "version_added": false
             },
             "webview_android": {
-              "version_added": false
+              "version_added": "87"
             }
           },
           "status": {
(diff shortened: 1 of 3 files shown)


css/properties/text-underline-offset.json
```diff
@@ -6,8 +6,7 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/text-underline-offset",
           "support": {
             "chrome": {
-              "version_added": false,
-              "notes": "See Chrome Platform Status entry for <a href='https://chromestatus.com/feature/5747636764147712'>text-decoration-thickness, text-underline-offset and from-font keyword for text-underline-position</a>."
+              "version_added": "87"
             },
             "chrome_android": {
               "version_added": false
```
ddbeck, line 12:
```suggestion
              "version_added": "87"
```

The same would go for `webview_android` below

ddbeck (changes requested):
Hi @stapelberg and welcome to BCD! Thank you for this PR.

I see that the changes you linked to cover several changes at once: `text-decoration-thickness`, `text-underline-offset` and the `from-font` keyword for `text-underline-position`. Would you mind if we expanded the scope to include all of these at once?

Either way, it also appears that Chrome intends to ship this regardless of platform, so we'll need to change the corresponding values for `"chrome_android"` and `"webview_android"` at the same time. I've left a line comment to illustrate.

Thank you!

css/properties/text-underline-offset.json
stapelberg: Done

stapelberg:
All done!

ddbeck (approved):
Outstanding. Thank you! 🎉

PR #7396, opened 2020-11-18
schema: Allow optional implementation_url

sideshowbarker:
This change updates the compat-data schema and docs to allow per-browser optional `implementation_url` values to be specified in the support data for any browser; detailed description:

> An optional changeset URL or commit URL for the change which implemented the feature in the source code for the associated browser; e.g. a https://trac.webkit.org/changeset/ URL for a changeset whose subject line is in the form *"Enable [feature name/description]."* or *"Enable [feature name/description] by default."*. The presence of an `implementation_url` value indicates that the associated browser has spec-conformant support for the feature in its source code. Therefore, an `implementation_url` value can be added even if the feature hasn't yet shipped in a stable release, but instead only in a Nightly, Canary, or Technology Preview release — or even if the feature hasn't yet shipped in any release at all, but has only landed in the source code for the associated browser.

https://github.com/mdn/browser-compat-data/issues/6896#issuecomment-707928572 and https://github.com/mdn/browser-compat-data/issues/6896#issuecomment-708876962 are where the idea for this change arose. In summary, the rationale is to have a record in BCD of what changeset landed support for a feature in a particular browser’s codebase. Among other use cases, we have downstream consumers who care only whether or not there’s an implementation of the feature in a particular browser engine — even if that implementation hasn’t yet actually shipped in a major release.


ddbeck:
@sideshowbarker Would you mind illustrating this with an example or two? What would some data that uses this field actually look like?

types.d.ts
```diff
@@ -195,6 +195,13 @@ export interface SimpleSupportStatement {
      */
     value_to_set?: string;
   }[];
+
+  /**
+   * An optional changeset URL or commit URL for the change which implemented the feature in the
+   * source code for the associated browser; e.g. a https://trac.webkit.org/changeset/ URL.
+   */
+  implementation_url?: string;
```
saschanaz, line 203: Should an array be allowed, because I see some notes with two tracking bugs?

https://github.com/mdn/browser-compat-data/blob/aefff297d4f0ef9ba8132b626f8882254345e852/api/AudioParam.json#L260

sideshowbarker:
> @sideshowbarker Would you mind illustrating this with an example or two? What would some data that uses this field actually look like?

I’ve raised https://github.com/mdn/browser-compat-data/pull/7417 as a concrete example. As noted there, it adds an `implementation_url` value for `WritableStream` in Safari and iOS Safari, per https://trac.webkit.org/changeset/266172/webkit, which says:

> Implementation is now ready and is up to date with the latest specification.

As https://github.com/mdn/browser-compat-data/pull/6873#issuecomment-707688645 notes, that change means this feature shipped in at least TP 114 and probably was also in since TP 113.

With my downstream-consumer hat on, I can say that once that PR got merged, I could use the data downstream to correct the problem I described in https://github.com/mdn/browser-compat-data/issues/6896#issuecomment-707928572.

types.d.ts
sideshowbarker: > Should an array be allowed, because I see some notes with two tracking bugs?

We could allow an array (as we do already for `spec_url`) — but the reason I restricted it to a single string in the initial patch here is that I was not aware for any known cases where we’d need to specify more than one URL.

> https://github.com/mdn/browser-compat-data/blob/aefff297d4f0ef9ba8132b626f8882254345e852/api/AudioParam.json#L260

OK, looking at that case, I see the subject of https://bugzil.la/1171438 is *“Implement the new behavior for AudioParam.value”*, so that’s clearly a (yet-unimplemented) feature/implementation request — while looking at https://bugzil.la/1567777 see that one looks like it’s just a bug report (and one which is actually marked as a duplicate of the https://bugzil.la/1171438 feature request…)

So to me at least, that’s not a case that demonstrates a concrete need for allowing an array for the `implemenation_url` value — because the https://bugzil.la/1567777 URL isn’t a changeset URL at all, and neither actually is the https://bugzil.la/1171438, since the feature hasn’t actually been implemented yet.

saschanaz: Cool, but that's just a one example. Searching by `https://[^d].+https://[^d]` gives 61 results, so probably something to consider.

sideshowbarker: > Cool, but that's just a one example. Searching by `https://[^dw].+https://[^dw]` gives 43 results, so probably something to consider.
> 
> Edit: Those may also turn out as just bug reports, though. A short skimming couldn't find a valid example.

Right. But you’ve make me realize that for Firefox features, rather than an https://bugzil.la URL, what we’d instead want to have in `implementation_url` values is a https://hg.mozilla.org/mozilla-central/rev URL — for example:

https://hg.mozilla.org/mozilla-central/rev/8213cdcc0378

The subject of that changeset is *“Implement FetchEvent.handled”* and it happens to have a corresponding bug at https://bugzilla.mozilla.org/show_bug.cgi?id=1617822 but it seems clear that we’d want to cite the https://hg.mozilla.org/mozilla-central/rev/8213cdcc0378 URL instead — and so, to generalize from that, we’d want to cite https://hg.mozilla.org/mozilla-central/rev URLs rather than https://bugzil.la or https://bugzilla.mozilla.org URLs.

And then to generalize even further, I think regardless of which browser it’s for, we’d want to restrict the value to being a URL for the browser project’s revision log rather than for the project’s bug tracker.

So that means we’d want either a https://trac.webkit.org/changeset/, https://hg.mozilla.org/mozilla-central/rev/, or 
https://chromium.googlesource.com/chromium/src or https://crrev.com/ URL.

Given that, I’ve pushed another commit here that updates the docs part of patch to explicitly mention those — and I’ve also updated the schema change itself to explicitly limit the value to a pattern that matches URLs in one of those forms.

ddbeck:
This and the example PR are really helpful for understanding the proposal. Well done, @sideshowbarker! Before I start fussing with the documentation and JSON schema, I've got a couple big questions to make sure we're on the right track:

## What's the meaning of `implementation_url` in successive support statements?

Suppose we have some feature that has a story that goes a bit like this:

1. The feature starts out in an immature state and gets put behind a build flag or gated to Nightly or is only available in Safari TP. The JSON looks like this:

   ```json
   {
     "version_added": "false",
     "implementation_url": "https://…"
   }
   ```

   A consumer looks at this support statement and can read it as "not shipping, but in development".

   But as a reviewer, I've got questions. Suppose it's built behind a flag, incrementally, and there's multiple commits at this point. Which one do we link to? The first commit? The most recent? The most complete?

2. Soon it ships behind a preference, then generally:

   ```json
   {
     "version_added": "100",
     "implementation_url": "https://…"
   },
   {
     "version_removed": "100",
     "version_added": "99",
     "flags": ["…"],
     "implementation_url": "https://…"
   }
   ```

   Intuitively, I'd imagine this says to a consumer, "this has shipped and here's the proof." As a reviewer, I make sure that the URLs point to the commit that ships the preference as off-by-default and the commit that flips it.

   But maybe that's not quite right. Maybe I just carry forward the original implementation URL, from its time as an unreleased feature? Is `implementation_url` meant to be evidence of the implementation for the specific support statement? Or for the feature in general? The current schema doc, as written, suggests that any one is _valid_, but it's not clear to me which to prefer.

3. But then an intractible security flaw is discovered and everybody pulls it, `SharedArrayBuffer`-style:

   ```json
   {
     "version_removed": "101",
     "version_added": "100",
     "implementation_url": "https://…"
   },
   {
     "version_removed": "100",
     "version_added": "99",
     "flags": ["…"],
     "implementation_url": "https://…"
   }
   ```

   Now I've got a weird scenario. To a consumer, it's unclear what the most recent support statement means with the URL. Does `implementation_url` say, "here's the citation it was _added_" or "here's the citation it was _removed_"? Or do we forbid `implementation_url` in ambiguous scenarios like this?

## What's spec-conformant?

The documentation says:

> The presence of an `implementation_url` value indicates that the associated browser has spec-conformant support for the feature in its source code.

This sentence raises a bunch of somewhat related questions for me:

- As a reviewer, how do I know if a new implementation is, in fact, spec-conformant?
- Should we never use `implementation_url` on non-standard features? Why not?
- Should we prohibit `implementation_url` on features that don't have a `spec_url` or are `"standard_track": false`?
- What's the meaning of `implementation_url` alongside `partial_implementation`? Is _partly_ implementing a specification "spec-conformant?"
- What happens if a specification changes and the browser has not yet implemented the new specification. Should we retroactively remove the `implementation_url`?

ddbeck:
Oops, hit submit a little soon. For that last section, I guess I'm asking: why is `implementation_url` tied to the specification in this way?

sideshowbarker:
> For that last section, I guess I'm asking: why is `implementation_url` tied to the specification in this way?

Short answer: I’m not wedded to a need for including any language tying it to the specification; it’s not strictly necessary — so rather than _“spec-conformant support”_, I think _“complete support”_ or _”non-partial support”_ or such would be fine.

I think the goal of the relevant language here should just be to make it clear that the point at which we add an `implementation_url` feature is not the time when a browser project first starts adding initial code for a feature, but instead the time when the project has finished implementing the feature to the degree that they can/will be able to assert _“our browser supports this feature”_ (in, for example, their feature-status page such as https://webkit.org/status/ or equivalent).

> But as a reviewer, I've got questions. Suppose it's built behind a flag, incrementally, and there's multiple commits at this point. Which one do we link to? The first commit? The most recent? The most complete?

The most complete. In the case of the WebKit project, for example, that’s often a commit/revision which has a subject line in the form _“Enable the foo feature”_ or _“Enable the foo feature by default”_ — but in general it can also be often by _“Implement the foo feature”_ or _“Support the foo feature”_.

Determining whether which particular commit/revision really adds usable support for a feature is a qualitative judgement that depends on the BCD change author having some experience and familiarity with each browser projects’ conventions.

While I can confidently assert that, given a particular browser-project commit/revision, I personally know how to figure out whether it really represents implementation of fully-usable support, I don’t know how to document the criteria for that succinctly without adding lots of extra words to the docs.

And it seems like the relevant part of the docs for this is already pretty long, so I’d suggest that a better place to try to document that guidance would probably be the part of the MDN article at https://developer.mozilla.org/en-US/docs/MDN/Contribute/Processes/Matching_features_to_browser_version, or an additional new article to supplement that.

> As a reviewer, I make sure that the URLs point to the commit that ships the preference as off-by-default and the commit that flips it.

Yes, that. I think we should always consider the `implementation_url` value to be mutable/disposable/replaceable. We should not try to make it into something that attempts to preserve the entire change history of multiple revisions.

> The current schema doc, as written, suggests that any one is _valid_, but it's not clear to me which to prefer.

I believe what makes sense is that we should always prefer the URL latest revision — in particular, in even if all that revision does it to flip a preference/flag value to cause the feature to be exposed by default. That is in fact what many of the relevant WebKit revisions actually do — the ones worded _“Enable the foo feature by default.”_

> Or do we forbid `implementation_url` in ambiguous scenarios like this?

That, I think. In cases like that, we just completely remove the `implementation_url`. That’s in keeping with the sense that the value of disposable; and when we ever get to a point where a particular `implementation_url` isn’t clearly serving a purpose for us — or worse, gets confusing or problematic (in, e.g., the way described above — then we just rip it out.

> As a reviewer, how do I know if a new implementation is, in fact, spec-conformant?

As I mentioned earlier, maybe we could preempt that question by replacing _“spec-conformant implementation”_ with _“complete implementation”_ or something. At least that gets around the problem of it being tied to a spec.

But that aside, generally, you shouldn’t need to independently determine whether the implementation is spec-conformant. I mean, we also don’t do that when browser projects assert _“This release adds support for this feature.”_ 

So instead in the case for pre-release landing of features, you just make a qualitative assessment based on your experience and your understanding of the relevant browser project’s conventions with respect to changeset messages; that is, you basically say, _“OK this is changeset message follows the same kind of pattern I am familiar with for features that end of being included in he release notes for the next major release.”_

But that said, if instead you really did want to independently confirm a browser has spec-conformant support, you could do that be running the relevant WPT tests and checking the results, and/or by reading the spec and running your own tests.

> Should we never use `implementation_url` on non-standard features? Why not?

That’s a good question. I’m not actually sure what the answer should be.

The problem with non-standard features is that there’s no way for anybody to ever evaluate how complete they are. So it actually makes no sense to state _“This browser has complete support for non-standard feature foo.”_

The problem is that without a standard spec that defines the requirements of the feature, nobody can even write tests that can be used to check how complete the support for a feature is in a particular browser.

> Should we prohibit `implementation_url` on features that don't have a `spec_url` or are `"standard_track": false`?

Again I’m actually not sure what the answer should be. But I’m inclined to say that we should just cross that bridge when we get there. That is, if/when a BCD contributor submits a patch/PR with an `implementation_url` for a feature that has no spec, then at that time we’ll have a concrete case to consider.

> Is _partly_ implementing a specification "spec-conformant?"

I think the answer to that is definitely No. "spec-conformant" means it fully implements the spec, without any known/documented non-conforming behavior.

> What's the meaning of `implementation_url` alongside `partial_implementation`?

I think the clear answer to that needs to be, it has no meaning, and we should prohibit it.

I don’t think JSON Schema provides us with a practical way to express prohibiting `partial_implementation` with `implementation_url` at the schema level — but certainly we can make the linter prohibit it. And we should.

> What happens if a specification changes and the browser has not yet implemented the new specification. Should we retroactively remove the `implementation_url`?

That’s a good question —  but it seems like probably another _“let’s cross that bridge when we get it”_, where once we hit a case of that actually happening, we’d have something concrete to consider.

But my initial sense of it is, if a spec changes to add new methods or attributes to an existing API, then those additions essentially constitute a new feature — and all browser projects are going to treat them as such, and document them as such. And for us with BCD, that means we’d anyway have a separate `implementation_url` added to each new method/attribute.

The other case to consider of course is the case where the spec changes to remove some existing feature. But that was brought up in an earlier comment, and to re-state what I said in my response to that: I think at the point where the relevant browser removes the feature, we completely yank the `implementation_url` from our data.

And even prior to that — and to the point about what _“spec-conformant”_ means, and whether we should prohibit `implementation_url` on features that don't have a `spec_url` or are `"standard_track": false` — I agree our data would be left in kind of a weird state if 1) a browser has implemented a feature and we’ve added an `implementation_url` for it, then 2) the feature is subsequently dropped from a spec, and 3) the browser has not yet removed their implementation of the feature.

But I think that’s really another kind of _“let’s cross that bridge when we get it”_ situation.

foolip:
What's the status of this? I've wanted to add a few links like this in the past few days, but it would fail the lint like https://github.com/mdn/browser-compat-data/pull/7417.

ddbeck:
This has come up a couple of times in the past few days. Re-adding it to my (ideally short-term) review queue

schemas/compat-data-schema.md
```diff
@@ -369,6 +369,19 @@ Example for two flags required:
 }
 ```
 
+#### `implementation_url`
+
+An optional changeset URL or commit URL for the revision that implemented the feature in the
```
ddbeck, line 374: A few things here:

1. What do you think of breaking this up visually?
2. Based on the discussion, can we drop "spec-conformant" bit? I think hashing out "is this URL good enough for `implementation_url`?" is probably better suited to the data guidelines; consumers ought to be able to trust that we've done some adequate vetting that the URL demonstrates support without needing to know the gritty details.
3. Wordsmithing the "even if the feature hasn't shipped" stuff, particularly to make "Safari Technology Preview" more searchable.

Here's sorta what I had in mind (open to modification/reinterpretation):

> An optional URL that links to the commit or changeset for the revision that implemented the feature in the source code for the browser or underlying engine.
> 
> For example, a URL from:
> 
> - `https://trac.webkit.org/changeset/`
> - `https://hg.mozilla.org/mozilla-central/rev/`
> - `https://crrev.com/`
> 
> Typically, an applicable change has a subject line in the form of _"Implement [feature]"_,
_"Support [feature]"_, or _"Enable [feature]"_.
>
> The presence of an `implementation_url` value indicates that the associated browser has support for the feature in its source code. Therefore, an `implementation_url` value can be added even if the feature
hasn't yet shipped in a stable version, but may be available in a development version (such as Firefox Nightly, Chrome Canary, or Safari Technology Preview).

schemas/compat-data.schema.json
```diff
@@ -39,6 +39,12 @@
             "required": ["type", "name"]
           }
         },
+        "implementation_url": {
+          "type": "string",
+          "format": "uri",
+          "pattern": "^https:\/\/(trac.webkit.org\/changeset\/|hg.mozilla.org\/mozilla-central\/rev\/|chromium.googlesource.com\/chromium\/src\/|crrev.com\/)",
```
ddbeck, line 45: This regex is a bit complex. Can we use [a `oneOf` array](https://json-schema.org/understanding-json-schema/reference/combining.html#oneof) to provide a set of patterns for each engine/vendor? Something like this, I think:

```
"oneOf": [
            {
              "pattern": "^https://trac\.webkit\.org/changeset/",
            },
            {
              "pattern": "^https://hg\.mozilla\.org\/mozilla-central",
            },
          ]
```

ddbeck (changes requested):
OK, there's some concrete suggestions in line comments.

One thing I'm thinking about though, after all the discussion (especially for all the questions about the corner cases), is that we might want to constrain ourselves initially. Perhaps _only_ use this for features in development, then drop such links when features properly ship. I'm at this point where I trust the schema to express lots of things _except_ for the thing that this plainly fixes (unreleased features). But I'm not sure I know how this proposal interacts with things that the schema already copes with, so maybe the answer, for now, is to simply not let those things interact.

queengooborg:
Hey @sideshowbarker, I think this is a great feature and would be very helpful to include!  It looks like this is awaiting a rebase and a response from you -- are you still up for working on this PR?

sideshowbarker:
> Hey @sideshowbarker, I think this is a great feature and would be very helpful to include! It looks like this is awaiting a rebase and a response from you -- are you still up for working on this PR?

I can’t justify the time needed to respond to the open comments on this and finish it up. I can no longer see adding this as a high priority, because the problem I’d originally hoped this could help solve was the problem of not being able  to add a version identifier for features that had shipped in Safari Technology Preview — but that problem was subsequently solved by the policy and schema were changed to allow `preview` as a version identifier.

And while I do still think there’s some ancillary value in being able to add implementation URLs to features, I can’t see enough value in it personally to justify taking the time needed to shepherd this PR the rest of way through the process. And once we had this, I don’t think I’d be spending time adding implementation URLs to any features — so unless others are planning to add them, I think this could end up just being something we put work into but never ended up actually using.

queengooborg:
That makes sense -- I guess with the collector as well, there's even less incentive to track down bug URLs and add them to BCD, so it might end up unmaintained like the `matches` object.  I'm down to close this PR then, or if it's a feature others still desire and would commit to adding data for, I'm more than happy to take over!

queengooborg:
I think this is still a helpful feature, so I've opened #16415 to continue the work done in this PR.  Thank you @sideshowbarker for getting the ball rolling on this!

Issue #7374, opened 2020-11-16
Publish pre-build bundles to npm


bershanskiy:
Please consider "bundling" all of the data for releases on npm to improve execution speed, reduce size of dependents' `node_modules` folder, and support legacy NodeJS versions.

Consider the following simple script that just loads BCD and stores all of it in a single file `export.json`.
```
'use strict';
const fs = require('fs');
const bcd = require('@mdn/browser-compat-data');
const string = JSON.stringify(bcd);
fs.writeFileSync('export.json', string);
```

Let's compare the current `@mdn/browser-compat-data` release and its "bundled" version. A minimal "bundled" version could include `export.json` produced by the script above and BCD's `LICENSE`, `README.md`, and `package.json`.

### Load time
Currently, `index.js` walks the directories and loads data files synchronously because it needs to be `require`'d by dependents. Numerous synchronous directory lookups and file reads cause it to be slow.

I flushed the file system cache before each measurement to make comparisons as fair as possible. I measured time during a few runs and results fluctuated a bit. On SSD, "real" time of `bundled.js` never exceeded 0.3s while "real" time of `current.js` never dropped below 1s. On HDD, "real" time of `bundled.js` never exceeded 1s while "real" time of `current.js` never dropped below 4s. (The "user" time of 0.000s looks suspicious to me, but it might be just Windows miss-reporting very small time intervals. The "real" time measurements do feel reasonable, especially the ones over 1s.)

#### Tests
```
$ cat current.js
'use strict';
const a = require('@mdn/browser-compat-data');

$ cat bundled.js
'use strict';
const a = require('./export.json');
```

#### Sample results
On a modern NVMe SSD:
```
$ time node bundled.js
real    0m0.265s
user    0m0.000s
sys     0m0.015s

$ time node current.js
real    0m1.089s
user    0m0.000s
sys     0m0.015s
```

On a slow HDD:
```
$ time node bundled.js
real    0m0.521s
user    0m0.000s
sys     0m0.000s

$ time node current.js
real    0m4.606s
user    0m0.015s
sys     0m0.030s
```

### File size
Size of `node_modules` containing current `@mdn/browser-compat-data` and `extend` (required to load BCD): 18,333,922 bytes
Size of minimal "bundled" version: 8,989,309 bytes

Size of minimal "bundled" version is the sum of size of `export.json` produced by the script above and sizes of BCD's `LICENSE`, `README.md`, and `package.json`.

[...]

PR #7398, opened 2020-11-18
Publish pre-built bundles to npm

bershanskiy:
This is a simple implementation of publishing "bundled" releases to npm, which are much smaller and load faster than the current releases. Background: https://github.com/mdn/browser-compat-data/issues/7374 .

## Benchmark (updated)
```
$ npm publish dist/ --dry-run --access public
$ npm publish dist/ --dry-run --access public
npm notice
npm notice 📦  @mdn/browser-compat-data@2.0.6
npm notice === Tarball Contents ===
npm notice 6.6kB   LICENSE
npm notice 6.7kB   README.md
npm notice 440.2kB data.json.gz
npm notice 337B    index.d.ts
npm notice 163B    index.js
npm notice 604B    package.json
npm notice 11.2kB  types.d.ts
npm notice === Tarball Details ===
npm notice name:          @mdn/browser-compat-data
npm notice version:       2.0.6
npm notice filename:      @mdn/browser-compat-data-2.0.6.tgz
npm notice package size:  450.1 kB
npm notice unpacked size: 465.8 kB
npm notice shasum:        7698d4a9ec5137486e11062d1eae762528066e81
npm notice integrity:     sha512-RRTkhnEmu3DuA[...]GeIUgaXxjZ6uQ==
npm notice total files:   7
npm notice
+ @mdn/browser-compat-data@2.0.6
```

## Release contents
 - The following files are released verbatim because dependents might actually want to read them:
    - LICENSE
    - README.md
    - index.d.ts and types.d.ts
 - package.json is stripped of entries not needed by dependents and npm (e.g., `scripts`, which describes script files already omitted from npm releases) 
 - index.js is "bundled"

- [x] Summarize your changes


vinyldarkscratch:
I've only just been skimming the code on my iPhone, but I'm liking where this PR is going!  Just a couple of thoughts:

- We should move the basic file copying from the GitHub Action into the dist generation script.  I think it's more reasonable to keep it all in one place, plus we can test it easier.
- For the version number, wouldn't you mean "v2.1.0" (not "v2.1.6")?  Regardless, I think we should leave the version bump to Daniel during the publish stage.

vinyldarkscratch:
I'm loving how this is looking, great work!  I love the idea of a smaller and faster package, not to mention that would remove the need for additional dependencies, or worry about the NodeJS version at all.  I spent some time reviewing this code, and I've just got a few suggestions:

- We should move all of the preparation work into the script, so that the copying of the files like README and other are all performed in the same script.  To me, it seems more reasonable to keep the process in one place from start to finish.
- Did you mean "v2.1.0" rather than "v2.1.6"?  Regardless, I'm thinking that we should keep the version bumping out of this PR and let it be performed as a part of the release process.
- Perhaps we should separate the JSON data into its own file, such as `data.json`?  Organizationally, I feel that makes a little more sense.
  - Extending on this, it might be nice to compress the bundle even further with a little gzip compression.  I was doing some experimentation and research, and by compressing the JSON, the file size drops down to 440 KB (vs 9 MB).  What's more, I'm not seeing performance drawbacks, and zlib functions are available since NodeJS v0.10!

vinyldarkscratch:
@ddbeck, what do you think?

bershanskiy:
@vinyldarkscratch Thanks for your comments!

> We should move all of the preparation work into the script, so that the copying of the files like README and other are all performed in the same script. To me, it seems more reasonable to keep the process in one place from start to finish.

Ok, will do.

> Perhaps we should separate the JSON data into its own file, such as data.json? Organizationally, I feel that makes a little more sense.

As is, I use `index.js` file name just to match previous name. I think we could change it to anything (e.g., `data.json`) and just need to set `package.json` key `main` to that name.

> Extending on this, it might be nice to compress the bundle even further with a little gzip compression. I was doing some experimentation and research, and by compressing the JSON, the file size drops down to 440 KB (vs 9 MB). What's more, I'm seeing no performance drawbacks (about 0.22s to load), and zlib functions are available since NodeJS v0.10 (confirmed it works in NodeJS v4)!

Are you suggesting that to reduce download size or at-rest size in `node_modules`?

As far as I know, npm already gzips packages in transit (and uncompresses them automatically). According to the a log of a dry `npm publish` run above, npm compresses the whole package to 449.1 KB (vs. 440 KB you saw for data). Since npm reports the whole package size, the extra 9.1 KB likely came from `LICENSE` and `README.md`, etc.

Storing package in compressed format in `node_modules` and then un-compressing it on every run would indeed save a lot of space. That's actually the case for most packages on npm, so I wonder why npm/node does not do it automatically. Perhaps, it's a co

vinyldarkscratch:
> As is, I use `index.js` file name just to match previous name. I think we could change it to anything (e.g., `data.json`) and just need to set `package.json` key `main` to that name.

I think `index.js` is a good name for an entry point!  I did some experimentation and unfortunately found that we are not able to specify a JSON file as the `main` key; it has to be a script.

> Are you suggesting that to reduce download size or at-rest size in `node_modules`?

I'm referring to the at-rest size when the package is downloaded and installed.

bershanskiy:
@vinyldarkscratch I added GZip, moved file copy code and changed around a few things. Most notably, now bundling script uses async I/O for file copying and checks for errors. Also, bundle now requires (and was tested) on NodeJS 4.2.0 released in 2015-- the earliest LTS release in 4.* branch, as far as I see.

bershanskiy:
Is there anything that I should do to advance this PR?

ddbeck:
I plan to review this (and the original issue) more thoroughly, but it may take some time (with some time off at the end of this week). My very quick impression here:

* This is a clever approach and solves some real problems. I want to explore the ideas more thoroughly.
* There's some risks here that I haven't thought about long enough yet to assess properly, but concern me nonetheless. A couple off the top of my head:
  * It looks like our tests never actually run against the code that ships to npm. It's not clear to me how I _would_ test the shipping code. That's… worrying, to say the least.
  * This introduces multiple new kinds of complexity (building _and_ compressing). I suspect I'd want to defer the compression stuff until later, but I haven't read closely enough to see if it's essential.
* This uses GitHub Actions to do some of the publishing work here, which makes local testing difficult. It seems like we should use idiomatic [npm life cycle scripts](https://docs.npmjs.com/cli/v6/using-npm/scripts) instead.

ddbeck:
Oh one more thought: I would also like to look at some of the existing tooling for building distributables (e.g., webpack) and make sure we're not reinventing something.

bershanskiy:
@ddbeck I'll split out the compression part to simplify review.

> It looks like our tests never actually run against the code that ships to npm. It's not clear to me how I would test the shipping code. That's… worrying, to say the least.

The current script assumes that the current system (`index.js` with extend) works. It only checks that all the file I/O operations returned no errors.

We could test the following:
 1. Do a npm publish dry run and check that npm actually picks up all files we care about and that their hashes are what we expect.
   Fortunately, most files (`README`, `LICENSE`) don't have to change frequently (and bundled `index.js` changes only in its `version`, so hashes can be pre-computed in advance for, e.g. 100 future releases). `data.json` is the only file that has to change every release.
 2. Add test that ensures that the bundle is really equivalent to the current system. For example, something trivial like:
     ```
     const real = require('./index');
     const bundle = require('./dist/index');
     deepCompare(real, bundle); // 
     ```

> This introduces multiple new kinds of complexity (building _and_ compressing). I suspect I'd want to defer the compression stuff until later, but I haven't read closely enough to see if it's essential.

Compression is not essential (I actually excluded it initially to simplify review). I would be glad to defer compression for later.

>Oh one more thought: I would also like to look at some of the existing tooling for building distributables (e.g., webpack) and make sure we're not reinventing something.

I don't think anyone bothered to write a tool for a use case as simple as this one. Most bundling tools do a lot of things like dependency resolution, tree shaking, transpiring, etc. In our case, none of that is needed.

bershanskiy:
I re-wrote the PR from scratch and made everything much simpler. Bundling script still assumes 

Regarding earlier comments by @ddbeck :

> It looks like our tests never actually run against the code that ships to npm. It's not clear to me how I would test the shipping code. That's… worrying, to say the least.

I can create the following process with an end-to-end test (that could be run as an Action or locally):
1. Create a bundle, checking that its content matches the current `require('index.js')`
2. Run `npm publish --dry-run` to ensure it picks up all the desired files
2. Push that bundle to npm as a "private" release
3. Download that private pre-release and ensure it matches the current `require('index.js')`
4. Only after that change visibility of already published tag to "public"

Unfortunately, only paying npm users can push private releases.

> we should use idiomatic npm life cycle scripts

I experimented with lifecycle scripts in #7513 and found them too inflexible. I found two main inconveniences:
 - having to use the same `package.json`, which means lots of unnecessary entries (`scripts`, `devDependencies`, etc.) and the same `main` (which among other things means `index.js` it has to determine when it's installed as a dependency or in development environment to ensure tests are [loading and testing](https://github.com/mdn/browser-compat-data/blob/f5d1bd642fab0e399e08c150b7d3ab73fa604feb/test/linter/test-versions.js#L11) the real data and not previously built bundle).
 - being able to run build steps and tests selectively in any order

ddbeck:
@bershanskiy Thank you for the updates on this. I'm sorry its taken me so long to come back to this. I do plan to return to this, but I'm reluctant to merge a change like this close the launch of Yari (our most visible dependent). Watch this space. 😄

bershanskiy:
@ddbeck Of course, take your time! Bug-free launch is more exciting than a rushed feature. :)

scripts/release-build.js
```diff
@@ -0,0 +1,94 @@
...
+function writeManifest() {
+  const dest = path.resolve('./', directory, './package.json');
+  const manifest = createManifest();
+  const promise = fs.writeFile(dest, manifest);
+  return promise;
+}
+
+async function main() {
+  // Remove existing files, if there are any
+  const deletedDir = await fs.rm(directory, {
```
foolip, line 73:
```suggestion
  await fs.rmdir(directory, {
```

It looks like `fs.rm` was added in Node 14, so this didn't work for me locally. (I stay on Node 12 because some web services I deploy still run on it, makes it easier to not accidentally depend on new stuff.)

Also this resolves with undefined on success, so `deletedDir` isn't accurate and the return value isn't important.

scripts/release-build.js
```diff
@@ -0,0 +1,94 @@
...
+    const dest = path.join(directory, file);
+    const promise = fs.copyFile(src, dest);
+    promises = promises.concat(promise);
+  }
+  return promises;
+}
+
+function createManifest() {
+  const full = require('../package.json');
+  const minimal = {};
```
foolip, line 39:
```suggestion
  const minimal = {main: 'data.json'};
```

I couldn't load the built module without this.

scripts/release-build.js
```diff
@@ -0,0 +1,94 @@
...
+    'bugs',
+    'homepage',
+    'types',
+  ];
+
+  for (const key of allowList) {
+    if (key in full) {
+      minimal[key] = full[key];
+    } else {
+      throw `Could not create a complete manifest! ${key} is missing!`;
```
foolip, line 58: Given this error, `allowList` is a bit of a misnomer. These are required, not merely allowed, so maybe `minimalKeys` as an alternative?

scripts/release-build.js
```diff
@@ -0,0 +1,94 @@
...
+  }
+  return JSON.stringify(minimal);
+}
+
+function writeManifest() {
+  const dest = path.resolve('./', directory, './package.json');
+  const manifest = createManifest();
+  const promise = fs.writeFile(dest, manifest);
+  return promise;
+}
```
foolip, line 69:
```suggestion
async function writeManifest() {
  const dest = path.resolve('./', directory, './package.json');
  const manifest = createManifest();
  await fs.writeFile(dest, manifest);
}
```

scripts/release-build.js
```diff
@@ -0,0 +1,94 @@
...
+  return string;
+}
+
+// Returns a promise for writing the data to JSON file
+function writeData() {
+  const dest = path.resolve('./', directory, './data.json');
+  const data = createDataBundle();
+  const promise = fs.writeFile(dest, data);
+  return promise;
+}
```
foolip, line 23:
```suggestion
async function writeData() {
  const dest = path.resolve('./', directory, './data.json');
  const data = createDataBundle();
  await fs.writeFile(dest, data);
}
```

I'm suggesting a few changes like this that make the code shorter and more obviously async.

scripts/release-build.js
```diff
@@ -0,0 +1,94 @@
...
+function copyFiles() {
+  let promises = [];
+  for (const file of verbatimFiles) {
+    const src = path.join('./', file);
+    const dest = path.join(directory, file);
+    const promise = fs.copyFile(src, dest);
+    promises = promises.concat(promise);
+  }
+  return promises;
+}
```
foolip, line 35:
```suggestion
async function copyFiles() {
  for (const file of verbatimFiles) {
    const src = path.join('./', file);
    const dest = path.join(directory, file);
    await fs.copyFile(src, dest);
  }
}
```

I'm thinking there are so few files to copy that parallelizing won't make it noticeably faster.

scripts/release-build.js
```diff
@@ -0,0 +1,94 @@
...
+  // Remove existing files, if there are any
+  const deletedDir = await fs.rm(directory, {
+    force: true,
+    recursive: true,
+  });
+  if (deletedDir !== undefined) throw deletedDir;
+
+  // Crate a new directory
+  const createdDir = await fs.mkdir(directory);
+  if (createdDir !== undefined) throw createdDir;
```
foolip, line 81:
```suggestion
  await fs.mkdir(directory);
```

scripts/release-build.js
```diff
@@ -0,0 +1,94 @@
...
+
+  const promises = [writeManifest(), writeData(), ...copyFiles()];
+  let errors = await Promise.all(promises);
+  errors = errors.filter(e => e !== undefined);
+  if (errors.length !== 0) {
+    // Re-throw all errors
+    throw errors;
+  }
```
foolip, line 89:
```suggestion
  await writeManifest();
  await writeData();
  await copyFiles();
  
  console.log('Data bundle is ready');
```

Suggesting making this as simple as possible, without parallelism.

scripts/release-build.js
```diff
@@ -0,0 +1,94 @@
...
+}
+
+// This is needed because NodeJS does not support top-level await.
+// Do not catch errors so that NodeJS fails on them.
+main().then(console.log('Data bundle is ready'));
```
foolip, line 94:
```suggestion
main().catch(e => {
  console.error(e);
  process.exit(1);
});
```

Without this there will be an unhandled promise rejection on error. Also, previously the `console.log` call would have happened immediately, it should have been `() => console.log(...)` but I suggest moving it into `main()` for simplicity.

foolip (commented):
Very excited about this direction! I tried it locally and made a bunch of suggestions.

scripts/release-build.js
bershanskiy: OK. We can always make it more parallel later if there is need for that. If I recall correctly, there was a large speedup from using async I/O in general, but impact of awaiting for `writeManifest`, `writeData` and `copyFiles` in parallel was measureable only on external HDD connected via USB2.0.

foolip: @bershanskiy I thought that your measurements were for loading the resulting package, not for preparing it? I think it would be fine if running this script took a long time, since it's only done by maintainers. It doesn't seem to actually be slow, however.

If one wanted to speed this up by parallelizing I/O, I think the loading of the JSON files in `index.js` is where the wins would be.

scripts/release-build.js
```diff
@@ -0,0 +1,89 @@
...
+
+  await writeManifest();
+  await writeData();
+  await copyFiles();
+
+  console.log('Data bundle is ready');
+}
+
+// This is needed because NodeJS does not support top-level await.
+// Do not catch errors so that NodeJS fails on them.
```
foolip, line 85: Oops, this comment doesn't work any more. This now does catch errors, so that node *does* fail in the right way.

scripts/release-build.js
bershanskiy: Done!

scripts/release-build.js
```diff
@@ -0,0 +1,89 @@
...
+// Returns a string representing data ready for writing to JSON file
+function createDataBundle() {
+  const bcd = require('../index.js');
+  const string = JSON.stringify(bcd);
+  return string;
+}
+
+// Returns a promise for writing the data to JSON file
+async function writeData() {
+  const dest = path.resolve('./', directory, './data.json');
```
foolip, line 19:
```suggestion
  const dest = path.resolve(directory, 'data.json');
```

I didn't know that `path.resolve` could take any number of arguments and be like `path.join`, but since it is, this simplification seems possible.

scripts/release-build.js
```diff
@@ -0,0 +1,89 @@
...
+      minimal[key] = full[key];
+    } else {
+      throw `Could not create a complete manifest! ${key} is missing!`;
+    }
+  }
+  return JSON.stringify(minimal);
+}
+
+async function writeManifest() {
+  const dest = path.resolve('./', directory, './package.json');
```
foolip, line 61:
```suggestion
  const dest = path.resolve(directory, 'package.json');
```

scripts/release-build.js
bershanskiy: Good catch! Does this comment look better?
> // This is needed because NodeJS does not support top-level await.
> // Also, make sure to log all errors and exit with failure code.

foolip:
This is looking close to done now, but a few of the concerns @ddbeck raised in https://github.com/mdn/browser-compat-data/pull/7398#issuecomment-733781277 remain.

We never test this in CI, so it could be entirely broken and tests would still pass. A simple kind of test for this would be to build the package, then require both the built package and the regular `index.js`, and use `assert.deepEquals`.

An alternative setup would be to always have the build step and require that for local development too, but I don't think that would really be an improvement, as one would have to run it basically for any change, or risk running the tests against the wrong thing.

scripts/release-build.js
bershanskiy: Done!

scripts/release-build.js
bershanskiy: Done!

scripts/release-build.js
bershanskiy: Done!

bershanskiy:
> We never test this in CI, so it could be entirely broken and tests would still pass. A simple kind of test for this would be to build the package, then require both the built package and the regular index.js, and use assert.deepEquals.

I added a Mocha test for this in `scripts/release-build.test.js`. The only reservation I have is that I have to specify timeout, since bundling script now takes over 2 seconds on GitHub Actions.

scripts/release-build.js
bershanskiy: Done!

bershanskiy:
@vinyldarkscratch A while ago you wrote:
> I think index.js is a good name for an entry point! I did some experimentation and unfortunately found that we are not able to specify a JSON file as the main key; it has to be a script.

Please note that now `main` is `'data.json'` and it works for me, for @foolip, and on GitHub Actions. Could you check whether this PR works for you locally now? I would hate to break something accidentally. Thank you.

scripts/release-build.js
```diff
@@ -0,0 +1,93 @@
...
+  const manifest = createManifest();
+  await fs.writeFile(dest, manifest);
+}
+
+async function main() {
+  // Remove existing files, if there are any
+  await fs
+    .rmdir(directory, {
+      force: true,
+      recursive: true,
```
ddbeck, line 71: By switching Node versions (with `nodenv`), I discovered that the `recursive` option doesn't work on Node 10. But it's EOL in ~2 weeks. I think we can tolerate the (unlikely) annoyance for that long.

ddbeck: Actually, now that I think on it some more: I guess it doesn't matter at all. If all we publish is a `.json` file, then we can adopt whatever version of Node we want for development.

(Though this also suggests that, should we desire to expose utilities for working with compat data, it would need to be as a separate package.)

ddbeck:
> > We never test this in CI, so it could be entirely broken and tests would still pass. A simple kind of test for this would be to build the package, then require both the built package and the regular index.js, and use assert.deepEquals.
> 
> I added a Mocha test for this in `scripts/release-build.test.js`. The only reservation I have is that I have to specify timeout, since bundling script now takes over 2 seconds on GitHub Actions.

Since this test is slow, perhaps we can put mocha after the linting, for the convenience of routine edits?

```
// in package.json
"test": "npm run lint && npm run mocha"
```

We can probably do something to improve the performance of `index.js` to properly fix this, though.

scripts/release-build.js
bershanskiy: > By switching Node versions (with nodenv), I discovered that the recursive option doesn't work on Node 10.

Yes, the [docs](https://nodejs.org/api/fs.html#fs_fspromises_rmdir_path_options) say `recursive` was introduced in 12.

> If all we publish is a .json file, then we can adopt whatever version of Node we want for development.

I'll confess: it was my secret plan all along :) . More recent NodeJS versions have nice features, so I wanted to simplify transition to modern NodeJS should you ever choose to.

> (Though this also suggests that, should we desire to expose utilities for working with compat data, it would need to be as a separate package.)

Does BCD expose any utilities now?

bershanskiy:
> Since this test is slow, perhaps we can put mocha after the linting, for the convenience of routine edits?

Done!

> We can probably do something to improve the performance of index.js to properly fix this, though.

Yes, `index.js` can be made more async by using generic file I/O and `JSON.parse()`. Problem is, `require()` is syncroneous, so BCD dependents can not `require()` and get the object immediatelly. They'd have to wait for a promise or provide a callback (yak!) just to get the data. I acyually started this PR partially because I could not find a way to improve `index.js`. If we publish JSON bundles, then we can make all building steps async and arbitrarily complex!

However, tests temselves can be improved (and made more async):
  - some stuff can be moved out of linter (and syncroneous calls) and into Mocha (for example, #9230)
  - some linting steps can be made async (e.g., [this](https://github.com/mdn/browser-compat-data/blob/main/test/test-format.js#L7)), but that makes console printout harder (e.g., spinners no longer work properly)

foolip:
@bershanskiy I think you're on the right track with the performance. Right now, we do I/O (`fs.readFileSync`) and parsing (`JSON.parse`) in tandem and thus aren't reading from disk as fast as we could.

I did an experiment to first create promises for all the file reads, to read everything in parallel. That can being down the (uncached) time by perhaps half a second, but doesn't do much when cached.

I think that loading all files in parallel isn't the optimum though, probably one could do better by reading one at a time but never stopping for `JSON.parse`. But since this will only affect contributors and not users of BCD, there's a limit to how complex it's worth making this.

foolip:
Here's the async variant of `load()` that I tried:
```js
async function load(...dirs) {
  const result = {};

  const stack = dirs.map(dir => path.resolve(__dirname, dir));
  const promises = [];

  while (stack.length) {
    const dir = stack.pop();
    const entries = await fs.readdir(dir, {withFileTypes: true});
    for (const ent of entries) {
      const entpath = path.join(dir, ent.name);
      if (ent.isDirectory()) {
        stack.push(entpath);
      } else if (path.extname(entpath) === '.json') {
        promises.push(fs.readFile(entpath).then(JSON.parse));
      }
    }
  }

  for (const p of promises) {
    const data = await p;
    extend(result, data);
  }

  return result;
}
```

A remaining problem is that it changes the order of the keys in the objects, which isn't even going to be deterministic on all platforms unless the `fs.readdir` results are sorted.

scripts/release-build.test.js
```diff
@@ -0,0 +1,15 @@
+'use strict';
+const assert = require('assert');
+const { execSync } = require('child_process');
+
+const prebuiltPath = '../build';
+
+describe('release-build', () => {
+  it('pre-built bundles are identical to the source', done => {
```
foolip, line 8:
```suggestion
  it('pre-built bundles are identical to the source', () => {
```

This test is fully synchronous. That means you actually won't need the timeout, because it's impossible for mocha to run and deem the test timed out. It can mark it as slow in the output, though.

For tests that need to be async, mocha supports async functions.

scripts/release-build.test.js
```diff
@@ -0,0 +1,15 @@
...
+    execSync('npm run release-build');
+    const regular = require('../');
```
foolip, line 10:
```suggestion
    const regular = require('..');
```

foolip (approved):
Looking good! Leaving final review for @ddbeck though.

scripts/release-build.test.js
```diff
@@ -0,0 +1,14 @@
...
+
+const prebuiltPath = '../build';
+
+describe('release-build', () => {
+  it('pre-built bundles are identical to the source', () => {
+    execSync('npm run release-build');
+    const regular = require('..');
+    const bundled = require(prebuiltPath);
+    assert.deepEqual(regular, bundled);
+  }).timeout(5000); // Timeout must be long enough for all the file I/O
```
foolip, line 13: Is the timeout needed, or does it only affect the color of the test in the output now?

bershanskiy: Yes, if a test exceeds the limit (in this case 2s), Mocha will fail it.

bershanskiy: I think so.
If default timeout is exceeded (2s), then the following message will be printed:
>   1) release-build
>       pre-built bundles are identical to the source:
>     Error: Timeout of 2000ms exceeded. For async tests and hooks, ensure "done()"
 is called; if returning a Promise, ensure it resolves. (C:\browser-compat-data\
scripts\release-build.test.js)
>      at processImmediate (node:internal/timers:464:21)

On "Maintenance LTS" (Node 10) configuration this test can easily exceed the limit. See this run from the most recent build on this branch:
https://github.com/mdn/browser-compat-data/runs/2351499801

scripts/release-build.js
ddbeck: > I'll confess: it was my secret plan all along :)

😄 Yeah, there's a ton of hidden benefits to this. For example, we could add internal-only fields to the schema that might make working with the data more ergonomic, but filter it out before building the final data for consumers.

> Does BCD expose any utilities now?

No, not yet. It's something we've maybe inched toward (see https://github.com/mdn/browser-compat-data/pull/9441), but have been sort of cautious about it. I rather like the idea of decoupling the data from such utilities anyway and this PR would sort of force that decision.

scripts/release-build.test.js
foolip: Hmm, I don't really understand what it could mean for a sync test to timeout, but it doesn't matter, this works :)

ddbeck:
OK, I've read back through the thread and I have one question remaining.

If I install this package from source (i.e., `npm install git@github.com:bershanskiy/browser-compat-data.git#bundle`), I get the development package (with `extend` as a dependency), instead of what we'll actually ship to npm.

(Compare this to https://github.com/mdn/browser-compat-data/pull/7513, where running `npm install git@github.com:bershanskiy/browser-compat-data.git#bundle` gets you the `dist`-style package, without any dependencies.)

Is that weird or surprising? This is an honest question. I'm legitimately unsure how significant that is.

scripts/release-build.test.js
ddbeck: I think it's Mocha that's timing out, not the test. It's a bit poorly documented, but my hunch here is that Mocha is trying to fail on slow tests (for example, to fail on infinite loops), rather than running endlessly.

foolip:
@ddbeck I'm no expert maintainer of npm packages, but a "installing from source should equal dist" rule doesn't seem like anyone can safely assume. Take https://github.com/w3c/webref which is the source of two packages, there isn't any URL at all one could install to get the equivalent of the `@webref/idl` package there.

So this seems fine to me.

Using lifecycle scripts might still be a good idea, however. The tradeoff is that any build step will require contributors to run the build script manually after any change, or risk being confused when their changes don't seem to work. Lots of packages work like that, so it's not unworkable, but does seem avoidable in this case.

ddbeck:
> @ddbeck I'm no expert maintainer of npm packages, but a "installing from source should equal dist" rule doesn't seem like anyone can safely assume. Take https://github.com/w3c/webref which is the source of two packages, there isn't any URL at all one could install to get the equivalent of the `@webref/idl` package there.

@foolip having an existing, real-world example to consider is really helpful. This works for me. Thank you!

ddbeck (approved):
OK, I'm going to merge this now. Outstanding and thank you for putting this together and for your patience, @bershanskiy! And thank you @foolip, for giving this a really excellent review too.

I've written a rather lengthy release note for this PR in https://github.com/mdn/browser-compat-data/pull/9865. If either of you has a chance to take a look at that before today's release, that would be especially welcome (though strictly optional).

PR #7402, opened 2020-11-18
Correct Safari data for HTML Element APIs

vinyldarkscratch:
This PR uses Safari (4-14) and Safari iOS (3-14) results from the mdn-bcd-collector project to correct the data for the HTML element APIs.  All results were double-checked for accuracy.

### DIFF
--- a/api/HTMLAnchorElement.json
+++ b/api/HTMLAnchorElement.json
@@ -268,10 +268,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": false
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": false
             },
             "samsunginternet_android": {
               "version_added": true
--- a/api/HTMLAreaElement.json
+++ b/api/HTMLAreaElement.json
@@ -220,10 +220,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": false
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": false
             },
             "samsunginternet_android": {
               "version_added": true
@@ -268,10 +268,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": false
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": false
             },
             "samsunginternet_android": {
               "version_added": true
@@ -604,10 +604,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": false
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": false
             },
             "samsunginternet_android": {
               "version_added": true
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -1981,10 +1981,10 @@
               "version_added": "45"
             },
             "safari": {
-              "version_added": false
+              "version_added": "10"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "10"
             },
             "samsunginternet_android": {
               "version_added": "8.0"
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -400,7 +400,7 @@
               "version_added": false
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "10.3"
             },
             "samsunginternet_android": {
               "version_added": "9.0"
@@ -703,7 +703,7 @@
               "version_added": "5.1"
             },
             "safari_ios": {
-              "version_added": "5.1"
+              "version_added": "5"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
@@ -799,7 +799,7 @@
               "version_added": "5"
             },
             "safari_ios": {
-              "version_added": "5"
+              "version_added": "4"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
@@ -1154,7 +1154,7 @@
               "version_added": "5.1"
             },
             "safari_ios": {
-              "version_added": "6"
+              "version_added": "5"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
@@ -1299,7 +1299,7 @@
               "version_added": "3"
             },
             "safari_ios": {
-              "version_added": "4"
+              "version_added": "≤3"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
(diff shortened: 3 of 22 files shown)


api/HTMLAnchorElement.json
```diff
@@ -268,10 +268,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": false
```
sideshowbarker, line 271: So I realize this patch is scoped to just updating the Safari data — but as far as I can see, all the data we have for this feature is wrong. There’s no `HTMLAnchorElement.prototype.media` in Blink or Gecko either — which is unsurprising since https://html.spec.whatwg.org/multipage/text-level-semantics.html#htmlanchorelement doesn’t actually define a `media` member for `HTMLAnchorElement`, not even in as an obsolete member in https://html.spec.whatwg.org/multipage/obsolete.html#HTMLAnchorElement-partial. And there is actually no https://developer.mozilla.org/docs/Web/API/HTMLAnchorElement/media article.

So it seems to me we should just remove this entire feature.

api/HTMLInputElement.json
```diff
@@ -318,10 +318,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": true
+              "version_added": false
```
sideshowbarker, line 321: Safari does in fact have `HTMLInputElement.prototype.formEnctype`. I guess the problem here is that we have a typo in our data: we have `formEncType` — capital _T_ — but the actual member name is `formEnctype` — lowercase _t_.

api/HTMLAreaElement.json
```diff
@@ -220,10 +220,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": false
```
sideshowbarker, line 223: There’s no `HTMLAreaElement.prototype.hreflang` in Blink or Gecko either — and https://html.spec.whatwg.org/multipage/image-maps.html#htmlareaelement doesn’t actually define a `hreflang` member for `HTMLAreaElement`, not even in as an obsolete member in https://html.spec.whatwg.org/multipage/obsolete.html#HTMLAreaElement-partiall. And there is actually no https://developer.mozilla.org/en-US/docs/Web/API/HTMLAreaElement/hreflang article.

So it seems to me we should just remove this entire feature.

sideshowbarker (commented):
This patch seems to have helped surface some instances of some ghost features in our data that should just be completely removed. I now feel like we should be looking into how we could somehow automate identifying other ghost features like this ones this patch has exposed. It seems like y’all might have the means to do that with mdn-bcd-collector stuff?

api/HTMLInputElement.json
foolip: Good catch! We were generating tests for both:
http://mdn-bcd-collector.appspot.com/tests/api/HTMLInputElement/formEnctype
http://mdn-bcd-collector.appspot.com/tests/api/HTMLInputElement/formEncType

I've sent https://github.com/foolip/mdn-bcd-collector/pull/807 to remove the custom IDL for the miscapitalized version.

@vinyldarkscratch can you (separately) rename this entry?

api/HTMLAnchorElement.json
foolip: Agree that this should be removed. I think it's worth applying this change first, however, since (eventually) the feature would then show up in a listing of features not supported anywhere. Per our data it's still in Edge and Firefox, and we should check that before removal.

@vinyldarkscratch can you split this out into a separate PR?

api/HTMLAreaElement.json
foolip: The data still says Edge and Firefox support this. Not sure if that's right, though.

api/HTMLElement.json
```diff
@@ -1981,10 +1981,10 @@
               "version_added": "45"
             },
             "safari": {
-              "version_added": "10"
+              "version_added": false
```
foolip, line 1984: @vinyldarkscratch can you back this out? It's `nonce`, updated in https://github.com/mdn/browser-compat-data/pull/7303.

api/HTMLImageElement.json
```diff
@@ -822,10 +822,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": false
```
foolip, line 825: This is also a capitalization issue. I've removed the custom IDL for this as well in https://github.com/foolip/mdn-bcd-collector/pull/807.

@vinyldarkscratch can you rename this entry too?

api/HTMLMarqueeElement.json
```diff
@@ -124,10 +124,10 @@
               "version_added": "10.1"
             },
             "safari": {
-              "version_added": "3"
+              "version_added": false
```
foolip, line 127: Ah, another miscapitalization. This is `bgcolor`, should be `bgColor`.

@vinyldarkscratch can you list all entries which are identical except for capitalization, and add this and any additional ones which are actually a typo in BCD to https://github.com/foolip/mdn-bcd-collector/pull/807?

api/HTMLMarqueeElement.json
```diff
@@ -359,10 +359,10 @@
...
```
foolip, line 362: `scrollamount` vs. `scrollAmount` issue here

api/HTMLMarqueeElement.json
```diff
@@ -406,10 +406,10 @@
...
```
foolip, line 409: `scrolldelay` vs. `scrollDelay` here

foolip (commented):
I've reviewed the cases where claimed support is being reduced only so far.

api/HTMLAnchorElement.json
vinyldarkscratch: Just to confirm: by splitting out, are you talking about this change, or removing the feature?

foolip: I mean doing the removal separately. Setting this to false here is fine.

vinyldarkscratch: Gotcha, thanks for the clarification!  Definitely, I'll get a PR for that.

foolip:
I've reviewed again. The changes to api.HTMLImageElement.lowSrc and api.HTMLInputElement.formEncType were not correctly reverted, resulting in a null&rarr;true change remaining. I'll set them back to null and merge this PR.

foolip (approved):

PR #7411, opened 2020-11-19
Remove IDBVersionChangeRequest as irrelevant

ddbeck:
This feature was [removed from WebKit (Chrome and Safari) over six years ago](https://bugs.webkit.org/show_bug.cgi?id=107711) and I could find no evidence that this was implemented elsewhere.

A follow up to https://github.com/mdn/browser-compat-data/pull/6934#issuecomment-730318051.

Pages to remove from MDN before merging:

- [x] https://wiki.developer.mozilla.org/en-US/docs/Web/API/IDBVersionChangeRequest (actually the whole tree from here should do the trick)
- [x] https://wiki.developer.mozilla.org/en-US/docs/Web/API/IDBVersionChangeRequest/setVersion


chrisdavidmills (approved):
LGTM. r+

chrisdavidmills:
Pages deleted, as requested.

PR #7416, opened 2020-11-19
Update clear-site-data.json

jakub-g:
- IE11 was released 2013; the feature is from 2016 hence no support
- Safari bug: https://bugs.webkit.org/show_bug.cgi?id=203215


sideshowbarker (approved):

sideshowbarker:
👍 Great to see some null values getting mutated into real data

PR #7426, opened 2020-11-20
Add Safari real values for Media APIs

vinyldarkscratch:
This PR adds real values for Safari for various Media APIs based upon results from the mdn-bcd-collector project.

### DIFF
--- a/api/MediaStreamTrack.json
+++ b/api/MediaStreamTrack.json
@@ -172,10 +172,10 @@
               "version_added": "44"
             },
             "safari": {
-              "version_added": null
+              "version_added": false
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": false
             },
             "samsunginternet_android": {
               "version_added": "8.0"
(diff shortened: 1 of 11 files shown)


api/MediaStreamTrack.json
```diff
@@ -849,10 +849,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": false
```
sideshowbarker, line 852: Ghost feature?

api/MediaStreamTrack.json
```diff
@@ -1041,10 +1041,10 @@
...
```
sideshowbarker, line 1044: Ghost feature?

sideshowbarker (commented):
ghost features…

api/MediaStreamTrack.json
vinyldarkscratch: There does seem to be a lot of these ghost features that we're finding...  I know that @saschanaz has been doing a little project to narrow down and remove these features without support!

foolip: This one's in https://w3c.github.io/webrtc-identity/identity.html#isolated-track, but I guess not implemented.

api/MediaStreamTrack.json
foolip: I think this was in Chrome at some point, but that's not reflected in the data.

foolip (approved):

PR #7427, opened 2020-11-20
Correct Safari data for various Media APIs

vinyldarkscratch:
This PR corrects the Safari data for various Media APIs using results from the mdn-bcd-collector project.

### DIFF
--- a/api/MediaElementAudioSourceNode.json
+++ b/api/MediaElementAudioSourceNode.json
@@ -32,7 +32,7 @@
             "version_added": "6"
           },
           "safari_ios": {
-            "version_added": true
+            "version_added": "6"
           },
           "samsunginternet_android": {
             "version_added": "1.0"
@@ -79,10 +79,10 @@
               "version_added": "42"
             },
             "safari": {
-              "version_added": null
+              "version_added": false
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": false
             },
             "samsunginternet_android": {
               "version_added": "6.0",
@@ -129,10 +129,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": false
+              "version_added": "6"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "6"
             },
             "samsunginternet_android": {
               "version_added": true
--- a/api/MediaElementAudioSourceNode.json
+++ b/api/MediaElementAudioSourceNode.json
@@ -129,10 +129,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": false
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": false
             },
             "samsunginternet_android": {
               "version_added": true
--- a/api/MediaSource.json
+++ b/api/MediaSource.json
@@ -315,7 +315,7 @@
               "version_added": "46"
             },
             "safari": {
-              "version_added": false
+              "version_added": "10"
             },
             "safari_ios": {
               "version_added": false
@@ -501,7 +501,7 @@
               "version_added": "14"
             },
             "safari": {
-              "version_added": "8"
+              "version_added": false
             },
             "safari_ios": {
               "version_added": false
@@ -563,7 +563,7 @@
               "version_added": "14"
             },
             "safari": {
-              "version_added": "8"
+              "version_added": "10.1"
             },
             "safari_ios": {
               "version_added": false
@@ -625,7 +625,7 @@
               "version_added": "14"
             },
             "safari": {
-              "version_added": "8"
+              "version_added": "10.1"
             },
             "safari_ios": {
               "version_added": false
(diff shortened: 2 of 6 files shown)


api/MediaStreamTrack.json
```diff
@@ -560,10 +560,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": false
```
sideshowbarker, line 563: I think the entire `isolated` feature here should be removed from your data — because as far as I can from https://w3c.github.io/mediacapture-main/ the MediaStreamTrack interface has no `isolated` member.

api/MediaStreamTrack.json
```diff
@@ -1041,10 +1041,10 @@
...
```
sideshowbarker, line 1044: I think we should remove the `readonly` feature entirely. It’s not part of the spec.

sideshowbarker (commented):
We seem to have some ghost features here that should be removed I think

api/MediaStreamTrack.json
foolip: This is from https://w3c.github.io/webrtc-identity/identity.html#isolated-track.

api/MediaStreamTrack.json
foolip: I think https://chromium.googlesource.com/chromium/src/+/bcc6621834abbd8f95c4d36e18d691ac495b0aa1 explains why this is here. It was never web-exposed in Chrome, but it was an internal concept. It would make sense to remove together with the `remote` entry, I think.

api/MediaElementAudioSourceNode.json
```diff
@@ -129,10 +129,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": false
```
foolip, line 132: I'm fairly sure this is wrong. https://mdn-bcd-collector.appspot.com/tests/api/MediaElementAudioSourceNode in Safari 14 shows an exception being thrown in a way that only makes sense if the property is supported, and the result of the test is null. Is the update code treated null as false? Is so, changes like this one are very likely. It would be great to handle such cases separately.

One can also tell from https://github.com/WebKit/webkit/blob/master/Source/WebCore/Modules/webaudio/MediaElementAudioSourceNode.idl that `mediaElement` has been in that IDL file for a very long time.

api/MediaSource.json
```diff
@@ -501,7 +501,7 @@
               "version_added": "14"
             },
             "safari": {
-              "version_added": "8"
+              "version_added": false
```
foolip, line 504: http://mdn-bcd-collector.appspot.com/tests/api/MediaSource/isTypeSupported is testing the wrong thing here, the original data is probably correct. unittest/unit/build.js even has unit tests for this API specifically, but still an instance is being tested. @vinyldarkscratch can you look into this?

api/MediaStreamAudioSourceNode.json
```diff
@@ -29,10 +29,10 @@
             "version_added": "14"
           },
           "safari": {
-            "version_added": "6"
+            "version_added": false
```
foolip, line 32: I suspect this change is incorrect and `MediaStreamAudioSourceNode` is supported. http://mdn-bcd-collector.appspot.com/tests/api/MediaStreamAudioSourceNode return false in many places making it hard to interpret the results.

@vinyldarkscratch can you check out why we're seeing a false result here?

api/MediaStreamTrack.json
```diff
@@ -29,10 +29,10 @@
             "version_added": "16"
           },
           "safari": {
-            "version_added": "11"
+            "version_added": false
```
foolip, line 32: http://mdn-bcd-collector.appspot.com/tests/api/MediaStreamTrack is similarly complex and I don't think these changes are correct.

foolip (changes requested):
There are a fair number of apparent test issues here. It might be a good idea to split out the changes that are added claimed support from those that are removing claimed support, since the latter requires much more scrutiny and will hold up the former.

api/MediaElementAudioSourceNode.json
vinyldarkscratch: Hmm, some version must have returned `false` in the results, as we've made sure that `null` is just that, and an all `null` feature will not be set as `false`.  I'll look into this some more; for now I set it manually by going to https://mdn-bcd-collector.appspot.com/tests/api/MediaElementAudioSourceNode/mediaElement in Safari 6.

api/MediaStreamTrack.json
vinyldarkscratch: Looks like there was an issue with the test in regards to importing promises that caused this.  I'm deploying a fix to the collector and have reverted changes.

api/MediaStreamAudioSourceNode.json
foolip: @vinyldarkscratch this conversation was resolved but it looks like the change still remains, so reopening.

foolip:
@vinyldarkscratch should any of the changes in api/MediaStreamAudioSourceNode.json remain, or should they be reverted?

foolip (approved):

PR #7434, opened 2020-11-20
Add CSSFontFaceRule and CSSImportRule APIs

vinyldarkscratch:
This PR adds BCD data for two CSS rule APIs: the font face rule, and the import rule.  Versions were determined using the mdn-bcd-collector project, along with a slight bit of mirroring (Firefox -> Fenix, Chrome -> CrA, Opera, Opera Android, Samsung Internet, WebView).

These two interfaces are a part of DOM Level 2:
https://www.w3.org/TR/DOM-Level-2-Style/css.html#CSS-CSSFontFaceRule
https://www.w3.org/TR/DOM-Level-2-Style/css.html#CSS-CSSImportRule

### DIFF
--- a/api/CSSFontFaceRule.json
+++ b/api/CSSFontFaceRule.json
@@ -5,40 +5,40 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/CSSFontFaceRule",
         "support": {
           "chrome": {
-            "version_added": false
+            "version_added": "1"
           },
           "chrome_android": {
-            "version_added": false
+            "version_added": "18"
           },
           "edge": {
-            "version_added": false
+            "version_added": "12"
           },
           "firefox": {
-            "version_added": false
+            "version_added": "3.5"
           },
           "firefox_android": {
-            "version_added": false
+            "version_added": "4"
           },
           "ie": {
-            "version_added": false
+            "version_added": "9"
           },
           "opera": {
-            "version_added": false
+            "version_added": "≤15"
           },
           "opera_android": {
-            "version_added": false
+            "version_added": "≤14"
           },
           "safari": {
-            "version_added": false
+            "version_added": "≤4"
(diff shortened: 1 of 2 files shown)


api/CSSFontFaceRule.json
```diff
@@ -0,0 +1,100 @@
+{
+  "api": {
+    "CSSFontFaceRule": {
+      "__compat": {
+        "mdn_url": "https://developer.mozilla.org/docs/Web/API/CSSFontFaceRule",
```
foolip, line 5: The current spec for this is https://drafts.csswg.org/css-fonts/#om-fontface, and it matches this, just the interface with a `style` property.

api/CSSImportRule.json
```diff
@@ -0,0 +1,196 @@
+{
+  "api": {
+    "CSSImportRule": {
+      "__compat": {
+        "mdn_url": "https://developer.mozilla.org/docs/Web/API/CSSImportRule",
```
foolip, line 5: This is spec'd in https://drafts.csswg.org/cssom/#the-cssimportrule-interface, and that matches the subfeatures being added here.

foolip (approved):

foolip:
@vinyldarkscratch in order to figure out which spec something is in, I would suggest not looking at https://www.w3.org/TR/, but instead search in https://github.com/w3c/webref/tree/master/ed/idl. The most up-to-date version of the spec is linked at the top of each file there.

@ddbeck is there anything special one should do when adding an entry?

ddbeck:
@foolip 

> is there anything special one should do when adding an entry?

No, unlike removals, I haven't been doing anything special for new entries. Let me know if you think I _should_ though. 😆

foolip:
@ddbeck alright, in that case I will just merge away without the release notes label :)

PR #7436, opened 2020-11-21
Add PerformanceElementTiming API

vinyldarkscratch:
This PR adds data for the PerformanceElementTiming API. Versions were determined using the mdn-bcd-collector project, along with a slight bit of manual testing in Chrome Android, WebView, and Opera.

These are a part of a WICG draft:
https://wicg.github.io/element-timing/#sec-performance-element-timing

### DIFF
--- a/api/PerformanceElementTiming.json
+++ b/api/PerformanceElementTiming.json
@@ -1,0 +1,532 @@
+{
+  "api": {
+    "PerformanceElementTiming": {
+      "__compat": {
+        "mdn_url": "https://developer.mozilla.org/docs/Web/API/PerformanceElementTiming",
+        "support": {
+          "chrome": {
+            "version_added": false
+          },
+          "chrome_android": {
+            "version_added": false
+          },
+          "edge": {
+            "version_added": false
+          },
+          "firefox": {
+            "version_added": false
+          },
+          "firefox_android": {
+            "version_added": false
+          },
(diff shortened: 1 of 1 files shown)


api/PerformanceElementTiming.json
```diff
@@ -0,0 +1,532 @@
...
+          "samsunginternet_android": {
+            "version_added": "12.0"
+          },
+          "webview_android": {
+            "version_added": "77"
+          }
+        },
+        "status": {
+          "experimental": false,
+          "standard_track": true,
```
foolip, line 46: https://wicg.github.io/element-timing/ has "unofficial draft" written so large it's hard to read the spec. Whether WICG should get the `standards_track` status is the subject of https://github.com/mdn/browser-compat-data/issues/6103, but in this case I think the spec itself claims to be so unofficial that we should probably set it to false.

foolip (commented):
Looks good apart from `standard_track` status.

jpmedley (approved):

api/PerformanceElementTiming.json
foolip: I've gone ahead and updated the statuses in https://github.com/mdn/browser-compat-data/pull/7436/commits/2ed81ae2d0ad9fd3a993f7d1e2264b4d857c6ca8.

PR #7437, opened 2020-11-21
Add Safari data for a few WebGL APIs

vinyldarkscratch:
This PR adds real values for Safari for a few WebGL APIs based upon results from the mdn-bcd-collector project.

### DIFF
--- a/api/WEBGL_compressed_texture_etc.json
+++ b/api/WEBGL_compressed_texture_etc.json
@@ -29,10 +29,10 @@
             "version_added": true
           },
           "safari": {
-            "version_added": null
+            "version_added": false
           },
           "safari_ios": {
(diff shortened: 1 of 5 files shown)


api/WEBGL_compressed_texture_etc.json
```diff
@@ -29,10 +29,10 @@
             "version_added": true
           },
           "safari": {
-            "version_added": null
+            "version_added": false
```
foolip, line 32: This is an unusual situation, I guess, but http://mdn-bcd-collector.appspot.com/tests/api/WEBGL_compressed_texture_etc does show false on Safari 14 on a physical MBP.

foolip (approved):

PR #7439, opened 2020-11-21
Add Safari data for DOMStringList/Map API

vinyldarkscratch:
This PR adds Safari data for the DOMStringList and DOMStringMap APIs based upon results from the mdn-bcd-collector project.

### DIFF
--- a/api/DOMStringList.json
+++ b/api/DOMStringList.json
@@ -172,10 +172,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": "10"
+              "version_added": "5.1"
             },
             "safari_ios": {
-              "version_added": "10"
+              "version_added": "5"
             },
             "samsunginternet_android": {
               "version_added": true
--- a/api/DOMStringList.json
+++ b/api/DOMStringList.json
@@ -29,10 +29,10 @@
             "version_added": true
           },
           "safari": {
-            "version_added": true
+            "version_added": "5.1"
           },
           "safari_ios": {
-            "version_added": true
+            "version_added": "5"
           },
           "samsunginternet_android": {
             "version_added": true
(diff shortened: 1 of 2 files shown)


api/DOMStringList.json
```diff
@@ -172,10 +172,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "10"
```
foolip, line 175: I'm pretty sure this is a problem with http://mdn-bcd-collector.appspot.com/tests/api/DOMStringList. The `length` property was probably just on the instance before this. Can you re-test with an instance?

api/DOMStringList.json
```diff
@@ -29,10 +29,10 @@
             "version_added": true
           },
           "safari": {
-            "version_added": true
+            "version_added": "5.1"
```
foolip, line 32: The only place in WebKit where `DOMStringList` is used is `location.ancestorOrigins`, but that says it was added in Safari 6. Can you check if that data also needs updating? It would be quite strange (but not impossible) if they weren't shipped together.

vinyldarkscratch: Interestingly enough, `DOMStringList` was exposed in Safari 5.1, but `location.ancestorOrigins` was introduced in Safari 6.

api/DOMStringList.json
vinyldarkscratch: I'll have to look into how to obtain an instance for this interface -- for now, I've manually updated this based upon the [IDL in Safari 5.1.7](https://trac.webkit.org/browser/webkit/releases/Apple/Safari%205.1.7/WebCore/dom/DOMStringList.idl).

api/DOMStringList.json
foolip: That's weird, but not impossible. Getting the compile- or runtime-flags correct in every case isn't easy, stuff like this happens. (Or `DOMStringList` was used for something else I don't know about in Safari 5.1.)

foolip (approved):

PR #7440, opened 2020-11-21
Update Safari data for Document API

vinyldarkscratch:
This PR updates the Safari and Safari iOS data for various features of the Document API based upon results from the mdn-bcd-collector project.

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -1092,7 +1092,7 @@
               "version_added": "5"
             },
             "safari_ios": {
-              "version_added": "5"
+              "version_added": "4"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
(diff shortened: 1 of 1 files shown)


api/Document.json
```diff
@@ -7940,7 +7940,7 @@
               "version_added": "10.1"
             },
             "safari_ios": {
-              "version_added": "10.3"
+              "version_added": false
```
foolip, line 7943: This does not seem plausible, that the `onpointerlockchange` event handler attribute should be supported in Safari but not Safari iOS, while `element.requestPointerLock()` is claimed to be supported in both.

What were the tests results for each here?

vinyldarkscratch: It appears that `api.Element.requestPointerLock` is not supported in Safari iOS, which lines up with this data.

foolip: Great, thanks!

foolip (approved):

PR #7442, opened 2020-11-21
Add IE data for CanvasRenderingContext2D API

vinyldarkscratch:
This PR adds Internet Explorer data for the CanvasRenderingContext2D API based upon results from the mdn-bcd-collector project.

### DIFF
--- a/api/CanvasRenderingContext2D.json
+++ b/api/CanvasRenderingContext2D.json
@@ -537,7 +537,7 @@
               "version_added": "4"
             },
             "ie": {
-              "version_added": true
+              "version_added": "9"
             },
             "opera": {
               "version_added": true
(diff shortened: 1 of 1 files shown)


api/CanvasRenderingContext2D.json
```diff
@@ -2445,7 +2445,7 @@
               "version_added": false
             },
             "ie": {
-              "version_added": null
+              "version_added": false
```
foolip, line 2448: This is setting `imageSmoothingQuality` to false. I find no indication that `msImageSmoothingQuality` was ever a thing, so this seems right.

api/CanvasRenderingContext2D.json
```diff
@@ -3269,7 +3269,7 @@
               ]
             },
             "ie": {
-              "version_added": true
+              "version_added": false
```
foolip, line 3272: This now matches the IE data for `addHitRegion`, good!

foolip (approved):
I looked most closely at the cases that weren't setting the version to "9".

PR #7444, opened 2020-11-21
Update IE data for WebGL APIs

vinyldarkscratch:
This PR updates the Internet Explorer data for various WebGL APIs based upon results from the mdn-bcd-collector project.

### DIFF
--- a/api/WebGLObject.json
+++ b/api/WebGLObject.json
@@ -20,7 +20,7 @@
             "version_added": null
           },
           "ie": {
-            "version_added": null
+            "version_added": "11"
           },
           "opera": {
             "version_added": null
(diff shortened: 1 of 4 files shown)


api/WebGLObject.json
```diff
@@ -20,7 +20,7 @@
             "version_added": null
           },
           "ie": {
-            "version_added": null
+            "version_added": "11"
```
foolip, line 23: Interesting. I guess it was only IE/Edge that had this base interface for `WebGLBuffer` and other interfaces. The Edge data will also need updating, since I assume it's not in new Edge.

foolip (approved):

PR #7447, opened 2020-11-21
Update IE data for FileList/FileReaderSync API

vinyldarkscratch:
This PR updates the Internet Explorer data for the FileList and FileReaderSync APIs based upon data from the mdn-bcd-collector project.

### DIFF
--- a/api/FileList.json
+++ b/api/FileList.json
@@ -67,7 +67,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": null
+              "version_added": "10"
             },
             "opera": {
               "version_added": true
@@ -115,7 +115,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": null
+              "version_added": "10"
             },
             "opera": {
               "version_added": true
--- a/api/FileReaderSync.json
+++ b/api/FileReaderSync.json
@@ -20,7 +20,7 @@
             "version_added": "8"
           },
           "ie": {
-            "version_added": false
+            "version_added": true
           },
           "opera": {
             "version_added": true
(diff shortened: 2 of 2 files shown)


api/FileList.json
```diff
@@ -67,7 +67,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": null
+              "version_added": "10"
```
foolip, line 70: Seems very plausible, the parent entry is also "10".

api/FileReaderSync.json
```diff
@@ -20,7 +20,7 @@
             "version_added": "8"
           },
           "ie": {
-            "version_added": true
+            "version_added": false
```
foolip, line 23: Per https://caniuse.com/filereadersync this was supported in IE10, and http://mdn-bcd-collector.appspot.com/tests/api/FileReaderSync is only checking for the interface object so false negatives are somewhat likely.

Can you check where the original data came from, and whether this was somehow supported in IE? Maybe the API shape was slightly different, or it was prefixed?

vinyldarkscratch: It looks like there's an issue with getting exposure for `DedicatedWorker`, so because it was only testing in shared workers, which is not supported in IE, it was marking this as `false`.  Reverting.

foolip: I see. I've filed https://github.com/foolip/mdn-bcd-collector/issues/815 to track this.

api/FileReaderSync.json
```diff
@@ -325,7 +325,7 @@
               "version_added": "8"
             },
             "ie": {
-              "version_added": true
+              "version_added": false
```
foolip, line 328: @vinyldarkscratch did you mean to leave this change? (This is the `worker_support` entry.)

vinyldarkscratch: I'm wondering why my git client's "revert" isn't working fully, hmm...

foolip (approved):

PR #7451, opened 2020-11-22
Update IE data for HTML element APIs

vinyldarkscratch:
This PR updates the Internet Explorer data for various HTML element APIs based upon results from the mdn-bcd-collector project.  Results were manually verified for confirmation.

### DIFF
--- a/api/HTMLAudioElement.json
+++ b/api/HTMLAudioElement.json
@@ -68,7 +68,7 @@
               "version_added": "4"
             },
             "ie": {
-              "version_added": "10"
+              "version_added": "9"
             },
             "opera": {
               "version_added": true
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -2356,7 +2356,7 @@
               "version_added": null
             },
             "ie": {
-              "version_added": false
+              "version_added": null
             },
             "opera": {
               "version_added": "47"
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -641,7 +641,7 @@
               "version_added": "4"
             },
             "ie": {
-              "version_added": null
+              "version_added": false
             },
             "opera": {
               "version_added": null
@@ -838,7 +838,7 @@
               "version_added": "4"
             },
             "ie": {
-              "version_added": null
+              "version_added": false
             },
             "opera": {
               "version_added": null
@@ -2356,7 +2356,7 @@
               "version_added": null
             },
             "ie": {
-              "version_added": null
+              "version_added": false
             },
             "opera": {
               "version_added": "47"
@@ -2452,7 +2452,7 @@
               "version_added": false
             },
             "ie": {
-              "version_added": true
+              "version_added": "≤6"
             },
             "opera": {
               "version_added": true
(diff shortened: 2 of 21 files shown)


foolip:
This is updating some entries which have just been renamed or still have open PRs to rename, in https://github.com/mdn/browser-compat-data/pull/7471 through https://github.com/mdn/browser-compat-data/pull/7475.

Can you regenerate these changes to get rid of the incorrect changes due to the name mismatch?

api/HTMLMarqueeElement.json
```diff
@@ -444,7 +444,7 @@
               "version_added": "65"
             },
             "ie": {
-              "version_added": "4"
+              "version_added": false
```
foolip, line 447: Another capitalization mismatch. @vinyldarkscratch can you tweak find-missing.js a bit to find all entries which have multiple capitalization variants and ensure that only valid and known cases like `defaultstatus` + `defaultStatus` remain?

api/HTMLTrackElement.json
```diff
@@ -48,7 +48,7 @@
             }
           ],
           "ie": {
-            "version_added": "5.5"
+            "version_added": "10"
```
foolip, line 51: Yeah, `<track>` was definitely not in IE 5.5 :)

api/HTMLVideoElement.json
```diff
@@ -513,7 +513,7 @@
               "version_added": false
             },
             "ie": {
-              "version_added": "10"
+              "version_added": false
```
foolip, line 516: This is `msIsStereo3D` and came up in another review as more complicated than it seems at first. Can you find that review and see what links we dug up? We should probably add some notes here to avoid a change like this slipping through.

foolip:
@vinyldarkscratch I see that changes remain for api.HTMLImageElement.lowSrc, api.HTMLInputElement.formEncType, and other miscapitalized APIs. Can you double check that all of the remaining changes are correct?

api/HTMLElement.json
```diff
@@ -2356,7 +2356,7 @@
               "version_added": null
             },
             "ie": {
-              "version_added": null
+              "version_added": false
```
foolip, line 2359: I'm removing this entry in https://github.com/mdn/browser-compat-data/pull/7484, `onModule` isn't a thing.

api/HTMLAudioElement.json
```diff
@@ -68,7 +68,7 @@
               "version_added": "4"
             },
             "ie": {
-              "version_added": "10"
+              "version_added": "9"
```
foolip, line 71: This now matches the api.HTMLAudioElement entry itself, makes sense!

api/HTMLMenuElement.json
```diff
@@ -165,7 +165,7 @@
               "version_added": "8"
             },
             "ie": {
-              "version_added": true
+              "version_added": "≤6"
```
foolip, line 168: This change seems a little bit suspect, why is only api.HTMLMenuElement.type updated but not the parent entry or api.HTMLMenuElement.compact? Don't the tests cover those?

foolip (approved):
In the interest of making progress and not creating large merge conflicts, I'm going to back out two of the changes I have questions about, api.HTMLElement.onModule and api.HTMLMenuElement.type. @vinyldarkscratch can you check those out afterwards?

PR #7452, opened 2020-11-22
Add IE data for MediaList API

vinyldarkscratch:
This PR updates the Internet Explorer data for the MediaList API based upon results from the mdn-bcd-collector project.  Results were manually verified for confirmation.

### DIFF
--- a/api/MediaList.json
+++ b/api/MediaList.json
@@ -20,7 +20,7 @@
             "version_added": true
           },
           "ie": {
-            "version_added": null
+            "version_added": "9"
           },
           "opera": {
             "version_added": true
@@ -67,7 +67,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": null
+              "version_added": "9"
             },
             "opera": {
               "version_added": true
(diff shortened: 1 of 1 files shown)


api/MediaList.json
```diff
@@ -20,7 +20,7 @@
             "version_added": true
           },
           "ie": {
-            "version_added": null
+            "version_added": "9"
```
foolip, line 23: I suspect this is later than it was actually introduced. http://mdn-bcd-collector.appspot.com/tests/api/MediaList looks for the interface object, but that's not an important part of the API. A `MediaList` is used on https://developer.mozilla.org/en-US/docs/Web/API/CSSMediaRule which is the object for https://developer.mozilla.org/en-US/docs/Web/CSS/@media.

Can you update these as a group? If `@media` really was supported since IE6, was there no rule object for it at all until IE9, or was it just not exposed?

foolip (approved):

foolip:
Ping @vinyldarkscratch

vinyldarkscratch:
Closing this PR and will create a new one with improved results!

PR #7453, opened 2020-11-22
Update IE data for TextTrack[Cue/List] API

vinyldarkscratch:
This PR updates the Internet Explorer data for the TextTrack, TextTrackCue, and TextTrackList APIs based upon results from the mdn-bcd-collector project.  Results were manually verified for confirmation.

### DIFF
--- a/api/TextTrack.json
+++ b/api/TextTrack.json
@@ -117,7 +117,7 @@
               "version_added": "31"
             },
             "ie": {
-              "version_added": "10"
+              "version_added": "11"
             },
             "opera": {
               "version_added": "15"
@@ -260,7 +260,7 @@
               "version_added": "31"
             },
             "ie": {
-              "version_added": "10"
+              "version_added": false
             },
             "opera": {
               "version_added": "15"
@@ -308,7 +308,7 @@
               "version_added": "31"
             },
             "ie": {
-              "version_added": "10"
+              "version_added": false
             },
             "opera": {
               "version_added": "15"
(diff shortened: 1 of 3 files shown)


api/TextTrackList.json
```diff
@@ -20,7 +20,7 @@
             "version_added": null
           },
           "ie": {
-            "version_added": null
+            "version_added": "10"
```
foolip, line 23: `TextTrackList` is used for https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/textTracks which is missing the IE data. Can you update that here too, to ensure they're in sync? It would be totally possible for the `TextTrackList` interface object to be missing, so if `textTracks` is supported earlier than this, we'd need to investigate. (IE9 seems too early for that to be possible, though.)

api/TextTrack.json
```diff
@@ -117,7 +117,7 @@
               "version_added": "31"
             },
             "ie": {
-              "version_added": "10"
+              "version_added": "11"
```
foolip, line 120: http://mdn-bcd-collector.appspot.com/tests/api/TextTrack only looks at the prototype, can you do some manual testing to confirm this with an instance? It seems a little odd for `addCue` and `removeCue` to have been delayed until IE11, while `TextTrackCue` was supported in IE10.

api/TextTrackList.json
vinyldarkscratch: Apparently the page just isn't updating with the most recent information, but going to `wiki.***` shows the up-to-date BCD.

foolip: Ah, so that entry was updated in https://github.com/mdn/browser-compat-data/pull/7246, now says IE10.

foolip (approved):

PR #7455, opened 2020-11-22
Update IE data for Node[List] API

vinyldarkscratch:
This PR updates the Internet Explorer data for the Node and NodeList APIs based upon results from the mdn-bcd-collector project.  Results were manually verified for confirmation.

### DIFF
--- a/api/Node.json
+++ b/api/Node.json
@@ -171,7 +171,7 @@
               "version_added": null
             },
             "ie": {
-              "version_added": null
+              "version_added": false
             },
             "opera": {
               "version_added": null
@@ -486,7 +486,7 @@
               "version_added": "4"
             },
             "ie": {
-              "version_added": true
+              "version_added": "≤6"
             },
             "opera": {
               "version_added": true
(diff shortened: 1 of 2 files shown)


api/Node.json
```diff
@@ -171,7 +171,7 @@
               "version_added": null
             },
             "ie": {
-              "version_added": null
+              "version_added": false
```
foolip, line 174: This `baseURIObject` entry should definitely be removed. Want to do that in a separate PR?

api/NodeList.json
```diff
@@ -163,7 +163,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": null
+              "version_added": "8"
```
foolip, line 166: `NodeList` is what's used for `childNodes`, but the `api.Node.childNodes` entry says IE5 and I've confirmed it's in IE7 at least. So what kind of object did IE return for `childNodes` before? I suspect this is just a problem of interface object not being exposed, and this should be set to an earlier version.

vinyldarkscratch: I've added a custom test for this, and have updated the results manually!

foolip (approved):

PR #7459, opened 2020-11-22
Add data for SVG element API features in IE 9

vinyldarkscratch:
This PR updates the Internet Explorer data for SVG element APIs based upon results from the mdn-bcd-collector project, cherry-picking changes involving IE 9. Results were manually verified for confirmation.

### DIFF
--- a/api/SVGCircleElement.json
+++ b/api/SVGCircleElement.json
@@ -67,7 +67,7 @@
               "version_added": true
             },
             "ie": {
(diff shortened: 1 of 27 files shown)


foolip (commented):
This is a huge change but easy to review, because all of the changes are either null&rarr;9 or true&rarr;9. Then I've just reviewed the list of entries updated, and it seems to make sense together.

PR #7460, opened 2020-11-22
Mark unsupported SVG element API features in IE

vinyldarkscratch:
This PR updates the Internet Explorer data for SVG element APIs based upon results from the mdn-bcd-collector project, cherry-picking changes that indicate no support. Results were manually verified for confirmation.


foolip (approved):
These are all null&rarr;false changes, and has mostly to do with SVG fonts and a few other bits, like the more recent additions on `SVGImageElement`. It all makes sense together, to me at least.

PR #7461, opened 2020-11-22
Add data for SVG element API features in IE 10

vinyldarkscratch:
This PR updates the Internet Explorer data for SVG element APIs based upon results from the mdn-bcd-collector project, cherry-picking changes involving IE 10. Results were manually verified for confirmation.


foolip (approved):
This seems to all be about SVG filters, and https://caniuse.com/svg-filters agrees that was added in IE10. Easy to review because the previous values were all true or null. Thanks for splitting it like this!

PR #7462, opened 2020-11-22
Correct SVG element API data for IE

vinyldarkscratch:
This PR corrects the existing Internet Explorer data for SVG element APIs based upon results from the mdn-bcd-collector project. Results were manually verified for confirmation.

### DIFF
--- a/api/SVGForeignObjectElement.json
+++ b/api/SVGForeignObjectElement.json
@@ -20,7 +20,7 @@
             "version_added": true
           },
           "ie": {
-            "version_added": "9"
+            "version_added": false
           },
           "opera": {
             "version_added": "≤12.1"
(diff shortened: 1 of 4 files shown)


api/SVGForeignObjectElement.json
```diff
@@ -20,7 +20,7 @@
             "version_added": true
           },
           "ie": {
-            "version_added": "9"
+            "version_added": false
```
foolip, line 23: https://developer.mozilla.org/en-US/docs/Web/SVG/Element/foreignObject#Browser_compatibility and https://stackoverflow.com/a/19739804 also indicate `<foreignObject>` was not supported in IE.

foolip (approved):

PR #7463, opened 2020-11-22
Mark unsupported Window API features for IE

vinyldarkscratch:
This PR sets the Internet Explorer data for Window API to false for unsupported features determined based upon results from the mdn-bcd-collector project. Results were manually verified for confirmation.

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -1336,7 +1336,7 @@
               "version_added": "6"
             },
             "ie": {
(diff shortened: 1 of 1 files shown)


foolip (approved):
These are all null&rarr;false changes and the list of updated entires looks sensible, I can't see any APIs that I know to be very old and I expect to have been in IE.

PR #7465, opened 2020-11-22
Add Chromium data for CanvasRenderingContext2D

vinyldarkscratch:
This PR adds real values for Chrome and its derivatives for the CanvasRenderingContext2D API using results from the mdn-bcd-collector project, along with a tiny bit of mirroring (which was all confirmed with mdn-bcd-collector results).

### DIFF
--- a/api/CanvasRenderingContext2D.json
+++ b/api/CanvasRenderingContext2D.json
@@ -522,10 +522,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/CanvasRenderingContext2D/beginPath",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "1"
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": "18"
             },
             "edge": {
               "version_added": "12"
@@ -540,10 +540,10 @@
               "version_added": true
             },
             "opera": {
-              "version_added": true
+              "version_added": "≤12.1"
             },
             "opera_android": {
-              "version_added": true
+              "version_added": "≤12.1"
             },
             "safari": {
               "version_added": "≤4"
@@ -552,10 +552,10 @@
               "version_added": "≤3"
             },
             "samsunginternet_android": {
-              "version_added": true
+              "version_added": "1.0"
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "1"
             }
           },
           "status": {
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
Nothing surprises me here but I might be a bit rusty. Anything I need to know when reviewing mdn-bcd-collector PRs?

vinyldarkscratch:
Welcome back, @Elchi3, glad to see you again!

Basically, the mdn-bcd-collector project is a feature detection system that Philip and I have been working on, something like what Confluence does.  Overall, reviewing is the same process as other PRs!

foolip:
@Elchi3 some tips for these PRs. I use https://github.com/mdn/browser-compat-data/pull/6862 for them, since that gives a compact of view of what has changed.

The majority of issues I look for and catch are false negatives, due to a variety of reasons:
- Interface object not being exposed, as I suspect in https://github.com/mdn/browser-compat-data/pull/7452#discussion_r528711247. There are *lots* of cases like this in the history of all browsers.
- Interface object being exposed but some things being on instances, which can happen for many reasons. https://github.com/mdn/browser-compat-data/pull/7439#discussion_r528657369 is a case where I suspect this.
- Dependencies for getting to the feature test itself, where any of those failing can lead to an incorrect result. I suspect this in https://github.com/mdn/browser-compat-data/pull/7427#discussion_r528648507.
- Bugs in the generated tests, causing them to test the wrong thing, as in https://github.com/mdn/browser-compat-data/pull/7427#discussion_r528645814.

False negatives aren't only a problem when setting something to true, but are always necessary to consider unless the tests return true since the very first version of a browser, since an observed change between version N and N+1 could be a false negative for version N. However, I spend a lot less time looking for these types of errors when a set of changes together seem to make sense.

Beyond false negatives, there are also some limitations of the update scripts themselves, some of them documented in https://github.com/foolip/mdn-bcd-collector/issues/571. In all of my reviewing so far, however, I haven't spotted a case where this was the root cause for some incorrect data change.

api/CanvasRenderingContext2D.json
```diff
@@ -1581,10 +1581,10 @@
                 "version_added": null
               },
               "opera": {
-                "version_added": null
+                "version_added": "46"
```
foolip, line 1584: This is changing the "SVGImageElement_source_image" entry, but it looks like it's just doing mirroring. We don't generate tests for this entry.

foolip (approved):
http://mdn-bcd-collector.appspot.com/tests/api/CanvasRenderingContext2D tests using `document.createElement('canvas')` so false negatives are quite unlikely here.

api.CanvasRenderingContext2D.drawImage.SVGImageElement_source_image is the only behavioral entry touched here, and the rest all look sensible to me, nothing that looks especially suspect.

Issue #7467, opened 2020-11-22
CSS: update `conic-gradient` compatibility for Firefox 


monkeypox8:
`conic-gradient` support has landed in [Firefox 83](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/83), but the double-position color stops entry has not been updated. [See this page in Firefox](https://developer.mozilla.org/en-US/docs/Web/CSS/conic-gradient()#Gradient_pie-chart) for verification.

### references
- https://bugzilla.mozilla.org/show_bug.cgi?id=1632351
- https://bugzilla.mozilla.org/show_bug.cgi?id=1175958

ddbeck:
Fixed by #7468. Thank you!

PR #7468, opened 2020-11-22
Conic-gradient support update

monkeypox8:
## Description
Adds double-position entry for Firefox 83 (see #7467)


ddbeck (approved):
Thank you! 🎉

PR #7469, opened 2020-11-23
Add BCD entry for fractionalSecondDigits option in DateTimeFormat

hamishwillee:
This adds a new entry for the new option `fractionalSecondDigits` added to the `Intl.DateTimeFormat()` constructor options. 
- Added to Firefox in FF84 - https://bugzilla.mozilla.org/show_bug.cgi?id=1645107
- Added to Chromium in 84 - https://www.chromestatus.com/feature/5704965743968256
- There is not yet a version of SamsungAndroid browser that has this.
- Looks like it is in Webkit/Safarit - but still trying to find out version: https://bugs.webkit.org/show_bug.cgi?id=215840
- It is in Nodejs but need help mapping to a version https://monorail-prod.appspot.com/p/v8/issues/detail?id=9284

Docs work being done for this can be tracked in https://github.com/mdn/sprints/issues/3897#issuecomment-730870217

### DIFF
--- a/javascript/builtins/intl/DateTimeFormat.json
+++ b/javascript/builtins/intl/DateTimeFormat.json
@@ -313,6 +313,56 @@
                   "deprecated": false
                 }
               }
+            },
+            "fractionalSecondDigits": {
+              "__compat": {
+                "support": {
+                  "chrome": {
+                    "version_added": "84"
+                  },
+                  "chrome_android": {
+                    "version_added": "84"
+                  },
+                  "edge": {
+                    "version_added": "84"
+                  },
+                  "firefox": {
+                    "version_added": "84"
+                  },
+                  "firefox_android": {
+                    "version_added": "84"
+                  },
+                  "ie": {
+                    "version_added": false
+                  },
+                  "nodejs": {
+                    "version_added": null
+                  },
+                  "opera": {
+                    "version_added": "70"
+                  },
+                  "opera_android": {
+                    "version_added": "60"
+                  },
+                  "safari": {
+                    "version_added": null
+                  },
+                  "safari_ios": {
+                    "version_added": null
+                  },
+                  "samsunginternet_android": {
+                    "version_added": false
+                  },
+                  "webview_android": {
+                    "version_added": "84"
+                  }
+                },
+                "status": {
+                  "experimental": false,
+                  "standard_track": true,
+                  "deprecated": false
+                }
+              }
             }
           },
           "format": {
--- a/javascript/builtins/intl/DateTimeFormat.json
+++ b/javascript/builtins/intl/DateTimeFormat.json
@@ -163,6 +163,56 @@
                 }
               }
             },
+            "fractionalSecondDigits": {
+              "__compat": {
+                "support": {
+                  "chrome": {
+                    "version_added": "84"
+                  },
+                  "chrome_android": {
+                    "version_added": "84"
+                  },
+                  "edge": {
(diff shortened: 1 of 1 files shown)


hamishwillee:
Hi @Constellation,
Can you please help me work out what version of Safari (Webkit) this change went into? https://bugs.webkit.org/show_bug.cgi?id=215840

hamishwillee:
@ddbeck Almost done, but would appreciate advice on how to map https://monorail-prod.appspot.com/p/v8/issues/detail?id=9284 to a particular version. This went in apparently in "m84" but I don't know how that maps to v8 releases numbers.

ddbeck:
> PS Note this fails because of the nulls for safari.

I tested this (using the test code from the WebKit patch[ in the linked bug](https://bugs.webkit.org/show_bug.cgi?id=215840)) in Safari 14 and it didn't work. Since we don't have any version later than that, so you can set the Safaris to to `false`.

> This went in apparently in "m84" but I don't know how that maps to v8 releases numbers.

Tracing Chrome to Node.js is a little bit fun. A bunch of details here, but you can jump to end for a wholly different approach.

A few resources that can help here:

* [OmahaProxy](https://omahaproxy.appspot.com/) can tell you recent Chrome releases' V8 version numbers. You can also use it to look up older releases, though there's a few steps involved (see below).
* The [V8 blog](https://v8.dev/blog) has release notes which are sometimes helpful.
* The Node.js [Previous releases](https://nodejs.org/en/download/releases/) page lists Node version numbers and their corresponding V8 versions.

To actually follow follow Chrome to V8 to Node.js precisely:

1. Wait until the Chrome release in question is beta or stable. V8 typically cuts a release when a Chrome release reaches beta, so this process might be unreliable for Canary versions.

2. Usually (always? I'm not sure), you can get a Chrome version's V8 version number by shifting the decimal place of the Chrome release one place to the left, as in Chrome 84 → V8 8.4. Often, this is good enough to jump to the last step, but follow the complete steps if you want to be certain.

3. Look up the exact version of V8 that shipped with the release of Chrome.

    For recent versions (e.g., beta and stable), [OmahaProxy](https://omahaproxy.appspot.com/) lists the V8 version in a table.

    For older versions, get a full version string. A good way to do this is to check the release notes via [`chrome.json`](https://github.com/mdn/browser-compat-data/blob/master/browsers/chrome.json#L583). In this case, I got "84.0.4147.105". In OmahaProxy's _Version Information_ widget, enter the version string and click **Lookup**. The V8 version is listed in the widget. In this case, `8.4.371.22`.

4. In the Node.js [Previous releases](https://nodejs.org/en/download/releases/) page, find the earliest release with same V8 version number. In this case, `14.6.0` (V8: `8.4.371.19`). Sometimes (as in this case) the most minor version number is different; I've not yet encountered a case where this mattered.

All that said, for Node.js, it wasn't too complex to test this directly (using [`nodenv`](https://github.com/nodenv/nodenv) to switch versions): I ran the same WebKit test code in 14.6 and 14.5. 14.6 shows the extra digits and 14.5 doesn't. 🎉

hamishwillee:
Thanks @ddbeck - that's a helpful description which I have taken a record of for myself (and verified for this case, though using nvm to switch node, as I am more familiar with it).

I don't actually have macOS so I can't test Safari (in future) unless you can suggest an alternative way of doing so?

Should be good now.

ddbeck:
Thanks, @hamishwillee!

PR #7471, opened 2020-11-23
Rename api.HTMLElement.formEncType -> formEnctype

vinyldarkscratch:
This PR fixes a capitalization issue for the HTMLElement API's `formEnctype` method.  Inspired by https://github.com/mdn/browser-compat-data/pull/7402#discussion_r527274048.

### DIFF
--- a/api/HTMLInputElement.json
+++ b/api/HTMLInputElement.json
@@ -289,9 +289,9 @@
           }
         }
       },
-      "formEncType": {
+      "formEnctype": {
         "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLInputElement/formEncType",
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLInputElement/formEnctype",
           "support": {
             "chrome": {
               "version_added": true


api/HTMLInputElement.json
```diff
@@ -289,9 +289,9 @@
           }
         }
       },
-      "formEncType": {
+      "formEnctype": {
         "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLInputElement/formEncType",
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLInputElement/formEnctype",
```
foolip, line 294: Both this and the old URL is 404. @ddbeck should we include the URLs in this case?

foolip:
@vinyldarkscratch can you update https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement as well once this is ready to merge?

vinyldarkscratch:
Certainly, can do!

api/HTMLInputElement.json
ddbeck: Dealing with 404s is a whole… thing. #5297 is open to remove them, but there's some process questions around that (how do we pick up those pages once they're created?) that are unanswered. For now, I suggest keeping the 404s.

foolip:
So after @ddbeck's guidance https://github.com/mdn/browser-compat-data/pull/7472 I think I can deal with this myself. I've updated MDN in https://wiki.developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement$compare?to=1655071&from=1641293 and since https://developer.mozilla.org/docs/Web/API/HTMLInputElement/formEnctype doesn't exist, nothing more will need to be done for the next release.

I'll go ahead and merge this now.

PR #7472, opened 2020-11-23
Rename api.HTMLImageElement.lowSrc -> lowsrc

vinyldarkscratch:
This PR corrects capitalization for the HTMLImageElement API's `lowsrc` feature.

### DIFF
--- a/api/HTMLImageElement.json
+++ b/api/HTMLImageElement.json
@@ -796,9 +796,9 @@
           }
         }
       },
-      "lowSrc": {
+      "lowsrc": {
         "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLImageElement/lowSrc",
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLImageElement/lowsrc",
           "support": {
             "chrome": {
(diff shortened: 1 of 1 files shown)


foolip:
This looks good, but I'm not sure how to do the rename dance with https://wiki.developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/lowSrc in a way that doesn't break the compat table there for longer than necessary. @ddbeck can you advise?

foolip:
I have done all the renaming in https://wiki.developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement and https://wiki.developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement/lowsrc other than the name in the BCD reference.

ddbeck:
@foolip Since I'll see this when I do the release notes anyway, I'll update the compat macro on the page at that time. Is that satisfactory? Do we need anything else to merge this?

foolip:
@ddbeck that sounds good, getting the timing right is the only thing I'm concerned about, since the worst case is that the compat table is broken for a week, and the best case is that nobody notices at all :)

I'll go ahead and merge this then.

PR #7473, opened 2020-11-23
Rename api.HTMLMarqueeElement.bgcolor -> bgColor

vinyldarkscratch:
This PR corrects capitalization for the HTMLMarqueeElement API's `bgColor` feature.

### DIFF
--- a/api/HTMLMarqueeElement.json
+++ b/api/HTMLMarqueeElement.json
@@ -96,7 +96,7 @@
           }
         }
       },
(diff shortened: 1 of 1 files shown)


foolip (approved):

foolip:
This is a simple one because https://developer.mozilla.org/en-US/docs/Web/API/HTMLMarqueeElement/bgColor doesn't exist.

PR #7476, opened 2020-11-23
Safari iOS doesn't support api.Element.requestPointerLock

vinyldarkscratch:
Based upon results from the mdn-bcd-collector project, I've determined that the `requestPointerLock` feature is actually _not_ supported on Safari iOS.

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -6935,15 +6935,9 @@
                 "prefix": "webkit"
               }
             ],
-            "safari_ios": [
-              {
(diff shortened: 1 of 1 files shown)


foolip:
Can you also update the many other related entries in Document.json, DocumentOrShadowRoot.json and GlobalEventHandlers.json?

foolip:
Scratch my previous request. Let's get this merged, there's always more that could be fixed.

foolip (approved):

PR #7477, opened 2020-11-23
Linter: Ensure that IE feature implies Edge's

### LINKED ISSUE #7458: Linter: disallow Edge 13+ as version if IE is truthy?
Since Microsoft Edge is a direct continuation from Internet Explorer, it is safe to assume that most everything present in IE will be in Edge 12.  However, I've noticed that at times, our data will have a version number besides Edge 12 when IE is truthy.  We should add a linter to make sure that if IE is truthy and Edge is truthy, Edge must be set to "12".

bershanskiy:
Fixes #7458.
Since Edge 12 inherited IE's engine, most often if IE supports something,
so does Edge 12.

- [x] Summarize your changes

### DIFF
--- a/test/lint.js
+++ b/test/lint.js
@@ -14,6 +14,7 @@ const {
   testVersions,
   testConsistency,
   testDescriptions,
+  testIEtoEdgeContinuety,
 } = require('./linter/index.js');
 const { IS_CI } = require('./utils.js');
 const testCompareFeatures = require('./test-compare-features');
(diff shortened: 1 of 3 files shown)


bershanskiy:
Please note: tests are failing because the new test finds likely false data.

bershanskiy:
I have updated the linter to handle multiple version ranges for both IE and Edge. Now, linter fails if IE has a support statement with truthly `version_added` and no indication of any prefixes, flags, etc., but Edge doesn't have one starting with `version_added: 12`.

queengooborg:
Hey @bershanskiy, thank you for your PR!  Since we're looking to deprecate and remove the data however (and since Edge is derived from Chrome now), a linter like this has become less desirable.

bershanskiy:
Thanks for the heads up @queengooborg! Please keep me posted on IE removal.

PR #7478, opened 2020-11-23
Mark ShadowRoot.prototype.innerHTML standard

sideshowbarker:
This change marks `ShadowRoot.prototype.innerHTML` as `standard_track:true` and `experimental:false.`

https://w3c.github.io/DOM-Parsing/#the-innerhtml-mixin (the `InnerHTML` interface mixin), normatively defines `ShadowRoot.prototype.innerHTML,` and all major browser engines expose `ShadowRoot.prototype.innerHTML,` so it’s not experimental.

### DIFF
--- a/api/ShadowRoot.json
+++ b/api/ShadowRoot.json
@@ -313,8 +313,8 @@
             }
           },
           "status": {
-            "experimental": true,
-            "standard_track": false,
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
Nice! :+1:

PR #7479, opened 2020-11-24
Drop notes from Safari 14 beta / TP in Locale

Constellation:
This patch drops that note since this problem does not exist in shipping Safari 14.


sideshowbarker (approved):

sideshowbarker:
👍 Good housekeeping

Issue #7480, opened 2020-11-24
Review AppCache removal in Chromium 90


hamishwillee:
Chromium 90 is planned to remove App Cache: https://www.chromestatus.com/feature/6192449487634432

- There is more information about timelines, and also for Safari/Webkit in https://web.dev/appcache-removal/
- FF support removed in FF84 (see https://github.com/mdn/browser-compat-data/pull/7481)

jculver050:
Close

queengooborg:
This was resolved in #7622!

PR #7481, opened 2020-11-24
Mark appcache (html.manifest) as removed in FF84

hamishwillee:
Appcache has been removed in FF84 as per https://bugzilla.mozilla.org/show_bug.cgi?id=1619673. This adds `version_removed` for the manifest and for the sub feature (secure contexts).  

Note:
- Other docs work for this can be tracked in https://github.com/mdn/sprints/issues/3901
- This won't disappear from [Chromium until v90](https://www.chromestatus.com/feature/6192449487634432). I created https://github.com/mdn/browser-compat-data/issues/7480 to track that. Note though "in effect" it was already removed as only sites with a certain "origin certificate" can use this feature [I have not noted that].

### DIFF
--- a/html/elements/html.json
+++ b/html/elements/html.json
@@ -63,7 +63,8 @@
               },
               "firefox": [
                 {
-                  "version_added": "3.5"
+                  "version_added": "3.5",
+                  "version_removed": "84"
                 },
                 {
                   "version_added": "3",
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Looks good. Thank you! 🎉

PR #7482, opened 2020-11-24
Deprecate HTMLFrameSetElement

sideshowbarker:
This change marks `HTMLFrameSetElement` and all its members `deprecated:true` per https://html.spec.whatwg.org/multipage/obsolete.html#htmlframesetelement. Related: https://github.com/mdn/browser-compat-data/issues/7255


Elchi3 (approved):
:+1:

PR #1260, opened 2018-03-03
adding compat data for api.HTMLFrameSetElement

RajuDasa:
[https://developer.mozilla.org/en-US/docs/Web/API/HTMLFrameSetElement](https://developer.mozilla.org/en-US/docs/Web/API/HTMLFrameSetElement)

PR #7483, opened 2020-11-24
Drop HTMLFrameSetElement.prototype.onlanguage

sideshowbarker:
This change completely removes the `HTMLFrameSetElement.prototype.onlanguage` feature from BCD. https://html.spec.whatwg.org/multipage/indices.html#attributes-3 shows there is no such attribute at all in HTML (and there never has been).  Testing in Firefox — which the current data claims got support for this feature in version 32 — shows that in fact there is no such attribute exposed in Firefox. And the data for all other browsers is null or false. So this seems like a phantom feature that just crept into our data by mistake.

### DIFF
--- a/api/HTMLFrameSetElement.json
+++ b/api/HTMLFrameSetElement.json
@@ -142,53 +142,6 @@
           }
         }
       },
-      "onlanguage": {
-        "__compat": {
-          "support": {
-            "chrome": {
-              "version_added": null
-            },
-            "chrome_android": {
-              "version_added": null
-            },
-            "edge": {
-              "version_added": null
-            },
-            "firefox": {
-              "version_added": "32"
-            },
-            "firefox_android": {
-              "version_added": "32"
-            },
-            "ie": {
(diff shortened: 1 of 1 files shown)


foolip:
This came from wiki migration in https://github.com/mdn/browser-compat-data/pull/1260. I think we can confidently say this was originally a typo of `onlanguagechange`, which does exist.

api/HTMLFrameSetElement.json
```diff
@@ -142,53 +142,6 @@
...
-              "version_added": null
-            },
-            "chrome_android": {
-              "version_added": null
-            },
-            "edge": {
-              "version_added": null
-            },
-            "firefox": {
-              "version_added": "32"
```
foolip, line 158: Just to be sure, I tested http://mdn-bcd-collector.appspot.com/tests/api/HTMLFrameSetElement/onlanguage in Firefox 32, and no, it's not there.

foolip (approved):

PR #7488, opened 2020-11-24
Fix for issue #6178

joneskj55:
This PR addresses issue #6178 by updating the version number in the String.json file as addressed in the issue.

### DIFF
--- a/javascript/builtins/String.json
+++ b/javascript/builtins/String.json
@@ -1742,7 +1742,7 @@
                 "version_added": "3.0"
               },
               "webview_android": {
-                "version_added": false
+                "version_added": "41"
               }
             },
             "status": {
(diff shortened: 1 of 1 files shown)


javascript/builtins/String.json
```diff
@@ -1794,7 +1794,7 @@
                 "version_added": "1.0"
               },
               "webview_android": {
-                "version_added": "1"
+                "version_added": "41"
```
saschanaz, line 1797: This modifies `.replace()` instead of `.repeat()`. Line 1745 should be fixed instead.

joneskj55: @saschanaz Thanks! I reverted the previous change and added another commit to this branch with the line 1745 change.

Elchi3 (approved):
Looks good to me. Thanks @joneskj55 and welcome to mdn-browser-compat-data! :tada:

joneskj55:
Thanks @Elchi3 !

PR #7489, opened 2020-11-25
Update modifier keys

MendyBerger:
- [x] Summarize your changes
Updated deprecation information on `KeyboardEvent.getModifierState()` to match the spec

https://www.w3.org/TR/uievents-key/#keys-modifier

mdn/sprints/issues/3587

### DIFF
--- a/api/KeyboardEvent.json
+++ b/api/KeyboardEvent.json
@@ -667,7 +667,7 @@
             "status": {
               "experimental": false,
               "standard_track": false,
-              "deprecated": true
+              "deprecated": false
             }
           }
         },
@@ -715,7 +715,7 @@
             "status": {
               "experimental": false,
               "standard_track": false,
-              "deprecated": true
+              "deprecated": false
             }
           }
         },
@@ -1101,7 +1101,7 @@
             "status": {
               "experimental": false,
               "standard_track": true,
-              "deprecated": false
+              "deprecated": true
             }
           }
         },
--- a/api/KeyboardEvent.json
+++ b/api/KeyboardEvent.json
@@ -666,7 +666,7 @@
             },
             "status": {
               "experimental": false,
-              "standard_track": false,
+              "standard_track": true,
               "deprecated": false
             }
           }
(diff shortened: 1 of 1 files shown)


api/KeyboardEvent.json
```diff
@@ -667,7 +667,7 @@
             "status": {
               "experimental": false,
               "standard_track": false,
```
ddbeck, line 669: suggestion: since this is specified, let's reflect that in the `standard_track` value

```suggestion
              "standard_track": false,
```

api/KeyboardEvent.json
```diff
@@ -715,7 +715,7 @@
...
```
ddbeck, line 717: suggestion: (again) since this is specified, let's reflect that in the `standard_track` value

```suggestion
              "standard_track": false,
```

api/KeyboardEvent.json
```diff
@@ -1101,7 +1101,7 @@
             "status": {
               "experimental": false,
               "standard_track": true,
```
ddbeck, line 1103: suggestion: reflect the standards status, the opposite of the others in this PR

```suggestion
              "standard_track": false,
```

ddbeck (changes requested):
Thank you for this PR, @MendyBerger. A few suggestions in line, to represent the specification situation better, but otherwise a welcome change.

api/KeyboardEvent.json
MendyBerger: Don't see any changes here

api/KeyboardEvent.json
MendyBerger: Don't see any changes here

api/KeyboardEvent.json
ddbeck: Sorry, copy and paste failure:

```suggestion
              "standard_track": true,
```

api/KeyboardEvent.json
ddbeck:
```suggestion
              "standard_track": true,
```

ddbeck (changes requested):

ddbeck (approved):
Looks great. Thank you! 🎉

PR #7493, opened 2020-11-25
Add/correct Firefox real data for DOM* APIs

vinyldarkscratch:
This PR adds real data for Firefox for various APIs prefixed with `DOM` based upon results from the mdn-bcd-collector project, along with a few corrections.

### DIFF
--- a/api/DOMParser.json
+++ b/api/DOMParser.json
@@ -65,7 +65,7 @@
               "version_added": "1"
             },
             "firefox_android": {
-              "version_added": true
+              "version_added": "4"
             },
             "ie": {
               "version_added": "9"
@@ -113,7 +113,7 @@
               "version_added": "1"
             },
             "firefox_android": {
-              "version_added": true
+              "version_added": "4"
             },
             "ie": {
               "version_added": "9"
--- a/api/DOMTokenList.json
+++ b/api/DOMTokenList.json
@@ -61,10 +61,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": true
+              "version_added": "3.6"
             },
             "firefox_android": {
-              "version_added": true
+              "version_added": "4"
             },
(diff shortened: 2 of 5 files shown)


api/DOMTokenList.json
```diff
@@ -61,10 +61,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": true
+              "version_added": "3.6"
```
foolip, line 64: The parent api.DOMTokenList isn't updated in this PR, and has Firefox versions 3/4. That's surprising given that http://mdn-bcd-collector.appspot.com/tests/api/DOMTokenList is just looking at the interface object. Was there an interface object exposed but with no members in Firefox 3, and the members were added in 3.6? I'm guessing that actually the whole thing was added in Firefox 3.6, but then can't the collector figure that out from the test results?

api/DOMParser.json
```diff
@@ -65,7 +65,7 @@
               "version_added": "1"
             },
             "firefox_android": {
-              "version_added": true
+              "version_added": "4"
```
foolip, line 68: The parent feature api.DOMParser is also set to 1/4, good.

api/DOMTokenList.json
vinyldarkscratch: Yeah, this interface was actually added in Firefox 3.6.  I had a little VM misconfiguration where my Firefox 3 VM was actually 3.6.  I've since then fixed it, but apparently I had a few PRs with this data.

I can follow-up with a corrections PR, or include the correction for this interface in this PR, whichever is more preferred?

foolip: If you could fix all of DOMTokenList in this PR that would be great. Other 3 vs. 3.6 mishaps as its own PR would be good.

foolip (approved):

PR #7495, opened 2020-11-25
Add Firefox data for FileList API

vinyldarkscratch:
This PR updates the Firefox data for the FileList API based upon results from the mdn-bcd-collector project, along with a little bit of mirroring to Firefox Android.

### DIFF
--- a/api/FileList.json
+++ b/api/FileList.json
@@ -61,10 +61,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": true
+              "version_added": "3"
             },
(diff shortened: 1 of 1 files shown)


api/FileList.json
```diff
@@ -61,10 +61,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": true
+              "version_added": "3"
```
foolip, line 64: This now matches the parent entry, makes sense!

foolip (approved):

PR #7496, opened 2020-11-25
Add real values for Firefox for HTML Element APIs

vinyldarkscratch:
This PR adds real values for Firefox for HTML element APIs using results from the mdn-bcd-collector project, including some mirroring to Firefox Android (after confirmation through the collector results).  Some mirroring includes copying notes or alt. name data.  This PR is a cherry-pick of changes that maintain truthiness.

### DIFF
--- a/api/HTMLInputElement.json
+++ b/api/HTMLInputElement.json
@@ -109,10 +109,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": true
+              "version_added": "4"
             },
             "firefox_android": {
-              "version_added": true
+              "version_added": "4"
             },
             "ie": {
               "version_added": "10"
@@ -160,7 +160,7 @@
               "version_added": "4"
             },
             "firefox_android": {
-              "version_added": true
+              "version_added": "4"
             },
             "ie": {
               "version_added": "10"
@@ -208,7 +208,7 @@
               "version_added": "3"
             },
             "firefox_android": {
-              "version_added": true
+              "version_added": "4"
             },
             "ie": {
               "version_added": "10",
@@ -258,7 +258,7 @@
               "version_added": "4"
             },
             "firefox_android": {
-              "version_added": true
+              "version_added": "4"
             },
             "ie": {
               "version_added": "10"
@@ -306,7 +306,7 @@
               "version_added": "4"
             },
             "firefox_android": {
-              "version_added": null
+              "version_added": "4"
             },
             "ie": {
               "version_added": true
@@ -354,7 +354,7 @@
               "version_added": "4"
             },
             "firefox_android": {
-              "version_added": true
+              "version_added": "4"
             },
             "ie": {
               "version_added": "10"
@@ -402,7 +402,7 @@
               "version_added": "4"
             },
             "firefox_android": {
-              "version_added": true
+              "version_added": "4"
             },
             "ie": {
               "version_added": "10"
@@ -450,7 +450,7 @@
               "version_added": "4"
             },
             "firefox_android": {
-              "version_added": true
+              "version_added": "4"
             },
             "ie": {
               "version_added": "10"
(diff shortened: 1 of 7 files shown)


foolip:
> This PR is a cherry-pick of changes that maintain truthiness.

Thanks for separating that out, much easier to review!

api/HTMLInputElement.json
```diff
@@ -643,7 +643,7 @@
               "version_added": "56"
             },
             "firefox_android": {
-              "version_added": true
+              "version_added": "56"
```
foolip, line 646: Looks like all of the `labels` entries (on multiple elements) were "56" for both Firefoxes, and this was the only outlier. Good!

api/HTMLMediaElement.json
```diff
@@ -484,7 +484,8 @@
               "notes": "Prior to Firefox 28, <code>canPlayType()</code> returned <code>probably</code> when asked about WebM audio or video files without the <code>codecs</code> parameter. Since multiple codecs are supported, this is not enough information to determine if a file can be played, so <code>maybe</code> is now correctly returned."
             },
             "firefox_android": {
-              "version_added": true
+              "version_added": "4",
+              "notes": "Prior to Firefox 28, <code>canPlayType()</code> returned <code>probably</code> when asked about WebM audio or video files without the <code>codecs</code> parameter. Since multiple codecs are supported, this is not enough information to determine if a file can be played, so <code>maybe</code> is now correctly returned."
```
foolip, line 488: This note is probably not strictly true for Android, where codec support depends on the OS version and device, not just the browser version. Researching a correct note would be a lot of work, so I would just omit it.

api/HTMLMediaElement.json
```diff
@@ -731,9 +732,16 @@
...
-              "version_added": true
-            },
+            "firefox_android": [
+              {
+                "version_added": "22"
+              },
+              {
+                "version_added": "14",
+                "version_removed": "22",
+                "alternative_name": "crossorigin"
```
foolip, line 742: Can you add this miscapitalized variant to the collector so that we can confirm the data as well eventually?

api/HTMLMediaElement.json
```diff
@@ -2972,7 +2980,8 @@
               "version_added": "20"
             },
             "firefox_android": {
-              "version_added": true
+              "prefix": "moz",
+              "version_added": "20"
```
foolip, line 2984: Coming across this entry I saw a bunch of things wrong so I've sent https://github.com/mdn/browser-compat-data/pull/7518 and https://github.com/foolip/mdn-bcd-collector/pull/831.

We haven't confirmed that the prefixed data is correct here to begin with, but let's assume that and update it if we find the desktop data to be wrong later.

api/HTMLInputElement.json
```diff
@@ -543,7 +543,7 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "3"
+              "version_added": "3.6"
```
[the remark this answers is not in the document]
vinyldarkscratch: I recently found out that my Firefox 3 VM was of version 3.6.29, rather than a 3.0.x release for some odd reason.  Since that PR, I fixed that VM, but I didn't realize I submitted any PRs with data involving that version.  In other words, any changes from 3 to 3.5/3.6 from my own PRs should be determined as the more accurate data.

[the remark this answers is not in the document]
foolip: OK, that explains it! As discussed in another PR, if you could fix 3 vs. 3.6 issues in a separate PR, that would be great!

api/HTMLMediaElement.json
foolip: Reopened, but doing this doesn't block this PR.

foolip (approved):

api/HTMLMediaElement.json
vinyldarkscratch: Tracking alt. names (and prefixes) in https://github.com/foolip/mdn-bcd-collector/issues/333, so we can verify its data!

PR #7497, opened 2020-11-25
Change null->false for Firefox for HTML element API features

vinyldarkscratch:
This PR sets lack of support for Firefox for a few features in the HTML element APIs using results from the mdn-bcd-collector project.

### DIFF
--- a/api/HTMLMediaElement.json
+++ b/api/HTMLMediaElement.json
@@ -677,7 +677,10 @@
               "version_added": "≤79"
             },
             "firefox": {
-              "version_added": null
+              "version_added": false
+            },
+            "firefox_android": {
+              "version_added": false
             },
             "ie": {
               "version_added": false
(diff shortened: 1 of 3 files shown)


api/HTMLTextAreaElement.json
```diff
@@ -61,10 +61,10 @@
               "version_added": "79"
             },
             "firefox": {
-              "version_added": null
+              "version_added": false
```
foolip, line 64: I took a closer look at `autocapitalize` and its history is long and twisty! Filed https://github.com/mdn/browser-compat-data/issues/7519, but at the end of the day this change is correct for Gecko.

foolip (approved):

api/HTMLMediaElement.json
```diff
@@ -677,7 +677,10 @@
               "version_added": "≤79"
             },
             "firefox": {
-              "version_added": null
+              "version_added": false
```
foolip, line 680: Yep, `controlsList` is a thing only in Chromium so far I think.

foolip (approved):
OK, sorted out the merge conflicts and confirmed that only the expected changes remain.

PR #7498, opened 2020-11-25
Correct Firefox data for HTMLBaseFontElement

vinyldarkscratch:
This PR corrects the Firefox data for the HTMLBaseFontElement API using results from the mdn-bcd-collector project.  As it turns out, this API was supported for a brief period in the first versions of Firefox.

### DIFF
--- a/api/HTMLBaseFontElement.json
+++ b/api/HTMLBaseFontElement.json
@@ -15,10 +15,11 @@
             "version_removed": "79"
           },
           "firefox": {
-            "version_added": false
+            "version_added": "1",
+            "version_removed": "4"
           },
           "firefox_android": {
-            "version_added": null
+            "version_added": false
           },
           "ie": {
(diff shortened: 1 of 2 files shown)


api/HTMLBaseFontElement.json
```diff
@@ -15,10 +15,11 @@
             "version_removed": "79"
           },
           "firefox": {
-            "version_added": false
+            "version_added": "1",
```
foolip, line 18: :) It's fun that we can add this data easily, even though I don't think anyone will start using `<basefont>` now as a result.

foolip (commented):
Can you also update basefont.json to match? This all seems to back to https://bugzilla.mozilla.org/show_bug.cgi?id=3875, a lively bug from 1999-2005 :)

foolip (approved):

PR #7499, opened 2020-11-25
Test version range comparisons in the linter

ddbeck:
While reviewing @foolip's https://github.com/mdn/browser-compat-data/pull/6558, I thought it was difficult to understand the linter's check that versions make sense with respect to their descendant features. I ended up writing some tests against that code. This attempts to bring those tests into the project properly.

I'll admit this has largely been an exercise in me trying to make sense of our own linter. But because we're not in the habit of testing things, a surprising amount changes here:

1. I expanded the glob to match tests nested more deeply. Maybe this is a bad practice. I'm not sure!
2. I made some changes to make things more testable. `test-consistency.js` exposes a function that touches the file system. Testing with the file system is a pain; instead, I've moved the class (`ConsistencyChecker`) that contains all the feature comparisons to its own module and tested it from there.
3. I actually added a few tests!


ddbeck:
I'm withdrawing this. This doesn't have a clear problem to solve, which I think is an ongoing issue with many of the infra PRs we have. I'll be auditing more with that in mind in the future.

PR #7500, opened 2020-11-25
Correct Firefox 1-4 data for HTML element APIs

vinyldarkscratch:
This PR uses results from the mdn-bcd-collector project to correct Firefox data between versions 1-4 for the HTML element APIs.

### DIFF
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -537,7 +537,7 @@
               ]
             },
             "firefox_android": {
-              "version_added": "3",
+              "version_added": "4",
               "notes": [
                 "Before Firefox 5, <code>click()</code> is only defined on buttons and inputs, and has no effect on text and file inputs.",
                 "Starting in Firefox for Android 79, the <code>click()</code> function works even when the element is not attached to a DOM tree."
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -785,7 +785,7 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "3.5"
+              "version_added": "2"
             },
             "firefox_android": {
               "version_added": "4"
@@ -1140,7 +1140,7 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "4"
+              "version_added": "1"
             },
             "firefox_android": {
               "version_added": "4"
@@ -3249,7 +3249,7 @@
               }
             ],
             "firefox": {
-              "version_added": "1.5"
+              "version_added": "1"
             },
             "firefox_android": {
               "version_added": "4"
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -537,7 +537,11 @@
               ]
             },
             "firefox_android": {
-              "version_added": "5"
+              "version_added": "3",
+              "notes": [
+                "Before Firefox 5, <code>click()</code> is only defined on buttons and inputs, and has no effect on text and file inputs.",
+                "Starting in Firefox for Android 79, the <code>click()</code> function works even when the element is not attached to a DOM tree."
+              ]
             },
             "ie": {
               "version_added": "5.5"
@@ -781,7 +785,7 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "2"
+              "version_added": "3.5"
             },
             "firefox_android": {
               "version_added": "4"
@@ -1136,7 +1140,7 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "1"
+              "version_added": "4"
             },
             "firefox_android": {
               "version_added": "4"
@@ -1443,7 +1447,7 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "1"
+              "version_added": "4"
             },
             "firefox_android": {
               "version_added": "4"
@@ -3245,7 +3249,7 @@
               }
             ],
             "firefox": {
-              "version_added": "1"
+              "version_added": "1.5"
             },
             "firefox_android": {
               "version_added": "4"
--- a/api/HTMLFieldSetElement.json
+++ b/api/HTMLFieldSetElement.json
@@ -156,7 +156,7 @@
                 "version_added": "≤18"
               },
               "firefox": {
-                "version_added": "1"
+                "version_added": "4"
               },
               "firefox_android": {
                 "version_added": "4"
--- a/api/HTMLFieldSetElement.json
+++ b/api/HTMLFieldSetElement.json
@@ -61,7 +61,7 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "4"
+              "version_added": "1"
             },
             "firefox_android": {
               "version_added": "4"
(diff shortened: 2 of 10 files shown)


foolip:
@vinyldarkscratch this changes a lot of data from Firefox 1 to 4, and I want to make sure this is correct. For each of these changes, it would be great if you could check:
- Is there another entry by the same name somewhere, and does its data make sense together with these changes? I see there are many autofocus entries, and it would make sense if they were all introduced at the same time.
- Is the member on this interface in the spec? If not, has it been moved between prototypes? I would expect to uncover more situations like https://github.com/mdn/browser-compat-data/issues/7519.
- What does the api.HTML*Element test ([example](http://mdn-bcd-collector.appspot.com/tests/api/HTMLFieldSetElement)) do? Was the element itself supported, or should the parent entry for a lot of these also change? It seems suspect that most of  HTMLFieldSetElement's members should have been introduced after the interface itself.

foolip:
I'll go ahead and review this myself, doing the checks in https://github.com/mdn/browser-compat-data/pull/7500#issuecomment-734775015.

api/HTMLElement.json
```diff
@@ -1136,7 +1140,7 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "1"
+              "version_added": "4"
```
foolip, line 1143: "hidden" is also a reflected attribute and listed in global_attributes.json. I will revert this and other changes like it, as it's possible the content attribute was supported before the reflected attribute, but when that's the case we might need to add notes.

api/HTMLElement.json
```diff
@@ -1443,7 +1447,7 @@
...
```
foolip, line 1450: This is `isContentEditable`, a plain IDL attribute, so I trust the test results.

api/HTMLElement.json
```diff
@@ -3245,7 +3249,7 @@
               }
             ],
             "firefox": {
-              "version_added": "1"
+              "version_added": "1.5"
```
foolip, line 3252: This is `tabIndex` a reflected global attributes. Will revert.

api/HTMLFieldSetElement.json
```diff
@@ -61,7 +61,7 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "1"
+              "version_added": "4"
```
foolip, line 64: This is `disabled`, a reflected attributes. Reverting.

api/HTMLFieldSetElement.json
```diff
@@ -109,7 +109,7 @@
               "version_added": "17"
             },
             "firefox": {
-              "version_added": "1"
+              "version_added": "4"
```
foolip, line 112: This is `elements`, a plain IDL attribute, so I trust the results.

api/HTMLFieldSetElement.json
```diff
@@ -253,7 +253,7 @@
               "version_added": "14"
             },
             "firefox": {
-              "version_added": "1"
+              "version_added": "4"
```
foolip, line 256: `name` is a reflected attribute. Reverting.

api/HTMLFieldSetElement.json
```diff
@@ -349,7 +349,7 @@
               "version_added": "17"
             },
             "firefox": {
-              "version_added": "1"
+              "version_added": "4"
```
foolip, line 352: `type` is a plain IDL attribute. Trusting.

api/HTMLFieldSetElement.json
```diff
@@ -397,7 +397,7 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "1"
+              "version_added": "4"
```
foolip, line 400: `validationMessage` and the following aren't reflected content attributes, so trusting.

api/HTMLMediaElement.json
```diff
@@ -1436,7 +1436,7 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "3.6"
+              "version_added": "3.5"
```
foolip, line 1439: Makes sense, 3.5 is when the rest of the initial `HTMLMediaElement` was shipped.

api/HTMLObjectElement.json
```diff
@@ -205,7 +205,7 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "1"
+              "version_added": "4"
```
foolip, line 208: http://mdn-bcd-collector.appspot.com/tests/api/HTMLObjectElement looks sensible and none of these things on `HTMLObjectElement` are reflected attributes, so I trust the results.

api/HTMLScriptElement.json
```diff
@@ -201,7 +201,7 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "3.5"
+              "version_added": "1"
```
foolip, line 204: There's a "defer" entry in script.json with notes, updating this would require a bit more care. I'll revert.

api/HTMLSelectElement.json
```diff
@@ -494,7 +494,7 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "4"
+              "version_added": "1"
```
foolip, line 497: http://mdn-bcd-collector.appspot.com/tests/api/HTMLSelectElement looks good and this is decreasing the version number, so trusting.

api/HTMLSelectElement.json
```diff
@@ -1168,7 +1168,7 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "2"
+              "version_added": "1"
```
foolip, line 1171: This is "type", I checked that it's not in select.json.

api/HTMLTableElement.json
```diff
@@ -781,7 +781,7 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "3",
+              "version_added": "1",
```
foolip, line 784: This is `insertRow()`, trusting this change.

api/HTMLVideoElement.json
```diff
@@ -14,7 +14,7 @@
             "version_added": "12"
           },
           "firefox": {
-            "version_added": "4"
+            "version_added": "3.5"
```
foolip, line 17: Indeed, `HTMLVideoElement` had to have been shipped together with `HTMLMediaElement` and `HTMLAudioElement`.

api/HTMLVideoElement.json
```diff
@@ -555,7 +555,7 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "4"
+              "version_added": "3.6"
```
foolip, line 558: This is the "poster" attribute, and the entry in video.json already says 3.6. Good!

foolip (approved):

foolip:
@vinyldarkscratch if you'd like to follow up on the reflected attributes that need to be checked against the content attribute data, please see the two commits I pushed before merging, cae4523e7851290450cdde44e647cc0a9c5e8756 and 17bb762f2143109aa107542ba512c07aebdbcc77.

PR #7501, opened 2020-11-25
Correct Firefox data for HTMLContentElement API

vinyldarkscratch:
This PR corrects the Firefox data for the HTMLContentElement API using results from the mdn-bcd-collector project.  The collector indicated that this API was supported by default without the need of enabling the flag, but it was disabled by default in Firefox 52.  While the flag is present in Firefox 53, it was removed in Firefox 59, which is close to the flag removal requirement, so I figured it made sense to just remove the flag data.

### DIFF
--- a/api/HTMLContentElement.json
+++ b/api/HTMLContentElement.json
@@ -15,25 +15,11 @@
           },
           "firefox": {
             "version_added": "28",
-            "version_removed": "59",
-            "flags": [
-              {
-                "type": "preference",
-                "name": "dom.webcomponents.enabled",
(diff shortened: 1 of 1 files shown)


foolip:
> While the flag is present in Firefox 53, it was removed in Firefox 59, which is close to the flag removal requirement, so I figured it made sense to just remove the flag data.

To confirm, Firefox 59 was released in 2018-03-13, which is more than two years ago. (If it had been close but less than two years ago, we should have waited.)

foolip (approved):

PR #7502, opened 2020-11-25
Correct Firefox version numbers for HTML element APIs

vinyldarkscratch:
This PR corrects the Firefox version numbers for various features within the HTML element APIs based upon results from the mdn-bcd-collector project.  This PR solely changes version numbers, and does not change true -> false or remove flags.

### DIFF
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -635,10 +635,10 @@
               "version_removed": "79"
             },
             "firefox": {
-              "version_added": "1"
+              "version_added": "8"
             },
             "firefox_android": {
-              "version_added": "4"
+              "version_added": "8"
             },
             "ie": {
               "version_added": false
@@ -1969,7 +1969,7 @@
               "version_added": "75"
             },
             "firefox_android": {
-              "version_added": false
+              "version_added": "79"
             },
             "ie": {
               "version_added": false
@@ -2254,10 +2254,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "3"
+              "version_added": "9"
             },
             "firefox_android": {
-              "version_added": "4"
+              "version_added": "9"
             },
             "ie": {
               "version_added": "5.5"
@@ -2302,10 +2302,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "3"
+              "version_added": "9"
             },
             "firefox_android": {
-              "version_added": "4"
+              "version_added": "9"
             },
             "ie": {
               "version_added": "5.5"
@@ -2398,10 +2398,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "3"
+              "version_added": "9"
             },
             "firefox_android": {
-              "version_added": "4"
+              "version_added": "9"
             },
             "ie": {
               "version_added": "5.5"
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -2254,10 +2254,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "9"
+              "version_added": "3"
             },
             "firefox_android": {
-              "version_added": "9"
+              "version_added": "4"
             },
             "ie": {
               "version_added": "5.5"
--- a/api/HTMLHeadElement.json
+++ b/api/HTMLHeadElement.json
@@ -61,12 +61,11 @@
               "version_removed": "79"
             },
             "firefox": {
-              "version_added": true,
-              "version_removed": "7"
+              "version_added": "1",
+              "version_removed": "4"
             },
             "firefox_android": {
-              "version_added": true,
-              "version_removed": "7"
+              "version_added": false
             },
             "ie": {
               "version_added": "≤6"
--- a/api/HTMLImageElement.json
+++ b/api/HTMLImageElement.json
@@ -715,7 +715,7 @@
               "version_added": "75"
             },
             "firefox_android": {
-              "version_added": false
+              "version_added": "79"
             },
             "ie": {
               "version_added": false
@@ -813,7 +813,7 @@
               "version_added": true
             },
             "firefox_android": {
-              "version_added": null
+              "version_added": true
             },
             "ie": {
               "version_added": null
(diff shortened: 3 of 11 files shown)


api/HTMLElement.json
```diff
@@ -2254,10 +2254,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "3"
+              "version_added": "9"
```
foolip, line 2257: Was `oncopy` on some other interface before Firefox 9? How about the copy_event entry in Element.json, does that need updating? And is there a good reason it's over on `Element` instead of here?

Ditto for the following `oncut` and `onpaste`.

api/HTMLHeadElement.json
```diff
@@ -61,12 +61,11 @@
               "version_removed": "79"
             },
             "firefox": {
-              "version_added": true,
-              "version_removed": "7"
+              "version_added": "1",
+              "version_removed": "4"
```
foolip, line 65: Here's Firefox 4 again. Was Firefox 4 very special, or why does it show up so much?

http://mdn-bcd-collector.appspot.com/tests/api/HTMLHeadElement/profile returns false for me in Firefox 3.6 on Windows XP (BrowserStack) as well, is something wrong here?

api/HTMLMediaElement.json
```diff
@@ -2296,10 +2296,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "3.5"
+              "version_added": "9"
```
foolip, line 2299: Can you update `error_event` as well? This doesn't seem right, but maybe the event itself was fired from 3.5 but the event handler attribute was missing. http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8721 is a test for it, it should be easy to turn into a custom test if we have async tests.

api/HTMLSourceElement.json
```diff
@@ -120,10 +120,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "3.5"
+              "version_added": "15"
```
foolip, line 123: Hmm, what's going on here? This isn't when `<picture>` support was shipped, but it does match the media attribute data in source.json.

Since the `<source media>` attribute has been used for both `<video>` and `<picture>` and the use in the former context was removed from the spec, it would be good to have some notes around this. Not necessary for this PR though, I expect this is making the data more correct.

api/HTMLImageElement.json
```diff
@@ -813,7 +813,7 @@
               "version_added": true
             },
             "firefox_android": {
-              "version_added": null
+              "version_added": true
```
foolip, line 816: Is this just mirroring the desktop data, or how do we know `lowsrc` is supported on current Firefox for Android?

api/HTMLVideoElement.json
```diff
@@ -124,7 +124,7 @@
               }
             ],
             "firefox_android": {
-              "version_added": false
+              "version_added": "42"
```
foolip, line 127: Not mirroring the flag entry, good :)

foolip:
> This PR solely changes version numbers, and does not change true -> false or remove flags.

This is much appreciated, but I'm finding myself still reviewing real &rarr; real changes in two groups, those that increase the version number and those that decrease the version number separately. The former is much more likely to be problematic and a result of issues with our methodology. Let's leave existing PRs alone, but it would be great to split future PRs along these lines. (Small PRs with less than 5 changes are fine to mix, I just don't want to hold up the good changes I trust because of doubts with the other half.)

api/HTMLElement.json
```diff
@@ -635,10 +635,10 @@
               "version_removed": "79"
             },
             "firefox": {
-              "version_added": "1"
+              "version_added": "8"
```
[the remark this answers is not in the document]
vinyldarkscratch: The original version number came from the wiki migration, and is inconsistent with the data of the [`<menu>` HTML element data](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menu).  Since a context menu needs a `<menu>` element to be defined, it doesn't make sense that this property is supported before this HTML element.

api/HTMLElement.json
vinyldarkscratch: We probably need to look into this a little bit deeper; it doesn't make sense that the event handlers are on HTMLElement, but the events themselves are on Element.

api/HTMLHeadElement.json
vinyldarkscratch: Hmmm...Firefox 3.6 on Windows XP and 7 are returning `true` for me.  I know that sometimes with the older Firefox versions, BrowserStack doesn't close the previous version before attempting to open the tabs, resulting in it opening the same version (so if you're on Firefox 4, but then switch to 3.6, it doesn't close Firefox 4 and ends up still using Firefox 4).

api/HTMLElement.json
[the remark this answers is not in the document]
foolip: OK, thanks for confirming!

api/HTMLHeadElement.json
foolip: Oh, that sounds concerning! Is this when running under Selenium, or manually?

vinyldarkscratch: Luckily, this is just when running tests manually; running in Selenium has no problem.  I've sent a message to BrowserStack support about it, hopefully the problem will be resolved soon!

api/HTMLImageElement.json
vinyldarkscratch: The collector has results for Firefox Android 83, and is updating Fenix data with the results -- so, this is confirmed supported on Fenix!

foolip: OK, excellent!

api/HTMLElement.json
vinyldarkscratch: Reverted these changes, will dive deeper into this another time.

api/HTMLObjectElement.json
```diff
@@ -1021,10 +1021,12 @@
               "version_added": false
             },
             "firefox": {
-              "version_added": "27"
+              "version_added": "27",
+              "version_removed": "68"
```
foolip, line 1025: Removal confirmed by https://bugzilla.mozilla.org/show_bug.cgi?id=1548773

foolip (approved):

PR #7505, opened 2020-11-25
Correct Firefox to false for unsupported HTML element APIs

vinyldarkscratch:
This PR uses results from the mdn-bcd-collector project to set various features to `false` for HTML element APIs, when they were initially marked as `true` (or had some version number).  Most of these features were already determined as "ghost features" (features in BCD that were never implemented or in any spec), others still marked as supported in Firefox Android when the desktop data was changed, and a few changes for features supported in Firefox Desktop but not Firefox Android.


api/HTMLMarqueeElement.json
```diff
@@ -438,10 +438,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "65"
+              "version_added": false
```
saschanaz, line 441: Firefox and Chrome have `trueSpeed` instead, so maybe just a typo.

vinyldarkscratch: Argh, we seem to have a lot of these miscapitalized entries...

foolip: Yep :) You've already opened https://github.com/foolip/mdn-bcd-collector/issues/812, and doing that soon would be good, as now we're relying on the reviewer to spot suspicious changes.

api/HTMLAnchorElement.json
```diff
@@ -253,10 +253,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "1"
+              "version_added": false
```
foolip, line 256: There is no `media` attribute on `HTMLAnchorElement` or the `HTMLHyperlinkElementUtils` mixin in the spec. I don't think there ever has been. Same for `HTMLAreaElement`. These entries should be removed instead.

foolip: I've sent https://github.com/mdn/browser-compat-data/pull/7538 to do that.

api/HTMLAreaElement.json
```diff
@@ -253,10 +253,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "1"
+              "version_added": false
```
foolip, line 256: Removing this `media` entry in https://github.com/mdn/browser-compat-data/pull/7538.

api/HTMLAreaElement.json
```diff
@@ -205,10 +205,10 @@
...
```
foolip, line 208: `hreflang` on `HTMLAreaElement` is also a mistaken entry which should be removed.

api/HTMLAreaElement.json
```diff
@@ -589,10 +589,10 @@
...
```
foolip, line 592: This `type` entry should also be removed, no such thing: https://html.spec.whatwg.org/#htmlareaelement

api/HTMLElement.json
```diff
@@ -832,10 +832,10 @@
               "version_removed": "79"
             },
             "firefox": {
-              "version_added": "1"
+              "version_added": false
```
foolip, line 835: Indeed, no `dropzone` in Gecko, it's been commented out since it was added in https://github.com/mozilla/gecko-dev/commit/f19ca9b71f06760fa782f23e2854119da2d5abe9

api/HTMLAreaElement.json
vinyldarkscratch: Opened #7539 to remove this.

api/HTMLAreaElement.json
vinyldarkscratch: Opened as #7540.

api/HTMLInputElement.json
```diff
@@ -1588,10 +1588,10 @@
               "version_added": "≤18"
             },
             "firefox": {
-              "version_added": "16"
+              "version_added": false
```
foolip, line 1591: Indeed not supported per http://mdn-bcd-collector.appspot.com/tests/api/HTMLInputElement/weight.

But what is `input.weight`, was it ever a thing?

I suspect not, since a `width` entry is missing I suspect this is just a typo of that which was never caught before. Can you rename the entry and update the data with collector results? It should probably be the same as for `height`, but not necessarily, it seems plausible `width` is actually older than `height`.

foolip (commented):
Well, pretty much everything about this PR was complicated :)

I would suggest breaking out everything complicated, leaving only the three entries on HTMLElement in this PR.

api/HTMLIFrameElement.json
```diff
@@ -234,7 +234,7 @@
               "version_added": "56"
             },
             "firefox_android": {
-              "version_added": true
+              "version_added": false
```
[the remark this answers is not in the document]
vinyldarkscratch: Odd, a range should have been generated for this...  Fixing now!

api/HTMLInputElement.json
```diff
@@ -1495,7 +1495,7 @@
               "version_added": "50"
             },
             "firefox_android": {
-              "version_added": "50"
+              "version_added": false
```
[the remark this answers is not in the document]
vinyldarkscratch: It looks like it is indeed disabled in Firefox Android by default.  I've updated the data and added the flag.

api/HTMLInputElement.json
vinyldarkscratch: I've created #7671 to resolve this issue.

api/HTMLIFrameElement.json
```diff
@@ -231,10 +231,12 @@
               "version_added": "15"
             },
             "firefox": {
-              "version_added": "56"
+              "version_added": "56",
+              "version_removed": "83"
```
foolip, line 235: Removal confirmed by https://bugzilla.mozilla.org/show_bug.cgi?id=1665252

api/HTMLInputElement.json
```diff
@@ -1543,7 +1550,7 @@
...
```
foolip, line 1553: This looked a bit suspect, but confirmed with http://mdn-bcd-collector.appspot.com/tests/api/HTMLInputElement that `webkitEntries` isn't enabled. It is behind a "dom.webkitBlink.filesystem.enabled" pref if you'd like to add it, but I won't block this PR on it.

foolip (approved):

PR #7506, opened 2020-11-25
Correct Firefox data for api.HTMLMediaElement.seekToNextFrame

vinyldarkscratch:
This PR uses results from the mdn-bcd-collector project to correct and update the Firefox data for the `seekToNextFrame` feature of the HTMLMediaElement API.  Initially, the feature was supported behind a flag; however, it was enabled by default in Firefox 56, but the data was never updated to reflect this.  This PR sets the version to 56 and removes the flag data, as it meets the flag removal requirements.

### DIFF
--- a/api/HTMLMediaElement.json
+++ b/api/HTMLMediaElement.json
@@ -3356,24 +3356,10 @@
               "version_added": false
             },
             "firefox": {
-              "version_added": "49",
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "media.seekToNextFrame",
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
I was skeptical that this is really shipping (given it is not standardized), but it is in fact there. Nice find!

Bugs that demand for standardization:
- https://bugzilla.mozilla.org/show_bug.cgi?id=1336404
- https://bugs.chromium.org/p/chromium/issues/detail?id=1065675

PR #7509, opened 2020-11-25
Add real values for Firefox for History API

vinyldarkscratch:
This PR uses results from the mdn-bcd-collector project to add real values for the History API for Firefox.

### DIFF
--- a/api/History.json
+++ b/api/History.json
@@ -61,10 +61,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": true
+              "version_added": "1"
             },
             "firefox_android": {
-              "version_added": true
+              "version_added": "4"
             },
             "ie": {
               "version_added": "10"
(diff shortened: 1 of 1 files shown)


api/History.json
```diff
@@ -497,10 +497,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": true
+              "version_added": "4"
```
foolip, line 500: Here's Firefox 4 again, but http://mdn-bcd-collector.appspot.com/tests/api/History/state is so simple that I don't think it can be wrong. Confirmed it returns false in Firefox 3 too.

api/History.json
```diff
@@ -452,7 +452,7 @@
               "version_added": "46"
             },
             "firefox_android": {
-              "version_added": true
+              "version_added": "46"
```
foolip, line 455: `history.scrollRestoration` is a relatively recent thing, makes sense.

foolip (approved):

PR #7511, opened 2020-11-26
adding missing pseudos

rachelandrew:
I've been working on completing our documentation for the Level 4 Selectors spec in https://github.com/mdn/sprints/issues/1686

While these pseudos don't currently have browser support I think there is an argument to include them in order that folk can see the state of support. There are lots of articles round the web that mention them, sometimes without any detail that they are not supported.

### DIFF
--- a/css/selectors/current.json
+++ b/css/selectors/current.json
@@ -1,0 +1,55 @@
+{
+  "css": {
+    "selectors": {
+      "current": {
+        "__compat": {
+          "description": "<code>:current</code>",
+          "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/:current",
+          "support": {
+            "chrome": {
+              "version_added": false
+            },
+            "chrome_android": {
+              "version_added": false
+            },
+            "edge": {
+              "version_added": false
+            },
+            "firefox": {
+              "version_added": false
+            },
+            "firefox_android": {
+              "version_added": false
+            },
+            "ie": {
+              "version_added": false
+            },
+            "opera": {
+              "version_added": false
+            },
+            "opera_android": {
+              "version_added": false
+            },
+            "safari": {
+              "version_added": false
+            },
+            "safari_ios": {
+              "version_added": false
+            },
+            "samsunginternet_android": {
+              "version_added": false
+            },
+            "webview_android": {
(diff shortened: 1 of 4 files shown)


lucalves (approved):

ddbeck (changes requested):
Hi Rachel, do you know of anything we might link to in comments here showing that one or more browsers are working on implementing these?

We're getting a bit cautions about including all-`false` features. We'd like to avoid including very speculative, immature proposals, or extremely historic features, since they're a bit of maintenance burden.

If we have some notes showing that these are live proposals, then it's easier to avoid removing these prematurely.

rachelandrew:
@ddbeck in general with all of the missing pseudos I have searched for implementation bugs and would have mentioned them if I had found them.

If we aren't going to include compat data for them that's fine by me. However I think there is a good reason to doc these missing pseudos as there is broad implementation of the spec, and a quick search on stack overflow etc. will turn up someone confused by something they think should work. 

I believe the linter will be upset by a lack of BCD for a feature. The alternative would be to have a formal way to add a line that says "This feature has not yet been implemented by any browser." which would serve the same purpose without the overhead of the BCD.

ddbeck:
Thank you for the added info, @rachelandrew. 

If we don't have evidence of implementation, then I'm inclined to not include it in BCD until we do. BCD should be able to have data for likely features, but probably not all proposed features. I think we've had a few cases of things getting into BCD that never went anywhere, then got pruned from specs, but were never cleaned up from BCD (so-called "ghost" features).

> I believe the linter will be upset by a lack of BCD for a feature. The alternative would be to have a formal way to add a line that says "This feature has not yet been implemented by any browser." which would serve the same purpose without the overhead of the BCD.

You're right about the linter. Yes, I think it makes sense to update the linter to support an explicit opt-out of BCD, like we do for things which don't have specifications. If fixing the linter and closing this PR seems like a reasonable course to you, then would you open an issue in mdn/sprints and tag me and Chris about it? I'm happy to work on this, with his OK.

rachelandrew:
done https://github.com/mdn/sprints/issues/3970

ddbeck:
@rachelandrew thank you!

PR #7512, opened 2020-11-26
Spread syntax is not experimental

lahmatiy:
Spread syntax is a part of [ECMAScript (ECMA-262)](https://tc39.es/ecma262/#sec-object-initializer) for years


Elchi3 (approved):
Yes. 

(we're trying to rework "experimental", so embarrassments like this will hopefully go away).

PR #7490, opened 2020-11-25
Remove the api.AudioBufferSourceNode.start entry

foolip:
The start() method is inherited from AudioScheduledSourceNode:
https://webaudio.github.io/web-audio-api/#AudioScheduledSourceNode

There's already an api.AudioScheduledSourceNode.start entry.

The MDN article has already been updated:
https://wiki.developer.mozilla.org/en-US/docs/Web/API/AudioBufferSourceNode$compare?to=1655102&from=1655101

PR #7516, opened 2020-11-27
Revert "Remove the api.AudioBufferSourceNode.start entry"

foolip:
Reverts mdn/browser-compat-data#7490


foolip:
@ddbeck I made a mistake! It looks like this week's release hasn't happened yet, but it would be great to revert this first and not mention https://github.com/mdn/browser-compat-data/pull/7490 in the release notes.

PR #7518, opened 2020-11-27
Update preservesPitch (shipped in Chromium 86)

foolip:
https://chromestatus.com/feature/5742134990733312
https://storage.googleapis.com/chromium-find-releases-static/f3e.html#f3ec10b358a091faeaa2c6c9586fe9ae2bb769a4

The existing data about a webkit-prefixed variant in Chromium-based
browsers are because it was once a no-op from its WebKit origins. This
no-op was removed (by me) back in 2013:
https://chromium.googlesource.com/chromium/src/+/fca4e7ec60a824062589c7a3dde726cb1ccef902

Adding partial_implementation entries for this isn't going to be of
practical benefit to anyone at all, so it's just dropped instead.


foolip:
@tguilbert-google FYI

jpmedley (approved):

PR #7520, opened 2020-11-27
Remove the api.Node.baseURIObject entry

foolip:
This is a Gecko "ChromeOnly" attribute not exposed to the web:
https://github.com/mozilla/gecko-dev/blob/8a4aa0c699d9ec281d1f576c9be1c6c1f289e4e7/dom/webidl/Node.webidl#L109

The string "baseURIObject" does not appear in Chromium or WebKit source,
apart from one instane in a gecko_dom.js as part of a test suite.

https://developer.mozilla.org/en-US/docs/Web/API/Node/baseURIObject
should be redirected to Node.

### DIFF
--- a/api/Node.json
+++ b/api/Node.json
@@ -151,54 +151,6 @@
           }
         }
       },
-      "baseURIObject": {
-        "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/Node/baseURIObject",
-          "support": {
-            "chrome": {
-              "version_added": null
-            },
-            "chrome_android": {
-              "version_added": null
-            },
-            "edge": {
-              "version_added": null
-            },
-            "firefox": {
-              "version_added": null
-            },
-            "firefox_android": {
-              "version_added": null
-            },
-            "ie": {
-              "version_added": false
-            },
-            "opera": {
-              "version_added": null
-            },
-            "opera_android": {
-              "version_added": null
-            },
-            "safari": {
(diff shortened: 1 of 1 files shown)


foolip:
Once approved, I can take care of redirecting https://developer.mozilla.org/en-US/docs/Web/API/Node/baseURIObject. Or if the reviewer would like to do it and merge directly, that's also OK.

foolip:
https://github.com/mdn/browser-compat-data/pull/7521 is very similar but with an extra error in the data thrown in. Dealing with them together would probably still make sense though.

Elchi3 (approved):
I created the redirect and also removed `baseURIObject` from the main `Node` page.

Is there a list you work from for these? I think there are (at least) 4 more that are mentioned here: https://wiki.developer.mozilla.org/en-US/docs/Web/API/Node#Obsolete_properties

foolip:
@Elchi3 I don't have a list, these were just the two I came across in https://github.com/mdn/browser-compat-data/pull/7515.

3 of the remaining ones in https://wiki.developer.mozilla.org/en-US/docs/Web/API/Node#Obsolete_properties are https://github.com/mdn/browser-compat-data/issues/6545, feedback on the proposed resolution appreciated.

The remaining `rootNode` indeed looks ripe for removal, however. I'll give it a go.

foolip:
The go was given in https://github.com/mdn/browser-compat-data/pull/7524.

PR #7525, opened 2020-11-27
Add real values for Firefox for Range API

vinyldarkscratch:
This PR adds versions for Firefox for a few features of the Range API based upon results from the mdn-bcd-collector project.


foolip (approved):
Ancient API, makes sense.

PR #7527, opened 2020-11-27
Correct Firefox data for Range API

vinyldarkscratch:
This PR corrects the data for the Range API for Firefox based upon results from the mdn-bcd-collector project.  Support was marked as Firefox 4 for this API, however our results indicate it is actually in Firefox 1, along with all of its features.

### DIFF
--- a/api/Range.json
+++ b/api/Range.json
@@ -14,7 +14,7 @@
             "version_added": "12"
           },
           "firefox": {
-            "version_added": "4",
+            "version_added": "1",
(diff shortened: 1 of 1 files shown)


api/Range.json
```diff
@@ -449,7 +449,8 @@
               "version_added": false
             },
             "firefox": {
-              "version_added": false
+              "version_added": "1",
```
foolip, line 452: Interesting! This entry should be remove, but not in this PR, and having this data is a good clue for when it is removed.

foolip (approved):
Range is old, makes sense, this is almost entirely changing Firefox 4 to 1. I guess someone just assumed 4 for both Firefox and Firefox Android at some point.

PR #7528, opened 2020-11-27
Add real values for Firefox for Selection API

vinyldarkscratch:
This PR adds real values for Firefox for the Selection API based upon results from the mdn-bcd-collector project.


foolip (approved):
This was a simple one :)

PR #7529, opened 2020-11-27
Correct Firefox data for Selection API

vinyldarkscratch:
This PR corrects a few versions for the Selection API for Firefox based upon results from the mdn-bcd-collector project.

### DIFF
--- a/api/Selection.json
+++ b/api/Selection.json
@@ -351,7 +351,7 @@
               "version_added": "12"
             },
             "firefox": {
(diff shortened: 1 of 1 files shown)


foolip (approved):
I noticed that https://developer.mozilla.org/en-US/docs/Web/API/Selection says this is experimental, so commented on https://github.com/mdn/browser-compat-data/issues/6905#issuecomment-735714986

PR #7530, opened 2020-11-27
Add real values for Firefox for SVG element APIs

vinyldarkscratch:
This PR adds version numbers for Firefox for the SVG element APIs and its features based upon results from the mdn-bcd-collector project.  This is a cherry-pick of changes that change null/true into a version number.

### DIFF
--- a/api/SVGImageElement.json
+++ b/api/SVGImageElement.json
@@ -14,10 +14,10 @@
             "version_added": "12"
           },
           "firefox": {
-            "version_added": true
+            "version_added": "1.5"
           },
           "firefox_android": {
-            "version_added": true
+            "version_added": "4"
           },
           "ie": {
             "version_added": "9"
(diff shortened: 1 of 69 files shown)


api/SVGPathElement.json
```diff
@@ -14,10 +14,10 @@
             "version_added": "12"
           },
           "firefox": {
-            "version_added": true
+            "version_added": "1.5"
```
foolip, line 17: Interesting, so `SVGPathElement` itself lives on, in a much smaller form:
https://github.com/mozilla/gecko-dev/blob/master/dom/webidl/SVGPathElement.webidl

api/SVGImageElement.json
```diff
@@ -158,10 +158,10 @@
               "version_added": "≤79"
             },
             "firefox": {
-              "version_added": null
+              "version_added": "63"
```
foolip, line 161: This (`decoding`) is the most recent addition out of all these, and it checks out: https://bugzilla.mozilla.org/show_bug.cgi?id=1416328

foolip (approved):
This is huge, but still fairly easy to have confidence in, given how the tests for this work. Example:
http://mdn-bcd-collector.appspot.com/tests/api.SVGSVGElement.contentStyleType

I also doubt checked the most recent addition, since overall I'd expect most of this to be ancient, and false negatives are more likely at the recent end. Looks good though :)

PR #7533, opened 2020-11-28
Correct versions for Firefox for Window API (P1)

vinyldarkscratch:
This PR is part 1 of a few to correct the versions for Firefox for various features of the Window API based upon results from the mdn-bcd-collector project.  This PR specifically touches data where the version added is reduced, or the data is changed from `false` to a version added+removed range.

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -1570,10 +1570,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "34"
+              "version_added": "1"
             },
             "firefox_android": {
-              "version_added": "34"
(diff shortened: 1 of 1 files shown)


api/Window.json
```diff
@@ -4204,10 +4204,12 @@
               "version_added": false
             },
             "firefox": {
-              "version_added": false
+              "version_added": "4",
```
foolip, line 4207: Well that makes much more sense than it never being supported at all!

foolip (approved):
Good stuff! Clarifies why these entries exist to begin with.

PR #7534, opened 2020-11-28
Correct data for Window.convertPointFromNodeToPage/PageToNode

vinyldarkscratch:
While I was going through the Window API to update the Firefox data from mdn-bcd-collector results, I noticed something really weird with the data for the `convertPointFromPageToNode` feature.  The notes were referring to `ondeviceorientationabsolute`, doesn't make sense for this feature.  I traced this back to #2021, which was the initial addition for this data, and I believe that it was meant to be added to `ondeviceorientation` instead.

This corrects the data for both this feature (`convertPointFromPageToNode`), and the sister feature (`convertPointFromNodeToPage`), based upon manual testing in a number of browsers.  The data currently in `ondeviceorientation` seems to be more up to date than this misplaced data.  (Note: this adds "true" without version numbers.  I plan to go through and add version numbers at a later time.)

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -1321,41 +1321,54 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Window/convertPointFromNodeToPage",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": true,
+              "version_removed": true,
+              "alternative_name": "webkitConvertPointFromNodeToPage"
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": true,
+              "version_removed": true,
+              "alternative_name": "webkitConvertPointFromNodeToPage"
             },
             "edge": {
-              "version_added": "≤18"
+              "version_added": false
             },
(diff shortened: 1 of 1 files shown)


api/Window.json
```diff
@@ -1321,41 +1321,54 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Window/convertPointFromNodeToPage",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": true,
+              "version_removed": true,
```
foolip, line 1325: I can confirm that these were once there and then removed, by me :)

https://chromium.googlesource.com/chromium/src/+/ed6edbcb38b468c0168c13ebe13fa2910fe32a41

I think that was Chrome 38, but OK to fill out later as suggested.

foolip (approved):

foolip:
I've sent https://github.com/foolip/mdn-bcd-collector/pull/842 to add the missing bits to the collector.

PR #1119, opened 2018-02-17
Add compat data for HTMLAreaElement

bunnybooboo:
This API has listed mixins, which I've not currently included, via discussion over at #472 

https://developer.mozilla.org/en-US/docs/Web/API/HTMLAreaElement

PR #7538, opened 2020-11-28
Remove media members of HTMLAnchorElement/HTMLAreaElement

foolip:
These entries both come from wiki migration:
https://github.com/mdn/browser-compat-data/pull/1116
https://github.com/mdn/browser-compat-data/pull/1119

These do not exist in the spec, a media IDL attribute exists only on
these three interfaces:
https://html.spec.whatwg.org/#htmllinkelement
https://html.spec.whatwg.org/#htmlsourceelement
https://html.spec.whatwg.org/#htmlstyleelement

BCD already has entries for those.

The existing data suggesting media is supported on a/area seems
mistaken, and can be confirmed with these tests:
http://mdn-bcd-collector.appspot.com/tests/api/HTMLAnchorElement/media
http://mdn-bcd-collector.appspot.com/tests/api/HTMLAreaElement/media

Both return false on Chrome 86, Edge 18, Firefox 82 and Safari 14.

### DIFF
--- a/api/HTMLAnchorElement.json
+++ b/api/HTMLAnchorElement.json
@@ -239,54 +239,6 @@
           }
         }
       },
-      "media": {
-        "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLAnchorElement/media",
-          "support": {
-            "chrome": {
-              "version_added": true
-            },
-            "chrome_android": {
-              "version_added": true
-            },
-            "edge": {
-              "version_added": "12"
-            },
-            "firefox": {
-              "version_added": "1"
-            },
-            "firefox_android": {
-              "version_added": "4"
-            },
-            "ie": {
-              "version_added": false
-            },
-            "opera": {
-              "version_added": true
-            },
-            "opera_android": {
-              "version_added": true
-            },
-            "safari": {
-              "version_added": false
-            },
-            "safari_ios": {
-              "version_added": false
-            },
-            "samsunginternet_android": {
-              "version_added": true
-            },
-            "webview_android": {
-              "version_added": true
-            }
-          },
-          "status": {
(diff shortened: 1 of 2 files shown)


api/HTMLAnchorElement.json
```diff
@@ -239,54 +239,6 @@
           }
         }
       },
-      "media": {
-        "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLAnchorElement/media",
```
foolip, line 244: https://developer.mozilla.org/docs/Web/API/HTMLAnchorElement/media is 404 and I've already cleaned up https://developer.mozilla.org/docs/Web/API/HTMLAnchorElement

api/HTMLAreaElement.json
```diff
@@ -239,54 +239,6 @@
           }
         }
       },
-      "media": {
-        "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLAreaElement/media",
```
foolip, line 244: 404 and I've cleaned up https://wiki.developer.mozilla.org/en-US/docs/Web/API/HTMLAreaElement

Elchi3 (approved):
In the wiki, this was initially documented in 2010, and the spec back then was https://www.w3.org/TR/2010/WD-html5-20100624/text-level-semantics.html#the-a-element where there is HTMLAnchorElement.media. However, I assume it got removed from the spec and wasn't really implemented? Anyway LGTM as it is gone from specs and implementations.

foolip:
I tracked down the removal to https://github.com/whatwg/html/commit/66fcb2357f205448fe2f40d7834a1e8ea2ed283b / https://www.w3.org/Bugs/Public/show_bug.cgi?id=20521

PR #7539, opened 2020-11-28
Remove ghost feature: api.HTMLAreaElement.hreflang

vinyldarkscratch:
This PR removes the ghost feature: api.HTMLAreaElement.hreflang.  This feature is not supported in any browsers (despite what the current data indicates).

### DIFF
--- a/api/HTMLAreaElement.json
+++ b/api/HTMLAreaElement.json
@@ -191,54 +191,6 @@
           }
         }
       },
-      "hreflang": {
-        "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLAreaElement/hreflang",
-          "support": {
-            "chrome": {
-              "version_added": true
-            },
-            "chrome_android": {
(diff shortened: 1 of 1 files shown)


foolip:
https://developer.mozilla.org/en-US/docs/Web/API/HTMLAreaElement will need to be updated too.

vinyldarkscratch:
Update made!

foolip:
I removed one additional mention of `hreflang` on that page, now merging this.

foolip (approved):

foolip:
While reviewing https://github.com/mdn/browser-compat-data/pull/7540 I discovered that actually these were in the HTML spec at one point, but they were removed in https://github.com/whatwg/html/issues/232 / https://github.com/whatwg/html/pull/236.

They were in https://www.w3.org/TR/html50/embedded-content-0.html#the-area-element, so I'll undo the edit I just made to the MDN page. We should still not have the data in BCD, though, since it was never implemented.

PR #7542, opened 2020-11-29
Add/update Chromium data for CanvasRenderingContext2D

vinyldarkscratch:
This PR adds and updates the Chromium data for the CanvasRenderingContext2D API based upon results from the mdn-bcd-collector project.  Many of the updates include mirroring data to Opera, Opera Android, Samsung Internet, and WebView (which all have been confirmed with collector results).

### DIFF
--- a/api/CanvasRenderingContext2D.json
+++ b/api/CanvasRenderingContext2D.json
@@ -1248,20 +1248,9 @@
             "chrome_android": {
               "version_added": "77"
             },
-            "edge": [
-              {
-                "version_added": "79"
-              },
-              {
-                "version_added": "≤79",
-                "flags": [
-                  {
-                    "type": "preference",
-                    "name": "Experimental Web Platform Features"
-                  }
-                ]
-              }
-            ],
+            "edge": {
+              "version_added": "79"
+            },
             "firefox": {
               "version_added": false
             },
--- a/api/CanvasRenderingContext2D.json
+++ b/api/CanvasRenderingContext2D.json
@@ -2439,11 +2439,11 @@
             ],
             "opera_android": [
               {
-                "version_added": "17"
+                "version_added": "18"
               },
               {
                 "version_added": true,
-                "version_removed": "17",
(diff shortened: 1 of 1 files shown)


foolip:
> Many of the updates include mirroring data to Opera, Opera Android, Samsung Internet, and WebView (which all have been confirmed with collector results).

Is this really the case? We don't have all the versions for these browsers to confirm the data. Was it the most recent for each which was tested, or additional ones too in the case of Opera perhaps?

api/CanvasRenderingContext2D.json
```diff
@@ -2369,9 +2390,16 @@
...
             ],
-            "chrome_android": {
-              "version_added": true
-            },
+            "chrome_android": [
+              {
+                "version_added": "30"
+              },
+              {
+                "version_added": true,
```
foolip, line 2398: `webkitImageSmoothingEnabled` is indeed very old, added in https://trac.webkit.org/changeset/117635/webkit.

Perhaps split out `imageSmoothingEnabled` updates separately, and also ensure the Safari data is correct at the same time? At the very least it's now missing support for `webkitImageSmoothingEnabled` in Safari, but maybe it's also conflating prefixed and unprefixed.

I've sent https://github.com/foolip/mdn-bcd-collector/pull/843 to help sort this out.

api/CanvasRenderingContext2D.json
```diff
@@ -3272,10 +3328,22 @@
               "version_added": false
             },
             "opera": {
-              "version_added": true
+              "version_added": true,
+              "flags": [
```
foolip, line 3332: OK, this flag mirroring is good because it makes it clear it's not supported by default, which the previous data implied.

foolip (commented):
Good stuff, just needs `imageSmoothingEnabled` split out since that's complicated.

foolip:
Ping @vinyldarkscratch.

foolip (approved):

PR #7551, opened 2020-12-01
Note support for window.home() in Opera 12.16 (Presto)

foolip:
Based on manually testing http://mdn-bcd-collector.appspot.com/tests/api/Window/home
in Opera 12.16 and Opera 15 in Windows 7 on BrowserStack.


ddbeck (approved):
Looks good. Thank you! 🎉

PR #7552, opened 2020-12-01
Set <menuitem> HTML element and contextmenu global attribute as non-standard

ddbeck:
These were [dropped from the spec due to a lack of vendor interest](https://github.com/whatwg/html/commit/e7e8c88ecdd7cdc96171f05ab6fe23b56dd48d8a).

I also tidied up or confirmed some data that followed from this discovery:

- For Chrome (and derivatives), this was [implemented behind a flag, then moved to a different flag, then completely removed circa Chrome 61](https://bugs.chromium.org/p/chromium/issues/detail?id=87553). Rather than sort out the history of the flags, I opted to remove them under our irrelevant flags guideline.
- I saw no evidence that IE or Edge implemented this.
- Safari [implemented `<menuitem>` but never shipped it](https://trac.webkit.org/changeset/224457/webkit). I couldn't find evidence that the `contextmenu` attribute was ever supported by Safari.
- The Firefox data looks plausible, despite [an attempt at removal](https://bugzilla.mozilla.org/show_bug.cgi?id=1372276).

I have no idea what prompted me to look at this stuff in the first place.

### DIFF
--- a/html/elements/menuitem.json
+++ b/html/elements/menuitem.json
@@ -52,7 +52,7 @@
           },
           "status": {
             "experimental": false,
-            "standard_track": true,
+            "standard_track": false,
             "deprecated": true
           }
         },
@@ -98,7 +98,7 @@
             },
             "status": {
               "experimental": false,
-              "standard_track": true,
+              "standard_track": false,
               "deprecated": true
             }
           }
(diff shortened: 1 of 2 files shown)


sideshowbarker (approved):
> I have no idea what prompted me to look at this stuff in the first place.

😆 I know that feeling… Pressing the button to raise a PR, and trying to think back and remember where I was at 2 hours ago or whatever, when for some reason incidentally related to another thing I was working on, I first decided to take a just few minutes to look at some BCD data…

PR #7553, opened 2020-12-01
Add CSSTransformComponent.toString method

rachelandrew:
The `toString()` method of `CSSTransformComponent` was missing and is causing a scripting error here: https://wiki.developer.mozilla.org/en-US/docs/Web/API/CSSTransformComponent/toString

This PR adds the method.

### DIFF
--- a/api/CSSTransformComponent.json
+++ b/api/CSSTransformComponent.json
@@ -142,6 +142,54 @@
             "deprecated": false
           }
         }
+      },
+      "toString": {
+        "__compat": {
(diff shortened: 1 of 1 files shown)


jpmedley (approved):

ddbeck (approved):

ddbeck:
It's not clear to me why part of the build canceled here (or why GitHub won't let me rerun it). I've run the tests locally and it works though, so I'm going to merge anyway.

PR #7554, opened 2020-12-02
Remove prefix claims for Web Speech API member subfeatures

foolip:
The members (attributes and methods) of the Web Speech API interfaces
are not themselves prefixes, the usage is

> new webkitSpeechRecognition().start()

not the doubly prefixed

> new webkitSpeechRecognition().webkitStart()

There are also issues with prefixes on some of the interfaces
themselves, and a lot of repetetive notes, but this commit is only about
the member subfeatures.


jpmedley (approved):

foolip:
Thanks @jpmedley! I've now sent https://github.com/mdn/browser-compat-data/pull/7565 to clean up some more here.

PR #7555, opened 2020-12-02
Add time pseudo-class selectors (:current, :future:, :past)

rachelandrew:
Working on https://github.com/mdn/sprints/issues/1686

Added the time-based pseudo-classes with information for Safari as discussed on the above issue. The `:current` pseudo is not supported anywhere.


ddbeck:
`:past` and `:future` appear to have been implemented before the Blink fork:

https://trac.webkit.org/changeset/138784/webkit
https://trac.webkit.org/changeset/140707/webkit

I don't know much about how this is used. Can we test to find out whether Chrome (etc.) support these too?

For `:current`, do we have any evidence that anyone's considering implementing this? I'm happy to have the data in here, but it would be nice to link out to additional information, so it doesn't get swept up as we clean out nonexistent features.

ddbeck:
> I don't know much about how this is used. Can we test to find out whether Chrome (etc.) support these too?

To elaborate on this thought a little: if we can show that Chrome supports the `:past` and `:future` features, then we can fairly safely assume that it has supported it since the first version.

Based on other conversations about PRs like this, I think we can drop the `:current` feature (for now).

ddbeck:
Hi @rachelandrew, do you plan to come back to this PR?

rachelandrew:
@ddbeck so you want me to delete the `:current` file?

ddbeck:
> @ddbeck so you want me to delete the `:current` file?

@rachelandrew yes, please!

ddbeck (approved):
Thank you! 🎉

PR #7557, opened 2020-12-02
Add :playing and :paused pseudo-class selectors

rachelandrew:
I am documenting the missing level 4 pseudo-classes in https://github.com/mdn/sprints/issues/1686 as far as I can tell these have no support at present. However as with the other unsupported pseudos I think it is helpful for authors to be able to see lack of support.

### DIFF
--- a/css/selectors/paused.json
+++ b/css/selectors/paused.json
@@ -1,0 +1,55 @@
+{
+  "css": {
+    "selectors": {
+      "paused": {
+        "__compat": {
+          "description": "<code>:paused</code>",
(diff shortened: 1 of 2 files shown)


ddbeck:
@rachelandrew Sorry to be a broken record here, but do we have any evidence that we could link to showing that implementation is a live prospect?

ddbeck:
I did some of my own searching here and couldn't find anything to show that this exists in any form except as a spec. I'm going to close this, but feel free to reopen if something turns up or the status changes. Thank you!

PR #7567, opened 2020-12-04
Complete nodejs data for functions

gilmoreorless:
Part of #6249. This fills in all the `true` and `null` values for all function-related entries for Node.js

Thanks to the new minimum version of `0.10.0`, most of the entries get assigned to that version. The rest were checked on [node.green](https://node.green) and cross-referenced against Chrome via the V8 version (plus some manual testing).

### DIFF
--- a/javascript/builtins/Function.json
+++ b/javascript/builtins/Function.json
@@ -25,7 +25,7 @@
               "version_added": "4"
             },
             "nodejs": {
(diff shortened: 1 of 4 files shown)


Elchi3 (approved):
Nothing here looks surprising to me, thanks so much for finishing this up @gilmoreorless :tada:

PR #7573, opened 2020-12-05
Add additional features for CSSCounterStyleRule

vinyldarkscratch:
This PR is a part of a project to add missing features to BCD that are from a spec in Editor's Draft or more and is supported in at least one browser.  This particular PR adds missing features for the CSSCounterStyleRule API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).

Spec: https://drafts.csswg.org/css-counter-styles-3/#the-csscounterstylerule-interface

### DIFF
--- a/api/CSSCounterStyleRule.json
+++ b/api/CSSCounterStyleRule.json
@@ -46,6 +46,523 @@
           "standard_track": true,
           "deprecated": false
         }
+      },
+      "additiveSymbols": {
+        "__compat": {
+          "support": {
+            "chrome": {
+              "version_added": false
+            },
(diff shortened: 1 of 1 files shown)


foolip (approved):
This was pretty straightforward. These are all the attributes in https://drafts.csswg.org/css-counter-styles-3/#the-csscounterstylerule-interface, and all of them were introduced in Firefox 33 and nowhere else. https://developer.mozilla.org/en-US/docs/Web/API/CSSCounterStyleRule documents them all but none of the subpages exist.

PR #7578, opened 2020-12-05
Add CSSFontFeatureValuesRule API

vinyldarkscratch:
This PR is a part of a project to add missing features to BCD that are from a spec in Editor's Draft or more and is supported in at least one browser.  This particular PR adds the CSSFontFeatureValuesRule API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).

Spec: https://drafts.csswg.org/css-fonts-4/

### DIFF
--- a/api/CSSFontFeatureValuesRule.json
+++ b/api/CSSFontFeatureValuesRule.json
@@ -1,0 +1,98 @@
+{
+  "api": {
+    "CSSFontFeatureValuesRule": {
+      "__compat": {
+        "support": {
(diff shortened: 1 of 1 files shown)


foolip (approved):
Curious but seems correct: https://github.com/mozilla/gecko-dev/blob/master/dom/webidl/CSSFontFeatureValuesRule.webidl has only this one attribute, and does not support `CSSFontFeatureValuesMap`.

PR #7581, opened 2020-12-05
Add/update data for dir HTML element

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from a spec in Editor's Draft or more and is supported in at least one browser.  This particular PR adds the missing HTMLDirectoryElement API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).

Spec: https://html.spec.whatwg.org/multipage/obsolete.html#htmldirectoryelement

Edit: this also updates the data for the HTML element itself.

### DIFF
--- a/api/HTMLDirectoryElement.json
+++ b/api/HTMLDirectoryElement.json
@@ -19,7 +19,7 @@
             "version_added": "4"
           },
           "ie": {
-            "version_added": "≤9"
+            "version_added": "≤6"
           },
           "opera": {
             "version_added": "≤12.1"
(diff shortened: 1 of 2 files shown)


api/HTMLDirectoryElement.json
```diff
@@ -0,0 +1,98 @@
+{
+  "api": {
+    "HTMLDirectoryElement": {
```
foolip, line 3: This data seems to contradict what we have in dir.json, and I suspect that data must be wrong. Because http://mdn-bcd-collector.appspot.com/tests/api/HTMLDirectoryElement involves `document.createElement('dir')`, if those tests return true then it also implies support for the element, not just the interface.

Simply pasting the same data would be fine I think, given how old all of this is and the ranges there's very little room for it to be wrong, so no need to dig deep here.

foolip (approved):

PR #7582, opened 2020-12-05
Add various orientation properties

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from a spec in Editor's Draft or more and is supported in at least one browser.  This particular PR adds missing features for the HTMLBodyElement API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).

Spec: https://compat.spec.whatwg.org/

### DIFF
--- a/api/HTMLBodyElement.json
+++ b/api/HTMLBodyElement.json
@@ -239,6 +239,53 @@
           }
         }
       },
+      "onorientationchange": {
+        "__compat": {
+          "support": {
+            "chrome": {
+              "version_added": false
+            },
+            "chrome_android": {
+              "version_added": true
+            },
+            "edge": {
+              "version_added": false
+            },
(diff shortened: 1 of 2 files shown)


foolip (commented):
Can you also add Window.orientation + Window.onorientationchange based on http://mdn-bcd-collector.appspot.com/tests/api/Window/orientation + http://mdn-bcd-collector.appspot.com/tests/api/Window/onorientationchange?

I'd expect the data to be the same for all of these.

Getting real version numbers for this might be hard, but for Safari iOS we should be able to get it, right?

vinyldarkscratch:
Window.orientation is already existent, so I've added onorientationchange.

foolip (approved):

PR #7586, opened 2020-12-05
Add missing features for HTMLEmbedElement

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from a spec in Editor's Draft or more and is supported in at least one browser.  This particular PR adds missing features for the HTMLEmbedElement API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).

Spec: https://html.spec.whatwg.org/multipage/iframe-embed-object.html#htmlembedelement

### DIFF
--- a/api/HTMLEmbedElement.json
+++ b/api/HTMLEmbedElement.json
@@ -96,7 +96,7 @@
           "status": {
             "experimental": false,
             "standard_track": true,
-            "deprecated": false
+            "deprecated": true
           }
         }
       },
--- a/api/HTMLEmbedElement.json
+++ b/api/HTMLEmbedElement.json
@@ -237,7 +237,7 @@
           "status": {
             "experimental": false,
             "standard_track": true,
-            "deprecated": false
+            "deprecated": true
           }
         }
       },
--- a/api/HTMLEmbedElement.json
+++ b/api/HTMLEmbedElement.json
@@ -52,6 +52,335 @@
           "standard_track": true,
           "deprecated": false
         }
+      },
+      "align": {
+        "__compat": {
+          "support": {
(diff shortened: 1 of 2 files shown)


foolip:
@vinyldarkscratch can you go through this and see which are reflected content attributes and add/update the corresponding entries in embed.json?

api/HTMLEmbedElement.json
```diff
@@ -52,6 +52,335 @@
...
+              "version_added": "1.0"
+            },
+            "webview_android": {
+              "version_added": "1"
+            }
+          },
+          "status": {
+            "experimental": false,
+            "standard_track": true,
+            "deprecated": false
```
foolip, line 99:
```suggestion
            "deprecated": true
```

https://html.spec.whatwg.org/multipage/obsolete.html#HTMLEmbedElement-partial

api/HTMLEmbedElement.json
```diff
@@ -52,6 +52,335 @@
...
```
foolip, line 240:
```suggestion
            "deprecated": true
```

https://html.spec.whatwg.org/multipage/obsolete.html#HTMLEmbedElement-partial

api/HTMLEmbedElement.json
```diff
@@ -52,6 +52,335 @@
           "standard_track": true,
           "deprecated": false
         }
+      },
+      "align": {
```
foolip, line 56: Looks like content attribute entries are still missing for "align" and maybe other IDL attributes added here.

foolip (approved):

PR #7587, opened 2020-12-06
Add checkValidity/setCustomValidity features of HTMLFieldSetElement API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from a spec in Editor's Draft or more and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6), for the HTMLFieldSetElement API.

Spec: https://html.spec.whatwg.org/multipage/form-elements.html#htmlfieldsetelement

### DIFF
--- a/api/HTMLFieldSetElement.json
+++ b/api/HTMLFieldSetElement.json
@@ -47,6 +47,53 @@
           "deprecated": false
         }
       },
+      "checkValidity": {
+        "__compat": {
+          "support": {
+            "chrome": {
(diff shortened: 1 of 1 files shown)


foolip:
Renamed to not seem like a dupe of https://github.com/mdn/browser-compat-data/pull/7583 :)

foolip (approved):

Issue #7519, opened 2020-11-27
Sort out data for autocapitalize


foolip:
The `autocapitalize` content attribute and reflected IDL attribute has a messy history, and isn't captured correctly in BCD.

The current spec for it is here:
https://html.spec.whatwg.org/multipage/interaction.html#autocapitalization
https://html.spec.whatwg.org/multipage/dom.html#htmlelement

The content attribute can be used on any elements that also uses `contenteditable`, and on `<input>`, `<textarea>` and `<form>` elements. Treating it as a [global attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autocapitalize) in BCD seems appropriate. But BCD also has entries in forms.json and textarea.json. These should probably be removed.

The IDL attribute should be on [`HTMLElement`](https://html.spec.whatwg.org/multipage/dom.html#htmlelement), but BCD also has an entry in HTMLTextAreaElement.json.

Implementation history:
- Chromium:
  - added to HTMLInputElement.idl + HTMLTextAreaElement.idl in https://chromium.googlesource.com/chromium/src/+/b9bf6131125ee5afd66d4e7da1fbede9380dd5e4
  - moved to HTMLElement.idl in https://chromium.googlesource.com/chromium/src/+/1de83967752e0cf4f36f3a013b4f1ab687b88e11
- Gecko:
  - added to HTMLElement.idl directly in https://bugzilla.mozilla.org/show_bug.cgi?id=1425291
- WebKit:
  - added (together with `autocorrect`) to HTMLFormElement.idl + HTMLInputElement.idl + HTMLTextAreaElement.idl in https://trac.webkit.org/changeset/160733/webkit
  - moved to HTMLElement.idl in https://trac.webkit.org/changeset/208624/webkit

Additionally, there's platform dependency for this, see https://github.com/mdn/browser-compat-data/issues/6914.

foolip:
I've sent https://github.com/foolip/mdn-bcd-collector/pull/832 to facilitate testing of the the attribute on HTMLFormElement + HTMLInputElement + HTMLTextAreaElement.

foolip:
Testing that stuff resulted in https://github.com/mdn/browser-compat-data/pull/7588. We shouldn't add those entries, but should figure out what `partial_implementation` notes should look like.

PR #7588, opened 2020-12-06
Add api.HTMLFormElement.autocapitalize

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from a spec in Editor's Draft or more and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6), for the HTMLFormElement API.

Spec: https://html.spec.whatwg.org/multipage/forms.html#htmlformelement

### DIFF
--- a/api/HTMLFormElement.json
+++ b/api/HTMLFormElement.json
@@ -143,6 +143,53 @@
           }
         }
       },
+      "autocapitalize": {
+        "__compat": {
+          "support": {
+            "chrome": {
+              "version_added": "66"
+            },
+            "chrome_android": {
+              "version_added": "66"
+            },
+            "edge": {
+              "version_added": "79"
+            },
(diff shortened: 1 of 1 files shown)


foolip (changes requested):
This is from custom IDL in the collector, the attribute isn't on `HTMLFormElement` in the spec. BCD already has an api.HTMLElement.autocapitalize entry.

What I think is needed is to remove all of the autocapitalize except api.HTMLElement.autocapitalize, html.global_attributes.autocapitalize, which is 3 different entries. Also some notes for the entries that remain about how they were previously on only some interfaces.

See https://github.com/mdn/browser-compat-data/issues/7519 for the whole mess.

PR #5442, opened 2020-01-03
Add "loading" attribute for img/iframe HTML elements

vinyldarkscratch:
Supersedes #5284 due to lack of response from the author.  This PR adds the `loading` attribute to the `` and <iframe>` elements, and fixes the linter errors within the original PR.

Enabled by default in Chrome 76: https://chromestatus.com/feature/5645767347798016
Bugzilla bug: https://bugzil.la/1542784 (seemingly only for ``, thus only added to `img.json`)
Webkit bug: https://webkit.org/b/196698

PR #7591, opened 2020-12-06
Add missing features for HTMLIFrameElement

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from a spec in Editor's Draft or more and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6), for the HTMLIFrameElement API.

Spec: https://html.spec.whatwg.org/multipage/semantics.html#htmliframeelement

### DIFF
--- a/api/HTMLIFrameElement.json
+++ b/api/HTMLIFrameElement.json
@@ -621,6 +621,53 @@
           }
         }
       },
+      "getSVGDocument": {
+        "__compat": {
+          "support": {
+            "chrome": {
+              "version_added": "1"
+            },
+            "chrome_android": {
+              "version_added": "18"
+            },
+            "edge": {
+              "version_added": "12"
+            },
+            "firefox": {
+              "version_added": "3.5"
+            },
+            "firefox_android": {
+              "version_added": "4"
+            },
+            "ie": {
+              "version_added": "9"
+            },
+            "opera": {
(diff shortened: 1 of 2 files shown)


api/HTMLIFrameElement.json
```diff
@@ -669,6 +716,53 @@
           }
         }
       },
+      "loading": {
```
foolip, line 719: This is a reflected attribute, can you also update the content attribute?

html/elements/iframe.json
```diff
@@ -521,10 +521,10 @@
           "__compat": {
             "support": {
               "chrome": {
-                "version_added": "76"
+                "version_added": "77"
```
foolip, line 524: 77 as opposed to 76 confirmed by https://storage.googleapis.com/chromium-find-releases-static/0fc.html#0fc50ca2d3c81f74f73aae8aab6079a516d9b8fb

76 came from https://github.com/mdn/browser-compat-data/pull/5442 and also did the same for `img`, so there's more data to fix. I've commented in https://github.com/mdn/browser-compat-data/pull/5442#issuecomment-742852922

foolip (approved):

PR #7592, opened 2020-12-06
Add missing features for HTMLInputElement

vinyldarkscratch:

Spec: https://html.spec.whatwg.org/multipage/semantics.html#htmlinputelement

Notice: the diff looks pretty weird for this one due to the similarities in the results between all of the features.  This PR doesn't modifying any existing features, however.


foolip:
This PR is huge, and will also require checking that reflected IDL attributes have their content attribute data updated. @vinyldarkscratch how about splitting this into smaller pieces, where the diff of each piece becomes more managable?

api/HTMLInputElement.json
```diff
@@ -95,33 +94,32 @@
           }
         }
       },
-      "autofocus": {
+      "align": {
```
foolip, line 97: `align` is listed in https://html.spec.whatwg.org/multipage/obsolete.html#HTMLInputElement-partial and should be deprecated. (But I can't make suggestions in that part of the diff.)

api/HTMLInputElement.json
```diff
@@ -1155,60 +2625,59 @@
...
-              "version_added": true
+              "version_added": "1"
             }
           },
           "status": {
-            "deprecated": false,
             "experimental": false,
-            "standard_track": true
+            "standard_track": true,
+            "deprecated": false
```
foolip, line 2671:
```suggestion
            "deprecated": true
```

https://html.spec.whatwg.org/multipage/obsolete.html#HTMLInputElement-partial

api/HTMLInputElement.json
```diff
@@ -191,47 +188,44 @@
           }
         }
       },
-      "files": {
+      "autocapitalize": {
```
foolip, line 191: `autocapitalize` shouldn't be added: https://github.com/mdn/browser-compat-data/issues/7519

foolip (commented):
Took another look, but this is still very large and difficult to review. Splitting it into smaller bits in a way that avoids moves becoming apparent changes would really help.

vinyldarkscratch:
I separated this PR into three other PRs, see #7647, #7648, and #7649.

PR #7595, opened 2020-12-06
Add missing features for link/script HTML elements

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from a spec in Editor's Draft or more and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6), for the HTMLLinkElement and HTMLScriptElement APIs.

Specs:
https://html.spec.whatwg.org/multipage/semantics.html#htmllinkelement
https://html.spec.whatwg.org/multipage/semantics.html#htmlscriptelement

Edit: this PR also updates some attributes within the HTML element BCD to maintain consistency.

### DIFF
--- a/api/HTMLLinkElement.json
+++ b/api/HTMLLinkElement.json
@@ -347,7 +347,7 @@
               "version_added": "78"
             },
             "firefox_android": {
-              "version_added": "78"
+              "version_added": "79"
             },
             "ie": {
               "version_added": false
@@ -394,7 +394,7 @@
               "version_added": "78"
             },
             "firefox_android": {
-              "version_added": "78"
+              "version_added": "79"
             },
             "ie": {
               "version_added": false
--- a/api/HTMLLinkElement.json
+++ b/api/HTMLLinkElement.json
@@ -95,6 +95,53 @@
           }
         }
       },
+      "charset": {
+        "__compat": {
+          "support": {
+            "chrome": {
+              "version_added": "1"
+            },
+            "chrome_android": {
+              "version_added": "18"
+            },
+            "edge": {
+              "version_added": "12"
+            },
+            "firefox": {
+              "version_added": "1"
+            },
+            "firefox_android": {
+              "version_added": "4"
+            },
+            "ie": {
+              "version_added": "≤6"
+            },
+            "opera": {
+              "version_added": "≤12.1"
+            },
+            "opera_android": {
+              "version_added": "≤12.1"
+            },
+            "safari": {
+              "version_added": "≤4"
+            },
+            "safari_ios": {
+              "version_added": "≤3"
+            },
+            "samsunginternet_android": {
+              "version_added": "1.0"
+            },
+            "webview_android": {
+              "version_added": "1"
+            }
+          },
+          "status": {
+            "experimental": false,
+            "standard_track": true,
+            "deprecated": false
+          }
+        }
+      },
       "crossOrigin": {
         "__compat": {
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLLinkElement/crossOrigin",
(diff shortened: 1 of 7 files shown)


api/HTMLLinkElement.json
```diff
@@ -95,6 +95,53 @@
...
+              "version_added": "1.0"
+            },
+            "webview_android": {
+              "version_added": "1"
+            }
+          },
+          "status": {
+            "experimental": false,
+            "standard_track": true,
+            "deprecated": false
```
foolip, line 141:
```suggestion
            "deprecated": true
```

https://html.spec.whatwg.org/multipage/obsolete.html#HTMLLinkElement-partial

api/HTMLLinkElement.json
```diff
@@ -287,6 +663,53 @@
...
```
foolip, line 709:
```suggestion
            "deprecated": true
```

https://html.spec.whatwg.org/multipage/obsolete.html#HTMLLinkElement-partial

api/HTMLLinkElement.json
```diff
@@ -340,6 +763,100 @@
...
```
foolip, line 810:
```suggestion
            "deprecated": true
```

https://html.spec.whatwg.org/multipage/obsolete.html#HTMLLinkElement-partial

api/HTMLLinkElement.json
```diff
@@ -95,6 +95,53 @@
           }
         }
       },
+      "charset": {
```
foolip, line 98: Content attribute needs updating too.

api/HTMLLinkElement.json
```diff
@@ -143,6 +190,335 @@
           }
         }
       },
+      "disabled": {
```
foolip, line 193: Content attribute needs updating too, and there are notes there that maybe should be replicated here.

api/HTMLLinkElement.json
```diff
@@ -143,6 +190,335 @@
...
+            }
+          },
+          "status": {
+            "experimental": false,
+            "standard_track": true,
+            "deprecated": false
+          }
+        }
+      },
+      "href": {
```
foolip, line 240: Content attribute needs updating too.

api/HTMLLinkElement.json
```diff
@@ -143,6 +190,335 @@
...
+            }
+          },
+          "status": {
+            "experimental": false,
+            "standard_track": true,
+            "deprecated": false
+          }
+        }
+      },
+      "hreflang": {
```
foolip, line 287: Content attribute needs updating too.

api/HTMLLinkElement.json
```diff
@@ -143,6 +190,335 @@
...
+            }
+          },
+          "status": {
+            "experimental": false,
+            "standard_track": true,
+            "deprecated": false
+          }
+        }
+      },
+      "imageSizes": {
```
foolip, line 334: Content attribute data is missing for this, needs to be added.

api/HTMLLinkElement.json
```diff
@@ -143,6 +190,335 @@
...
+            }
+          },
+          "status": {
+            "experimental": false,
+            "standard_track": true,
+            "deprecated": false
+          }
+        }
+      },
+      "imageSrcset": {
```
foolip, line 381: Also missing the content attribute.

api/HTMLLinkElement.json
```diff
@@ -143,6 +190,335 @@
...
+            }
+          },
+          "status": {
+            "experimental": false,
+            "standard_track": true,
+            "deprecated": false
+          }
+        }
+      },
+      "integrity": {
```
foolip, line 428: Content attribute needs updating too.

api/HTMLLinkElement.json
```diff
@@ -143,6 +190,335 @@
...
+            }
+          },
+          "status": {
+            "experimental": false,
+            "standard_track": true,
+            "deprecated": false
+          }
+        }
+      },
+      "media": {
```
foolip, line 475: Content attribute needs updating too.

api/HTMLLinkElement.json
```diff
@@ -287,6 +663,53 @@
           }
         }
       },
+      "rev": {
```
foolip, line 666: Content attribute needs updating too.

api/HTMLLinkElement.json
```diff
@@ -340,6 +763,100 @@
             "deprecated": false
           }
         }
+      },
+      "target": {
```
foolip, line 767: Content attribute needs updating too.

api/HTMLLinkElement.json
```diff
@@ -340,6 +763,100 @@
...
+            }
+          },
+          "status": {
+            "experimental": false,
+            "standard_track": true,
+            "deprecated": false
+          }
+        }
+      },
+      "type": {
```
foolip, line 814: Content attribute needs updating too.

html/elements/link.json
```diff
@@ -262,25 +263,119 @@
...
+              }
+            },
+            "status": {
+              "experimental": false,
+              "standard_track": true,
+              "deprecated": false
+            }
+          }
+        },
+        "imageSizes": {
```
foolip, line 294:
```suggestion
        "imagesizes": {
```

html/elements/link.json
```diff
@@ -262,25 +263,119 @@
...
+              }
+            },
+            "status": {
+              "experimental": false,
+              "standard_track": true,
+              "deprecated": false
+            }
+          }
+        },
+        "imageSrcset": {
```
foolip, line 341:
```suggestion
        "imagesrcset": {
```

api/HTMLScriptElement.json
```diff
@@ -330,6 +330,54 @@
...
+          "support": {
+            "chrome": {
+              "version_added": "45"
+            },
+            "chrome_android": {
+              "version_added": "45"
+            },
+            "edge": {
+              "version_added": "17",
+              "notes": "EdgeHTML did not pass all <a href='https://w3c-test.org/subresource-integrity/subresource-integrity.html'>W3C tests</a> for this property."
```
foolip, line 344: I think this is a new note. Similar to in https://github.com/mdn/browser-compat-data/pull/7233, I don't think it's great to link to tests directly, when interpreting what the results mean takes a lot of knowledge.

https://wpt.fyi/results/subresource-integrity/subresource-integrity.html?product=edge-18 shows the results of the last Edge run we had of WPT. I'm not sure what the results mean in terms of what a web developer needs to know. I'd stick with `partial_implementation` unless it can be summarized what did and did not work.

html/elements/script.json
```diff
@@ -210,7 +210,7 @@
               },
               "edge": {
                 "version_added": "17",
-                "partial_implementation": true
+                "notes": "EdgeHTML did not pass all <a href='https://w3c-test.org/subresource-integrity/subresource-integrity.html'>W3C tests</a> for this attribute."
```
foolip, line 213: See previous note on linking WPT results.

api/HTMLLinkElement.json
```diff
@@ -143,6 +190,342 @@
...
+              "version_added": "1",
+              "notes": "In Chrome and other Blink-based browsers, adding the <code>disabled</code> attribute using JavaScript does not remove the stylesheet from <code>document.styleSheets</code>."
+            },
+            "chrome_android": {
+              "version_added": "18",
+              "notes": "In Chrome and other Blink-based browsers, adding the <code>disabled</code> attribute using JavaScript does not remove the stylesheet from <code>document.styleSheets</code>."
+            },
+            "edge": {
+              "version_added": "12",
+              "notes": "Since Edge 79, adding the <code>disabled</code> attribute using JavaScript does not remove the stylesheet from <code>document.styleSheets</code>."
```
foolip, line 206: Did you check if this works in Edge 18? I think the correct per-spec behavior for `disabled` was only implemented by Firefox, and suspect Edge 18 might have had the same problem as Chromium here.

html/elements/link.json
```diff
@@ -147,45 +147,46 @@
...
               },
               "chrome_android": {
-                "version_added": true,
+                "version_added": "18",
                 "notes": "In Chrome and other Blink-based browsers, adding the <code>disabled</code> attribute using JavaScript does not remove the stylesheet from <code>document.styleSheets</code>."
               },
               "edge": {
-                "version_added": "12"
+                "version_added": "12",
+                "notes": "Since Edge 79, adding the <code>disabled</code> attribute using JavaScript does not remove the stylesheet from <code>document.styleSheets</code>."
```
foolip, line 159: See previous comment on testing `disabled` in Edge 18.

api/HTMLScriptElement.json
vinyldarkscratch: What do you recommend for a note, then?  If we're setting `partial_implementation`, we _must_ include a note explaining why (https://github.com/mdn/browser-compat-data/blob/master/docs/data-guidelines.md#partial_implementation-requires-a-note).

foolip: I don't know what the test failures mean for a web developer, and that's really the problem. Without understanding what's missing from the implementation, setting `partial_implementation` isn't really possible.

If there's a way to avoid figuring this out in this PR that would be great. Otherwise, can you try to dig into the tests and understand what kind of code works differently in EdgeHTML compared to spec and other engines? It could be inconsequential stuff that isn't worth mentioning to web developers, or it could be that the feature doesn't work at all in common cases, I can't tell from a quick glance.

foolip:
@vinyldarkscratch I see you pushed some fixes, but is this ready for review again? Looks like there are issues left to resolve?

api/HTMLScriptElement.json
vinyldarkscratch: I can't tell either, I don't know enough about this feature to really say.  Removing the note so we can get this in.

api/HTMLLinkElement.json
vinyldarkscratch: Yep, looks like this works correctly in Edge 18, so the behavior started in Edge 79.

Test Code:
```html
<!DOCTYPE html>
<html>
	<head>
		<meta content="width=device-width">
		<style id="persistent-style">
			/* Persistent styling */
			body {
				background-color: black;
				color: white;
				margin: 0;
				padding: 0;
			}
			canvas {
				background-color: white;
			}
		</style>
		<style id="test-style">
			@media (max-width: 600px) {
				body {
					background-color: #111;
					color: white;
				}
			}
		</style>
		<link rel="stylesheet" href="https://fonts.googleapis.com/css2?family=Roboto&display=swap" disabled />
	</head>
	<body>
		<div id="test"></div>

		<script>
			alert(document.styleSheets.length + " / 3");
		</script>
	</body>
</html>
```

foolip (approved):

PR #7597, opened 2020-12-06
Add HTMLMediaElement.getStartDate API (#7597)

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from a spec in Editor's Draft or more and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6), for the HTMLMediaElement API.

Spec: https://html.spec.whatwg.org/multipage/semantics.html#htmlmediaelement


api/HTMLMediaElement.json
```diff
@@ -47,6 +47,429 @@
           "deprecated": false
         }
       },
+      "HAVE_CURRENT_DATA": {
```
foolip, line 50: These are constants and should be reverted per discussion in https://github.com/mdn/browser-compat-data/issues/7585.

api/HTMLMediaElement.json
```diff
@@ -1392,6 +1815,53 @@
...
+              "version_added": false
+            },
+            "opera": {
+              "version_added": false
+            },
+            "opera_android": {
+              "version_added": false
+            },
+            "safari": {
+              "version_added": "9"
```
foolip, line 1846: Oh neat, didn't realize Safari had implemented this.

foolip (approved):

PR #7598, opened 2020-12-06
Add missing features for HTMLMetaElement API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from a spec in Editor's Draft or more and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6), for the HTMLMetaElement API.

Spec: https://html.spec.whatwg.org/multipage/semantics.html#htmlmetaelement

### DIFF
--- a/api/HTMLMetaElement.json
+++ b/api/HTMLMetaElement.json
@@ -231,7 +231,7 @@
           "status": {
             "experimental": false,
             "standard_track": true,
-            "deprecated": false
+            "deprecated": true
           }
         }
       }
--- a/api/HTMLMetaElement.json
+++ b/api/HTMLMetaElement.json
@@ -46,6 +46,194 @@
           "standard_track": true,
           "deprecated": false
         }
+      },
+      "content": {
+        "__compat": {
+          "support": {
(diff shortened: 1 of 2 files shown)


api/HTMLMetaElement.json
```diff
@@ -46,6 +46,194 @@
...
+              "version_added": "1.0"
+            },
+            "webview_android": {
+              "version_added": "1"
+            }
+          },
+          "status": {
+            "experimental": false,
+            "standard_track": true,
+            "deprecated": false
```
foolip, line 234:
```suggestion
            "deprecated": true
```

https://html.spec.whatwg.org/multipage/obsolete.html#HTMLMetaElement-partial

foolip (commented):
Can you sync the content attribute data too?

foolip:
> Can you sync the content attribute data too?

Since the PR was updated, just noting this is still needed.

foolip (approved):

PR #7599, opened 2020-12-06
Add missing Option constructor of HTMLOptionElement

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from a spec in Editor's Draft or more and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6), for the HTMLOptionElement API.

Spec: https://html.spec.whatwg.org/multipage/semantics.html#htmloptionelement

### DIFF
--- a/api/HTMLOptionElement.json
+++ b/api/HTMLOptionElement.json
@@ -47,6 +47,53 @@
           "deprecated": false
         }
       },
+      "Option": {
+        "__compat": {
+          "support": {
+            "chrome": {
+              "version_added": "1"
+            },
+            "chrome_android": {
+              "version_added": "18"
+            },
+            "edge": {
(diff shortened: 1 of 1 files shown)


api/HTMLOptionElement.json
```diff
@@ -47,6 +47,53 @@
           "deprecated": false
         }
       },
+      "Option": {
+        "__compat": {
```
foolip, line 51: Can you link to https://developer.mozilla.org/en-US/docs/Web/API/HTMLOptionElement/Option and use the description `"<code>Option()</code> constructor"` for this?

foolip (approved):

PR #7600, opened 2020-12-06
Add missing "integrity" attribute of HTMLScriptElement

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from a spec in Editor's Draft or more and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6), for the HTMLScriptElement API.

Spec: https://html.spec.whatwg.org/multipage/semantics.html#htmlscriptelement

### DIFF
--- a/api/HTMLScriptElement.json
+++ b/api/HTMLScriptElement.json
@@ -330,6 +330,53 @@
           }
         }
       },
+      "integrity": {
+        "__compat": {
+          "support": {
+            "chrome": {
+              "version_added": "45"
+            },
+            "chrome_android": {
+              "version_added": "45"
+            },
+            "edge": {
+              "version_added": "17"
+            },
+            "firefox": {
+              "version_added": "43"
+            },
+            "firefox_android": {
+              "version_added": "43"
+            },
+            "ie": {
+              "version_added": false
+            },
+            "opera": {
+              "version_added": "32"
+            },
+            "opera_android": {
+              "version_added": "32"
+            },
+            "safari": {
+              "version_added": "11.1"
+            },
+            "safari_ios": {
+              "version_added": "11.3"
+            },
+            "samsunginternet_android": {
+              "version_added": "5.0"
+            },
+            "webview_android": {
+              "version_added": "43"
+            }
+          },
+          "status": {
+            "experimental": false,
+            "standard_track": true,
+            "deprecated": false
+          }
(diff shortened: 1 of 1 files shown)


foolip:
The content attribute data already exists:
https://github.com/mdn/browser-compat-data/blob/28ff6dd19e3f3a485c9eccf875f113325c09cf70/html/elements/script.json#L202-L249

api/HTMLScriptElement.json
```diff
@@ -330,6 +330,53 @@
...
+        "__compat": {
+          "support": {
+            "chrome": {
+              "version_added": "45"
+            },
+            "chrome_android": {
+              "version_added": "45"
+            },
+            "edge": {
+              "version_added": "17"
```
foolip, line 343: The content attribute entry has `partial_implementation` for Edge. Can you check why that is and whether we should have the same here? It should almost certainly only be for Edge 17-18 though.

api/HTMLScriptElement.json
```diff
@@ -330,6 +330,53 @@
...
+              "version_added": "32"
+            },
+            "opera_android": {
+              "version_added": "32"
+            },
+            "safari": {
+              "version_added": "11.1"
+            },
+            "safari_ios": {
+              "version_added": "11.3"
```
foolip, line 364: The Safari iOS data for the content attribute is false. I'm guessing that's just wrong, can you update that and other bits of html/elements/script.json to match?

Because it was all added in https://trac.webkit.org/changeset/216347/webkit, we can be confident the ranges should be the same.

foolip (commented):
Since they were all added at once, can you also include `HTMLLinkElement` changes in this PR?

vinyldarkscratch:
I've merged this into #7595 to link the HTMLLinkElement and HTMLScriptElement changes together!  Will apply review comments and changes there.

api/HTMLScriptElement.json
vinyldarkscratch: It looks like `partial_implementation` is marked because some of the [W3C tests fail](https://w3c-test.org/subresource-integrity/subresource-integrity.html), and was added in #2765.  It should have been added with a note, though...

PR #7602, opened 2020-12-06
Fix version numbers for HTMLTrackElement

vinyldarkscratch:
This PR updates the version numbers using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6) for the HTMLTrackElement API.


api/HTMLTrackElement.json
```diff
@@ -75,6 +75,194 @@
           "deprecated": false
         }
       },
+      "ERROR": {
```
foolip, line 78: These are constants, and it looks like for example constants on `HTMLMediaElement` aren't in BCD. @ddbeck @Elchi3 should we represent constants in BCD? In some ways they're part of the attribute using them, in this case the `readyState` attribute, but they are actual attributes on the interface object and prototype, so unlike enums they're not invisible.

I couldn't find any open issue about this.

api/HTMLTrackElement.json
```diff
@@ -5,10 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLTrackElement",
         "support": {
           "chrome": {
-            "version_added": "23"
+            "version_added": "17"
```
foolip, line 8: @vinyldarkscratch can you break out these fixes into a separate PR?

api/HTMLTrackElement.json
foolip: Turns out it's being discussed in https://github.com/mdn/browser-compat-data/issues/7585, let's talk over there.

api/HTMLTrackElement.json
vinyldarkscratch: These changes fix consistency, so I'm not able to remove them from this PR.

foolip: Ah, OK then. (They could be broken out and landed first, but too much work.)

foolip: Oh, but actually, all that remains of this PR is now updating these version numbers. But then lots of subfeatures are left as "23" and this was all introduced together, so that must be wrong.

It's not super important to fix this data now, but if you want to, can you update more of the track-related data together in this PR?

vinyldarkscratch:
Will have to dig into this a little bit more.  Since this PR no longer adds new version numbers, I'll close it and reopen a new one.

PR #7608, opened 2020-12-07
Update Firefox Android support for browsingData.

agi:
We just added support for browsingData in Firefox for Android.

Code landed in https://bugzilla.mozilla.org/show_bug.cgi?id=1625233.

- [x] Summarize your changes

### DIFF
--- a/webextensions/api/browsingData.json
+++ b/webextensions/api/browsingData.json
@@ -15,10 +15,15 @@
                 "firefox": {
                   "version_added": "53"
                 },
-                "firefox_android": {
-                  "version_added": "56",
-                  "version_removed": "79"
-                },
+                "firefox_android": [
+                  {
+                    "version_added": "56",
+                    "version_removed": "79"
+                  },
+                  {
+                    "version_added": "85"
+                  }
+                ],
                 "opera": {
                   "version_added": true
                 },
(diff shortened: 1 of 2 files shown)


webextensions/api/browsingData.json
```diff
@@ -16,8 +16,7 @@
                   "version_added": "53"
                 },
                 "firefox_android": {
-                  "version_added": "56",
-                  "version_removed": "79"
+                  "version_added": "85"
```
ddbeck, line 19: Was the previous data here wrong? Or did it actually happen that it was implemented in Fennec and then appeared in Fenix in version 85? If it's the latter, then we should structure this data as an array of support statements:

```
[
  {
     "version_added": "85"
  },
  {
     "version_removed": "79",
     "version_added: "56"
  }
]
```

(this would apply to the other instances like this in the diff, as well)

ddbeck (commented):
Hi @agi and thanks for this PR. I had a question about the history in line comments, which might have some consequences for the structure of the data.

webextensions/api/browsingData.json
agi: It was implemented in Fennec and now we are re-implementing it in Fenix. Thanks for your comment! I updated the patch.

ddbeck (approved):
Looks good. Thank you! 🎉

PR #7610, opened 2020-12-08
Mark Application Cache features non-standard

sideshowbarker:
Per https://github.com/whatwg/html/pull/6153 and https://github.com/whatwg/html/commit/e4330d5, the Application Cache feature is no longer part of the HTML standard. So this change marks both the `applicationCache` member of `SharedWorkerGlobalScope` and the `manifest` attribute of the `html` element as `standard_track:false`.

---

I’ve also already updated the following MDN articles:

* https://developer.mozilla.org/en-US/docs/Web/API/Window/applicationCache
* https://developer.mozilla.org/en-US/docs/Web/HTML/Element/html/manifest

### DIFF
--- a/api/SharedWorkerGlobalScope.json
+++ b/api/SharedWorkerGlobalScope.json
@@ -90,8 +90,8 @@
           },
           "status": {
             "experimental": false,
-            "standard_track": true,
-            "deprecated": false
+            "standard_track": false,
+            "deprecated": true
           }
         }
       },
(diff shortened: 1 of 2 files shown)


ddbeck (approved):
Thanks, @sideshowbarker!

I'm guessing the actual data here is probably a mess. But if I ignore it a while longer, maybe we can just rip these features out. 😆

PR #7612, opened 2020-12-08
Update Opera Android 61 Current

vsn4ik:


Elchi3 (approved):
Confirming info also in https://forums.opera.com/topic/45652/opera-for-android-61

Thanks for your PR! :+1:

PR #7613, opened 2020-12-08
Rename NDEFReader.prototype.onerror to NDEFReader.prototype.onreadingerror

sideshowbarker:
This change renames `NDEFReader.prototype.onerror` to `NDEFReader.prototype.onreadingerror,` per the spec change at https://github.com/w3c/web-nfc/commit/a71bff9 (https://github.com/w3c/web-nfc/pull/601).

---

No browsers yet implement this feature, so alternatively, we could just remove it from BCD.

I’ve already also updated MDN for this, to make https://developer.mozilla.org/docs/Web/API/NDEFReader/onerror a redirect to https://developer.mozilla.org//docs/Web/API/NDEFReader/onreadingerror

### DIFF
--- a/api/NDEFReader.json
+++ b/api/NDEFReader.json
@@ -96,9 +96,9 @@
           }
         }
       },
-      "onerror": {
+      "onreadingerror": {
         "__compat": {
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Looks good. Thank you! 🎉

PR #7614, opened 2020-12-08
Document 'text/html' support in Clipboard API for Chrome

pbadenski:
Chrome added support for 'text/html' MIME type in Chrome 86.

Source: https://www.chromestatus.com/feature/5357049665814528

### DIFF
--- a/api/Clipboard.json
+++ b/api/Clipboard.json
@@ -53,6 +53,11 @@
           "support": {
             "chrome": [
               {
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
I see this is your first PR to the project. Welcome @pbadenski!

This looks fine to me, though it's weird data to begin with—I've opened #7991 to improve this later.

Thank you! 🎉

PR #7616, opened 2020-12-08
Drop OverconstrainedError message and name members

sideshowbarker:
This change removes the `message` and `name` members from the data for the `OverconstrainedError` interface, per the spec change at https://github.com/w3c/mediacapture-main/pull/746, which redefined `OverconstrainedError` as as extending `DOMException` (which has `message` and name `members` that `OverconstrainedError` inherits).

---

I’ve already redirected the relevant MDN articles to the equivalent `DOMException` articles.


ddbeck (approved):
Thank you! 🎉

PR #7617, opened 2020-12-08
Mark OverconstrainedErrorEvent non-standard

sideshowbarker:
This change marks `OverconstrainedErrorEvent` as `standard_track:false` and `deprecated:true`.

It was implemented only on WebKit, and is no longer part of any spec; https://github.com/w3c/mediacapture-main/issues/573 removed it.


ddbeck (approved):
Looks good. Thank you! 🎉

PR #7618, opened 2020-12-08
Edge 87 adds support for text-[decoration-thickness,underline-offset]

36degrees:
Also add support for `from-font` in `text-underline-position`.

The same changes that were made for Chrome 87 in 116a3bd (#7395) also apply to Edge 87.

Verified by testing in Edge 86 and 87 on BrowserStack, with the CSS:

```
a {
  text-decoration-color: red;
  text-decoration-thickness: 3px;
  text-underline-offset: 3px;
}
```

In Edge 86, the underline is red but only the default user agent thickness, and not offset. Developer tools also flags the `text-decoration-thickness` and `text-underline-offset` properties as invalid.

Edge 86 with text-decoration-thickness and text-underline-offset not having any effect

In Edge 87, it has 3px thickness and is offset, and has the same appearance that it does in Chrome 87)

Edge 86 with text-decoration-thickness and text-underline-offset working as expected

- [x] Summarize your changes

### DIFF
--- a/css/properties/text-decoration-thickness.json
+++ b/css/properties/text-decoration-thickness.json
@@ -12,7 +12,7 @@
               "version_added": "87"
             },
             "edge": {
-              "version_added": false
+              "version_added": "87"
             },
             "firefox": [
               {
--- a/css/properties/text-underline-offset.json
+++ b/css/properties/text-underline-offset.json
@@ -12,7 +12,7 @@
               "version_added": "87"
             },
             "edge": {
-              "version_added": false
+              "version_added": "87"
             },
             "firefox": [
               {
--- a/css/properties/text-underline-position.json
+++ b/css/properties/text-underline-position.json
@@ -163,7 +163,7 @@
                 "version_added": "87"
               },
               "edge": {
-                "version_added": false
+                "version_added": "87"
               },
               "firefox": {
                 "version_added": "74"


36degrees:
It looks like there's [a bug with the implementation in Chrome 87](https://bugs.chromium.org/p/chromium/issues/detail?id=1154537&q=text-decoration-thickness&can=1) (and therefore also Edge 87) where `text-decoration-thickness` is not applied unless either:

- `text-underline-offset` is set to something other than `auto` OR 
- `text-decoration-color` is set to something other than `currentColor`

Is this sort of thing worth adding a note for? I'm not clear on how fixes for bugs like this land, and whether it'll be resolved before Chrome / Edge 88.

ddbeck (approved):
Hi @36degrees. Thank you for this PR and welcome to BCD!

Since the immediate change here looks good to me, I'm going to approve this and merge it, so it'll make today's release of BCD.

To your question about the bug, there's two routes we could take here:

One option, if it looks like a fix is coming along shortly, is to wait until we know what version is going to ship the bug fix and add a note in a follow up PR (e.g., `"notes": "Before version 89, …"`).

The other option is to open an issue, add the note in a PR (without the "Before version" text but with a link to the Chromium bug), and then follow up with a PR later to revise the note to reflect the fix.

I hope that helps. Thank you again for this PR!

PR #7625, opened 2020-12-10
Note an issue with text-decoration-thickness in Chrome, Edge

36degrees:
Following on from https://github.com/mdn/browser-compat-data/pull/7618#issuecomment-741643427.

The `text-decoration-thickness` property does not work unless either `text-underline-offset` is set to something other than `auto` or `text-decoration-color` is set to something other than `currentColor`.

See https://crbug.com/1154537.

Still seeing this in Chrome 88 / Edge 88 betas, but Chrome Canary 89 does not have this issue. However, I’m not sure exactly how the Chrome / Edge release process works and so am not sure if this fix might end up in Chrome 88 as well, so not able to provide a bounded version range until Chrome 88 is stable.

- [x] Summarize your changes

### DIFF
--- a/css/properties/text-decoration-thickness.json
+++ b/css/properties/text-decoration-thickness.json
@@ -6,13 +6,16 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/text-decoration-thickness",
           "support": {
             "chrome": {
-              "version_added": "87"
+              "version_added": "87",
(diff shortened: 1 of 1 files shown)


ddbeck (changes requested):
Thanks for opening this follow up, @36degrees! One suggestion here: webview_android is also based on Chromium. Can we apply the note to that engine too?

36degrees:
@ddbeck – certainly, good catch. I've updated the commit to add the same note to webview_android.

ddbeck (approved):
Thank you! 🎉

PR #7627, opened 2020-12-10
Correct data errors in certain CSS interfaces.

jpmedley:
The corrects a few errors with some CSS interfaces.

None of them have `@@iterator` which must be [present in the the IDL](https://heycam.github.io/webidl/#es-iterator). They are in neither [the spec](https://drafts.css-houdini.org/css-typed-om/), nor Chrome's implemention of it:

* [CSSNumericArray](https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/core/css/cssom/css_numeric_array.idl)
* [CSSTransformValue](https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/core/css/cssom/css_transform_value.idl)
* [CSSUnparsedValue](https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/core/css/cssom/css_unparsed_value.idl)

The setters and getters listed in the IDL were misinterpreted as functions. For example, instead of there being a method named `CSSNumericArray.CSSNumbericValue` there is actually the ability to treat an instance of the interface as an array. In other words, `CSSNumbericArray[n]` would be a valid programming reference.

### DIFF
--- a/api/CSSNumericArray.json
+++ b/api/CSSNumericArray.json
@@ -47,54 +47,6 @@
           "deprecated": false
         }
       },
-      "CSSNumericValue": {
-        "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/CSSNumericArray/CSSNumericValue",
-          "support": {
-            "chrome": {
-              "version_added": "66"
-            },
-            "chrome_android": {
-              "version_added": "66"
-            },
-            "edge": {
-              "version_added": "79"
-            },
-            "firefox": {
-              "version_added": false
-            },
-            "firefox_android": {
-              "version_added": false
-            },
-            "ie": {
-              "version_added": false
-            },
-            "opera": {
-              "version_added": "53"
-            },
-            "opera_android": {
-              "version_added": "47"
-            },
-            "safari": {
-              "version_added": false
-            },
-            "safari_ios": {
-              "version_added": false
-            },
-            "samsunginternet_android": {
-              "version_added": "9.0"
-            },
-            "webview_android": {
-              "version_added": "66"
-            }
-          },
-          "status": {
-            "experimental": true,
-            "standard_track": true,
-            "deprecated": false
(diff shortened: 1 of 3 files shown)


jpmedley:
cc: @rachelandrew

ddbeck:
> None of them have `@@iterator` which must be [present in the the IDL](https://heycam.github.io/webidl/#es-iterator).

Do [iterable declarations](https://heycam.github.io/webidl/#idl-iterable) not imply the `@@iterator` symbol? It looks like all three have the declaration.

This possibly relates to #6367.

foolip:
> Do [iterable declarations](https://heycam.github.io/webidl/#idl-iterable) not imply the `@@iterator` symbol? It looks like all three have the declaration.

These three interfaces all have `iterable<T>` declarations in https://drafts.css-houdini.org/css-typed-om/, and https://heycam.github.io/webidl/#es-iterable is what says that results in a `@@iterator` property being added. (A bunch of other things are also implied, and it's worth knowing that which are different for `iterable<T>` and `iterable<K, V>`.)

ddbeck:
Hi @jpmedley, do you plan to come back to this PR?

jpmedley:
I shouldn't have removed `@@iterable`. The reason I thought it should be removed is because I'm still confused about. It doesn't look to me like you ever write this in a script. What you actually write is `instanceVariable[Symbol.ite.rator]()`.

ddbeck:
@jpmedley If you want to change the way BCD names the iterator symbol, then I think we need to take that up under a separate issue. The `@@` notation for well-known symbols seems conventional within BCD and on MDN (see [`Array`'s `@@iterator`](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/@@iterator) for an example); if we're going to change one, we ought to be consistent about it.

PR #7628, opened 2020-12-10
Update Intl.ListFormat for Safari 14.1

ezfe:
Safari 14.1 adds support for these operations. 14.1 is still in testing with Safari Technology Preview. I've added 14.1 as the version added for Safari on Desktop.

I tested this by running the MDN code examples for each section with Safari Technology Preview 117.

Webkit Release Notes: https://webkit.org/blog/11364/release-notes-for-safari-technology-preview-117/

I've never made a PR here before so I hope this is all good.

- [x] Summarize your changes

### DIFF
--- a/browsers/safari.json
+++ b/browsers/safari.json
@@ -173,6 +173,11 @@
           "status": "current",
           "engine": "WebKit",
           "engine_version": "610.1.28"
+        },
+        "14.1": {
+          "release_notes": "https://developer.apple.com/safari/technology-preview/release-notes/"
+          "status": "beta",
+          "engine": "WebKit"
         }
       }
     }
--- a/browsers/safari.json
+++ b/browsers/safari.json
@@ -175,7 +175,7 @@
           "engine_version": "610.1.28"
         },
         "14.1": {
-          "release_notes": "https://developer.apple.com/safari/technology-preview/release-notes/"
+          "release_notes": "https://developer.apple.com/safari/technology-preview/release-notes/",
           "status": "beta",
           "engine": "WebKit"
         }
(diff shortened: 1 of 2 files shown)


browsers/safari.json
```diff
@@ -173,6 +173,11 @@
           "status": "current",
           "engine": "WebKit",
           "engine_version": "610.1.28"
+        },
+        "14.1": {
+          "release_notes": "https://developer.apple.com/safari/technology-preview/release-notes/",
```
sideshowbarker, line 178: We can’t use the TP release notes URL here. Instead we need to wait until the 14.1 beta is released. And I think that’s not going to happen until several weeks from now  Or months, really.

sideshowbarker (changes requested):
Thanks much for raising this PR. It’s great to have this, but we can’t merge it yet. For one thing, the change to the `browsers/safari.json` file should be in a separate PR. So please remove it from the branch for this PR. The changes to the `javascript/builtins/intl/ListFormat.json` look great, but we can’t merge them until after the Safari 14.1 beta release.

browsers/safari.json
sideshowbarker: > @sideshowbarker Does MDN not consider STP to be a tracked version?

For Safari, unfortunately our current policy is to not track STP versions, but instead only major releases. Since for Safari, major releases only happen twice a year, that’s not super optimal. But anyway, that’s the state of things.

> If I need to wait until there's a version shipping as beta I'll wait until then to revisit this PR of course.

Yeah I think that in this case we unfortunately need to wait

ezfe: Sounds good - no worries.

PR #7630, opened 2020-12-11
Add DOMRectList API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from a spec in Editor's Draft or more and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6), for the DOMRectList API.

Spec: https://drafts.fxtf.org/geometry/#DOMRectList

### DIFF
--- a/api/DOMRectList.json
+++ b/api/DOMRectList.json
@@ -1,0 +1,145 @@
+{
+  "api": {
+    "DOMRectList": {
+      "__compat": {
+        "support": {
+          "chrome": {
+            "version_added": "68"
+          },
+          "chrome_android": {
+            "version_added": "68"
+          },
+          "edge": {
+            "version_added": "79"
+          },
+          "firefox": {
+            "version_added": "27"
+          },
+          "firefox_android": {
+            "version_added": "27"
+          },
+          "ie": {
+            "version_added": false
+          },
+          "opera": {
+            "version_added": "55"
+          },
+          "opera_android": {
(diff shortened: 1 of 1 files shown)


api/DOMRectList.json
```diff
@@ -0,0 +1,145 @@
+{
+  "api": {
+    "DOMRectList": {
```
foolip, line 3: `DOMRect` and `DOMRectList` are a bit special, they're what `element.getBoundingClientRect()` and `element.getClientRects()` respectively return, and those are very old APIs. But originally they returned objects called `ClientRect` and `ClientRectList`, except those interfaces weren't exposed.

So the support statements should go back at least as far as https://developer.mozilla.org/en-US/docs/Web/API/Element/getClientRects#Browser_compatibility as `alternative_name` statements.

http://software.hixie.ch/utilities/js/live-dom-viewer/saved/8745 should reveal what the name of it in each browser was. Testing it in Firefox 26 I see `ClientRect` and I suspect that's the case for other browsers too.

Can you add these older support statements and check it for consistency with DOMRect.json + DOMRectReadOnly.json as well?

vinyldarkscratch: Since the API doesn't seem to be exposed at first, I'll recreate this PR with data from a custom test.

PR #7631, opened 2020-12-11
Add ElementInternals API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from a spec in Editor's Draft or more and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6), for the ElementInternals API.

Spec: https://html.spec.whatwg.org/multipage/custom-elements.html#elementinternals


api/ElementInternals.json
```diff
@@ -0,0 +1,2213 @@
...
+            "version_added": "77"
+          }
+        },
+        "status": {
+          "experimental": false,
+          "standard_track": true,
+          "deprecated": false
+        }
+      },
+      "ariaAtomic": {
```
foolip, line 49: These aren't from the spec linked in the description, they're from https://w3c.github.io/aria/#ARIAMixin.

Can you split them out into a separate PR and add them to Element.json at the same time?

Chrome 81 as the version is confirmed by https://storage.googleapis.com/chromium-find-releases-static/dbe.html#dbef8d4838807211e2940f0929c8109cdf06b9d5

vinyldarkscratch: Ack, thought I excluded these.  Removing now!

foolip (approved):

PR #7633, opened 2020-12-11
Add EventCounts API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from a spec in Editor's Draft or more and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6), for the EventCounts API.

Spec: https://wicg.github.io/event-timing/
IDL: https://github.com/w3c/webref/blob/master/ed/idl/event-timing.idl

### DIFF
--- a/api/EventCounts.json
+++ b/api/EventCounts.json
@@ -41,8 +41,8 @@
           }
         },
         "status": {
-          "experimental": false,
-          "standard_track": true,
+          "experimental": true,
+          "standard_track": false,
           "deprecated": false
         }
       },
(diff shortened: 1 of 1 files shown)


api/EventCounts.json
```diff
@@ -0,0 +1,380 @@
...
+          "samsunginternet_android": {
+            "version_added": false
+          },
+          "webview_android": {
+            "version_added": "85"
+          }
+        },
+        "status": {
+          "experimental": false,
+          "standard_track": true,
```
foolip, line 45: For some previous PRs I pushed changes to change this to false for WICG specs. Can you do that for this and other open PRs where the spec URL given starts with https://wicg.github.io/?

foolip (approved):

PR #7638, opened 2020-12-11
Add RTCError API and update RTCErrorEvent API data

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from a spec in Editor's Draft or more and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6), for the RTCError API.

Spec: https://w3c.github.io/webrtc-pc/
IDL: https://github.com/w3c/webref/blob/master/ed/idl/webrtc.idl

Edit: this also updates the data for RTCErrorEvent.

### DIFF
--- a/api/RTCError.json
+++ b/api/RTCError.json
@@ -1,0 +1,333 @@
+{
+  "api": {
+    "RTCError": {
+      "__compat": {
+        "support": {
+          "chrome": {
+            "version_added": "74"
+          },
+          "chrome_android": {
+            "version_added": "74"
+          },
+          "edge": {
+            "version_added": "79"
+          },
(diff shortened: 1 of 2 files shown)


foolip:
Could you update RTCErrorEvent.json together with this? That has a lot of null data, but the two should go together. I fully expect, however, that they weren't exposed at the same time, and that we'll need custom tests to properly test for one or both of them.

foolip (approved):
Thanks, now it's clear that these things go together!

PR #7639, opened 2020-12-11
Add StyleMedia API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from a spec ~~in Editor's Draft or more~~ and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6), for the StyleMedia API.

Spec (deprecated): https://web.archive.org/web/20100206142043/http://dev.w3.org/csswg/cssom-view#the-stylemedia-interface
IDL: https://github.com/foolip/mdn-bcd-collector/blob/main/custom-idl/cssom-view.idl

StyleMedia was originally named Media, and was removed from the spec shortly after it was renamed to StyleMedia.

### DIFF
--- a/api/StyleMedia.json
+++ b/api/StyleMedia.json
@@ -43,8 +43,8 @@
         },
         "status": {
           "experimental": false,
-          "standard_track": true,
-          "deprecated": false
+          "standard_track": false,
+          "deprecated": true
         }
       },
       "matchMedium": {
@@ -90,8 +90,8 @@
           },
           "status": {
             "experimental": false,
-            "standard_track": true,
-            "deprecated": false
+            "standard_track": false,
+            "deprecated": true
           }
         }
       },
@@ -138,8 +138,8 @@
           },
           "status": {
             "experimental": false,
-            "standard_track": true,
-            "deprecated": false
+            "standard_track": false,
+            "deprecated": true
           }
         }
       }
--- a/api/StyleMedia.json
+++ b/api/StyleMedia.json
@@ -1,0 +1,148 @@
+{
+  "api": {
+    "StyleMedia": {
+      "__compat": {
+        "support": {
+          "chrome": {
+            "version_added": false
+          },
+          "chrome_android": {
+            "version_added": false
+          },
+          "edge": {
+            "version_added": "12",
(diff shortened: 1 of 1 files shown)


foolip:
This should be added together with an entry for `window.styleMedia`, which BCD is missing. Chrome does in fact support this, it's just that the `StyleMedia` interface is not exposed. Other than that, the relationship between `styleMedia` and `StyleMedia` is like that between `navigator` and `Navigator`. The interface object itself is more or less irrelevant.

api/StyleMedia.json
```diff
@@ -0,0 +1,148 @@
...
+          "samsunginternet_android": {
+            "version_added": false
+          },
+          "webview_android": {
+            "version_added": false
+          }
+        },
+        "status": {
+          "experimental": false,
+          "standard_track": true,
```
foolip, line 46: This isn't in https://drafts.csswg.org/cssom-view/ or on a standards track, we're getting it from our custom IDL associated with cssom-view.

foolip:
Since comments might be overlooked with so many PRs, this isn't ready, this is supported in Chrome.

vinyldarkscratch:
Gotcha -- I'll reopen this with a custom test that will be introduced in mdn-bcd-collector v1.1.7!

gsnedders:
> Gotcha -- I'll reopen this with a custom test that will be introduced in mdn-bcd-collector v1.1.7!

As far as I can tell, this never happened. (I was trying to find out if Chrome still supported this!)

queengooborg:
Sorry about that!  It looks like for some reason, this interface isn't present in mdn-bcd-collector v4.0 (the current published version), but is present in v5.0.  Once a new collector version launches, I'll be able to get this added in!

foolip:
It's here now: http://mdn-bcd-collector.appspot.com/tests/api/StyleMedia

We had a mishap where all of the custom tests were removed, but they're now restored in 5.0.0.

queengooborg:
I have just opened a PR to add this now!  #15000

PR #7640, opened 2020-12-11
Add TextTrackCueList

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from a spec in Editor's Draft or more and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6), for the TextTrackCueList API.

Spec: https://html.spec.whatwg.org/multipage/media.html#texttrackcuelist

### DIFF
--- a/api/TextTrackCueList.json
+++ b/api/TextTrackCueList.json
@@ -109,7 +109,7 @@
               "version_added": "31"
             },
             "firefox_android": {
-              "version_added": "32"
+              "version_added": "31"
             },
             "ie": {
               "version_added": "10"
@@ -118,7 +118,7 @@
               "version_added": "31"
             },
             "opera_android": {
-              "version_added": "31"
+              "version_added": "32"
             },
             "safari": {
               "version_added": "10"
--- a/api/TextTrackCueList.json
+++ b/api/TextTrackCueList.json
@@ -1,0 +1,145 @@
+{
+  "api": {
+    "TextTrackCueList": {
+      "__compat": {
+        "support": {
+          "chrome": {
+            "version_added": "23"
+          },
+          "chrome_android": {
+            "version_added": "25"
+          },
+          "edge": {
+            "version_added": "12"
+          },
+          "firefox": {
+            "version_added": "31"
+          },
+          "firefox_android": {
+            "version_added": "31"
+          },
+          "ie": {
+            "version_added": "10"
+          },
+          "opera": {
+            "version_added": "≤12.1"
+          },
+          "opera_android": {
+            "version_added": "≤12.1"
+          },
+          "safari": {
+            "version_added": "6"
+          },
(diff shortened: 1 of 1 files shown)


api/TextTrackCueList.json
```diff
@@ -0,0 +1,145 @@
+{
+  "api": {
+    "TextTrackCueList": {
+      "__compat": {
+        "support": {
+          "chrome": {
+            "version_added": "23"
```
foolip, line 7: This doesn't match TextTrackCue.json, in particular the `cues` entry there which is an instance of `TextTrackCueList` and these would have to have been introduced together. Perhaps a case of the interface object not being exposed?

vinyldarkscratch: It's very possible -- I confirmed the source of the data for `TextTrackCue` (#3763), and it checks out.  I'll reopen this after mdn-bcd-collector v1.1.7, which features custom tests for these APIs.

vinyldarkscratch: Hmm...it seems that the results are the same for the v1.1.7 collector.

foolip: Can you poke around manually? Was it really the case that part of the text track stuff was introduced in Chrome 9, but TextTrackCueList not until 23? The related APIs worth checking are HTMLTrackElement, HTMLMediaElement.textTracks, and TextTrack* + VTT*.

foolip: I found the WebKit changeset which introduced the interface: https://trac.webkit.org/changeset/97926/webkit

At the time interfaces had to be explicitly exposed in DOMWindow.idl, which was done for TextTrackCue but not TextTrackCueList.

I think we can then confidently say this should be supported from the same versions as `api.TextTrackCue`, `api.HTMLTrackElement` and `api.HTMLMediaElement.textTracks`. I'll check if the data agrees.

api/TextTrackCueList.json
```diff
@@ -0,0 +1,145 @@
...
+            }
+          },
+          "status": {
+            "experimental": false,
+            "standard_track": true,
+            "deprecated": false
+          }
+        }
+      },
+      "length": {
```
foolip, line 96: This should match the parent feature, I'm sure it's a testing bug that led to "44" here. I'll push a change.

foolip (approved):
I've convinced myself using http://mdn-bcd-collector.appspot.com/tests/api/TextTrackCue that this data is correct, but that there are errors in the `TextTrackCue` entry. I'll fix that.

foolip:
I've sent https://github.com/mdn/browser-compat-data/pull/8726 to fix the inconsistencies discovered while reviewing this.

PR #7644, opened 2020-12-11
Add XPathEvaluator

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from a spec in Editor's Draft or more and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6), for the XPathEvaluator API.

Spec: https://dom.spec.whatwg.org/#interface-xpathevaluator

### DIFF
--- a/api/XPathEvaluator.json
+++ b/api/XPathEvaluator.json
@@ -48,6 +48,7 @@
       },
       "XPathEvaluator": {
         "__compat": {
+          "description": "<code>XPathEvaluator()</code> constructor",
           "support": {
             "chrome": {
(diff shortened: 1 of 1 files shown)


foolip (approved):

foolip (changes requested):
Actually no, https://developer.mozilla.org/en-US/docs/Web/API/XPathEvaluator exists, so lots of URLs can be added here.

foolip (approved):
Yay!

PR #7645, opened 2020-12-11
Add missing XR APIs and API features

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from a spec in Editor's Draft or more and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6), for the XR APIs.

Specs:
https://immersive-web.github.io/anchors/
https://immersive-web.github.io/dom-overlays/
https://immersive-web.github.io/hit-test/
https://immersive-web.github.io/layers/
https://immersive-web.github.io/webxr/
https://immersive-web.github.io/webxr-ar-module/
https://immersive-web.github.io/webxr-hand-input/

### DIFF
--- a/api/XRAnchor.json
+++ b/api/XRAnchor.json
@@ -1,0 +1,145 @@
+{
+  "api": {
+    "XRAnchor": {
+      "__compat": {
+        "support": {
+          "chrome": {
+            "version_added": "85"
+          },
+          "chrome_android": {
+            "version_added": "85"
+          },
+          "edge": {
+            "version_added": "85"
+          },
+          "firefox": {
+            "version_added": false
+          },
+          "firefox_android": {
+            "version_added": false
+          },
+          "ie": {
+            "version_added": false
+          },
+          "opera": {
+            "version_added": false
+          },
+          "opera_android": {
+            "version_added": false
+          },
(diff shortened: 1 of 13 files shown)


api/XRAnchor.json
```diff
@@ -0,0 +1,145 @@
...
+          "samsunginternet_android": {
+            "version_added": false
+          },
+          "webview_android": {
+            "version_added": false
+          }
+        },
+        "status": {
+          "experimental": false,
+          "standard_track": true,
```
foolip, line 45: Looks like there's both a CG (https://www.w3.org/community/immersive-web/) and. a WG (https://www.w3.org/immersive-web/) for this, so I guess `standards_track` true is appropriate here.

But `experimental` is set to true for some of what already exists in api/XRFrame.json, but not all existing XR* stuff it seems.

Can you go over and check what's most common and make sure we set the same statuses for all of XR* unless there's nothing differentiating them? I'd default to `experimental` true and `standards_track` true if the existing pattern is not clear.

foolip (approved):

PR #7648, opened 2020-12-11
Add missing (obsolete) attributes for HTMLInputElement API

vinyldarkscratch:

Spec: https://html.spec.whatwg.org/multipage/obsolete.html#HTMLInputElement-partial

### DIFF
--- a/api/HTMLInputElement.json
+++ b/api/HTMLInputElement.json
@@ -47,53 +47,6 @@
           "deprecated": false
         }
       },
-      "align": {
-        "__compat": {
-          "support": {
-            "chrome": {
-              "version_added": "1"
-            },
-            "chrome_android": {
-              "version_added": "18"
-            },
-            "edge": {
-              "version_added": "12"
-            },
-            "firefox": {
-              "version_added": "1"
-            },
-            "firefox_android": {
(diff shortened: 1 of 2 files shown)


foolip (commented):
Can you check if the content attributes are in BCD or need adding/updating?

vinyldarkscratch:
I was planning on updating the HTML element data as a follow-up PR (as to focus on accomplishing our 90% goal).  I can do it now if you'd like, though?

api/HTMLInputElement.json
```diff
@@ -47,6 +47,53 @@
           "deprecated": false
         }
       },
+      "align": {
```
foolip, line 50: Can you add "align" to input.json? `<input align=right type=image>` confirms it does something.

api/HTMLInputElement.json
```diff
@@ -1355,6 +1402,53 @@
           }
         }
       },
+      "useMap": {
```
foolip, line 1405: Can you add "usemap" to input.json? It's been supported since the dawn of time, using the same data as for this entry would work.

foolip (approved):

PR #7651, opened 2020-12-11
Add real Chromium/Opera versions for History API

vinyldarkscratch:
This PR updates the Chromium and Opera versions for the History API based upon results from the mdn-bcd-collector project.


foolip (approved):
This is ancient stuff indeed, makes sense.

PR #7653, opened 2020-12-11
Add Chromium real values for MediaError API

vinyldarkscratch:
This PR updates the Chromium data for the MediaError API based upon results from the mdn-bcd-collector project.

### DIFF
--- a/api/MediaError.json
+++ b/api/MediaError.json
@@ -5,10 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/MediaError",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "1"
           },
           "chrome_android": {
-            "version_added": true
+            "version_added": "18"
           },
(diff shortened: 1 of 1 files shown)


api/MediaError.json
```diff
@@ -52,10 +52,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/MediaError/code",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "43"
```
foolip, line 55: This is an attribute and http://mdn-bcd-collector.appspot.com/tests/api/MediaError/code is only looking at the prototype, so this 43 is when attributes were moved to prototypes. I'm pretty sure this actually goes back to when the interface itself was introduced, which in turn should match when api.HTMLMediaElement.error` was introduced.

vinyldarkscratch:
I'll reopen this PR when collector custom tests are written!

PR #7654, opened 2020-12-11
Add Chromium real values for MediaList API

vinyldarkscratch:
This PR updates the Chromium data for the MediaList API based upon results from the mdn-bcd-collector project.

### DIFF
--- a/api/MediaList.json
+++ b/api/MediaList.json
@@ -244,10 +244,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/MediaList/mediaText",
           "support": {
             "chrome": {
-              "version_added": "44"
+              "version_added": true
             },
             "chrome_android": {
-              "version_added": "44"
+              "version_added": true
             },
             "edge": {
               "version_added": "≤18"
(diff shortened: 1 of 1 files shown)


api/MediaList.json
```diff
@@ -196,10 +196,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/MediaList/length",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "44"
```
foolip, line 199: I doubt this is right. http://mdn-bcd-collector.appspot.com/tests/api/MediaList doesn't involve an instance, and `length` is even more special than some other attributes.

A custom test could probably be added based on https://developer.mozilla.org/en-US/docs/Web/API/MediaList#Examples

api/MediaList.json
```diff
@@ -244,10 +244,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/MediaList/mediaText",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "44"
```
foolip, line 247: This is also an attribute and probably incorrect.

foolip (approved):

PR #7657, opened 2020-12-11
Add real values for Chromium for NamedNodeMap API

vinyldarkscratch:
This PR adds real values for Chromium for the NamedNodeMap API based upon results from the mdn-bcd-collector project.

### DIFF
--- a/api/NamedNodeMap.json
+++ b/api/NamedNodeMap.json
@@ -218,10 +218,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/NamedNodeMap/length",
           "support": {
             "chrome": {
-              "version_added": "44"
+              "version_added": true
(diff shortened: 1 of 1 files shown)


api/NamedNodeMap.json
```diff
@@ -218,10 +218,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/NamedNodeMap/length",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "44"
```
foolip, line 221: I don't think this is right, it's an attribute on prototype problem. http://mdn-bcd-collector.appspot.com/tests/api/NamedNodeMap might need a custom test to fix this.

foolip (approved):

PR #7659, opened 2020-12-11
Add Chromium real values for Node API

vinyldarkscratch:
This PR adds real values for Chromium for the Node API based upon results from the mdn-bcd-collector project.

### DIFF
--- a/api/Node.json
+++ b/api/Node.json
@@ -108,10 +108,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Node/baseURI",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "1"
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": "18"
             },
             "edge": {
               "version_added": "12"
@@ -126,10 +126,10 @@
               "version_added": false
             },
             "opera": {
-              "version_added": true
+              "version_added": "≤12.1"
             },
             "opera_android": {
-              "version_added": true
+              "version_added": "≤12.1"
             },
             "safari": {
               "version_added": "7"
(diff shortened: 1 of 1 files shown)


api/Node.json
```diff
@@ -489,10 +489,11 @@
               "version_added": false
             },
             "opera": {
-              "version_added": null
+              "version_added": "≤12.1",
+              "version_removed": "15"
             },
             "opera_android": {
-              "version_added": null
+              "version_added": false
```
foolip, line 496: If `node.getFeature()` was in Opera for desktop then it was almost certainly in Opera for Android too.

api/Node.json
```diff
@@ -636,10 +638,11 @@
               "version_added": "9"
             },
             "opera": {
-              "version_added": null
+              "version_added": "≤12.1",
+              "version_removed": "15"
             },
             "opera_android": {
-              "version_added": null
+              "version_added": false
```
foolip, line 645: Same issue here, `hasAttributes` was probably supported in Presto on Android.

api/Node.json
```diff
@@ -1603,7 +1616,7 @@
               "notes": "Only supported on <code>Element</code>."
             },
             "opera": {
-              "version_added": "7"
+              "version_added": "15"
```
foolip, line 1619: This can't be right, Presto definitely had `parentElement`. It might have been only on `Element`, perhaps, like IE?

vinyldarkscratch: Ack, this update slipped by -- ah well, I'll add a note and revert the version change!

api/Node.json
```diff
@@ -1603,10 +1618,12 @@
               "notes": "Only supported on <code>Element</code>."
             },
             "opera": {
-              "version_added": "7"
+              "version_added": "7",
+              "notes": "Before Opera 15, this feature was only supported on <code>Element</code>."
```
foolip, line 1622: I did some quick testing to confirm that `document.body.parentElement` works in Opera 12.16, but `Text` and `Attr` nodes didn't have `parentElement`. Those are the other node types where it most likely might have existing, so I think this was `Element` only.

foolip (approved):

PR #7661, opened 2020-12-11
Update Storage Access API availability on Chrome

### LINKED ISSUE #7510: Storage Access API data is not accurate
The Storage Access API (`document.{request|has}StorageAccess`) is not available by default in Chrome unless a [flag](chrome://flags/#storage-access-api) is set.

[Chrome bug](https://bugs.chromium.org/p/chromium/issues/detail?id=989663)
[Chrome Platform Status](https://www.chromestatus.com/feature/5612590694662144)

Not sure if this is the right format for this bug so let me know how I can help!

andybons:
This change updates the availability of the Storage Access API in Chrome

+ [Chrome bug](https://bugs.chromium.org/p/chromium/issues/detail?id=989663)
+ [Chrome Platform Status](https://www.chromestatus.com/feature/5612590694662144)

The [first commit](https://chromium.googlesource.com/chromium/src.git/+/da12cf8726dc8414f0d61624afebff46b83c8f1a) was in [version 78](https://chromium.googlesource.com/chromium/src/+/da12cf8726dc8414f0d61624afebff46b83c8f1a/chrome/VERSION), not 85, and the flag remains off by default.

I’m not sure how MS Edge’s versions work and when it was turned on by default, so that version hasn’t been altered.

Fixes #7510

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -6303,10 +6303,17 @@
               "version_added": false
             },
             "opera": {
-              "version_added": "71"
+              "version_added": "65",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "#storage-access-api",
+                  "value_to_set": "Enabled"
+                }
+              ]
             },
             "opera_android": {
-              "version_added": "60"
+              "version_added": false
             },
             "safari": {
               "version_added": "11.1"
@@ -6318,7 +6325,7 @@
               "version_added": false
             },
             "webview_android": {
-              "version_added": "85"
+              "version_added": false
             }
           },
           "status": {
@@ -10322,10 +10329,17 @@
               "version_added": false
             },
             "opera": {
-              "version_added": "71"
+              "version_added": "65",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "#storage-access-api",
+                  "value_to_set": "Enabled"
+                }
+              ]
             },
             "opera_android": {
-              "version_added": "60"
+              "version_added": false
             },
             "safari": {
               "version_added": "11.1"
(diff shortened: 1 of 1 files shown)


lucalves (approved):

ddbeck (changes requested):
Thanks for this PR @andybons. I see that it's your first to the project—welcome to BCD!

The [Edge release notes report](https://docs.microsoft.com/en-us/deployedge/microsoft-edge-relnote-stable-channel#version-85056441-august-27) that this did ship in Edge 85 (but not Chrome), which is unusual, but fine.

There's one more change I'd like to see in this PR before merging, plus a couple of optional ones:

- WebView doesn't support any of Chrome's flags. Please set that value to `false`. We typically try to deal with the Chromes and WebView together.
- (optional) In #6836, the erroneous Chrome data was mirrored to Opera. Desktop Opera should show the same flag as Chrome but with version 65 and Opera for Android should be `false` (because it doesn't expose flags).

Thank you for your help with this!

andybons:
Sorry for delay, here. Holidays in U.S. are slowing me down 🎄 

Will follow up with those changes!

andybons:
> WebView doesn't support any of Chrome's flags. Please set that value to false. We typically try to deal with the Chromes and WebView together.

Done

> (optional) In #6836, the erroneous Chrome data was mirrored to Opera. Desktop Opera should show the same flag as Chrome but with version 65 and Opera for Android should be false (because it doesn't expose flags).

Done

ddbeck (approved):
This is a really nice clean up. Thank you and happy holidays, @andybons!

PR #7667, opened 2020-12-12
firefox dropped support for HTMLCanvasElement.getContext powerPrefere…

kkota1:
https://bugzilla.mozilla.org/show_bug.cgi?id=1496459

Edit 12/20:
This was broken ~~either never supported on Windows, or dropped~~ around version ~71

Copied from bugzilla:

Steps to reproduce:

1. Start Firefox on a Windows system that has an Nvidia optimus configuration containing an integrated and discrete GPU.  (Make sure that the NVIDIA control panel has _not_ been set to force use of the discrete GPU for Firefox)

2. Go to https://bug1349799.bmoattachments.org/attachment.cgi?id=8941348

3. Hit the '2' key to select the high performance option

Actual results:

Note that the integrated GPU is used

Expected results:

The discrete (Nvidia) GPU should be used for WebGL contexts that are created setting "powerPreference" to "high-performance".

### DIFF
--- a/api/HTMLCanvasElement.json
+++ b/api/HTMLCanvasElement.json
@@ -314,10 +314,10 @@
                 "version_added": false
               },
               "firefox": {
-                "version_added": "63"
+                "version_added": false
               },
               "firefox_android": {
-                "version_added": "63"
+                "version_added": false
               },
               "ie": {
                 "version_added": false


ddbeck:
Hi @kkota1 and welcome to BCD! Thanks for this PR. You wrote:

> This was either never supported on Windows, or dropped around version ~71

Do you have any more information about this ever having been implemented in Firefox? If we can't show that this was implemented anywhere, we may want to consider dropping this phantom feature from BCD instead of modifying it.

kkota1:
I will look around shortly. My secondhand understanding is that it
currently works in Firefox for Mac

kkota1:
@ddbeck the FF source code used to default powerPreference to 'high-performance'. It's now written to default to 'low-power', but the feature is broken, not dropped as I originally said.

References to feature/source:

https://bugzilla.mozilla.org/show_bug.cgi?id=1561073
https://bugzilla.mozilla.org/show_bug.cgi?id=1562812

ddbeck:
> @ddbeck the FF source code used to default powerPreference to 'high-performance'. It's now written to default to 'low-power', but the feature is broken, not dropped as I originally said.

Thanks for the update, @kkota1! Given this additional information, I'd suggest adding one or more notes to the feature in BCD in the [`HTMLCanvasElement.json`](https://github.com/mdn/browser-compat-data/blob/master/api/HTMLCanvasElement.json#L317) file. You could use the note text to briefly describe the actual behavior, then link to the corresponding bug(s).

ddbeck:
Hi @kkota1, do you plan to return to this PR?

ddbeck:
I've opened #9060 to follow up on this. That PR will close this one, when it merges. Thanks for alerting us to the issues with this data, @kkota1!

Issue #7668, opened 2020-12-13
Rename PaymentRequest.paymentAddress to shippingAddress


saschanaz:
To match the spec https://developer.mozilla.org/docs/Web/API/PaymentRequest/shippingAddress

It has been shippingAddress for years, probably from the start. https://github.com/w3c/payment-request/commit/576550c4de9f47feb889f8f7e587b2853e2da668

PR #7669, opened 2020-12-13
PaymentRequest.p.paymentAddress → shippingAddress

sideshowbarker:
This change renames the BCD `paymentAddress` subfeature of the `PaymentRequest` feature to `shippingAddress`.

As noted in https://github.com/mdn/browser-compat-data/issues/7668, it’s been `PaymentRequest.prototype.shippingAddress` “for years, probably from the start”.

---

cc @marcoscaceres for confirmation

### DIFF
--- a/api/PaymentRequest.json
+++ b/api/PaymentRequest.json
@@ -503,56 +503,6 @@
           }
         }
       },
-      "paymentAddress": {
-        "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/PaymentRequest/paymentAddress",
-          "support": {
-            "chrome": {
-              "version_added": "61"
-            },
-            "chrome_android": {
-              "version_added": "53"
-            },
-            "edge": {
-              "version_added": "15"
-            },
-            "firefox": {
-              "version_added": false,
-              "notes": "Available only in nightly builds as <code>shippingAddress</code>."
-            },
-            "firefox_android": {
-              "version_added": false,
-              "notes": "Available only in nightly builds as <code>shippingAddress</code>."
-            },
-            "ie": {
-              "version_added": false
-            },
-            "opera": {
-              "version_added": "47"
-            },
-            "opera_android": {
-              "version_added": "44"
-            },
-            "safari": {
-              "version_added": null
-            },
-            "safari_ios": {
-              "version_added": null
-            },
-            "samsunginternet_android": {
-              "version_added": "6.0"
-            },
-            "webview_android": {
-              "version_added": false
-            }
-          },
-          "status": {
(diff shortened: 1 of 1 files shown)


api/PaymentRequest.json
```diff
@@ -604,36 +605,35 @@
...
             },
             "edge": {
-              "version_added": "≤18",
-              "version_removed": "79"
+              "version_added": "15"
             },
             "firefox": {
               "version_added": false,
-              "notes": "Available only in nightly builds."
+              "notes": "Available only in nightly builds as <code>shippingAddress</code>."
```
saschanaz, line 623: This note is now invalid:

```suggestion
              "flags": [
                {
                  "type": "preference",
                  "name": "dom.payments.request.enabled"
                }
              ]
```

api/PaymentRequest.json
```diff
@@ -604,36 +605,35 @@
...
             "opera": {
-              "version_added": false
+              "version_added": "47"
             },
             "opera_android": {
-              "version_added": false
+              "version_added": "44"
             },
             "safari": {
               "version_added": null
```
saschanaz, line 639: Safari support should probably be 11.1 (the version that first enabled Payment Request), as it seems it has have the attribute since the beginning.

https://webkit-search.igalia.com/webkit/source/Source/WebCore/Modules/paymentrequest/PaymentRequest.idl

api/PaymentRequest.json
```diff
@@ -604,36 +605,35 @@
...
             },
             "firefox": {
               "version_added": false,
-              "notes": "Available only in nightly builds."
+              "notes": "Available only in nightly builds as <code>shippingAddress</code>."
             },
             "firefox_android": {
               "version_added": false,
-              "notes": "Available only in nightly builds."
+              "notes": "Available only in nightly builds as <code>shippingAddress</code>."
```
sideshowbarker, line 627:
```suggestion
              "notes": "Available only in nightly builds as <code>shippingAddress</code>."
```
```suggestion
```

api/PaymentRequest.json
sideshowbarker: > Safari support should probably be 11.1 (the version that first enabled Payment Request), as it seems it has have the attribute since the beginning.
> 
> https://webkit-search.igalia.com/webkit/source/Source/WebCore/Modules/paymentrequest/PaymentRequest.idl

Thanks — will raise a separate PR for that

ddbeck (approved):
This looks good to me.

Do you want to wait for Marcos's confirmation before merging?

foolip:
Marcos has an away status, so I would suggest not waiting.

PR #7670, opened 2020-12-13
Correct data for "loading" attr in img/iframe

vinyldarkscratch:
This PR is a follow-up to #5442.  In that PR, the data was based upon information from ChromeStatus, which stated 76.  However, @foolip looked into it further and had confirmed that it was actually implemented in Chrome 77.  This PR fixes the data within the `img` element accordingly, as well as resolves a typo in the Opera version number for the attribute on the `iframe` element.

### DIFF
--- a/html/elements/iframe.json
+++ b/html/elements/iframe.json
@@ -539,7 +539,7 @@
                 "version_added": false
               },
               "opera": {
-                "version_added": "60"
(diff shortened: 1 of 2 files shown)


foolip (approved):

foolip:
Thanks for sorting this out!

PR #7672, opened 2020-12-13
Add missing features for Attr API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from a spec in Editor's Draft or more and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6), for the Attr API.

Additionally, this PR updates the main IE data for consistency.  (Support was confirmed.)

Spec: https://dom.spec.whatwg.org/#interface-attr

### DIFF
--- a/api/Attr.json
+++ b/api/Attr.json
@@ -20,7 +20,7 @@
             "version_added": "4"
           },
           "ie": {
-            "version_added": "8"
+            "version_added": "≤6"
           },
           "opera": {
             "version_added": "8"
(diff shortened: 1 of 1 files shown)


api/Attr.json
```diff
@@ -103,6 +103,53 @@
           }
         }
       },
+      "name": {
```
foolip, line 106: Surprising that this was missing, but it's listed in https://developer.mozilla.org/en-US/docs/Web/API/Attr with red link color, and I've confirmed it's *not* incorrectly in Node.json already, so this looks good!

api/Attr.json
```diff
@@ -159,6 +206,53 @@
           }
         }
       },
+      "ownerElement": {
```
foolip, line 209: This is in the same situation as "name" above.

api/Attr.json
```diff
@@ -214,6 +308,100 @@
             "deprecated": false
           }
         }
+      },
+      "specified": {
```
foolip, line 312: This is a funny attribute, it always returns true :)

At some point I think the API was able to *create* attributes that didn't yet exist via an `Attr` node, but no more...

foolip (approved):

PR #7673, opened 2020-12-13
Add Chromium versions for Range API

vinyldarkscratch:
This PR adds real values for Chrome and Chrome-based browsers for the Range API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).

### DIFF
--- a/api/Range.json
+++ b/api/Range.json
@@ -55,10 +55,10 @@
           "description": "<code>Range()</code> constructor",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "29"
             },
             "chrome_android": {
-              "version_added": true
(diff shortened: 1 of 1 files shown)


api/Range.json
```diff
@@ -440,10 +440,12 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Range/compareNode",
           "support": {
             "chrome": {
-              "version_added": false
+              "version_added": "1",
+              "version_removed": "45"
```
foolip, line 444: Yep, removed in https://bugs.chromium.org/p/chromium/issues/detail?id=445218

foolip (approved):

PR #7677, opened 2020-12-13
Add versions for IE for FileReaderSync API

vinyldarkscratch:
This PR adds real values for Internet Explorer for the FileReaderSync API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).

### DIFF
--- a/api/FileReaderSync.json
+++ b/api/FileReaderSync.json
@@ -20,7 +20,7 @@
             "version_added": "8"
           },
           "ie": {
-            "version_added": true
+            "version_added": "10"
           },
           "opera": {
             "version_added": true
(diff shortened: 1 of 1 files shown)


api/FileReaderSync.json
```diff
@@ -119,7 +119,7 @@
               "version_added": "8"
             },
             "ie": {
-              "version_added": true
+              "version_added": "11"
```
foolip, line 122: Curious that `readAsBinaryString` would be added after `readAsArrayBuffer`, but I see no reason to doubt the results.

foolip (approved):

PR #7678, opened 2020-12-13
Add/update IE+Edge versions for HTMLAllCollection API

vinyldarkscratch:
This PR adds real values for Internet Explorer for the HTMLAllCollection API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Additionally, this updates the Edge data based upon the IE data.


foolip (approved):
The data makes more sense now, thanks!

PR #7692, opened 2020-12-13
Add IE/Edge versions for SVGTransform[List] APIs

vinyldarkscratch:
This PR adds real values for IE and Edge for the SVGTransform and SVGTransformList APIs, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).

### DIFF
--- a/api/SVGTransformList.json
+++ b/api/SVGTransformList.json
@@ -57,7 +57,7 @@
               "version_added": true
             },
             "edge": {
-              "version_added": "79"
+              "version_added": "≤79"
             },
             "firefox": {
               "version_added": "9"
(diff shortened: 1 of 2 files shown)


api/SVGTransformList.json
```diff
@@ -57,7 +57,7 @@
               "version_added": true
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "79"
```
foolip, line 60: This doesn't seem like it can be right. A `FooList` interface without a `length` attribute would be very unusual. http://mdn-bcd-collector.appspot.com/tests/api/SVGTransformList tests the prototype only, so I think that's the problem here.

foolip (changes requested):

foolip (approved):

PR #7694, opened 2020-12-13
Add Firefox versions for DOMRect API

vinyldarkscratch:
This PR adds real values for Firefox for the DOMRect API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy

### DIFF
--- a/api/DOMRect.json
+++ b/api/DOMRect.json
@@ -20,10 +20,10 @@
             }
           ],
           "firefox": {
-            "version_added": true
+            "version_added": "27"
           },
           "firefox_android": {
-            "version_added": true
+            "version_added": "27"
           },
           "ie": [
             {


api/DOMRect.json
```diff
@@ -20,10 +20,10 @@
             }
           ],
           "firefox": {
-            "version_added": true
+            "version_added": "27"
```
foolip, line 23: This seems odd, since no subfeature in https://developer.mozilla.org/en-US/docs/Web/API/DOMRect#Browser_compatibility was introduced in Firefox 27. What was `DOMRect` like in Firefox 27, did it not have any members?

foolip: I've had a look in Firefox 27. `DOMRect` was exposed, and had members `left`, `top`, `right`, `bottom`, `width` and `height`. `document.body.getBoundingClientRect()` returned ` DOMRect`, and a `DOMRectList` was also exposed.

foolip: I used http://software.hixie.ch/utilities/js/live-dom-viewer/ to check what happened before Firefox 27, and then `document.body.getBoundingClientRect()` would return a `ClientRect`. So 27 is when `ClientRect` was renamed to `DOMRect` and exposed, but still didn't have a constructor. So this change is correct.

foolip: I've sent https://github.com/mdn/browser-compat-data/pull/8711 to add the `ClientRect` alt. name.

PR #7695, opened 2020-12-13
Add Firefox versions for HTMLAllCollection API

vinyldarkscratch:
This PR adds real values for Firefox for the HTMLAllCollection API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests used: https://mdn-bcd-collector.appspot.com/tests/api/HTMLAllCollection

### DIFF
--- a/api/HTMLAllCollection.json
+++ b/api/HTMLAllCollection.json
@@ -14,10 +14,10 @@
             "version_added": "≤18"
           },
           "firefox": {
-            "version_added": true
+            "version_added": "32"
           },
           "firefox_android": {
-            "version_added": true
+            "version_added": "32"
           },
           "ie": {
             "version_added": null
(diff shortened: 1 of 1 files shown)


api/HTMLAllCollection.json
```diff
@@ -14,10 +14,10 @@
             "version_added": "≤18"
           },
           "firefox": {
-            "version_added": true
+            "version_added": "32"
```
foolip, line 17: The `document.all` entry says Firefox 24. What kind of a thing was `document.all` in Firefox 24 if not an `HTMLAllCollection`?

vinyldarkscratch: Seems that it was an `HTMLCollection` rather than an `HTMLAllCollection`.  This appears to be common amongst all of the browsers.

foolip: Aha, this situation again. Well that explains it!

foolip (approved):

PR #7698, opened 2020-12-13
Add Firefox versions for HTMLOptionsCollection API

vinyldarkscratch:
This PR adds real values for Firefox for the HTMLOptionsCollection API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/HTMLOptionsCollection.json
+++ b/api/HTMLOptionsCollection.json
@@ -14,10 +14,10 @@
             "version_added": "12"
           },
           "firefox": {
-            "version_added": true
+            "version_added": "1"
           },
           "firefox_android": {
-            "version_added": true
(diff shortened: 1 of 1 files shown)


api/HTMLOptionsCollection.json
```diff
@@ -61,10 +61,10 @@
               "version_added": "≤18"
             },
             "firefox": {
-              "version_added": true
+              "version_added": "4"
```
foolip, line 64: Looks like `add` and `remove` were added slightly after the main interface, but that's reasonable, it started out as a readonly representation I guess.

foolip (approved):

PR #7699, opened 2020-12-13
Add Firefox versions for MediaKeySession API

vinyldarkscratch:
This PR adds real values for Firefox for the MediaKeySession API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/MediaKeySession

### DIFF
--- a/api/MediaKeySession.json
+++ b/api/MediaKeySession.json
@@ -14,10 +14,10 @@
             "version_added": "≤18"
           },
           "firefox": {
-            "version_added": true
+            "version_added": "38"
           },
           "firefox_android": {
-            "version_added": true
+            "version_added": "38"
           },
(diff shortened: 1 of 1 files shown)


api/MediaKeySession.json
```diff
@@ -349,10 +349,10 @@
               "version_added": "79"
             },
             "firefox": {
-              "version_added": true
+              "version_added": "52"
```
foolip, line 352: Oops, yet another case of onfoo attributes being added later. I don't doubt the results though.

foolip (approved):

PR #7700, opened 2020-12-13
Add Firefox versions for MediaKeyStatusMap API

vinyldarkscratch:
This PR adds real values for Firefox for the MediaKeyStatusMap API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/MediaKeyStatusMap

### DIFF
--- a/api/MediaKeyStatusMap.json
+++ b/api/MediaKeyStatusMap.json
@@ -14,10 +14,10 @@
             "version_added": "≤18"
           },
           "firefox": {
-            "version_added": true
+            "version_added": "38"
           },
           "firefox_android": {
-            "version_added": true
+            "version_added": "38"
           },
           "ie": {
             "version_added": false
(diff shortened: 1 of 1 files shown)


api/MediaKeyStatusMap.json
```diff
@@ -109,10 +109,10 @@
               "version_added": "13"
             },
             "firefox": {
-              "version_added": null
+              "version_added": "47"
```
foolip, line 112: This looks a bit suspect, that `forEach` should have been added later than `@@iterator`, since both are part of `iterable<T>` in Web IDL. Unfortunately https://github.com/mozilla/gecko-dev/blame/master/dom/webidl/MediaKeyStatusMap.webidl did not have any clues. Can you double check what happened here manually?

foolip: I've gone ahead and tested http://mdn-bcd-collector.appspot.com/tests/api/MediaKeyStatusMap in Firefox 46 and 47 and now can't doubt these results. All subfeatures except this test true in Firefox 46, so it really must have been missing. And `MediaKeyStatusMap` doesn't inherit from anything, so we don't need to test an instance to rule out that it was just moved around.

foolip (approved):

PR #7710, opened 2020-12-14
Drop stray &lt;/&gt; in CSS function descriptions

sideshowbarker:
This change updates the `description` fields for several CSS functions, to remove some stray `&lt;`/`&gt;` character references.

---

Or alternatively, I can update the patch in this branch to just remove the `description` fields completely — because for these cases, that `description` field doesn’t seem to be serving much purpose.

### DIFF
--- a/css/types/calc.json
+++ b/css/types/calc.json
@@ -3,7 +3,7 @@
     "types": {
       "calc": {
         "__compat": {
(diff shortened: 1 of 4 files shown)


ddbeck (approved):
I like this way of fixing these. Thank you! 🎉

Issue #7709, opened 2020-12-14
Remove invalid deprecation flags


saschanaz:
While I was working on https://github.com/saschanaz/types-web/commit/7c382a0386413120df13be89b5d50cb05bc3ec5e I found some invalid flags:

* `DOMPointReadOnly#matrixTransform`: I see no reason to mark this as deprecated. [This is a valid API](https://drafts.fxtf.org/geometry/#dom-dompointreadonly-matrixtransform).
* `HTMLImageElement#useMap`: [Same](https://html.spec.whatwg.org/multipage/image-maps.html#attr-hyperlink-usemap).
* `HTMLMapElement#areas`: While a selector call can easily replace this, this is still a useful shortcut and [the spec does not mark it as deprecated](https://html.spec.whatwg.org/multipage/image-maps.html#dom-map-areas).
* `HTMLMapElement#name`: This is actually "required" by [the spec](https://html.spec.whatwg.org/multipage/image-maps.html#attr-map-name).
   >The attribute must be present and must have a non-empty value with no ASCII whitespace
* `HTMLScriptElement#type`: You know, `type=module`.
* `SVGPathElement#getPointAtLength` and `SVGPathElement#getTotalLength`: Both are moved to the superclass SVGGeometryElement. Those should be just removed.

sideshowbarker:
> I found some invalid flags:
> …
> `HTMLImageElement#useMap`: [Same](https://html.spec.whatwg.org/multipage/image-maps.html#attr-hyperlink-usemap).
> …
> `HTMLScriptElement#type`: You know, `type=module`.

oops, those two are my fault — thanks for catching those. I’ve raised #7711 and #7712 with fixes

sideshowbarker:
> `DOMPointReadOnly#matrixTransform`: I see no reason to mark this as deprecated. [This is a valid API](https://drafts.fxtf.org/geometry/#dom-dompointreadonly-matrixtransform).

➡ #7713 

> `HTMLMapElement#areas`: While a selector call can easily replace this, this is still a useful shortcut and [the spec does not mark it as deprecated](https://html.spec.whatwg.org/multipage/image-maps.html#dom-map-areas).
> 
> `HTMLMapElement#name`: This is actually "required" by [the spec](https://html.spec.whatwg.org/multipage/image-maps.html#attr-map-name).
>       > The attribute must be present and must have a non-empty value with no ASCII whitespace

➡️ #7719

saschanaz:
[...]

PR #7712, opened 2020-12-14
Undeprecate the "type" attribute of the script element and HTMLScriptElement interface

sideshowbarker:
This change marks the `type` content/markup attribute of the `script` element and the `type` IDL/DOM attribute of the `HTMLScriptElement` interface `deprecated:false`. They were mistakenly marked `deprecated:true` in https://github.com/mdn/browser-compat-data/commit/a6f6d39.

Related: https://github.com/mdn/browser-compat-data/issues/7709

### DIFF
--- a/api/HTMLScriptElement.json
+++ b/api/HTMLScriptElement.json
@@ -588,7 +588,7 @@
           "status": {
             "experimental": false,
             "standard_track": true,
-            "deprecated": true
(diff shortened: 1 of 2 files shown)


vinyldarkscratch (approved):
Yep, it's not deprecated, thanks!

PR #7713, opened 2020-12-14
Undeprecate DOMPointReadOnly.p.matrixTransform

sideshowbarker:
This change marks `DOMPointReadOnly.prototype.matrixTransform` as `deprecated:false`. Per https://drafts.fxtf.org/geometry/#dom-dompointreadonly-matrixtransform it’s not currently deprecated (and it’s not clear it should ever have been marked `deprecated:true` to begin with).

Related: https://github.com/mdn/browser-compat-data/issues/7709

### DIFF
--- a/api/DOMPointReadOnly.json
+++ b/api/DOMPointReadOnly.json
@@ -190,7 +190,7 @@
           "status": {
             "experimental": false,
             "standard_track": true,
-            "deprecated": true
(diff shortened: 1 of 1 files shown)


vinyldarkscratch (approved):
Looks like it was marked as deprecated from the original wiki tables, but it's certainly not deprecated.

PR #7731, opened 2020-12-14
Add Firefox versions for SVGRenderingIntent API

vinyldarkscratch:
This PR adds real values for Firefox for the SVGRenderingIntent API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/SVGRenderingIntent

### DIFF
--- a/api/SVGRenderingIntent.json
+++ b/api/SVGRenderingIntent.json
@@ -14,10 +14,10 @@
             "version_added": null
           },
           "firefox": {
-            "version_added": null
+            "version_added": false
           },
           "firefox_android": {
-            "version_added": null
+            "version_added": false
           },
           "ie": {
             "version_added": null


api/SVGRenderingIntent.json
```diff
@@ -14,10 +14,10 @@
             "version_added": null
           },
           "firefox": {
-            "version_added": null
+            "version_added": false
```
foolip, line 17: From https://developer.mozilla.org/docs/Web/API/SVGRenderingIntent it looks like we should probably remove this entry. But this PR is fine.

foolip (approved):

PR #7732, opened 2020-12-14
Add Firefox versions for SVGSolidcolorElement API

vinyldarkscratch:
This PR adds real values for Firefox for the SVGSolidcolorElement API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/SVGSolidcolorElement

### DIFF
--- a/api/SVGSolidcolorElement.json
+++ b/api/SVGSolidcolorElement.json
@@ -14,10 +14,10 @@
             "version_added": null
           },
           "firefox": {
-            "version_added": null
+            "version_added": false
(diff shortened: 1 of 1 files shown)


foolip (approved):
https://developer.mozilla.org/docs/Web/API/SVGSolidcolorElement should be removed, but this PR is correct.

PR #7733, opened 2020-12-14
Add Firefox versions for SVGStylable API

vinyldarkscratch:
This PR adds real values for Firefox for the SVGStylable API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/SVGStylable.json
+++ b/api/SVGStylable.json
@@ -14,10 +14,12 @@
             "version_added": "≤18"
           },
           "firefox": {
-            "version_added": null
+            "version_added": "1.5",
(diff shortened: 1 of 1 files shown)


foolip (approved):
The way https://developer.mozilla.org/docs/Web/API/SVGStylable is described it sounds a bit like a mixin, but I guess this is more like `SVGZoomAndPan` or `SVGUnitTypes`, things that are kinda like mixins but predate Web IDL mixins by a lot, and where the interface object was exposed.

PR #7738, opened 2020-12-14
Add Firefox versions for SVGTransformable API

vinyldarkscratch:
This PR adds real values for Firefox for the SVGTransformable API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/SVGTransformable

### DIFF
--- a/api/SVGTransformable.json
+++ b/api/SVGTransformable.json
@@ -14,10 +14,12 @@
             "version_added": "≤18"
           },
           "firefox": {
-            "version_added": null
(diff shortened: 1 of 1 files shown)


foolip (changes requested):
Let's figure out what to do in https://github.com/mdn/browser-compat-data/issues/7752

PR #7742, opened 2020-12-14
Add Firefox versions for SVGVKernElement API

vinyldarkscratch:
This PR adds real values for Firefox for the SVGVKernElement API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/SVGVKernElement

### DIFF
--- a/api/SVGVKernElement.json
+++ b/api/SVGVKernElement.json
@@ -14,10 +14,10 @@
             "version_added": null
           },
           "firefox": {
-            "version_added": null
+            "version_added": false
           },
           "firefox_android": {
-            "version_added": null
+            "version_added": false
           },
           "ie": {
             "version_added": false


foolip (commented):
https://developer.mozilla.org/en-US/docs/Web/API/SVGVKernElement is full of nulls. Perhaps we should set it to false everywhere, or delete the entry?

vinyldarkscratch:
It appears that this is still supported in the most recent edition of Safari, though it is unsupported in all other browsers.

foolip:
OK, in that case the entry should stay. I'll merge this.

foolip (approved):

PR #7743, opened 2020-12-14
Add Firefox versions for SVGZoomAndPan API

vinyldarkscratch:
This PR adds real values for Firefox for the SVGZoomAndPan API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/SVGZoomAndPan

### DIFF
--- a/api/SVGZoomAndPan.json
+++ b/api/SVGZoomAndPan.json
@@ -14,10 +14,12 @@
             "version_added": "≤79"
           },
           "firefox": {
-            "version_added": null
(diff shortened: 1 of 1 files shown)


foolip (approved):
This is a weird interface, but it's been exposed because it has constants, so isn't like most in https://github.com/mdn/browser-compat-data/issues/7752.

PR #7749, opened 2020-12-14
Add missing SVGPointList API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from a spec in Editor's Draft or more and is supported in at least one browser.  This particular PR adds the missing SVGPointList API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).

Spec: https://svgwg.org/svg2-draft/shapes.html#InterfaceSVGPointList

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/SVGPointList

### DIFF
--- a/api/SVGPointList.json
+++ b/api/SVGPointList.json
@@ -285,13 +285,13 @@
         "__compat": {
           "support": {
             "chrome": {
-              "version_added": "44"
+              "version_added": true
             },
             "chrome_android": {
-              "version_added": "44"
+              "version_added": true
             },
             "edge": {
-              "version_added": "79"
+              "version_added": true
             },
             "firefox": {
               "version_added": "5"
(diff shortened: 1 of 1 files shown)


api/SVGPointList.json
```diff
@@ -0,0 +1,474 @@
...
+            "standard_track": true,
+            "deprecated": false
+          }
+        }
+      },
+      "length": {
+        "__compat": {
+          "support": {
+            "chrome": {
+              "version_added": "44"
```
foolip, line 288: Here's yet another `length` member with suspicious results. Maybe we should treat all `length` attributes as special in the collector?

I assume that this was really added with the interface itself, so the data for many browsers here need updated.

api/SVGPointList.json
```diff
@@ -0,0 +1,474 @@
...
+            "standard_track": true,
+            "deprecated": false
+          }
+        }
+      },
+      "numberOfItems": {
+        "__compat": {
+          "support": {
+            "chrome": {
+              "version_added": "44"
```
foolip, line 335: `numberOfItems` seems to have the same issue as `length`.

foolip (changes requested):

api/SVGPointList.json
foolip: I've filed https://github.com/mdn/browser-compat-data/issues/7755 about other instances of this that might have slipped through review.

foolip (approved):

PR #7751, opened 2020-12-14
[HOTFIX] Remove duplicate "width" feature from HTMLInputElement API

vinyldarkscratch:
Two PRs that were recently merged had both added a "width" feature to the HTMLInputElement API, causing a failure of the linter.  This PR is a quick fix to remove the duplicate feature and combine the two entries.

### DIFF
--- a/api/HTMLInputElement.json
+++ b/api/HTMLInputElement.json
@@ -2699,53 +2699,6 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLInputElement/width",
           "support": {
             "chrome": {
-              "version_added": true
-            },
-            "chrome_android": {
-              "version_added": true
-            },
-            "edge": {
-              "version_added": "≤18"
-            },
-            "firefox": {
-              "version_added": "16"
-            },
-            "firefox_android": {
-              "version_added": "16"
-            },
-            "ie": {
-              "version_added": false
-            },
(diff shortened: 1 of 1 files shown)


vinyldarkscratch:
Notice: Since this is a fix for a failing build in `master`, I merged this without review to resolve the build, though I know I'm technically not supposed to do that.

ddbeck:
> Two PRs that were recently merged

@vinyldarkscratch Can you cite the PRs at issue here? And link to the failing actions? I'd like to understand the problem and the fix better.

vinyldarkscratch:
Of course!  The two PRs are #7647 and #7671.  Both of the PRs added the `width` feature for the HTMLInputElement API, but since the second of those PRs had performed a rename of a ghost feature (adding `width` in its place), Git didn't produce any merge conflicts, thus resulting in both PRs mergeable without issues. 
 https://github.com/mdn/browser-compat-data/runs/1550069646  (Admittedly, the linter's error message isn't too helpful, maybe we can create an update to the linter for it to catch and alert duplicates better?)

ddbeck:
Thanks, @vinyldarkscratch! Seems like an unusual case (two PRs against the same data, merged closely to each other). This doesn't happen terribly often (I can think of maybe one other case this year), so I'm OK with letting this fix be the end of it, unless this starts happening routinely.

PR #7753, opened 2020-12-14
Add/correct Safari versions for HTMLAllCollection API

vinyldarkscratch:
This PR adds and corrects real values for Safari for the HTMLAllCollection API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/HTMLAllCollection

### DIFF
--- a/api/HTMLAllCollection.json
+++ b/api/HTMLAllCollection.json
@@ -124,7 +124,7 @@
               "version_added": "32"
             },
             "safari": {
-              "version_added": true
+              "version_added": "10"
             },
             "safari_ios": {
               "version_added": true
--- a/api/HTMLAllCollection.json
+++ b/api/HTMLAllCollection.json
@@ -124,10 +124,10 @@
               "version_added": "32"
             },
             "safari": {
-              "version_added": "10"
+              "version_added": true
             },
             "safari_ios": {
-              "version_added": "10"
+              "version_added": true
             },
             "samsunginternet_android": {
               "version_added": "4.0"
--- a/api/HTMLAllCollection.json
+++ b/api/HTMLAllCollection.json
@@ -29,10 +29,10 @@
             "version_added": true
           },
           "safari": {
-            "version_added": true
+            "version_added": "5"
           },
           "safari_ios": {
-            "version_added": true
+            "version_added": "4"
           },
           "samsunginternet_android": {
             "version_added": true
@@ -76,10 +76,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": "6"
+              "version_added": "5"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "4"
             },
             "samsunginternet_android": {
               "version_added": true
@@ -127,7 +127,7 @@
               "version_added": "10"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "10"
             },
             "samsunginternet_android": {
               "version_added": "4.0"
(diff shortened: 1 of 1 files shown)


api/HTMLAllCollection.json
```diff
@@ -29,10 +29,10 @@
             "version_added": true
           },
           "safari": {
-            "version_added": true
+            "version_added": "5"
```
foolip, line 32: This is later than the support for `document.all`. Was `document.all` something else before this?

vinyldarkscratch: Looks like `document.all` just returned an `HTMLCollection` in older Safari versions:
Screen Shot 2020-12-14 at 04 56 40

foolip: It's tempting to suggest a `partial_implementation` for the range where `document.all` was a `HTMLCollection`, but unfortunately it seems that wouldn't capture what matters to web developers most.

Most of the things that work with an `HTMLAllCollection` also work with `HTMLAllCollection`. The ability to call `document.all` is only spec'd for `HTMLAllCollection`, so I tested http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8753 in Safari 4. That shows that their `HTMLCollection` was also possible to call as a function. So I think that `HTMLCollection` was later forked into `HTMLAllCollection` and in many small changes over time the two reached their current state.

Without careful testing, it'll be hard to write a note that says something that can be usefully interpreted by web developers. http://wpt.live/html/infrastructure/common-dom-interfaces/collections/htmlallcollection.html is pretty exhaustive, but it won't run on old Safari. Most of the tests were fixed between Safari 10.1 and 11.1, and so not related to `HTMLCollection` vs. `HTMLAllCollection`.

So... I don't have any suggestion for a note that would be useful.

api/HTMLAllCollection.json
```diff
@@ -127,7 +127,7 @@
               "version_added": "10"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "10"
```
foolip, line 130: This is the `length` property. Pretty sure it was supported from the time the interface was introduced.

api/HTMLAllCollection.json
vinyldarkscratch: Hmm...  From what I can tell, both `HTMLCollection` and `HTMLAllCollection` have almost the same IDL, with a few slight differences.  I'm thinking that it may not be relevant enough to make a note, possibly?

api/HTMLAllCollection.json
```diff
@@ -124,7 +124,7 @@
               "version_added": "32"
             },
             "safari": {
-              "version_added": "10"
+              "version_added": true
```
foolip, line 127: This is changing "10" to true, but I've confirmed with http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8758 that it was supported in Safari 9, so this data was not correct. Better to leave it as true and come back to it later. (We haven't generally done this, but it's fine to do in rare cases I think.)

api/HTMLAllCollection.json
foolip: I agree, I don't think the differences are large enough to warrant a note.

foolip (approved):

api/HTMLAllCollection.json
foolip: Or actually, I'll revert this change, since otherwise the commit message doesn't work. This is removing versions, not adding/correcting :)

foolip, line 127:
```suggestion
              "version_added": "10"
```

PR #7756, opened 2020-12-14
Add Safari versions for HTMLCollection API

vinyldarkscratch:
This PR adds real values for Safari for the `HTMLCollection` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/HTMLCollection

### DIFF
--- a/api/HTMLCollection.json
+++ b/api/HTMLCollection.json
@@ -32,7 +32,7 @@
             "version_added": "4"
           },
           "safari_ios": {
-            "version_added": "3.2"
+            "version_added": "≤3"
           },
           "samsunginternet_android": {
             "version_added": "1.0"
@@ -76,10 +76,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "≤4"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "≤3"
             },
             "samsunginternet_android": {
               "version_added": true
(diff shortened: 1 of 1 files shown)


api/HTMLCollection.json
```diff
@@ -76,10 +76,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "≤4"
```
foolip, line 79: This ideally should have failed the lint. The parent feature was version "4", so "≤4" doesn't make sense, it couldn't have been a release before 4.

I see you added some custom tests in https://github.com/foolip/mdn-bcd-collector/pull/872, will that allow us to check for support in Safari 3 and see if we can resolve this to use either see  "4", or "≤4" everything?

foolip:
I still think it's pretty likely this whole interface was supported before Safari 4. `HTMLCollection` was in the earliest generated bindings checked into WebKit, with `namedItem` and `item` functions. Not sure about `length`, it might be special and supported in some other way than via the prototype. But at least I'm pretty sure parts of `HTMLCollection` was supported since Safari 1.

vinyldarkscratch:
Good point, yeah -- I just ran a quick test and can confirm it's in at least Safari 3.  Added a range!

foolip (approved):

foolip:
Thanks @vinyldarkscratch!

PR #7758, opened 2020-12-14
Add Safari versions for HTMLFormControlsCollection API

vinyldarkscratch:
This PR adds real values for Safari for the `HTMLFormControlsCollection` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/HTMLFormControlsCollection

### DIFF
--- a/api/HTMLFormControlsCollection.json
+++ b/api/HTMLFormControlsCollection.json
@@ -29,10 +29,10 @@
             "version_added": true
           },
           "safari": {
-            "version_added": true
+            "version_added": "6.1"
           },
           "safari_ios": {
-            "version_added": true
+            "version_added": "7"
           },
           "samsunginternet_android": {
             "version_added": true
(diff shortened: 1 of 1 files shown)


api/HTMLFormControlsCollection.json
```diff
@@ -29,10 +29,10 @@
             "version_added": true
           },
           "safari": {
-            "version_added": true
+            "version_added": "6.1"
```
foolip, line 32: I think there's a problem of a missing interface object here. `HTMLFormControlsCollection` is the type of `form.elements`, and that was added in Safari 3 per https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement#Browser_compatibility

foolip: I looked this up in source, `HTMLFormControlsCollection` was added in https://trac.webkit.org/changeset/135093/webkit which does map to Safari 6.1, and `form.elements` was a `HTMLCollection` before that, confirmed by http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8800 in Safari 5.1.

foolip (approved):

PR #7761, opened 2020-12-14
Add Safari versions for HTMLMediaElement API

vinyldarkscratch:
This PR adds real values for Safari for the `HTMLMediaElement` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/HTMLMediaElement.json
+++ b/api/HTMLMediaElement.json
@@ -1798,7 +1798,10 @@
               "version_added": "29"
             },
             "safari": {
-              "version_added": null
+              "version_added": "12.1"
+            },
+            "safari_ios": {
+              "version_added": "12.2"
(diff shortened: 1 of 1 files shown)


api/HTMLMediaElement.json
```diff
@@ -1798,7 +1798,10 @@
               "version_added": "29"
             },
             "safari": {
-              "version_added": null
+              "version_added": "12.1"
```
foolip, line 1801: Matches https://developer.mozilla.org/en-US/docs/Web/API/MediaKeys#Browser_compatibility, good!

foolip (approved):

PR #7762, opened 2020-12-14
Add Safari versions for HTMLOptionsCollection API

vinyldarkscratch:
This PR adds real values for Safari for the `HTMLOptionsCollection` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/HTMLOptionsCollection.json
+++ b/api/HTMLOptionsCollection.json
@@ -29,10 +29,10 @@
             "version_added": true
           },
           "safari": {
-            "version_added": true
+            "version_added": "6.1"
           },
           "safari_ios": {
-            "version_added": true
+            "version_added": "7"
           },
           "samsunginternet_android": {
             "version_added": true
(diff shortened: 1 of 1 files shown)


api/HTMLOptionsCollection.json
```diff
@@ -29,10 +29,10 @@
             "version_added": true
           },
           "safari": {
-            "version_added": true
+            "version_added": "6.1"
```
foolip, line 32: `HTMLOptionsCollection` is used for `select.options`, and https://developer.mozilla.org/en-US/docs/Web/API/HTMLSelectElement/options#Browser_compatibility says that's supported since earlier than this. http://mdn-bcd-collector.appspot.com/tests/api/HTMLOptionsCollection is testing the prototype only, can you double check if this shouldn't be supported earlier?

foolip: I've sent https://github.com/mdn/browser-compat-data/pull/8727, closing this PR.

PR #7763, opened 2020-12-14
Add Safari versions for Node API

vinyldarkscratch:
This PR adds real values for Safari for the `Node` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Node

### DIFF
--- a/api/Node.json
+++ b/api/Node.json
@@ -352,7 +352,7 @@
               "version_added": "≤4"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "≤3"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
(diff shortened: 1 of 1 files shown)


api/Node.json
```diff
@@ -1111,10 +1113,12 @@
               "version_removed": "33"
             },
             "safari": {
-              "version_added": null
+              "version_added": "≤4",
```
foolip, line 1116: This and the following edits are for the attributes in https://github.com/mdn/browser-compat-data/issues/6545. The data is correct though, so I'll merge this to make it easier to get all the versions right when the cleanup suggested in that issue happens.

foolip (approved):

PR #7770, opened 2020-12-14
Add Safari versions for SVGAnimated* APIs

vinyldarkscratch:
This PR adds real values for Safari for various SVG animation-based APIs, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).

### DIFF
--- a/api/SVGAnimatedString.json
+++ b/api/SVGAnimatedString.json
@@ -29,10 +29,10 @@
             "version_added": true
           },
           "safari": {
-            "version_added": true
+            "version_added": "5"
           },
           "safari_ios": {
-            "version_added": true
+            "version_added": "4"
           },
           "samsunginternet_android": {
             "version_added": false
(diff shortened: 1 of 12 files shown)


api/SVGAnimatedString.json
```diff
@@ -76,10 +76,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "6.1"
```
foolip, line 79: Looks like this is the only SVGAnimated* interface which has the `animVal` and `baseVal` entries, even though there are 12 interfaces that have them in the SVG spec's IDL.

Anyway, I don't trust this version, Safari 6.1 keeps coming up as the version for added attributes, and I think it's because we're testing the prototype only. The interface without these attributes wouldn't make any sense. Can you double check?

vinyldarkscratch: Yeah, we're testing the prototype rather than an instance (I'm not yet certain how to create an instance for these).  Let me revert changes that set "6.1" for attributes.

foolip: Thanks!

foolip (approved):

PR #7771, opened 2020-12-14
Add Chromium versions for HTMLAllCollection API

vinyldarkscratch:
This PR adds real values for Chromium for the `HTMLAllCollection` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/HTMLAllCollection

### DIFF
--- a/api/HTMLAllCollection.json
+++ b/api/HTMLAllCollection.json
@@ -5,10 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLAllCollection",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "4"
           },
           "chrome_android": {
-            "version_added": true
+            "version_added": "18"
           },
           "edge": {
             "version_added": "12"
@@ -23,10 +23,10 @@
             "version_added": "11"
           },
           "opera": {
-            "version_added": true
+            "version_added": "15"
           },
           "opera_android": {
-            "version_added": true
+            "version_added": "14"
           },
           "safari": {
             "version_added": true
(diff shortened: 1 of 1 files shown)


api/HTMLAllCollection.json
```diff
@@ -5,10 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLAllCollection",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "4"
```
foolip, line 8: `document.all` is marked as supported in Chrome 1. Was it `HTMLCollection` between Chrome 1 and 3? Same basic issue as in https://github.com/mdn/browser-compat-data/pull/7753.

foolip: Per https://github.com/mdn/browser-compat-data/pull/7753#discussion_r542362347 it was indeed a `HTMLCollection` originally. In that PR we went with just representing when `HTMLAllCollection` was added without adding a pedantic note about it on `document.all`, so let's do the same here.

api/HTMLAllCollection.json
```diff
@@ -23,10 +23,10 @@
             "version_added": "11"
           },
           "opera": {
-            "version_added": true
+            "version_added": "15"
```
foolip, line 26: http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8803 in Opera 12.16 shows `document.all` is a `HTMLCollection`, so this seems right.

foolip (approved):

PR #7779, opened 2020-12-14
Add Chromium versions for HTMLFormControlsCollection API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `HTMLFormControlsCollection` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/HTMLFormControlsCollection

### DIFF
--- a/api/HTMLFormControlsCollection.json
+++ b/api/HTMLFormControlsCollection.json
@@ -98,7 +98,7 @@
               "version_added": true
             },
             "samsunginternet_android": {
-              "version_added": "≤13.0"
+              "version_added": "1.5"
             },
             "webview_android": {
               "version_added": "≤37"
--- a/api/HTMLFormControlsCollection.json
+++ b/api/HTMLFormControlsCollection.json
@@ -5,10 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLFormControlsCollection",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "25"
           },
           "chrome_android": {
-            "version_added": true
+            "version_added": "25"
           },
           "edge": {
             "version_added": "79"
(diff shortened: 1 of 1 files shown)


api/HTMLFormControlsCollection.json
```diff
@@ -5,10 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLFormControlsCollection",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "25"
```
foolip, line 8: https://developer.mozilla.org/en-US/docs/Web/API/HTMLFormElement#Browser_compatibility says `form.elements` was in Chrome 1, so this might be a case of interface object not being exposed. Can you check what kind of an object `form.elements` was in Chrome 24?

foolip: I figured out the story for WebKit in https://github.com/mdn/browser-compat-data/pull/7758 which helps resolve this. It was added in https://trac.webkit.org/changeset/135093/webkit, which was [WebKit trunk 537.20.0](https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Configurations/Version.xcconfig?rev=135093) and that indeed is Chrome 25. So this was right!

api/HTMLFormControlsCollection.json
```diff
@@ -23,10 +23,10 @@
             "version_added": false
           },
           "opera": {
-            "version_added": true
+            "version_added": "15"
```
foolip, line 26: http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8800 in Opera 12.16 shows `form.elements` was a plain `HTMLCollection` there, so this seems right too.

foolip (approved):

PR #7780, opened 2020-12-14
Add Opera versions for HTMLElement API

vinyldarkscratch:
This PR adds real values for Opera and Opera Android for the `HTMLElement` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/HTMLElement

### DIFF
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -596,7 +596,7 @@
               "version_added": "9"
             },
             "opera_android": {
-              "version_added": true
+              "version_added": "10.1"
             },
             "safari": {
               "version_added": "3"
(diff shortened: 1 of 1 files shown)


api/HTMLElement.json
```diff
@@ -841,10 +841,12 @@
               "version_added": false
             },
             "opera": {
-              "version_added": null
+              "version_added": "≤12.1",
+              "version_removed": "15"
```
foolip, line 845: This seems inconsistent with the data for Chrome. If `dropzone` was removed in Chrome 59, would it not have been in Opera 15?

vinyldarkscratch: Surprisingly enough, it was not introduced to Opera Blink.

foolip: Huh. I guess it might have required some platform integration work which was never done and the feature was disabled. Thanks for confirming!

foolip (approved):

PR #7783, opened 2020-12-14
Add Chromium versions for HTMLOptionsCollection API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `HTMLOptionsCollection` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/HTMLOptionsCollection.json
+++ b/api/HTMLOptionsCollection.json
@@ -5,10 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLOptionsCollection",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "25"
           },
           "chrome_android": {
-            "version_added": true
+            "version_added": "25"
           },
           "edge": {
             "version_added": "12"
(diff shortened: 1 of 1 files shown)


api/HTMLOptionsCollection.json
```diff
@@ -5,10 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLOptionsCollection",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "25"
```
foolip, line 8: I have the same doubt here as in https://github.com/mdn/browser-compat-data/pull/7762#discussion_r542636041. I think this was supported earlier.

foolip:
Closing this in favor of https://github.com/mdn/browser-compat-data/pull/7783. That doesn't update the Opera data, but that's left as true so we can come back to it.

foolip:
Uh, no, it's closed in favor of https://github.com/mdn/browser-compat-data/pull/8727 :)

PR #7786, opened 2020-12-14
Add Chromium versions for HTMLVideoElement API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `HTMLVideoElement` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/HTMLVideoElement

### DIFF
--- a/api/HTMLVideoElement.json
+++ b/api/HTMLVideoElement.json
@@ -5,10 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLVideoElement",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "1"
           },
           "chrome_android": {
-            "version_added": true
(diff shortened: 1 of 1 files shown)


api/HTMLVideoElement.json
```diff
@@ -119,7 +119,7 @@
               "version_added": "69"
             },
             "opera_android": {
-              "version_added": false
+              "version_added": "59"
```
foolip, line 122: Yay, thanks for updating this! We must have hundreds upon hundreds of entries incorrectly set to false for various Chromium-based browsers...

foolip (approved):

PR #7788, opened 2020-12-14
Add Chromium versions for RTCIceCandidate API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `RTCIceCandidate` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/RTCIceCandidate

### DIFF
--- a/api/RTCIceCandidate.json
+++ b/api/RTCIceCandidate.json
@@ -5,10 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/RTCIceCandidate",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "23"
           },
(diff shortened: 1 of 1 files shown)


foolip (approved):

foolip:
I had a look at https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection#Browser_compatibility and see that the "icecandidate" event was added in Chrome 24, and then other ice stuff in Chrome 25, 26 and 28. So this is consistent with that whole feature growing into place during that time.

Issue #7752, opened 2020-12-14
Sort out SVG's mixin-like interfaces


foolip:
[SVG 1.1](https://www.w3.org/TR/SVG11/single-page.html) defined some interfaces which were a lot like mixins, with wording like this:

> Interface `SVGTests` defines an interface which applies to all elements which have attributes ‘requiredFeatures’, ‘requiredExtensions’ and ‘systemLanguage’.

or:

> Interface `SVGTransformable` contains properties and methods that apply to all elements which have attribute ‘transform’.

However, these predated Web IDL mixins, and some of them were implemented in a way which exposed the interface object for them.

A number of interfaces like this are in BCD/MDN:
- https://developer.mozilla.org/en-US/docs/Web/API/SVGAnimatedPoints
- https://developer.mozilla.org/en-US/docs/Web/API/SVGExternalResourcesRequired
- https://developer.mozilla.org/en-US/docs/Web/API/SVGFilterPrimitiveStandardAttributes
- https://developer.mozilla.org/en-US/docs/Web/API/SVGStylable
- https://developer.mozilla.org/en-US/docs/Web/API/SVGTests
- https://developer.mozilla.org/en-US/docs/Web/API/SVGTransformable
- https://developer.mozilla.org/en-US/docs/Web/API/SVGURIReference

This list might not be complete.

We should probably delete all of these pages and their BCD entries, but in the process we should also make sure that the members are instead of the other correct interfaces.

Note that https://developer.mozilla.org/en-US/docs/Web/API/SVGUnitTypes is *not* to be removed, because https://svgwg.org/svg2-draft/types.html#InterfaceSVGUnitTypes defines it as a plain interface, because there are constants on it that are accessed through the interface object.

foolip:
I noticed this while reviewing @vinyldarkscratch's PRs, and only after a merged a few PRs making changes to these entries: https://github.com/mdn/browser-compat-data/pull/7716 + https://github.com/mdn/browser-compat-data/pull/7720 + https://github.com/mdn/browser-compat-data/pull/7733

There are additional open PRs with this issue, and I'll link those to this issue instead of merging them.

foolip:
https://developer.mozilla.org/en-US/docs/Web/API/SVGZoomAndPan is also not in this category, since it also has constants on it.

foolip:
Oops, https://github.com/mdn/browser-compat-data/pull/7718 is another case already merged.

ddbeck:
[...]

PR #7793, opened 2020-12-14
Add Chromium versions for SVGAnimatedPoints API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `SVGAnimatedPoints` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.


foolip (changes requested):
I'd prefer to sort this out in https://github.com/mdn/browser-compat-data/issues/7752.

PR #7796, opened 2020-12-14
Add Chromium versions for SVGElement API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `SVGElement` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/SVGElement.json
+++ b/api/SVGElement.json
@@ -344,11 +344,11 @@
         "__compat": {
           "support": {
             "chrome": {
-              "version_added": true,
+              "version_added": "1",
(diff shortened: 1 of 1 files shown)


foolip (approved):
These are the attributes removed in https://groups.google.com/a/chromium.org/g/blink-dev/c/jjwLLSG_hGY/m/Ovi-nvEeDwAJ

PR #7818, opened 2020-12-14
Add Chromium versions for SVGRenderingIntent API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `SVGRenderingIntent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/SVGRenderingIntent

### DIFF
--- a/api/SVGRenderingIntent.json
+++ b/api/SVGRenderingIntent.json
@@ -5,10 +5,12 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/SVGRenderingIntent",
         "support": {
           "chrome": {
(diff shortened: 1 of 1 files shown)


foolip (approved):
Seeing this number made me file https://github.com/mdn/browser-compat-data/issues/7844, but that's not an issue here since we're not using https://omahaproxy.appspot.com/ at all here.

PR #7822, opened 2020-12-14
Add Chromium versions for SVGStylable API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `SVGStylable` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.


foolip (changes requested):
This is a case of https://github.com/mdn/browser-compat-data/issues/7752

PR #7829, opened 2020-12-14
Add Chromium versions for SVGUnitTypes API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `SVGUnitTypes` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/SVGUnitTypes


foolip (approved):
This is mentioned in https://github.com/mdn/browser-compat-data/issues/7752 but not one of the now removed interfaces.

PR #7832, opened 2020-12-14
Add Chromium versions for XPathExpression API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `XPathExpression` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/XPathExpression

### DIFF
--- a/api/XPathExpression.json
+++ b/api/XPathExpression.json
@@ -5,10 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/XPathExpression",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "29"
           },
           "chrome_android": {
-            "version_added": true
+            "version_added": "29"
           },
           "edge": {
             "version_added": "12"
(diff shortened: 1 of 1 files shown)


api/XPathExpression.json
```diff
@@ -5,10 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/XPathExpression",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "29"
```
foolip, line 8: The data added in https://github.com/mdn/browser-compat-data/pull/7644 says that `XPathEvaluator.createExpression()` was supported since Chrome 1, so this doesn't seem right. Maybe the interface object wasn't exposed?

foolip (changes requested):

api/XPathExpression.json
vinyldarkscratch: That seems the most likely -- I'll reopen this with custom tests introduced in 1.1.7!

PR #7837, opened 2020-12-14
Add Chromium versions for Plugin API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `Plugin` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/Plugin.json
+++ b/api/Plugin.json
@@ -5,10 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/Plugin",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "1"
           },
           "chrome_android": {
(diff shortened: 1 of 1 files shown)


api/Plugin.json
```diff
@@ -300,10 +300,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Plugin/version",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": false
```
foolip, line 303: OK, I guess that `version` might have been an IE+Edge thing then.

foolip (approved):

PR #7842, opened 2020-12-14
Add missing features for CSSStyleValue API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from a spec in Editor's Draft or more and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6), for the `CSSStyleValue` API.

Spec: https://drafts.css-houdini.org/css-typed-om-1/

IDL: https://github.com/w3c/webref/blob/master/ed/idl/css-typed-om.idl

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/CSSStyleValue

### DIFF
--- a/api/CSSStyleValue.json
+++ b/api/CSSStyleValue.json
@@ -142,6 +142,53 @@
             "deprecated": false
           }
         }
+      },
+      "toString": {
+        "__compat": {
+          "support": {
+            "chrome": {
+              "version_added": "66"
+            },
+            "chrome_android": {
+              "version_added": "66"
+            },
+            "edge": {
+              "version_added": "79"
+            },
+            "firefox": {
+              "version_added": false
+            },
+            "firefox_android": {
+              "version_added": false
+            },
+            "ie": {
+              "version_added": false
+            },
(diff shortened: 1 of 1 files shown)


api/CSSStyleValue.json
```diff
@@ -142,6 +142,53 @@
             "deprecated": false
           }
         }
+      },
+      "toString": {
```
foolip, line 146: https://mdn-bcd-collector.appspot.com/tests/api/CSSStyleValue/toString has the problem described in https://github.com/foolip/mdn-bcd-collector/issues/841.

I think this data is correct, but if `toString` had been added later, the test wouldn't have caught that. Can you double check this one?

vinyldarkscratch: Can confirm this one works in Chrome 66!

foolip: Thanks!

foolip (approved):

PR #7845, opened 2020-12-15
Add missing features for CanvasRenderingContext2D API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from a spec in Editor's Draft or more and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6), for the `CanvasRenderingContext2D` API.

Spec: https://html.spec.whatwg.org/multipage/canvas.html#canvasrenderingcontext2d

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/CanvasRenderingContext2D

### DIFF
--- a/api/CanvasRenderingContext2D.json
+++ b/api/CanvasRenderingContext2D.json
@@ -2123,6 +2123,53 @@
           }
         }
       },
+      "getContextAttributes": {
+        "__compat": {
+          "support": {
+            "chrome": {
+              "version_added": "32"
+            },
+            "chrome_android": {
+              "version_added": "32"
+            },
+            "edge": {
+              "version_added": "79"
+            },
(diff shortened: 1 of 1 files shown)


api/CanvasRenderingContext2D.json
```diff
@@ -2123,6 +2123,53 @@
           }
         }
       },
+      "getContextAttributes": {
```
foolip, line 2126: Yep, looks like this isn't in Gecko or WebKit. They do have `getContextAttributes` for WebGL though.

foolip (approved):

PR #7850, opened 2020-12-15
Add Bluetooth.getDevices

hamishwillee:
- [x] Summarize your changes

This adds documentation for the new Bluetooth.getDevices() which was recently added to chromium (build 85) under a flag. Some docs on this here:
- https://docs.google.com/document/d/1RF4D-60cQJWR1LoQeLBxxigrxJwYS8nLOE0qWmBF1eo/edit
- https://googlechrome.github.io/samples/web-bluetooth/get-devices.html
- https://www.chromestatus.com/feature/4797798639730688

I tested this for Chrome, Edge, Opera for desktop on Windows. I rolled out to equivalent versions for some of the others. There are a couple of questions inline too.

FYI, this is one of the missing entries listed in https://github.com/mdn/browser-compat-data/issues/7585

PS When this is done will need to add compatibility table to https://developer.mozilla.org/en-US/docs/Web/API/Bluetooth/getDevices

### DIFF
--- a/api/Bluetooth.json
+++ b/api/Bluetooth.json
@@ -269,14 +269,7 @@
               "version_added": false
             },
             "webview_android": {
-              "version_added": "85",
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "#enable-experimental-web-platform-features",
-                  "value_to_set": "enabled"
-                }
-              ]
+              "version_added": false
             }
           },
           "status": {
--- a/api/Bluetooth.json
+++ b/api/Bluetooth.json
@@ -203,6 +203,89 @@
           }
         }
       },
+      "getDevices": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/Bluetooth/getDevices",
+          "support": {
+            "chrome": {
+              "version_added": "85",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "#enable-experimental-web-platform-features",
+                  "value_to_set": "enabled"
+                }
+              ]
+            },
+            "chrome_android": {
+              "version_added": "85"
+            },
+            "edge": {
+              "version_added": "85",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "#enable-experimental-web-platform-features",
+                  "value_to_set": "enabled"
+                }
+              ]
+            },
+            "firefox": {
+              "version_added": false
+            },
+            "firefox_android": {
+              "version_added": false
+            },
+            "ie": {
+              "version_added": false
+            },
+            "opera": {
+              "version_added": "85",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "#enable-experimental-web-platform-features",
+                  "value_to_set": "enabled"
+                }
+              ]
+            },
+            "opera_android": {
+              "version_added": "61",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "#enable-experimental-web-platform-features",
+                  "value_to_set": "enabled"
+                }
+              ]
+            },
+            "safari": {
+              "version_added": false
+            },
+            "safari_ios": {
+              "version_added": false
+            },
+            "samsunginternet_android": {
+              "version_added": false
+            },
+            "webview_android": {
+              "version_added": "85",
+              "flags": [
+                {
+                  "type": "preference",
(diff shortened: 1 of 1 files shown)


api/Bluetooth.json
```diff
@@ -203,6 +203,89 @@
...
+                  "type": "preference",
+                  "name": "#enable-experimental-web-platform-features",
+                  "value_to_set": "enabled"
+                }
+              ]
+            }
+          },
+          "status": {
+            "experimental": true,
+            "standard_track": false,
```
hamishwillee, line 284: @ddbeck The parent `Bluetooth` and friends had `true` for this. However the spec https://webbluetoothcg.github.io/web-bluetooth/ seems to indicate it is not - " It is not a W3C Standard nor is it on the W3C Standards Track. " - perhaps I misunderstand

api/Bluetooth.json
```diff
@@ -203,6 +203,89 @@
...
+              "version_added": false
+            },
+            "safari_ios": {
+              "version_added": false
+            },
+            "samsunginternet_android": {
+              "version_added": false
+            },
+            "webview_android": {
+              "version_added": "85",
```
hamishwillee, line 272: Note, this one I did not test (don't know how to) and whether the flag is relevant. Thoughts @ddbeck ?

api/Bluetooth.json
```diff
@@ -203,6 +203,89 @@
...
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "#enable-experimental-web-platform-features",
+                  "value_to_set": "enabled"
+                }
+              ]
+            },
+            "chrome_android": {
+              "version_added": "85"
```
ddbeck, line 221: Is this actually not behind a preference on Android, but is behind a preference on desktop?

api/Bluetooth.json
ddbeck: WebView doesn't support any of Chrome's flags. So if this is behind a flag on Chrome for Android, then it's likely to be `false` for WebView (this is also the cause of the linter error—the parent feature is `false` for WebView).

api/Bluetooth.json
ddbeck: Historically, there was some uncertainty for a bit about whether we should accept WICG proposals as standards track or not. But I think we have consensus now that we should take specs at their word when they say "It is not a W3C Standard nor is it on the W3C Standards Track".

Want to open a follow-up PR or issue to fix these values? 😄

hamishwillee: @ddbeck I hate you :-) . Done in #8012

api/Bluetooth.json
hamishwillee: @ddbeck This IS behind a preference on desktop (or at least the ones I tested - chrome, opera, edge).  I _do not know_ whether it is behind a preference on android (links don't say). How do we find out?

api/Bluetooth.json
hamishwillee: OK, I have made this false.

api/Bluetooth.json
ddbeck: I ended up [testing this](https://codepen.io/ddbeck/pen/ZEpyvLa?editors=1111) on my phone. In Chrome 87, this is still behind the flag.

api/Bluetooth.json
```diff
@@ -203,6 +203,82 @@
...
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "#enable-experimental-web-platform-features",
+                  "value_to_set": "enabled"
+                }
+              ]
+            },
+            "opera_android": {
+              "version_added": "61",
```
ddbeck, line 253: Sorry I missed this previously. We ought to set this to `false`, actually. Opera for Android doesn't support flags, last I checked. We really ought to do some enforcement of this.

ddbeck (changes requested):

api/Bluetooth.json
hamishwillee: Thanks! Fixed

api/Bluetooth.json
hamishwillee: Fixed (set to false)

hamishwillee:
Thanks @ddbeck , I think this addresses your comments now

ddbeck (approved):
Looks great. Thank you! 🎉

PR #7857, opened 2020-12-15
Add Safari versions for CountQueuingStrategy API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `CountQueuingStrategy` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/CountQueuingStrategy

### DIFF
--- a/api/CountQueuingStrategy.json
+++ b/api/CountQueuingStrategy.json
@@ -125,10 +125,10 @@
               "version_added": "43"
             },
             "safari": {
-              "version_added": false
+              "version_added": "10.1"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "10.3"
             },
             "samsunginternet_android": {
               "version_added": "7.0"
--- a/api/CountQueuingStrategy.json
+++ b/api/CountQueuingStrategy.json
@@ -53,10 +53,10 @@
             "version_added": "43"
           },
           "safari": {
-            "version_added": null
+            "version_added": "10.1"
           },
           "safari_ios": {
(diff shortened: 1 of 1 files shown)


api/CountQueuingStrategy.json
```diff
@@ -125,10 +125,10 @@
               "version_added": "43"
             },
             "safari": {
-              "version_added": null
+              "version_added": false
```
foolip, line 128: `new CountQueuingStrategy({ highWaterMark: 10 })` works for me in Safari.

The result of https://mdn-bcd-collector.appspot.com/tests/api/CountQueuingStrategy/CountQueuingStrategy in Safari 14 is null, so why was this entry set to false? Is the update-bcd script now treating null as false? If so I'll need to pay a lot more attention to any false values being added.

foolip (changes requested):

api/CountQueuingStrategy.json
foolip: I've filed https://github.com/foolip/mdn-bcd-collector/issues/883 about this.

foolip: Since the constructor was there since the interface was added in https://trac.webkit.org/changeset/188127/webkit, I'll update this to match the version when the interface was exposed.

foolip (approved):

PR #7863, opened 2020-12-15
Add Safari versions for MimeType API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `MimeType` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/MimeType

### DIFF
--- a/api/MimeType.json
+++ b/api/MimeType.json
@@ -29,10 +29,11 @@
             "version_added": true
           },
           "safari": {
-            "version_added": true
+            "version_added": "≤4"
           },
           "safari_ios": {
-            "version_added": true
+            "version_added": "≤3",
+            "version_removed": "8"
           },
           "samsunginternet_android": {
             "version_added": true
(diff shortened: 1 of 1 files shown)


api/MimeType.json
```diff
@@ -29,10 +29,11 @@
             "version_added": true
           },
           "safari": {
-            "version_added": true
+            "version_added": "≤4"
           },
           "safari_ios": {
-            "version_added": true
+            "version_added": "≤3",
+            "version_removed": "8"
```
foolip, line 36: This makes sense. No plugin support on iOS.

foolip (dismissed):

api/MimeType.json
foolip: Hmm, but then the same doesn't apply for `MimeTypeArray` in https://github.com/mdn/browser-compat-data/pull/7864, that's weird. I'm guessing that `navigator.mimeTypes` still exists, but that it's always an empty array? That would make sense to minimize the risk of breaking existing web content. Can you confirm if that's the case in Safari 8 or newer?

foolip: I tried https://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8805 on Safari iOS 10 and indeed `navigator.mimeTypes` exists but is empty. So this seems right.

foolip (approved):

PR #7868, opened 2020-12-15
Add Safari versions for PushSubscriptionOptions API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `PushSubscriptionOptions` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/PushSubscriptionOptions.json
+++ b/api/PushSubscriptionOptions.json
@@ -29,10 +29,10 @@
             "version_added": true
           },
           "safari": {
-            "version_added": null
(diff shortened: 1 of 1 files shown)


foolip (approved):
By the name I was pretty sure this was a dictionary, but it's not, it's an interface: https://www.w3.org/TR/push-api/

PR #7885, opened 2020-12-15
Update Firefox data for Gamepad extensions

vinyldarkscratch:
This PR updates the data for Gamepad features behind the `dom.gamepad-extensions.enabled` flag in Firefox, using results from the mdn-bcd-collector project (as well as the existing notes).  These features were enabled by default in Firefox 55; as such, this PR replaces the data with `version_added: "55"`, thus removing the now-irrelevant flag data (since it has been enabled by default over two years ago).

### DIFF
--- a/api/Gamepad.json
+++ b/api/Gamepad.json
@@ -462,18 +462,10 @@
               "version_removed": "79"
             },
             "firefox": {
-              "version_added": true,
-              "flags": [
-                {
-                  "type": "preference",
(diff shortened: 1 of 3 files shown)


foolip (approved):
Thanks for the cleanup, much better!

I confirmed that http://mdn-bcd-collector.appspot.com/tests/api/GamepadPose passes on Firefox for Android, to be sure that data was not merely being mirrored, since it was previously set to false.

PR #7888, opened 2020-12-15
Added in manifest info

Yash-Singh1:
- [x] Summarize your changes

# Data

Look at the commit messages for sources

### DIFF
--- a/html/manifest/scope.json
+++ b/html/manifest/scope.json
@@ -7,7 +7,7 @@
           "spec_url": "https://w3c.github.io/manifest/#scope-member",
           "support": {
             "chrome": {
-              "version_added": null
+              "version_added": "53"
             },
             "chrome_android": {
               "version_added": null
--- a/html/manifest/scope.json
+++ b/html/manifest/scope.json
@@ -10,7 +10,7 @@
               "version_added": "53"
             },
             "chrome_android": {
-              "version_added": null
+              "version_added": "53"
             },
             "edge": {
               "version_added": null
--- a/html/manifest/description.json
+++ b/html/manifest/description.json
@@ -34,7 +34,7 @@
               "version_added": null
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "11.3"
             },
             "samsunginternet_android": {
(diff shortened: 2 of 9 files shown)


ddbeck:
Hi @Yash-Singh1 and thanks for your patience while we got around to reviewing this PR.

Looking this over, I don't think route to a PR is going to work. For many of these features, it looks like the caniuse data is reporting that the current version is supported, not a true version added value (e.g., some of the features that are version 87 in your PR are now shown as 88). This data also appears to be intermingled with BCD's own data.

To move these details forward, I think we need some independent source of information about the features' support, such as Chrome bugs or commits or direct tests. If you'd like to work on updating these values and citing sources for them in this PR, then we can leave this open while you work on that, or we can close it until another opportunity arises.

Thank you!

Yash-Singh1:
@ddbeck I just updated the PR. I wasn’t able to find samsung and webkit data though, so I just removed those.

html/manifest/scope.json
```diff
@@ -7,7 +7,7 @@
           "spec_url": "https://w3c.github.io/manifest/#scope-member",
           "support": {
             "chrome": {
-              "version_added": null
+              "version_added": "53"
             },
             "chrome_android": {
               "version_added": null
```
ddbeck, line 13: Is there a reason this was set only for Chrome and not Chrome for Android? This is different from the others in this PR.

ddbeck (approved):
Thank for filling in all the details via the commit messages, @Yash-Singh1! Looks good. 🎉

ddbeck (commented):
Oh wait, I forgot I had one question about the scope data. 🤦  Just need to answer that before we go ahead with merging this.

html/manifest/scope.json
Yash-Singh1: I think I just forgot to add it in

Yash-Singh1: 91851d8

ddbeck (approved):
This is a really well-developed update. Thank you! 🎉

PR #7893, opened 2020-12-15
Add Firefox versions for MutationRecord API

vinyldarkscratch:
This PR adds real values for Firefox and Firefox Android for the `MutationRecord` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/MutationRecord.json
+++ b/api/MutationRecord.json
@@ -14,10 +14,10 @@
             "version_added": "12"
           },
           "firefox": {
-            "version_added": true
+            "version_added": "14"
           },
           "firefox_android": {
-            "version_added": true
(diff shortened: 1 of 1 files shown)


api/MutationRecord.json
```diff
@@ -14,10 +14,10 @@
             "version_added": "12"
           },
           "firefox": {
-            "version_added": true
+            "version_added": "14"
```
foolip, line 17: This matches the data for when MutationObserver was added, so makes sense!

foolip (approved):

PR #7897, opened 2020-12-15
Add Firefox versions for Request API

vinyldarkscratch:
This PR adds real values for Firefox and Firefox Android for the `Request` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Request

### DIFF
--- a/api/Request.json
+++ b/api/Request.json
@@ -1199,7 +1199,7 @@
               "version_added": "39"
             },
             "firefox_android": {
-              "version_added": "≤83"
+              "version_added": "39"
             },
             "ie": {
               "version_added": false
--- a/api/Request.json
+++ b/api/Request.json
@@ -749,7 +749,7 @@
               }
             ],
             "firefox_android": {
-              "version_added": true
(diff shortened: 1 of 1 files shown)


api/Request.json
```diff
@@ -749,7 +749,7 @@
               }
             ],
             "firefox_android": {
-              "version_added": true
+              "version_added": "39"
```
foolip, line 752: Matches when `fetch()` was added, makes sense!

api/Request.json
```diff
@@ -1054,10 +1054,10 @@
               "version_added": "15"
             },
             "firefox": {
-              "version_added": null
+              "version_added": false
```
foolip, line 1057: Seems right, `keepalive` support is tracked by https://bugzilla.mozilla.org/show_bug.cgi?id=1342484

foolip (approved):

PR #7900, opened 2020-12-16
Add Safari versions for PaymentAddress API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `PaymentAddress` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/PaymentAddress

### DIFF
--- a/api/PaymentAddress.json
+++ b/api/PaymentAddress.json
@@ -280,10 +280,12 @@
               "version_added": "44"
             },
             "safari": {
-              "version_added": null
+              "version_added": "11.1",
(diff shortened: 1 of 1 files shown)


foolip (approved):
This matches the rest of https://developer.mozilla.org/en-US/docs/Web/API/PaymentAddress#Browser_compatibility, makes sense

PR #7901, opened 2020-12-16
Add Safari versions for PaymentRequest API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `PaymentRequest` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/PaymentRequest

### DIFF
--- a/api/PaymentRequest.json
+++ b/api/PaymentRequest.json
@@ -337,7 +337,7 @@
               "version_added": "11.1"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "11.3"
             },
             "samsunginternet_android": {
               "version_added": false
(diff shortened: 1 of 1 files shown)


api/PaymentRequest.json
```diff
@@ -534,10 +534,10 @@
               "version_added": "44"
             },
             "safari": {
-              "version_added": null
+              "version_added": false
```
foolip, line 537: Indeed, no `paymentAddress` or `requestId` in https://github.com/WebKit/WebKit/blob/main/Source/WebCore/Modules/paymentrequest/PaymentRequest.idl.

foolip (approved):

Issue #7886, opened 2020-12-15
html.elements.script - Chromium does not properly implement ESM in XHTML documents


00ff0000red:
#### What information was incorrect, unhelpful, or incomplete?

To date, no Chromium-based browser "properly" implements modular JavaScript, that is, simply adding `type="module"` to a script element, when in an XHTML document.

#### What did you expect to see?

I believe that an Implementation note should be added for all Chromium-based browsers to the `type.module` section of the table, warning developers about the specific Implementation of ESM in XHTML.

In Chromium, one cannot load a module script without also having the `async` attribute present.

Therefore, in Chromium's current implementation, these two are valid:
```xhtml
<script src="./modules/main.mjs" type="module" async="async" />
<script src="./modules/main.mjs" type="module" async="async" defer="defer" />
```
While this fails:
```xhtml
<script src="./modules/main.mjs" type="module" />
```

#### Did you test this? If so, how?

The (still unresolved) Chromium bug report should be enough proof:
https://crbug.com/717643

But, for verification, I tested the Firefox, Edge, and Chrome browsers to confirm this; Firefox seems to support ESM in XHTML fine, Chrome and Edge logged an error and did not load the latter script.
For clarification, Firefox supports the "normal" loading, whereas Chrome/Edge requires the `async` attribute.

I believe that Safari does not support ESM in XHTML at all, yet have not tested it.

<details>
<summary>MDN page report details</summary>

* Query: `html.elements.script`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/script
* Report started: 2020-12-15T18:48:58.420Z

</details>

PR #7902, opened 2020-12-16
Add Chromium implementation note for loading script

00ff0000red:
Adds an implementation note warning developers about module script loading (`<script type="module">`) for XHTML documents in Chromium-based browsers.

There is a Chromium bug report about this, revealing that ESM loading is not supposed to work at all in XHTML, but it also reveals that a workaround was found and that the `nomodule` attribute does not help for feature detecting this in XHTML.
https://crbug.com/717643

They had never implemented this, so it should be safe to assume that this applies to all Chromium-based browsers, across all versions.

Can anyone check the current browsers for insurance and non Chromium browsers, i.e. Safari, Firefox?

I had tested this on a recent version of Chrome and Edge by opening an XHTML document that had contained a module script, and observed the effects.
(I had also run a few specific tests, ex: global scope `this === undefined`, and checked if the code was executed in strict mode)

I have no first-hand information regarding any other browser.

Closes https://github.com/mdn/browser-compat-data/issues/7886

- [x] Summarize your changes

### DIFF
--- a/html/elements/script.json
+++ b/html/elements/script.json
@@ -558,13 +558,16 @@
             "__compat": {
               "support": {
                 "chrome": {
-                  "version_added": "61"
+                  "version_added": "61",
+                  "notes": "Chrome does not load module scripts without the <code>async</code> attribute when the page is served as XHTML (<code>application/xhtml+xml</code>) - <a href='https://crbug.com/717643'>Chromium Issue #717643</a>"
                 },
                 "chrome_android": {
-                  "version_added": "61"
+                  "version_added": "61",
+                  "notes": "Chrome does not load module scripts without the <code>async</code> attribute when the page is served as XHTML (<code>application/xhtml+xml</code>) - <a href='https://crbug.com/717643'>Chromium Issue #717643</a>"
                 },
                 "edge": {
-                  "version_added": "16"
+                  "version_added": "16",
+                  "notes": "Chrome does not load module scripts without the <code>async</code> attribute when the page is served as XHTML (<code>application/xhtml+xml</code>) - <a href='https://crbug.com/717643'>Chromium Issue #717643</a>"
                 },
                 "firefox": [
                   {
@@ -602,10 +605,12 @@
                   "version_added": false
                 },
                 "opera": {
-                  "version_added": "48"
+                  "version_added": "48",
+                  "notes": "Chrome does not load module scripts without the <code>async</code> attribute when the page is served as XHTML (<code>application/xhtml+xml</code>) - <a href='https://crbug.com/717643'>Chromium Issue #717643</a>"
                 },
                 "opera_android": {
-                  "version_added": "45"
+                  "version_added": "45",
+                  "notes": "Chrome does not load module scripts without the <code>async</code> attribute when the page is served as XHTML (<code>application/xhtml+xml</code>) - <a href='https://crbug.com/717643'>Chromium Issue #717643</a>"
                 },
                 "safari": {
                   "version_added": "10.1"
@@ -614,10 +619,12 @@
                   "version_added": "10.3"
                 },
                 "samsunginternet_android": {
-                  "version_added": "8.0"
+                  "version_added": "8.0",
+                  "notes": "Chrome does not load module scripts without the <code>async</code> attribute when the page is served as XHTML (<code>application/xhtml+xml</code>) - <a href='https://crbug.com/717643'>Chromium Issue #717643</a>"
                 },
                 "webview_android": {
-                  "version_added": "61"
+                  "version_added": "61",
+                  "notes": "Chrome does not load module scripts without the <code>async</code> attribute when the page is served as XHTML (<code>application/xhtml+xml</code>) - <a href='https://crbug.com/717643'>Chromium Issue #717643</a>"
                 }
               },
               "status": {
--- a/html/elements/script.json
+++ b/html/elements/script.json
@@ -567,7 +567,7 @@
                 },
                 "edge": {
                   "version_added": "16",
-                  "notes": "Chrome does not load module scripts without the <code>async</code> attribute when the page is served as XHTML (<code>application/xhtml+xml</code>) - <a href='https://crbug.com/717643'>Chromium Issue #717643</a>"
+                  "notes": "Module scripts without the <code>async</code> attribute do not load when the page is served as XHTML (<code>application/xhtml+xml</code>). See <a href='https://crbug.com/717643'>bug 717643</a>."
                 },
                 "firefox": [
                   {
--- a/html/elements/script.json
+++ b/html/elements/script.json
@@ -559,11 +559,11 @@
               "support": {
                 "chrome": {
                   "version_added": "61",
-                  "notes": "Chrome does not load module scripts without the <code>async</code> attribute when the page is served as XHTML (<code>application/xhtml+xml</code>). See <a href='https://crbug.com/717643'>bug 717643</a>."
+                  "notes": "Module scripts without the <code>async</code> attribute do not load when the page is served as XHTML (<code>application/xhtml+xml</code>). See <a href='https://crbug.com/717643'>bug 717643</a>."
                 },
                 "chrome_android": {
                   "version_added": "61",
-                  "notes": "Chrome does not load module scripts without the <code>async</code> attribute when the page is served as XHTML (<code>application/xhtml+xml</code>) - <a href='https://crbug.com/717643'>Chromium Issue #717643</a>"
+                  "notes": "Module scripts without the <code>async</code> attribute do not load when the page is served as XHTML (<code>application/xhtml+xml</code>). See <a href='https://crbug.com/717643'>bug 717643</a>."
                 },
                 "edge": {
                   "version_added": "16",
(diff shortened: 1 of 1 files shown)


html/elements/script.json
```diff
@@ -558,13 +558,16 @@
             "__compat": {
               "support": {
                 "chrome": {
-                  "version_added": "61"
+                  "version_added": "61",
+                  "notes": "Chrome does not load module scripts without the <code>async</code> attribute when the page is served as XHTML (<code>application/xhtml+xml</code>) - <a href='https://crbug.com/717643'>Chromium Issue #717643</a>"
```
ddbeck, line 562: We usually format bug links more minimally, like this:

```suggestion
                  "notes": "Chrome does not load module scripts without the <code>async</code> attribute when the page is served as XHTML (<code>application/xhtml+xml</code>). See <a href='https://crbug.com/717643'>bug 717643</a>."
```

html/elements/script.json
```diff
@@ -558,13 +558,16 @@
...
                 },
                 "chrome_android": {
-                  "version_added": "61"
+                  "version_added": "61",
+                  "notes": "Chrome does not load module scripts without the <code>async</code> attribute when the page is served as XHTML (<code>application/xhtml+xml</code>) - <a href='https://crbug.com/717643'>Chromium Issue #717643</a>"
                 },
                 "edge": {
-                  "version_added": "16"
+                  "version_added": "16",
+                  "notes": "Chrome does not load module scripts without the <code>async</code> attribute when the page is served as XHTML (<code>application/xhtml+xml</code>) - <a href='https://crbug.com/717643'>Chromium Issue #717643</a>"
```
ddbeck, line 570: We also try to avoid mentioning unrelated browsers in notes. You can either rework the note to avoid the browser name (when it makes sense to do so), like this:

```suggestion
                  "notes": "Module scripts without the <code>async</code> attribute do not load when the page is served as XHTML (<code>application/xhtml+xml</code>). See <a href='https://crbug.com/717643'>bug 717643</a>."
```

(this is nice, as you can copy and paste to several notes)

Or you can substitute the correct browser name, like this:


```suggestion
                  "notes": "Edge does not load module scripts without the <code>async</code> attribute when the page is served as XHTML (<code>application/xhtml+xml</code>). See <a href='https://crbug.com/717643'>bug 717643</a>."
```

ddbeck (changes requested):
Thanks for this PR. I see it's your first to the repo—welcome to BCD! I've made a few style suggestions in line comments, but otherwise, this looks like a good change. Thank you!

html/elements/script.json
00ff0000red: My apologies, I had actually more or less copied the format that some of the implementation notes for "defer" had, should those be restyled too?

html/elements/script.json
00ff0000red: I'm not sure about how anyone else would like it formatted, but I feel like saying "Chromium" (I'm not sure why I used Chrome everywhere though) would more accurately reflect the underlying reason behind these particular browsers not supporting this feature.

It's not that a particular browser didn't implement it, instead, it's because they all use the same software as a frame, which doesn't support it.

But, if you disagree, I could commit one of your suggestions instead.

00ff0000red:
I currently don't have access to another browser, or a site like [browserstack.com](https://browserstack.com/), but it would be great to get data on those last two browsers that I had mentioned (Firefox and Safari).

html/elements/script.json
ddbeck: Sorry, I didn't realize you had copied from elsewhere. There's no need to change the old notes in this PR.

html/elements/script.json
ddbeck: Early in this project's history, there was was decision made to avoid engine names where possible and use the names of browsers. Apparently, these names are more widely recognized among developers than the engines. Thanks for your understanding on this.

ddbeck:
> I currently don't have access to another browser, nor do I have access to a site like browserstack.com, but it would be great to get data on those last two browsers that I had mentioned (Firefox and Safari).

I have access to some testing tools. I'm happy to test this, if you can point me toward a simple test case.

html/elements/script.json
```diff
@@ -558,13 +558,16 @@
...
+                  "notes": "Module scripts without the <code>async</code> attribute do not load when the page is served as XHTML (<code>application/xhtml+xml</code>). See <a href='https://crbug.com/717643'>bug 717643</a>."
                 },
                 "chrome_android": {
-                  "version_added": "61"
+                  "version_added": "61",
+                  "notes": "Module scripts without the <code>async</code> attribute do not load when the page is served as XHTML (<code>application/xhtml+xml</code>). See <a href='https://crbug.com/717643'>bug 717643</a>."
                 },
                 "edge": {
-                  "version_added": "16"
+                  "version_added": "16",
```
ddbeck, line 569: Speaking of testing, I'd be good to double-check this point, too. It's possible the behavior of EdgeHTML (Edge <79) is different from Chromium-based Edge (Edge ≥79)

00ff0000red: > I have access to some testing tools. I'm happy to test this, if you can point me toward a simple test case.

Oh sorry for not getting back to you sooner, I'll probably just link the test case that I had sent to Chromium previously.
I should have it sent within a few minutes (I might take a few hours though).

00ff0000red:
@ddbeck I'm a bit late, but I just threw together a quick test that should log a message to the browser console for all of the three conditions:
* browser can load normal scripts (if this fails, we can ignore it)
* browser can load module scripts (normally)
* browser can load module scripts asynchronously

The browsers that I initially noted in my changes could load scripts asynchronously, but not "normally."

Just open this web page: https://38k3l.csb.app/index.xhtml
and check the console.

I threw in a few quick checks to make sure that the code is executed as a module, if any logs say `false`, then we have a whole different problem.

html/elements/script.json
ddbeck: Upon testing: EdgeHTML fully supported this and Chromium Edge has the Chrome behavior. So we'll need an array of support statements here:

1. `"version_added": "79"` with the same note as Chrome
2. `"version_added": "16"` and `"version_removed": "79"` (no note)

html/elements/script.json
```diff
@@ -558,13 +558,16 @@
...
+                  "notes": "Module scripts without the <code>async</code> attribute do not load when the page is served as XHTML (<code>application/xhtml+xml</code>). See <a href='https://crbug.com/717643'>bug 717643</a>."
                 },
                 "firefox": [
```
ddbeck, line 572: Upon testing: Firefox data seems fine

html/elements/script.json
```diff
@@ -602,10 +605,12 @@
                   "version_added": false
                 },
                 "opera": {
```
ddbeck, line 607: (I'm assuming that Opera, Samsung, and WebView are the same as Chrome.)

html/elements/script.json
```diff
@@ -602,10 +605,12 @@
...
-                  "version_added": "48"
+                  "version_added": "48",
+                  "notes": "Module scripts without the <code>async</code> attribute do not load when the page is served as XHTML (<code>application/xhtml+xml</code>). See <a href='https://crbug.com/717643'>bug 717643</a>."
                 },
                 "opera_android": {
-                  "version_added": "45"
+                  "version_added": "45",
+                  "notes": "Module scripts without the <code>async</code> attribute do not load when the page is served as XHTML (<code>application/xhtml+xml</code>). See <a href='https://crbug.com/717643'>bug 717643</a>."
                 },
                 "safari": {
```
ddbeck, line 615: From testing, Safari was the surprise: it appears that modules are not supported in XHTML at all. Only the `text/javascript` script ran. I expect that this is also true for Safari for iOS. Perhaps we could add a note like `Module scripts do not load when the page is served as XHTML (<code>application/xhtml+xml</code>).` to both browsers

ddbeck (changes requested):
Thank you for the test case, @00ff0000red. That's perfect. I used it to test in Sauce Labs and BrowserStack. Results in line comments.

html/elements/script.json
00ff0000red: Interesting, EdgeHTML supported this, I'm surprised! Updating as appropriate.
As for Firefox, it's good to see that they haven't regressed since I last checked.

html/elements/script.json
00ff0000red: I would *like* to just presume the same applies to Opera, Samsung, and WebView, but, it seems that caniuse.com and MDN have a discrepancy regarding Opera's module support, and I remember quite literally, telling someone that WebView didn't support modules only a few months ago, maybe I was misinformed?

Regardless, it's probably safe to make the assumption.

html/elements/script.json
00ff0000red: Actually, this was no surprise to me, thanks for confirming!
I've had someone say one of my sites didn't work before, and it happened to be that they used Mac & Safari.

ddbeck (changes requested):
One more thing to adjust here, then we can merge this, I think. Thank you, @00ff0000red.

html/elements/script.json
```diff
@@ -558,14 +558,23 @@
...
                 },
+                "edge": [
+                  {
+                    "version_added": "16",
+                    "version_removed": "79"
+                  },
+                  {
+                    "version_added": "79",
+                    "notes": "Module scripts without the <code>async</code> attribute do not load when the page is served as XHTML (<code>application/xhtml+xml</code>). See <a href='https://crbug.com/717643'>bug 717643</a>."
+                  }
```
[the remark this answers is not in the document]
00ff0000red: Oh, I was thinking like I normally do, think how JavaScript's `Array#push` adds to the end, not the start.

I'll do it right now.

00ff0000red:
@ddbeck You're welcome, and thank you for helping get the specifics of Safari's implementation, and thanks to MDN for getting this data out there!

ddbeck (approved):
This looks ready now. Thank you, @00ff0000red! 🎉

PR #7903, opened 2020-12-16
Add Safari versions for PerformanceServerTiming API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `PerformanceServerTiming` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/PerformanceServerTiming

### DIFF
--- a/api/PerformanceServerTiming.json
+++ b/api/PerformanceServerTiming.json
@@ -29,10 +29,10 @@
             "version_added": "47"
           },
           "safari": {
-            "version_added": null
+            "version_added": false
           },
           "safari_ios": {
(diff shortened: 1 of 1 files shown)


foolip (approved):
Interesting, https://github.com/WebKit/WebKit/blob/main/Source/WebCore/page/PerformanceServerTiming.idl exists, but indeed doesn't seem to be enabled in Safari 14.

PR #7904, opened 2020-12-16
Correct Safari data for PerformanceTiming API

vinyldarkscratch:
This PR corrects the Safari data for the PerformanceTiming API based upon results from the mdn-bcd-collector project.

### DIFF
--- a/api/PerformanceTiming.json
+++ b/api/PerformanceTiming.json
@@ -32,7 +32,7 @@
             "version_added": "8"
           },
           "safari_ios": {
-            "version_added": "8"
+            "version_added": "9"
           },
           "samsunginternet_android": {
             "version_added": "1.0"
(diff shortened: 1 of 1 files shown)


api/PerformanceTiming.json
```diff
@@ -988,10 +988,10 @@
               "version_added": "32"
             },
             "safari": {
-              "version_added": true
+              "version_added": "10.1"
```
foolip, line 991: This (`toJSON`) is the only outlier in this PR. Indeed it was added later, in https://trac.webkit.org/changeset/205161/webkit, and the timing (Aug 29, 2016 ) is right for 10.1. (It was slightly before 10 was released, but that must have been branched months earlier.)

foolip (approved):

PR #7906, opened 2020-12-16
Add Safari versions for SVGFilterPrimitiveStandardAttributes API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `SVGFilterPrimitiveStandardAttributes` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.


foolip (changes requested):
This is a case of https://github.com/mdn/browser-compat-data/issues/7752

PR #7908, opened 2020-12-16
Add Safari versions for SVGLengthList API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `SVGLengthList` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/SVGLengthList

### DIFF
--- a/api/SVGLengthList.json
+++ b/api/SVGLengthList.json
@@ -29,10 +29,10 @@
             "version_added": "15"
           },
           "safari": {
-            "version_added": true
+            "version_added": "5"
           },
           "safari_ios": {
-            "version_added": true
+            "version_added": "4"
           },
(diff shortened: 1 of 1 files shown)


api/SVGLengthList.json
```diff
@@ -29,10 +29,10 @@
             "version_added": "15"
           },
           "safari": {
-            "version_added": true
+            "version_added": "5"
```
foolip, line 32: Matches when `SVGAnimatedLengthList` was introduced, so makes sense.

foolip (approved):

PR #7913, opened 2020-12-16
Add Safari versions for SVGStringList API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `SVGStringList` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/SVGStringList

### DIFF
--- a/api/SVGStringList.json
+++ b/api/SVGStringList.json
@@ -31,10 +31,10 @@
             "version_added": "14"
           },
           "safari": {
-            "version_added": true
+            "version_added": "5"
           },
           "safari_ios": {
-            "version_added": true
+            "version_added": "4"
           },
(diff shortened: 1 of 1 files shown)


api/SVGStringList.json
```diff
@@ -77,10 +77,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": null
+              "version_added": "13.1"
```
foolip, line 80: This is `length`, I don't trust this data. Needs to test an instance to be sure.

foolip (approved):

PR #7920, opened 2020-12-16
Add Safari versions for PannerNode API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `PannerNode` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/PannerNode

### DIFF
--- a/api/PannerNode.json
+++ b/api/PannerNode.json
@@ -32,7 +32,7 @@
             "version_added": "6"
           },
           "safari_ios": {
-            "version_added": null
+            "version_added": "6"
           },
           "samsunginternet_android": {
             "version_added": "1.0"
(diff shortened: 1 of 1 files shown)


api/PannerNode.json
```diff
@@ -79,10 +79,10 @@
               "version_added": "42"
             },
             "safari": {
-              "version_added": null
+              "version_added": false
```
foolip, line 82: No `PannerNode` constructor seems right, that's coming soon though.

foolip (approved):

PR #7926, opened 2020-12-16
:focus-visible CSS pseudo-class is enabled in Firefox 85

rachelandrew:
Working on: https://github.com/mdn/content/issues/292

`:focus-visible` is enabled in Firefox 85, so this PR adds that information to the BCD.

Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1445482

### DIFF
--- a/css/selectors/focus-visible.json
+++ b/css/selectors/focus-visible.json
@@ -46,14 +46,24 @@
                 }
               ]
             },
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Looks good. Thank you! 🎉

PR #7927, opened 2020-12-16
Add Safari versions for StyleSheetList API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `StyleSheetList` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/StyleSheetList

### DIFF
--- a/api/StyleSheetList.json
+++ b/api/StyleSheetList.json
@@ -124,10 +124,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "10"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "10"
             },
             "samsunginternet_android": {
               "version_added": true


api/StyleSheetList.json
```diff
@@ -124,10 +124,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "10"
```
foolip, line 127: This is a `length` property and I don't trust the results without testing an instance.

foolip (changes requested):

api/StyleSheetList.json
vinyldarkscratch: Ack, didn't mean to create this PR...

PR #7928, opened 2020-12-16
Update Safari versions for TextTrack API

vinyldarkscratch:
This PR updates and corrects the real values for Safari (Desktop and iOS/iPadOS) for the `TextTrack` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/TextTrack

### DIFF
--- a/api/TextTrack.json
+++ b/api/TextTrack.json
@@ -657,7 +657,7 @@
               "version_added": "8"
             },
             "safari_ios": {
-              "version_added": "14"
+              "version_added": "13"
             },
             "samsunginternet_android": {
               "version_added": false
--- a/api/TextTrack.json
+++ b/api/TextTrack.json
@@ -269,7 +269,7 @@
               "version_added": false
             },
             "safari": {
-              "version_added": "6"
+              "version_added": "7"
             },
             "safari_ios": {
               "version_added": "7"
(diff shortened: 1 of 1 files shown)


api/TextTrack.json
```diff
@@ -657,7 +657,7 @@
               "version_added": "8"
             },
             "safari_ios": {
-              "version_added": "13"
+              "version_added": "14"
```
foolip, line 660: This is the `sourceBuffer` attribute. Is `partial_implementation` and a note about iPadOS as in https://github.com/mdn/browser-compat-data/blob/master/api/SourceBuffer.json needed here too, or is it actually exposed on iOS? If it is, I would assume it doesn't really do anything.

vinyldarkscratch: This is weird, it's exposed on both iOS and iPadOS 13...  What is going on with the versions?

foolip: Was this a bug with the update-bcd script? If so can you file a bug and investigate so it doesn't happen again?

foolip (approved):

PR #7929, opened 2020-12-16
Update Safari versions for TextTrackList API

vinyldarkscratch:
This PR updates and corrects the real values for Safari (Desktop and iOS/iPadOS) for the `TextTrackList` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/TextTrackList.json
+++ b/api/TextTrackList.json
@@ -225,7 +225,7 @@
               "version_added": "6"
             },
             "safari_ios": {
-              "version_added": "6.1"
+              "version_added": "7"
             },
             "samsunginternet_android": {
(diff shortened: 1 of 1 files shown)


api/TextTrackList.json
```diff
@@ -222,10 +222,10 @@
               "version_added": "32"
             },
             "safari": {
-              "version_added": "10"
+              "version_added": "6"
```
foolip, line 225: `length` property fixes, yes!

foolip (approved):

PR #7930, opened 2020-12-16
Add Safari versions for URL API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `URL` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/URL

### DIFF
--- a/api/URL.json
+++ b/api/URL.json
@@ -314,10 +314,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "7"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "7"
             },
             "samsunginternet_android": {
               "version_added": true
@@ -365,10 +365,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "7"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "7"
             },
             "samsunginternet_android": {
               "version_added": true
@@ -970,10 +970,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "11"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "11"
             },
             "samsunginternet_android": {
               "version_added": "10.0"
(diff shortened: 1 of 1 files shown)


api/URL.json
```diff
@@ -970,10 +970,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "11"
```
foolip, line 973: https://mdn-bcd-collector.appspot.com/tests/api/URL/toJSON tests the prototype only. I think this is probably right still, but I'd be more comfortable with testing an instance, which is easy in this case.

api/URL.json
```diff
@@ -1018,10 +1018,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "6.1"
```
foolip, line 1021: `webkitURL` was added in Safari 6, and I'd expect `toString` to be there from the start. Could you look double check this with an instance / custom test?

vinyldarkscratch: This has an instance test, actually, but it may need its own custom test.

vinyldarkscratch: Hmm...actually, it looks like this data is correct.  `.toString()` in Safari 6 returns `[object URL]`, but in Safari 6.2 (6.1) it returns `https://mdn-bcd-collector.appspot.com` (given `new webkitURL('https://mdn-bcd-collector.appspot.com')`).

api/URL.json
vinyldarkscratch: Although the test is only testing the prototype, this is correct; `toJSON()` isn't defined on the instance until Safari 11.

foolip: Thanks for confirming!

api/URL.json
foolip: Aha. Excellent find! Not that Safari 6 support matters any more, but this is pretty important issue, since `new webkitURL(url).toString()` is a convenience for normalizing a URL that wouldn't work then.

foolip (approved):

PR #7931, opened 2020-12-16
Update Safari versions for FontFace API

vinyldarkscratch:
This PR updates and corrects the real values for Safari (Desktop and iOS/iPadOS) for the `FontFace` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/FontFace

### DIFF
--- a/api/FontFace.json
+++ b/api/FontFace.json
@@ -125,7 +125,7 @@
               "version_added": "44"
             },
             "safari": {
-              "version_added": "11"
+              "version_added": "10"
             },
             "safari_ios": {
               "version_added": "10"
--- a/api/FontFace.json
+++ b/api/FontFace.json
@@ -128,7 +128,7 @@
               "version_added": "11"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "10"
             },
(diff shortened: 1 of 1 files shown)


api/FontFace.json
```diff
@@ -128,7 +128,7 @@
               "version_added": "11"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "10"
```
foolip, line 131: Is it really right that this was added in Safari iOS 10, but in Safari 11 for macOS? That the above "11" wasn't updated to "10" suggests that, or was the desktop data not used for this PR?

vinyldarkscratch: I did some more testing, and it looks like there's some odd issue with the FontFace tests in Safari 11, so the collector didn't adjust the version for this.  I did some manual testing and confirmed it was supported since Safari 10.

foolip: Excellent, thank you!

foolip (approved):

PR #7940, opened 2020-12-16
Add Safari versions for XPathResult API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `XPathResult` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/XPathResult

### DIFF
--- a/api/XPathResult.json
+++ b/api/XPathResult.json
@@ -76,10 +76,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": "7"
+              "version_added": "≤4"
             },
             "safari_ios": {
-              "version_added": "7"
+              "version_added": "≤3"
             },
             "samsunginternet_android": {
               "version_added": true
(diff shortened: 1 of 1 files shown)


api/XPathResult.json
```diff
@@ -76,10 +76,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": null
+              "version_added": "7"
```
foolip, line 79: `invalidIteratorState` and `resultType` were in https://trac.webkit.org/browser/webkit/trunk/WebCore/xpath/XPathResult.idl?rev=14234 15 years ago, so I think this a problem of attributes on prototypes. We'll need to test an instance here.

foolip (changes requested):

foolip (approved):

PR #7942, opened 2020-12-16
Add Safari versions for Comment API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `Comment` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Comment

### DIFF
--- a/api/Comment.json
+++ b/api/Comment.json
@@ -29,10 +29,10 @@
             "version_added": true
           },
           "safari": {
-            "version_added": true
+            "version_added": "≤4"
           },
           "safari_ios": {
-            "version_added": true
+            "version_added": "≤3"
           },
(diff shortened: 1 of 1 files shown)


api/Comment.json
```diff
@@ -77,10 +77,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": null
+              "version_added": "7"
```
foolip, line 80: I thought this might be wrong, but confirmed by testing `new Comment('a')` on Safari 6 manually.

foolip (approved):

PR #7945, opened 2020-12-16
Add Safari versions for ImageData API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `ImageData` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/ImageData

### DIFF
--- a/api/ImageData.json
+++ b/api/ImageData.json
@@ -77,7 +77,7 @@
               "version_added": null
             },
             "safari": {
-              "version_added": "7"
+              "version_added": "6.1"
             },
             "safari_ios": {
               "version_added": "7"
--- a/api/ImageData.json
+++ b/api/ImageData.json
@@ -77,10 +77,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": "7"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "7"
             },
             "samsunginternet_android": {
               "version_added": "4.0"


api/ImageData.json
```diff
@@ -77,10 +77,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": "7"
```
foolip, line 80: This is a constructor and I don't trust our process for constructors given two recent cases of incorrectly inferring false. Can you verify this manually?

foolip: I've filed https://github.com/foolip/mdn-bcd-collector/issues/883 about those doubts.

vinyldarkscratch: Not sure why "7" was produced here; the constructor test is working, hmm...  I checked in Safari 6.1 with the collector, and it returns `true`.  Correcting.

api/ImageData.json
```diff
@@ -77,10 +77,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": "6.1"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "7"
```
foolip, line 83: Does the iOS version also need to be updated?

vinyldarkscratch: Just the desktop data; Safari iOS 6.1 is the same WebKit version as Safari iOS 6.

foolip: I see. A bit of a complicated version mapping :)

foolip (approved):

PR #7946, opened 2020-12-16
Add Safari versions for MutationRecord API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `MutationRecord` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/MutationRecord.json
+++ b/api/MutationRecord.json
@@ -29,10 +29,10 @@
             "version_added": true
           },
           "safari": {
-            "version_added": true
+            "version_added": "7"
           },
           "safari_ios": {
-            "version_added": true
+            "version_added": "7"
           },
           "samsunginternet_android": {
             "version_added": true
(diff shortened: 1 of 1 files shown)


api/MutationRecord.json
```diff
@@ -29,10 +29,10 @@
             "version_added": true
           },
           "safari": {
-            "version_added": true
+            "version_added": "7"
```
foolip, line 32: Going by https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver#Browser_compatibility it seems more likely this was introduced in Safari 6, but perhaps the interface wasn't exposed then?

vinyldarkscratch: It looks like `MutationObserver` was supported in Safari 6 behind the `WebKit` prefix, so it's probably the same case here.  Digging deeper into this.

vinyldarkscratch: Hmm...nope, not a prefix.  It _is_ supported in Safari 6.1, though.  I wonder why "6.2" keeps getting translated into "7"...

foolip: BCD doesn't have 6.2 as a version. The version matching logic should map it to 6.1, but maybe there's a bug?

foolip (approved):

PR #7947, opened 2020-12-16
Add Safari versions for Notation API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `Notation` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Notation

### DIFF
--- a/api/Notation.json
+++ b/api/Notation.json
@@ -29,10 +29,12 @@
             "version_added": null
           },
           "safari": {
-            "version_added": null
+            "version_added": "≤4",
+            "version_removed": "9"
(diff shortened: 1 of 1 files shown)


foolip (approved):
Since these are removed and the entry should be removed I'll blindly accept the `version_added` data, although I think it's likely this was earlier. It won't matter in the end.

PR #7948, opened 2020-12-16
Add Safari versions for Notification API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `Notification` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Notification

### DIFF
--- a/api/Notification.json
+++ b/api/Notification.json
@@ -258,7 +258,7 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "11"
             },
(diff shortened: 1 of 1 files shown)


foolip (approved):
This makes sense together with https://github.com/mdn/browser-compat-data/pull/7949.

PR #7950, opened 2020-12-16
Update Safari versions for Path2D API

vinyldarkscratch:
This PR updates and corrects the real values for Safari (Desktop and iOS/iPadOS) for the `Path2D` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Path2D

### DIFF
--- a/api/Path2D.json
+++ b/api/Path2D.json
@@ -29,10 +29,10 @@
             "version_added": "24"
           },
           "safari": {
-            "version_added": "10"
+            "version_added": "7"
           },
           "safari_ios": {
-            "version_added": true
+            "version_added": "7"
           },
           "samsunginternet_android": {
             "version_added": "3.0"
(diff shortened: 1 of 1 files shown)


api/Path2D.json
```diff
@@ -78,10 +78,10 @@
               "version_added": "24"
             },
             "safari": {
-              "version_added": "10"
+              "version_added": "7"
```
foolip, line 81: This is the constructor and the "7" made me skeptical after a bunch of other suspicious "7" values today. But this seems right in manual testing, and the rest of the data makes sense in light of this.

foolip (approved):

PR #7951, opened 2020-12-16
Update Safari versions for PeriodicWave API

vinyldarkscratch:
This PR updates and corrects the real values for Safari (Desktop and iOS/iPadOS) for the `PeriodicWave` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/PeriodicWave

### DIFF
--- a/api/PeriodicWave.json
+++ b/api/PeriodicWave.json
@@ -29,10 +29,10 @@
             "version_added": "14"
           },
           "safari": {
-            "version_added": "6"
+            "version_added": "7"
           },
           "safari_ios": {
-            "version_added": true
+            "version_added": "7"
           },
           "samsunginternet_android": {
             "version_added": "1.0"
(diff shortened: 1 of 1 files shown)


api/PeriodicWave.json
```diff
@@ -29,10 +29,10 @@
             "version_added": "14"
           },
           "safari": {
-            "version_added": "6"
+            "version_added": "7"
```
foolip, line 32: The original 6 makes more sense that's when `createPeriodicWave` and `OscillatorNode` introduced. Was this a case of the interface object not being exposed?

vinyldarkscratch: Oh, oops... I could have sworn there was a custom test for this, my bad!

foolip: We have custom tests for lots of audio nodes, so it's easy to get them mixed up :)

PR #7952, opened 2020-12-16
Add Safari versions for PluginArray API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `PluginArray` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/PluginArray.json
+++ b/api/PluginArray.json
@@ -29,10 +29,10 @@
             "version_added": false
           },
           "safari": {
-            "version_added": true
+            "version_added": "≤4"
           },
           "safari_ios": {
-            "version_added": true
+            "version_added": "≤3"
           },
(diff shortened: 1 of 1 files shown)


api/PluginArray.json
```diff
@@ -128,10 +128,10 @@
               "version_added": false
             },
             "safari": {
-              "version_added": true
+              "version_added": "10"
```
foolip, line 131: A `length` property introduced much later than the interface. I don't believe it :)

foolip (changes requested):

foolip (approved):

PR #7954, opened 2020-12-16
Add Safari versions for Text API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `Text` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Text

### DIFF
--- a/api/Text.json
+++ b/api/Text.json
@@ -26,10 +26,10 @@
             "version_added": true
           },
           "safari": {
-            "version_added": true
+            "version_added": "≤4"
           },
           "safari_ios": {
-            "version_added": true
+            "version_added": "≤3"
           },
           "webview_android": {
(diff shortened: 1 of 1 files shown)


api/Text.json
```diff
@@ -224,10 +224,12 @@
               "version_removed": "32"
             },
             "safari": {
-              "version_added": false
+              "version_added": "≤4",
+              "version_removed": "10.1"
```
foolip, line 228: Yep, that happened: https://trac.webkit.org/changeset/204236/webkit

foolip (approved):

PR #7955, opened 2020-12-16
Add Safari versions for TimeRanges API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `TimeRanges` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/TimeRanges

### DIFF
--- a/api/TimeRanges.json
+++ b/api/TimeRanges.json
@@ -29,10 +29,10 @@
             "version_added": true
           },
           "safari": {
-            "version_added": "5.1"
+            "version_added": "≤4"
           },
           "safari_ios": {
-            "version_added": "5"
+            "version_added": "≤3"
           },
           "samsunginternet_android": {
             "version_added": true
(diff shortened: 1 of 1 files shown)


api/TimeRanges.json
```diff
@@ -124,10 +124,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "10"
```
foolip, line 127: Another `length` property, almost certainly wrong.

foolip (changes requested):

api/TimeRanges.json
```diff
@@ -29,10 +29,10 @@
             "version_added": true
           },
           "safari": {
-            "version_added": true
+            "version_added": "5.1"
```
foolip, line 32: Hmm, so https://mdn-bcd-collector.appspot.com/tests/api/TimeRanges only tests the prototype, and api.HTMLMediaElement.buffered which is a `TimeRanges` is supported since Safari 3.2. Can you add custom tests for this?

foolip (changes requested):

foolip (approved):

PR #7956, opened 2020-12-16
Add Safari versions for WebSocket API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `WebSocket` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/WebSocket.json
+++ b/api/WebSocket.json
@@ -111,10 +111,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": "5"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "4.2"
             },
             "samsunginternet_android": {
               "version_added": true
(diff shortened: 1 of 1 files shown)


api/WebSocket.json
```diff
@@ -159,10 +159,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "7"
```
foolip, line 162: I don't totally trust that all of these attributes were introduced at the same time, but after the interface itself. Can you test an instance of `WebSocket` to make sure?

vinyldarkscratch: Sure thing -- had the same issue with Chrome data for WebSocket, so I'm less trusting of the Safari data.  I'll close this PR and reopen it when we have a custom test!

PR #7958, opened 2020-12-16
updating fx status now current is 84

chrisdavidmills:
I've updated both the `firefox` and `firefox_android` data


vinyldarkscratch (approved):
Looks good to me, thanks!

PR #7963, opened 2020-12-17
Update Firefox versions for MediaError API

vinyldarkscratch:
This PR updates and corrects the real values for Firefox and Firefox Android for the `MediaError` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/MediaError

### DIFF
--- a/api/MediaError.json
+++ b/api/MediaError.json
@@ -14,10 +14,10 @@
             "version_added": "12"
           },
           "firefox": {
-            "version_added": "3.5"
+            "version_added": "4"
           },
           "firefox_android": {
-            "version_added": true
+            "version_added": "4"
           },
           "ie": {
             "version_added": "9"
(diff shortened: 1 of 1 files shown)


api/MediaError.json
```diff
@@ -14,10 +14,10 @@
             "version_added": "12"
           },
           "firefox": {
-            "version_added": "3.5"
+            "version_added": "4"
```
foolip, line 17: The original "3.5" matches the data for api.HTMLMediaElement.error which is an instance of `MediaError`. This is more like an interface object not exposed. Can you check this manually?

foolip: I've tested http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8806 in Firefox 3.6 on BrowserStack and `video.error` is there. To make sure it wasn't just always null I checked when it was added and as part of that came across https://bugzilla.mozilla.org/show_bug.cgi?id=589561#c4. `MediaError` was originally `HTMLMediaError` and `TimeRanges` was `HTMLTimeRanges`.

I've sent https://github.com/mdn/browser-compat-data/pull/8734 to fix this.

foolip:
Closing in favor of https://github.com/mdn/browser-compat-data/pull/8734.

PR #7964, opened 2020-12-17
Add Firefox versions for MediaQueryList API

vinyldarkscratch:
This PR adds real values for Firefox and Firefox Android for the `MediaQueryList` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/MediaQueryList

### DIFF
--- a/api/MediaQueryList.json
+++ b/api/MediaQueryList.json
@@ -17,7 +17,7 @@
             "version_added": "6"
           },
           "firefox_android": {
-            "version_added": true
+            "version_added": "6"
           },
           "ie": {
             "version_added": "10"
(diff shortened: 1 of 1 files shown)


api/MediaQueryList.json
```diff
@@ -17,7 +17,7 @@
             "version_added": "6"
           },
           "firefox_android": {
-            "version_added": true
+            "version_added": "6"
```
foolip, line 20: This matches the data for `window.matchMedia` which returns a `MediaQueryList`, so I believe it!

foolip (approved):

PR #7969, opened 2020-12-17
Add Chromium versions for DOMStringMap API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `DOMStringMap` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/DOMStringMap

### DIFF
--- a/api/DOMStringMap.json
+++ b/api/DOMStringMap.json
@@ -6,10 +6,10 @@
         "spec_url": "https://html.spec.whatwg.org/multipage/dom.html#domstringmap",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "7"
           },
           "chrome_android": {
-            "version_added": true
+            "version_added": "18"
           },
           "edge": {
             "version_added": "12"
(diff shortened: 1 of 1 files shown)


api/DOMStringMap.json
```diff
@@ -6,10 +6,10 @@
         "spec_url": "https://html.spec.whatwg.org/multipage/dom.html#domstringmap",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "7"
```
foolip, line 9: `DOMStringMap` is the type of `element.dataset`, and that was added in Chrome 8 per our data. It's possible the interface was accidentally shipped before the thing using it, so I won't check further. If `element.dataset` was actually shipped in Chrome 7 too, we should eventually discover that using the collector as well.

foolip (approved):

PR #7973, opened 2020-12-17
Correct Chromium data for TextTrack API; add versions for TextTrackList API

vinyldarkscratch:
This PR corrects the data for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `TextTrack` API, and adds `TextTrackList` API real values, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/TextTrackList.json
+++ b/api/TextTrackList.json
@@ -5,10 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/TextTrackList",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "23"
           },
           "chrome_android": {
-            "version_added": true
+            "version_added": "25"
           },
           "edge": {
             "version_added": "≤18"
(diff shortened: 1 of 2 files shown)


api/TextTrackList.json
```diff
@@ -5,10 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/TextTrackList",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "23"
```
foolip, line 8: This matches api.HTMLMediaElement.textTracks, but `TextTrack` says 18 which I doubt is correct. Could you re-validate that data and if it's wrong update it in this PR as well?

vinyldarkscratch: Yep, collector's saying TextTrack is supported since Chrome 23.  Updates coming up!

foolip (approved):

PR #7974, opened 2020-12-17
Add Chromium versions for TrackDefault API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `TrackDefault` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/TrackDefault


foolip (approved):
Yep, this was never shipped I think.

PR #7978, opened 2020-12-17
Add Chromium versions for URL API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `URL` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/URL

### DIFF
--- a/api/URL.json
+++ b/api/URL.json
@@ -212,7 +212,7 @@
               "version_added": "1.5"
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "≤37"
             }
           },
           "status": {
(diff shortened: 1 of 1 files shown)


foolip (approved):
https://mdn-bcd-collector.appspot.com/tests/api/URL tries both `new URL` and `new webkitURL` so I have high confidence in these results.

PR #7980, opened 2020-12-17
Add Chromium versions for XPathExpression API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `XPathExpression` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/XPathExpression

### DIFF
--- a/api/XPathExpression.json
+++ b/api/XPathExpression.json
@@ -23,10 +23,10 @@
             "version_added": false
           },
           "opera": {
-            "version_added": "15"
+            "version_added": "≤12.1"
           },
           "opera_android": {
-            "version_added": "14"
+            "version_added": "≤12.1"
           },
           "safari": {
             "version_added": null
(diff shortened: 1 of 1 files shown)


api/XPathExpression.json
```diff
@@ -23,10 +23,10 @@
             "version_added": false
           },
           "opera": {
-            "version_added": true
+            "version_added": "15"
```
foolip, line 26: I'm pretty sure this was supported in Presto. `XPathEvaluator` is the strangest thing, both an interface and mixed into `Document` at the same time. So there's actually a `document.createExpression()` method too, which our data says was supported in Presto.

Can you update the custom tests for this to try creating an instance with both methods?

foolip (changes requested):

api/XPathExpression.json
vinyldarkscratch: Ah, looks like there's a slight issue with the test, so it generated a range and I probably accidentally ignored it.  I double-checked in BrowserStack and confirmed Presto support!

foolip: Thanks!

foolip (approved):

PR #7986, opened 2020-12-17
Add Chromium versions for PluginArray API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `PluginArray` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Additionally, this adjusts a note on a few methods to match our typical wording, and copies it to all of the Chrome-based browsers.

### DIFF
--- a/api/PluginArray.json
+++ b/api/PluginArray.json
@@ -5,10 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/PluginArray",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "1"
           },
           "chrome_android": {
-            "version_added": true
+            "version_added": "18"
           },
           "edge": {
             "version_added": "12"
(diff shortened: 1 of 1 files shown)


api/PluginArray.json
```diff
@@ -23,10 +23,10 @@
             "version_added": null
           },
           "opera": {
-            "version_added": true
+            "version_added": "15"
```
foolip, line 26: I wonder if this wasn't supported in Opera Presto. `PluginArray` is the type for `navigator.plugins`, can you update that entry at the same time and see if the values are the same? I suspect that the interface object just wasn't exposed, and that this needs to test an instance, which is simply `navigator.plugins`.

vinyldarkscratch: Looks like this is a case of unexposed interface, you're right.  Adding a custom test, and manually updated the data.

foolip (approved):

PR #7988, opened 2020-12-17
Add Opera versions for TextMetrics API

vinyldarkscratch:
This PR adds real values for Opera and Opera Android for the `TextMetrics` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/TextMetrics

### DIFF
--- a/api/TextMetrics.json
+++ b/api/TextMetrics.json
@@ -94,10 +94,10 @@
               "version_added": false
             },
             "opera": {
-              "version_added": null
+              "version_added": "64"
             },
             "opera_android": {
-              "version_added": null
+              "version_added": "55"
             },
             "safari": {
               "version_added": true
@@ -166,10 +166,10 @@
               "version_added": false
             },
             "opera": {
-              "version_added": null
+              "version_added": "64"
             },
             "opera_android": {
-              "version_added": null
+              "version_added": "55"
             },
             "safari": {
               "version_added": true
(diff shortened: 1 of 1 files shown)


api/TextMetrics.json
```diff
@@ -602,10 +602,10 @@
               "version_added": false
             },
             "opera": {
-              "version_added": null
+              "version_added": "73"
```
foolip, line 605: The Chrome data suggests this is behind a flag and thus not shipped. Can you update that too if it's not correct?

api/TextMetrics.json
```diff
@@ -602,10 +602,10 @@
...
             },
             "opera_android": {
-              "version_added": null
+              "version_added": false
```
foolip, line 608: I'd be surprised if this API was different between desktop and mobile, can you verify this isn't in Chrome for Android? Then I'd be more inclined to trust this.

api/TextMetrics.json
vinyldarkscratch: #7989 adds the Chrome data, which adds versions for when it was no longer behind a flag.  Realize I should have put these changes with that PR, sorry!

api/TextMetrics.json
vinyldarkscratch: This _was_ added in Chrome Android, enabled by default in Chrome 87.  There's no Opera Android version that corresponds to Chrome 87, though, hence `false`.

foolip: Oh, I see, thanks!

api/TextMetrics.json
foolip: Aha, so Chrome 87, which is the same as Opera 73. Thanks!

foolip (approved):

Issue #7843, opened 2020-12-14
Attributes incorrectly claimed to be added in Chrome 43


foolip:
Chrome 43 fixed its bindings so that Web IDL attributes appear on the prototypes:
https://developers.google.com/web/updates/2015/04/DOM-attributes-now-on-the-prototype-chain

Because of this, tests like `'attributeName' in InterfaceName` can be misleading, suggesting that something was added in Chrome 43. This has come up a lot in @vinyldarkscratch's and my work on https://github.com/foolip/mdn-bcd-collector.

Not all cases of this have been caught in review, and `npm run traverse chrome api 43` lists a lot of features. Some of these were introduced earlier, for example api.HTMLMediaElement.volume is much older.

We need to go through any attributes claimed to be added in Chrome 43 and check if they were really available earlier.

foolip:
https://github.com/mdn/browser-compat-data/issues/7755 is similar, but for Chrome 44.

foolip:
While reviewing https://github.com/mdn/browser-compat-data/pull/7990 I spotted some that I'm almost certain are incorrect and need fixing as part of this issue, namely api.HTMLMediaElement.*. Those are all very old APIs, not introduced in Chrome 43.

ddbeck:
@foolip What's the trick to testing that attributes are supported before 43? Testing on actual instances, instead of the prototype?

foolip:
@ddbeck exactly, an instance is needed, and that's what collector fixes like https://github.com/foolip/mdn-bcd-collector/pull/1247 ends up doing.

queengooborg:
I think that we've pretty much fixed this, and any additional instances should be catchable by the collector.  I'm going to close this issue and consider it to be a part of the bigger collector synchronization goal.

PR #7990, opened 2020-12-17
Add Chromium versions for TimeRanges API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `TimeRanges` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/TimeRanges

### DIFF
--- a/api/TimeRanges.json
+++ b/api/TimeRanges.json
@@ -5,10 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/TimeRanges",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "6"
           },
           "chrome_android": {
-            "version_added": true
+            "version_added": "18"
           },
           "edge": {
             "version_added": "12"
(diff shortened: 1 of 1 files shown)


api/TimeRanges.json
```diff
@@ -5,10 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/TimeRanges",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "6"
```
foolip, line 8: This doesn't add up with the attributes that return `TimeRanges`, but commented on https://github.com/mdn/browser-compat-data/issues/7843#issuecomment-747482568 instead of asking for that to be fixed in this PR.

foolip (approved):

PR #7995, opened 2020-12-17
Add IE/Edge versions for StyleSheetList API

vinyldarkscratch:
This PR adds real values for Internet Explorer and Edge for the `StyleSheetList` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/StyleSheetList

### DIFF
--- a/api/StyleSheetList.json
+++ b/api/StyleSheetList.json
@@ -20,7 +20,7 @@
             "version_added": "31"
           },
           "ie": {
-            "version_added": "9"
+            "version_added": "8"
           },
           "opera": {
             "version_added": true
@@ -67,7 +67,7 @@
               "version_added": "31"
             },
             "ie": {
-              "version_added": "9"
+              "version_added": "8"
             },
             "opera": {
               "version_added": true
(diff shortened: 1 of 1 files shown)


foolip (changes requested):
This should match https://mdn-bcd-collector.appspot.com/tests/api/Document/styleSheets but that works in IE8 as well. So I tested https://mdn-bcd-collector.appspot.com/tests/api/StyleSheetList and that too returns true for IE8. Did something go wrong with data collection here, or was it the update-bcd script that did the wrong thing?

vinyldarkscratch:
Hmm, not sure what happened here, that's weird...

api/StyleSheetList.json
```diff
@@ -20,7 +20,7 @@
             "version_added": "31"
           },
           "ie": {
-            "version_added": null
+            "version_added": "8"
```
foolip, line 23: I tested https://mdn-bcd-collector.appspot.com/tests/api/Document/styleSheets in IE8, but no earlier version. Now I tested it and it return true back to IE6. Can you check if `document.styleSheets` is a `StyleSheetList` instance in IE6 and if so update this?

foolip (changes requested):

foolip (approved):

PR #8001, opened 2020-12-17
Add IE/Edge versions for Plugin API

vinyldarkscratch:
This PR adds real values for Internet Explorer and Edge for the `Plugin` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/Plugin.json
+++ b/api/Plugin.json
@@ -20,7 +20,7 @@
             "version_added": "4"
           },
           "ie": {
-            "version_added": null
+            "version_added": "11"
           },
           "opera": {
             "version_added": "15"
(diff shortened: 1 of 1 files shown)


api/Plugin.json
```diff
@@ -20,7 +20,7 @@
             "version_added": "4"
           },
           "ie": {
-            "version_added": null
+            "version_added": "11"
```
foolip, line 23: This should match https://github.com/mdn/browser-compat-data/pull/8002#pullrequestreview-555222775 when the correct version for `navigator.plugins` has been found. It might not be possible to reliably test an instance here though, since there might be no supported plugins.

foolip (changes requested):

vinyldarkscratch:
For ease of review, I'll merge this into #8000, as well as include data for `navigator.plugins` in it, since they all go together.

PR #8002, opened 2020-12-17
Add IE/Edge versions for Plugin and PluginArray APIs

vinyldarkscratch:
This PR adds real values for Internet Explorer and Edge for the `Plugin` and `PluginArray` APIs, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.6).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/PluginArray.json
+++ b/api/PluginArray.json
@@ -20,7 +20,7 @@
             "version_added": "4"
           },
           "ie": {
-            "version_added": "11"
+            "version_added": "≤6"
           },
           "opera": {
             "version_added": "≤12.1"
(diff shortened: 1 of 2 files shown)


api/PluginArray.json
```diff
@@ -20,7 +20,7 @@
             "version_added": "4"
           },
           "ie": {
-            "version_added": null
+            "version_added": "11"
```
foolip, line 23: This should match https://mdn-bcd-collector.appspot.com/tests/api/Navigator/plugins but that tests true in IE10. Can you test an instance (`navigator.plugins`) here?

foolip (changes requested):

foolip (approved):

Issue #7851, opened 2020-12-15
ContactPicker API only on Android?


hamishwillee:
[Contact Picker API](https://developer.mozilla.org/en-US/docs/Web/API/Contact_Picker_API) indicates support on Chrome 80+. However I think this is only supported on Android.

This is based on chrome test where I ran code on desktop:
```
  let test1 = window.ContactsManager;
  let test2 = navigator.contacts;
```
However both cases these are `undefined`. I then looked at https://web.dev/contact-picker/ and this states "only in android".

> Note, test in docs is `const supported = ('contacts' in navigator && 'ContactsManager' in window);` - but there is no indication that `window.ContactsManager` is defined other than that mention of it.

Can we confirm?

Further, we have https://github.com/mdn/browser-compat-data/blob/master/api/ContactsManager.json , but I can't find a defined attribute `navigator.contacts` in https://github.com/mdn/browser-compat-data/blob/master/api/Navigator.json . Would this just be an omission?

ddbeck:
cc: @jpmedley since you added the original `ContactsManager` data in #6277.

I suspect the missing Navigator feature is an omission (and I'd be happy to see a PR adding it).

The https://web.dev/contact-picker/ article links to [a Chrome bug](https://bugs.chromium.org/p/chromium/issues/detail?id=860467) that appears to substantiate the Android-only aspect, though I don't know if there's been any work on expanding that since or what that means for WebView. Perhaps Joe can shed some light on that part of the story.

hamishwillee:
@ddbeck Thanks very much. I've created draft PR for adding Navigator.contacts #8010 - obviously has to be draft until the above questions are resolved. 

@jpmedley in addition to the above, do you know if any additional work has happened to standardize this (currently it is marked as non-standard in the docs and there is no spec listed).

ddbeck:
There's [a spec](https://wicg.github.io/contact-api/spec/), but it's one of these WICG "not a W3C Standard nor is it on the W3C Standards Track" specifications

hamishwillee:
OK, I've added spec in https://github.com/mdn/content/pull/750 (thanks) and added BCD update for ContactsManager in 
https://github.com/mdn/browser-compat-data/pull/8446

When the BCD entries go in then I will add a note in the [API docs](https://developer.mozilla.org/en-US/docs/Web/API/Contact_Picker_API) that this is currently only enabled for android

Rumyra:
[...]

PR #8010, opened 2020-12-18
Add Navigator.contacts

hamishwillee:
This adds `Navigator.contacts`, as discussed in #7851. This is not complete - it still depends on discussion in #7851.

What I've done for now is matched the [ContactsManager BCD info](https://developer.mozilla.org/en-US/docs/Web/API/Contact_Picker_API#Browser_compatibility), but only included the android chromium builds. This might therefore be incorrect if, for example, Opera on Android did not choose to add this.

### DIFF
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -534,6 +534,54 @@
           }
         }
       },
+      "contacts": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/Navigator/contacts",
+          "support": {
+            "chrome": {
+              "version_added": false
+            },
+            "chrome_android": {
+              "version_added": "80"
+            },
+            "edge": {
+              "version_added": false
+            },
+            "firefox": {
+              "version_added": false
+            },
+            "firefox_android": {
+              "version_added": false
+            },
+            "ie": {
+              "version_added": false
+            },
+            "opera": {
+              "version_added": false
+            },
+            "opera_android": {
+              "version_added": "57"
+            },
+            "safari": {
+              "version_added": false
+            },
+            "safari_ios": {
+              "version_added": false
+            },
+            "samsunginternet_android": {
+              "version_added": "13.0"
+            },
+            "webview_android": {
+              "version_added": "80"
+            }
+          },
+          "status": {
+            "experimental": true,
+            "standard_track": false,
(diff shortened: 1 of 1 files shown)


api/Navigator.json
```diff
@@ -534,6 +534,54 @@
...
       },
+      "contacts": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/Navigator/contacts",
+          "support": {
+            "chrome": {
+              "version_added": false
+            },
+            "chrome_android": {
+              "version_added": "80"
```
foolip, line 545: Confirmed supported on Chrome for Android and *not* macOS using https://mdn-bcd-collector.appspot.com/tests/api/Navigator/contacts

Moreover, Chrome 80 is confirmed by https://storage.googleapis.com/chromium-find-releases-static/967.html#967ef324a2761a4615713e3fe152423d1715d7f6

api/Navigator.json
```diff
@@ -534,6 +534,54 @@
...
+              "version_added": false
+            },
+            "safari_ios": {
+              "version_added": false
+            },
+            "samsunginternet_android": {
+              "version_added": "13.0"
+            },
+            "webview_android": {
+              "version_added": "80"
```
foolip, line 575: This isn't a given, but the API isn't listed in https://source.chromium.org/chromium/chromium/src/+/master:android_webview/tools/system_webview_shell/test/data/webexposed/not-webview-exposed.txt;l=1;drc=cfec54548aa72c255b28526f3ff2bca676dd6dbb so it's probably right.

foolip (approved):
The data here looks correct.

ddbeck:
> it still depends on discussion in #7851

I think @foolip's comments might have satisfied this. If you agree, then this could come out of draft status. 🙂

vinyldarkscratch (approved):
+1 review -- I was just about to open up a PR for this, then I noticed this PR.  Data matches with what I found, and the status data looks good, this is looking good to me!  I also agree that we can take it out of draft status.

hamishwillee:
Thanks! Out of draft then @ddbeck . FYI I'm likely to be gone until 13th or 14th Jan now.

foolip:
Thanks @hamishwillee!

PR #8012, opened 2020-12-18
Mark Bluetooth API as non-standard track

hamishwillee:
This changes all Bluetooth API classes from `"standard_track": true,` to `false`. As discussed in #7850 the [spec](https://webbluetoothcg.github.io/web-bluetooth/) specifically states that they are non-standard.


ddbeck (approved):
I'll trade hated for done. 😈 Thank you, @hamishwillee! 🎉

Issue #7849, opened 2020-12-15
Tag everything with worker support?


hamishwillee:
Worker support of APIs is quite a challenge to [track and maintain docs](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API/Functions_and_classes_available_to_workers#APIs_available_in_workers) for -  it can vary across an API and be spotty across browsers. It also isn't necessarily obvious to a user how they go about using some APIs.

I think it might make sense to "standardise" this in BCD and roll it out more or less for every single Web API.
- There are four types of workers - dedicated, shared, service, chrome. I don't know if we'd need to indicate these separately or not. 
- It is useful to note the API used to access the feature in both main thread (often via window or window.navigator) and the web service (usually some context)

There are a few good examples where this has been done in an adhoc way:
-  `FileReaderSync` ([mdn]((https://developer.mozilla.org/en-US/docs/Web/API/FileReaderSync) | [bcd](https://github.com/mdn/browser-compat-data/blob/master/api/FileReaderSync.json)) has separate keys indicating support for service workers and for other workers.
  ![image](https://user-images.githubusercontent.com/5368500/102153342-17973200-3ecb-11eb-9d74-f121595c3fa2.png)
- `CacheStorage` ([mdn](https://developer.mozilla.org/en-US/docs/Web/API/CacheStorage)) has a note indicating how it is used in both web workers and main thread.
   ![image](https://user-images.githubusercontent.com/5368500/102153462-5d53fa80-3ecb-11eb-82c4-f2c421e2dc9d.png)

FYI @chrisdavidmills @Elchi3 @ddbeck Would appreciate your thinking on this. To me this is a similar problem to documenting secure contexts.

ddbeck:
You're right that we ought to be more consistent about this. A couple thoughts on how we might do this:

First, I suppose we should have a worker support entry for things which have been specified as having worker support (stuff with `[Exposed=(Window,Worker)]` and the like in the IDL), or things which have worker support and don't have a specification (or for features for which either of these statements was historically true). We probably don't need a `worker_support` entry for every feature. All that said, I don't think we've done a great job of reviewing for this specifically. Perhaps this might be of interest to @foolip and @vinyldarkscratch's testing efforts.

[...]

PR #8014, opened 2020-12-18
Remove service_workers_support from api/FileReaderSync

hamishwillee:
There is a [Worker support guideline](https://github.com/mdn/browser-compat-data/blob/master/docs/data-guidelines.md#web-workers-worker_support) so `service_workers_support` is incorrect. This follows on from this discussion:  https://github.com/mdn/browser-compat-data/issues/7849#issuecomment-745445095

@ddbeck The way I have done this  is to just add a note indicating that service workers are not supported (with the version, if known). 

Are there other alternatives? i.e. should/can I do multiple worker supports like this?:
```
 "worker_support": {
        "__compat": {
...

 "worker_support": {
        "__compat": {
          "description": "Available in service workers",
...
````

Or something like this (feels confusing).
```json
 "worker_support": {
        "__compat": {
          "support": {
            "chrome": [
              {
                "version_added": true,
              },
              {
                "version_added": true,
                "version_removed": "59",
                "notes": "Service workers not supported from Chrome 59"
              },
            ],
```

### DIFF
--- a/api/FileReaderSync.json
+++ b/api/FileReaderSync.json
@@ -251,57 +251,56 @@
       },
       "worker_support": {
         "__compat": {
-          "description": "Available in workers (except for service workers)",
+          "description": "Available in workers",
           "support": {
             "chrome": {
               "version_added": true,
-              "notes": "Not supported in service workers from Chrome 59"
+              "notes": "From Chrome 59, not supported in service workers."
             },
             "chrome_android": {
               "version_added": true,
-              "notes": "Not supported in service workers from Chrome 59"
+              "notes": "From Chrome 59, not supported in service workers."
             },
             "edge": {
               "version_added": "12",
-              "notes": "Not supported in service workers"
+              "notes": "Not supported in service workers."
             },
             "firefox": {
               "version_added": "8",
-              "notes": "Not supported in service workers from FF61"
+              "notes": "From Firefox 61, not supported in service workers."
             },
             "firefox_android": {
               "version_added": "8",
-              "notes": "Not supported in service workers from FF61"
+              "notes": "Form Firefox 61, not supported in service workers."
             },
             "ie": {
-              "version_added": true,
-              "notes": "Not supported in service workers"
+              "version_added": true
             },
             "opera": {
               "version_added": true,
-              "notes": "Not supported in service workers from Opera 46"
+              "notes": "From Opera 46, not supported in service workers."
             },
             "opera_android": {
               "version_added": true,
-              "notes": "Not supported in service workers from Opera 43"
+              "notes": "From Opera 43, not supported in service workers."
             },
             "safari": {
               "version_added": true,
               "version_removed": "6.1",
-              "notes": "Service workers not supported"
+              "notes": "Not supported in service workers."
             },
             "safari_ios": {
               "version_added": true,
               "version_removed": "7",
-              "notes": "Service workers not supported"
+              "notes": "Not supported in service workers."
             },
             "samsunginternet_android": {
               "version_added": true,
-              "notes": "Not supported in service workers from version 7.0"
+              "notes": "From version 7.0, not supported in service workers."
             },
             "webview_android": {
               "version_added": true,
-              "notes": "Not supported in service workers from version 59"
+              "notes": "From version 59, not supported in service workers."
             }
           },
           "status": {
--- a/api/FileReaderSync.json
+++ b/api/FileReaderSync.json
@@ -251,7 +251,7 @@
       },
       "worker_support": {
         "__compat": {
-          "description": "Available in workers",
+          "description": "Available in workers (except for service workers)",
           "support": {
             "chrome": {
               "version_added": true,
--- a/api/FileReaderSync.json
+++ b/api/FileReaderSync.json
@@ -271,7 +271,7 @@
             },
             "firefox_android": {
               "version_added": "8",
-              "notes": "Form Firefox 61, not supported in service workers."
+              "notes": "From Firefox 61, not supported in service workers."
             },
             "ie": {
               "version_added": true
--- a/api/FileReaderSync.json
+++ b/api/FileReaderSync.json
@@ -249,103 +249,59 @@
           }
         }
       },
-      "service_workers_support": {
-        "__compat": {
-          "description": "Service workers support",
-          "support": {
-            "chrome": {
-              "version_added": true,
-              "version_removed": "59"
-            },
-            "chrome_android": {
-              "version_added": true,
-              "version_removed": "59"
-            },
-            "edge": {
-              "version_added": null
-            },
-            "firefox": {
-              "version_added": true,
-              "version_removed": "61"
-            },
-            "firefox_android": {
-              "version_added": true,
-              "version_removed": "61"
-            },
-            "ie": {
-              "version_added": false
-            },
-            "opera": {
-              "version_added": true,
-              "version_removed": "46"
-            },
-            "opera_android": {
-              "version_added": true,
-              "version_removed": "43"
-            },
-            "safari": {
-              "version_added": false
-            },
-            "safari_ios": {
-              "version_added": false
-            },
-            "samsunginternet_android": {
-              "version_added": true,
-              "version_removed": "7.0"
-            },
-            "webview_android": {
-              "version_added": true,
-              "version_removed": "59"
-            }
-          },
-          "status": {
-            "experimental": false,
-            "standard_track": false,
-            "deprecated": true
-          }
-        }
-      },
       "worker_support": {
         "__compat": {
           "description": "Available in workers",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": true,
+              "notes": "Service workers not supported from Chrome 59"
             },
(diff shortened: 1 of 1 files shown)


hamishwillee:
Note, same question for caches. In chrome, for example, support was added for service workers in version 40. In version 43 support was added for other workers. 

So I what is the right way to do this - added in version 40, and a note stating that initially only for service workers. Or two worker_support lines, or one worker support line but with notes about what happened in each version?

ddbeck:
It's hard to follow the diff for removals and data changes in one commit. Can you put into words the support story you're trying to tell here? The diff and the examples given seem to be telling competing stories. What was supported (and unsupported) and in what versions did support change?

That said, I can rule out one thing: multiple `"worker_support"` features. JSON keys have to be unique, or else we'll just clobber one bit of data with another.

Data like this also seems unlikely:

```
            "chrome": [
              {
                "version_added": true,
              },
              {
                "version_added": true,
                "version_removed": "59",
                "notes": "Service workers not supported from Chrome 59"
              },
            ],
```

The second support statement says support was dropped in version 59. The note is a bit strange, since it reiterates what `"version_removed": "59"` is already saying: from version 59, the feature is unsupported (though the first support statement says that support was reintroduced in some unknown version).

hamishwillee:
> Can you put into words the support story you're trying to tell here? 

Hi @ddbeck 
All I have done (see diff) is deleted the `service_workers_support` key and captured the version information about service worker support  in a note under `worker_support`.

So to answer your question, this is supposed to say:
- workers, other than service workers are supported
- Info of versions in which there is support for workers
- Info about versions in which there _was_ support for service workers.

This is probably OK for this case but I'm not sure it is the "right way". How would YOU present for chrome  this imaginary case.
- dedicated workers are supported from version 4, 
- shared workers are supported from version 8
- service workers are supported from version 2 to 6. 
- chrome workers are not supported

api/FileReaderSync.json
```diff
@@ -249,103 +249,59 @@
           }
         }
       },
-      "service_workers_support": {
+      "worker_support": {
         "__compat": {
-          "description": "Service workers support",
+          "description": "Available in workers (except for service workers)",
```
ddbeck, line 254: The linter won't accept any description but "Available in workers" here (though I appreciate what you're getting at).

```suggestion
          "description": "Available in workers",
```

api/FileReaderSync.json
```diff
@@ -249,103 +249,59 @@
...
           "support": {
             "chrome": {
               "version_added": true,
-              "version_removed": "59"
+              "notes": "Not supported in service workers from Chrome 59"
```
ddbeck, line 258: nit: style

```suggestion
              "notes": "From Chrome 59, not supported in service workers."
```

api/FileReaderSync.json
```diff
@@ -249,103 +249,59 @@
...
             },
             "chrome_android": {
               "version_added": true,
-              "version_removed": "59"
+              "notes": "Not supported in service workers from Chrome 59"
```
ddbeck, line 262: nit: style

```suggestion
              "notes": "From Chrome 59, not supported in service workers."
```

api/FileReaderSync.json
```diff
@@ -249,103 +249,59 @@
...
             "edge": {
-              "version_added": null
+              "version_added": "12",
+              "notes": "Not supported in service workers"
             },
             "firefox": {
-              "version_added": true,
-              "version_removed": "61"
+              "version_added": "8",
+              "notes": "Not supported in service workers from FF61"
```
ddbeck, line 270: nit: style

```suggestion
              "notes": "From Firefox 61, not supported in service workers."
```

api/FileReaderSync.json
```diff
@@ -249,103 +249,59 @@
...
             },
             "chrome_android": {
               "version_added": true,
-              "version_removed": "59"
+              "notes": "Not supported in service workers from Chrome 59"
             },
             "edge": {
-              "version_added": null
+              "version_added": "12",
+              "notes": "Not supported in service workers"
```
ddbeck, line 266: nit

```suggestion
              "notes": "Not supported in service workers."
```

api/FileReaderSync.json
```diff
@@ -249,103 +249,59 @@
...
-              "version_added": true,
-              "version_removed": "61"
+              "version_added": "8",
+              "notes": "Not supported in service workers from FF61"
             },
             "firefox_android": {
-              "version_added": true,
-              "version_removed": "61"
+              "version_added": "8",
+              "notes": "Not supported in service workers from FF61"
```
ddbeck, line 274: nit: style

```suggestion
              "notes": "Form Firefox 61, not supported in service workers."
```

api/FileReaderSync.json
```diff
@@ -249,103 +249,59 @@
...
             },
             "ie": {
-              "version_added": false
+              "version_added": true,
+              "notes": "Not supported in service workers"
```
ddbeck, line 278: Maybe we should omit this note? If I remember correctly, IE doesn't support service workers in any case (e.g., doesn't implement `navigator.serviceWorker`)

```suggestion
              "version_added": true
```

api/FileReaderSync.json
```diff
@@ -249,103 +249,59 @@
...
             },
             "opera": {
               "version_added": true,
-              "version_removed": "46"
+              "notes": "Not supported in service workers from Opera 46"
```
ddbeck, line 282: nit: style

```suggestion
              "notes": "From Opera 46, not supported in service workers."
```

api/FileReaderSync.json
```diff
@@ -249,103 +249,59 @@
...
             },
             "opera_android": {
               "version_added": true,
-              "version_removed": "43"
+              "notes": "Not supported in service workers from Opera 43"
```
ddbeck, line 286: nit: style

```suggestion
              "notes": "From Opera 43, not supported in service workers."
```

api/FileReaderSync.json
```diff
@@ -249,103 +249,59 @@
...
-              "version_added": true
-            },
-            "opera_android": {
-              "version_added": true
-            },
-            "safari": {
-              "version_added": true,
-              "version_removed": "6.1"
+              "version_removed": "6.1",
+              "notes": "Service workers not supported"
```
ddbeck, line 291: suggestion: for consistency with other notes in this PR

```suggestion
              "notes": "Not supported in service workers."
```

api/FileReaderSync.json
```diff
@@ -249,103 +249,59 @@
...
             },
             "safari_ios": {
               "version_added": true,
-              "version_removed": "7"
+              "version_removed": "7",
+              "notes": "Service workers not supported"
```
ddbeck, line 296: suggestion: for consistency with other notes in this PR

```suggestion
              "notes": "Not supported in service workers."
```

api/FileReaderSync.json
```diff
@@ -249,103 +249,59 @@
...
             },
             "samsunginternet_android": {
-              "version_added": true
+              "version_added": true,
+              "notes": "Not supported in service workers from version 7.0"
```
ddbeck, line 300:
```suggestion
              "notes": "From version 7.0, not supported in service workers."
```

api/FileReaderSync.json
```diff
@@ -249,103 +249,59 @@
...
             },
             "webview_android": {
-              "version_added": true
+              "version_added": true,
+              "notes": "Not supported in service workers from version 59"
```
ddbeck, line 304: nit: style

```suggestion
              "notes": "From version 59, not supported in service workers."
```

ddbeck (changes requested):
Thanks for the update, @hamishwillee.

First, to answer your hypothetical:

> How would YOU present for chrome this imaginary case.
> 
> * dedicated workers are supported from version 4,
> * shared workers are supported from version 8
> * service workers are supported from version 2 to 6.
> * chrome workers are not supported

Right now, given our schema, probably with something like this:

```json
{
  "version_added": "2",
  "notes": [
    "From version 8, shared workers are supported.",
    "From version 4, dedicated workers are supported.",
    "From version 2 to 6, service workers are supported; from version 7, service workers are unsupported."
  ]
}
```

Hypothetical aside, the new diff here makes a lot more sense to me. For example, in Chrome supported initially and generally, then dropped from service workers specifically. Suggests in line comments are just some style and consistency nits.

Also, just so I'm doing my due diligence here: I take it you did some sort of testing to figure out the support data here?

hamishwillee:
> Hypothetical aside, the new diff here makes a lot more sense to me. For example, in Chrome supported initially and generally, then dropped from service workers specifically. Suggests in line comments are just some style and consistency nits.

Agree - what you have suggested makes sense for the current data.

NOTE, that support for workers implies support for dedicated workers, shared workers and service workers. It does not imply support for the {FF-only) chrome workers - which I propose we completely ignore.

> Also, just so I'm doing my due diligence here: I take it you did some sort of testing to figure out the support data here?

No. There is no new data in this change - it is just a reformatting of the data that was already present.

ddbeck (approved):
Looks good. Thank you! 🎉

PR #8019, opened 2020-12-18
Move <menuitem> behind a flag in Firefox 85

rachelandrew:
Working on https://github.com/mdn/content/issues/294

Moving `<menuitem>` behind a flag. I've set it as removed in 85 then a flag added in 85, let me know if there is a better way to represent this.

Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1680596

### DIFF
--- a/html/elements/menuitem.json
+++ b/html/elements/menuitem.json
@@ -23,6 +23,10 @@
                     "name": "dom.menuitem.enabled",
                     "value_to_set": "true"
                   }
+                ],
+                "notes": [
+                  "Only works for <code>&lt;menuitem&gt;</code> elements defined within a <code>&lt;menu&gt;</code> element assigned to an element via the <code>contextmenu</code> attribute.",
+                  "The <code>&lt;menuitem&gt;</code> element requires a closing tag."
                 ]
               },
               {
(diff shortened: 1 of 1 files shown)


html/elements/menuitem.json
```diff
@@ -14,20 +14,46 @@
...
+                  }
+                ]
+              },
+              {
+                "version_removed": "85",
+                "version_added": "8",
+                "notes": [
+                  "Only works for <code>&lt;menuitem&gt;</code> elements defined within a <code>&lt;menu&gt;</code> element assigned to an element via the <code>contextmenu</code> attribute.",
+                  "The <code>&lt;menuitem&gt;</code> element requires a closing tag."
+                ]
```
ddbeck, line 34: I'd copy this to the support statement above

html/elements/menuitem.json
```diff
@@ -14,20 +14,46 @@
...
```
ddbeck, line 54: Same here

ddbeck (changes requested):
This looks really good, @rachelandrew. The only thing I'd suggest is preserving the notes from the unflagged support in the flagged support. From the bug, it doesn't look like the implementation changed at all—which makes sense a prelude to removal. Line comments to hopefully make that clearer.

ddbeck (approved):
Looks good. Thank you! 🎉

PR #8034, opened 2020-12-19
Update data for NavigatorConcurrentHardware mixin API

vinyldarkscratch:
This PR uses results from the mdn-bcd-collector project (v1.1.7) to set real values for the NavigatorConcurrentHardware mixin for all browsers. Results were manually confirmed for accuracy.  The collector confirmed the existing results, as well as indicated that IE has no support (expected, since it was supported since Edge 15), and Safari _briefly_ supported it in 10.1.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Navigator/hardwareConcurrency

### DIFF
--- a/api/NavigatorConcurrentHardware.json
+++ b/api/NavigatorConcurrentHardware.json
@@ -11,7 +11,7 @@
             "version_added": "37"
           },
           "edge": {
-            "version_added": "≤18"
+            "version_added": "15"
           },
           "firefox": {
             "version_added": "48"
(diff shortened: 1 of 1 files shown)


api/NavigatorConcurrentHardware.json
```diff
@@ -29,10 +29,12 @@
             "version_added": "24"
           },
           "safari": {
-            "version_added": false
+            "version_added": "10.1",
```
foolip, line 32: Interesting! :)

foolip (approved):

Issue #7517, opened 2020-11-27
Consistent use of partial_implementation and notes for prototype moves


foolip:
Over time, many members of `HTMLElement` or `SVGElement` have been moved up to `Element`, and all members of `HTMLDocument` have been moved up `Document`. The way this is represented in BCD varies a lot. A sample:

https://github.com/mdn/browser-compat-data/blob/5e7ff0a7194088756dedebb561d776abaf7ed03d/api/Document.json#L255-L256

https://github.com/mdn/browser-compat-data/blob/5e7ff0a7194088756dedebb561d776abaf7ed03d/api/Document.json#L1614

https://github.com/mdn/browser-compat-data/blob/5e7ff0a7194088756dedebb561d776abaf7ed03d/api/Document.json#L916

https://github.com/mdn/browser-compat-data/blob/5e7ff0a7194088756dedebb561d776abaf7ed03d/api/Element.json#L1684-L1685

https://github.com/mdn/browser-compat-data/blob/5e7ff0a7194088756dedebb561d776abaf7ed03d/api/Element.json#L1434-L1435

Since this problem is receding into the past and isn't something developers need to keep worrying about, I believe that we should consistently use `partial_implementation` and a note for the versions of each browser where the member was on a child interface and thus not supported on all instances of the interface, and separately have a support statement without notes from when the problem was fixed. Like this:

https://github.com/mdn/browser-compat-data/blob/5e7ff0a7194088756dedebb561d776abaf7ed03d/api/Document.json#L3108-L3118

foolip:
I should also say this only applies for moves *up* the prototype chain. https://github.com/mdn/browser-compat-data/issues/3463 is about the problem of moves more generally, and not all are the same.

jpmedley:
LGTM

ddbeck:
Having a rule we can apply to at least some of these moves is a step in the right direction. I like this and I'd be happy to see a PR with a data guideline codifying it. I have two questions, though:

Does this proposal require that the member appear only on the specified prototype and not the prototype it moved from? Or is this proposal only for the data on the destination prototype and agnostic about is previous place on the prototype chain?

Is the note text here part of the proposal? Do you think the note text above could act as a template for all such moves? (This would be nice to have, as it might help us locate such moves later, should we come up with a more comprehensive solution.)

jpmedley:
>Does this proposal require that the member appear only on the specified prototype and not the prototype it moved from?

[...]

PR #8040, opened 2020-12-19
Correct note text for cancelBubble prototype move

ddbeck:
This is a follow up to #7360. It rewrites the notes to use the model text in #7517.


foolip (approved):

foolip:
Hmm, should we not also have `version_removed` and `partial_implementation`?

foolip (approved):

PR #8041, opened 2020-12-19
Add missing features for CSSStyleSheet API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.7), for the `CSSStyleSheet` API.

Spec: https://wicg.github.io/construct-stylesheets/

IDL: https://github.com/w3c/webref/blob/master/ed/idl/construct-stylesheets.idl

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/CSSStyleSheet

### DIFF
--- a/api/CSSStyleSheet.json
+++ b/api/CSSStyleSheet.json
@@ -47,6 +47,53 @@
           "deprecated": false
         }
       },
+      "CSSStyleSheet": {
+        "__compat": {
+          "support": {
+            "chrome": {
+              "version_added": "66"
+            },
+            "chrome_android": {
+              "version_added": "66"
+            },
+            "edge": {
+              "version_added": "79"
+            },
+            "firefox": {
+              "version_added": false
+            },
+            "firefox_android": {
+              "version_added": false
+            },
+            "ie": {
+              "version_added": null
+            },
+            "opera": {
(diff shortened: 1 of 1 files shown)


api/CSSStyleSheet.json
```diff
@@ -47,6 +47,54 @@
...
+              "version_added": "79"
+            },
+            "firefox": {
+              "version_added": false
+            },
+            "firefox_android": {
+              "version_added": false
+            },
+            "ie": {
+              "version_added": null
```
foolip, line 70: Did the test not work in IE? I'm pretty sure this should be false.

foolip: I've tested https://mdn-bcd-collector.appspot.com/tests/api/CSSStyleSheet in IE11 and given the way it fails ("Object doesn't support this action") I'll indeed change this to false.

foolip:
```suggestion
              "version_added": false
```

foolip (approved):

PR #8042, opened 2020-12-19
Add missing features for CharacterData API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.7), for the `CharacterData` API.

Spec: https://dom.spec.whatwg.org/

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/CharacterData

### DIFF
--- a/api/CharacterData.json
+++ b/api/CharacterData.json
@@ -97,6 +97,53 @@
           }
         }
       },
+      "after": {
+        "__compat": {
+          "support": {
+            "chrome": {
+              "version_added": "54"
+            },
+            "chrome_android": {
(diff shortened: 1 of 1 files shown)


foolip:
Some of these are defined on mixing. Can you check they're not already elsewhere in BCD?

vinyldarkscratch:
Argh...mixins...

PR #8044, opened 2020-12-19

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.7), for the `DOMMatrix` API.

Spec: https://drafts.fxtf.org/geometry/

### DIFF
--- a/api/DOMMatrix.json
+++ b/api/DOMMatrix.json
@@ -96,6 +96,288 @@
           }
         }
       },
+      "invertSelf": {
+        "__compat": {
+          "support": {
+            "chrome": {
+              "version_added": "61"
+            },
+            "chrome_android": {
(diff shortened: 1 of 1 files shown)


api/DOMMatrix.json
```diff
@@ -96,6 +96,288 @@
           }
         }
       },
+      "invertSelf": {
```
foolip, line 99: Confirmed this and the other new entries here aren't already in some other part of BCD, none of the strings appear anywhere.

foolip (approved):

PR #8049, opened 2020-12-19
Correct data for DocumentOrShadowRoot mixin API

vinyldarkscratch:
While working on adding missing features of the Document API, the script was attempting to add data for features in this mixin (the mdn-bcd-collector flattens mixins).  However, before undoing those changes, I noticed that the new data mismatched what was in here.  This PR fixes the data accordingly.  All data was confirmed via manual testing.

### DIFF
--- a/api/DocumentOrShadowRoot.json
+++ b/api/DocumentOrShadowRoot.json
@@ -14,7 +14,7 @@
             "version_added": "12"
           },
           "firefox": {
-            "version_added": "3"
+            "version_added": "1"
           },
           "firefox_android": {
             "version_added": "4"
--- a/api/DocumentOrShadowRoot.json
+++ b/api/DocumentOrShadowRoot.json
@@ -5,40 +5,40 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/DocumentOrShadowRoot",
         "support": {
           "chrome": {
-            "version_added": "53"
+            "version_added": "1"
           },
           "chrome_android": {
-            "version_added": "53"
+            "version_added": "18"
           },
           "edge": {
             "version_added": "12"
           },
           "firefox": {
-            "version_added": true
+            "version_added": "3"
           },
           "firefox_android": {
-            "version_added": true
+            "version_added": "4"
           },
           "ie": {
-            "version_added": true
+            "version_added": "≤6"
           },
           "opera": {
-            "version_added": "40"
+            "version_added": "≤12.1"
           },
           "opera_android": {
-            "version_added": "41"
+            "version_added": "≤12.1"
           },
           "safari": {
-            "version_added": true
+            "version_added": "4"
           },
           "safari_ios": {
-            "version_added": true
+            "version_added": "3.2"
           },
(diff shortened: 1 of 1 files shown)


api/DocumentOrShadowRoot.json
```diff
@@ -5,40 +5,40 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/DocumentOrShadowRoot",
         "support": {
           "chrome": {
-            "version_added": "53"
+            "version_added": "1"
```
foolip, line 8: FWIW I think support statements for the mixin itself make no sense whatsoever. While the members of a mixin could be interpreted as "support on all the interfaces that include this mixin", support for the mixin itself doesn't really mean anything. But oh well, let's get rid of mixins another day.

api/DocumentOrShadowRoot.json
```diff
@@ -52,28 +52,28 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/DocumentOrShadowRoot/activeElement",
           "support": {
             "chrome": {
-              "version_added": "53"
+              "version_added": "1"
```
foolip, line 55: So here's the problem with mixins. `ShadowRoot` didn't exist in Chrome 1. But `document.activeElement` was there. So what do we do?

foolip (commented):
Mixins, mixins...

api/DocumentOrShadowRoot.json
vinyldarkscratch: Based upon our data, `ShadowRoot` is also not around in Chrome 53 either (though results from the `update-bcd` script indicate that it is).  It appears that was my fault (#3403); I'll open up a PR to fix the Chromium version data.

I think that overall, this is probably fine to set versions for these features earlier than when `ShadowRoot` was around.  I remember that when we were talking about how to mixins before, we had talked about basing support on when the method/attribute was supported on all of the interfaces that include it (disregarding versions when an interface is not available) -- this is basically how that data would look.

foolip: In https://github.com/mdn/browser-compat-data/issues/472#issuecomment-696628301 I list two options, and this would be the first ("any") option. @saschanaz also preferred that.

@ddbeck @Elchi3 do you think we should go ahead an update existing mixin data like this, or should we leave it alone and delete mixins entirely?

ddbeck: 1. If you're going to update mixin data, then the "any" option seems to be the only sustainable approach.
2. We should probably come up with a plan to de-mixin the data. But there's… a lot to unpack there, so I imagine it will take some time.
3. I suspect mixins will endure long enough that it won't be a waste of time to update the mixin data and I would welcome updates to it. But I would understand if, all other things being equal, you prioritized updating data for concrete interfaces ahead of mixins.

foolip: Thanks @ddbeck! Let's go with the "any" option for now then.

foolip (approved):

PR #8052, opened 2020-12-19
Remove the api.CharacterData.ChildNode entry

foolip:
ChildNode is a mixin, not an interface as described, and mixins are
completely invisible to web developers; they're a spec author
convenience.

This is a bit of cleanup that makes sense regardless of the outcome of
https://github.com/mdn/browser-compat-data/issues/472 more broadly.

### DIFF
--- a/api/CharacterData.json
+++ b/api/CharacterData.json
@@ -47,56 +47,6 @@
           "deprecated": false
         }
       },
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Thank you! 🎉

Issue #7936, opened 2020-12-16
Support data for constants?


saschanaz:
I had to manually remove some undersupported `CSSRule` constants while working on https://github.com/saschanaz/types-web/commit/88295d4f3fc491098b8f6cf1c9bc98f1514ede11, and found there is no support data for the constants except for `KeyboardEvent`. Would it be helpful to include them?

vinyldarkscratch:
There's been some discussion on this, for example https://github.com/mdn/browser-compat-data/pull/7580#issuecomment-739839319.  I can't find the exact PR/issue where it was mentioned, but overall we figured that constant data is not useful to web developers.

ddbeck:
Based on the discussion on #7585, I think we have consensus that we shouldn't include constants (unless we discover there are real compatibility issues raised by constants). I'd welcome a data guideline to document this policy (or I'll get to it… eventually).

sideshowbarker:
> Based on the discussion on #7585, I think we have consensus that we shouldn't include constants (unless we discover there are real compatibility issues raised by constants). I'd welcome a data guideline to document this policy (or I'll get to it… eventually).

I went ahead and raised #8062 with proposed language.

ddbeck:
We now have a guideline rejecting new constants in BCD. See https://github.com/mdn/browser-compat-data/pull/8062.

Repeating myself for anyone watching here: we'll need a separate issue for finding and pruning constants that have already slipped in, if that's something we want to do, but I think we've answered the question that kicked off the issue in the first place. If there's no objections, I'll close this one.

PR #8062, opened 2020-12-20
Add data guideline regarding constants

sideshowbarker:
See https://github.com/mdn/browser-compat-data/issues/7936#issuecomment-747390760 and https://github.com/mdn/browser-compat-data/issues/7585#issuecomment-739907251

### DIFF
--- a/docs/data-guidelines.md
+++ b/docs/data-guidelines.md
@@ -181,6 +181,8 @@ This guideline was proposed in [#6906](https://github.com/mdn/browser-compat-dat
 
 Don't include data for constants in BCD. The rationale for not including them is that they're not known to be a source of any compatibility issues.
 
+For example, although the UI Events specification defines a [`DOM_KEY_LOCATION_STANDARD`](https://w3c.github.io/uievents/#dom-keyboardevent-dom_key_location_standard) constant, we don't include data for it in BCD.
+
 This guideline was proposed in [#7936](https://github.com/mdn/browser-compat-data/issues/7585), based in part on previous discussion in [#7585](https://github.com/mdn/browser-compat-data/issues/7585).
 
 ## Release lines and backported features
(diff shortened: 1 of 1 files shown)


foolip:
I spotted some constants in https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent, are there others that will need cleaning up? If any of them have notes we should read them before deciding the entries have to go :)

docs/data-guidelines.md
```diff
@@ -177,6 +177,12 @@ However, this guideline does not apply to features where the browser's expected
 
 This guideline was proposed in [#6906](https://github.com/mdn/browser-compat-data/issues/6906).
 
+## Constants
+
+We don't include data for constants in BCD. The rationale for not including them is that they're not known to be a source of any compatibility issues.
```
ddbeck, line 182:
```suggestion
Don't include data for constants in BCD. The rationale for not including them is that they're not known to be a source of any compatibility issues.
```

docs/data-guidelines.md
```diff
@@ -177,6 +177,12 @@ However, this guideline does not apply to features where the browser's expected
...
+
```
ddbeck, line 183: It'd be nice to have a concrete example of a constant that we would omit from BCD.

ddbeck (changes requested):
Thank you for starting this, @sideshowbarker!

docs/data-guidelines.md
```diff
@@ -177,6 +177,12 @@ However, this guideline does not apply to features where the browser's expected
 
 This guideline was proposed in [#6906](https://github.com/mdn/browser-compat-data/issues/6906).
 
+## Constants
```
ddbeck, line 180: Oh, I missed one thing: can you update the ToC at the top of this file? Line 5 has instructions for doing this easily-ish

sideshowbarker: Will do

docs/data-guidelines.md
sideshowbarker: Agreed. Will add

sideshowbarker: > It'd be nice to have a concrete example of a constant that we would omit from BCD.

OK, added

docs/data-guidelines.md
sideshowbarker: > Oh, I missed one thing: can you update the ToC at the top of this file? Line 5 has instructions for doing this easily-ish

OK, ToC updated

ddbeck (approved):
OK, I'm happy with this and I'm going to merge it. Note that this only rejects adding new constants data to BCD. We'll need a separate issue for finding and pruning constants that have already slipped in, if that's something we want to do.

Thank you, @sideshowbarker!

PR #8068, opened 2020-12-20
Add spec_url data in html/manifest/*.json sources

sideshowbarker:
This change adds spec URLs in the `html/manifest/*.json` sources for all features with an `mdn_url` for an article with a Specification(s) table. Relates to https://github.com/mdn/browser-compat-data/issues/6765


Elchi3 (approved):
Very consistent, very nice! :+1:

PR #8071, opened 2020-12-20
Add missing ContactAddress API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing `ContactAddress` API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.7).

Spec: https://wicg.github.io/contact-api/spec/
IDL: https://github.com/w3c/webref/blob/master/ed/idl/contact-api.idl
Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/ContactAddress

### DIFF
--- a/api/ContactAddress.json
+++ b/api/ContactAddress.json
@@ -1,0 +1,53 @@
+{
+  "api": {
+    "ContactAddress": {
+      "__compat": {
+        "support": {
+          "chrome": {
+            "version_added": false
+          },
+          "chrome_android": {
+            "version_added": true
+          },
+          "edge": {
+            "version_added": false
+          },
+          "firefox": {
+            "version_added": "13",
+            "version_removed": "23"
+          },
+          "firefox_android": {
+            "version_added": "13",
+            "version_removed": "23"
+          },
+          "ie": {
+            "version_added": false
+          },
+          "opera": {
+            "version_added": false
+          },
+          "opera_android": {
+            "version_added": true
+          },
+          "safari": {
+            "version_added": false
+          },
+          "safari_ios": {
+            "version_added": false
+          },
+          "samsunginternet_android": {
+            "version_added": false
+          },
+          "webview_android": {
+            "version_added": true
+          }
+        },
+        "status": {
+          "experimental": false,
+          "standard_track": false,
+          "deprecated": false
+        }
+      }
+    }
+  }
+}
--- a/api/ContactAddress.json
+++ b/api/ContactAddress.json
@@ -17,7 +17,7 @@
             "version_removed": "23"
           },
           "firefox_android": {
-            "version_added": "13",
+            "version_added": "14",
             "version_removed": "23"
           },
           "ie": {
--- a/api/ContactAddress.json
+++ b/api/ContactAddress.json
@@ -13,12 +13,10 @@
             "version_added": false
           },
           "firefox": {
-            "version_added": "13",
-            "version_removed": "23"
+            "version_added": false
           },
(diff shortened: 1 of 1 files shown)


foolip (commented):
This is in https://wicg.github.io/contact-api/spec/#contactaddress, but why was it removed from Firefox? It looks like we don't have an entry for `navigator.contacts` which would be the entry point for this, so I'm skeptical about adding this.

api/ContactAddress.json
```diff
@@ -0,0 +1,53 @@
+{
+  "api": {
+    "ContactAddress": {
+      "__compat": {
+        "support": {
+          "chrome": {
+            "version_added": false
+          },
+          "chrome_android": {
+            "version_added": true
```
foolip, line 10:
```suggestion
            "version_added": "84"
```

Based on https://storage.googleapis.com/chromium-find-releases-static/1a3.html#1a3c88d905e7b6079dd1d98ae39f1fad7a39f219

api/ContactAddress.json
```diff
@@ -0,0 +1,53 @@
...
+          },
+          "safari": {
+            "version_added": false
+          },
+          "safari_ios": {
+            "version_added": false
+          },
+          "samsunginternet_android": {
+            "version_added": false
```
foolip, line 39: Indeed, doesn't seem to be enabled in the latest Samsung Internet.

api/ContactAddress.json
```diff
@@ -0,0 +1,53 @@
...
+            "version_removed": "23"
+          },
+          "ie": {
+            "version_added": false
+          },
+          "opera": {
+            "version_added": false
+          },
+          "opera_android": {
+            "version_added": true
```
foolip, line 30:
```suggestion
            "version_added": "60"
```

Mapped from Chrome, but also tested in Opera 61 to confirm it's not disabled.

api/ContactAddress.json
```diff
@@ -0,0 +1,53 @@
...
+            "version_added": false
+          },
+          "safari_ios": {
+            "version_added": false
+          },
+          "samsunginternet_android": {
+            "version_added": false
+          },
+          "webview_android": {
+            "version_added": true
```
foolip, line 42:
```suggestion
            "version_added": "84"
```

api/ContactAddress.json
```diff
@@ -0,0 +1,53 @@
...
+          }
+        },
+        "status": {
+          "experimental": false,
```
foolip, line 46:
```suggestion
          "experimental": true,
```

api/ContactAddress.json
```diff
@@ -0,0 +1,53 @@
...
+            "version_added": false
+          },
+          "chrome_android": {
+            "version_added": "84"
+          },
+          "edge": {
+            "version_added": false
+          },
+          "firefox": {
+            "version_added": "13",
```
foolip, line 16: I've confirmed there was something there in Firefox back then, but it was part of a totally different API, removed in https://github.com/mozilla/gecko-dev/commit/80af1b3196490d7e2b1e6fae66329a6462623e75, although then `ContactAddress` was a dictionary so I don't think that's the right commit for a removal version.

Since this is something totally different and a very long time ago, I'm just going to set the data for Firefox to false. If our tooling keeps suggesting adding it, we can think about a workaround to suppress it.

foolip (approved):

PR #8072, opened 2020-12-20
Add missing DOMRectList API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing `DOMRectList` API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.7).

Spec: https://drafts.fxtf.org/geometry/

### DIFF
--- a/api/DOMRectList.json
+++ b/api/DOMRectList.json
@@ -1,0 +1,145 @@
+{
+  "api": {
+    "DOMRectList": {
+      "__compat": {
+        "support": {
+          "chrome": {
+            "version_added": "2"
+          },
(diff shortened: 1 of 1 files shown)


foolip (commented):
Can you add alt. name statements for when this was called ClientRectList in various browsers?

foolip:
I spotted that the test is actually broken, so I've filed https://github.com/foolip/mdn-bcd-collector/issues/901 and will close this PR, can't trust the results.

PR #8075, opened 2020-12-21
Add missing OffscreenCanvasRenderingContext2D API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing `OffscreenCanvasRenderingContext2D` API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.7).

Spec: https://html.spec.whatwg.org/multipage/canvas.html#offscreencanvasrenderingcontext2d

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/OffscreenCanvasRenderingContext2D

### DIFF
--- a/api/OffscreenCanvasRenderingContext2D.json
+++ b/api/OffscreenCanvasRenderingContext2D.json
@@ -1,0 +1,98 @@
+{
+  "api": {
+    "OffscreenCanvasRenderingContext2D": {
+      "__compat": {
+        "support": {
+          "chrome": {
+            "version_added": "69"
+          },
+          "chrome_android": {
+            "version_added": "69"
+          },
+          "edge": {
+            "version_added": "79"
+          },
+          "firefox": {
+            "version_added": false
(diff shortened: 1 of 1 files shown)


foolip (commented):
No support for http://mdn-bcd-collector.appspot.com/tests/api/OffscreenCanvasRenderingContext2D/commit it seems.

But http://mdn-bcd-collector.appspot.com/tests/api/OffscreenCanvasRenderingContext2D lists many more members which come from mixins. Should those not be added here as well?

vinyldarkscratch:
I've been removing results generated for features in mixins.  I can write a follow-up PR to add the features from mixins, though?

foolip:
Sure, doing those separately would be fine.

foolip (approved):

PR #8087, opened 2020-12-21
Update Safari versions for GlobalEventHandlers API

vinyldarkscratch:
This PR updates and corrects the real values for Safari (Desktop and iOS/iPadOS) for the `GlobalEventHandlers` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.7).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/GlobalEventHandlers.json
+++ b/api/GlobalEventHandlers.json
@@ -204,10 +204,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": "11"
+              "version_added": "9"
             },
             "safari_ios": {
-              "version_added": "11"
+              "version_added": "9"
             },
             "samsunginternet_android": [
               {
@@ -282,10 +282,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": "11"
+              "version_added": "9"
             },
             "safari_ios": {
-              "version_added": "11"
+              "version_added": "9"
             },
             "samsunginternet_android": [
               {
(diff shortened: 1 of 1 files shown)


api/GlobalEventHandlers.json
```diff
@@ -204,10 +204,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": "9"
```
foolip, line 207: Was `onanimationend` perhaps supported on `Element` or `HTMLElement` in Safari 9? Same for changes below.

api/GlobalEventHandlers.json
```diff
@@ -3891,10 +3891,10 @@
               "version_added": "14"
             },
             "safari": {
-              "version_added": "1.3"
```
foolip, line 3894: Was `onselectionchange` perhaps supported earlier on some other interface?

foolip (changes requested):

api/GlobalEventHandlers.json
vinyldarkscratch: Interestingly enough, it wasn't supported on `Document`, but it was supported on all other interfaces.  Reverting these changes.

api/GlobalEventHandlers.json
vinyldarkscratch: Nope, no other interfaces supported it when testing in Safari 4, so I'm not sure where the "1.3" came from.

foolip (approved):

PR #8090, opened 2020-12-21
Update IE/Edge versions for GlobalEventHandlers API

vinyldarkscratch:
This PR updates and corrects the real values for Internet Explorer and Edge for the `GlobalEventHandlers` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.7).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/GlobalEventHandlers.json
+++ b/api/GlobalEventHandlers.json
@@ -953,7 +953,7 @@
               "version_added": false
             },
             "ie": {
-              "version_added": "10"
+              "version_added": "9"
             },
             "opera": {
               "version_added": "12"
@@ -1001,7 +1001,7 @@
               "version_added": false
             },
             "ie": {
-              "version_added": "10"
+              "version_added": "9"
             },
             "opera": {
               "version_added": "12"
@@ -1049,7 +1049,7 @@
               "version_added": false
             },
             "ie": {
-              "version_added": "10"
+              "version_added": "9"
             },
             "opera": {
               "version_added": "12"
@@ -1146,7 +1146,7 @@
               "version_added": false
             },
             "ie": {
-              "version_added": "10"
+              "version_added": "9"
             },
             "opera": {
               "version_added": "12"
@@ -1194,7 +1194,7 @@
               "version_added": false
             },
             "ie": {
-              "version_added": "10"
+              "version_added": "9"
             },
             "opera": {
               "version_added": "12"
@@ -1242,7 +1242,7 @@
               "version_added": false
             },
             "ie": {
-              "version_added": "10"
+              "version_added": "9"
             },
             "opera": {
               "version_added": "12"
@@ -1290,7 +1290,7 @@
               "version_added": false
             },
             "ie": {
-              "version_added": "10"
+              "version_added": "9"
             },
             "opera": {
               "version_added": "12"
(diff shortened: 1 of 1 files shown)


api/GlobalEventHandlers.json
```diff
@@ -3609,7 +3609,7 @@
               "version_added": "45"
             },
             "edge": {
-              "version_added": "12"
+              "version_added": "79"
```
foolip, line 3612: Can you double check why this was originally 12? Maybe the event itself is supported?

api/GlobalEventHandlers.json
```diff
@@ -4281,7 +4281,7 @@
               "version_added": true
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": false
```
foolip, line 4284: Can you check why this was set to supported before?

vinyldarkscratch: When Blink-versions of Edge had come out, all of the existing Edge data set to `true` and `null` was converted into ranges of `≤18` and `≤79` respectively, as per mirroring from Chrome Desktop.  Alas, the tests here have confirmed no support in Edge, so this is more like setting `null` to `false`.

api/GlobalEventHandlers.json
vinyldarkscratch: The `12` came from mirroring from Internet Explorer, it looks like.  However, since support was disconfirmed in IE, this data is also inaccurate.

api/GlobalEventHandlers.json
foolip: I see, thanks!

api/GlobalEventHandlers.json
```diff
@@ -3618,7 +3618,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": true
+              "version_added": false
```
foolip, line 3621: http://mdn-bcd-collector.appspot.com/tests/api/Window/onresize tests true in IE11, so there's a problem with the methodology.

api/GlobalEventHandlers.json
```diff
@@ -4821,7 +4821,7 @@
...
```
foolip, line 4824: I tested http://mdn-bcd-collector.appspot.com/tests/api/Window/onwheel and the other http://mdn-bcd-collector.appspot.com/tests/api/*/onwheel tests in IE11 and all return false, so it seems like the original data was wrong. The event itself might be supported though.

api/GlobalEventHandlers.json
vinyldarkscratch: Hmm, it's available on the `Window` API, but not on any of the other APIs.  Should we just add a note, or `partial_implementation` with a version number?

api/GlobalEventHandlers.json
foolip: I see, thanks!

api/GlobalEventHandlers.json
foolip: I think we should split mixins into their interfaces which would make partial impl. statements in this situation unnecessary. For now, how about just setting it to the version from when it was supported on any interface?

It sort of doesn't matter what we do since all mixin data will have to be revisited later anyway.

vinyldarkscratch: Sounds good -- I added partial implementation statements mentioning how it's only available on `Window` in IE and EdgeHTML, so that we have the data in here for whatever we'll end up doing with mixins!

foolip (approved):

PR #8091, opened 2020-12-21
Update Firefox versions for GlobalEventHandlers API

vinyldarkscratch:
This PR updates and corrects the real values for Firefox and Firefox Android for the `GlobalEventHandlers` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.7).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/GlobalEventHandlers.json
+++ b/api/GlobalEventHandlers.json
@@ -947,10 +947,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "3.5"
+              "version_added": "9"
             },
             "firefox_android": {
-              "version_added": false
+              "version_added": "9"
             },
             "ie": {
               "version_added": "10"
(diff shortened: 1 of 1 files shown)


api/GlobalEventHandlers.json
```diff
@@ -947,10 +947,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "3.5"
+              "version_added": "9"
```
foolip, line 950: There's a lot of changes from 3.5. Can you check where that data came from? Maybe it was supported on one of the other interfaces?

vinyldarkscratch: This data came from the initial wiki migration.  Doing some additional testing around, I couldn't find any interface that supported these event handlers, so it's probably that the event is supported in Firefox 3.5, but not the event handlers.

foolip: Yes, that's a common situation.

foolip (approved):

foolip:
And that was the last PR for the year, local time :)

vinyldarkscratch:
Happy New Years, @foolip!  I've still got 9 hours to go myself.  😛

foolip:
A belated happy new year to you too, @vinyldarkscratch!

PR #8106, opened 2020-12-21
Remove redundant Firefox notes from HTMLHyperlinkElementUtils

vinyldarkscratch:
This PR is a sister to #8105, which removes the redundant notes from Firefox.  Web developers would not see a difference due to the renaming of its mixin and the copying of its members to other interfaces.  As such, it's not really useful to mention the mixin rename.


foolip (approved):
Yay!

PR #8111, opened 2020-12-22
Add Chromium versions for HTMLHyperlinkElementUtils API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `HTMLHyperlinkElementUtils` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.7).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/HTMLAnchorElement

~~Notice: the diff may conflict with #8105 as they both perform modifications to the same lines.~~

### DIFF
--- a/api/HTMLHyperlinkElementUtils.json
+++ b/api/HTMLHyperlinkElementUtils.json
@@ -5,11 +5,11 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLHyperlinkElementUtils",
         "support": {
           "chrome": {
-            "version_added": true,
+            "version_added": "1",
             "notes": "Starting in Chrome 52, the members of this interface were moved to <a href='https://developer.mozilla.org/docs/Web/API/URL'>URL</a>"
           },
           "chrome_android": {
-            "version_added": true,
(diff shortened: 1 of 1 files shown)


foolip:
The test link there doesn't work. I think you copied the data from an interface that includes this mixin, can you link the tests that were used and the approach? Writing it up in the mixin issue and linking from PRs like this would be great.

vinyldarkscratch:
Ah yeah, I should probably update my PR creation script to account for mixing as well.  Long story short, the `add-new-bcd` script in the collector had added the flattened members on all the interfaces that implement the mixin, such as HTMLAnchorElement.  To obtain the version numbers, I copied the results that were added to `HTMLAnchorElement` API, and then checked other APIs that implement the mixin, matching up version numbers and using the latest version number (taking into consideration the versions before some interfaces aren't implemented).

foolip:
Thanks for the explanation! Testing HTMLAnchorElement makes sense.

foolip (approved):

PR #8112, opened 2020-12-22
Add HTMLElement.attachInternals API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.7), for the `HTMLElement` API.

Spec: https://html.spec.whatwg.org/multipage/dom.html#htmlelement

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/HTMLElement

### DIFF
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -369,6 +369,53 @@
           }
         }
       },
+      "attachInternals": {
+        "__compat": {
+          "support": {
+            "chrome": {
+              "version_added": "77"
+            },
+            "chrome_android": {
+              "version_added": "77"
+            },
+            "edge": {
+              "version_added": "79"
+            },
+            "firefox": {
+              "version_added": false
+            },
+            "firefox_android": {
+              "version_added": false
+            },
+            "ie": {
+              "version_added": false
+            },
+            "opera": {
+              "version_added": "64"
+            },
+            "opera_android": {
+              "version_added": "55"
+            },
+            "safari": {
+              "version_added": false
+            },
+            "safari_ios": {
+              "version_added": false
+            },
+            "samsunginternet_android": {
+              "version_added": "12.0"
+            },
+            "webview_android": {
+              "version_added": "77"
+            }
+          },
+          "status": {
+            "experimental": false,
+            "standard_track": true,
+            "deprecated": false
+          }
+        }
+      },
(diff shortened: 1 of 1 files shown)


api/HTMLElement.json
```diff
@@ -416,6 +463,53 @@
           }
         }
       },
+      "autofocus": {
+        "__compat": {
+          "support": {
+            "chrome": {
+              "version_added": "79"
```
foolip, line 470: Autofocus behavior is very old. Can you check if we have other "autofocus" entries anywhere in BCD and what their data is?

vinyldarkscratch: Yeah, I can dig into this more.  My best guess is that it may be due to the `autofocus` attribute being defined on the individual elements (or on the `Element` API), and then moved to `HTMLElement`?  The collector tests `HTMLElement` specifically.

foolip: Yeah, it might have been on HTMLInputElement originally, that's worth checking.

vinyldarkscratch: Okay, so it looks like they were initially defined on individual HTML elements (`button`, `input`, `select`, and `textarea`), and then in [Chrome 79](https://storage.googleapis.com/chromium-find-releases-static/0fb.html#0fbf9bdd80bfef802dcc57ab0c295a95a3fa39b3), they were moved to the `HTMLOrForeignElement` mixin.  Should we add a note as such?

foolip: Sounds like this is similar to https://github.com/mdn/browser-compat-data/issues/7519 and `partial_implementation` is appropriate. Problem is figuring out what the note should say, we need to figure out all the places that autofocus can work, and if any of those weren't supported from the beginning.

I half expect that the actual support hasn't changes and that it's only the IDL that's changed, so that there won't be a note that makes sense to web developers. But testing first makes sense.

foolip: In order to unblock the other half of this PR, I've filed https://github.com/mdn/browser-compat-data/issues/8745 and will revert this bit and land this PR.

foolip (approved):

PR #8114, opened 2020-12-22
Add missing features for HTMLInputElement API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.7), for the `HTMLInputElement` API.

Spec: https://html.spec.whatwg.org/multipage/input.html#htmlinputelement

### DIFF
--- a/api/HTMLInputElement.json
+++ b/api/HTMLInputElement.json
@@ -284,6 +284,53 @@
           }
         }
       },
+      "capture": {
+        "__compat": {
+          "support": {
+            "chrome": {
+              "version_added": false
+            },
+            "chrome_android": {
+              "version_added": true
+            },
+            "edge": {
+              "version_added": false
+            },
+            "firefox": {
+              "version_added": false
+            },
+            "firefox_android": {
+              "version_added": true
+            },
+            "ie": {
+              "version_added": false
+            },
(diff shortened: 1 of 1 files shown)


api/HTMLInputElement.json
```diff
@@ -284,6 +284,53 @@
...
         }
       },
+      "capture": {
+        "__compat": {
+          "support": {
+            "chrome": {
+              "version_added": false
+            },
+            "chrome_android": {
+              "version_added": true
```
foolip, line 294: Is it possible to figure out when it was added? I can dig through the source code if not.

vinyldarkscratch: I'd be happy to track down the version number, or if you'd like to track it down, go for it!  (I mainly wanted to focus on just getting the entries in first.)

foolip: I'd need to be at a computer, but okay let's merge this and fill in the version later.

foolip (approved):

PR #8129, opened 2020-12-22
Add missing features for NodeList API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.7), for the `NodeList` API.

Spec: https://dom.spec.whatwg.org/

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/NodeList

### DIFF
--- a/api/NodeList.json
+++ b/api/NodeList.json
@@ -334,6 +334,55 @@
             "deprecated": false
           }
         }
+      },
+      "@@iterator": {
+        "__compat": {
+          "support": {
+            "chrome": {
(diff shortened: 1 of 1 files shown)


foolip (commented):
This has been gone for over two years, so I don't think we should add it.

PR #8134, opened 2020-12-22
Add missing features for PerformanceLongTaskTiming API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.7), for the `PerformanceLongTaskTiming` API.

Spec: https://w3c.github.io/longtasks/

IDL: https://github.com/w3c/webref/blob/master/ed/idl/longtasks.idl

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/PerformanceLongTaskTiming

### DIFF
--- a/api/PerformanceLongTaskTiming.json
+++ b/api/PerformanceLongTaskTiming.json
@@ -97,6 +97,53 @@
             "deprecated": false
           }
         }
+      },
+      "toJSON": {
+        "__compat": {
+          "support": {
+            "chrome": {
+              "version_added": "58"
+            },
+            "chrome_android": {
+              "version_added": "58"
+            },
+            "edge": {
+              "version_added": "79"
+            },
+            "firefox": {
+              "version_added": false
+            },
+            "firefox_android": {
+              "version_added": false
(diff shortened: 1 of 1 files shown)


api/PerformanceLongTaskTiming.json
```diff
@@ -97,6 +97,53 @@
             "deprecated": false
           }
         }
+      },
+      "toJSON": {
```
foolip, line 101: Since the test tests the prototype, can you verify this isn't inherited and a false positive?

vinyldarkscratch: Just did some testing in Chrome in BrowserStack, it looks like this is accurate!

foolip: Great, thank you!

foolip (approved):

PR #8142, opened 2020-12-22
Add Chromium versions for DOMImplementation API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `DOMImplementation` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.7).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/DOMImplementation

### DIFF
--- a/api/DOMImplementation.json
+++ b/api/DOMImplementation.json
@@ -23,10 +23,10 @@
             "version_added": "6"
           },
           "opera": {
-            "version_added": "15"
+            "version_added": "≤12.1"
           },
           "opera_android": {
-            "version_added": "14"
+            "version_added": "≤12.1"
           },
           "safari": {
             "version_added": "≤4"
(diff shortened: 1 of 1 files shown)


api/DOMImplementation.json
```diff
@@ -23,10 +23,10 @@
             "version_added": "6"
           },
           "opera": {
-            "version_added": true
+            "version_added": "15"
```
foolip, line 26: Can you double check this? Most of document.implementation should have been in Presto too.

vinyldarkscratch: Didn't have a custom test for this, will introduce one in v1.1.8.  I'll adjust the data.

foolip: Thanks!

foolip (approved):

PR #8148, opened 2020-12-22
Mark expandEntityReferences support in createTreeWalker as deprecated

saschanaz:
It's dropped from:

* Blink in 2013 (but it already was being ignored before the removal, per the commit message) https://source.chromium.org/chromium/chromium/src/+/116ac6de3914734f7136874089a55a3922113f1d
* WebKit in 2015 https://webkit-search.igalia.com/webkit/commit/a29881a6fedfed8be60c3d76c5b4324041c7979c

- [x] Summarize your changes


saschanaz:
It seems the feature itself is removed (or at least effectively disabled) together with the interface `EntityReference`, so probably should be matched with it.

Or should this be just removed?

api/Document.json
```diff
@@ -2767,13 +2767,14 @@
             "description": "<code>expandEntityReferences</code> parameter",
             "support": {
               "chrome": {
-                "version_added": "4"
+                "version_added": false
```
saschanaz, line 2770: [Per the comment](https://source.chromium.org/chromium/chromium/src/+/116ac6de3914734f7136874089a55a3922113f1d:third_party/WebKit/Source/core/dom/Document.idl;dlc=954f614cc71dbb9a8f5990600de09f7cf0cf53ad):

```
// Historically, this argument was never implemented and has been ignored.
```

api/Document.json
```diff
@@ -2787,28 +2788,32 @@
...
               },
               "opera_android": {
-                "version_added": true
+                "version_added": true,
+                "version_removed": "14"
               },
               "safari": {
-                "version_added": "3"
+                "version_added": "3",
+                "version_removed": "10.1"
```
saschanaz, line 2800: This is copied from #7278.

ddbeck (approved):
> Or should this be just removed? There is no `EntityReference` in BCD anyway, it seemingly was removed from MDN many years ago.

Let's leave this in until next January. Edge's removal isn't old enough yet to merit removal under the irrelevance guideline.

Thank you!

PR #8172, opened 2020-12-23
Add missing features for ServiceWorkerGlobalScope API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.7), for the `ServiceWorkerGlobalScope` API.

Spec: https://wicg.github.io/cookie-store/, https://wicg.github.io/content-index/spec/, https://wicg.github.io/periodic-background-sync/, https://w3c.github.io/ServiceWorker/

IDL: https://github.com/w3c/webref/blob/master/ed/idl/service-workers.idl

### DIFF
--- a/api/ServiceWorkerGlobalScope.json
+++ b/api/ServiceWorkerGlobalScope.json
@@ -196,6 +196,53 @@
           }
         }
       },
+      "cookieStore": {
+        "__compat": {
+          "support": {
+            "chrome": {
+              "version_added": "87"
+            },
+            "chrome_android": {
+              "version_added": "87"
+            },
+            "edge": {
+              "version_added": "87"
+            },
+            "firefox": {
+              "version_added": false
+            },
+            "firefox_android": {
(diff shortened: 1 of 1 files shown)


api/ServiceWorkerGlobalScope.json
```diff
@@ -1383,6 +1571,53 @@
...
+              "version_added": false
+            },
+            "opera": {
+              "version_added": "66"
+            },
+            "opera_android": {
+              "version_added": "57"
+            },
+            "safari": {
+              "version_added": false
```
foolip, line 1602: No support for `self.serviceWorker` in Safari is surprising, but the test is as simple as it can be so I trust the results...

foolip (approved):

PR #8176, opened 2020-12-23
Add missing features for TextTrackCue API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.7), for the `TextTrackCue` API.

Spec: https://html.spec.whatwg.org/multipage/media.html#texttrackcue

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/TextTrackCue

Depends on data added in #8177 and #8179.

### DIFF
--- a/api/TextTrackCue.json
+++ b/api/TextTrackCue.json
@@ -241,6 +241,100 @@
           }
         }
       },
+      "onenter": {
+        "__compat": {
+          "support": {
+            "chrome": {
+              "version_added": "23"
+            },
+            "chrome_android": {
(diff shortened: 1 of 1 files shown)


foolip:
@vinyldarkscratch can you rebase now?

foolip (approved):

PR #8184, opened 2020-12-23
Add missing features for Window API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.7), for the `Window` API.

Spec: https://html.spec.whatwg.org/multipage/window-object.html#window, https://html.spec.whatwg.org/multipage/obsolete.html#Window-partial, https://wicg.github.io/cookie-store/, https://drafts.csswg.org/css-nav-1/, https://drafts.csswg.org/css-nav-1/, https://drafts.csswg.org/cssom-view/, https://w3c.github.io/webappsec-trusted-types/dist/spec/

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Window

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -4559,55 +4559,6 @@
           }
         }
       },
-      "navigate": {
-        "__compat": {
-          "support": {
-            "chrome": {
-              "version_added": false
-            },
-            "chrome_android": {
-              "version_added": false
-            },
-            "edge": {
-              "version_added": false
-            },
-            "firefox": {
-              "version_added": false
-            },
-            "firefox_android": {
-              "version_added": false
-            },
-            "ie": {
-              "version_added": "≤6"
-            },
-            "opera": {
-              "version_added": "≤12.1",
-              "version_removed": "15"
-            },
-            "opera_android": {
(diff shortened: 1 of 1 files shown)


api/Window.json
```diff
@@ -4418,6 +4559,55 @@
           }
         }
       },
+      "navigate": {
```
foolip, line 4562: I don't think we should add this entry, as it's not in any recent browser release.

foolip (changes requested):

foolip:
@vinyldarkscratch can you check if any of these entries also existing somewhere else in BCD, in particular on WindowOrWorkerGlobalScope?

vinyldarkscratch:
Before submitting this PR, I had double-checked to make sure that the only entries being added were directly implemented on the `Window` interface!

foolip:
Okay, in that case let me review properly...

foolip (approved):

PR #8201, opened 2020-12-23
Add Chromium versions for SVGAnimatedPoints API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `SVGAnimatedPoints` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.8).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/SVGAnimatedPoints.json
+++ b/api/SVGAnimatedPoints.json
@@ -5,10 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/SVGAnimatedPoints",
         "support": {
           "chrome": {
-            "version_added": null
+            "version_added": "1"
           },
           "chrome_android": {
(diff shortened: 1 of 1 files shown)


foolip:
Is the custom test for this really correct? The instance is the return value of createElementNS, but AFAIK all elements are called something ending with "Element".

vinyldarkscratch:
This interface is a mixin, but initially, it was its own interface (similar to the story explained in https://github.com/mdn/browser-compat-data/pull/8235#discussion_r548331997).

foolip:
Ah, I see, a custom test for a mixin by testing something that includes the mixin. That makes sense!

foolip (approved):

PR #8202, opened 2020-12-23
Update Firefox versions for SVGAnimatedPoints API

vinyldarkscratch:
This PR updates and corrects the real values for Firefox and Firefox Android for the `SVGAnimatedPoints` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.8).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/SVGAnimatedPoints.json
+++ b/api/SVGAnimatedPoints.json
@@ -14,12 +14,10 @@
             "version_added": "≤18"
           },
           "firefox": {
-            "version_added": "1.5",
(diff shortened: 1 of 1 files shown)


foolip:

foolip:
I'm going to close this. The previous state with addition and removal gives better clues about what happened here when trying to resolve this in https://github.com/mdn/browser-compat-data/issues/7752.

PR #8208, opened 2020-12-23
Add missing features for SVGAnimatedPoints API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.8), for the `SVGAnimatedPoints` API.

Depends on data added in #8201, #8202, #8203, and #8204.

### DIFF
--- a/api/SVGAnimatedPoints.json
+++ b/api/SVGAnimatedPoints.json
@@ -48,6 +48,100 @@
           "standard_track": true,
           "deprecated": false
         }
+      },
+      "animatedPoints": {
+        "__compat": {
+          "support": {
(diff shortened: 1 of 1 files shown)


foolip:

foolip:
In addition to https://github.com/mdn/browser-compat-data/issues/7752, this is now a mixin, and I don't think we should be adding subfeatures to mixins. Closing this in favor of the open issues.

PR #8235, opened 2020-12-23
Update Firefox versions for SVGFilterPrimitiveStandardAttributes API

vinyldarkscratch:
This PR updates and corrects the real values for Firefox and Firefox Android for the `SVGFilterPrimitiveStandardAttributes` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.8).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/SVGFilterPrimitiveStandardAttributes.json
+++ b/api/SVGFilterPrimitiveStandardAttributes.json
@@ -14,12 +14,10 @@
             "version_added": null
           },
           "firefox": {
-            "version_added": "3",
-            "version_removed": "22"
+            "version_added": "3"
           },
           "firefox_android": {
-            "version_added": "4",
-            "version_removed": "22"
+            "version_added": "4"
           },
           "ie": {
             "version_added": null


api/SVGFilterPrimitiveStandardAttributes.json
```diff
@@ -14,12 +14,10 @@
             "version_added": null
           },
           "firefox": {
-            "version_added": "3",
-            "version_removed": "22"
```
foolip, line 18: Wasn't it removed? What's the story with the existing data here?

vinyldarkscratch: Short answer: mixin...

Long answer: This interface had a bit of a weird history.  In [SVG 1.1](https://www.w3.org/TR/SVG11/filters.html#InterfaceSVGFilterPrimitiveStandardAttributes), it was defined as its own standalone interface.  But then, in [SVG 2](https://drafts.fxtf.org/filter-effects-1/#InterfaceSVGFilterPrimitiveStandardAttributes), it had become a mixin.  The collector has custom IDL for the SVG 1.1 interface, so the results it is generating are based upon the assumption that it is an interface on its own, not a mixin.

I opened up this PR by copying the versions for the new BCD added into interfaces that implement the mixin.  I've also now removed the old interface definition.

foolip:

foolip:
Closing this like https://github.com/mdn/browser-compat-data/pull/8202, because I think the previous state gives better clues about what happened here until we resolve the situation.

PR #8236, opened 2020-12-23
Add missing features for SVGFilterPrimitiveStandardAttributes API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.8), for the `SVGFilterPrimitiveStandardAttributes` API.

Spec: https://drafts.fxtf.org/filter-effects-1/

IDL: https://github.com/w3c/webref/blob/master/ed/idl/filter-effects.idl

Depends on data added in #8232, #8233, #8234, and #8235.

### DIFF
--- a/api/SVGFilterPrimitiveStandardAttributes.json
+++ b/api/SVGFilterPrimitiveStandardAttributes.json
@@ -48,6 +48,241 @@
           "standard_track": true,
           "deprecated": false
         }
+      },
+      "height": {
+        "__compat": {
+          "support": {
+            "chrome": {
+              "version_added": "5"
+            },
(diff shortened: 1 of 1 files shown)


foolip:

foolip:
Closing this like https://github.com/mdn/browser-compat-data/pull/8208, this is currently a mixin and I don't think we should add more subfeatures to mixins.

Issue #8150, opened 2020-12-22
api.SVGElement.className - Add and mark it as deprecated


saschanaz:
#### What information was incorrect, unhelpful, or incomplete?

The SVGElement page says it's deprecated but has no corresponding BCD data.

#### What did you expect to see?

A corresponding row that says it's deprecated

#### Did you test this? If so, how?

N/A

#### Can you link to any release notes, bugs, pull requests, or MDN pages related to this?

https://developer.mozilla.org/en-US/docs/Web/API/SVGElement

https://svgwg.org/svg2-draft/types.html#InterfaceSVGElement

>This attribute is deprecated and may be removed in a future version of this specification. Authors are advised to use Element.classList instead.

saschanaz:
#8245

PR #8245, opened 2020-12-23
Add missing features for SVGElement API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.8), for the `SVGElement` API.

### DIFF
--- a/api/SVGElement.json
+++ b/api/SVGElement.json
@@ -96,6 +96,53 @@
           }
         }
       },
+      "className": {
+        "__compat": {
+          "support": {
+            "chrome": {
+              "version_added": "22"
+            },
+            "chrome_android": {
+              "version_added": "25"
+            },
+            "edge": {
+              "version_added": "13"
+            },
+            "firefox": {
+              "version_added": "20"
+            },
+            "firefox_android": {
+              "version_added": "20"
+            },
+            "ie": {
+              "version_added": false
+            },
+            "opera": {
+              "version_added": "15"
+            },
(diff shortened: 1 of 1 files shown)


saschanaz:
Closes #8150 👍

api/SVGElement.json
```diff
@@ -96,6 +96,53 @@
...
+              "version_added": "1.5"
+            },
+            "webview_android": {
+              "version_added": "≤37"
+            }
+          },
+          "status": {
+            "experimental": false,
+            "standard_track": true,
+            "deprecated": false
```
saschanaz, line 142:
```suggestion
            "deprecated": true
```

api/SVGElement.json
```diff
@@ -96,6 +96,53 @@
           }
         }
       },
+      "className": {
```
foolip, line 99: Where else in BCD do we already have a "className" entry?

saschanaz: `api.Element.className` exists but this one is different. It returns an object instead of a string, and thus is deprecated.

foolip: Oh, so both exist. This will shadow the one on Element, so I guess deprecating simply means "don't use" in this case then.

foolip (approved):

PR #8249, opened 2020-12-23
Add missing features for SVGNumberList API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.8), for the `SVGNumberList` API.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/SVGNumberList

### DIFF
--- a/api/SVGNumberList.json
+++ b/api/SVGNumberList.json
@@ -46,6 +46,429 @@
           "standard_track": true,
           "deprecated": false
         }
+      },
+      "appendItem": {
+        "__compat": {
+          "support": {
+            "chrome": {
+              "version_added": "6"
+            },
+            "chrome_android": {
+              "version_added": "18"
+            },
+            "edge": {
+              "version_added": "12"
+            },
+            "firefox": {
+              "version_added": "3"
+            },
+            "firefox_android": {
+              "version_added": "4"
+            },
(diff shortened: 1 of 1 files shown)


api/SVGNumberList.json
```diff
@@ -46,6 +46,429 @@
...
+            }
+          },
+          "status": {
+            "experimental": false,
+            "standard_track": true,
+            "deprecated": false
+          }
+        }
+      },
+      "length": {
```
foolip, line 285: We have custom tests now, but it's still surprising that `length` wasn't supported from the beginning. Any idea what happened here?

vinyldarkscratch: It seems that these interfaces initially defined a property called `numberOfItems` instead of `length` (probably since `length` is commonly used for distance in SVG or something).  `length` came later to standardize the name across all of the web APIs.  At least, that's my assumption.

foolip: That makes sense, yeah.

foolip (approved):

PR #8258, opened 2020-12-23
Add missing features for SVGComponentTransferFunctionElement API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.8), for the `SVGComponentTransferFunctionElement` API.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/SVGComponentTransferFunctionElement

### DIFF
--- a/api/SVGComponentTransferFunctionElement.json
+++ b/api/SVGComponentTransferFunctionElement.json
@@ -46,6 +46,335 @@
           "standard_track": true,
           "deprecated": false
         }
+      },
+      "amplitude": {
+        "__compat": {
+          "support": {
+            "chrome": {
+              "version_added": "5"
+            },
+            "chrome_android": {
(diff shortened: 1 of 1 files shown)


foolip:
@vinyldarkscratch I see we're not testing an instance here, can you add a custom test?

vinyldarkscratch:
The results for this were actually added based upon manual testing, I forgot to add a note as such!  I added a custom test so collector 1.1.9 will match with these results.

foolip:
Ah, excellent!

foolip (approved):

PR #8263, opened 2020-12-24
CSSRule.type is deprecated

rachelandrew:
Working on https://github.com/mdn/content/issues/344

The CSSRule.type property is deprecated in the specification: https://drafts.csswg.org/cssom/#ref-for-dom-cssrule-type

I have been updating the docs to reflect this, this PR flags it as deprecated in BCD.


ddbeck (approved):
Thank you! 🎉

PR #8268, opened 2020-12-24
Add Chromium versions for WEBGL_compressed_texture_s3tc API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `WEBGL_compressed_texture_s3tc` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.8).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/WEBGL_compressed_texture_s3tc

### DIFF
--- a/api/WEBGL_compressed_texture_s3tc.json
+++ b/api/WEBGL_compressed_texture_s3tc.json
@@ -14,8 +14,7 @@
             }
           ],
           "chrome_android": {
-            "version_added": true,
-            "prefix": "WEBKIT_"
+            "version_added": false
           },
           "edge": {
             "version_added": "≤18"
@@ -46,8 +45,7 @@
             }
           ],
           "opera_android": {
-            "version_added": true,
-            "prefix": "WEBKIT_"
+            "version_added": false
           },
           "safari": {
             "version_added": "8"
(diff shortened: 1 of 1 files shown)


api/WEBGL_compressed_texture_s3tc.json
```diff
@@ -6,22 +6,17 @@
...
-              "version_added": true
+              "version_added": "26"
             },
             {
               "version_added": true,
               "prefix": "WEBKIT_"
             }
           ],
+          "chrome_android": {
+            "version_added": true,
```
foolip, line 17: Does this mean it was unprefixed in Chrome desktop but remains prefixed-only on Android? That sounds quite unusual, can you double check?

api/WEBGL_compressed_texture_s3tc.json
```diff
@@ -43,46 +38,31 @@
...
-              "version_added": true
+              "version_added": "15"
             },
             {
               "version_added": true,
               "prefix": "WEBKIT_"
             }
           ],
+          "opera_android": {
+            "version_added": true,
```
foolip, line 49: Same question as for Chrome Android here and below for Samsung+WebView.

api/WEBGL_compressed_texture_s3tc.json
vinyldarkscratch: I should have checked the prefixed support in Android as well; it looks like this API isn't supported in Chrome Android at all.  Fixing the data.

foolip (approved):

PR #8269, opened 2020-12-24
Add Chromium versions for WEBGL_compressed_texture_s3tc_srgb API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `WEBGL_compressed_texture_s3tc_srgb` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.8).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/WEBGL_compressed_texture_s3tc_srgb

### DIFF
--- a/api/WEBGL_compressed_texture_s3tc_srgb.json
+++ b/api/WEBGL_compressed_texture_s3tc_srgb.json
@@ -5,12 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/WEBGL_compressed_texture_s3tc_srgb",
         "support": {
           "chrome": {
-            "version_added": "60",
-            "version_removed": "78"
+            "version_added": "60"
           },
           "chrome_android": {
-            "version_added": "60",
-            "version_removed": "78"
+            "version_added": "60"
           },
           "edge": {
             "version_added": "≤79"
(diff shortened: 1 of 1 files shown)


api/WEBGL_compressed_texture_s3tc_srgb.json
```diff
@@ -5,10 +5,12 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/WEBGL_compressed_texture_s3tc_srgb",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "60",
+            "version_removed": "78"
```
foolip, line 9: https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/modules/webgl/webgl_compressed_texture_s3tc_srgb.idl is still in Chromium, so I wonder if this is a testing issue? Perhaps it's hardware dependent and the hardware used changed?

vinyldarkscratch: Just checked this more carefully, and it most likely is a hardware-dependent thing, yeah -- updating!

foolip (approved):

PR #8271, opened 2020-12-24
Add Chromium versions for WEBGL_draw_buffers API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `WEBGL_draw_buffers` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.8).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/WEBGL_draw_buffers

### DIFF
--- a/api/WEBGL_draw_buffers.json
+++ b/api/WEBGL_draw_buffers.json
@@ -5,10 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/WEBGL_draw_buffers",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "36"
           },
           "chrome_android": {
-            "version_added": true
+            "version_added": false
           },
           "edge": {
             "version_added": "17"
@@ -37,10 +37,10 @@
             "version_added": false
           },
           "opera": {
-            "version_added": true
+            "version_added": "23"
           },
           "opera_android": {
-            "version_added": true
+            "version_added": false
           },
           "safari": {
             "version_added": "9"
@@ -49,10 +49,10 @@
             "version_added": false
           },
           "samsunginternet_android": {
-            "version_added": true
+            "version_added": false
           },
           "webview_android": {
-            "version_added": true
+            "version_added": false
           }
         },
         "status": {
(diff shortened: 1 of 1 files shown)


api/WEBGL_draw_buffers.json
```diff
@@ -5,10 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/WEBGL_draw_buffers",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "36"
           },
           "chrome_android": {
-            "version_added": true
+            "version_added": false
```
foolip, line 11: The previous data was true. Are you sure it should be false, or is it not supported on some hardware?

foolip: Reading https://developer.mozilla.org/en-US/docs/Web/API/WEBGL_draw_buffers reveals that this extension is baked into WebGL 2, and the BCD entry is api.WebGL2RenderingContext.drawBuffers.

That is supported on Android, which makes it seem a bit implausible that this extension couldn't be supported on Android.

So further digging still required.

foolip:
@kenrussell perhaps you can advise here? Do you know if support for the WEBGL_draw_buffers extension is platform-dependent in some way? Testing https://mdn-bcd-collector.appspot.com/tests/api/WEBGL_draw_buffers on various Android devices it seems to not be supported, but if the functionality is the same as WebGL 2's `drawBuffers` I'm not sure why that would be...

kenrussell:
Support for many WebGL extensions is platform-dependent. The WebGL 1.0 WEBGL_draw_buffers extension is different from WebGL 2's built-in functionality. WebGL 2.0 only supports the functionality with ESSL 3.00 shaders, but the WebGL 1.0 extension only accepts ESSL 1.00 shaders. We decided not to risk transpiling ESSL 1.00 shaders to ESSL 3.00. Essentially, without the underlying OpenGL ES 2.0 extension - which is missing on most Android devices - it's very difficult to emulate this extension on top of OpenGL ES 3.0.

We recommend all users to upgrade to WebGL 2.0. With recent joint collaboration between Google and Apple, WebGL 2.0 should be universally available soon.

foolip:
Thanks @kenrussell! It sounds like it would be fair to say it's not supported on Android here in BCD. I did some searching and found https://en.wikipedia.org/wiki/OpenGL_ES#Platform_usage, and indeed support for OpenGL ES 2.0 seems very rare.

I've filed https://github.com/mdn/browser-compat-data/issues/8750 about looking over WebGL extensions in BCD more generally, as I expect this situation is quite common, but we usually haven't noticed it.

foolip (approved):

Issue #7844, opened 2020-12-14
Audit Chrome 45 versions


foolip:
The "Find Releases" tool at the bottom of https://omahaproxy.appspot.com/ is often used by BCD contributors to figure out which Chrome release a certain Chromium commit was first included in.

Back in [Sep 2015](https://chromium.googlesource.com/chromium/src/+/b59b6df51a249895fbba24f92b661f744e031546), the Blink and Chromium repositories were merged. That was for Chrome 47. Whether related to that merge or not, I have observed that trying to use the "Find Releases" tool with commits much older than that often incorrectly says that something was in Chrome 45, despite it actually being in an earlier release.

As an example, https://storage.googleapis.com/chromium-find-releases-static/243.html#24375f20cd603d56b8847b329209964c1281ba7a says "initially landed in 45.0.2454.101" but looking up the time of the commit with https://www.chromium.org/developers/calendar gives Chrome 43. I've seen this many times.

I would treat Chrome 45 as worthy of extra scrutiny in review, and moreover think we should go over the list of features in `npm run traverse chrome api 45` and make sure the data can be confirmed by http://mdn-bcd-collector.appspot.com/. Where we have no tests, the existing data might well be wrong because of this reason. 

@ddbeck @Elchi3 @jpmedley @sideshowbarker @vinyldarkscratch FYI

jculver050:
Close

foolip:
See https://github.com/mdn/browser-compat-data/pull/11631#discussion_r673354547 about `CSSGroupingRule`, a change that *did* happen in Chrome 45 but should perhaps be tweaked per https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#apis-moved-on-the-prototype-chain

foolip:
Burndown list to verify or fixed based on `npm run traverse chrome all 45`:

- [ ] api.CSSGroupingRule.* - https://github.com/mdn/browser-compat-data/issues/7844#issuecomment-883580883
- [ ] api.CSSKeyframesRule.appendRule (as insertRule)
- [ ] api.Client.postMessage
- [ ] api.Document.pointerlockchange_event (webkit prefix)
- [ ] api.Document.pointerlockerror_event (webkit prefix)
- [ ] api.FetchEvent.isReload
- [ ] api.FontFace.load
- [ ] api.FontFace.load
- [ ] api.FontFace.loaded
- [ ] api.FontFace.loaded
[...]

PR #8279, opened 2020-12-24
Add Chromium versions for PerformanceEntry API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `PerformanceEntry` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.8).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/PerformanceEntry.json
+++ b/api/PerformanceEntry.json
@@ -38,12 +38,24 @@
             "version_added": "8.5.0",
             "notes": "Stability: Experimental"
           },
-          "opera": {
-            "version_added": "33"
-          },
-          "opera_android": {
-            "version_added": "33"
-          },
+          "opera": [
+            {
+              "version_added": "33"
+            },
+            {
+              "version_added": "15",
+              "prefix": "webkit"
+            }
+          ],
+          "opera_android": [
+            {
+              "version_added": "33"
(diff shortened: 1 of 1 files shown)


foolip (approved):

api/PerformanceEntry.json
```diff
@@ -284,10 +296,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/PerformanceEntry/toJSON",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "45"
```
foolip, line 299: This is suspect because of https://github.com/mdn/browser-compat-data/issues/7844. The collector doesn't use commit mapping at all of course, but I double checked to confirm 45 anyway:
- `PerformanceEntry.prototype.hasOwnProperty('toJSON')` tested in Chrome 44 and 45
- https://chromium.googlesource.com/chromium/src/+/64d41ea9eeb27ea0897a190b62e44e2f1d57ed48 is Jul 9th, 2015
- https://www.chromium.org/developers/calendar/ branch point for 45 is Jul 10th, 2015

PR #8281, opened 2020-12-24
Add Chromium versions for RTCDataChannel API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `RTCDataChannel` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.8).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/RTCDataChannel


foolip:
This will need to test an instance, I'm pretty sure the interface object was originally not exposed.

PR #8282, opened 2020-12-24
Add Chromium versions for Comment API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `Comment` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.8).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Comment

### DIFF
--- a/api/Comment.json
+++ b/api/Comment.json
@@ -8,7 +8,7 @@
             "version_added": "1"
           },
           "chrome_android": {
-            "version_added": true
+            "version_added": "18"
           },
           "edge": {
             "version_added": "12"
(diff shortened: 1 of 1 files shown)


api/Comment.json
```diff
@@ -8,7 +8,7 @@
             "version_added": "1"
           },
           "chrome_android": {
-            "version_added": true
+            "version_added": "18"
```
foolip, line 11: Can you test an instance for the basic support? This is the DOM node for a `<!-- comment -->` and I'm sure this was supported in Chrome 1.

vinyldarkscratch: This line is for Chrome Android, actually!  Chrome Desktop is already set to 1, and the collector agrees.  :)

foolip: Oh, I wasn't paying attention :) Thanks!

foolip (approved):

PR #8301, opened 2020-12-25
Add WebView versions for Blob API

vinyldarkscratch:
This PR adds real values for WebView Android for the `Blob` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.8).  The collector obtains results based upon the latest WebView version (to determine if it is supported), then version numbers are copied from Chrome Android.  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Blob

### DIFF
--- a/api/Blob.json
+++ b/api/Blob.json
@@ -185,7 +185,7 @@
               "version_added": "1.0"
             },
             "webview_android": {
-              "version_added": false
+              "version_added": "≤37"
             }
           },
           "status": {
@@ -263,9 +263,16 @@
                 "version_removed": "1.5"
               }
             ],
-            "webview_android": {
-              "version_added": true
-            }
+            "webview_android": [
+              {
+                "version_added": "≤37"
+              },
+              {
+                "version_added": "≤37",
+                "version_removed": "≤37",
+                "prefix": "webkit"
+              }
+            ]
           },
           "status": {
             "experimental": false,
@@ -410,7 +417,7 @@
               "version_added": "1.0"
             },
             "webview_android": {
-              "version_added": false
+              "version_added": "≤37"
             }
           },
           "status": {
--- a/api/Blob.json
+++ b/api/Blob.json
@@ -263,16 +263,9 @@
                 "version_removed": "1.5"
               }
             ],
-            "webview_android": [
-              {
-                "version_added": "≤37"
-              },
-              {
-                "version_added": "≤37",
-                "version_removed": "≤37",
-                "prefix": "webkit"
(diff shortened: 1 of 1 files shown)


api/Blob.json
```diff
@@ -263,9 +263,16 @@
...
             ],
-            "webview_android": {
-              "version_added": true
-            }
+            "webview_android": [
+              {
+                "version_added": "≤37"
+              },
+              {
+                "version_added": "≤37",
```
foolip, line 271: The same range for both added and removed is something I'd half expect to fail the lint. @ddbeck @Elchi3 do you think this is OK to add?

foolip: What I've suggested in other PRs is to skip prefixed statements like this when both added and removed are <=37, since it's not giving much information at all.

ddbeck: Two version ranges feels like a `null` in disguise. My take: omitting this support statement is in the spirit of the irrelevant feature/flag removal guidelines and we should do that.

@foolip are there many more of these? If it's not common, we can probably say this a judgment call and move on. But if it's coming up repeatedly, we can write up a guideline against it.

foolip: This has come up in multiple PRs that @vinyldarkscratch has produced based on the collector project. It might occur for any prefixes APIs removed some time before Chrome 37, of which there are actually quite a few, I would guess between 5 and 25 such cases if we went looking.

I don't think this is analogous to irrelevant flags, however, prefixed APIs are just support with a different name, and I think we should accept the data and reject its removal where we can know what the real ranges are.

But anyway, I think we can deal with this by not adding these support statements for now. I'll file an issue about whether it should pass the lint later.

foolip (approved):

api/Blob.json
ddbeck: @foolip did you open an issue on whether `version_added == version_removed` should fail the lint?

Also, to clarify something I said before, I realize this is not analogous to irrelevant flags, but I do think this particular case is analogous to irrelevant features. Suppose that, in an alternate universe, BCD has distinct features for `api.Blob.slice` and `api.Blob.webkitslice` (instead of prefixes). The `webkitslice` feature would be eligible for removal, since it went away over 2 years go. That was the point I was trying to make with dropping the prefixed statements here.

foolip: I am back at work now and this was one of my TODOs. I've filed https://github.com/mdn/browser-compat-data/issues/8684 now.

PR #8311, opened 2020-12-25
Add WebView versions for DOMParser API

vinyldarkscratch:
This PR adds real values for WebView Android for the `DOMParser` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.8).  The collector obtains results based upon the latest WebView version (to determine if it is supported), then version numbers are copied from Chrome Android.  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/DOMParser

### DIFF
--- a/api/DOMParser.json
+++ b/api/DOMParser.json
@@ -86,7 +86,7 @@
               "version_added": "1.0"
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "1"
             }
           },
           "status": {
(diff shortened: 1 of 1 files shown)


api/DOMParser.json
```diff
@@ -181,7 +181,7 @@
                 "version_added": "3.0"
               },
               "webview_android": {
-                "version_added": "37"
+                "version_added": "4.4.3"
```
foolip, line 184: <=37 seems better here, more conservative.

foolip (approved):

PR #8323, opened 2020-12-25
Remove redundant prefix property for File.webkitRelativePath

vinyldarkscratch:
While updating data for the File API for Chrome, I noticed that Chrome had a prefix attribute for an already-prefixed attribute, `webkitRelativePath`.  It doesn't make sense to have a prefix listed for an already-prefixed attribute, so this PR removes said `prefix` property.


foolip (approved):

foolip:
Good catch, thanks!

PR #8329, opened 2020-12-25
Update WebView versions for GeolocationCoordinates API

vinyldarkscratch:
This PR updates and corrects the real values for WebView Android for the `GeolocationCoordinates` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.8).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/GeolocationCoordinates

### DIFF
--- a/api/GeolocationCoordinates.json
+++ b/api/GeolocationCoordinates.json
@@ -88,7 +88,7 @@
               "version_added": "79"
             },
             {
-              "version_added": true,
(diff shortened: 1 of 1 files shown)


foolip (approved):
These make more sense as ranges, nothing suggests this was added in exactly 37.

PR #8334, opened 2020-12-25
Add WebView versions for HTMLMediaElement API

vinyldarkscratch:
This PR adds real values for WebView Android for the `HTMLMediaElement` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.8).  The collector obtains results based upon the latest WebView version (to determine if it is supported), then version numbers are copied from Chrome Android.  Results are manually confirmed for accuracy.

### DIFF
--- a/api/HTMLMediaElement.json
+++ b/api/HTMLMediaElement.json
@@ -357,7 +357,7 @@
               "version_added": "4.0"
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "43"
             }
           },
           "status": {
@@ -653,7 +653,7 @@
               "version_added": "4.0"
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "43"
             }
           },
           "status": {
@@ -811,7 +811,7 @@
               "version_added": "4.0"
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "43"
             }
           },
           "status": {
@@ -859,7 +859,7 @@
               "version_added": "4.0"
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "43"
             }
           },
           "status": {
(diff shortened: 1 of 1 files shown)


api/HTMLMediaElement.json
```diff
@@ -357,7 +357,7 @@
               "version_added": "4.0"
             },
             "webview_android": {
-              "version_added": "43"
+              "version_added": true
```
foolip, line 360: Why this change? A lot of the "43" in this file are wrong and would be fixed by testing the instance, maybe fix all of those first to ensure that no known-wrong data is being mirrored or overwritten in this PR? We should be able to fill in a real value here after the 43s and true values elsewhere are fixed.

vinyldarkscratch: I actually don't know why I did this.  I think I might've been thinking "I'll set it to `true` and fix these later", but we've got an issue open to track these down.  Reverting these changes so that this PR is only increasing the amount of real values, not decreasing.

foolip: Yep, let's get back to these in https://github.com/mdn/browser-compat-data/issues/7843.

api/HTMLMediaElement.json
```diff
@@ -1003,7 +1003,7 @@
               "version_added": "5.0"
             },
             "webview_android": {
-              "version_added": "49"
+              "version_added": false
```
foolip, line 1006: Yep, confirmed by crbug.com/521319. I'll send a PR to link that issue after merging this.

api/HTMLMediaElement.json
```diff
@@ -3648,7 +3651,7 @@
...
```
foolip, line 3654: Confirmed by crbug.com/589500.

foolip (approved):

PR #8341, opened 2020-12-25
Remove duplicated prefixes in IDB APIs

vinyldarkscratch:
While working on the IDB APIs, I noticed that each of the individual properties had the same prefix data as the main API data.  However, this doesn't make sense; why would each of the members also be prefixed when the API itself is prefixed?  Additionally, I have verified this via manual testing.  This PR removes this copied prefix data.


foolip:
Excellent, thank you!

foolip:
This is big enough that I'll need to review the diff using my diffing tool, so can't do it on my phone just now.

foolip (approved):

PR #8349, opened 2020-12-25
Add WebView versions for MouseEvent API

vinyldarkscratch:
This PR adds real values for WebView Android for the `MouseEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.8).  The collector obtains results based upon the latest WebView version (to determine if it is supported), then version numbers are copied from Chrome Android.  Results are manually confirmed for accuracy.

### DIFF
--- a/api/MouseEvent.json
+++ b/api/MouseEvent.json
@@ -876,7 +876,7 @@
                 "version_added": "37"
               },
               {
-                "version_added": true,
+                "version_added": "≤37",
                 "version_removed": "37",
                 "prefix": "webkit"
               }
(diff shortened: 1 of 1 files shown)


vinyldarkscratch:
This fails the linter due to a bug fixed by #5576.  :(

foolip:
I guess there's no way to land this until the linter bug is fixed then?

vinyldarkscratch:
Only one of the three can be added without worrying about the linter, but for the other ones, the linter MUST be fixed.

PR #8350, opened 2020-12-25
Add WebView versions for MutationObserver API

vinyldarkscratch:
This PR adds real values for WebView Android for the `MutationObserver` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.8).  The collector obtains results based upon the latest WebView version (to determine if it is supported), then version numbers are copied from Chrome Android.  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/MutationObserver

### DIFF
--- a/api/MutationObserver.json
+++ b/api/MutationObserver.json
@@ -164,11 +164,11 @@
             ],
             "webview_android": [
               {
-                "version_added": true
+                "version_added": "≤37"
               },
               {
-                "version_added": true,
-                "version_removed": true,
+                "version_added": "≤37",
+                "version_removed": "≤37",
                 "prefix": "WebKit"
               }
             ]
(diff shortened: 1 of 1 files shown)


api/MutationObserver.json
```diff
@@ -164,11 +164,11 @@
             ],
             "webview_android": [
               {
-                "version_added": true
+                "version_added": "≤37"
               },
               {
-                "version_added": true,
-                "version_removed": true,
+                "version_added": "≤37",
```
foolip, line 170: I'm not a fan of these added+removed using the same value, but here both were true so it's still an improvement.

foolip (approved):

PR #8362, opened 2020-12-25
Add WebView versions for Performance API

vinyldarkscratch:
This PR adds real values for WebView Android for the `Performance` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.8).  The collector obtains results based upon the latest WebView version (to determine if it is supported), then version numbers are copied from Chrome Android.  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Performance

### DIFF
--- a/api/Performance.json
+++ b/api/Performance.json
@@ -116,16 +116,9 @@
                 "version_removed": "2.0"
               }
             ],
-            "webview_android": [
-              {
-                "version_added": "≤37"
-              },
-              {
-                "version_added": "≤37",
-                "version_removed": "≤37",
-                "prefix": "webkit"
-              }
-            ]
+            "webview_android": {
+              "version_added": "≤37"
+            }
           },
           "status": {
             "experimental": false,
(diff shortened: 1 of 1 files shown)


vinyldarkscratch:
I could not add ranges for `getEntries`, `getEntriesByName`, and `getEntriesByType` due to the bug resolved in #5576.

api/Performance.json
```diff
@@ -195,9 +195,16 @@
...
             ],
-            "webview_android": {
-              "version_added": true
-            }
+            "webview_android": [
+              {
+                "version_added": "≤37"
+              },
+              {
+                "version_added": "≤37",
```
foolip, line 203: I'm not too sure about adding ranges like this which give very little information about when it was supported.

api/Performance.json
```diff
@@ -393,9 +400,16 @@
...
             ],
-            "webview_android": {
-              "version_added": true
-            }
+            "webview_android": [
+              {
+                "version_added": "≤37"
+              },
+              {
+                "version_added": true,
```
foolip, line 408: This too doesn't seem that great. Just skipping the prefixed entries for this until we have better data to mirror seems best to me.

foolip (changes requested):

foolip (approved):

PR #8376, opened 2020-12-26
Add WebView versions for SpeechRecognition API

vinyldarkscratch:
This PR adds real values for WebView Android for the `SpeechRecognition` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.8).  The collector obtains results based upon the latest WebView version (to determine if it is supported), then version numbers are copied from Chrome Android.  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/SpeechRecognition

### DIFF
--- a/api/SpeechRecognition.json
+++ b/api/SpeechRecognition.json
@@ -158,7 +158,7 @@
               "notes": "You'll need to serve your code through a web server for recognition to work."
             },
             "webview_android": {
-              "version_added": true,
+              "version_added": "4.4.3",
               "notes": "You'll need to serve your code through a web server for recognition to work."
             }
           },
@@ -208,7 +208,7 @@
               "version_added": true
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "4.4.3"
             }
           },
           "status": {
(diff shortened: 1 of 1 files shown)


api/SpeechRecognition.json
```diff
@@ -158,7 +158,7 @@
               "notes": "You'll need to serve your code through a web server for recognition to work."
             },
             "webview_android": {
-              "version_added": true,
+              "version_added": "4.4.3",
```
foolip, line 161: Can we do <=37 instead as a more conservative assumption?

api/SpeechRecognition.json
```diff
@@ -309,7 +309,7 @@
               "notes": "You'll need to serve your code through a web server for recognition to work."
             },
             "webview_android": {
-              "version_added": true,
+              "version_added": false,
```
foolip, line 312: Why is this (`continuous`) set to false when it’s supported on Chrome for Android?

Same question for other true &rarr; false changes below.

vinyldarkscratch: Ah, I see what it is -- this interface is supported under a `webkit` prefix, so the collector isn't getting the correct results.  I'll close this PR for now.

PR #8377, opened 2020-12-26
Add WebView versions for SpeechRecognitionEvent API

vinyldarkscratch:
This PR adds real values for WebView Android for the `SpeechRecognitionEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.8).  The collector obtains results based upon the latest WebView version (to determine if it is supported), then version numbers are copied from Chrome Android.  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/SpeechRecognitionEvent

### DIFF
--- a/api/SpeechRecognitionEvent.json
+++ b/api/SpeechRecognitionEvent.json
@@ -47,7 +47,7 @@
           },
           "webview_android": {
             "prefix": "webkit",
-            "version_added": true,
+            "version_added": "4.4.3",
             "notes": "You'll need to serve your code through a web server for recognition to work."
           }
         },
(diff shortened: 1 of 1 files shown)


api/SpeechRecognitionEvent.json
```diff
@@ -47,7 +47,7 @@
           },
           "webview_android": {
             "prefix": "webkit",
-            "version_added": true,
+            "version_added": "4.4.3",
```
foolip, line 50: Since this API requires platform integration I'm not too confident about an exact number based on inference from Chrome for Android. Can we go with <=37 instead?

foolip (approved):

PR #8384, opened 2020-12-26
Add WebView versions for WheelEvent API

vinyldarkscratch:
This PR adds real values for WebView Android for the `WheelEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.8).  The collector obtains results based upon the latest WebView version (to determine if it is supported), then version numbers are copied from Chrome Android.  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/WheelEvent

### DIFF
--- a/api/WheelEvent.json
+++ b/api/WheelEvent.json
@@ -38,7 +38,7 @@
             "version_added": true
           },
           "webview_android": {
-            "version_added": "4.4.3"
+            "version_added": "≤37"
           }
         },
         "status": {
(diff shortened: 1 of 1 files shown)


foolip (commented):
Can we go with <=37 here unless we can be very sure? I'm less confident in assumptions based on Chrome for Android before 37 when WebView was more special.

foolip (approved):

PR #8386, opened 2020-12-26
Add WebView versions for WorkerGlobalScope API

vinyldarkscratch:
This PR adds real values for WebView Android for the `WorkerGlobalScope` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.8).  The collector obtains results based upon the latest WebView version (to determine if it is supported), then version numbers are copied from Chrome Android.  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/WorkerGlobalScope

### DIFF
--- a/api/WorkerGlobalScope.json
+++ b/api/WorkerGlobalScope.json
@@ -666,7 +666,7 @@
               "version_added": "4.0"
             },
             "webview_android": {
-              "version_added": false
+              "version_added": "40"
             }
           },
           "status": {
@@ -714,7 +714,7 @@
               "version_added": "4.0"
             },
             "webview_android": {
-              "version_added": false
+              "version_added": "40"
             }
           },
           "status": {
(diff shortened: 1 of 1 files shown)


api/WorkerGlobalScope.json
```diff
@@ -666,7 +666,7 @@
               "version_added": "4.0"
             },
             "webview_android": {
-              "version_added": true
+              "version_added": false
```
foolip, line 669: Why false here? It’s in Chrome for Android...

api/WorkerGlobalScope.json
```diff
@@ -714,7 +714,7 @@
...
```
foolip, line 717: Same here.

api/WorkerGlobalScope.json
vinyldarkscratch: Looks like this is a case of a bad test in the collector (I'm surprised that the rest of the tests worked properly).  I'll have a fix for this in collector v1.1.9; reverting and adding version numbers.

foolip: Great, thanks!

foolip (approved):

PR #8423, opened 2020-12-28
Add WebView versions for Window API

vinyldarkscratch:
This PR adds real values for WebView Android for the `Window` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.8).  The collector obtains results based upon the latest WebView version (to determine if it is supported), then version numbers are copied from Chrome Android.  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Window

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -2126,7 +2126,7 @@
               "version_added": null
             },
             "webview_android": {
-              "version_added": false
+              "version_added": null
             }
           },
           "status": {
--- a/api/Window.json
+++ b/api/Window.json
@@ -393,7 +393,7 @@
                 "version_added": "43"
               },
               {
-                "version_added": true,
+                "version_added": "1",
                 "prefix": "webkit"
               }
             ]
@@ -472,7 +472,7 @@
                 "version_added": "43"
               },
               {
-                "version_added": true,
+                "version_added": "1",
                 "prefix": "webkit"
               }
             ]
@@ -551,7 +551,7 @@
                 "version_added": "43"
               },
               {
-                "version_added": true,
+                "version_added": "1",
                 "prefix": "webkit"
               }
             ]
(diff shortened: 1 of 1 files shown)


api/Window.json
```diff
@@ -2126,7 +2126,7 @@
               "version_added": null
             },
             "webview_android": {
-              "version_added": null
+              "version_added": false
```
foolip, line 2129: This is null for all other Chromium-based browsers, and false for the rest. Seems like some work looking into this entry is needed, can you deal with it separately to simplify this PR?

api/Window.json
```diff
@@ -3528,7 +3528,7 @@
                 "version_added": false
               },
               "webview_android": {
-                "version_added": true
+                "version_added": false
```
foolip, line 3531: Matches Chrome data, so seems sensible.

api/Window.json
```diff
@@ -4709,7 +4709,7 @@
               "version_added": "7.0"
             },
             "webview_android": {
-              "version_added": false
+              "version_added": "57"
```
foolip, line 4712: This is the `onappinstalled` event handler attribute. Is web app installation via WebView really supported? Seems like it likely isn't, although the event handler attribute might still be incorrectly exposed I guess. Can you look a bit closer and see if this needs a note?

vinyldarkscratch: The collector gets data for WebView specifically, so this __is__ actually supported in WebView, though it's probably not supposed to be.

foolip: Okay, probably a case of a missing feature flag for the event handler attribute, although the event will never be fired.

foolip (approved):

PR #8433, opened 2020-12-29
Add Chromium versions for MediaQueryListEvent API

vinyldarkscratch:
This PR adds real values for Chromium for the MediaQueryListEvent API, as well as corrects the WebView data, based upon commit history.  Doing some digging around, I found that the API and all of its members were available since [Chrome 39](https://storage.googleapis.com/chromium-find-releases-static/5fc.html#5fc555b830a9b5b1536e242d8754d7e507bcc75a).  Additionally, the API is [not disabled in WebView](https://source.chromium.org/chromium/chromium/src/+/master:android_webview/tools/system_webview_shell/test/data/webexposed/not-webview-exposed.txt;l=1?q=not-webview-exposed&sq=&ss=chromium), and considering the `false` values had come from the initial wiki migration, I'm led to simply disregard them.

### DIFF
--- a/api/MediaQueryListEvent.json
+++ b/api/MediaQueryListEvent.json
@@ -5,10 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/MediaQueryListEvent",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "45"
           },
           "chrome_android": {
-            "version_added": true
+            "version_added": "45"
           },
           "edge": {
             "version_added": "≤79"
@@ -23,10 +23,10 @@
             "version_added": false
           },
           "opera": {
-            "version_added": true
+            "version_added": "32"
           },
           "opera_android": {
-            "version_added": true
+            "version_added": "32"
           },
           "safari": {
             "version_added": "14"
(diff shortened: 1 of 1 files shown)


foolip:
Chrome 45 cannot be trusted due to https://github.com/mdn/browser-compat-data/issues/7844. Can you check the time of the commit against https://www.chromium.org/developers/calendar to double check?

foolip:
I agree that the WebView data here is very likely wrong, this is an API that doesn't require much in the way of platform integration.

vinyldarkscratch:
Hmm, yeah, the date suggests a much earlier version than 45.  Marking as draft for now, will come back to this soon!

vinyldarkscratch:
Alrighty, this one's ready to go now!  I checked the commit date and mapped it against the Chrome release concordance spreadsheet Joe had shared, comparing the feature freeze dates to the commit date, which puts the commit (Aug 30, 2014) to land in Chrome 39 (Sep 12, 2014).  I then confirmed this via manual testing in BrowserStack.

foolip:
Thanks @vinyldarkscratch!

foolip (approved):

PR #8434, opened 2020-12-29
Add Chromium versions for CanvasRenderingContext2D API hit regions

vinyldarkscratch:
This PR adds real values for Chromium for the hit regions features of the CanvasRenderingContext2D API based upon Chromium commit history.  Tracking down the exact commit when this was added, I found it was introduced in [Chrome 45](https://storage.googleapis.com/chromium-find-releases-static/fcb.html#fcb1539c3aca167b9564907b6cf2592a1bc8cc8f), behind a flag.  Additionally, I had determined supported `addHitRegion` parameters by skimming through the [source C++ code](https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/modules/canvas/canvas2d/canvas_rendering_context_2d.cc;l=1230;drc=58bee6c238631d129396b09df273b99bb0143e8b) itself to find references, along with [commit history for the `fillRule` option](https://storage.googleapis.com/chromium-find-releases-static/503.html#5031c5f9c5dc07384bf1ce00de4f924e1437864c).

### DIFF
--- a/api/CanvasRenderingContext2D.json
+++ b/api/CanvasRenderingContext2D.json
@@ -275,6 +275,18 @@
                     "name": "Experimental Web Platform Features"
                   }
                 ]
+              },
+              "safari": {
+                "version_added": false
+              },
+              "safari_ios": {
+                "version_added": false
+              },
+              "samsunginternet_android": {
+                "version_added": false
+              },
+              "webview_android": {
+                "version_added": false
               }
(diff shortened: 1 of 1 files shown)


foolip:
This also needs double checking due to https://github.com/mdn/browser-compat-data/issues/7844.

PR #8442, opened 2020-12-29
Remove hit regions members of CanvasRenderingContext2D API

vinyldarkscratch:
This PR removes the hit regions members of the CanvasRenderingContext2D API.  The members qualify for irrelevance removal, as they are not supported in any browser by default within the last two years, and have since been removed from the spec.

Created based upon https://github.com/mdn/browser-compat-data/pull/8435#issuecomment-752145922.


foolip:
@Elchi3 said there was an open PR already, but I can't find it. Maybe go with this one then?

vinyldarkscratch:
It was mentioned that the PR was for the MDN docs content, rather than a BCD PR: https://github.com/mdn/content/pull/469

foolip:
I see. In that case it probably makes sense to merge it together with that PR. I'll leave this PR for someone else.

sideshowbarker (approved):

PR #8456, opened 2020-12-30
adding forced-color-adjust CSS property

rachelandrew:
BCD for `forced-color-adjust` as I am working on the page for MDN in https://github.com/mdn/content/issues/678

This is currently behind the `forced-colors` flag in Chrome 89, I've tested this.

The property is a standardized version of `-ms-high-contrast-adjust` which shipped in IE10 https://docs.microsoft.com/en-us/previous-versions/hh771863%28v%3dvs.85%29

### DIFF
--- a/css/properties/forced-color-adjust.json
+++ b/css/properties/forced-color-adjust.json
@@ -1,0 +1,63 @@
+{
+  "css": {
+    "properties": {
+      "forced-color-adjust": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/forced-color-adjust",
+          "support": {
+            "chrome": {
+              "version_added": "89",
+              "flags": [
+                {
+                  "type": "preference",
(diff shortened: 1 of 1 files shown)


rachelandrew:
Looking at this some more I think this went into Chrome flagged with the media feature in 79, the bug says it's on track to ship in 89 but it is currently still behind a pref. I can update the BCD if it does ship.

sideshowbarker:
Marked as draft per https://github.com/mdn/browser-compat-data/pull/8456#issuecomment-752619309. Or if that’s wrong and the intent is that we should merge this first, then update BCD later if it does ship, then please let me know, and I’ll go ahead and merge this.

rachelandrew:
@sideshowbarker I think merge it, it's useful for people to know this is there behind a flag if they are interested in the property. I'll update it if/when it does land.

sideshowbarker (approved):

PR #8462, opened 2020-12-30
Update Chromium versions for CanvasPattern API

vinyldarkscratch:
This PR updates and corrects the real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `CanvasPattern` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.8).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/CanvasPattern.json
+++ b/api/CanvasPattern.json
@@ -5,7 +5,7 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/CanvasPattern",
         "support": {
           "chrome": {
-            "version_added": "4"
+            "version_added": "6"
           },
           "chrome_android": {
             "version_added": "18"


api/CanvasPattern.json
```diff
@@ -5,7 +5,7 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/CanvasPattern",
         "support": {
           "chrome": {
-            "version_added": "4"
+            "version_added": "6"
```
foolip, line 8: This ought to match the data at https://github.com/mdn/browser-compat-data/blob/f3d419658370750830c5e8450922cba17b311000/api/CanvasRenderingContext2D.json#L1029, which says Chrome 1. Maybe the interface object just wasn't exposed originally?

vinyldarkscratch: Most likely so -- I'll reopen this PR with a custom test introduced in collector v1.1.9!

PR #8463, opened 2020-12-30
Update Opera versions for CanvasPattern API

vinyldarkscratch:
This PR updates and corrects the real values for Opera and Opera Android for the `CanvasPattern` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.8).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/CanvasPattern.json
+++ b/api/CanvasPattern.json
@@ -70,10 +70,10 @@
               "version_added": false
             },
             "opera": {
-              "version_added": "9"
+              "version_added": "55"
             },
             "opera_android": {
-              "version_added": "10.1"
+              "version_added": "48"
             },
             "safari": {
               "version_added": "11.1"


api/CanvasPattern.json
```diff
@@ -70,10 +70,10 @@
               "version_added": false
             },
             "opera": {
-              "version_added": "9"
+              "version_added": "55"
```
foolip, line 73: Are you sure about this? https://github.com/mdn/browser-compat-data/blob/f3d419658370750830c5e8450922cba17b311000/api/CanvasRenderingContext2D.json#L1046-L1051 says `CanvasPattern` was supported in Presto. Maybe we need to test an instance of `CanvasPattern` for this?

vinyldarkscratch: Good point; I'll recreate this PR using the collector instead, adding a custom test!

PR #8464, opened 2020-12-30
Set Chrome versions for api.CanvasPattern.setTransform.dommatrix

vinyldarkscratch:
This PR sets the Chromium versions for the `setTransform` method's `DOMMatrix` parameter for the `CanvasPattern` API based upon commit history.  It appears that the property was added at the same time as the method itself.

Source: https://source.chromium.org/chromium/chromium/src/+/f7028d961fb1cfd5d0f4abaecb582c61e0231c1f

### DIFF
--- a/api/CanvasPattern.json
+++ b/api/CanvasPattern.json
@@ -99,13 +99,13 @@
             "description": "<code>DOMMatrix</code> parameter supported",
             "support": {
               "chrome": {
-                "version_added": true
+                "version_added": "68"
               },
               "chrome_android": {
-                "version_added": true
+                "version_added": "68"
               },
               "edge": {
-                "version_added": true
+                "version_added": "79"
               },
               "firefox": {
                 "version_added": "79"
@@ -117,10 +117,10 @@
                 "version_added": false
               },
               "opera": {
-                "version_added": true
+                "version_added": "55"
               },
               "opera_android": {
-                "version_added": true
+                "version_added": "48"
               },
               "safari": {
                 "version_added": true
@@ -129,10 +129,10 @@
                 "version_added": true
               },
               "samsunginternet_android": {
-                "version_added": true
+                "version_added": "10.0"
               },
               "webview_android": {
-                "version_added": true
+                "version_added": "68"
               }
             },
             "status": {
--- a/api/CanvasPattern.json
+++ b/api/CanvasPattern.json
@@ -96,7 +96,7 @@
         },
         "dommatrix": {
           "__compat": {
-            "description": "<code>DOMMatrix</code> parameter supported",
+            "description": "Accepts a <code>DOMMatrix2DInit</code>-like object parameter",
             "support": {
               "chrome": {
                 "version_added": "68"


foolip (changes requested):
In https://source.chromium.org/chromium/chromium/src/+/f7028d961fb1cfd5d0f4abaecb582c61e0231c1f the argument was `SVGMatrix`, and then https://chromium.googlesource.com/chromium/src/+/26ea97726ef213927ed86a06836dd16a8b33275f (Chrome 68) changed it to the dictionary `DOMMatrix2DInit`, which is per spec.

https://github.com/mdn/browser-compat-data/pull/6365 added this entry to reflect https://bugzilla.mozilla.org/show_bug.cgi?id=1565997 which also changed the parameter from `SVGMatrix` to `DOMMatrix2DInit`. So I think to match that we should set Chrome 68 here.

Maybe the description should also be changed however to better reflect "it's now possible to use anything object which has `DOMMatrix2DInit` members, not just `SVGMatrix`", but I'm not sure how to phrase that. @ddbeck @Elchi3 WDYT?

ddbeck:
> Maybe the description should also be changed however to better reflect "it's now possible to use anything object which has `DOMMatrix2DInit` members, not just `SVGMatrix`", but I'm not sure how to phrase that. @ddbeck @Elchi3 WDYT?

Maybe something like, `Accepts a <code>DOMMatrix2DInit</code>-like object parameter`?

api/CanvasPattern.json
```diff
@@ -99,13 +99,13 @@
             "description": "<code>DOMMatrix</code> parameter supported",
```
foolip, line 99:
```suggestion
            "description": "Accepts a <code>DOMMatrix2DInit</code>-like object parameter",
```

foolip:
@ddbeck I've added that as a suggestion. Here are some other cases of dictionary/sequence arguments added later which we might look at for inspiration:

https://github.com/mdn/browser-compat-data/blob/6a10bd073bd46cd2509db0adf9d944036fcbca4c/api/Element.json#L1118

https://github.com/mdn/browser-compat-data/blob/6a10bd073bd46cd2509db0adf9d944036fcbca4c/api/EventTarget.json#L228

https://github.com/mdn/browser-compat-data/blob/6a10bd073bd46cd2509db0adf9d944036fcbca4c/api/URLSearchParams.json#L121

There's not really any consistency and I don't have a suggestion for how to improve the suggestion after looking at these :)

ddbeck:
Sorry, @foolip, I'm not quite sure what I'm meant to do here.

Are you looking for me to write another version of the description text? I don't think there's an obviously better or more consistent description text to go with here, given the details. So we can go with my suggestion or with a reasonable-sounding alternative. I'm not particularly opinionated on this one.

Or were you looking for a general review? Because this looks fine, though we ought to adopt _some_ change of the description text.

foolip:
@ddbeck I just wanted to show the existing forms and see if that would make you want to match some aspects of those other wordings. Since neither of us have a suggested improvement to the wording, I'll apply it and merge.

foolip (approved):

foolip:
I've also fixed the Safari versions in https://github.com/mdn/browser-compat-data/pull/8755.

PR #8465, opened 2020-12-30
Add Chromium versions for Path params of CanvasRenderingContext2D methods

vinyldarkscratch:
This PR adds Chromium versions for the `Path` parameters of various methods (`fill`, `stroke`, `drawFocusIfNeeded`, `clip`, `isPointInPath`, `isPointInStroke`) of the CanvasRenderingContext2D API, based upon commit history and dates.  By seeing the commit date and comparing it to the feature freeze for the Chrome versions, I've found that support was [enabled by default in Chrome 39](https://source.chromium.org/chromium/chromium/src/+/9a0ff42f9f7b5b5100bb57a77ff06456f01eaa27).

### DIFF
--- a/api/CanvasRenderingContext2D.json
+++ b/api/CanvasRenderingContext2D.json
@@ -833,10 +833,10 @@
             "description": "<code>Path</code> parameter",
             "support": {
               "chrome": {
-                "version_added": null
+                "version_added": "39"
               },
               "chrome_android": {
-                "version_added": null
+                "version_added": "39"
               },
               "edge": {
                 "version_added": "≤18"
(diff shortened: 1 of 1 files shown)


foolip:
I'm afraid that https://source.chromium.org/chromium/chromium/src/+/9a0ff42f9f7b5b5100bb57a77ff06456f01eaa27 is not the commit that enabled this. That removed the flag, but it already had `status=stable` so was enabled at some previous time. That was https://source.chromium.org/chromium/chromium/src/+/7e5f5d7b414aec49ac4c5821faf8a77666bb9635, which was Chrome 36 going by commit time and https://www.chromium.org/developers/calendar.

vinyldarkscratch:
Thanks for the double-check!  I should certainly pay more attention to when the flags are enabled by default...  😅   I've made the appropriate changes!

foolip:
Thanks @vinyldarkscratch!

foolip (approved):

PR #8476, opened 2020-12-31
Add Chromium versions for api.CanvasRenderingContext2D.currentTransform

vinyldarkscratch:
This PR adds Chromium versions for the CanvasRenderingContext2D API's `currentTransform` attribute based upon commit history.  Finding the [commit to add this feature](https://source.chromium.org/chromium/chromium/src/+/59ad6176c26b21afd28c5c64d0f294d3a72c52a6), I have mapped it to Chrome 32 as per its date and the according feature freeze.

### DIFF
--- a/api/CanvasRenderingContext2D.json
+++ b/api/CanvasRenderingContext2D.json
@@ -1123,6 +1123,7 @@
           "support": {
             "chrome": {
               "version_added": "32",
+              "version_removed": "68",
               "flags": [
                 {
                   "type": "preference",
@@ -1132,6 +1133,7 @@
             },
             "chrome_android": {
               "version_added": "32",
+              "version_removed": "68",
               "flags": [
                 {
                   "type": "preference",
@@ -1140,7 +1142,8 @@
               ]
             },
             "edge": {
-              "version_added": "≤18"
+              "version_added": "≤18",
+              "version_removed": "79"
             },
             "firefox": {
               "version_added": false,
@@ -1154,6 +1157,7 @@
             },
             "opera": {
               "version_added": "19",
+              "version_removed": "55",
               "flags": [
                 {
                   "type": "preference",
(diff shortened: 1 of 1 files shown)


foolip:
This again is a tricky case of interpreting Blink's IDL files. The attribute was added with `[EnabledAtRuntime=ExperimentalCanvasFeatures]` which means one has to also take into account whether that feature was enabled.

Is this not a case where http://mdn-bcd-collector.appspot.com/tests/api/CanvasRenderingContext2D/currentTransform can be used to have better confidence in the results? That tests suggests this isn't supported yet.

foolip:
Oh, I didn't read the diff carefully. This was a flagged entry...

api/CanvasRenderingContext2D.json
```diff
@@ -1122,7 +1122,7 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/CanvasRenderingContext2D/currentTransform",
           "support": {
             "chrome": {
-              "version_added": true,
+              "version_added": "32",
```
foolip, line 1125: Can you also mark it as removed in 68 based on https://storage.googleapis.com/chromium-find-releases-static/26e.html#26ea97726ef213927ed86a06836dd16a8b33275f?

I expect we can eventually delete this entry entirely, but might as well make this update to make that more clear.

foolip (changes requested):

api/CanvasRenderingContext2D.json
vinyldarkscratch: Thanks for catching this!  There's a lot of cases I need to watch out for in the commit history, haha...  😅 

Yeah, it looks like this qualifies as an irrelevant feature!

vinyldarkscratch: I opened up #8486 to remove the entire feature.

foolip (approved):

PR #8479, opened 2020-12-31
Add/update Chromium data for api.CanvasRenderingContext2D.imageSmoothingEnabled

vinyldarkscratch:
This PR adds the Chromium version for the prefixed version of the `imageSmoothingEnabled` attribute of the `CanvasRenderingContext2D` API based upon commit history, as well as copies the data to other Chromium browsers.  Based upon the [commit](https://source.chromium.org/chromium/chromium/src/+/a2112e8cdfbb4888002ce3dff9f90c3e48936003) date, I've determined that it had been added in Chrome 21.

### DIFF
--- a/api/CanvasRenderingContext2D.json
+++ b/api/CanvasRenderingContext2D.json
@@ -2421,14 +2421,21 @@
                 "version_added": "30"
               },
               {
-                "version_added": true,
+                "version_added": "21",
                 "version_removed": "30",
                 "prefix": "webkit"
               }
             ],
-            "chrome_android": {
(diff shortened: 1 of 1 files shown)


foolip:
Can you check if we get the same data from http://mdn-bcd-collector.appspot.com/tests/api/CanvasRenderingContext2D/webkitImageSmoothingEnabled?

vinyldarkscratch:
Forgot that the collector has some of the prefixed properties!  The collector agrees, good news!

foolip (approved):

PR #8486, opened 2020-12-31
Remove api.CanvasRenderingContext2D.currentTransform

vinyldarkscratch:
This PR removes the deprecated `currentTransform` member of the `CanvasRenderingContext2D` API.  Version numbers for Chrome are added in #8476, which indicate that it was removed in Chrome 68 (released over two years ago).  Since it's not supported in any other browser, and is deprecated, it qualifies for removal.


api/CanvasRenderingContext2D.json
```diff
@@ -1117,116 +1117,6 @@
...
-                  "type": "preference",
-                  "name": "Experimental Web Platform Features"
-                }
-              ]
-            },
-            "chrome_android": {
-              "version_added": false
-            },
-            "edge": {
-              "version_added": "≤18"
```
foolip, line 1137: Was it in Edge? I'm not sure what we should do with things that were in IE or EdgeHTML but not in Chromium, Gecko or WebKit. That's not really a well explored part of the removal guideline, right?

api/CanvasRenderingContext2D.json
```diff
@@ -1117,116 +1117,6 @@
...
-              "version_added": false
-            }
-          },
-          "status": {
-            "experimental": true,
-            "standard_track": false,
-            "deprecated": true
-          }
-        },
-        "DOMMatrix_return_value": {
```
foolip, line 1175: This removal wasn't intentional I guess?

vinyldarkscratch: It was -- this is a sub feature of the removed attribute.

api/CanvasRenderingContext2D.json
vinyldarkscratch: Well, the removal guidelines say no browser versions released in the past two years, so since Edge 18 was released October 2018, I'd say we're good?

foolip: If we apply this guideline fully, won't we be removing a bunch of APIs that IE/Edge have which are necessary alternatives to other APIs which yet aren't simple aliases of the cross-browser APIs?

Listing all APIs which could be removed under the guideline might help show if there's an issue like this.

wieseaka (changes requested):
[]()

api/CanvasRenderingContext2D.json
foolip: Oh, didn't spot that.

api/CanvasRenderingContext2D.json
ddbeck: I'd prefer to keep this data until at least 15 Jan 2022, another year.

The relevant bit of the irrelevance feature guideline says a feature can be removed if it "was implemented and has since been removed from all browsers dating back two or more years ago." I admit the wording is a bit awkward, but it's supposed to mean that two years have elapsed since the removal, not the ship date of the last version containing that feature (otherwise we wouldn't have bothered with the subsequent bullet in the guideline, with the five year horizon).

foolip:
I'll go ahead and close this, a year from now it will probably conflict, and hopefully we'll have scripts to enumerate candidates for removal which would re-surface this.

PR #8487, opened 2020-12-31
Add Safari versions for api.MediaQueryList.EventListener_objects

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `EventListener_objects` member of the `MediaQueryList` API, based upon commit history and date.

Commit: https://github.com/WebKit/WebKit/commit/dc95aca04b6b10c4023c14921fad1463c4bb5f0a#diff-d1998af1820803564b137487dc59bba1a9dbc682eb764335351af22f02463818

### DIFF
--- a/api/MediaQueryList.json
+++ b/api/MediaQueryList.json
@@ -78,10 +78,10 @@
               "version_added": false
             },
             "safari": {
-              "version_added": null
+              "version_added": "14"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "14"
             },
             "samsunginternet_android": {
               "version_added": "5.0"


foolip:
Using the date isn't a very reliable method for WebKit since one also has to know the branch date, and that isn't publishing anywhere, although one can figure it out after the fact.

If you have a checkout of WebKit, you could list all branches and tags which contain this commit and figure it out that way.

Otherwise, maybe write a custom test for this?

foolip:
I checked it out myself and can confirm. `Object.getPrototypeOf(MediaQueryList)==EventTarget` is true in Safari 14, and the change was made after Safari 13.1 was released. Also, WebKit trunk was then [610.1.11](https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Configurations/Version.xcconfig?rev=260243), which also maps to Safari 14.

foolip (approved):

PR #8569, opened 2021-01-04
Add Safari versions for ApplicationCache API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `ApplicationCache` API, based upon manual testing.

Test Code Used: `'ApplicationCache' in self`

### DIFF
--- a/api/ApplicationCache.json
+++ b/api/ApplicationCache.json
@@ -33,10 +33,10 @@
             "version_removed": "60"
           },
           "safari": {
-            "version_added": "≤4"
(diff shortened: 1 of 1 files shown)


foolip:
I love the "Test Code Used" bit, makes it much easier to review at a glance!

foolip:
I wonder if a better test would be for `window.applicationCache` however? That's the actual instance here, the interface object might not have been exposed from the beginning.

sideshowbarker (approved):

PR #8672, opened 2021-01-06
Add Safari versions for SVGAnimatedPoints API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `SVGAnimatedPoints` API, based upon manual testing.

Test Code Used: `var instance = document.createElementNS('http://www.w3.org/2000/svg', 'polygon')`


ddbeck (approved):
Thank you! 🎉

PR #8680, opened 2021-01-07
add BCD for HTMLPortalElement

rachelandrew:
The API HTMLPortalElement is behind a flag in Chrome.

- https://web.dev/hands-on-portals/
- https://wicg.github.io/portals/

I'm writing the docs for it, at the moment the MDN page doesn't exist.

### DIFF
--- a/api/HTMLPortalElement.json
+++ b/api/HTMLPortalElement.json
@@ -1,0 +1,68 @@
+{
+  "api": {
+    "HTMLPortalElement": {
+      "__compat": {
+        "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLPortalElement",
+        "support": {
+          "chrome": {
+            "version_added": "85",
+            "flags": [
+              {
+                "type": "preference",
+                "name": "#enable-portals",
+                "value_to_set": "enabled"
+              }
+            ],
+            "notes": "See <a href='https://www.chromestatus.com/feature/4828882419056640'>Chrome Platform Status</a>."
+          },
+          "chrome_android": {
+            "version_added": "85",
+            "flags": [
+              {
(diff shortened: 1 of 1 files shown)


ddbeck:
The first link talks about origin trials. Is the flag the only requirement? Or do you need both the flag and the trial enrollment right now?

rachelandrew:
To run it on your site you need the origin trial, to play with it you can toggle the flag. Origin Trials let you expose things to people who then don't need to toggle flags.

ddbeck:
Thanks for the clarification, Rachel. The reason I asked is that, in the past, @jpmedley has asked us to not include features affected by Chrome origin trials. Not sure if that quite applies here. Tagging him for review.

rachelandrew:
@jpmedley let me know as I'll need to do BCD for the other interfaces and the `<portal>` element.

rachelandrew:
Looks like the trial finished in November, so it's just the flags: https://developer.chrome.com/origintrials/#/view_trial/-7680889164480380927

ddbeck (approved):
OK, checks out. Thank you! 🎉

jpmedley:
MDN has agreed not to include origin trials in BCD. We would prefer that rule for flags as well. MDN accepts flag data, but anyone working for Google is asked to not submit such data.

ddbeck:
@jpmedley I tested that the element was recognized with the flag on (without origin enrollment), so I'm satisfied that we have not included origin data here.

When it comes to the Chrome flag policy question, I'd refer you to my previous guidance on approaching this problem in https://github.com/mdn/browser-compat-data/pull/5504#issuecomment-591587479. Policy discussion should take place in #3318 (or a new issue or PR, as needed).

PR #8682, opened 2021-01-08
Add note about nightly-only for private_class_fields (fixed #7446)

### LINKED ISSUE #7446: Private_class_fields Browser compatibility only in nightly versions of firefox functioning
see https://searchfox.org/mozilla-central/rev/1d34bd022de0b55c81d9db6026f69bda1d4a86d2/js/xpconnect/src/XPCJSContext.cpp#1014

and 

https://bugzilla.mozilla.org/show_bug.cgi?id=1677433

arai-a:
Private class fields are enabled only in nightly, behind pref.
https://searchfox.org/mozilla-central/rev/1d34bd022de0b55c81d9db6026f69bda1d4a86d2/js/xpconnect/src/XPCJSContext.cpp#1017-1020

Like other nightly-only feature, added `"notes": "Available only in nightly builds.",`.

### DIFF
--- a/javascript/classes.json
+++ b/javascript/classes.json
@@ -489,7 +489,8 @@
               "version_added": "79"
             },
             "firefox": {
-              "version_added": "81",
+              "version_added": false,
+              "notes": "Available only in nightly builds. See <a href='https://bugzil.la/1562054'>bug 1562054</a>.",
               "flags": [
                 {
                   "type": "preference",
@@ -504,7 +505,8 @@
               ]
             },
             "firefox_android": {
-              "version_added": "81",
+              "version_added": false,
+              "notes": "Available only in nightly builds. See <a href='https://bugzil.la/1562054'>bug 1562054</a>.",
               "flags": [
                 {
                   "type": "preference",


javascript/classes.json
```diff
@@ -490,6 +490,7 @@
             },
             "firefox": {
```
ddbeck, line 491: We ought to replace what we have here (and in Firefox for Android) with something like:

```
{
  "version_added": false
}
```

It's a bit of a quirky situation though, so we could plausibly provide a "See <a href='bugzil.la/1562054'>bug 1562054</a>." note here, if you like.

ddbeck (changes requested):
Hi @arai-a and thank you for this PR. I see this is your first PR on the project—welcome!

The nightly-only stuff has been a bit of an ongoing consistency problem. If it's gated to nightly, then the `version_added` value isn't strictly correct. Unfortunately, I think the right thing to do is to report this as unsupported. A little more detail in a line comment.

Thank you!

javascript/classes.json
arai-a: is it okay to add the version information about which nightly supports the feature, in "Notes" section?
or should we avoid that?

ddbeck: I think it's best to avoid it, at least in this case. There's very little expectation that developers are going to run old nightly releases and the version numbers can confuse things for stable users.

javascript/classes.json
```diff
@@ -489,7 +489,8 @@
               "version_added": "79"
             },
             "firefox": {
-              "version_added": "81",
+              "version_added": false,
+              "notes": "Available only in nightly builds. See <a href='https://bugzilla.mozilla.org/show_bug.cgi?id=1562054'>bug 1562054</a>.",
```
ddbeck, line 493: These links need shortening, which is why the linter is failing. See https://github.com/mdn/browser-compat-data/pull/8682/checks?check_run_id=1668964873#step:5:2339

ddbeck (changes requested):
We need to satisfy the linter here and then we can merge. Thank you, @arai-a!

ddbeck (approved):
Thank you! 🎉

PR #8700, opened 2021-01-11
Add Chrome for Android version to Web NFC

beaufortfrancois:
This PR adds Chrome for Android version 89 to Web NFC. See https://groups.google.com/a/chromium.org/g/blink-dev/c/aoj06_plqPc/m/D-6ahhfuAwAJ

Disclaimer: I'm an editor of the spec.


ddbeck (approved):
Looks good. Thank you! 🎉

PR #8701, opened 2021-01-11
Add Opera Android version for CSS @property

rachelandrew:
Fixes: https://github.com/mdn/content/issues/1102

Issue reported that Opera Android had support, this seems to be from version 60 based on Chromium 85.


ddbeck (approved):
Seems likely. Thank you! 🎉

PR #8712, opened 2021-01-12
Add Safari support for Animation.currentTime/play/pause. Fixes #7992

### LINKED ISSUE #7992: api.Animation.currentTime - Safari support

`Animation.prototype.currentTime` is reported to not exist at all on Safari. This seems to apply also for `.pause` and `.play` (and possibly others).

Considering I can use these features in Safari 14 and Safari 13.1 at least, it seems this information is at least inaccurate for Safari 14 and Safari 13.1.

I tested with [this CodeSandbox](https://codesandbox.io/s/animation-sync-bwbz7?file=/src/App.js) which allows you to add and remove spinners whilst the animations stay in sync using a combo of `currentTime`, `pause`, and `play`. You can toggle the sync code on/off for new spinners with a checkbox, and the not synced ones are shown in red. You can remove a spinner by clicking on it. 

Screenshot from the demo showing some in sync and some not (though they were added to the page at different times). You should really try the demo out first though. 

image

It's written in React but it should be a fairly accessible demo even for those who are not familiar. The important code is this, which runs when the spinners are added to the page

```js
const node = ref.current; // Get our new spinner from React
const spinners = Array.from(document.querySelectorAll(".Spinner"));

if (spinners.length > 1) {
  for (const spinner of spinners) {
    spinner.getAnimations()[0].pause();
  }

  const spinner = spinners.find((spinner) => spinner !== node);

  node.getAnimations()[0].currentTime = spinner.getAnimations()[0].currentTime;

  for (const spinner of spinners) {
    spinner.getAnimations()[0].play();
  }
}
```

This *definitely* works in Safari 14 running on macOS Catalina which I've ran on my machine. I've also tested this demo on Safari 13.1 on macOS Catalina via Browserstack where it works. I don't have access to a Safari 13 instance but this demo doesn't work on Safari 12 (though this is because getAnimations isn't support there at all). 

<details>

* Query: `api.Animation.currentTime`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/Animation/currentTime
* Report started: 2020-12-17T16:18:58.398Z

</details>

nathggns:
This addresses #7992 by indicating that Animation.currentTime/play/pause support was introduced in Safari 13.1 (and mobile Safari 13.4). It's likely that there are probably more properties of Animation available in Safari, but these are the ones I was able to verify (as detailed in the ticket)

### DIFF
--- a/api/Animation.json
+++ b/api/Animation.json
@@ -350,10 +350,10 @@
               "version_added": "26"
             },
             "safari": {
-              "version_added": false
+              "version_added": "13.1"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "13.4"
             },
             "samsunginternet_android": {
               "version_added": "4.0"
@@ -916,10 +916,10 @@
               "version_added": "26"
             },
             "safari": {
-              "version_added": false
+              "version_added": "13.1"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "13.4"
             },
             "samsunginternet_android": {
               "version_added": "4.0"
@@ -1090,10 +1090,10 @@
               "version_added": "26"
             },
             "safari": {
-              "version_added": false
+              "version_added": "13.1"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "13.4"
             },
             "samsunginternet_android": {
               "version_added": "4.0"


ddbeck (approved):
Excellent work. Thank you, @nathggns! 🎉

Issue #8709, opened 2021-01-12
Safari supports pointer events since version 13


Rezakarimi66:
#### What information was incorrect, unhelpful, or incomplete?
Safari supports pointer events.  

#### What did you expect to see?
https://caniuse.com/?search=pointerevent

#### Did you test this? If so, how?
No

<details>
<summary>MDN page report details</summary>

* Query: `api.HTMLElement.pointerenter_event`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/pointerenter_event
* Report started: 2021-01-12T07:37:56.350Z

</details>

ddbeck:
Thanks for reporting this, @Rezakarimi66.

The [Safari 13 release notes](https://developer.apple.com/documentation/safari-release-notes/safari-13-release-notes) seem to back up caniuse's data:

> Added support for the Pointer Events API enabling consistent access to mouse, trackpad, touch, and Apple Pencil events

Narrowly, this can be fixed with an edit to the Safari and Safari for iOS data in the `pointerenter_event` feature in [`HTMLElement.json`](https://github.com/mdn/browser-compat-data/blob/master/api/HTMLElement.json#L2510-L2590). I'd welcome a PR addressing this specifically, if you (or another contributor) would like to fix that specifically.

It's likely, however, that we'd want to test and update the other events in [the Pointer Events specification](https://w3c.github.io/pointerevents/), too, such as `pointerover` and `pointerdown`. There's a number of features involved, unfortunately, so the first step is probably to make a list of the features that the spec involves.

ddbeck:
#8713 fixes this for `pointerenter`. 👍

ddbeck:
I missed some previous related issues. But since this one contains more information, I'm marking them as duplicates of this one.

- #8747 - `pointerup`
- #7898, #8788 - `pointerdown`

ddbeck:
#9455 fixes a bunch of these.

Also, for the record, the Pointer Events spec defines these events:

- `gotpointercapture`
- `lostpointercapture`
- `pointercancel`
- `pointerdown`
- `pointerenter`
- `pointerleave`
- `pointermove`
- `pointerout`
- `pointerover`
- `pointerrawupdate`
- `pointerup`

ddbeck:
When complete, #9596 ought to resolve the remaining Safari pointer events data.

PR #8713, opened 2021-01-12
Safari supports pointerenter events since version 13

Rezakarimi66:
It seems that Safari supports pointerevnts since version 13 ( https://developer.apple.com/documentation/safari-release-notes/safari-13-release-notes ). I just updated the related json file. I have not test it yet on a real device. 
Related issue: https://github.com/mdn/browser-compat-data/issues/8709

### DIFF
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -2570,10 +2570,10 @@
               "version_added": null
             },
             "safari": {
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Thank you, @Rezakarimi66! 🎉

PR #8726, opened 2021-01-12
Fix Chrome + Samsung Internet versions for TextTrackCue

foolip:
This was definitely wrong, because Chrome 9 was shipped before
https://trac.webkit.org/changeset/97926/webkit which added this
interface to WebKit.

http://mdn-bcd-collector.appspot.com/tests/api/TextTrackCue was tested
in Chrome 22 and 23 to confirm.

The new data is consitent with data for HTMLTrackElement, TextTrack and
other APIs that all go together for the basic <track> functionality.

The Samsung Internet data must have been mirrored from Chrome and thus
fixed too. The data for other browsers is left alone even though there
are likely errors. In particular the Safari version numbers are
inconsistent, but not updated here.

Follow-up to https://github.com/mdn/browser-compat-data/pull/7640.

### DIFF
--- a/api/TextTrackCue.json
+++ b/api/TextTrackCue.json
@@ -5,10 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/TextTrackCue",
         "support": {
           "chrome": {
-            "version_added": "9"
+            "version_added": "23"
           },
           "chrome_android": {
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Looks good. Thank you! 🎉

PR #8727, opened 2021-01-12
Add data for HTMLOptionsCollection in WebKit

foolip:
http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8802 was
tested in Safari 4 to confirm that `select.options` was supported as
a `HTMLOptionsCollection`, but the interface object wasn't exposed.

https://trac.webkit.org/changeset/15321/webkit is what made
`select.options` a `HTMLOptionsCollection` instead of `HTMLCollection`.

Based on the date (2006) and the release branch we can say this was in
Safari 3:
https://trac.webkit.org/browser/webkit/releases/Apple/Safari%203.0/WebCore/html/HTMLOptionsCollection.idl

The remove() method was added later in
https://trac.webkit.org/changeset/28584/webkit, which was in Safari 3.1:
https://trac.webkit.org/browser/webkit/releases/Apple/Safari%203.1/WebCore/html/HTMLOptionsCollection.idl

Versions for other browsers were inferred based on the WebKit version.

### DIFF
--- a/api/HTMLOptionsCollection.json
+++ b/api/HTMLOptionsCollection.json
@@ -5,10 +5,10 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLOptionsCollection",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "1"
           },
           "chrome_android": {
-            "version_added": true
+            "version_added": "18"
           },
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Thank you! 🎉

PR #8735, opened 2021-01-13
Mark TimeRanges as shipped in Firefox 4

foolip:
This matches api.HTMLMediaElement.buffered, the first property that used
TimeRanges.

This was implemented in https://bugzilla.mozilla.org/show_bug.cgi?id=589561,
which by the date also matches Firefox 4.

State of the IDL file in the last patch on the bug:
https://hg.mozilla.org/mozilla-central/file/37f61e9d618a32ade3367c514598454b9b735a54/dom/interfaces/html/nsIDOMTimeRanges.idl

### DIFF
--- a/api/TimeRanges.json
+++ b/api/TimeRanges.json
@@ -14,10 +14,10 @@
             "version_added": "12"
           },
           "firefox": {
-            "version_added": true
+            "version_added": "4"
           },
           "firefox_android": {
-            "version_added": true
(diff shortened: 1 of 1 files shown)


foolip:
This is a simple update I came across when looking at https://github.com/mdn/browser-compat-data/pull/8734 using the same bug.

sideshowbarker (approved):

PR #8741, opened 2021-01-13
Fix crossOriginIsolated status for Chrome

Niek:
See: https://www.chromestatus.com/feature/5690888397258752

### DIFF
--- a/api/WindowOrWorkerGlobalScope.json
+++ b/api/WindowOrWorkerGlobalScope.json
@@ -576,7 +576,7 @@
               "version_added": "87"
             },
             "edge": {
-              "version_added": false
+              "version_added": "87"
             },
             "firefox": {
               "version_added": "72"
@@ -591,7 +591,7 @@
               "version_added": false
             },
             "opera": {
-              "version_added": false
+              "version_added": "73"
             },
             "opera_android": {
               "version_added": false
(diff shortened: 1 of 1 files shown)


api/WindowOrWorkerGlobalScope.json
```diff
@@ -570,10 +570,10 @@
...
             "chrome": {
-              "version_added": false
+              "version_added": "87"
             },
             "chrome_android": {
-              "version_added": false
+              "version_added": "87"
             },
             "edge": {
               "version_added": false
```
ddbeck, line 579: suggestion: mirror to other Chromium-derived browsers

According to the [intent to ship](https://groups.google.com/a/chromium.org/g/blink-dev/c/cJ5dXIcQCsc) and some direct testing of my own, this shipped on all of these browsers except WebView. Can we set Edge to 87 and Opera (desktop) to 73?

```suggestion
              "version_added": "87"
```

ddbeck (changes requested):
Hi @Niek and thank you for this PR. I see that it's your first on the project—welcome!

This looks good, but I have one suggestion (see line comment) to fill out the data for the other Chromium-derived browsers. Thank you again!

api/WindowOrWorkerGlobalScope.json
Niek: Thanks for the suggestion @ddbeck - I have updated the PR now.

ddbeck (approved):
Looks great. Thank you! 🎉

PR #8033, opened 2020-12-19
Add missing getInstalledRelatedApps feature for Navigator API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from a spec in Editor's Draft or more and is supported in at least one browser.  This particular PR adds missing features, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.7), for `Navigator.getInstalledRelatedApps()`.

Spec: https://wicg.github.io/get-installed-related-apps/spec/
IDL: https://github.com/w3c/webref/blob/master/ed/idl/get-installed-related-apps.idl
Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Navigator/getInstalledRelatedApps

PR #8742, opened 2021-01-13
Add notes for getInstalledRelatedApps desktop platform support

foolip:
Windows support was enabled in Chromium 85:
https://storage.googleapis.com/chromium-find-releases-static/f6a.html#f6a7ad6cdaebeae308f52b168936155e6e62443f

Additional sources:
https://groups.google.com/a/chromium.org/g/blink-dev/c/wB7mDJqrUZ8/m/Hbj0qxj8BQAJ
https://groups.google.com/a/chromium.org/g/blink-dev/c/aq1Zxc8D_uI/m/zHkFrzs5AQAJ
https://chromestatus.com/feature/5695378309513216

Follow-up to https://github.com/mdn/browser-compat-data/pull/8033.

### DIFF
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -1115,10 +1115,13 @@
             "chrome": [
               {
                 "version_added": "85",
+                "partial_implementation": true,
                 "notes": "Only supported on Windows. Resolves with an empty array on other platforms."
               },
               {
                 "version_added": "80",
+                "version_removed": "85",
+                "partial_implementation": true,
                 "notes": "Always resolves with an empty array."
               }
             ],
@@ -1128,10 +1131,13 @@
             "edge": [
               {
                 "version_added": "85",
+                "partial_implementation": true,
                 "notes": "Only supported on Windows. Resolves with an empty array on other platforms."
               },
               {
                 "version_added": "80",
+                "version_removed": "85",
+                "partial_implementation": true,
                 "notes": "Always resolves with an empty array."
               }
             ],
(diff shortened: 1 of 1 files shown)


api/Navigator.json
```diff
@@ -1112,15 +1112,29 @@
       "getInstalledRelatedApps": {
         "__compat": {
           "support": {
-            "chrome": {
-              "version_added": "80"
-            },
+            "chrome": [
+              {
+                "version_added": "85",
+                "notes": "Only supported on Windows. Resolves with an empty array on other platforms."
```
ddbeck, line 1118: I'm not well-versed in this API but I think this might be a case where partial_implementation applies: https://github.com/mdn/browser-compat-data/blob/master/docs/data-guidelines.md#operating-system-limitations-imply-partial_implementation I'd be interested in getting your take on this

foolip: I also don't know this API well, but I think it could be argued both ways. A native app has to be associated with an origin in some way to be related, and I guess that's an app store feature or some metadata in the installed app. Both depend on some changes to the OS I presume. So there are probably some platforms where there can't be any related apps by definition, and then returning an empty array is correct, not an incomplete implementation.

However, I don't know what the state of this on all platforms are, so I can't say if implementation is possible or not.

Calling it partial seems OK with me, it calls attention to the problem and isn't clearly wrong.

foolip: I've added `partial_implementation`, and also `version_removed` even though nothing was really removed, but the initial note applies only to a range of version.

ddbeck (approved):
Thank you! 🎉

PR #8755, opened 2021-01-14
Add/update Safari versions for CanvasPattern.setTransform

foolip:
This was implemented in https://trac.webkit.org/changeset/225121/webkit,
where the argument was a DOMMatrix2DInit dictionary from the beginning.

That revision maps to WebKit trunk at version 605.1.15:
https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Configurations/Version.xcconfig?rev=225121

Given that and the date, the right versions are 11.1 / 11.3, not 11,
which was released before this change.

### DIFF
--- a/api/CanvasPattern.json
+++ b/api/CanvasPattern.json
@@ -79,7 +79,7 @@
               "version_added": "11.1"
             },
             "safari_ios": {
-              "version_added": "11"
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Thank you! 🎉

PR #8768, opened 2021-01-15
Replace matrix builds with with named jobs

ddbeck:
As a prerequisite to #7171 and turning on branch protection rules, this PR replaces the matrix builds with jobs named for the specific release types of Node.js that we're targeting (Maintenance LTS and Active LTS). This allows us to require the release types succeed, without naming the exact version numbers in the branch protection rules.

~As a prerequisite to #7171 and turning on branch protection rules, this PR adds a new job to our testing workflow to summarize the results of all the matrix builds. This allows us to require the the complete build matrix succeeds, without specifying what builds make up that matrix.~

This was raised as a potential issue by @foolip in https://github.com/mdn/browser-compat-data/issues/7171#issuecomment-723052182:

> When one changes which checks are required, all open PRs turn into a pending state, and typically require rebasing to trigger the new check. For WPT, this has been pretty annoying to deal with, because we have so many open PRs.
> 
> For BCD, it would be smart to think twice about what the names of the checks we require are. Make sure we're happy with them, and that they don't include anything like version numbers which will change over time.

This ought to avoid an easy-to-anticipate problem for BCD: invalidating the state of PRs when the supported versions of Node.js change.

You can see what a failure would look like in https://github.com/ddbeck/browser-compat-data/pull/14.

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -10,20 +10,27 @@ on:
     branches: [ ] # Test on push to every branch
 
 jobs:
-  test:
-
+  maintenance_lts:
+    name: "Node.js maintenance LTS"
     runs-on: ubuntu-latest
-
-    strategy:
-      matrix:
-        node-version: [ 10.x, 14.x ]
-
     steps:
     - uses: actions/checkout@v2
-    - name: Use Node.js ${{ matrix.node-version }}
-      uses: actions/setup-node@v2-beta
+    - uses: actions/setup-node@v2-beta
       with:
-        node-version: ${{ matrix.node-version }}
+        node-version: 10.x
+    - run: npm install
+    - run: npm test
+      env:
+        FORCE_COLOR: 3
+
+  active_lts:
+    name: "Node.js active LTS"
+    runs-on: ubuntu-latest
+    steps:
+    - uses: actions/checkout@v2
+    - uses: actions/setup-node@v2-beta
+      with:
+        node-version: 14.x
     - run: npm install
     - run: npm test
       env:
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -10,7 +10,7 @@ on:
     branches: [ ] # Test on push to every branch
 
 jobs:
-  build:
+  test:
 
     runs-on: ubuntu-latest
 
@@ -28,3 +28,17 @@ jobs:
     - run: npm test
       env:
         FORCE_COLOR: 3
+
+  # This succeeds if all the matrix builds succeed. This allows us to set branch
+  # protection rules on this job in particular, rather than having to change
+  # branch protection rules when the matrix changes. See
+  # https://github.community/t/status-check-for-a-matrix-jobs/127354/7.
+  testall:
+    if: ${{ always() }}
+    runs-on: ubuntu-latest
+    name: Test (matrix)
+    needs: test
+    steps:
+      - name: Check test matrix status
+        if: ${{ needs.test.result != 'success' }}
+        run: exit 1
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -10,27 +10,21 @@ on:
     branches: [ ] # Test on push to every branch
 
 jobs:
-  maintenance_lts:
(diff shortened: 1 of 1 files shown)


.github/workflows/node.js.yml
```diff
@@ -28,3 +28,17 @@ jobs:
...
+
+  # This succeeds if all the matrix builds succeed. This allows us to set branch
+  # protection rules on this job in particular, rather than having to change
+  # branch protection rules when the matrix changes. See
+  # https://github.community/t/status-check-for-a-matrix-jobs/127354/7.
+  testall:
+    if: ${{ always() }}
+    runs-on: ubuntu-latest
+    name: Test (matrix)
+    needs: test
```
foolip, line 40: Given this job depenency I don't think the step checking for success should be necessary. If the matrix job fails, this job should just never run. Not sure if specifying no steps at all is a problem though?

foolip (approved):
This will work, but note that it will add at least a few seconds to every PR, as the no-op job still needs to be started on a runner, which might take time to provision when load is high.

An alternative approach would be to avoid version names in job names, with names like "Node.js LTS" and similar.

ddbeck:
> An alternative approach would be to avoid version names in job names, with names like "Node.js LTS" and similar.

OK, I really overthought the problem here. You're right that really what I need to do here is drop matrix builds, have two jobs that name what _kind_ of Node releases we're testing against, and skip this weird summary test. That will make the configuration more explicit about intent and simplify branch protection rules selection. I'll work on an update to this PR to make that happen.

For a bit of background here, though: I was trying to adapt our existing workflow but the matrix builds set job names automatically (e.g., "build (10.x)") and, as far as I can tell, there's no way to override the generated name of a matrix build. Inexplicably, I did not think to not use matrix builds.

Changes forthcoming.

.github/workflows/node.js.yml
ddbeck: Based on your other comments, I'm going to eliminate the need for this job, but, for the record:

The idea here is to ensure that 1) the summary runs after the completion of the matrix builds and 2) it always runs, whether the matrix builds passed or failed. This means we get a definitive pass or fail.

Ordinarily, if a dependency fails, then GitHub Actions skips the dependent job, putting it into that gray indeterminate state. But the gray state doesn't count as a failure and won't block a merge if selected as part of the branch protection rules.

ddbeck:
OK, I've broken up the matrix into distinct jobs. I also wrote a new title and description for this PR, to reflect the change better.

.github/workflows/node.js.yml
```diff
@@ -10,20 +10,27 @@ on:
     branches: [ ] # Test on push to every branch
 
 jobs:
-  build:
-
+  maintenance_lts:
+    name: "Maintenance LTS"
```
foolip, line 14: Looking at https://nodejs.org/en/about/releases/ I see there often be two versions in maintenance, and short windows of two active. But I guess we'll test the oldest and newest, and I think this naming is better than trying to be more pedantic about it.

foolip (approved):

foolip:
> as far as I can tell, there's no way to override the generated name of a matrix build

There actually is, but it's a bit roundabout, see example in https://github.com/foolip/mdn-bcd-collector/blob/main/.github/workflows/pull_request.yml. I think just not using matrix builds is fine though.

geien (approved):

Elchi3 (approved):
I've read the discussion and it reads all plausible to me. Please merge at your discretion.

ddbeck:
I'm sorry, y'all. I couldn't well enough alone. I pushed two changes:

- I used the matrix strategy that @foolip mentioned (it's not obvious you can use matrix builds this way!) since I think this is more obviously one set of steps for two different environments. This should prevent them from mistakenly going out of sync.
- I put a code comment on the "Maintenance LTS" details indicating that we should use the oldest release

I'll resist the urge to retitle this PR _again_ until after merging, so I don't break any email threads.

foolip (approved):

Elchi3 (approved):
Cool! I think Philip and you know a lot more about this than I do but sounds plausible to me. Please merge at your discretion.

PR #8783, opened 2021-01-18
Add worker_support info for CacheStorage

hamishwillee:
[CacheStorage](https://developer.mozilla.org/en-US/docs/Web/API/CacheStorage) is defined in the service worker spec but also works in the main thread and other worker types. This PR updates CacheStorage information to have a  `worker_support` key.

What I did was duplicate the top level support for `CacheStorage` as the key `worker_support`. My assumption being that worker support went in for the first version unless otherwise documented (seems reasonable because this is in the service worker spec)

image

There is a note indicating that for Chrome it became accessible through WorkerWindowScope (and window) in version 43. My assumption is therefore that this is a case where the class was available, but inaccessible. So I have updated the worker support information to 43 for that case. 

There is also a note for FF on all methods: *"notes": "<a href='https://www.mozilla.org/en-US/firefox/organizations/'>Extended Support Releases (ESR)</a> before Firefox 78 ESR do not support service workers and the Push API."*. I copied this to just mention service workers.

@ddbeck 
- Does this make sense?
- The note about ESR versions appears on all methods. Is it necessary - ie propose to remove it except for the parent CacheStorege.

### DIFF
--- a/api/CacheStorage.json
+++ b/api/CacheStorage.json
@@ -6,17 +6,11 @@
         "support": {
           "chrome": {
             "version_added": "40",
-            "notes": [
-              "Accessible from <code>Window.caches</code> from version 43.",
-              "Accessible from <code>WorkerGlobalScope.caches</code> from version 43."
-            ]
+            "notes": "Version 40 to 42: supported in service workers only. Version 43 enabled use in main thread and other worker types."
           },
           "chrome_android": {
             "version_added": "40",
-            "notes": [
-              "Accessible from <code>Window</code> from version 43.",
-              "Accessible from <code>WorkerGlobalScope</code> from version 43."
-            ]
+            "notes": "Version 40 to 42: supported in service workers only. Version 43 enabled use in main thread and other worker types."
           },
           "edge": {
             "version_added": "≤18"
@@ -404,10 +398,12 @@
           "description": "Available in workers",
           "support": {
             "chrome": {
-              "version_added": "43"
+              "version_added": "40",
+              "notes": "Version 40 to 42: supported in service workers only. Version 43 enabled use in main thread and other worker types."
             },
             "chrome_android": {
-              "version_added": "43"
+              "version_added": "40",
+              "notes": "Version 40 to 42: supported in service workers only. Version 43 enabled use in main thread and other worker types."
             },
             "edge": {
               "version_added": "≤18"
--- a/api/CacheStorage.json
+++ b/api/CacheStorage.json
@@ -6,11 +6,11 @@
         "support": {
           "chrome": {
             "version_added": "40",
-            "notes": "Version 40 to 42: supported in service workers only. Version 43 enabled use in main thread and other worker types."
+            "notes": "Before version 43, only service workers are supported. From version 43, all worker types and the main thread are supported."
           },
           "chrome_android": {
             "version_added": "40",
-            "notes": "Version 40 to 42: supported in service workers only. Version 43 enabled use in main thread and other worker types."
+            "notes": "Before version 43, only service workers are supported. From version 43, all worker types and the main thread are supported"
           },
           "edge": {
             "version_added": "≤18"
@@ -26,10 +26,12 @@
             "version_added": false
           },
           "opera": {
-            "version_added": "27"
+            "version_added": "27",
+            "notes": "Before version 30, only service workers are supported. From version 30, all worker types and the main thread are supported"
           },
           "opera_android": {
-            "version_added": "27"
+            "version_added": "27",
+            "notes": "Before version 30, only service workers are supported. From version 30, all worker types and the main thread are supported"
           },
           "safari": {
             "version_added": "11.1"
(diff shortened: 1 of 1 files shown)


api/CacheStorage.json
```diff
@@ -7,8 +7,8 @@
           "chrome": {
             "version_added": "40",
             "notes": [
-              "Accessible from <code>Window</code> from version 43.",
-              "Accessible from <code>WorkerGlobalScope</code> from version 43."
+              "Accessible from <code>Window.caches</code> from version 43.",
```
hamishwillee, line 10: Note, I decided not to remove this from the top level version info - because otherwise the assumption might (on quick scan) that this can be used from version 40).  Should we remove from the methods though - e.g. `CacheStorage.match()`?

hamishwillee:
@ddbeck Further, looking at [Cache](https://developer.mozilla.org/en-US/docs/Web/API/Cache) (which is used with `CacheStorage`) the BCD says support from chrome 43 but From 40 to 42, this was only available on service workers.

That's a bit confusing. It kind of indicates to me that perhaps the note for CacheStorage is incorrect "Accessible from <code>WorkerGlobalScope</code> from version 43." - and this should be from version 40 (with other corresponding changes).

Thoughts?

ddbeck:
My first (possibly wrong!) impressions:

- We ought to remove the `Accessible from` notes outright. These notes basically say which `api.{SomeInterface}.caches` features exist. We have, for example, an `api.WindowOrWorkerGlobalScope.caches` feature that ought to cover these details (API moves might be appropriate on _that_ feature, but not here).
- I'm not quite sure what "the class was available, but inaccessible" means. Does this mean that `CacheStorage` is only sometimes in the global scope and that's what these values reflect?

hamishwillee:
@ddbeck Thanks very much. So I have updated in line with your comments and my interpretation. Essentially that CacheStorage was supported in Chrome in service workers from 40, and other workers/main thread from version 43. 

This is heavily implied by the `Cache` docs which explicitly state that.

hamishwillee:
PS For Cache the BCD currently says we have `version-added` 43 and note "From 40 to 42, this was only available on service workers.". Our previous discussion was that the version should reflect when the first worker was added and the note reflect any variation. 

So propose for `Cache`:
- change version-added to 40 on the top level docs and leave the note.
- Then add `worker_support` that duplicates this pattern

Make sense?

api/CacheStorage.json
```diff
@@ -398,6 +392,57 @@
...
           }
         }
+      },
+      "worker_support": {
+        "__compat": {
+          "description": "Available in workers",
+          "support": {
+            "chrome": {
+              "version_added": "40",
+              "notes": "Version 40 to 42: supported in service workers only. Version 43 enabled use in main thread and other worker types."
```
ddbeck, line 402: I like how much clearer these notes and data are. A suggestion here to reframe this as consequential only for very old browsers:

```suggestion
              "notes": "Before version 43, only service workers are supported. From version 43, all worker types and the main thread are supported."
```

api/CacheStorage.json
```diff
@@ -398,6 +392,57 @@
...
+            },
+            "chrome_android": {
+              "version_added": "40",
+              "notes": "Version 40 to 42: supported in service workers only. Version 43 enabled use in main thread and other worker types."
```
ddbeck, line 406:
```suggestion
              "notes": "Before version 43, only service workers are supported. From version 43, all worker types and the main thread are supported."
```

api/CacheStorage.json
```diff
@@ -6,17 +6,11 @@
         "support": {
           "chrome": {
             "version_added": "40",
-            "notes": [
-              "Accessible from <code>Window</code> from version 43.",
-              "Accessible from <code>WorkerGlobalScope</code> from version 43."
-            ]
+            "notes": "Version 40 to 42: supported in service workers only. Version 43 enabled use in main thread and other worker types."
```
ddbeck, line 9: Since this happened five years ago and isn't getting any younger, let's put this in terms that shows it's historical and let `worker_support` do most of the, um, work. 😃 

```suggestion
            "notes": "Before version 43, this is supported in service workers only."
```

I've put a suggestion here for Chrome and Chrome for Android, but could we also:

- `opera_android`: add the same note, except with version 30
- `opera_android`: add the same note, except with version 30
- `samsunginternet_android`: remove all the notes (not applicable)
- `webview_android`: Apply the same change

api/CacheStorage.json
```diff
@@ -6,17 +6,11 @@
...
           },
           "chrome_android": {
             "version_added": "40",
-            "notes": [
-              "Accessible from <code>Window</code> from version 43.",
-              "Accessible from <code>WorkerGlobalScope</code> from version 43."
-            ]
+            "notes": "Version 40 to 42: supported in service workers only. Version 43 enabled use in main thread and other worker types."
```
ddbeck, line 13:
```suggestion
            "notes": "Before version 43, this is supported in service workers only."
```

ddbeck (changes requested):
Overall, I'm really happy with this. I think it's telling a much clearer story! And the proposal for Cache makes sense. I think we just need one more pass on this. Suggestions in line comments.

api/CacheStorage.json
hamishwillee: Thanks @ddbeck . I committed this. Slightly tempted by putting the current behaviour first because as you imply, the old behaviour isn't so interesting. So this would/could be. 

> From version 43, all worker types and the main thread are supported. Before version 43, only service workers were supported.

hamishwillee:
Hi @ddbeck 
Thanks, works for me. All your suggestions addressed (I think). I'll look at Cache once this is accepted.

ddbeck (approved):
Thank you, @hamishwillee! 🎉

hamishwillee:
A win! Thanks @ddbeck

PR #8784, opened 2021-01-18
Add worker_support to PushManager

hamishwillee:
Effectively this just duplicates the main support for PushManager in a worker_support key.

Note that I did test that PushManager is supported in ALL worker types on Chrome and Opera (i.e. you can have a push manager in a dedicated worker or service worker or shared worker - using https://worker-playground.glitch.me/

### DIFF
--- a/api/PushManager.json
+++ b/api/PushManager.json
@@ -607,7 +607,7 @@
             }
           },
           "status": {
-            "experimental": true,
+            "experimental": false,
             "standard_track": true,
             "deprecated": false
           }
--- a/api/PushManager.json
+++ b/api/PushManager.json
@@ -565,6 +565,9 @@
             },
             "edge": [
               {
+                "version_added": "17"
+              },
+              {
                 "version_added": "16",
                 "flags": [
                   {
@@ -572,9 +575,6 @@
                     "name": "Enable service workers"
                   }
                 ]
-              },
-              {
-                "version_added": "17"
               }
             ],
             "firefox": {
--- a/api/PushManager.json
+++ b/api/PushManager.json
@@ -552,6 +552,66 @@
             "deprecated": true
           }
         }
+      },
+      "worker_support": {
+        "__compat": {
+          "description": "Available in workers",
+          "support": {
+            "chrome": {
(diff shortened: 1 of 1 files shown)


api/PushManager.json
```diff
@@ -552,6 +552,66 @@
...
+                "version_added": "16",
+                "flags": [
+                  {
+                    "type": "preference",
+                    "name": "Enable service workers"
+                  }
+                ]
+              },
+              {
+                "version_added": "17"
```
ddbeck, line 577: The more general support statement needs to go first (first version_added: 17, then the one behind a flag). I realize this was inherited from above, but I'd rather not duplicate the problem here.

api/PushManager.json
```diff
@@ -552,6 +552,66 @@
...
+            },
+            "samsunginternet_android": {
+              "version_added": "4.0"
+            },
+            "webview_android": {
+              "version_added": false
+            }
+          },
+          "status": {
+            "experimental": true,
```
ddbeck, line 610: This has two implementations, so we can use non-experimental here

```suggestion
            "experimental": false,
```

ddbeck (changes requested):
A couple minor suggestions here, @hamishwillee. Thank you!

api/PushManager.json
hamishwillee: Fixed. So is the general rule that the most recent version in a set appears first? If so can the tooling notify of that?

hamishwillee:
Thanks. Changes made.

api/PushManager.json
ddbeck: That's the general rule, but there's a bunch of subtleties that make automation a challenge. The real ranking is rather more complex. It's something like: most-recent available unflagged and unprefixed, then available prefixed, then available flagged, then anything removed, with ties going to things without contemporaneously applicable notes or partial flags. 😅 

It's probably better to think of the order as an authorial decision. What's most widely applicable and useful to web developers right now? That's probably what should go first.

ddbeck (approved):
Thank you! 🎉

PR #8785, opened 2021-01-18
Clear-Site-Data requires secure context

hamishwillee:
Clear-Site-Data header requires secure context, but this is not documented - see this issue: https://github.com/mdn/content/issues/991
- The requirement for secure context is not in the draft but is visible in Firefox code https://searchfox.org/mozilla-central/source/toolkit/components/clearsitedata/ClearSiteData.cpp#166
- It is also noted in the [W3C test code](https://github.com/w3c/webappsec-clear-site-data/tree/master/demo) (which was submitted/edited by the spec author) from this note in the readme: **(NOTE: Clear-Site-Data only works on localhost and https.**

This essentially just adds the `secure_context_required` keyword. I have added `false` for browsers that don't support the header and `true` for those that do. 

@ddbeck 
- Do we *have* to try get versions on this? 
- How much further testing is reasonable? I guess I could try set up a remote server and verify that this works locally but not via HTTP remotely, and perhaps bisect some versions. But I'd rather not if it can be avoided :-)

### DIFF
--- a/http/headers/clear-site-data.json
+++ b/http/headers/clear-site-data.json
@@ -300,28 +300,28 @@
             "description": "Secure context required",
             "support": {
               "chrome": {
-                "version_added": true
+                "version_added": "61"
               },
               "chrome_android": {
-                "version_added": true
+                "version_added": "61"
               },
               "edge": {
-                "version_added": true
+                "version_added": "≤79"
               },
               "firefox": {
-                "version_added": true
+                "version_added": "62"
               },
               "firefox_android": {
-                "version_added": true
+                "version_added": "62"
               },
               "ie": {
                 "version_added": false
               },
               "opera": {
-                "version_added": true
+                "version_added": "48"
               },
               "opera_android": {
-                "version_added": true
+                "version_added": "45"
               },
               "safari": {
                 "version_added": false
@@ -330,10 +330,10 @@
                 "version_added": false
               },
               "samsunginternet_android": {
-                "version_added": true
+                "version_added": "8.0"
               },
               "webview_android": {
-                "version_added": true
+                "version_added": "61"
               }
             },
             "status": {
--- a/http/headers/clear-site-data.json
+++ b/http/headers/clear-site-data.json
@@ -338,7 +338,7 @@
             },
             "status": {
               "experimental": false,
-              "standard_track": true,
+              "standard_track": false,
               "deprecated": false
             }
           }
--- a/http/headers/clear-site-data.json
+++ b/http/headers/clear-site-data.json
@@ -295,6 +295,54 @@
             }
           }
         },
+        "secure_context_required": {
+          "__compat": {
+            "description": "Secure context required",
+            "support": {
+              "chrome": {
+                "version_added": true
+              },
+              "chrome_android": {
+                "version_added": true
+              },
+              "edge": {
+                "version_added": true
+              },
+              "firefox": {
+                "version_added": true
+              },
+              "firefox_android": {
+                "version_added": true
+              },
+              "ie": {
+                "version_added": false
+              },
+              "opera": {
+                "version_added": true
+              },
+              "opera_android": {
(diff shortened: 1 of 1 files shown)


hamishwillee:
@ddbeck Ping :-)

ddbeck:
@hamishwillee, answers incoming:

> Do we _have_ to try get versions on this?

The general answer to this question is: if a feature is not already in a category that we require (via the linter) real version data for (e.g., `css.*`), then we should make a reasonable effort to get version data, from testing, trawling bugs, etc. If something's easy to test for (e.g., global API exposure), then we should, but bugs, multiple reports from web developers, etc. should also satisfy. So consider this a license to _not_ test in this case.

That said, I was able to learn a little bit about this for Chrome and Firefox from these bugs:

https://bugs.chromium.org/p/chromium/issues/detail?id=607897
https://bugzilla.mozilla.org/show_bug.cgi?id=1268889

Based on the discussion in each, I think it's safe to assume that both engines required a secure context from initial implementation, even though it's not in the spec. You should be able to pull down the support statements from the parent feature to fill in the version numbers. 🚀

http/headers/clear-site-data.json
```diff
@@ -295,6 +295,54 @@
...
+              "samsunginternet_android": {
+                "version_added": true
+              },
+              "webview_android": {
+                "version_added": true
+              }
+            },
+            "status": {
+              "experimental": true,
+              "standard_track": true,
```
ddbeck, line 341: If this isn't in the spec, then we should set this to `false`, even though it's a de facto standard.

And maybe we ought to open an issue against the spec? Though I'm not sure what it would say, exactly.

```suggestion
              "standard_track": false,
```

http/headers/clear-site-data.json
```diff
@@ -295,6 +295,54 @@
...
+              },
+              "samsunginternet_android": {
+                "version_added": true
+              },
+              "webview_android": {
+                "version_added": true
+              }
+            },
+            "status": {
+              "experimental": true,
```
ddbeck, line 340:
```suggestion
              "experimental": false,
```

hamishwillee:
@ddbeck Thanks very much for the guidance. Have updated as you suggested. 

Should I be adding spec_info? If not, should be good to merge.

ddbeck (approved):
Thank you! 🎉

hamishwillee:
@ddbeck Just looked at secure_context_required on clear-site-data live and noted the icon that this is "non-standard. 

Technically that is correct, but it is implemented almost everywhere which makes the note associated with this incorrect "Non-standard. Expect poor cross-browser support.". I think it was kind of "implied". Can/should I do anything about this?

image

ddbeck:
Three thoughts on this, @hamishwillee:

1. Unspecified behavior, even if it's broadly supported, is still non-standard. We shouldn't paper over omissions by spec authors. Ordinarily, I'd suggest opening an issue against the spec, but…
3. I took a closer look at the spec. It's a bit indirect, but the spec does imply in at least two places ([1](https://w3c.github.io/webappsec-clear-site-data/#clear-response), [2](https://w3c.github.io/webappsec-clear-site-data/#incomplete)) that the header shouldn't do anything in potentially insecure contexts. I'd welcome a follow-up PR that flips the `standard_track` value here.
2. That "Expect poor cross-browser support" should probably be dropped from the table. That's probably correct in 95% of cases, but not all. The actual support data should tell readers whether they should expect cross-browser support. If you want to fix this, then I suggest opening a PR against Yari (see [`legend.tsx`](https://github.com/mdn/yari/blob/251285aff1bc930e38fe58fdfb32ee98912c6e75/client/src/document/ingredients/browser-compatibility-table/legend.tsx#L11).

hamishwillee:
Thanks very much. FYI opened #9408 for the standard_track change and suggestion on yari here https://github.com/mdn/yari/pull/3176

PR #8796, opened 2021-01-19
Update Safari support info for static class variables

### LINKED ISSUE #8468: javascript.classes.static - Safari on iOS 14 is not compatible

The page says iOS Safari 9 supports static variables. This does not seem to be the case for iOS Safari 14.

Running iOS 14.2 on an iPhone 11 Pro Max.

This should either not display compatibility or it should say that it doesn’t work above a certain version number.

Tested this on my device (running iOS 14.2 on an iPhone 11 Pro Max) using the demo on the page. I get the following output:

> Error: Unexpected token '='. Expected an opening '(' before a method's parameter list.

I get the same behavior on Safari on my desktop computer, with the same error in the console, from another page that uses the same syntax.

<details>

* Query: `javascript.classes.static`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Classes/static
* Report started: 2020-12-30T19:25:28.741Z

</details>

rileyjohngibbs:
As of Safari 14, this feature is not supported on desktop or mobile. This PR updates the `classes.json` file to reflect that `static` is not supported by Safari for either platform.

Closes #8468.

### DIFF
--- a/javascript/classes.json
+++ b/javascript/classes.json
@@ -734,10 +734,10 @@
               }
             ],
             "safari": {
-              "version_added": "9"
+              "version_added": false
             },
             "safari_ios": {
-              "version_added": "9"
+              "version_added": false
             },
             "samsunginternet_android": [
               {


ddbeck (approved):
Looks great. Thank you, @rileyjohngibbs! 🎉

PR #8803, opened 2021-01-20
Add Chrome support for overflow-clip-margin and overflow: clip

rachelandrew:
Chrome is shipping `overflow:clip` and the new `overflow-clip-margin` property in 89. Adding data for these (I'm going to document `overflow-clip-margin`). Also removing experimental flag for `overflow: clip` as Firefox already implements it so that is two engines. 

https://www.chromestatus.com/feature/5638444178997248

### DIFF
--- a/css/properties/overflow-clip-margin.json
+++ b/css/properties/overflow-clip-margin.json
@@ -44,7 +44,7 @@
             }
           },
           "status": {
-            "experimental": false,
+            "experimental": true,
             "standard_track": true,
             "deprecated": false
           }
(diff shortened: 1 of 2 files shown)


css/properties/overflow-clip-margin.json
```diff
@@ -0,0 +1,55 @@
...
+            },
+            "samsunginternet_android": {
+              "version_added": false
+            },
+            "webview_android": {
+              "version_added": "89"
+            }
+          },
+          "status": {
+            "experimental": false,
```
ddbeck, line 47:
```suggestion
            "experimental": true,
```

ddbeck (changes requested):
One little fix to make before merging. Thank you, @rachelandrew!

rachelandrew:
@ddbeck done :)

ddbeck (approved):
Thank you! 🎉

PR #8807, opened 2021-01-20
Add top-level await

Rumyra:
- [x] Summarize your changes

Updating info for top level await

https://bugs.chromium.org/p/v8/issues/detail?id=9344

This works for me in Chrome 87 with #enable-javascript-harmony flag enabled

Passed

https://github.com/mdn/content/issues/297


ddbeck (changes requested):
Hi @Rumyra, while reviewing the linked bug I saw that this is [shipping enabled-by-default in 89](https://bugs.chromium.org/p/v8/issues/detail?id=9344#c50) (on all platforms). Can we include that as well? Here are some notes that might help:

- Chrome: new support statement for Chrome 89
- Chrome Android: copy all from Chrome
- Opera: copy flagged support statement from Chrome, but use version 73 instead (too early to show unflagged data for Opera)
- Opera Android: leave as `false` (doesn't support flags, too early for unflagged)
- WebView: copy unflagged version 89 support statement only (doesn't support flags)

Rumyra:
Thanks @ddbeck a quick question: Chrome new support statement - do we keep the existing flagged statement as well?

jpmedley:
If allowed by the BCD rules it should be removed. After a feature arrives
in stable, Chrome advises against using anything other for developer
experimentation, and against asking users to flip flags. Any bugs found in
the flagged version will be fixed in stable only (assuming they weren't
found before the feature went to stable).
Joe Medley | Technical Writer, Chrome DevRel | jmedley@google.com |
 816-678-7195
*If an API's not documented it doesn't exist.*

Rumyra:
Thanks @jpmedley 👍

ddbeck:
> Thanks @ddbeck a quick question: Chrome new support statement - do we keep the existing flagged statement as well?

> If allowed by the BCD rules it should be removed.

What @jpmedley suggests is permissible in this case since we never added the flag data to begin with, though it's a rather exceptional case. We never got consensus about what to do with so-called "dead flags" (see #3318), which means that, for flags which are already checked in, you can't remove flags except under narrow circumstances (see #6018).

javascript/operators/await.json
```diff
@@ -70,11 +70,11 @@
             "spec_url": "https://tc39.es/proposal-top-level-await/#introduction",
             "support": {
               "chrome": {
-                "version_added": false,
+                "version_added": "89",
                 "notes": "See <a href='https://crbug.com/v8/9344'>bug 9344</a>"
```
ddbeck, line 74: If we're not going to include the flag data, then we ought to remove the note too.

```suggestion
```

ddbeck (changes requested):
One last suggestion ahead of merging. Thank you, @Rumyra!

Rumyra:
Thank so much @ddbeck that's really helpful 👍

ddbeck (approved):
Thank you! 🎉

PR #8817, opened 2021-01-21
Add a couple more A-C HTML element spec URLs

sideshowbarker:
This change adds spec URLs in the `html/elements/a.json` and `html/elements/area.json` data, for the `noopener` and `noreferrer` attributes, which have their own MDN articles.

This is a follow-up to https://github.com/mdn/browser-compat-data/pull/8810 (c84b9e1), which added the spec URLs for the HTML elements themselves whose names are in the A-C range.


Elchi3 (approved):
Great catch, thanks!

PR #8844, opened 2021-01-24
Fix page_action.show_matches support for Android

Rob--W:
When the feature initially landed, there was no support for mobile: https://bugzilla.mozilla.org/show_bug.cgi?id=1419842

`page_action.show_matches` support was added when the implementation was re-implemented from scratch based on the desktop implementation at https://bugzilla.mozilla.org/show_bug.cgi?id=1530402


ddbeck (approved):
Looks good. Thank you! 🎉

PR #8850, opened 2021-01-25
Add FF android webshare support

### LINKED ISSUE #8833: api.Navigator.share - Issue with "Navigator.share()": …
MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/share

"No Support" on Mobile Firefox

"Partial/Full Support" (not native Android Share, but own Implementation) 

https://mdn.github.io/dom-examples/web-share/

<details>

* Folder: `en-us/web/api/navigator/share`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/share
* GitHub URL: https://github.com/mdn/content/blob/master/files/en-us/web/api/navigator/share/index.html
* Document last modified: 2021-01-12T16:34:08.000Z

</details>

hamishwillee:
Fixes #8833

#8833 points out that the web share API works on FF for Android but BCD indicates not supported. 

What I have done is:
- Verified that Navigator.share does indeed work on current FF for android using https://mdn.github.io/dom-examples/web-share/
- Verified that Navigator.canShare works using the [webshare tests](http://www1.www.wpt.live:8000/web-share/canShare-files.tentative.https.html)
- Verified custom gecko integration added in [Bugzilla 1402369](https://bugzilla.mozilla.org/show_bug.cgi?id=1402369) - you can see [code here](https://searchfox.org/mozilla-central/source/mobile/android/geckoview/src/main/java/org/mozilla/geckoview/GeckoSession.java#5062)

Bugzilla indicates this going in in FF72, but there is no FF72 for android so I updated BCD with the first version after that - FF79. 

Note, https://github.com/mdn/browser-compat-data/issues/4983 was created to track these sorts of items.

### DIFF
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -2506,8 +2506,7 @@
               "version_added": false
             },
             "firefox_android": {
-              "version_added": "79",
-              "notes": "Firefox Mobile displays custom share dialog (instead of the platform-native dialog indicated by the specification)."
+              "version_added": "79"
             },
             "ie": {
               "version_added": false
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -445,7 +445,7 @@
               "version_added": false
             },
             "firefox_android": {
-              "version_added": false
+              "version_added": "79"
             },
             "ie": {
               "version_added": false
@@ -2506,7 +2506,8 @@
               "version_added": false
             },
             "firefox_android": {
-              "version_added": false
+              "version_added": "79",
+              "notes": "Firefox Mobile displays custom share dialog (instead of the platform-native dialog indicated by the specification)."
             },
             "ie": {
               "version_added": false
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -445,7 +445,7 @@
               "version_added": false
             },
             "firefox_android": {
-              "version_added": "79"
+              "version_added": false
             },
             "ie": {
               "version_added": false
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -2506,7 +2506,9 @@
               "version_added": false
             },
             "firefox_android": {
-              "version_added": "79"
+              "version_added": "79",
+              "partial_implementation": true,
+              "notes": "Firefox does not support sharing files."
             },
             "ie": {
               "version_added": false


api/Navigator.json
```diff
@@ -2506,7 +2506,8 @@
               "version_added": false
             },
             "firefox_android": {
-              "version_added": false
+              "version_added": "79",
+              "notes": "Firefox Mobile displays custom share dialog (instead of the platform-native dialog indicated by the specification)."
```
saschanaz, line 2510: AFAIK it's using Android feature to customize the UI, which should be still considered as "native". Chrome also shows customized UI and also several non-browser apps e.g. OneDrive.

A quick search shows how. https://stackoverflow.com/q/32478266

saschanaz: I don't see a spec requirement to be "native" BTW. The spec says the user agent "can" use the OS share system, not "must". https://w3c.github.io/web-share/#examples-of-share-targets

hamishwillee: Fair enough. In the abstract it even says "The available share targets are not specified here; they are provided by the user agent.",

I will remove that.

hamishwillee:
Thanks @saschanaz . Fixed as you suggested. Change now relatively simple.

saschanaz:
Cool! Could you mark it as partial implementation as Firefox doesn't support `files` yet? https://bugzilla.mozilla.org/show_bug.cgi?id=1635700

hamishwillee:
@saschanaz Thanks. I added this to the `share` value. Not to `canShare` as my assumption is that will return false to anything that it does not know about. 

Just FMI if I had added supported for web share and THEN files support was added to the spec, would it be expected we would go through and mark all the APIs as "partial"? Seems a little odd.

saschanaz:
> 
> 
> @saschanaz Thanks. I added this to the `share` value. Not to `canShare` as my assumption is that will return false to anything that it does not know about.
> 
> Just FMI if I had added supported for web share and THEN files support was added to the spec, would it be expected we would go through and mark all the APIs as "partial"? Seems a little odd.

Fair enough, maybe we should add a subitem instead, something like `file_support`.

api/Navigator.json
```diff
@@ -445,7 +445,7 @@
               "version_added": false
             },
             "firefox_android": {
-              "version_added": false
+              "version_added": "79"
```
saschanaz, line 448: Did you confirm it? Because we haven't added support for `navigator.canShare`.

hamishwillee: Hmmm. I ran http://wpt.live/web-share/canShare-files.tentative.https.html test on my phone with latest FF and it showed success with a single file. Given that we know files are not supported, this is perhaps a false positive.

I will revert that.

hamishwillee: Reverted.

saschanaz: That test is a little confusing. It tests whether passing a file without wrapping it with an array fails or not, and it does fail because it's not supported at all anyway. 😄

hamishwillee: So much for being "clever" in my due diligence. In any case, this is now the most trivial review for you ever!

ddbeck (approved):
Looks good. Thank you! 🎉

PR #8853, opened 2021-01-25
Add worker_support for XMLHttpRequest

hamishwillee:
Partial fix for https://github.com/mdn/content/issues/1515

XMLHttpRequest is intended to be used in workers
- Testing shows it cannot be used in serviceworkers but can in the other types (using https://worker-playground.glitch.me/)
- Docs indicated [XMLHttpRequest.responseXML](https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/responseXML) returns null in worker
- [This doc](https://developer.mozilla.org/en-US/docs/Web/API/Web_Workers_API) indicates XMLHttpRequest.channel returns null in workers (but other mentions of that item do not, and this should be verified).

What I've done is add "worker_support" key just to `XMLHttpRequest` (taking the support version from XMLHttpRequest) and adding a note for **every browser**: "notes": "Not supported in service workers.".
- I would prefer to add the note just once for all browsers but there does not appear to be a way (?) - because description is "hard coded" by toolchain to "available in workers".
- I have not added the notes about responseXML and channel because I wonder if these should got somehow as sub properties? Would appreciate advice on right way to do that?

@ddbeck FYI

### DIFF
--- a/api/XMLHttpRequest.json
+++ b/api/XMLHttpRequest.json
@@ -2097,7 +2097,7 @@
           "description": "Available in workers",
           "support": {
             "chrome": {
-              "version_added": "1",
+              "version_added": "4",
               "notes": "Not supported in service workers."
             },
             "chrome_android": {
@@ -2109,7 +2109,7 @@
               "notes": "Not supported in service workers."
             },
             "firefox": {
-              "version_added": "1",
+              "version_added": "3.5",
               "notes": "Not supported in service workers."
             },
             "firefox_android": {
@@ -2117,23 +2117,23 @@
               "notes": "Not supported in service workers."
             },
             "ie": {
-              "version_added": "7",
+              "version_added": "10",
               "notes": "Not supported in service workers."
             },
             "opera": {
-              "version_added": "8",
+              "version_added": "10.6",
               "notes": "Not supported in service workers."
             },
             "opera_android": {
-              "version_added": "10.1",
+              "version_added": "11",
               "notes": "Not supported in service workers."
             },
             "safari": {
-              "version_added": "1.2",
+              "version_added": "4",
               "notes": "Not supported in service workers."
             },
             "safari_ios": {
-              "version_added": "1",
+              "version_added": "5.1",
               "notes": "Not supported in service workers."
             },
             "samsunginternet_android": {
@@ -2141,7 +2141,7 @@
               "notes": "Not supported in service workers."
             },
             "webview_android": {
-              "version_added": "1",
+              "version_added": "4",
               "notes": "Not supported in service workers."
             }
           },
--- a/api/XMLHttpRequest.json
+++ b/api/XMLHttpRequest.json
@@ -2091,6 +2091,66 @@
             "deprecated": false
           }
         }
+      },
+      "worker_support": {
+        "__compat": {
+          "description": "Available in workers",
+          "support": {
+            "chrome": {
+              "version_added": "1",
+              "notes": "Not supported in service workers."
+            },
+            "chrome_android": {
+              "version_added": "18",
+              "notes": "Not supported in service workers."
+            },
+            "edge": {
+              "version_added": "12",
+              "notes": "Not supported in service workers."
+            },
+            "firefox": {
+              "version_added": "1",
+              "notes": "Not supported in service workers."
+            },
+            "firefox_android": {
+              "version_added": "4",
+              "notes": "Not supported in service workers."
+            },
+            "ie": {
+              "version_added": "7",
+              "notes": "Not supported in service workers."
+            },
+            "opera": {
+              "version_added": "8",
+              "notes": "Not supported in service workers."
+            },
+            "opera_android": {
+              "version_added": "10.1",
+              "notes": "Not supported in service workers."
+            },
+            "safari": {
+              "version_added": "1.2",
+              "notes": "Not supported in service workers."
+            },
+            "safari_ios": {
+              "version_added": "1",
+              "notes": "Not supported in service workers."
+            },
+            "samsunginternet_android": {
+              "version_added": "1.0",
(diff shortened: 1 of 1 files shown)


hamishwillee:
@ddbeck Ping.

ddbeck:
> * I would prefer to add the note just once for all browsers but there does not appear to be a way (?) - because description is "hard coded" by toolchain to "available in workers".

Yeah, I understand it's a bit annoying, but I don't expect that we'll be able to do anything about this any time soon. I think the way you've handled this here is fine.

> * I have not added the notes about responseXML and channel because I wonder if these should got somehow as sub properties? Would appreciate advice on right way to do that?

Hmm. It sort of depends on how these things work. I've seen conflicting details on this (e.g., whether it's exposed or whether it's always `null`). If it's always `null` then I don't think we do anything (or add a note that says to expect this). If it's not exposed in workers, then I suppose we ought to have a worker support subfeature for each of the affected features.

That said, I probably would not expend much effort on the compat story if all the browsers have the same behavior. BCD should help web developers understand the availability and interoperability of web APIs, not every quirk of their specification.

api/XMLHttpRequest.json
```diff
@@ -2091,6 +2091,66 @@
             "deprecated": false
           }
         }
+      },
+      "worker_support": {
+        "__compat": {
+          "description": "Available in workers",
+          "support": {
+            "chrome": {
+              "version_added": "1",
```
ddbeck, line 2100: I have a hard time believing any of these supported workers before, for example, supporting `Worker()`. Can you re-check all these versions against the worker APIs to make sure we're not suggesting that worker support preceded workers themselves? (e.g., I would expect `api.XMLHttpRequest.worker_support` ≥ `api.Worker`, or some other similar comparison, to hold)

ddbeck (changes requested):

hamishwillee:
Thanks very much. For the responseXML/channel being null issue I think we ignore it for now - I kind of think a subfeature is the right thing to do, but not all that helpful (as you indicated).

api/XMLHttpRequest.json
hamishwillee: Good point - we are - below are the compatibilities from the BCD for workers and shared workers.

Thingy | Chrome | Edge | Firefox | Internet Explorer | Opera | Safari | WebView Android | Chrome Android | Firefox Android | Opera Android | iOS Safari | Samsung Internet
-- |  -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | -- | --
XMLHttpRequest | 1 | 12 | 1 | 7 | 8 | 1.2 | 1 | 18 | 4 | 10.1 | 1 | 1.0
Worker | 4 | 12 | 3.5 | 10 | 10.6 | 4 | 4 | 18 | 4 | 11 | 5.1 | 1.0
SharedWorker | 4 | 79 | 29 | X | 10.6 | No support 5 — 6.1 | X | X | 33 | No support 11 — 14 | No support 5.1 — 7 | No support4.0 — 5.0

I propose that we set the version to true for all of these and modify the note from:

> "notes": "Not supported in service workers."
 
 TO
 
 > "notes": "Not supported in service workers. Supported in dedicated and shared workers when they are supported by the platform (check Worker and SharedWorker compatibility for details).

Of course this may not be 100% accurate but it does convey that you need to check the Worker/Shared worker. 

What do you think?

hamishwillee:
One more point I ran into in https://developer.mozilla.org/en-US/docs/Web/API/XMLHttpRequest/Using_XMLHttpRequest#workers - "Setting overrideMimeType does not work from a Worker. See bug 678057 for more details. Other browsers may handle this differently."

Should we add a subtype for overrideMimeType indicating false for all browsers?

api/XMLHttpRequest.json
ddbeck: I think setting `true` values and using the note here muddies things a little too much. In general, let's treat this support as "supported in at least one or more worker types". More concretely:

- Let's set all the versions to whatever the support is on `api.Worker`. This feels like a pretty safe assumption and if we're off by a version or two then it seems pretty harmless for something that happened a decade or more ago. And if we're wrong on the internet, it's likely someone will be along shortly to tell us.
- Let's stick to the original "Not supported in service workers." note. It's hard to deal with `SharedWorker` in a way that doesn't just duplicate `SharedWorker`'s compat data ("From version 1 to 4, `SharedWorker` is unsupported.") or creates some ambiguity (e.g., implying that shared workers exist but don't support this API with a something like "Before version 4, not supported in shared workers."). The edge here shouldn't be too sharp anyway: you'll get an error trying to use a non-existent `SharedWorker`, not `XMLHttpRequest`.

ddbeck:
> Should we add a subtype for overrideMimeType indicating false for all browsers?

The linked bug says it was fixed some 10 years ago. Like the other missing features of `XMLHttpRequest` in workers, it could have its own subfeature, but I'd probably find more interesting things to work on.

api/XMLHttpRequest.json
hamishwillee: Thanks @ddbeck - Makes sense to me. Done!

ddbeck (approved):
Thank you! 🎉

hamishwillee:
Thanks again for all your patience. I feel like I'm not getting much better at this.

hamishwillee:
How long does it usually take from merging to BCD being updated on MDN? Only asking because this one still isn't live after 3 days.

ddbeck:
Typically, it shows up a day or two after the subsequent release of `@mdn/browser-compat-data`, which I usually do on Thursdays. This was merged Monday, so it went into yesterday's release. I can see that the new package was merged into mdn/yari yesterday, so my guess is that the next deployment of MDN will show this.

hamishwillee:
Thanks. Live now :-)

PR #8864, opened 2021-01-26
Update Firefox versions to account for Firefox 85 release

chrisdavidmills:
Firefox 85 is launched today. I've got everything else in place, and this PR just updates the browser data for Fx and FxA to recognise v85 as the current version. Officially I think the release happens at 8am PT, so if you want, you can wait till then. But if you do want to just merge and be done with it, I doubt a few hours will make much difference really.


chrisdavidmills:
@ddbeck can I get you to look quickly at this teensy little one?

ddbeck (approved):
Thank you! 🎉

PR #8869, opened 2021-01-26
Remove Chromium 89 from String.at.

jpmedley:
The `String.at()` function did not make the cut for Chromium 89. I set the value to `false`. Even though it will probably be in 90, we won't know for sure for as six weeks.


Elchi3 (changes requested):
You probably also want to change `Array.at` and `TypedArray.at`, right?

ddbeck:
Related previous PR: #8721

jpmedley:
@Elchi3 Done.

Elchi3 (approved):
@jpmedley most excellent 👍

PR #8877, opened 2021-01-27
Rename :-webkit-autofill to :autofill

rachelandrew:
Working on https://github.com/mdn/content/issues/1725

Firefox has implemented `:autofill` adding the webkit prefixed version as an alias, as this is now part of a spec moving the property to the standard name with the webkit versions as prefixed versions.


jeanMS741 (approved):

jeanMS741 (approved):

css/selectors/autofill.json
```diff
@@ -0,0 +1,66 @@
...
+              "version_added": "18",
+              "prefix": "-webkit-"
+            },
+            "edge": {
+              "version_added": "79",
+              "prefix": "-webkit-"
+            },
+            "firefox": {
+              "version_added": "86",
+              "notes": "Firefox has also implemented -webkit-autofill as an alias."
```
ddbeck, line 23: This note should be reported as two support statements in an array, one with a prefix and one without.

css/selectors/autofill.json
```diff
@@ -0,0 +1,66 @@
...
+            },
+            "firefox_android": {
+              "version_added": "86",
+              "notes": "Firefox has also implemented -webkit-autofill as an alias."
```
ddbeck, line 27: As above, this note should be expressed as a separate support statement.

ddbeck (changes requested):
One change needed to the Firefox data here. Thank you, @rachelandrew!

ddbeck (approved):
Thank you, @rachelandrew! 🎉

PR #8881, opened 2021-01-27
Mark MediaSource's onsourceclose as not supported in Firefox

foolip:
This is due to an unfortunate typo in Gecko's Web IDL.

### DIFF
--- a/api/MediaSource.json
+++ b/api/MediaSource.json
@@ -533,24 +533,13 @@
             "edge": {
               "version_added": "17"
             },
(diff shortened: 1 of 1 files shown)


foolip:
Discovered via https://github.com/foolip/mdn-bcd-collector/issues/917, as Confluence listed Firefox's extra `onsourceclosed` attributes.

ddbeck (approved):
Oh, that's odd. Thank you!

PR #8898, opened 2021-01-29
Update windows.create BCD to indicate focused:false is currently ignored

caitmuenster:
Fixes https://github.com/mdn/content/issues/1723 by adding a note to the BCD table in https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/windows/create that `focused:false` is currently ignored.

### DIFF
--- a/webextensions/api/windows.json
+++ b/webextensions/api/windows.json
@@ -656,7 +656,8 @@
               "firefox": {
                 "notes": [
                   "'url' and 'tabId options can't both be set together.",
-                  "The returned 'Window' object contains the 'tabs' property only from version 52 onwards."
+                  "The returned 'Window' object contains the 'tabs' property only from version 52 onwards.",
+                  "Firefox currently ignores the 'focused:false' option."
                 ],
                 "version_added": "45"
               },
--- a/webextensions/api/windows.json
+++ b/webextensions/api/windows.json
@@ -657,7 +657,7 @@
                 "notes": [
                   "'url' and 'tabId options can't both be set together.",
                   "The returned 'Window' object contains the 'tabs' property only from version 52 onwards.",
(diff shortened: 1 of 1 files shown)


Rob--W (approved):
LGTM

Could have been a separate property to mark the level of support for `focused:true` by Firefox version, but I don't really mind the current approach.

caitmuenster:
Thanks Rob!

webextensions/api/windows.json
```diff
@@ -656,7 +656,8 @@
               "firefox": {
                 "notes": [
                   "'url' and 'tabId options can't both be set together.",
-                  "The returned 'Window' object contains the 'tabs' property only from version 52 onwards."
+                  "The returned 'Window' object contains the 'tabs' property only from version 52 onwards.",
+                  "Firefox currently ignores the 'focused:false' option."
```
ddbeck, line 660: Thanks for this PR, @caitmuenster. I trust the other review here on the substance of the note, but I do have a couple of style concerns.

- "Currently" is a bit of a moving target and difficult to maintain (we don't reconfirm every note for every release). Can we constrain this information at all? For example, if this is a new behavior, can it be anchored to a specific version number (e.g., "From version X, Firefox ignores the `focused: false` option."). Or if it's been true from inception, can we drop the word "currently"?
- I think it'd be nice to put the quoted text in code format (even though the other notes didn't do that)

```suggestion
                  "Firefox currently ignores the <code>focused: false</code> option."
```

ddbeck (changes requested):

caitmuenster:
Thanks @ddbeck! Just pushed a commit with your suggestions.

ddbeck (approved):
Looks good. Thank you! 🎉

Issue #8834, opened 2021-01-23
api.Window.requestIdleCallback - Safari iOS supported behind a flag


nobody5050:
#### What information was incorrect, unhelpful, or incomplete?

under compatibility mobile safari is listed as “not compatible”

#### What did you expect to see?

It should be labeled as “experimental”

#### Did you test this? If so, how?

On a mobile device I went to settings > safari > advanced > experimental features and noticed that requestidlecallback is an enable-able feature. This exists on any version above iOS 13.6

<details>
<summary>MDN page report details</summary>

* Query: `api.Window.requestIdleCallback`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/Window/requestIdleCallback
* Report started: 2021-01-23T18:19:31.123Z

</details>

ddbeck:
Thanks for reporting this issue, @nobody5050.

This does appear to be a preference available in Safari (both desktop and iOS). To fix this, we'd need to update the support data for [`requestIdleCallback` in `Window.json`](https://github.com/mdn/browser-compat-data/blob/c5ca113ab8166255ae8e2ba287e19e7768138afc/api/Window.json#L7433-L7506). Specifically, we'd need a support statement with a `flags` array (e.g., `[ { "type": "preference", "name": "requestIdleCallback" } ]`).

That leaves the version numbers themselves. Did you test with versions of Safari iOS before 13.6? Or was 13.6 the earliest you tested with, but no earlier? If it's the latter, some additional testing or research may be necessary to determine the correct `version_added` value (likewise on desktop).

If you (or another contributor) has more insight into the history of this feature, I'd welcome an update on this issue or a PR resolving it. Thank you!

queengooborg:
This preference has since been removed from Safari, so I'm going to close this issue following our irrelevant flag data guidelines.

PR #8899, opened 2021-01-29
Remove several needless "Enabled by default" notes

ddbeck:
I stumbled on (via #8834) a bunch of notes that needlessly use notes to describe their own support statements.

In the case of `requestIdleCallback`, I did dig up the preference name. In the other cases, it was self evident based on adjacent support statements.

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -7445,24 +7445,34 @@
             },
             "firefox": [
               {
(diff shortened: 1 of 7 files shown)


Petemtl (approved):

Elchi3 (approved):
Nice finding! :)

PR #8921, opened 2021-02-01
Add the ::target-text pseudo

rachelandrew:
Adding the `::target-text` pseudo. I'll be creating the rest of the docs.

Spec: https://drafts.csswg.org/css-pseudo/#selectordef-target-text
Chrome Status: https://www.chromestatus.com/feature/5689463273422848

### DIFF
--- a/css/selectors/target-text.json
+++ b/css/selectors/target-text.json
@@ -25,7 +25,7 @@
               "version_added": false
             },
             "opera": {
-              "version_added": false
+              "version_added": "75"
             },
             "opera_android": {
               "version_added": false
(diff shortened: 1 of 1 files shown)


css/selectors/target-text.json
```diff
@@ -0,0 +1,55 @@
...
+              "version_added": false
+            },
+            "firefox_android": {
+              "version_added": false
+            },
+            "ie": {
+              "version_added": false
+            },
+            "opera": {
+              "version_added": false
```
ddbeck, line 28: Likely addition?

```suggestion
              "version_added": "75"
```

ddbeck (changes requested):
One possible suggestion here, otherwise this looks good. Thank you!

ddbeck (approved):
Looks good. Thank you! 🎉

PR #8925, opened 2021-02-01
Update DocumentFragment.json

romainmenke:
Updated compat data for `DocumentFragment constructor` :

- Chrome 29 is the first version with a working constructor
- Chrome 29 corresponds to Opera 16
- Edge 17 is the first version with a working constructor
- ~iOS/Safari 7 also has a working constructor~ kept it at `8` after looking through the git history

Ran this code in all browsers from Browserstack :

```js
var fixture = document.getElementById('the-fixture');

var fragment = new DocumentFragment();
var child = document.createElement('div');
child.id = "child-added-with-js";

fragment.appendChild(child);

fixture.appendChild(fragment);

var result = document.getElementById('child-added-with-js');

// cb checks for "true"
cb(
	!!result &&
	result.parentNode === fixture
);
```

https://romainmenke.github.io/web-tests/ (search for `DocumentFragment`)

### DIFF
--- a/api/DocumentFragment.json
+++ b/api/DocumentFragment.json
@@ -53,13 +53,13 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/DocumentFragment/DocumentFragment",
           "support": {
             "chrome": {
-              "version_added": "28"
+              "version_added": "29"
             },
             "chrome_android": {
               "version_added": true
             },
             "edge": {
-              "version_added": "≤18"
+              "version_added": "17"
             },
             "firefox": {
               "version_added": "24"
@@ -71,16 +71,16 @@
               "version_added": false
             },
             "opera": {
-              "version_added": "15"
+              "version_added": "16"
             },
             "opera_android": {
               "version_added": true
             },
             "safari": {
-              "version_added": "8"
+              "version_added": "7"
             },
             "safari_ios": {
-              "version_added": "8"
+              "version_added": "7"
             },
             "samsunginternet_android": {
               "version_added": true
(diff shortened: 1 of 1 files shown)


api/DocumentFragment.json
```diff
@@ -53,13 +53,13 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/DocumentFragment/DocumentFragment",
           "support": {
             "chrome": {
-              "version_added": "28"
+              "version_added": "29"
```
foolip, line 56: I can confirm this by testing http://mdn-bcd-collector.appspot.com/tests/api/DocumentFragment/DocumentFragment on Chrome 28 and 29.

api/DocumentFragment.json
```diff
@@ -53,13 +53,13 @@
...
             },
             "chrome_android": {
               "version_added": true
```
foolip, line 59:
```suggestion
              "version_added": "29"
```

api/DocumentFragment.json
```diff
@@ -53,13 +53,13 @@
...
             },
             "edge": {
-              "version_added": "≤18"
+              "version_added": "17"
```
foolip, line 62: Also confirmed by testing Edge 16+17.

api/DocumentFragment.json
```diff
@@ -71,7 +71,7 @@
               "version_added": false
             },
             "opera": {
-              "version_added": "15"
+              "version_added": "16"
             },
             "opera_android": {
               "version_added": true
```
foolip, line 77:
```suggestion
              "version_added": "16"
```

foolip (approved):

foolip:
Thanks for this @romainmenke, and welcome to BCD! I'll go ahead and mirror (as we call it) to the other Chromium-based browsers and merge this.

PR #8926, opened 2021-02-01
Cache: Add worker support info

hamishwillee:
Updates `Cache` to use the same patterns as done in `CacheStorage`: #8783

This is an information reformatting, not "new" information. 

@ddbeck Note, this also adds a couple of full stops to the CacheStorage docs.


ddbeck (changes requested):
Looks good. Thank you! 🎉

PR #8932, opened 2021-02-02
Add data for Set.prototype.keys (alias of values)

foolip:
This is part of the spec and the MDN page redirects:
https://tc39.es/ecma262/#sec-set.prototype.keys
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/keys

All that's missing is the support in BCD. It looks as though `keys` and
`values` were introduced at the same time in all cases.

https://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8881 was
tested in the following browsers:
- Chrome 37 + 38 on Windows 7
- Edge 15 on Windows 10 (not 12, so leaves room for error)
- Firefox 23 + 24 on Windows 7
- Safari on iOS 8 (7 not tested)

Node.JS 0.12.18 was manually tested, and Set.prototype.keys is there.

The remaining changes are Chromium-based browsers and assumed to match
Chrome desktop.

### DIFF
--- a/javascript/builtins/Set.json
+++ b/javascript/builtins/Set.json
@@ -748,45 +748,117 @@
             "mdn_url": "https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Set/values",
             "spec_url": "https://tc39.es/ecma262/#sec-set.prototype.values",
             "support": {
-              "chrome": {
-                "version_added": "38"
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Tidy! Thank you, @foolip. 🎉

PR #8948, opened 2021-02-03
css.properties.list-style-image - Add a note for Firefox <image> support

rachelandrew:
Firefox 86 implements any `<image>` for `list-style-type`.

I have tested and it seems other browsers have supported this from a long way back Safari 6.2/Chrome 26/IE10 so I think just having a note for Firefox is probably sufficient. We could make it in to a subfeature, but I don't think this is a thing a lot of people run into.

### DIFF
--- a/css/properties/list-style-image.json
+++ b/css/properties/list-style-image.json
@@ -15,7 +15,8 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "1"
+              "version_added": "1",
+              "notes": "Prior to Firefox 86 this property did not accept an <image> type, and required the url of an image."
             },
             "firefox_android": {
               "version_added": "4"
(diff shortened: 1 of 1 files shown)


css/properties/list-style-image.json
```diff
@@ -15,7 +15,8 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "1"
+              "version_added": "1",
+              "notes": "Prior to Firefox 86 this property did not accept an <image> type, and required the url of an image."
```
ddbeck, line 19: some style suggestions here

```suggestion
              "notes": "Before Firefox 86, this property did not accept an <code>&lt;image&gt;</code> type, and required the URL of an image."
```

css/properties/list-style-image.json
```diff
@@ -15,7 +15,8 @@
...
             },
             "firefox_android": {
               "version_added": "4"
```
ddbeck, line 22: Is it likely that the note also applies to Android? If so we may as well copy it over

ddbeck (changes requested):
A couple of suggestions for you. Thank you, @rachelandrew!

ddbeck (approved):
Thank you! 🎉

PR #8951, opened 2021-02-03
Update Chrome/Safari versions for api.CSSStyleSheet.insertRule.optional_index

vinyldarkscratch:
Follow-up to #8461.  By running `document.styleSheets[0].insertRule('p{ color: black;}');` in Safari 1.0, I have determined that the `index` parameter was optional since the very first WebKit release.  As such, this marks Safari as "1", and corrects Chrome to "1" (the original data came from the wiki migration so I don't trust it).

### DIFF
--- a/api/CSSStyleSheet.json
+++ b/api/CSSStyleSheet.json
@@ -271,7 +271,7 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "3"
+              "version_added": "1"
             },
             "safari_ios": {
               "version_added": "1"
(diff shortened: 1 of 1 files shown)


api/CSSStyleSheet.json
```diff
@@ -294,10 +294,10 @@
             "description": "<code>index</code> is optional",
             "support": {
               "chrome": {
-                "version_added": "60"
+                "version_added": "1"
```
foolip, line 297: Yep, it's optional in Chrome. When an attempt was made to make it non-optional in https://bugs.chromium.org/p/chromium/issues/detail?id=319695, that didn't work out and was reverted. So I think to be pedantic it was non-optional in Chrome 31, but no real need to capture that here I think.

foolip (approved):

Issue #8854, opened 2021-01-25
Intended meaning/precision of WebKit engine version?


foolip:
Since Safari 9, WebKit engine version numbers in [`safari.json`](https://github.com/mdn/browser-compat-data/blob/e05c9e9ad76ff6023d5c717059d682e8b4e778f8/browsers/safari.json) and [`safari_ios.json`](https://github.com/mdn/browser-compat-data/blob/e05c9e9ad76ff6023d5c717059d682e8b4e778f8/browsers/safari_ios.json) are given with 3 version components. For example, Safari 14 is 610.1.28:
https://github.com/mdn/browser-compat-data/blob/e05c9e9ad76ff6023d5c717059d682e8b4e778f8/browsers/safari.json#L170-L176

However, current Safari 14.0.2 has "16610.3.7.1.9" in "About Safari", which I take to be WebKit 610.3.7.

Neither 610.1.28 or 610.3.7 are WebKit trunk versions, those reached only 610.1.26 and 610.2.2 before bumping to 611.

Since I've used these numbers to figure out the Safari version for a change from WebKit trunk, knowing how to interpret it is important.

So, should this be the WebKit version given in the earliest release of that Safari version, or should it be the version of WebKit trunk it was based on?

foolip:
cc @gsnedders

foolip:
One thing I'm not sure about is versions from Safari release branches. The [Safari 13.0 branch](https://trac.webkit.org/browser/webkit/releases/Apple/Safari%2013.0/WebCore/Configurations/Version.xcconfig) has version 608.1.49, while [13.0.1](https://trac.webkit.org/browser/webkit/releases/Apple/Safari%2013.0.1/WebCore/Configurations/Version.xcconfig) is 608.2.11, matching BCD. I'd assume BCD was wrong here, but 13.0.1 is the [earliest run](https://wpt.fyi/results/?run_id=334500001) we have in wpt.fyi, and I'm not sure how we'd be certain if 13.0 was the first stable release.

ddbeck:
I think a lot of the Safari and WebKit engine version numbers are best guesses. Look at a PR like #3877, for example. Before Safari froze the UA string, we had a plausible way of tracking the underlying engine by taking what Safari itself was reporting about the engine version. Now we don't have that I'm not sure what to do about it.

[...]

PR #8952, opened 2021-02-03
Add Safari version for api.HTMLImageElement.lowsrc

vinyldarkscratch:
This PR replaces the ranged value for Safari for `api.HTMLImageElement.lowsrc` with a specific version number, based upon commit history.

Safari 3: https://trac.webkit.org/browser/webkit/tags/old/Safari-521.11/WebCore/html/HTMLImageElement.idl
Safari 3.1: https://trac.webkit.org/browser/webkit/tags/old/Safari-5525.13/WebCore/html/HTMLImageElement.idl#L43

### DIFF
--- a/api/HTMLImageElement.json
+++ b/api/HTMLImageElement.json
@@ -822,10 +822,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": "≤4"
+              "version_added": "3.1"
             },
             "safari_ios": {
-              "version_added": "≤3"
+              "version_added": "2"
             },
             "samsunginternet_android": {
               "version_added": true
--- a/api/HTMLImageElement.json
+++ b/api/HTMLImageElement.json
@@ -822,7 +822,7 @@
               "version_added": true
             },
             "safari": {
-              "version_added": "3.1"
+              "version_added": "3"
             },
             "safari_ios": {
               "version_added": "2"


foolip:
After @gsnedders provided a lot of info https://github.com/mdn/browser-compat-data/issues/8854 I'm not super trusting of the tags for WebKit/Safari releases, and will probably look longer and harder at changes based on source.

In this case, looking at the Safari 3.0 release tag seems to conflict:
https://trac.webkit.org/browser/webkit/releases/Apple/Safari%203.0/WebCore/html/HTMLImageElement.idl#L38

Another way of going about it is finding the revision that introduced it:
https://trac.webkit.org/changeset/18940/webkit

Based on the date it seems more likely this was already in Safari 3, especially since it was a regression.

The way to settle this for sure would be to test real Safari 3. But if that's too hard, I'd just go with Safari 3 as the more likely correct answer.

gsnedders:
Safari 2.0.4 was WebKit-417.9, Safari 3.0.4 was WebKit-5523.10.3, FYI.

foolip:
Thanks Sam! Based on https://trac.webkit.org/browser/webkit/tags/old/Safari-5523.10.3/WebCore/html/HTMLImageElement.idl then I think we can safely say this was introduced in Safari 3.

api/HTMLImageElement.json
```diff
@@ -822,10 +822,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": "≤4"
+              "version_added": "3.1"
```
foolip, line 825:
```suggestion
              "version_added": "3"
```

foolip (approved):

PR #8953, opened 2021-02-03
Add Safari versions for HTMLLinkElement API properties

vinyldarkscratch:
This PR adds the Safari versions for the `HTMLLinkElement` API's attributes based upon commit history.  Looking at the IDL, all of the attributes that were marked as "≤4" were present since the API was introduced (Safari 3).  https://trac.webkit.org/browser/webkit/tags/old/Safari-521.11/WebCore/html/HTMLLinkElement.idl

### DIFF
--- a/api/HTMLLinkElement.json
+++ b/api/HTMLLinkElement.json
@@ -123,10 +123,10 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "≤4"
+              "version_added": "3"
             },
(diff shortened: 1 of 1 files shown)


foolip (approved):
https://trac.webkit.org/browser/webkit/releases/Apple/Safari%203.0/WebCore/html/HTMLLinkElement.idl, which I'd trust a little bit more than the other tag, and it agrees about these members.

PR #8955, opened 2021-02-03
Add Safari versions for api.SVGMaskElement.mask[Content]Units

vinyldarkscratch:
This PR adds Safari versions for the `maskUnits` and `maskContentUnits` attributes of the `SVGMaskElement` API based upon commit history.  https://trac.webkit.org/browser/webkit/tags/old/Safari-5525.13/WebCore/svg/SVGMaskElement.idl

### DIFF
--- a/api/SVGMaskElement.json
+++ b/api/SVGMaskElement.json
@@ -122,10 +122,10 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "≤4"
+              "version_added": "3.1"
             },
             "safari_ios": {
-              "version_added": "≤3"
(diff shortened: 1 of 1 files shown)


foolip (approved):
As in https://github.com/mdn/browser-compat-data/pull/8952 I'm not sure about the accuracy of using these tags. These were introduced in https://trac.webkit.org/changeset/26393/webkit and per https://trac.webkit.org/browser/webkit/trunk/WebCore/Configurations/Version.xcconfig?rev=26393 that's 523.11.0, which does map to 3.1. So this seems right!

PR #8961, opened 2021-02-04
Mark object-fit/object-position as supported in Chrome 32 (not 31)

foolip:
The data for the other Chromium-based browsers was correct, and the
unusual offset between Opera and Chrome is how this was noticed. (The
offset was always 13 until Opera 69, but was 12 here.)

Confirmed by testing in Chrome 31 and 32:
http://mdn-bcd-collector.appspot.com/tests/css/properties/object-fit
http://mdn-bcd-collector.appspot.com/tests/css/properties/object-position
https://fregante.github.io/object-fit-images/demo/

### DIFF
--- a/css/properties/object-fit.json
+++ b/css/properties/object-fit.json
@@ -6,10 +6,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/object-fit",
           "support": {
             "chrome": {
-              "version_added": "31"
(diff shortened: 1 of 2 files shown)


foolip:
Oops, the same update is needed for `object-position` too. I'll push that too.

ddbeck (approved):
Nice find. Thank you, @foolip! 🎉

PR #8962, opened 2021-02-04
Better reflect EdgeHTML's partial support for object-fit/object-position

foolip:
The URL linked in the original note is gone, but the issue can be seen
here: https://web.archive.org/web/20171208010253/https://developer.microsoft.com/en-us/microsoft-edge/platform/issues/13603873/

To confirm that this also applied to object-position, a test using
object-position for video was run on Edge 18 and 80 to confirm:
https://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8889

In Edge 18, object-position has no effect. Changing the demo to use
`object-fit: fill` also confirms that object-fit didn't work in Edge 18.

### DIFF
--- a/css/properties/object-fit.json
+++ b/css/properties/object-fit.json
@@ -11,10 +11,17 @@
             "chrome_android": {
               "version_added": "31"
             },
-            "edge": {
-              "version_added": "16",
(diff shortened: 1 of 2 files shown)


ddbeck (approved):
Thank you! 🎉

Issue #8964, opened 2021-02-04
Remove three_value_syntax subfeatures


foolip:
These subfeatures were added in https://github.com/mdn/browser-compat-data/pull/4882 and the current state can be seen here:
https://developer.mozilla.org/en-US/docs/Web/CSS/mask-position#browser_compatibility
https://developer.mozilla.org/en-US/docs/Web/CSS/object-position#browser_compatibility
https://developer.mozilla.org/en-US/docs/Web/CSS/perspective-origin#browser_compatibility
https://developer.mozilla.org/en-US/docs/Web/CSS/position_value#browser_compatibility
https://developer.mozilla.org/en-US/docs/Web/CSS/shape-outside#browser_compatibility

Support was removed from WebKit in https://bugs.webkit.org/show_bug.cgi?id=189142 and matching the version numbers that rolled out with Safari 13.1 / iOS 13.4.

I started fixing the data but there's too much. In addition to the WebKit change, there are lots of prefixed entries which don't have the `version_removed` bits, and support is still claimed for Firefox for Android. The entries should also be made non-standard and deprecated.

Since the entries won't be found by looking for features removed everywhere, I'm filing this issue to eventually remove them, two years after the Safari 13.1/13.4 release date, which will be 2022-03-24.

cc @rachelandrew

jculver050:
Close

PR #8965, opened 2021-02-04
Mark three-value syntax for object-position as removed everywhere

foolip:
This was the motivating case for https://github.com/mdn/browser-compat-data/issues/8964,
and fixing just this entry ensures that this will eventually resurface
when removing irrelevant features removed from all browsers.

WebKit removal: https://trac.webkit.org/changeset/251668/webkit

WebKit trunk was then at version 609.1.9.

### DIFF
--- a/css/properties/object-position.json
+++ b/css/properties/object-position.json
@@ -81,7 +81,8 @@
                 "version_removed": "70"
               },
               "firefox_android": {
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Thank you! 🎉

PR #8967, opened 2021-02-04
Add IE versions for Attr API

vinyldarkscratch:
This PR replaces the ranged versions for IE ≤6 with specific version numbers.  Looking at `api.Element.getAttributeNode` (the method to get an `Attr` instance), it marks support in IE as 6, so it wouldn't make sense for the interface or its members to be supported before then.  (Confirmed by running https://mdn-bcd-collector.appspot.com/tests/api/Attr on IE 5.5.)

### DIFF
--- a/api/Attr.json
+++ b/api/Attr.json
@@ -20,7 +20,7 @@
             "version_added": "4"
           },
           "ie": {
-            "version_added": "6"
+            "version_added": "5.5"
           },
           "opera": {
             "version_added": "8"
@@ -328,7 +328,7 @@
               "version_added": "4"
             },
             "ie": {
-              "version_added": "6"
+              "version_added": "5.5"
             },
             "opera": {
               "version_added": "≤12.1"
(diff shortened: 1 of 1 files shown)


foolip:
Another way to (currently) get `Attr` instances is `element.attributes`, and per https://developer.mozilla.org/en-US/docs/Web/API/Element/attributes#browser_compatibility that's supported since IE 5.5. Since https://mdn-bcd-collector.appspot.com/tests/api/Attr depends on `getAttributeNode` it doesn't provide additional evidence that this *wasn't* support in IE 5.5.

Can you check if `element.attributes` and the `Attr` interface object itself is in IE 5.5?

foolip:
The changes https://github.com/mdn/browser-compat-data/pull/8986 point in the direction of `Attr` actually being supported earlier than IE6.

vinyldarkscratch:
Using `element.attributes` did reveal that there's support in IE 5.5, for the main API and one of its subfeatures.  Data updated!

foolip (approved):

PR #8968, opened 2021-02-04
Add IE versions for CharacterData API

vinyldarkscratch:
This PR replaces the ranges of ≤6 with just "6" for the members of the CharacterData API for IE.  Since the API itself was implemented in IE 6, it's not possible for its members to be implemented earlier.  (I'll be writing a linter update to prevent this from happening.)

### DIFF
--- a/api/CharacterData.json
+++ b/api/CharacterData.json
@@ -67,7 +67,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": "≤6"
+              "version_added": "6"
             },
             "opera": {
               "version_added": "≤12.1"
@@ -115,7 +115,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": "≤6"
+              "version_added": "6"
             },
             "opera": {
               "version_added": "≤12.1"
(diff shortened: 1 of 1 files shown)


foolip:
I'd like to trust that the existing data for api.CharacterData is correct, but I don't quite. https://dom.spec.whatwg.org/#characterdata explains that "`CharacterData` is an abstract interface and does not exist as node. It is used by `Text`, `ProcessingInstruction`, and `Comment` nodes."

That makes we wonder when `Text` was supported in IE, and it seems quite implausible that it wasn't until IE 6. So, if `Text` was supported in IE 5.5, wasn't it a `CharacterData`? It would be great to check these two things in IE 5.5:
- Do either or both of `window.CharacterData` and `window.Text` exist?
- What kind of object is the first child of an element parsed from `<div>bla</div>`? Does it have the members of `CharacterData`?

It might be that the interface itself wasn't on the prototype chain originally, but that all of the members were supported on `Text` instances. That would make for a messy situation, where maybe just setting the support for api.CharacterData to ≤6 as well might be most expedient.

vinyldarkscratch:
Admittedly, this one was a more programmatic "≤6 -> 6" change, I didn't actually dive in and double-check the data.  However, considering my PR for the `Text` API, this was certainly around earlier.

I think it also makes sense to merge this PR with #9007 considering the data relies upon one another.  Merging the two PRs now!

PR #8969, opened 2021-02-04
A subfeature cannot have a range at same version as non-range parent

vinyldarkscratch:
This PR adds an additional update to the consistency check to disallow a subfeature having a range that is at the same version as a parent with a non-range.  In other words, if a parent is set to `6`, and a subfeature is set to `≤6`, the linter throws an error as the subfeature cannot imply earlier support than its parent.

Because there are such consistency issues already in our data, the linter will *fail* for the time being.  I will be writing PRs to resolve this.


foolip:
@queengooborg looks like there are lint failures, data needs updating?

PR #8970, opened 2021-02-04
[api] Remove range if parent is at same version

vinyldarkscratch:
This PR corrects the issues within the `api/` folder caught by the linter updates in #8969, where it disallows a subfeature having a range that is at the same version as a parent with a non-range (ex. if the parent is `6`, the subfeature cannot be `≤6` since it implies potential earlier support).

### DIFF
--- a/api/AudioScheduledSourceNode.json
+++ b/api/AudioScheduledSourceNode.json
@@ -115,7 +115,7 @@
               "version_added": "18"
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "79"
             },
             "firefox": {
               "version_added": "53"
--- a/api/HTMLLegendElement.json
+++ b/api/HTMLLegendElement.json
@@ -66,7 +66,7 @@
               "version_added": "4"
             },
             "ie": {
-              "version_added": "≤6"
+              "version_added": "6"
             },
             "opera": {
               "version_added": "≤12.1"
@@ -113,7 +113,7 @@
               "version_added": "4"
             },
             "ie": {
-              "version_added": "≤6"
+              "version_added": "6"
             },
             "opera": {
               "version_added": "≤12.1"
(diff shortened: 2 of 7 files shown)


foolip:
For a bunch of these cases I'm not too sure I trust the data for the parent feature. Can you trust it as untrustworthy and verify if the data we're assuming is correct here is in fact correct?

api/AudioScheduledSourceNode.json
```diff
@@ -115,7 +115,7 @@
               "version_added": "18"
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "79"
```
foolip, line 118: I'm skeptical about this one since https://developer.mozilla.org/en-US/docs/Web/API/AudioBufferSourceNode#browser_compatibility + https://developer.mozilla.org/en-US/docs/Web/API/OscillatorNode#browser_compatibility were in Edge 12.

api/HTMLLegendElement.json
```diff
@@ -66,7 +66,7 @@
               "version_added": "4"
             },
             "ie": {
-              "version_added": "≤6"
+              "version_added": "6"
```
foolip, line 69: Let's assume this is correct, I did in https://github.com/mdn/browser-compat-data/pull/8972.

api/AudioScheduledSourceNode.json
vinyldarkscratch: Surprisingly enough, `AudioScheduledSourceNode` wasn't present in Edge 15 or 18, either with the standard name or the alternate name used in other browsers at first, `AudioSourceNode`.

vinyldarkscratch:
Just to be on the safe side, I gave this one a much closer look, and it appears that everything's in order!  All of the parent's data came from recent PRs from the collector (which were aided by additional manual testing, it looks like), or other credible PRs.

foolip:
Thanks for double checking, that assuages my fears :)

foolip (approved):

PR #8971, opened 2021-02-04
[css] Remove range if parent is at same version

vinyldarkscratch:
This PR corrects the issues within the `css/` folder caught by the linter updates in #8969, where it disallows a subfeature having a range that is at the same version as a parent with a non-range (ex. if the parent is `6`, the subfeature cannot be `≤6` since it implies potential earlier support).

### DIFF
--- a/css/properties/background-repeat.json
+++ b/css/properties/background-repeat.json
@@ -140,7 +140,7 @@
                 "version_added": "2.0"
               },
               "webview_android": {
-                "version_added": "37"
+                "version_added": "≤37"
               }
             },
             "status": {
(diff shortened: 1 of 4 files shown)


ddbeck:
Did you test these, @vinyldarkscratch? Or is this a mechanical change based on the data already here?

I looked at each of these and wondered if the relationship shouldn't be reversed and the ranges passed up to the parents. The case is easiest to see in the webviews, where it really looks like cases of missing `≤` on the parent features.

That said, I haven't gone through the git blame to check the history of any of these.

vinyldarkscratch:
Man, I should have checked on these before making this PR...looks like pretty much all of these are the case of bad mirroring (and mine at that).

PR #8972, opened 2021-02-04
html.elements.legend: Remove range if parent is at same version

vinyldarkscratch:
This PR corrects the issues within the `html/` folder caught by the linter updates in #8969, where it disallows a subfeature having a range that is at the same version as a parent with a non-range (ex. if the parent is `6`, the subfeature cannot be `≤6` since it implies potential earlier support).

### DIFF
--- a/html/elements/legend.json
+++ b/html/elements/legend.json
@@ -68,7 +68,7 @@
                 "version_added": "4"
               },
               "ie": {
-                "version_added": "≤6"
+                "version_added": "6"
               },
               "opera": {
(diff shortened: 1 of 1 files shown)


foolip (approved):
https://developer.mozilla.org/en-US/docs/Web/API/HTMLLegendElement#browser_compatibility agrees, so I'll assume `<legend>` was indeed introduced in IE6.

chandonglee (commented):
What's disturbing?

PR #8973, opened 2021-02-04
Add IE/Edge versions for HTMLBaseFontElement API

vinyldarkscratch:
This PR adds real values for Internet Explorer and Edge for the `HTMLBaseFontElement` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.1.8).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/HTMLBaseFontElement

### DIFF
--- a/api/HTMLBaseFontElement.json
+++ b/api/HTMLBaseFontElement.json
@@ -22,7 +22,7 @@
             "version_added": false
           },
           "ie": {
-            "version_added": "≤6"
+            "version_added": "5"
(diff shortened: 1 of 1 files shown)


foolip (approved):
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/basefont#browser_compatibility could also be updated based on this.

PR #8976, opened 2021-02-04
Update note for Firefox support for CSS flex-direction with reverse

### LINKED ISSUE #8883: CSS flex-direction - Firefox supports overflow
According [to Bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1042151) starting from Firefox version 81 this is no longer true:

> Does not support overflow when using `*-reverse`. 

- https://github.com/mdn/browser-compat-data/blob/master/css/properties/flex-direction.json
- https://caniuse.com/mdn-css_properties_flex-direction

joneskj55:
This PR addresses issue #8883 for both Firefox and Firefox for Android.

Fixes #8883

### DIFF
--- a/css/properties/flex-direction.json
+++ b/css/properties/flex-direction.json
@@ -38,7 +38,7 @@
                 "partial_implementation": true,
                 "notes": [
                   "Since Firefox 28, multi-line flexbox is supported.",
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Looks great. Thank you! 🎉

PR #8977, opened 2021-02-04
Add IE/Edge versions for HTMLDirectoryElement API

vinyldarkscratch:
This PR adds real values for Internet Explorer and Edge for the `HTMLDirectoryElement` API, based upon manual testing.

Test Code Used: `var instance = document.createElement('dir');`


foolip (approved):
https://developer.mozilla.org/en-us/docs/Web/HTML/Element/dir#browser_compatibility could also be updated based on this.

PR #8978, opened 2021-02-04
Add IE/Edge versions for HTMLDocument API

vinyldarkscratch:
This PR adds real values for Internet Explorer and Edge for the `HTMLDocument` API, based upon manual testing.

Test Code Used: `var instance = document;`

### DIFF
--- a/api/HTMLDocument.json
+++ b/api/HTMLDocument.json
@@ -20,7 +20,7 @@
             "version_added": "4"
           },
           "ie": {
-            "version_added": "≤6"
+            "version_added": "4"
           },
           "opera": {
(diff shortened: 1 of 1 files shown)


api/HTMLDocument.json
```diff
@@ -20,7 +20,7 @@
             "version_added": "4"
           },
           "ie": {
-            "version_added": "≤6"
+            "version_added": "4"
```
foolip, line 23: This matches https://developer.mozilla.org/en-us/docs/Web/API/Window/document#browser_compatibility.

foolip (approved):

PR #8979, opened 2021-02-04
Add IE/Edge versions for HTMLElement API

vinyldarkscratch:
This PR adds real values for Internet Explorer and Edge for the `HTMLElement` API, based upon manual testing.

Test Code Used: `var instance = document.createElement('element'); instance.outerText !== undefined;`

### DIFF
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -2387,7 +2387,7 @@
               "version_added": false
             },
             "ie": {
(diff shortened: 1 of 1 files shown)


foolip (approved):
This also matches https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/innerText#browser_compatibility

PR #8980, opened 2021-02-04
Add IE/Edge versions for HTMLEmbedElement API

vinyldarkscratch:
This PR adds real values for Internet Explorer and Edge for the `HTMLEmbedElement` API, based upon manual testing.

Test Code Used: `var instance = document.createElement('embed'); instance.*;`


foolip (approved):
https://developer.mozilla.org/en-US/docs/Web/HTML/Element/embed#browser_compatibility could also be updated to say IE6.

Issue #8798, opened 2021-01-19
api.Blob.arrayBuffer - Safari 14 compatibility


poicooler:
#### What information was incorrect, unhelpful, or incomplete?
It is said that it is not supported in Safari.

#### What did you expect to see?
Blob.arraybuffer works on Safari 14

#### Did you test this? If so, how?
Yes, imported a file and `file.arrayBuffer()` works

<details>
<summary>MDN page report details</summary>

* Query: `api.Blob.arrayBuffer`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/Blob/arrayBuffer
* Report started: 2021-01-19T19:36:43.800Z

</details>

ddbeck:
Hi @poicooler, and thanks for reporting this issue.

A quick test shows that `Blob.arrayBuffer()` was first exposed in Safari 14 (it's not available in Safari 13). It's probably safe to assume that this is true in Safari for iOS as well.

I'd welcome a PR from you (or another contributor) to update [`Blob.json`](https://github.com/mdn/browser-compat-data/blob/master/api/Blob.json#L101-L149) to show that the `arrayBuffer` feature was added in version 14.

Thank you!

foolip:
Fixed by @joneskj55 in https://github.com/mdn/browser-compat-data/pull/8996. \o/

pastelmind:
It seems iOS Safari 14 does not support `blob.arrayBuffer`.

- Tested on iPhone 12 mini (iOS 14.1)
- My user agent: `Mozilla/5.0 (iPhone; CPU iPhone OS 14_1 like Mac OS X) AppleWebkit/605.1.15 (KHTML, like Gecko) Version/14.0 Mobile/15E148 Safari/604.1`
- Test link: https://runjs.co/s/ZWfL02tRY

foolip:
@pastelmind can you file a new issue about this? I just tested https://runjs.co/s/ZWfL02tRY and https://mdn-bcd-collector.appspot.com/tests/api/Blob/arrayBuffer in Safari iOS 14.5 and it is supported there. It's possible it wasn't in Safari 14, but it sounds like I tested just that in https://github.com/mdn/browser-compat-data/pull/8996#pullrequestreview-584566856.

foolip:
I also tested https://mdn-bcd-collector.appspot.com/tests/api/Blob/arrayBuffer on iOS 14 on BrowserStack, and it's also supported.

pastelmind:
[...]

PR #8996, opened 2021-02-04
Update Safari version for arrayBuffer()

joneskj55:
This PR addresses issue #8798 updating data for Safari and Safari iOS.

### DIFF
--- a/api/Blob.json
+++ b/api/Blob.json
@@ -128,10 +128,10 @@
               "version_added": "54"
             },
             "safari": {
-              "version_added": false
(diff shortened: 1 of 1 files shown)


foolip (approved):
I can confirm by testing https://mdn-bcd-collector.appspot.com/tests/api/Blob/arrayBuffer on Safari 14 on both macOS and iOS. I also confirmed it's *not* in Safari 13.1 on macOS.

foolip:
Thanks @joneskj55!

PR #9002, opened 2021-02-04
Add IE/Edge versions for NodeList API

vinyldarkscratch:
This PR adds real values for Internet Explorer and Edge for the `NodeList` API, based upon manual testing.

Test Code Used: https://mdn-bcd-collector.appspot.com/tests/api/NodeList

### DIFF
--- a/api/NodeList.json
+++ b/api/NodeList.json
@@ -20,7 +20,7 @@
             "version_added": "4"
           },
           "ie": {
-            "version_added": "≤6"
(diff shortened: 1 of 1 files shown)


foolip (approved):
https://mdn-bcd-collector.appspot.com/tests/api/NodeList depends on `document.childNodes`, but I'm pretty sure that's the earliest API using `NodeList`, so this seems right.

PR #9007, opened 2021-02-04
Add IE versions for CharacterData/Text API

vinyldarkscratch:
This PR adds real values for Internet Explorer and Edge for the `Text` API (and in turn, the `CharacterData` mixin it implements), based upon manual testing.

Test Code Used: https://mdn-bcd-collector.appspot.com/tests/api/Text & https://mdn-bcd-collector.appspot.com/tests/api/CharacterData

Resolves an issue blocking #8969.

### DIFF
--- a/api/CharacterData.json
+++ b/api/CharacterData.json
@@ -20,7 +20,7 @@
             "version_added": true
           },
           "ie": {
-            "version_added": "6"
+            "version_added": "5"
           },
           "opera": {
             "version_added": "≤12.1"
@@ -115,7 +115,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": "6"
+              "version_added": "5"
             },
             "opera": {
               "version_added": "≤12.1"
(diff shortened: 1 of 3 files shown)


foolip:
https://mdn-bcd-collector.appspot.com/tests/api/Text depends on `el.childNodes` and it seems quite possible that text nodes had a DOM representation in IE before `childNodes` was supported. What did your testing in IE 4 reveal, in what manner did the test fail?

vinyldarkscratch:
It is possible that the `Text` API was implemented in IE 4, but so far I've found no way to obtain an instance for it.  None of the methods I've found to potentially get a `Text` node (`document.all`, `Node.childNodes`, `document.createTextNode`...) are supported in IE 4, but all of them are supported in IE 5.

foolip:
@vinyldarkscratch let's go with IE 5 then :)

I just spotted a minor issue in the description. `CharacterData` it not a mixin, it's an abstract interface. The difference is that `Text` inherits from it via the prototype chain, and things like `CharacterData.prototype.substringData` exist.

foolip (approved):

PR #9015, opened 2021-02-05
Require that version_removed is either a string or true

foolip:
Setting it explicitly to null or false is meaningless.

### DIFF
--- a/schemas/compat-data.schema.json
+++ b/schemas/compat-data.schema.json
@@ -43,8 +43,20 @@
           "type": "boolean",
           "description": "A boolean value indicating whether or not the implementation of the sub-feature deviates from the specification in a way that may cause compatibility problems. It defaults to false (no interoperability problems expected). If set to true, it is recommended that you add a note explaining how it diverges from the standard (such as that it implements an old version of the standard, for example)."
         },
-        "version_added": { "$ref": "#/definitions/version_value" },
-        "version_removed": { "$ref": "#/definitions/version_value" },
+        "version_added": {
+          "description": "A string the value true, indicating which browser version added this feature, or the value false indicating the feature is not supported, or the value null indicating support is unknown.",
+          "allOf": [
+            { "type": ["string", "boolean", "null"] },
+            { "not": { "enum": ["true", "false", "null"] } }
+          ]
+        },
+        "version_removed": {
+          "description": "A string or the value true, indicating which browser version removed this feature.",
+          "allOf": [
+            { "type": ["string", "boolean"] },
+            { "not": { "enum": ["true", false] } }
+          ]
+        },
         "notes": {
           "description": "A string or array of strings containing additional information.",
           "anyOf": [
@@ -64,7 +76,7 @@
       "required": ["version_added"],
       "anyOf": [
         {
-          "not": {"required": ["prefix","alternative_name"]}
+          "not": {"required": ["prefix", "alternative_name"]}
         },
         {
           "oneOf": [
@@ -258,13 +270,6 @@
         "^(?!__compat)[a-zA-Z_0-9-$@]*$" : { "$ref": "#/definitions/webextensions_identifier" }
       },
       "additionalProperties": false
-    },
-
-    "version_value": {
-      "allOf": [
-        { "type": ["string", "boolean", "null"] },
-        { "not": { "enum": ["true", "false", "null"] } }
-      ]
     }
   },
 
--- a/schemas/compat-data.schema.json
+++ b/schemas/compat-data.schema.json
@@ -51,7 +51,7 @@
           ]
         },
         "version_removed": {
-          "description": "A string or the value true, indicating which browser version removed this feature.",
+          "description": "A string, indicating which browser version removed this feature, or the value true, indicating that the feature was removed in an unknown version.",
           "allOf": [
             { "type": ["string", "boolean"] },
             { "not": { "enum": ["true", false] } }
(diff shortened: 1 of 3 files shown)


foolip:
This still needs work to actually enforce it, but I'm not sure if it's should go in `compat-data.schema.json`, in `test-versions.js`, or both?

ddbeck:
Change `compat-data.schema.json`. Then stuff like the VS Code settings will automatically pick this up. Let's mess with the linter only if there's confusion around the errors raised by the schema change.

schemas/compat-data.schema.json
```diff
@@ -43,8 +43,20 @@
           "type": "boolean",
           "description": "A boolean value indicating whether or not the implementation of the sub-feature deviates from the specification in a way that may cause compatibility problems. It defaults to false (no interoperability problems expected). If set to true, it is recommended that you add a note explaining how it diverges from the standard (such as that it implements an old version of the standard, for example)."
         },
-        "version_added": { "$ref": "#/definitions/version_value" },
-        "version_removed": { "$ref": "#/definitions/version_value" },
+        "version_added": {
+          "description": "A string the value true, indicating which browser version added this feature, or the value false indicating the feature is not supported, or the value null indicating support is unknown.",
+          "allOf": [
+            { "type": ["string", "boolean", "null"] },
+            { "not": { "enum": ["true", "false", "null"] } }
```
foolip, line 50: Since it's not obvious: the "true", "false" and "null" here merely disallow those *strings*, presumably to help highlight such mishaps in the editor. (The linter would also complain about it.)

schemas/compat-data.schema.json
```diff
@@ -43,8 +43,20 @@
...
+          ]
+        },
+        "version_removed": {
+          "description": "A string or the value true, indicating which browser version removed this feature.",
+          "allOf": [
+            { "type": ["string", "boolean"] },
+            { "not": { "enum": ["true", false] } }
```
foolip, line 57: This is not the most obvious way of saying that true is an allowed value, but convenient since the string "true" is already excluded here.

foolip:
@ddbeck I've updated `compat-data.schema.json` now.

schemas/compat-data.schema.json
```diff
@@ -43,8 +43,20 @@
           "type": "boolean",
           "description": "A boolean value indicating whether or not the implementation of the sub-feature deviates from the specification in a way that may cause compatibility problems. It defaults to false (no interoperability problems expected). If set to true, it is recommended that you add a note explaining how it diverges from the standard (such as that it implements an old version of the standard, for example)."
         },
-        "version_added": { "$ref": "#/definitions/version_value" },
-        "version_removed": { "$ref": "#/definitions/version_value" },
+        "version_added": {
+          "description": "A string the value true, indicating which browser version added this feature, or the value false indicating the feature is not supported, or the value null indicating support is unknown.",
```
ddbeck, line 47:
```suggestion
          "description": "A string (indicating which browser version added this feature), the value true (indicating support added in an unknown version), the value false (indicating the feature is not supported), or the value null (indicating support is unknown).",
```

schemas/compat-data.schema.json
ddbeck: Yep. History of this is in #485.

schemas/compat-data.schema.json
```diff
@@ -43,8 +43,20 @@
...
+          "allOf": [
+            { "type": ["string", "boolean", "null"] },
+            { "not": { "enum": ["true", "false", "null"] } }
+          ]
+        },
+        "version_removed": {
+          "description": "A string or the value true, indicating which browser version removed this feature.",
```
ddbeck, line 54:
```suggestion
          "description": "A string, indicating which browser version removed this feature, or the value true, indicating that the feature was removed in an unknown version.",
```

ddbeck (changes requested):
A couple little suggestions on the wording of the descriptions, but nothing major. Also, I checked out this branch locally and confirmed that the schema changed worked as expected by seeing my editor correctly flag some now-invalid changes. 👍

foolip:
@ddbeck suggestions applied!

ddbeck (approved):
Thank you!

ddbeck:
@Elchi3 this is a (what looks to me a non-breaking) schema change, so I thought I'd give you a chance to review it too. Though if you're not concerned about it, dismiss the review request and I'll go ahead.

foolip:
Given that the data and schema go together, what would constitute a breaking schema change?

Since this change makes the schema stricter it should be breaking by some definition, but I'm not sure how in practice.

ddbeck:
This was my thinking: suppose we changed the data, but didn't change the schema. That wouldn't be breaking, though it would be the case that our data, at present, doesn't contain any `"version_removed": false` or `"version_removed": null` data (but still could, in the future—this is current state of non-real values in `css` for example). Consumers might (should, even) have code to handle such data, but none would appear in practice.

But by changing the schema, we're saying to consumers, "you can safely discard any code that expects `"version_removed": false` or `"version_removed": null`". Now that I've spelled this out, this sounds like a feature, so it's semver minor (e.g., the next version could be v3.2.0, not v3.1.4).

That said, if we wanted to play things safe, we could treat any schema change as breaking.

foolip:
Ah, right, the schema is our "API" and consumers can now make some more assumptions about the data, which is exactly why I made the change.

I don't particularly mind which part of the version is bumped, but could see the argument for either minor or patch version. Bumping the minor version might bring more attention to the change, allowing more consumers to simplify their code.

Elchi3 (approved):
Makes sense to me! We should have had this from this start :) 
Semver minor sounds good to me, too.

ddbeck:
Thanks, @foolip and @Elchi3! 🎉

I've started draft PR #9404 for this week's release.

PR #9017, opened 2021-02-05
Update picture-in-picture feature policy for Chrome

### LINKED ISSUE #8260: http.headers.Feature-Policy.picture-in-picture - Supported in Chrome

The Browser compatibility table on https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Feature-Policy/picture-in-picture appears to be out of date. It shows that `picture-in-picture` is not supported on any browser. I think it is supported on some browsers.

I expected the table to indicate support for it in some browsers, because based on my testing it appears that it is supported.

In Chrome 87 (on macOS), I loaded a test page structured as follows. The main frame has an iframe to a page on another origin that has a video and on user action will launch picture-in-picture. Triggering the picture-in-picture causes the video to switch to picture-in-picture mode. (So far, this is all expected behavior and doesn't have to do with the `picture-in-picture` feature policy.) Now, I modified the iframe to add `allow="picture-in-picture 'none'"` and reload the page. With that change, when I try to trigger the picture-in-picture, it doesn't switch to that mode, and I get in the console "Uncaught (in promise) DOMException: Failed to execute 'requestPictureInPicture' on 'HTMLVideoElement': Access to the feature "picture-in-picture" is disallowed by feature policy."

This appears to be evidence that Chrome 87 supports the `picture-in-picture` feature policy.

I haven't tested in other browsers (or even other versions of Chrome to determine when this support was introduced), so unfortunately I don't have advice for how the table should be updated.

<details>

* Query: `http.headers.Feature-Policy.picture-in-picture`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Feature-Policy/picture-in-picture
* Report started: 2020-12-23T22:27:36.234Z

</details>

joneskj55:
This PR fixes #8260 which updates the picture-in-picture support for Chrome.

### DIFF
--- a/http/headers/feature-policy.json
+++ b/http/headers/feature-policy.json
@@ -1344,7 +1344,7 @@
             "mdn_url": "https://developer.mozilla.org/docs/Web/HTTP/Headers/Feature-Policy/picture-in-picture",
             "support": {
               "chrome": {
-                "version_added": false
+                "version_added": 71
               },
               "chrome_android": {
                 "version_added": false
--- a/http/headers/feature-policy.json
+++ b/http/headers/feature-policy.json
@@ -1344,7 +1344,7 @@
             "mdn_url": "https://developer.mozilla.org/docs/Web/HTTP/Headers/Feature-Policy/picture-in-picture",
             "support": {
               "chrome": {
-                "version_added": 71
+                "version_added": "71"
               },
               "chrome_android": {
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Looks great. Thank you! 🎉

PR #9026, opened 2021-02-07
Mark Animation constructor and members as shipped in Safari 13.1/13.4

foolip:
Only the parent api.Animation feature was updated in https://github.com/mdn/browser-compat-data/pull/5963.

This updates members according to mdn-bcd-collector results from Safari
13/13.1 on macOS and 13.3/13.4 on iOS.

### DIFF
--- a/api/Animation.json
+++ b/api/Animation.json
@@ -150,10 +150,10 @@
               "version_added": "45"
             },
             "safari": {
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Looks good. Thank you! 🎉

PR #9041, opened 2021-02-09
Fix Opera Android version for min-width CSS property

vinyldarkscratch:
This PR fixes the mirroring of Opera to Opera Android for the CSS `min-width` property.  For some odd reason, I had mirrored this improperly.


ddbeck (approved):
Thank you! 🎉

PR #9042, opened 2021-02-09
Fix mirroring of Chrome -> WebView for css.properties.word-spacing

vinyldarkscratch:
This PR fixes the WebView data for the `word-spacing` CSS property. A while back, I had performed mirroring, however at the time I had not accounted for ranged versions, or Chrome 1. This resolves said issues, as anything in Chrome 1 would certainly be in WebView 1.


ddbeck (approved):
Thanks! 🎉

PR #9043, opened 2021-02-09
Fix mirroring of Chrome -> WebView for css.selectors.hover

vinyldarkscratch:
This PR fixes the WebView data for the `hover` CSS selection. A while back, I had performed mirroring, however at the time I had not accounted for ranged versions, or Chrome 1. This resolves said issues, as anything in Chrome 1 would certainly be in WebView 1.

### DIFF
--- a/css/selectors/hover.json
+++ b/css/selectors/hover.json
@@ -41,7 +41,7 @@
               "version_added": "1.0"
             },
             "webview_android": {
-              "version_added": "37"
+              "version_added": "1"
             }
           },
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
> as anything in Chrome 1 would certainly be in WebView 1

I don't think this is strictly true (IIRC, the oldest WebViews were WebKit-derived independently of Chrome). In this case, the feature was in a Safari that long predates Android, so I'm OK with making this assumption here.

Issue #8829, opened 2021-01-22
api.URLSearchParams - URLSearchParams.prototype.toString doesn't work as expected until Firefox 54


romainmenke:
#### What information was incorrect, unhelpful, or incomplete?

`URLSearchParams.prototype.toString` is stated as supported (without notice)  from Firefox 29.

#### What did you expect to see?

Working `URLSearchParams.prototype.toString` from Firefox 29.

In practice you get `[Object Object]` URL encoded.

This works correctly from Firefox 54

#### Did you test this? If so, how?

Ran this code in all browsers available in Browserstack :

```js
var params = new URLSearchParams({ key: "730d67" });
params.toString() === "key=730d67";
```

also see : https://github.com/Financial-Times/polyfill-library/pull/983

<details>
<summary>MDN page report details</summary>

* Query: `api.URLSearchParams`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams
* Report started: 2021-01-22T18:46:34.190Z

</details>

ddbeck:
Hi @romainmenke and thanks for reporting this issue.

It's possible this might actually be a problem with the `URLSearchParams()` constructor (`api.URLSearchParams.URLSearchParams`), not `toString` per se. It looks like Firefox didn't accept records as the `init` argument until Firefox 54: https://bugzilla.mozilla.org/show_bug.cgi?id=1331580.

For example, compare the output from this in Firefox 53 and 54:

```js
var z = new URLSearchParams({'abc': 123});
console.log(z.get('abc')); // null, then 123
z.get('abc') === "123"; // false, then true
```

I'd suggest instead that we correct this by adding a subfeature to `api.URLSearchParams.URLSearchParams` (like `USVString_sequence`) in [`URLSearchParams.json`](https://github.com/mdn/browser-compat-data/blob/5d2366a7237c9fbb63ad38bd69e7a0c534d29f3c/api/URLSearchParams.json#L61) that represents support for record objects as `init`. I'd welcome a PR (from you or another contributor) seeding this data with Firefox versions alone, though experimentation in other browsers would also be welcome.

Thanks again!

romainmenke:
@ddbeck I agree. Wrote more tests to get a better view of browser support for records as init arguments.
Will get back to you when I have some results.

romainmenke:
Still need to run more tests but at first glance it seems these all landed at different times :
- `new URLSearchParams( "foo=bar" ); // firefox 29`
[...]

PR #9049, opened 2021-02-09
details for the different types in URLSearchParams.constructor

romainmenke:
fixes : #8829

In short : 

`URLSearchParams.constructor` supports multiple types of init args.
Support for each type was added in different versions for most browser engines.

This change is intended to make it more clear which browsers support which init types.

_I am not yet used to all the naming conventions here so any feedback is super welcome!_

### DIFF
--- a/api/URLSearchParams.json
+++ b/api/URLSearchParams.json
@@ -181,7 +181,7 @@
                 "version_added": "17"
               },
               "firefox": {
-                "version_added": "53"
+                "version_added": "54"
               },
               "firefox_android": {
                 "version_added": "54"
@@ -196,7 +196,7 @@
                 "version_added": "48"
               },
               "opera_android": {
-                "version_added": "48"
+                "version_added": "45"
               },
               "safari": {
                 "version_added": "11"
@@ -259,7 +259,7 @@
                 "version_added": "7.2"
               },
               "webview_android": {
-                "version_added": "61"
+                "version_added": "58"
               }
             },
             "status": {
--- a/api/URLSearchParams.json
+++ b/api/URLSearchParams.json
@@ -167,57 +167,6 @@
             }
           }
         },
-        "sequence": {
-          "__compat": {
-            "description": "sequence for <code>init</code> object",
-            "support": {
-              "chrome": {
-                "version_added": "58"
-              },
-              "chrome_android": {
-                "version_added": "58"
-              },
-              "edge": {
-                "version_added": "17"
-              },
-              "firefox": {
-                "version_added": "53"
-              },
-              "firefox_android": {
-                "version_added": "53"
-              },
-              "ie": {
-                "version_added": false
-              },
-              "nodejs": {
-                "version_added": "7.10.0"
-              },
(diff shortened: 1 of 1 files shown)


api/URLSearchParams.json
```diff
@@ -133,32 +184,25 @@
                 "version_added": "53"
```
ddbeck, line 184: It looks like this should also be version 54 (making desktop and Android line up)

```suggestion
                "version_added": "54"
```

api/URLSearchParams.json
```diff
@@ -133,32 +184,25 @@
...
-              ],
+              "nodejs": {
+                "version_added": "7.10.0"
+              },
               "opera": {
                 "version_added": "48"
               },
               "opera_android": {
-                "version_added": "45"
+                "version_added": "48"
```
ddbeck, line 199: I think this ought to be 45 still (Opera desktop and Android version numbers aren't lined up, unfortunately).

```suggestion
                "version_added": "45"
```

api/URLSearchParams.json
```diff
@@ -173,6 +217,57 @@
...
+                "version_added": "11"
+              },
+              "safari_ios": {
+                "version_added": "11"
+              },
+              "samsunginternet_android": {
+                "version_added": "7.2"
+              },
+              "webview_android": {
+                "version_added": "61"
```
ddbeck, line 262: To sync up with Chrome and Chrome for Android

```suggestion
                "version_added": "58"
```

ddbeck (changes requested):
@romainmenke This looks really good. Well done! I've made a few small suggestions in line comments for your consideration. Thank you!

api/URLSearchParams.json
romainmenke: Good catch! 
I didn't know that they started drifting apart at some point.

api/URLSearchParams.json
```diff
@@ -173,6 +217,57 @@
...
+                "version_added": false
+              },
+              "nodejs": {
+                "version_added": "7.10.0"
+              },
+              "opera": {
+                "version_added": "45"
+              },
+              "opera_android": {
+                "version_added": "45"
```
romainmenke, line 250:
```suggestion
                "version_added": "43"
```

I think Opera Android 43 (with Blink 59) is the first release with support.

Opera Android 42 has Blink 55

romainmenke:
@ddbeck Thank you for the review! I applied the changes and I think I also spotted one more Opera Android version that needs to be updated.

api/URLSearchParams.json
ddbeck: Good catch! This squares with https://github.com/mdn/browser-compat-data/blob/master/browsers/opera_android.json#L206

I'm going to apply this, approve, and merge. 👍

ddbeck (approved):
Excellent work, @romainmenke! Thank you! 🎉

romainmenke:
Awesome 🎉  thank you for taking to review this and follow up on everything 🙂

Issue #7991, opened 2020-12-17
api.Clipboard.{read,write} - Split MIME types into subfeatures?


ddbeck:
#### What information was incorrect, unhelpful, or incomplete?

The [_Clipboard API and events_ spec](https://www.w3.org/TR/clipboard-apis/) calls for supporting [several mandatory data types](https://www.w3.org/TR/clipboard-apis/#mandatory-data-types-x) when reading and writing from the clipboard. But actual implementation varies (see #7614).

#### What did you expect to see?

We ought to create subfeatures for each of the specified types and track their support independently of the `read` and `write` methods.

#### Can you link to any release notes, bugs, pull requests, or MDN pages related to this?

- #7614 - inciting PR here
- #6313 - earliest case of this

jculver050:
Close

PR #9058, opened 2021-02-10
Update Firefox Clipboard API support

evilpie:
- ClipboardItem is now implemented behind a pref.
- clipboard.read and clipboad.write are also behind the same pref.
- clipboard.write now supports more types, not just text.
- Chrome's implementation of clipboard.read is marked as partial, so maybe Firefox's should be as well? Sadly the spec here is a very bad state. By the same token clipboard.write could be marked as partial... 
- It's also interesting to note that the previous implementation of read/write used DataTransfer, which was completely different from the specified ClipboardItem.

https://bugzilla.mozilla.org/show_bug.cgi?id=1619947
https://bugzilla.mozilla.org/show_bug.cgi?id=1689886

The last bug hasn't been merged yet, so waiting on that.

### DIFF
--- a/api/Clipboard.json
+++ b/api/Clipboard.json
@@ -103,6 +103,7 @@
               {
                 "version_added": "63",
                 "version_removed": "87",
+                "partial_implementation": true,
                 "flags": [
                   {
                     "type": "preference",
@@ -110,7 +111,10 @@
                     "value_to_set": "true"
                   }
                 ],
-                "notes": "Currently works just like <code>readText()</code>; non-text content is not currently supported."
+                "notes": [
+                  "This method returns a <code>DataTransfer</code> object instead of an array of <code>ClipboardItem</code> objects.",
+                  "Firefox only supports reading the clipboard in browser extensions, using the <code>\"clipboardRead\"</code> extension permission."
+                ]
               }
             ],
             "firefox_android": [
@@ -128,6 +132,7 @@
               {
                 "version_added": "63",
                 "version_removed": "87",
+                "partial_implementation": true,
                 "flags": [
                   {
                     "type": "preference",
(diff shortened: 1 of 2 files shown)


evilpie:
(Landed now)

ddbeck:
> Chrome's implementation of clipboard.read is marked as partial, so maybe Firefox's should be as well? Sadly the spec here is a very bad state. By the same token clipboard.write could be marked as partial...

No, don't mark Firefox's support as partial. Instead, we ought to break this down into subfeatures that express the supported MIME types. https://github.com/mdn/browser-compat-data/issues/7991 describes what ought to happen with that.

ddbeck:
> It's also interesting to note that the previous implementation of read/write used DataTransfer, which was completely different from the specified ClipboardItem.

I'm not sure what this means. Is this something we should add as a note? Can you propose some text? Or would this be better suited to a separate PR?

evilpie:
> > It's also interesting to note that the previous implementation of read/write used DataTransfer, which was completely different from the specified ClipboardItem.
> 
> I'm not sure what this means. Is this something we should add as a note? Can you propose some text? Or would this be better suited to a separate PR?

I think we should add a note. Maybe something like for e.g. read: "In Firefox this API returns a DataTransfer object instead of an array (called sequence in WebIDL?) of ClipboardItems". write would be s/returns/accepts/.

ddbeck:
@evilpie

> I think we should add a note. Maybe something like for e.g. read: "In Firefox this API returns a DataTransfer object instead of an array (called sequence in WebIDL?) of ClipboardItems". write would be s/returns/accepts/.

I like this a lot. It solves a lot of problems with the existing note text, too. Running with your text a bit, I think the note on those support statements would be:

* Read: `This method returns a <code>DataTransfer</code> object instead of an array of <code>ClipboardItem</code> objects.`
* Write: `This method accepts a <code>DataTransfer</code> object instead of an array of <code>ClipboardItem</code> objects.`

This actually strikes me as a sufficiently different behavior that we should use the partial implementation flag on the range with this behavior.

Do we still need a second note for each covering non-text? Something like `Non-text content is not supported.`?

evilpie:
Seems to me like the note about non-text in read was actually not correct anyway.

ddbeck (approved):
This is a big improvement to the data here. Thank you, @evilpie!

PR #9060, opened 2021-02-11
api.HTMLCanvasElement.getContext.powerPreference works on macOS only

ddbeck:
This is a substitute to the stalled PR #7667. The feature was only ever implemented in Firefox on macOS. I did some checking and it has not been implemented in Chromium or WebKit, so I set the experimental flag to `true` while I was here.

https://bugzilla.mozilla.org/show_bug.cgi?id=1349799
https://bugzilla.mozilla.org/show_bug.cgi?id=1496459

Closes #7667.

### DIFF
--- a/api/HTMLCanvasElement.json
+++ b/api/HTMLCanvasElement.json
@@ -314,10 +314,12 @@
                 "version_added": false
               },
               "firefox": {
-                "version_added": "63"
+                "version_added": "63",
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
The details no one told me when I documented this! 😄 Thanks for digging into this story!

PR #9068, opened 2021-02-11
Fix IE data for Element API

vinyldarkscratch:
This PR fixes the Element API data for Internet Explorer.  Previously, the API was listed as supported since IE 4 (apparently, by my doing).  However, so far I haven't found a single way to obtain an Element instance in IE 4, whether through variables or methods.

I believe that when I initially set IE to "4" for this API, I must have made a typo on my document and not realized it to be so.  The data for the modified subfeatures also came from the initial wiki page migration, so I'm less likely to trust its data.

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -20,7 +20,7 @@
             "version_added": "4"
           },
           "ie": {
-            "version_added": "4"
+            "version_added": "5"
           },
           "opera": {
             "version_added": "8"
@@ -4048,7 +4048,7 @@
               "version_added": "4"
             },
             "ie": {
-              "version_added": "4"
+              "version_added": "5"
             },
             "opera": {
               "version_added": "9.5"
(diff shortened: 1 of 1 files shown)


foolip (approved):
https://developer.mozilla.org/en-US/docs/Web/API/Node#browser_compatibility says IE5, so this makes sense.

foolip:
Hmm, but have you tried `document.forms`? That should be in IE4, or https://developer.mozilla.org/en-US/docs/Web/API/Document/forms#browser_compatibility is wrong too :)

vinyldarkscratch:
So it looks like `document.forms` _is_ present in IE 4, good catch!  Using `document.forms`, I've now determined that the `Element` API is indeed supported earlier.  I'm going to open up an issue to revise the `Element` and `Node` APIs for IE 4/5 data based upon this.

PR #9074, opened 2021-02-12
feComposite support for operator lighter property

hamishwillee:
FF86 introduces support for the "lighter" filter attribute - see [bug 1518099](https://bugzilla.mozilla.org/show_bug.cgi?id=1518099).

The current BCD on feComposite is a bit of a mess - for example it says that this is already supported in FF and is not supported in Chrome, neither of which is true. 

Bug info for main platforms here:
- [Chrome](https://bugs.chromium.org/p/chromium/issues/detail?id=439037) - release 45, which I got [from here](https://github.com/chromium/chromium/commit/5a681e630d857171bab7f21dfc4f109cfb03d41f). I have rolled out that support to equivalent versions for Opera, Samsung etc.
- [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1518099) - FF86
- [Webkit 141376](https://bugs.webkit.org/show_bug.cgi?id=141376) - fixed in v528+

Following on from those bugs it is pretty clear that feComposite must be supported in Blink and Webkit derived browsers. Have marked as true. For Blink have marked as version version where Blink was supported.

Docs bug is https://github.com/mdn/content/issues/1727

This may also affect SVGFECompositeElement but still [checking](https://bugzilla.mozilla.org/show_bug.cgi?id=1518099#c9).


svg/elements/feComposite.json
```diff
@@ -24,16 +24,16 @@
               "version_added": null
             },
             "opera": {
-              "version_added": false
+              "version_added": true
```
hamishwillee, line 27: Note, these must all be true/supported because you can't have support for "lighter" if the parent item is not supported (as per bugs). I am not sure what versions. 

If you believe chrome then this was in from release 1 (webkit and from then in Blink). Should I leave these alone or assume that support came in earliest matching version of Blink/Webkit for the other browsers? That would mean, for example, that Opera would change to version 15 and I'd assume there was no support in Presto. Thoughts?

svg/elements/feComposite.json
```diff
@@ -108,10 +108,12 @@
                 "version_added": "≤18"
               },
               "firefox": {
-                "version_added": true
+                "version_added": true,
+                "notes": "BackgroundImage/BackgroundAlpha are not supported (see <a href='https://bugzil.la/1171438'>bug 1171438</a>)."
```
hamishwillee, line 112: This is documented in [bug 437554](https://bugzilla.mozilla.org/show_bug.cgi?id=437554) 
- Not yet clear if this also affects the `in` attribute too. 
- [this comment](https://bugzilla.mozilla.org/show_bug.cgi?id=437554#c18) indicates that this also affects other browsers. I'm not sure though. The "workaround" code shows this fails on FF but not the others - run this on both different browsers to see: https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/in#result

evilpie: The bug reference seems wrong.

hamishwillee: @evilpie You're right. Fixed, and rolled out to both in and in2.

svg/elements/feComposite.json
```diff
@@ -24,19 +24,19 @@
               "version_added": null
             },
             "opera": {
-              "version_added": false
+              "version_added": "15"
```
hamishwillee, line 27: Chrome has full support from version 1. It got Blink in v28 which corresponds to Opera 15)

svg/elements/feComposite.json
```diff
@@ -24,19 +24,19 @@
...
             },
             "opera_android": {
-              "version_added": null
+              "version_added": "14"
```
hamishwillee, line 30: Chrome has full support from version 1. It got Blink in v28 which corresponds to Opera Android 14)

svg/elements/feComposite.json
```diff
@@ -24,19 +24,19 @@
...
-              "version_added": null
+              "version_added": true
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": true
             },
             "samsunginternet_android": {
-              "version_added": true
+              "version_added": "1.5"
```
hamishwillee, line 39: Chrome has full support from version 1. It got Blink in v28 which corresponds to Samsung Internet 1.5

svg/elements/feComposite.json
```diff
@@ -24,19 +24,19 @@
...
-              "version_added": false
+              "version_added": "15"
             },
             "opera_android": {
-              "version_added": null
+              "version_added": "14"
             },
             "safari": {
-              "version_added": null
+              "version_added": true
```
hamishwillee, line 33: Marked this true as there are webkit bugs against lighter for operator - which implies that feComposite must be supported.

svg/elements/feComposite.json
```diff
@@ -24,19 +24,19 @@
...
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": true
```
hamishwillee, line 36: Marked this true as there are webkit bugs against lighter for operator - which implies that feComposite must be supported.

api/SVGFECompositeElement.json
```diff
@@ -51,40 +51,52 @@
         "__compat": {
           "support": {
             "chrome": {
-              "version_added": "5"
+              "version_added": "5",
+              "notes": "BackgroundImage/BackgroundAlpha are not supported (see <a href='https://crbug.com/137230'>Chromium bug 137230</a>)."
```
hamishwillee, line 55: FYI added all these notes to mirror `feComposite` (which this API does).

api/SVGFECompositeElement.json
```diff
@@ -329,6 +351,54 @@
           }
         }
       },
+      "lighterForError": {
```
hamishwillee, line 354: Robert (domain expert) [confirmed](https://bugzilla.mozilla.org/show_bug.cgi?id=1518099#c13) lighterForError should be added to SVGFECompositeElement too. I have matched the feComposite versions exactly except for Safari. 

For Safari the `operator` was not added until version 6, so I pushed the implementation of operator value lighter to be that same version. That is not "rock solid" (i.e. no bug confirmation) but very likely.

api/SVGFECompositeElement.json
```diff
@@ -329,6 +353,54 @@
...
+            "edge": {
+              "version_added": "≤18"
+            },
+            "firefox": {
+              "version_added": "86"
+            },
+            "firefox_android": {
+              "version_added": "86"
+            },
+            "ie": {
```
hamishwillee, line 375: Note, IE doesn't work support lighter for feComposite, so set false in that and here too (verified by testing).

svg/elements/feComposite.json
```diff
@@ -52,40 +52,52 @@
...
               },
               "chrome_android": {
-                "version_added": true
+                "version_added": true,
+                "notes": "BackgroundImage/BackgroundAlpha are not supported (see <a href='https://crbug.com/137230'>Chromium bug 137230</a>)."
               },
               "edge": {
-                "version_added": "≤18"
+                "version_added": "≤18",
+                "notes": "BackgroundImage/BackgroundAlpha are not supported from version 79(see <a href='https://crbug.com/137230'>Chromium bug 137230</a>)."
```
hamishwillee, line 64: FYI version 79 is when edge moved to Blink, and lost the ability to support BackgroundImage.

svg/elements/feComposite.json
```diff
@@ -52,40 +52,52 @@
...
+                "notes": "BackgroundImage/BackgroundAlpha are not supported (see <a href='https://bugzil.la/437554'>bug 437554</a>)."
               },
               "firefox_android": {
-                "version_added": true
+                "version_added": true,
+                "notes": "BackgroundImage/BackgroundAlpha are not supported (see <a href='https://bugzil.la/437554'>bug 437554</a>)."
               },
               "ie": {
-                "version_added": null
+                "version_added": true,
```
hamishwillee, line 75: I tested and this works (except for operator value `lighter`).  I have been told by Robert only edge prior to Blinks supports BackgroundImage/Alpha but have no issue to track - hence see note below has no bug number.

svg/elements/feComposite.json
```diff
@@ -202,28 +226,28 @@
                 "version_added": "≤79"
               },
               "firefox": {
-                "version_added": null
+                "version_added": true
```
hamishwillee, line 229: according to Robert `in`, `in2` and all the `k` values are supported by everything. So these are all moved from null to true.

hamishwillee:
@evilpie @ddbeck This should now be ready for review. I've added notes throughout on what I've done. General things to note to make review easier are that according to Robert, the Moz domain expert:
- if feComposite is supported, then `in`, `in2` and all the `k` values are supported by everything. So these are all moved from `null` to at least `true` (if version not known).
- feComposite and SVGFECompositeElement mirror each other in terms of functionality. That's why I updated that to have lighter and BackgroundImage. Versions might not completely match up though.
- BackgroundImage/BackgroundAlpha only supported by Edge prior to Blink support.

api/SVGFECompositeElement.json
ddbeck: Sorry, but we should drop this feature entirely. For reflected attributes, we typically just track whether the attribute exists as part of the API. This is us mostly following the convention of specifications for these things: the IDL usually specifies the reflection, while [the specification for the element defines valid values](https://drafts.fxtf.org/filter-effects/#feCompositeElement).

(Plus—and this is definitely not your fault, since you're following the structure from the element data—the structure is wrong. This ought to be a subfeature of `operator`, not a peer to `operator`.)

svg/elements/feComposite.json
```diff
@@ -332,44 +356,43 @@
         },
         "lighterForError": {
```
ddbeck, line 357: This is not your fault, but I wouldn't mind if we fixed this in the process of doing all this work here:

`lighterForError` should be a subfeature `lighter_value` (description: `<code>lighter<code> value`) of the `operator` feature. We could work through all the other changes that need to happen, then, as a final polish to this PR, move the feature underneath `operator`.

ddbeck (changes requested):
Wow, this is a complicated one! Thanks for digging into this, @hamishwillee. I've got a bunch of suggestions, but it took me a while to understand the whole story here, so it might be a bit confused in places. If there's any questions about my comments, please say something. Thank you!

api/SVGFECompositeElement.json
hamishwillee: Essentially yes. You can specify those of course (they are "valid" options) but the only thing they they work on is pre-Blink-Edge. So "no current browser". 

My read of the defect reports is because this is very hard to do, not because it isn't a good idea. Even the example in the SVG spec uses these. I have just removed that example from the docs. Also yes, that some browsers won't ever fix this.

But my concern is that without the note everyone will think all the spec values for in1 and in2 will work. I think you're saying "that doesn't matter". I will take your direction and update appropriately :-)

api/SVGFECompositeElement.json
hamishwillee: @ddbeck So to summarize, I just remove this addition altogether - we simply don't track things supported at this level in the BCD. Sounds nuts, but there you go.

svg/elements/feComposite.json
hamishwillee: We do not. Will update.

svg/elements/feComposite.json
hamishwillee: We do not. Will update.

svg/elements/feComposite.json
hamishwillee: Will do.

svg/elements/feComposite.json
hamishwillee: Will do.

hamishwillee:
@ddbeck Thanks for the review. I think I have now done the fixes. Hope I got it right.

hamishwillee:
PS Partially the response confused me because I don't know what a reflective property is.

api/SVGFECompositeElement.json
ddbeck: 1. To clear up a point of confusion: reflected attributes set and get the source element's attributes. For example, given page source `<feComposite in1="somevalue">`, `SVGFECompositeElement.in1` returns `"somevalue"` and `SVGFECompositeElement.in1 = "NEW"` sets the element to `<feComposite in1="NEW">`. In terms of behavior, reflected attributes provide access to underlying element but don't really add anything more.

   Consequently, to avoid duplication, we document support for attribute values on the element itself, not the interface to that element. We figure that people don't really learn about, for example, the attributes of HTML elements by reading the `HTML*Element` interface docs.

2. > my concern is that without the note everyone will think all the spec values for in1 and in2 will work
    
    I didn't realize the spec used these (unsupported) values as examples and I appreciate that you're trying to guard against misunderstanding here. In that case, I think the right way to handle this is to have subfeatures for `in1` and `in2` (on the `svg.elements.feComposite.{in1,in2}` features) that represent each of the values for it and provide complete compat data for each, instead of notes. I think that gives readers a clearer idea of what is and is not supported.

    There's one additional bit of nuance here, however: because spec authors write ahead of implementation, there are often features which are in specs and never implemented. BCD can be a source of additional confusion, as folks get the idea that a feature in BCD means that it's a live concern. This is why we avoid all `false` features, where possible.
    
    In this case, the `BackgroundImage` and `BackgroundAlpha` values were once supported by at least one browser, so that deserves an entry.

3. I was trying to avoid repetitious notes. Repetitious notes are hard to follow. When every entry has a note and the text is very similar between them, then it's hard to know what's actually noteworthy. That's why I was leaning toward writing notes for the exceptional, not usual, case.

api/SVGFECompositeElement.json
ddbeck: Yes, we should only have data on the element, not the interface. I got into this more in https://github.com/mdn/browser-compat-data/pull/9074#discussion_r579041812. But let me know if you have more questions—it's not an obvious distinction.

api/SVGFECompositeElement.json
ddbeck: Two additional thoughts here:

- I missed that IE supported BackgroundImage and BackgroundAlpha as well. I think my general point still stands, however, that these are interesting because they're different from the others.
- I said that we should "have subfeatures for `in1` and `in2` (on the `svg.elements.feComposite.{in1,in2}` features) that represent each of the values". I am not asking you to add features for every value yourself as part of this PR (that's optional). I just meant that `svg.elements.feComposite.{in1,in2}.BackgroundImage` and `svg.elements.feComposite.{in1,in2}.BackgroundAlpha` would be germane here.

api/SVGFECompositeElement.json
hamishwillee: @ddbeck What I think you're saying is that we don't document the bits of an SVG element that reflects a javascript API -  we'd just document the javascript behaviour. Right?

So I understand why you wanted me to delete the `lighterForError` from the reflected API  svg/elements/feComposite.json , but THIS is api/SVGFECompositeElement - the API from which we reflect. Which makes me think that it should have stayed. 

What am I missing?

api/SVGFECompositeElement.json
hamishwillee: @ddbeck You didn't miss anything about IE. I'm an idiot - those notes were supposed to be on Edge. Fixed. 

But my feeling is that this is probably worth merging now because it is "not wrong" and adds a lot of corrected information about supported versions. 

There are ongoing discussions like https://github.com/mdn/browser-compat-data/pull/9074#discussion_r579043292 and this one on possibly adding subfeatures, but I'd happily do those as a post process. Up to you.

api/SVGFECompositeElement.json
ddbeck: > Or did you mean that we document on the element but not the javascript API? Either way, doesn't one of them need comments about the features? The comments here said delete this from both /api and /svg

I meant on the element, not the API. Sorry, I didn't mean to suggest deleting all of them everywhere. I think the various threads on this PR made it hard to keep track of which thing was which. I'll post a standalone review that summarizes things that need to change to get this merged.

api/SVGFECompositeElement.json
```diff
@@ -57,7 +57,8 @@
               "version_added": "18"
             },
             "edge": {
-              "version_added": "12"
+              "version_added": "12",
+              "notes": "Before Edge 79, <code>BackgroundImage</code> and <code>BackgroundAlpha</code> were supported."
```
ddbeck, line 61: Remove this note (the one on `svg.elements.feComposite.in` covers it)

svg/elements/feComposite.json
```diff
@@ -295,81 +297,32 @@
...
-              }
-            },
-            "status": {
-              "experimental": false,
-              "standard_track": true,
-              "deprecated": false
-            }
-          }
-        },
-        "lighterForError": {
```
ddbeck, line 333: Three things at this point:

- reinstate `lighterForError` (again, sorry for the earlier confusion; I only meant for the `api` equivalent feature to be droppped)
- move `lighterForError` one level deeper, as a child to `operator`
- rename `lighterForError` to `lighter` and rewrite the description to `<code>lighter</code> value`

api/SVGFECompositeElement.json
```diff
@@ -104,7 +105,8 @@
...
```
ddbeck, line 109: Remove this note (the one on `svg.elements.feComposite.in2` covers it)

api/SVGFECompositeElement.json
```diff
@@ -57,7 +57,8 @@
               "version_added": "18"
```
ddbeck, line 57: I just noticed that a few lines above this, that this feature is erroneously named `in1` instead of `in`.

ddbeck (changes requested):
Here's a summary of what needs to change before we merge this:

## API features

- Remove the notes (new in this diff) from the `api` features
- Rename `in1` to `in` (to match the name of the element's attributes and the spec—this is new and I only just noticed it)

## SVG features

- Reinstate `lighterForError` (SVG only, not API)
- Move `lighterForError` one level deeper, as a child to `operator`
- Rename `lighterForError` to `lighter` and rewrite the description to `<code>lighter</code> value`

hamishwillee:
Thanks for the clear direction @ddbeck . I'm pretty sure this captures it.

chrisdavidmills:
@ddbeck this one is mentioned in the 86 rel notes: https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/86#svg

Can you have a quick look at @hamishwillee 's latest fixes and get it merged if you are happy? Thanks!

ddbeck (approved):
Thank you! 🎉

foolip:
Sorry to revive an old PR, but just a note that I found the Chrome, Edge and Safari versions here to be incorrect and fixed them in https://github.com/mdn/browser-compat-data/pull/16278. For Chromium and WebKit, it appears that commits were mapped to versions in an incorrect way. The Chrome mistake is probably due to https://github.com/mdn/browser-compat-data/issues/7844, but I don't know what happened for WebKit.

Issue #8946, opened 2021-02-03
api.Notification.data - Chrome for Android should be marked as supporting the feature


fpetitit:
#### What information was incorrect, unhelpful, or incomplete?

The page mentions that Chrome for Android does not support Notification.data, but it does.

#### What did you expect to see?

Chrome for Android should be marked as supporting the feature.

#### Did you test this? If so, how?

Yes, on a Motorola E4 device using Android 7.1.1 and Chrome 88 (last version at the moment).

<details>
<summary>MDN page report details</summary>

* Query: `api.Notification.data`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/Notification/data
* Report started: 2021-02-03T13:08:36.870Z

</details>

ddbeck:
Thanks for reporting this, @fpetitit.

[This Chrome Platform Status entry](https://chromestatus.com/feature/5736434757533696) says this shipped in Chrome 44 and testing confirms that this is true. This feature seems to have appeared on all the Chromiums simultaneously.

To fix this, the `data` feature in [`Notification.json`](https://github.com/mdn/browser-compat-data/blob/master/api/Notification.json#L329-L376) needs to be updated to show that it was added in Chrome 44, Chrome for Android 44, WebView 44, Opera 31, and Opera for Android 32.

I'd welcome a PR from you or another contributor that makes this change. Thank you again!

fpetitit:
PR created : https://github.com/mdn/browser-compat-data/pull/9077

ddbeck:
Fixed by #9077.

Issue #7923, opened 2020-12-16
api.Notification.Notification - Chrome Android is not supported


O-Ami:
#### What information was incorrect, unhelpful, or incomplete?
Notification constructor is not supported on Chrome Android. It throws an error `Failed to construct 'Notification': Illegal constructor. Use ServiceWorkerRegistration.showNotification() instead.` [According to google developer doc](https://developers.google.com/web/updates/2015/05/notifying-you-of-changes-to-notifications#android_notifications) and [Chrome Issue Tracker](https://developer.mozilla.org/en-US/docs/Web/API/Notification/Notification#Browser_compatibility), this was never supported on android. 

#### What did you expect to see?
On Browser compatibility for `Notification constructor` ([Link](https://developer.mozilla.org/en-US/docs/Web/API/Notification/Notification#Browser_compatibility)), it should state `No` for Chrome Android

#### Did you test this? If so, how?
I've tested `new Notification('test')` on Chrome Android and it throws the above error. It is working on desktop.

<details>
<summary>MDN page report details</summary>

* Query: `api.Notification.Notification`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/Notification/Notification
* Report started: 2020-12-16T16:03:07.208Z

</details>

ddbeck:
Thanks for opening this issue, @O-Ami. You're right that the data is off here. It's also unsupported in WebView (listed in [`not-webview-exposed.txt`](https://source.chromium.org/chromium/chromium/src/+/master:android_webview/tools/system_webview_shell/test/data/webexposed/not-webview-exposed.txt)).

I'd welcome a PR (from you or another contributor), setting the `chrome_android` and `webview_android` values to `false` throughout [`Notification.json`](https://github.com/mdn/browser-compat-data/blob/master/api/Notification.json).

ddbeck:
#8113 fixed this for Chrome for Android. Leaving the derivative browsers (webview_android and likely opera_android and samsung_internet, now that I think of it.)

ddbeck:
OK, I goofed here, which I discovered while reviewing #9077. I misunderstood the linked documentation. The story with `Notification` on Android appears to be more complex than I first realized. Here's the main story:

1. WebView does not support notifications. At all. See [bug 551446](https://bugs.chromium.org/p/chromium/issues/detail?id=551446).
[...]

PR #9077, opened 2021-02-12
api.Notification.data - Some browsers does support feature

fpetitit:
Cf. issue https://github.com/mdn/browser-compat-data/issues/8946

-> the data feature in Notification.json needs to be updated to show that it was added in Chrome 44, Chrome for Android 44, Opera 31, and Opera for Android 32.

Official documentation listing the availability of this feature for these browsers : https://chromestatus.com/feature/5736434757533696 (but it seems that Android WebView still does not support Notifications : https://bugs.chromium.org/p/chromium/issues/detail?id=551446)

### DIFF
--- a/api/Notification.json
+++ b/api/Notification.json
@@ -15,7 +15,7 @@
             }
           ],
           "chrome_android": {
-            "version_added": "44"
+            "version_added": true
           },
           "edge": {
             "version_added": "14"
--- a/api/Notification.json
+++ b/api/Notification.json
@@ -15,7 +15,7 @@
             }
           ],
           "chrome_android": {
-            "version_added": false
+            "version_added": "44"
           },
           "edge": {
             "version_added": "14"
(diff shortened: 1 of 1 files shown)


api/Notification.json
```diff
@@ -15,7 +15,7 @@
             }
           ],
           "chrome_android": {
-            "version_added": false
+            "version_added": "44"
```
ddbeck, line 18: Thanks to your PR, I discovered that the whole `api.Notification` data for Chrome for Android needs some work. For now, I'd ask you to you to change this to `true` (see https://github.com/mdn/browser-compat-data/issues/7923#issuecomment-778359481 if you want all the detail, but it's not directly relevant to this PR).

```suggestion
            "version_added": true
```

ddbeck (changes requested):
Thanks for this PR, @fpetitit. I have a suggestion for you that doesn't actually have anything to do with this PR, but it would be helpful to deal with some other issues. After that, we can merge this one. Thank you!

ddbeck (approved):

PR #9078, opened 2021-02-12
Move "masonry" under "grid-template-rows" so it is treated as a value rather than a property

Fyrd:
Currently the "masonry" value of `grid-template-rows` appears as a sibling of `grid-template-rows` rather than a child. As a result it is interpreted on caniuse (and perhaps other places) as its own property, rather than a value.

This PR moves it as a child of `grid-template-rows` to correct this issue.

### DIFF
--- a/css/properties/grid-template-rows.json
+++ b/css/properties/grid-template-rows.json
@@ -224,6 +224,62 @@
             }
           }
         },
+        "masonry": {
+          "__compat": {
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Nice catch. Thank you! 🎉

GPHemsley:
Should this change also have been made to `grid-template-columns`?

(Original PR was #7249.)

ddbeck:
@GPHemsley Good catch. I've opened https://github.com/mdn/browser-compat-data/pull/14562 to fix this.

PR #9080, opened 2021-02-13
Mirror Safari -> Safari iOS for HTMLElement, HTMLLIElement, ImageData

vinyldarkscratch:
This PR performs some mirroring of Safari to Safari iOS versions for three APIs (`HTMLElement`, `HTMLLIElement`, and `ImageData`).


foolip (approved):
These seem unlikely to differ between platforms.

PR #9083, opened 2021-02-13
Add Edge versions for AudioParam API

vinyldarkscratch:
This PR adds real values for Microsoft Edge for the `AudioParam` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.3.1).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/AudioParam

### DIFF
--- a/api/AudioParam.json
+++ b/api/AudioParam.json
@@ -58,7 +58,7 @@
               "version_added": "68"
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "79"
             },
             "firefox": {
               "version_added": false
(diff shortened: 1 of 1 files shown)


foolip (changes requested):
I think https://mdn-bcd-collector.appspot.com/tests/api/AudioParam is probably a broken test, the results are all false in both current Chrome and in Edge 18.

foolip:
I fixed these tests in https://github.com/foolip/mdn-bcd-collector/pull/973, which should be deployed soon.

vinyldarkscratch:
Looks like the results are the same for this PR as in collector v1.3.3 (which has the updated tests).

foolip (approved):
I've looked at https://mdn-bcd-collector.appspot.com/tests/api/AudioParam in Edge 18 and 80 and indeed this looks right.

PR #9129, opened 2021-02-15
Note Window.name privacy behavior in Firefox 86+

Rumyra:
- [x] Summarize your changes

Updating window.name note as it's no longer behind pref

https://bugzilla.mozilla.org/show_bug.cgi?id=1685089

Works in nightly for me (which I realise is 87 but that's enough to confirm for me it happened in 86 as per bug)

Tests pass

https://github.com/mdn/content/issues/1730


chrisdavidmills (changes requested):
Hi @Rumyra , just providing a drive-by review of this one. Mostly good, just a couple of bits:

1. Why did you remove the `<code>` elements from `window.name`?
2. If it is enabled in 86, surely it is available behind a pref from 82-85?
3. If that is the case, I think it would be better to have two separate entries for this one
   * One that just shows `version_added` as 86, with nothing else.
   * One that shows `varion_added` as 82, and `version_removed` as 85, with the flag information included in the proper flag structure rather than a note (see https://developer.mozilla.org/en-US/docs/MDN/Structures/Compatibility_tables#including_browser_preferences_or_flags)

Have I got this right, @ddbeck ?

ddbeck:
Yes, I think that's on the right track, @chrisdavidmills.

I'd probably retain the note on the second support statement though. Together, it'd look something like this:

```
{
  "version_added": "87",
},
{
  "version_added": "82",
  "version_removed": "87",
  "flags": [
    "type": "preference",
    "name": "privacy.window.name.update.enabled",
    "value_to_set": "true"
  ],
  "notes": "From Firefox 86, if a new page from another domain…"
}
```

I'd also like to see the code formatting reinstated for the note text that remains.

chrisdavidmills:
Perfect, thanks a lot @ddbeck !

Rumyra:
Thank you so much both!

My query about this is `window.name` has been around since v1, only the part about it not persisting has been modified, set behind a pref (82) and then shipped (86). Before I do another PR for this, I want to check this is ok:

```
{
  "version_added": "86",
},
{
  "version_added": "82",
  "version_removed": "86",
  "flags": [
    "type": "preference",
    "name": "privacy.window.name.update.enabled",
    "value_to_set": "true"
  ],
  "notes": "From Firefox 86, if a new page from another domain…"
},
{
  "version_added": "1",
  "version_removed": "82"
// possible note here?
},
```

And with `<code>` back in - apologies, I think that came out from the discussion in the issue #1730

api/Window.json
```diff
@@ -4525,11 +4525,11 @@
             },
             "firefox": {
               "version_added": "1",
-              "notes": "From Firefox 82, if a new page from another domain is loaded into a tab, then <code>window.name</code> is set to the empty string (the original string is restored if the tab reverts back to the original page). This change prevents some cross-site attacks. See <a href='https://bugzil.la/444222'>bug 444222</a>."
+              "notes": "From Firefox 86, if a new page from another domain is loaded into a tab, then window.name is set to an empty string (the original string is restored if the tab reverts back to the original page). This change prevents some cross-site attacks. From Firefox 82 to 86 this feature can enabled using the preference privacy.window.name.update.enabled. See <a href='https://bugzil.la/1685089'>bug 1685089</a> and <a href='https://bugzil.la/444222'>bug 444222</a>."
```
ddbeck, line 4528:
```suggestion
              "notes": "Before Firefox 86, if a new page from another domain is loaded into a tab, then <code>window.name</code> is not set to an empty string, which can allow some cross-site attacks. See <a href='https://bugzil.la/1685089'>bug 1685089</a> and <a href='https://bugzil.la/444222'>bug 444222</a>."
```

api/Window.json
```diff
@@ -4525,11 +4525,11 @@
...
             },
             "firefox_android": {
               "version_added": "4",
-              "notes": "From Firefox 82, if a new page from another domain is loaded into a tab, then <code>window.name</code> is set to the empty string (the original string is restored if the tab reverts back to the original page). This change prevents some cross-site attacks. See <a href='https://bugzil.la/444222'>bug 444222</a>."
+              "notes": "From Firefox 86, if a new page from another domain is loaded into a tab, then window.name is set to an empty string (the original string is restored if the tab reverts back to the original page). This change prevents some cross-site attacks. From Firefox 82 to 86 this feature can enabled using the preference privacy.window.name.update.enabled. See <a href='https://bugzil.la/1685089'>bug 1685089</a> and <a href='https://bugzil.la/444222'>bug 444222</a>."
```
ddbeck, line 4532:
```suggestion
              "notes": "Before Firefox 86, if a new page from another domain is loaded into a tab, then <code>window.name</code> is not set to an empty string, which can allow some cross-site attacks. See <a href='https://bugzil.la/1685089'>bug 1685089</a> and <a href='https://bugzil.la/444222'>bug 444222</a>."
```

ddbeck (changes requested):
@Rumyra I'm really sorry. I've badly misled you here and I'm grateful that you caught the error. I got hung up on the flag and misread the original note text. I thought it was a pref for the whole feature, not the narrow behavior change. 🤦‍♂️

I went back and reread the bug and some links there. The fact that [the spec changed](https://github.com/whatwg/html/pull/5871) to make this standard behavior makes me think we should come at this from a slightly different angle: alert web developers to the old behavior, not the new one. In other words, we can pretend that Firefox historically did something weird and that the new behavior is correct, especially since [it seems likely that Chrome will implement this change too](https://bugs.chromium.org/p/chromium/issues/detail?id=706350). This also suggests eliding the flag entirely—it's more like an implementation detail of a bug fix than the introduction of a new feature.

In a line comment, I've shown how I'd handle this, now that I understand better, though I'm open to other phrasing, if you think it's unclear.

Finally, thanks for being patient with me on this one!

api/Window.json
Rumyra: This looks great to me @ddbeck and please don't be sorry! I feel I learn something new every time I do a BCD PR ☺️ and am very grateful for your continued feedback!

I wholeheartedly agree with this approach, esp considering Chrome are very likely to push this soon too.

Also I assume then that we don't need to mention it being behind a pref from 82 - 86?

ddbeck: > Also I assume then that we don't need to mention it being behind a pref from 82 - 86?

Yes, I think so. If it hadn't shipped yet and the shipping version was indeterminate, I might suggest a mentioning it. But it'll be a historical footnote starting tomorrow (which will be before this data even gets into a BCD release).

ddbeck (approved):
Thank you! 🎉

PR #9130, opened 2021-02-15
Add <form rel> initial compat data

kiding:
- [x] Summarize your changes
    Adding `<form rel>` compat data as per the revised spec: https://github.com/whatwg/html/pull/4332.
    - Mozilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1509346
    - Chromium: https://bugs.chromium.org/p/chromium/issues/detail?id=931209
    - WebKit: https://bugs.webkit.org/show_bug.cgi?id=194398
    - No browser has implemented this feature as far as I know.
    - Tested with Firefox 85, Chrome 88, Safari 14, and IE 11.
    Fell into this rabbit hole while preparing https://github.com/mdn/content/pull/2236. Should follow up there once this data is available.

### DIFF
--- a/html/elements/form.json
+++ b/html/elements/form.json
@@ -473,6 +473,53 @@
             }
           }
         },
+        "rel": {
+          "__compat": {
+            "support": {
+              "chrome": {
+                "version_added": false
+              },
+              "chrome_android": {
+                "version_added": false
+              },
+              "edge": {
+                "version_added": false
+              },
+              "firefox": {
+                "version_added": false
+              },
+              "firefox_android": {
+                "version_added": false
+              },
+              "ie": {
+                "version_added": false
+              },
+              "opera": {
+                "version_added": false
+              },
+              "opera_android": {
+                "version_added": false
+              },
+              "safari": {
+                "version_added": false
+              },
+              "safari_ios": {
+                "version_added": false
+              },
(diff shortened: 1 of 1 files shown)


ddbeck:
Hi @kiding. Thanks for opening this PR. I see that it's your first on the project—welcome!

This looks pretty good, but there's a blocker here: we've been avoiding adding features until there's some evidence of implementation. Reading over the linked bugs, I think it's a little too early to bring this into BCD. I'd be happy to see the return of this data in the future, as soon as one or more browser vendors start implementing it.

Thanks for your understanding!

kiding:
Thanks for the reply, @ddbeck! 

While I’m not entirely on board with the guideline—especially in that applying it to WHATWG-standardized features might be a bit too much—I more or less understand the rationale.

Let me amend mdn/content to reduce confusion instead, then come back for this PR when any implementation effort surfaces. Thank you!

ddbeck:
@kiding I agree that the guideline isn't ideal, but it's an unfortunate practicality for the time being. We're working to link features to specifications, which might permit us some more nuance when that work is complete (e.g., allowing any feature in with recognized spec, while requiring implementations for non-standard features). Thanks again for your understanding on this and I'm looking forward to the eventual return of this PR. 😃

PR #9135, opened 2021-02-16
feImage note: doesn't work with doc fragments on Firefox

hamishwillee:
Firefox does not support document fragments in `feImage` - as per bug https://bugzilla.mozilla.org/show_bug.cgi?id=455986

What this means is that you can reference a png, svg or other document from an external source, but not a fragment - e.g. a path or circle you define in the doc. It's an irritating limitation.

### DIFF
--- a/svg/elements/feImage.json
+++ b/svg/elements/feImage.json
@@ -62,7 +62,7 @@
               },
               "firefox": {
                 "version_added": true,
-                "notes": "Document fragments (including references to fragments in current document) are not supported (see <a href='https://bugzil.la/455986'>bug 455986</a>)."
+                "notes": "Document fragments (including references to fragments in the current document) are not supported (see <a href='https://bugzil.la/455986'>bug 455986</a>)."
               },
               "firefox_android": {
                 "version_added": true,
--- a/svg/elements/feImage.json
+++ b/svg/elements/feImage.json
@@ -66,7 +66,7 @@
               },
               "firefox_android": {
                 "version_added": true,
-                "notes": "Document fragments (including references to fragments in current document) are not supported (see <a href='https://bugzil.la/455986'>bug 455986</a>)."
+                "notes": "Document fragments (including references to fragments in the current document) are not supported (see <a href='https://bugzil.la/455986'>bug 455986</a>)."
               },
               "ie": {
                 "version_added": true
--- a/svg/elements/feImage.json
+++ b/svg/elements/feImage.json
@@ -159,7 +159,7 @@
               },
               "firefox": {
                 "version_added": "4",
-                "notes": "Document fragments (including references to fragments in current document) are not supported (see <a href='https://bugzil.la/455986'>bug 455986</a>)."
+                "notes": "Document fragments (including references to fragments in the current document) are not supported (see <a href='https://bugzil.la/455986'>bug 455986</a>)."
               },
               "firefox_android": {
                 "version_added": true,
--- a/svg/elements/feImage.json
+++ b/svg/elements/feImage.json
@@ -156,10 +156,12 @@
                 "version_added": "12"
               },
               "firefox": {
-                "version_added": "4"
+                "version_added": "4",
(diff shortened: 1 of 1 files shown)


ddbeck:
@hamishwillee Is there any help we can give here, to get this out of draft status?

hamishwillee:
@ddbeck Good to review now. 

An "associated" question though. I have documented this FF limitation for the `feImage`, but this also affects [SVGFEImageElement](https://developer.mozilla.org/en-US/docs/Web/API/SVGFEImageElement), which inherits the same href and xlink:href values from [SVGURIReference](https://developer.mozilla.org/en-US/docs/Web/API/SVGURIReference). So the question is, how should this be documented in SVGFEImageElement/SVGURIReference given that SVGURIReference is kind of a mixin? Or should this just be ignored because SVGFEImageElement reflects feImge?

ddbeck:
> how should this be documented in SVGFEImageElement/SVGURIReference given that SVGURIReference is kind of a mixin? Or should this just be ignored because SVGFEImageElement reflects feImge?

Let's just ignore it. Two points in favor doing so:

1. It's reflection, so I think it's fair to regard the noted issue is a problem of the underlying element rather than the API.
2. For `SVGURIReference` specifically, the data will soon go away. It is a true mixin and it's on the list of interfaces to demix in https://github.com/mdn/browser-compat-data/issues/8929#issuecomment-778502425.

I'll start my actual review now. 😃

svg/elements/feImage.json
```diff
@@ -61,10 +61,12 @@
                 "version_added": "12"
               },
               "firefox": {
-                "version_added": true
+                "version_added": true,
+                "notes": "Document fragments (including references to fragments in current document) are not supported (see <a href='https://bugzil.la/455986'>bug 455986</a>)."
```
ddbeck, line 65: nit: I think it reads better with a definite article here. Sorry, this is a very small one

```suggestion
                "notes": "Document fragments (including references to fragments in the current document) are not supported (see <a href='https://bugzil.la/455986'>bug 455986</a>)."
```

svg/elements/feImage.json
```diff
@@ -61,10 +61,12 @@
...
               },
               "firefox_android": {
-                "version_added": true
+                "version_added": true,
+                "notes": "Document fragments (including references to fragments in current document) are not supported (see <a href='https://bugzil.la/455986'>bug 455986</a>)."
```
ddbeck, line 69:
```suggestion
                "notes": "Document fragments (including references to fragments in the current document) are not supported (see <a href='https://bugzil.la/455986'>bug 455986</a>)."
```

svg/elements/feImage.json
```diff
@@ -156,10 +158,12 @@
                 "version_added": "12"
               },
               "firefox": {
-                "version_added": "4"
+                "version_added": "4",
+                "notes": "Document fragments (including references to fragments in current document) are not supported (see <a href='https://bugzil.la/455986'>bug 455986</a>)."
```
ddbeck, line 162:
```suggestion
                "notes": "Document fragments (including references to fragments in the current document) are not supported (see <a href='https://bugzil.la/455986'>bug 455986</a>)."
```

svg/elements/feImage.json
```diff
@@ -156,10 +158,12 @@
...
               },
               "firefox_android": {
-                "version_added": true
+                "version_added": true,
+                "notes": "Document fragments (including references to fragments in current document) are not supported (see <a href='https://bugzil.la/455986'>bug 455986</a>)."
```
ddbeck, line 166:
```suggestion
                "notes": "Document fragments (including references to fragments in the current document) are not supported (see <a href='https://bugzil.la/455986'>bug 455986</a>)."
```

ddbeck (changes requested):
A little idea before merging in line comments. Thank you, @hamishwillee!

hamishwillee:
Thanks @ddbeck . Makes sense. I merged your change.

ddbeck (approved):
Thank you! 🎉

PR #9137, opened 2021-02-16
Replace ranged Edge ≤79 versions with 79 for Web Bluetooth

foolip:
These APIs were not supported in EdgeHTML. This is a simple
search-replace and leaves many issues with the data.


ddbeck (approved):
This seems reasonable to me. 👍

PR #8113, opened 2020-12-22
api.Notification is unsupported in Chrome for Android

joneskj55:
This PR addresses issue #7923, updating chrome_android data.

PR #9139, opened 2021-02-16
Revert changes that marked api.Notification as unsupported in Chrome for Android

ddbeck:
In #7923, I wrongly proposed marking `api.Notification` as unsupported in Chrome for Android, which was carried out in #8113. In fact, the interface is only unsupported in WebView on Android. See https://github.com/mdn/browser-compat-data/issues/7923#issuecomment-778359481 for a bit more detail about why I got it wrong.

This reverts PR #8113 / commit d2ad54011d6cc672a42d1f748faaf3f0acb565fc. We still need to properly test for this interface because the story is so strange; see https://github.com/mdn/browser-compat-data/issues/7923#issuecomment-778361043 and the response to it for more on that question.


foolip:
@jpmedley would you be able to review this? This situation seems complicated and I'm not very familiar with this API.

ddbeck:
I just fixed the merge conflict, but this data continues to be a problem (e.g., see https://github.com/mdn/browser-compat-data/issues/9273)

foolip (approved):

PR #9168, opened 2021-02-18
Update Pointer Events for Firefox for Android

chrisdavidmills:
The compat data for [Pointer Events](https://developer.mozilla.org/en-US/docs/Web/API/Pointer_events) on Firefox for Android has gotten very out of date. This pull request updates all the relevant data points.

Note that this PR is part of the required work to close https://github.com/mdn/browser-compat-data/issues/4983. It would be great if we could chip away at it together until we can close it.

Note also that 79 was the chosen support version, as discussed in https://github.com/mdn/browser-compat-data/issues/4983#issuecomment-669943788

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -4573,7 +4573,7 @@
             ],
             "firefox_android": [
               {
-                "version_added": false
+                "version_added": "79"
               },
               {
                 "version_added": "41",
@@ -6414,7 +6414,7 @@
             ],
             "firefox_android": [
               {
-                "version_added": false
+                "version_added": "79"
               },
               {
                 "version_added": "41",
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -1475,17 +1475,22 @@
                 ]
               }
             ],
-            "firefox_android": {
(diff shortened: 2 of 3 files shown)


api/Element.json
```diff
@@ -4573,7 +4573,7 @@
             ],
             "firefox_android": [
               {
-                "version_added": false
+                "version_added": "79"
```
ddbeck, line 4576: A recurring question I've got: a lot (all?) of these features land on Desktop in 59, but in 79 on Android. Is it right that Fennec never enabled Pointer Events by default, but Fenix did?

chrisdavidmills: Yeah, the features got enabled at different times. they eventually got enabled for Fenix in version 70: see https://bugzilla.mozilla.org/show_bug.cgi?id=1507495, but Fenix was not publicly released until 79.

ddbeck: OK, great. Thank you!

ddbeck (approved):
Thank you, @chrisdavidmills! 🎉

PR #9169, opened 2021-02-18
Remove irrelevant flag data and mixin notes for WindowOrWorkerGlobalScope

foolip:
The flag data can be removed under the irrelevant flag data guideline:
https://github.com/mdn/browser-compat-data/blob/master/docs/data-guidelines.md#removal-of-irrelevant-flag-data

The notes say "now defined on WindowOrWorkerGlobalScope mixin" which
does not communicate any information. If some of these things were
previously not exposed in workers (or window) then that should be noted
directly.


Elchi3 (approved):
Looking forward to having all this a lot clearer. Thanks a lot!

PR #9182, opened 2021-02-19
Update Edge versions for AnimationTimeline API

vinyldarkscratch:
This PR updates and corrects the real values for Microsoft Edge for the `AnimationTimeline` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.3.1).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/AnimationTimeline

### DIFF
--- a/api/AnimationTimeline.json
+++ b/api/AnimationTimeline.json
@@ -11,7 +11,7 @@
             "version_added": false
           },
           "edge": {
-            "version_added": "≤79"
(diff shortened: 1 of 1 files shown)


foolip (approved):
The Chrome data also needs updating, but this seems correct.

PR #9185, opened 2021-02-19
Add @property descriptors

rachelandrew:
These are the descriptors for `"property`. Spec: https://drafts.css-houdini.org/css-properties-values-api-1/#at-property-rule

I'm going to document them for: https://github.com/mdn/content/issues/1884


ddbeck (changes requested):
Other descriptors are included as subfeatures of their corresponding at-rule (e.g. `css.properties.font-face.src` for the `src` descriptor). Could you move these features to that part of the tree?

css/at-rules/property.json
```diff
@@ -48,6 +48,153 @@
             "standard_track": true,
             "deprecated": false
           }
+        },
+        "inherits": {
+          "__compat": {
+            "description": "inherits",
```
ddbeck, line 54: suggestion: if we're going to have description, we ought to expand on the the key for this feature

```suggestion
            "description": "<code>inherits</code> descriptor",
```

css/at-rules/property.json
```diff
@@ -48,6 +48,153 @@
...
+            "status": {
+              "experimental": true,
+              "standard_track": true,
+              "deprecated": false
+            }
+          }
+        },
+        "initial-value": {
+          "__compat": {
+            "description": "initial-value",
```
ddbeck, line 103:
```suggestion
            "description": "<code>initial-value</code> descriptor",
```

css/at-rules/property.json
```diff
@@ -48,6 +48,153 @@
...
+            "status": {
+              "experimental": true,
+              "standard_track": true,
+              "deprecated": false
+            }
+          }
+        },
+        "syntax": {
+          "__compat": {
+            "description": "syntax",
```
ddbeck, line 152:
```suggestion
            "description": "<code>syntax</code> descriptor",
```

ddbeck (changes requested):
A few small suggestions in line comments. Thank you!

ddbeck (approved):
Thank you! 🎉

PR #9186, opened 2021-02-19
Add Edge versions for CacheStorage API

vinyldarkscratch:
This PR adds real values for Microsoft Edge for the `CacheStorage` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.3.3).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/CacheStorage


foolip (approved):
Makes perfect since looking at https://developer.mozilla.org/en-US/docs/Web/API/CacheStorage#browser_compatibility

PR #9189, opened 2021-02-19
Add Edge versions for CSSPropertyRule API

vinyldarkscratch:
This PR adds real values for Microsoft Edge for the `CSSPropertyRule` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.3.3).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/CSSPropertyRule

### DIFF
--- a/api/CSSPropertyRule.json
+++ b/api/CSSPropertyRule.json
@@ -11,7 +11,7 @@
             "version_added": "78"
           },
           "edge": {
-            "version_added": "≤79"
+            "version_added": "85"
           },
           "firefox": {
             "version_added": false
@@ -58,7 +58,7 @@
               "version_added": "78"
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "85"
             },
             "firefox": {
               "version_added": false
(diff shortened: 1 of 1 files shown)


api/CSSPropertyRule.json
```diff
@@ -11,7 +11,7 @@
             "version_added": "78"
           },
           "edge": {
-            "version_added": "≤79"
+            "version_added": "85"
```
foolip, line 14: It's surprising if this was enabled in Chrome 78 but not until Edge 85. Is the Chrome data incorrect, or what's going on?

foolip: The answer is that https://mdn-bcd-collector.appspot.com/tests/api/CSSPropertyRule also starts returning true in Chrome 85.

foolip: OK, I've commented in https://github.com/mdn/browser-compat-data/pull/6278#issuecomment-782663612. I'll merge this PR and send another to fix the rest.

PR #9192, opened 2021-02-19
Remove irrelevant flags in Firefox for Animation API

vinyldarkscratch:
This PR removes irrelevant flag data for Firefox and Firefox Android for the `Animation` API as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/master/docs/data-guidelines.md#removal-of-irrelevant-flag-data).

### DIFF
--- a/api/Animation.json
+++ b/api/Animation.json
@@ -27,38 +27,12 @@
           "edge": {
             "version_added": "79"
           },
-          "firefox": [
-            {
-              "version_added": "48"
-            },
-            {
-              "version_added": "46",
-              "version_removed": "48",
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "dom.animations-api.core.enabled",
-                  "value_to_set": "true"
-                }
-              ]
-            }
-          ],
-          "firefox_android": [
-            {
-              "version_added": "48"
-            },
-            {
-              "version_added": "46",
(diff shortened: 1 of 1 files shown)


api/Animation.json
```diff
@@ -1599,58 +1161,20 @@
...
-                "notes": "This property is supported in Firefox 48 but is read-only. It became writable in Firefox 49."
-              }
-            ],
+            "firefox": {
+              "version_added": "75",
+              "notes": "Currently only the getter is supported"
+            },
+            "firefox_android": {
+              "version_added": "63",
+              "flags": [
```
foolip, line 1170: Is the existing data really correct here, or was the Firefox entry updated at some point while forgetting the Firefox for Android entry?

saschanaz: I suspect it's the latter. https://bugzilla.mozilla.org/show_bug.cgi?id=1619178

foolip: OK, so let's merge this PR and then I'll send a separate one to fix this.

foolip (approved):

api/Animation.json
foolip: I've sent https://github.com/mdn/browser-compat-data/pull/9436.

PR #9195, opened 2021-02-19
Remove KeyboardEvent.DOM_KEY_* constants

foolip:
These should be removed per the recent guideline on the matter:
https://github.com/mdn/browser-compat-data/blob/master/docs/data-guidelines.md#constants

These appear to be the last such constants in BCD.


Elchi3 (approved):
The mdn_urls also don't exist (we won't create pages for each constant either)

PR #9199, opened 2021-02-19
Add data for Web Serial API.

jpmedley:
https://www.chromestatus.com/feature/6577673212002304

### DIFF
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -2384,7 +2384,7 @@
       },
       "serial": {
         "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/Navigator/serviceWorker",
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/Navigator/serial",
           "support": {
             "chrome": {
               "version_added": "89"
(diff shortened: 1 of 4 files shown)


api/Navigator.json
```diff
@@ -2382,6 +2382,54 @@
           }
         }
       },
+      "serial": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/Navigator/serviceWorker",
```
foolip, line 2387: This URL and that in `WorkerNavigator` needs to be updated.

jpmedley: Fixed. PTAL

api/WorkerNavigator.json
```diff
@@ -144,6 +144,54 @@
           }
         }
       },
+      "serial": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/Navigator/serviceWorker",
```
foolip, line 149: This URL also needs updating.

foolip:
@jpmedley I tried verifying this with collector results but then noticed it's Chrome 89 which hasn't been released yet. Would you like to merge this before Chrome 89 is released?

api/WorkerNavigator.json
jpmedley: Fixed. PTAL

PR #9206, opened 2021-02-21
Remove html.elements.script.version_parameter entry

foolip:
Firefox 59 was released 2018-03-13, so this can be removed:
https://github.com/mdn/browser-compat-data/blob/master/docs/data-guidelines.md#removal-of-irrelevant-features


Elchi3 (approved):
Thanks for removing irrelevant content!

PR #9222, opened 2021-02-22
FF87 removes the non-standard caption-side values

rachelandrew:
Working on https://github.com/mdn/content/issues/2502

The non-standard values of `caption-side` are being removed. I don't know what the policy is with something which is now in no browsers, but did have support?

### DIFF
--- a/css/properties/caption-side.json
+++ b/css/properties/caption-side.json
@@ -62,10 +62,12 @@
                 "version_added": false
               },
               "firefox": {
-                "version_added": "1"
+                "version_added": "1",
+                "version_removed": "87"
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Thank you, @rachelandrew.

> I don't know what the policy is with something which is now in no browsers, but did have support?

You've handled this correctly. Eventually (in two years), we can drop this feature as irrelevant but, for now, we leave it in. The actual guideline is here: https://github.com/mdn/browser-compat-data/blob/master/docs/data-guidelines.md#removal-of-irrelevant-features.

PR #9224, opened 2021-02-22
Note Safari bug resolving CSS variables and url()

### LINKED ISSUE #9075: css.properties.custom-property - not supporting images in Safari
If I use image URLs in CSS variables, they are not working in Safari (This works fine in Chrome and Firefox)

Tested in Safari Version 14.0.3 (16610.4.3.1.4)
macOS Big Sur Version 11.2.1 

```
:root {
    --download-icon: url('../img/download.png');
}

@media (prefers-color-scheme: dark) {
    :root {
        --download-icon: url('../img/download_white.png');
    }
}

.document-link {
    background: var(--download-icon) no-repeat center;
}
```

<details>

* Query: `css.properties.custom-property`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/CSS/Using_CSS_custom_properties
* Report started: 2021-02-12T10:13:32.342Z

</details>

ddbeck:
In certain circumstances, Safari will incorrectly resolve `url()` values relative to the document instead of the stylesheet. It's a bit of an edge case, but there's two WebKit bugs and a BCD issue about it (#9075), so I figured that justified a note.

See for details about the bug itself and the fix: https://trac.webkit.org/changeset/267951/webkit

Fixes #9075

### DIFF
--- a/css/properties/custom-property.json
+++ b/css/properties/custom-property.json
@@ -83,10 +83,12 @@
               "version_added": "36"
             },
             "safari": {
-              "version_added": "9.1"
+              "version_added": "9.1",
+              "notes": "Shorthand property values containing <code>url()</code> and a CSS variable resolve the <code>url()</code> relative to the document root instead of the stylesheet. See <a href='https://webkit.org/b/198512'>bug 198512</a>."
             },
             "safari_ios": {
-              "version_added": "9.3"
+              "version_added": "9.3",
+              "notes": "Shorthand property values containing <code>url()</code> and a CSS variable resolve the <code>url()</code> relative to the document root instead of the stylesheet. See <a href='https://webkit.org/b/198512'>bug 198512</a>."
             },
             "samsunginternet_android": {
               "version_added": "5.0"


foolip:
Going by the timing the fix will be in Safari 14.1. Is there anything that will serve as a reminder to add another note-free entry when that is released?

ddbeck:
@foolip Hrm. We have a couple of options. We could speculatively rewrite the note as "Before Safari 14.1…" or I could leave the original issue open and consider this only a partial resolution. It feels sufficiently edge case-ish that we probably don't need two support statements here.

foolip:
It's a bit of an aside, but it sounds like we have the opposite preferences for how to deal with notes that aren't super important. This is a case where I think two support statements would hide away the note and mean that nobody needs to read it, while having it as a note on a single statement would make it appear to be relevant indefinitely.

(FWIW, I do think a lot of existing notes in BCD would also benefit from being hidden away.)

ddbeck:
Yeah, you're right. At least on MDN's current tables, there's two competing considerations:

1. If you use multiple support statements, it can create apparent discontinuities. For example, if you have two support statements, version 1-90 (with a note) and version 91+ (no note), then it hides the note, but it makes it appear as if support wasn't added until version 91.

2. If you use one support statement, notes add noise, even if the note applies to a very old version.

I come down on the first consideration being more important, but I've been in the second camp before too.

If the table were smarter, then we might not have this problem. It could report the greatest continuous range (1+), even if it were reported in multiple statements. Or it'd provide a useful summary ("Supported for 10+ years") and every feature would have an optional detailed timeline hidden away.

In this particular case, I'd be fine with deleting this note once Safari ships a fix.

foolip:
@ddbeck good point, I didn't think about the impact of showing "91" without a note when skimming the table without looking at any notes or expanding hidden statements.

foolip:
> In this particular case, I'd be fine with deleting this note once Safari ships a fix.

If we wait another week it will probably have shipped. Nevertheless I'll approve the PR so you can decide what to do :)

foolip (approved):

ddbeck:
(This is fixed in Safari 14.1, so I'm closing this without merging.)

PR #9241, opened 2021-02-24
Deprecate CSP plugin-types directive

sideshowbarker:
https://github.com/w3c/webappsec-csp/commit/e9895ed (https://github.com/w3c/webappsec-csp/issues/394) completely removed the `plugin-types` directive from the CSP spec. So this patch marks it `deprecated:true` and `standard_track:false`.

### DIFF
--- a/http/headers/content-security-policy.json
+++ b/http/headers/content-security-policy.json
@@ -842,8 +842,8 @@
               },
               "status": {
                 "experimental": false,
-                "standard_track": true,
-                "deprecated": false
+                "standard_track": false,
+                "deprecated": true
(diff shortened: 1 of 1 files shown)


Elchi3 (changes requested):
Thanks! Please remove the bug reference for Firefox. I don't think it will be implemented anymore.

sideshowbarker:
> Thanks! Please remove the bug reference for Firefox. I don't think it will be implemented anymore.

OK — thanks, removed now

Elchi3 (approved):
👍

PR #9243, opened 2021-02-24
Update text-underline-offset.json

danilo-znamerovszkij:
Opera already supports text-underline-offset 

https://live.browserstack.com/dashboard#os=OS+X&os_version=Catalina&browser=Opera&browser_version=73.0&zoom_to_fit=true&full_screen=true&url=www.google.com&speed=1

tested locally and using the code pen from CSS tricks in Opera on Windows (BrowserStack) and locally with Opera on Mac:
https://css-tricks.com/almanac/properties/t/text-underline-offset/

### DIFF
--- a/css/properties/text-underline-offset.json
+++ b/css/properties/text-underline-offset.json
@@ -36,7 +36,7 @@
               "version_added": false
             },
             "opera": {
-              "version_added": true
+              "version_added": "73"
             },
             "opera_android": {
               "version_added": false
--- a/css/properties/text-underline-offset.json
+++ b/css/properties/text-underline-offset.json
@@ -36,7 +36,7 @@
               "version_added": false
             },
             "opera": {
-              "version_added": false
+              "version_added": true
             },
             "opera_android": {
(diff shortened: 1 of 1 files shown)


css/properties/text-underline-offset.json
```diff
@@ -36,7 +36,7 @@
               "version_added": false
             },
             "opera": {
-              "version_added": false
+              "version_added": true
```
ddbeck, line 39:
```suggestion
              "version_added": "73"
```

Setting this to 73 will satisfy the linter. I might have assumed this based on the Chrome to Opera version mapping, but the BrowserStack link was very helpful to confirm!

ddbeck (changes requested):
Hi @danilo-znamerovszkij and thanks for this PR. I see this is your first to BCD—welcome! I have one suggestion in a line comment, then we can get this merged. Thank you again!

ddbeck (approved):
Looks great. Thank you! 🎉

PR #9245, opened 2021-02-24
Merge Safari iOS 13.3 into 13

foolip:
As discussed when considering adding 14.3:
https://github.com/mdn/browser-compat-data/pull/8838#discussion_r563202188


Elchi3 (approved):
Since this removes 13.3 from the browser releases file, the build would fail I you haven't caught all instances of 13.3. It doesn't fail, so looks good to me.

PR #9250, opened 2021-02-24
Change Safari support from null to true based on tests

foolip:
This updates features where both Safaris had nulls, where
mdn-bcd-collector results from Safari 14 on macOS and on iOS 14 both
indicate support.

This is an appropriate category to update in bulk, since false positives
in the tests are quite uncommon, and as an extra safety the true values
will have to be revisited to determine the exact versions later.

### DIFF
--- a/api/ClipboardEvent.json
+++ b/api/ClipboardEvent.json
@@ -77,10 +77,10 @@
               "version_added": "43"
             },
             "safari": {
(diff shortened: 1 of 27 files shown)


Elchi3 (approved):
Thanks, I think this kind of bulk update is alright without much double checking.

PR #9251, opened 2021-02-24
Add note about CSS caret-color behavior on iOS

### LINKED ISSUE #9142: css.properties.caret-color - iOS Safari does not support this correctly
When testing the caret-color css on iOS Safari (14.2) the caret color does not change as expected, so it is not correctly supported by this browser.

In both the CSS Demo and Examples sections, when you select an input that has a custom color set, the caret color should reflect that.

Tested on iOS 14.2 Safari:
- Go to the Examples section near the bottom of the page.
- Tap in the first input field that uses the default caret: caret is blue
- Tap in the second input field with a custom caret color of red: caret is blue
- Tap in the paragraph with a custom caret color of green: caret is red
- Tap in the first input field that uses the default caret: caret is green

<details>

* Query: `css.properties.caret-color`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/CSS/caret-color
* Report started: 2021-02-16T17:05:24.086Z

</details>

ddbeck:
Fixes #9142, which attempts to describe the odd, buggy behavior of `caret-color` on iOS.

### DIFF
--- a/css/properties/caret-color.json
+++ b/css/properties/caret-color.json
@@ -33,7 +33,9 @@
               "version_added": "11.1"
             },
             "safari_ios": {
-              "version_added": "11.3"
+              "version_added": "11.3",
+              "partial_implementation": true,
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
LGTM! 👍

Issue #9144, opened 2021-02-16
api.GlobalEventHandlers.onselectstart - onselectstart not called on iOS


otacke:
#### What information was incorrect, unhelpful, or incomplete?

- The table lists iOS among the devices supporting the `onselectstart` handler. `onselectstart` does not to be supported by iOS 12 at least.
- That behavior seems to be known at least since 2016, cmp. https://gist.github.com/robertknight/ad9f1bca8280b3fa325817aa3ed9935c, so other iOS versions are likely to be affected, too. 

- CanIUse lists iOS 14 as minimum version, see https://caniuse.com/?search=onselectstart

#### What did you expect to see?
I expected the `onselectstart` handler to be called when starting to select text.

#### Did you test this? If so, how?
Yes. I used https://codepen.io/otacke/pen/wvodWBo to verify that the `onselectstart` handler is in fact not called.

<details>
<summary>MDN page report details</summary>

* Query: `api.Document.selectstart_event`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/Document/selectstart_event
* Report started: 2021-02-16T19:32:09.578Z

</details>

ddbeck:
Hi @toacke and thanks for raising this issue and the related #9143. Apparently `selectstart` has [a number](https://github.com/w3c/selection-api/issues/53) of [issues associated with it](https://bugs.webkit.org/showdependencytree.cgi?id=60424&hide_resolved=0).

A quick test shows that `onselectstart` does exist in the interface (e.g., `"onselectstart" in document` is `true`) but if it never fires fire, that's a bit misleading. I propose that we mark the feature as supported with a `partial_implementation` set to `true` and a note saying, "The `selectstart` event never fires and never invokes this handler."

To fix this, these need to be updated:

- `onselectstart` (partial implementation) in https://github.com/mdn/browser-compat-data/blob/e07c447ba01fd6fef734ab88cf40ba8ff55ce517/api/Document.json#L10758
- `onselectstart` (partial implementation) in https://github.com/mdn/browser-compat-data/blob/ee47fa52c6e57dd06b40a0e8f9aa072176226ab1/api/GlobalEventHandlers.json#L4155
- `selectstart` (`"version_added": false`) https://github.com/mdn/browser-compat-data/blob/e07c447ba01fd6fef734ab88cf40ba8ff55ce517/api/Document.json#L10758

I would welcome a PR from you (or another contributor) making these changes.

Thank you again!

[...]

PR #9256, opened 2021-02-24
Add notes about selectstart event handling on Safari iOS

otacke:
Add notes about `selectstart` event never firing on Safari iOS

Related issues: fixes https://github.com/mdn/browser-compat-data/issues/9143 and fixes https://github.com/mdn/browser-compat-data/issues/9144

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -10814,9 +10814,7 @@
             "safari_ios": {
               "version_added": false,
               "partial_implementation": true,
-              "notes": [
-                "The <code>selectstart</code> event never fires and never invokes this handler."
-              ]
+              "notes": "The <code>selectstart</code> event never fires and never invokes this handler."
             },
             "samsunginternet_android": {
               "version_added": true
--- a/api/Document.json
+++ b/api/Document.json
@@ -10812,9 +10812,7 @@
               "version_added": "1.3"
             },
             "safari_ios": {
-              "version_added": false,
-              "partial_implementation": true,
-              "notes": "The <code>selectstart</code> event never fires and never invokes this handler."
+              "version_added": false
             },
             "samsunginternet_android": {
               "version_added": true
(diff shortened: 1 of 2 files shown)


api/Document.json
```diff
@@ -10812,7 +10812,9 @@
               "version_added": "1.3"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": false,
+              "partial_implementation": true,
+              "notes": "The <code>selectstart</code> event never fires and never invokes this handler."
```
ddbeck, line 10817: Sorry, I messed up my explanation in the issues. This is my fault, completely.

Since we're setting this as `"version_added": false` we should drop the partial and note on this one. This is a subtlety in how we record data for events: the event itself `selectstart` never fires, so it's `false`, but the handler exists so it's truthy. Sorry for the confusion here.

```suggestion
              "version_added": false
```

ddbeck (changes requested):
One suggestion for you, @otacke (stemming from a confused comment I made on the original issue). Thank you!

api/Document.json
otacke: @ddbeck Thanks a lot again for contributing your time. I am fully aware that you'd have been quicker making the changes yourself and I hope that your investment will yield some return (me engaging more in MDN - have at least a "to do" labelled "read MDN contribution docs" on my list). Updated ...

otacke: @ddbeck Oh, could I have just clicked on "Commit suggestion" in order to add that to my pull-request? Just learned something new about github, I guess.

ddbeck: > you'd have been quicker making the changes yourself

Maybe, but I find I'm much less likely to do something foolish (or at least catch it before committing) if I have to explain it to someone else first. 😄

ddbeck (approved):
@otacke Thank you! 🎉

PR #9257, opened 2021-02-24
Add a script to query BCD by support in a specific release

foolip:
Exclude scripts/ from the published package so that none of this becomes
part of the public "API" for @mdn/browser-compat-data, yet.

### DIFF
--- a/utils/query2.js
+++ b/utils/query2.js
@@ -1,0 +1,160 @@
+'use strict';
+
+const assert = require('assert');
+const compareVersions = require('compare-versions');
+
+// Like compareVersions, but returns NaN if it's not possible to compare
+// the versions because one or both is ranged (starts with ≤). A ranged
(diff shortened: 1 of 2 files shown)


foolip:
@ddbeck @Elchi3 @sideshowbarker as discussed today, one of the scripts I had sitting around with tests and all.

I want this for https://github.com/foolip/mdn-bcd-collector/issues/875, where I want to compare the results of a specific browser release to the support in BCD, without having to deal with interpreting support statements.

foolip:
So after submitting this I realize that I never implemented an important part: finding a BCD entry using a path. @ddbeck is that one of the scripts you have sitting around? A basic one that works for only "x.y.z" is easy, but it would also be handy to support everything that `traverseFeatures` does but in a reusable form.

queengooborg:
This has been untouched for over two years.  Is there still interest in this?  If not, we should close it.

PR #9263, opened 2021-02-25
Remove notes about an error on chromestatus.com

foolip:
I have edited the chromestatus.com entry to fix the mistake.


Elchi3 (approved):
Always happy to see less notes :)

PR #9289, opened 2021-02-27
Add spec_url for CSS selectors G-O

sideshowbarker:


Elchi3 (approved):
Looks good to me. I haven't checked if all of these link to sections, but I think that's minor. These all seem to link to current specs and that matters to me.

PR #9297, opened 2021-02-28
Revert "Mark ConstantSourceNode as shipped in Edge 79 (not earlier)"

hugoglezcon:
Reverts mdn/browser-compat-data#9278


foolip:
@hugoglezcon can you say more about why this should be reverted?

ddbeck:
@foolip I'm closing this and the other revert PR. This user has been unresponsive in the past: https://github.com/mdn/browser-compat-data/pull/8676

PR #9299, opened 2021-02-28
Add notes about FF preferences that overridde CSS prefers-color-scheme

hamishwillee:
The [prefers-color-scheme](https://developer.mozilla.org/en-US/docs/Web/CSS/@media/prefers-color-scheme) media query has a Firefox specific note up the top about preference over-rides for this feature. That note should be in the BCD. 

image

This follows on from discussion in https://github.com/mdn/content/pull/2562#discussion_r580196154

### DIFF
--- a/css/at-rules/media.json
+++ b/css/at-rules/media.json
@@ -1285,7 +1285,11 @@
                 "version_added": "79"
               },
               "firefox": {
-                "version_added": "67"
+                "version_added": "67",
+                "notes": [
+                  "The feature can be overridden by Firefox preference <code>privacy.resistFingerprinting</code>. Set to <code>true</code> to always return <code>light</code>.",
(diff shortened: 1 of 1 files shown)


hamishwillee:
@ddbeck When you get a moment would be great if this could be sanity checked.

ddbeck:
I don't think either of these notes belong in BCD. I think it's a given that there will be lots of ways for a user to select their preferred color scheme and Firefox providing these particular methods is not a compatibility distinction worth mentioning. Moreover, I think these notes suggest that Firefox is doing something non-conforming, when it's not. [From the spec](https://drafts.csswg.org/mediaqueries-5/#prefers-color-scheme):

> The method by which the user expresses their preference can vary. It might be a system-wide setting exposed by the Operating System, or a setting controlled by the User Agent.

If the contents of these notes are genuinely useful to web developers, then I think the page content needs to address that in a more direct why, providing some context for relaying this information. Stuffing them into a note box on the page doesn't get at _why_ we're telling web developers about this, but neither does stuffing the same notes into the compat table. For instance, maybe these settings are useful tools for testing different color schemes; I'd much rather learn about that in the context of a section (or page) about testing `prefers-color-scheme`.

hamishwillee:
@ddbeck Thanks for your detailed answer.

PR #9303, opened 2021-03-01
HTTP Referrer-Policy - change to default

hamishwillee:
The **default** HTTP `Referrer-Policy` changed from `no-referrer-when-downgrade` to `strict-origin-when-cross-origin` in  Firefox 87 (https://bugzilla.mozilla.org/show_bug.cgi?id=1589074) as a result of this spec change: https://github.com/whatwg/fetch/pull/1066. The change will also affect chrome and others.

In addition, the MDN page has [some notes](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Referrer-Policy#browser_compatibility) about setting the default which I understand we'd now include in BCD (reproduced below):

image

This first draft just puts all of these things as notes . What I THINK needs to happen is that we have a new subfeature to indicate the version at which `strict-origin-when-cross-origin` is used as the default. So if this is false or unknown the assumption is that the default is  `no-referrer-when-downgrade`. Does that make sense?  The other notes from the box above would then be re-worked without all the detail about the "default default" :-)

Does that make sense? If so, any suggestions for the feature name and description?

@ddbeck Your advice appreciated.

Note, this impacts FF content update: https://github.com/mdn/content/issues/2516

### DIFF
--- a/http/headers/referrer-policy.json
+++ b/http/headers/referrer-policy.json
@@ -15,10 +15,20 @@
               "version_added": "79"
             },
             "firefox": {
-              "version_added": "50"
+              "version_added": "50",
+              "notes": [
+                "Firefox version 87 and later have a default <code>Referrer-Policy</code> of <code>strict-origin-when-cross-origin</code>. This default can be changed using the preference <code>network.http.referer.userControlPolicy</code>, where <code>0 = no-referrer</code>, <code>1 = same-origin</code>, <code>2 = strict-origin-when-cross-origin</code>, <code>3 = no-referrer-when-downgrade</code>.",
+                "Firefox versions 59 to 86 have a default <code>Referrer-Policy</code> of <code>3 = no-referrer-when-downgrade (default)</code>. This default can be change using the preference <code>network.http.referer.defaultPolicy</code> (and <code>network.http.referer.defaultPolicy.pbmode</code> for private networks), where <code>0 = no-referrer</code>, <code>1 = same-origin</code>, <code>2 = strict-origin-when-cross-origin</code>, <code>3 = no-referrer-when-downgrade (default)</code>.",
+                "Firefox versions 53 to 58 allow the default <code>Referrer-Policy</code> to be set using the preference <code>network.http.referer.userControlPolicy</code>, where <code>0 = no-referrer</code>, <code>1 = same-origin</code>, <code>2 = strict-origin-when-cross-origin</code>, <code>3 = no-referrer-when-downgrade (default)</code>."
+              ]
             },
             "firefox_android": {
-              "version_added": "50"
+              "version_added": "50",
+              "notes": [
+                "Firefox version 87 and later have a default <code>Referrer-Policy</code> of <code>strict-origin-when-cross-origin</code>. This default can be changed using the preference <code>network.http.referer.userControlPolicy</code>, where <code>0 = no-referrer</code>, <code>1 = same-origin</code>, <code>2 = strict-origin-when-cross-origin</code>, <code>3 = no-referrer-when-downgrade</code>.",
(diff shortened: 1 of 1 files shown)


ddbeck:
1. > have a new subfeature to indicate the version
 
   Right. I'd probably have a subfeature of `http.headers.Referrer-Policy` with a description like "Default policy is `strict-origin-when-cross-origin`".  Maybe name the feature `default_strict-origin-when-cross-origin`

1. For the notes in the screenshot:

    - I'd remove the notes from the page.
    - I'd ignore the `network.http.referer.userControlPolicy` setting. It's not a preview of a forthcoming feature so I don't think it's part of the compat story.
    - I'd include a Firefox 59-87 support statement for the default behind a flag (`network.http.referer.defaultPolicy`), in the aforementioned new feature. Given the schema, you'll have to mention the private browsing mode in a note separately.

hamishwillee:
@ddbeck Thanks. I have added the new subfeature `default_strict-origin-when-cross-origin` with 87 for FF and currently "null" for everything else (I will do more investigation there). I have removed these notes from the parent feature:
```
,
              "notes": [
                "Firefox version 87 and later have a default <code>Referrer-Policy</code> of <code>strict-origin-when-cross-origin</code>. This default can be changed using the preference <code>network.http.referer.userControlPolicy</code>, where <code>0 = no-referrer</code>, <code>1 = same-origin</code>, <code>2 = strict-origin-when-cross-origin</code>, <code>3 = no-referrer-when-downgrade</code>.",
                "Firefox versions 59 to 86 have a default <code>Referrer-Policy</code> of <code>3 = no-referrer-when-downgrade (default)</code>. This default can be change using the preference <code>network.http.referer.defaultPolicy</code> (and <code>network.http.referer.defaultPolicy.pbmode</code> for private networks), where <code>0 = no-referrer</code>, <code>1 = same-origin</code>, <code>2 = strict-origin-when-cross-origin</code>, <code>3 = no-referrer-when-downgrade (default)</code>.",
                "Firefox versions 53 to 58 allow the default <code>Referrer-Policy</code> to be set using the preference <code>network.http.referer.userControlPolicy</code>, where <code>0 = no-referrer</code>, <code>1 = same-origin</code>, <code>2 = strict-origin-when-cross-origin</code>, <code>3 = no-referrer-when-downgrade (default)</code>."
              ]
```

But I don't understand the rest of the instruction, in particular 

> - I'd ignore the `network.http.referer.userControlPolicy` setting. It's not a preview of a forthcoming feature so I don't think it's part of the compat story.
> - I'd include a Firefox 59-87 support statement for the default behind a flag (network.http.referer.defaultPolicy), in the aforementioned new feature. Given the schema, you'll have to mention the private browsing mode in a note separately.

`network.http.referer.userControlPolicy`  and `network.http.referer.defaultPolicy` do the same thing - they allow you to set the default preference (from `strict-origin-when-cross-origin` and  `no-referrer-when-downgrade` respectively), while  `network.http.referer.defaultPolicy.pbmode` lets you set the default for private mode browsing, which happens to already be `strict-origin-when-cross-origin` - so nothing has changed for private mode.

Point being if it isn't worth documenting one, I don't see why we'd document any as a compatibility thing. 

If this is worth doing, you're going to have to show me how. I don't understand how this could work in the schema. Sorry!

http/headers/referrer-policy.json
```diff
@@ -48,6 +48,54 @@
...
+              "opera_android": {
+                "version_added": "60"
+              },
+              "safari": {
+                "version_added": false
+              },
+              "safari_ios": {
+                "version_added": false
+              },
+              "samsunginternet_android": {
```
hamishwillee, line 85: FYI false, because v85 version isn't available yet

http/headers/referrer-policy.json
```diff
@@ -48,6 +48,54 @@
             "deprecated": false
           }
         },
+        "default_strict-origin-when-cross-origin": {
+          "__compat": {
+            "description": "Default policy is strict-origin-when-cross-origin",
```
ddbeck, line 53:
```suggestion
            "description": "Default policy is <code>strict-origin-when-cross-origin</code>",
```

ddbeck (changes requested):
Sorry, I misunderstood the default policy setting bug (I thought it was a way to opt-in into the new default, but that's exactly right). One small change needed here, then we'll be ready to go.

hamishwillee:
Done. Thanks very much!

ddbeck (approved):
Thank you! 🎉

PR #9305, opened 2021-03-01
updating link matching behavior for :link, :visited, :any-link

rachelandrew:
I am working on https://github.com/mdn/content/issues/2506 the discussion there explains why I am making these changes.

Of the three pseudo-classes `:visited` had a separate subsection about this, I've removed that and updated all three to add notes explaining current status.

### DIFF
--- a/css/selectors/any-link.json
+++ b/css/selectors/any-link.json
@@ -8,7 +8,8 @@
           "support": {
             "chrome": [
               {
-                "version_added": "65"
+                "version_added": "65",
+                "notes": "Chromium has never matched <code>&lt;link&gt;</code> elements with link pseudo-classes."
               },
               {
                 "prefix": "-webkit-",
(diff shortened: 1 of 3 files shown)


ddbeck (changes requested):
> I've removed that and updated all three to add notes explaining current status.

I think we need to go the opposite direction here and make the other features more closely resemble `:visited` with a `not_match_link` subfeature. This has three benefits:

1. We don't add an asterisk to most of the top-row `:visited` entries for something that's (now) standardized.
1. We don't add a couple dozen notes for data that can be represented structurally. The subfeature gets the benefit of consistency checks, while notes don't.
1. We can reduce the absolute number of notes by dropping all the redundant "Chromium has never…" notes.

rachelandrew:
@ddbeck I'd discussed this with Chris in the linked discussion and he had suggested going the route that I did, so maybe argue the case there and I'll make it whatever the decision is rather than going back and forth.

rachelandrew:
I'm going to close this PR and start over, as that is likely going to be easier.

PR #9311, opened 2021-03-02
Update forced-colors Chrome version status

dlibby-:
Forced colors mode is shipping in Chrome 89. Updating the version information to reflect this.

See: 
https://chromium.googlesource.com/chromium/src/+/0b1c596b6732f6bc9d440b12af53d3c74aa3c4ac
https://www.chromestatus.com/feature/5757293075365888#details
https://blog.chromium.org/2021/01/chrome-89-beta-advanced-hardware.html

I'm also working on a PR to update the MDN docs for these CSS features.

### DIFF
--- a/css/at-rules/media.json
+++ b/css/at-rules/media.json
@@ -672,7 +672,7 @@
               }
             },
             "status": {
-              "experimental": true,
+              "experimental": false,
               "standard_track": true,
               "deprecated": false
             }
--- a/css/at-rules/media.json
+++ b/css/at-rules/media.json
@@ -611,9 +611,22 @@
             "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/@media/forced-colors",
             "spec_url": "https://drafts.csswg.org/mediaqueries-5/#forced-colors",
             "support": {
-              "chrome": {
-                "version_added": "89"
-              },
+              "chrome": [
+                {
+                  "version_added": "79",
+                  "flags": [
+                    {
+                      "type": "preference",
+                      "name": "#forced-colors",
+                      "value_to_set": "Enabled"
+                    }
+                  ],
+                  "notes": "See <a href='https://crbug.com/970285'>bug 970285</a>."
+                },
+                {
+                  "version_added": "89"
+                }
+              ],
               "chrome_android": {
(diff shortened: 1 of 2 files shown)


dlibby-:
@foolip - PTAL thanks! (I don't seem to have permissions to add a reviewer myself)

css/at-rules/media.json
```diff
@@ -612,15 +612,7 @@
...
-                "version_added": "79",
-                "flags": [
-                  {
-                    "type": "preference",
-                    "name": "#forced-colors",
-                    "value_to_set": "Enabled"
-                  }
-                ],
-                "notes": "See <a href='https://crbug.com/970285'>bug 970285</a>."
+                "version_added": "89"
```
foolip, line 615: I am very happy to see flag data removed, but the [guideline](https://github.com/mdn/browser-compat-data/blob/master/docs/data-guidelines.md#removal-of-irrelevant-flag-data) is to do so only after the feature has been enabled by default for two years. So, what you'll need to do here instead is to use an array of two support statements, also adding `"version_added": "89"` for the flags entry.

dlibby-: Apologies for missing the guidance (should have read more closely). This is fixed now.

foolip: I doubt anyone reads the guidelines before a specific one is relevant in a PR, I haven't read all of them at least :)

css/at-rules/media.json
```diff
@@ -611,17 +611,22 @@
...
+                    {
+                      "type": "preference",
+                      "name": "#forced-colors",
+                      "value_to_set": "Enabled"
+                    }
+                  ],
+                  "notes": "See <a href='https://crbug.com/970285'>bug 970285</a>."
+                },
+                {
+                  "version_added": "89"
```
foolip, line 627: Unfortunately the order of entries matter, and the first one is the one that'll be shown on MDN by default. So you'll need to change the order to put this first.

dlibby-: Oops! Fixed. Thanks for bearing with me!

foolip (approved):

foolip:
Thanks @dlibby-!

PR #9333, opened 2021-03-04
PerformanceObserver.supportedEntryTypes in Firefox

jakub-g:
Bugzilla says v66: https://bugzilla.mozilla.org/show_bug.cgi?id=1512225
but I tested on BrowserStack that it's available only in v68

### DIFF
--- a/api/PerformanceObserver.json
+++ b/api/PerformanceObserver.json
@@ -256,7 +256,7 @@
             }
           },
           "status": {
-            "experimental": true,
+            "experimental": false,
             "standard_track": true,
             "deprecated": false
           }
--- a/api/PerformanceObserver.json
+++ b/api/PerformanceObserver.json
@@ -228,7 +228,7 @@
               "version_added": "68"
             },
             "firefox_android": {
-              "version_added": null
+              "version_added": "68"
             },
             "ie": {
               "version_added": false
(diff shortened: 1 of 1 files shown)


api/PerformanceObserver.json
```diff
@@ -225,7 +225,7 @@
               "version_added": "≤79"
             },
             "firefox": {
-              "version_added": null
+              "version_added": "68"
             },
             "firefox_android": {
               "version_added": null
```
ddbeck, line 231: This is supported on Android too and I can't find anything that suggests it would've landed in a different version. Let's set this one too?

```suggestion
              "version_added": "68"
```

ddbeck (changes requested):
Thanks for this PR, @jakub-g!

One suggestion in a line comment and I have a request that falls outside the diff: could you flip `experimental` to `false`? There's 3 implementations, so I don't think it's going anywhere.

api/PerformanceObserver.json
jakub-g: I couldn't test on Android but yes I assume it should be supported there too.

jakub-g:
@ddbeck thx for review, applied suggested changes

ddbeck (approved):
Thank you! 🎉

PR #9334, opened 2021-03-04
PerformancePaintTiming in Firefox

jakub-g:
Tested with BrowsterStack

Release docs:
https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/84#apis

### DIFF
--- a/api/PerformancePaintTiming.json
+++ b/api/PerformancePaintTiming.json
@@ -17,7 +17,7 @@
             "version_added": "84"
           },
           "firefox_android": {
-            "version_added": null
+            "version_added": "84"
           },
           "ie": {
             "version_added": false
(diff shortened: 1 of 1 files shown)


api/PerformancePaintTiming.json
```diff
@@ -14,7 +14,7 @@
             "version_added": "≤79"
           },
           "firefox": {
-            "version_added": null
+            "version_added": "84"
           },
           "firefox_android": {
             "version_added": null
```
ddbeck, line 20: Looks like this also appeared in Android

```suggestion
            "version_added": "84"
```

ddbeck (changes requested):
One suggestion to expand on this a little. Thank you, @jakub-g.

jakub-g:
@ddbeck updated :)

ddbeck (approved):
Thank you! 🎉

PR #9342, opened 2021-03-05
Add a note of CSSStyleDeclaration's @@iterator

aleen42:
I have confirmed that:

1. `CSSStyleDeclaration` has no `@@iterator` under all IEs as they don't support `Symbol`.
2. Chrome has implemented `CSSStyleDeclaration[Symbol.iterator]` since 51. Under 50 and below, we have to transform it into Array by calling `Array.from()` before iteration.
3. Firefox has implemented with `Symbol.iterator` at the same time since 36.
4. Opera has implemented with `Symbol.iterator` at the same time since 30.

Feel free to update the `null` value of other browsers when ensuring the version added or nonexistent feature.

### DIFF
--- a/api/CSSStyleDeclaration.json
+++ b/api/CSSStyleDeclaration.json
@@ -544,6 +544,54 @@
             "deprecated": false
           }
         }
+      },
+      "@@iterator": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/CSSStyleDeclaration/@@iterator",
+          "support": {
+            "chrome": {
+              "version_added": "51"
+            },
+            "chrome_android": {
+              "version_added": "51"
+            },
+            "edge": {
+              "version_added": "18"
+            },
+            "firefox": {
+              "version_added": "36"
+            },
+            "firefox_android": {
+              "version_added": "36"
+            },
+            "ie": {
+              "version_added": false
+            },
+            "opera": {
+              "version_added": "38"
+            },
+            "opera_android": {
+              "version_added": "41"
+            },
+            "safari": {
+              "version_added": "11"
+            },
+            "safari_ios": {
+              "version_added": "11"
+            },
+            "samsunginternet_android": {
+              "version_added": "5.0"
+            },
+            "webview_android": {
+              "version_added": "51"
+            }
+          },
+          "status": {
+            "experimental": false,
+            "standard_track": true,
+            "deprecated": false
+          }
+        }
       }
     }
   }


aleen42:
any progress?

ddbeck:
Thanks for waiting while I got to reviewing this, @aleen42. I took a look at this, but I'm not sure how you got the version numbers here. Did you do some testing or use some other source to set these values? At least one value here is quite surprising (Opera 30 uses a Blink version before Chrome 51), so I want to make sure we're recording something that's well attested. Thank you again!

ddbeck:
My apologies for the delay, @aleen42. What I was really asking for was: do you have some sample code to test with? I can use BrowserStack or Sauce Labs to run through tests to fill in the `null` values, but I thought it might be more consistent if I used the same test you did.

aleen42:
1. Check `Symbol`:

```js
Symbol; // => assert undefined or not
```

2. Check `Symbol.iterator`:

```js
Symbol.iterator; // => assert undefined or not
```

3. Check the `CSSStyleDeclarations[Symbol.iterator]`

```js
const iterator = getComputedStyle(document.body)[Symbol.iterator]; // assert undefined or not
iterator.next(); // assert work well or not
```

aleen42:
@ddbeck I have not tested other browsers like Firefox Android, Chrome Android, Safari, Safari iOS, and it needs you to review whether the value can be set with `null` temporarily.

aleen42:
@ddbeck Any progress?

ddbeck:
@aleen42 My apologies for the delay here. I mistakenly thought this was blocked by another issue. I'll revisit these test cases soon and review. Thanks for your patience!

aleen42:
@ddbeck Thanks and relax.

api/CSSStyleDeclaration.json
```diff
@@ -544,6 +544,55 @@
...
+      "@@iterator": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/CSSStyleDeclaration/@@iterator",
+          "support": {
+            "chrome": {
+              "version_added": "51",
+              "notes": "Before Chrome 51, we can iterate it after calling <code>Array.from()</code> to transfer it."
+            },
+            "chrome_android": {
+              "version_added": null
```
ddbeck, line 557:
```suggestion
              "version_added": "51"
```

api/CSSStyleDeclaration.json
```diff
@@ -544,6 +544,55 @@
...
+            },
+            "edge": {
+              "version_added": null
```
ddbeck, line 560:
```suggestion
              "version_added": "18"
```

api/CSSStyleDeclaration.json
```diff
@@ -544,6 +544,55 @@
...
+            },
+            "firefox": {
+              "version_added": "36"
+            },
+            "firefox_android": {
+              "version_added": null
```
ddbeck, line 566:
```suggestion
              "version_added": "36"
```

api/CSSStyleDeclaration.json
```diff
@@ -544,6 +544,55 @@
...
+            },
+            "ie": {
+              "version_added": false
+            },
+            "opera": {
+              "version_added": "30"
```
ddbeck, line 572: I couldn't reproduce your result here. This is what I got from testing.

```suggestion
              "version_added": "38"
```

api/CSSStyleDeclaration.json
```diff
@@ -544,6 +544,55 @@
...
+            },
+            "opera_android": {
+              "version_added": null
```
ddbeck, line 575: Confirmed that it works in the current Opera for Android, the version number based on the underlying engine version.

```suggestion
              "version_added": "41"
```

api/CSSStyleDeclaration.json
```diff
@@ -544,6 +544,55 @@
...
+            },
+            "safari": {
+              "version_added": null
```
ddbeck, line 578:
```suggestion
              "version_added": "11"
```

api/CSSStyleDeclaration.json
```diff
@@ -544,6 +544,55 @@
...
+            },
+            "safari_ios": {
+              "version_added": null
```
ddbeck, line 581:
```suggestion
              "version_added": "11"
```

api/CSSStyleDeclaration.json
```diff
@@ -544,6 +544,55 @@
...
+            },
+            "samsunginternet_android": {
+              "version_added": null
```
ddbeck, line 584:
```suggestion
              "version_added": "5.0"
```

api/CSSStyleDeclaration.json
```diff
@@ -544,6 +544,55 @@
...
+            },
+            "webview_android": {
+              "version_added": null
```
ddbeck, line 587:
```suggestion
              "version_added": "51"
```

api/CSSStyleDeclaration.json
```diff
@@ -544,6 +544,55 @@
...
           }
         }
+      },
+      "@@iterator": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/CSSStyleDeclaration/@@iterator",
+          "support": {
+            "chrome": {
+              "version_added": "51",
+              "notes": "Before Chrome 51, we can iterate it after calling <code>Array.from()</code> to transfer it."
```
ddbeck, line 554: I don't think that we should include this note. Two reasons:

1. This is very likely true (excepting the version number) for every browser, so we'd probably have to copy and maintain it for each browser.
2. The browsers this applies to are all more than five years old. It's rather obscure at this point.

That said, if you think it's still worth mentioning, I suggest you open an issue (or PR) for the https://developer.mozilla.org/docs/Web/API/CSSStyleDeclaration page on mdn/content. Since it's a strategy that would make sense for any browser, it could just be a general suggestion on the page.

```suggestion
              "version_added": "51"
```

ddbeck (changes requested):
I've gone ahead and done some testing (or using our underlying engine data) to figure out the missing values here. I also have some ideas about the original note proposed. Thank you!

api/CSSStyleDeclaration.json
aleen42: I will confirm this later under Opera 30

aleen42: It seems my fault in the following testing

ddbeck (approved):
Thank you! 🎉

Issue #9167, opened 2021-02-18
css.selectors.focus-visible - supported in Edge


liesahead:
#### What information was incorrect, unhelpful, or incomplete?
The new versions of edge support `focus-visible` (I tested it myself and caniuse.com also states the same)

![image](https://user-images.githubusercontent.com/32300784/108326114-87c77a00-71d2-11eb-8ec7-cf0445b64541.png)
![image](https://user-images.githubusercontent.com/32300784/108326275-bba29f80-71d2-11eb-8076-a3cd80a56fad.png)

#### Did you test this? If so, how?
Opened MDN `focus-visible` page and tested it in edge.

<details>
<summary>MDN page report details</summary>

* Query: `css.selectors.focus-visible`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/CSS/:focus-visible
* Report started: 2021-02-17T11:58:40.179Z

</details>

ddbeck:
Hi @liesahead and thanks for raising this issue.

It does appear that this data is out of data.

To fix this, the Edge data in [`focus-visible.json`](https://github.com/mdn/browser-compat-data/blob/f5417741503d02a893039fe06f0d8b4f6b4d9c24/css/selectors/focus-visible.json#L4) needs to be updated to reflect (it'll match the data for Chrome, in this case).

I would welcome a PR from you (or another contributor) making these changes.

Thank you again!

ddbeck:
Fixed by #9345! 🎉

PR #9345, opened 2021-03-05
Update focus-visible for Edge

joneskj55:
This PR updates `focus-visible` for Edge as referenced in issue #9167

### DIFF
--- a/css/selectors/focus-visible.json
+++ b/css/selectors/focus-visible.json
@@ -36,16 +36,21 @@
                 ]
               }
             ],
-            "edge": {
-              "version_added": "86",
-              "flags": [
-                {
-                  "name": "#enable-experimental-web-platform-features",
-                  "type": "preference",
-                  "value_to_set": "enabled"
-                }
-              ]
-            },
+            "edge": [
+              {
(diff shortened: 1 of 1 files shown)


css/selectors/focus-visible.json
```diff
@@ -37,7 +37,7 @@
               }
             ],
             "edge": {
```
ddbeck, line 39: I think I poorly described the solution in the original issue. I think what needs to happen here is:

1. Copy the data from Chrome completely (i.e., the array for `"chrome"`, above). This way we'll have two support statements: one without a flag and one with.
2. For Edge only, change version "67" to "79", reflecting the first version Edge was based on Chromium.

ddbeck (changes requested):
Thanks for starting this PR, @joneskj55!

css/selectors/focus-visible.json
joneskj55: I think I understand. Just updated the PR. Please let me know if it is what is expected.

ddbeck (approved):
Looks great. Thank you! 🎉

Issue #9172, opened 2021-02-18
html.elements.input.input-time - is now supported by safari


timstokman:
#### What information was incorrect, unhelpful, or incomplete?

The page claims no support with safari for this input type.

#### What did you expect to see?

Safari added support for this in october 2020, see: https://bugs.webkit.org/show_bug.cgi?id=214946 and https://bugs.webkit.org/show_bug.cgi?id=200416

It's supported from TP 115 onwards apparently.

#### Did you test this? If so, how?

I asked a colleague to test this jsfiddle on his iPad: https://jsfiddle.net/2etfd6vm/

<details>
<summary>MDN page report details</summary>

* Query: `html.elements.input.input-time`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/time
* Report started: 2021-02-18T14:31:31.818Z

</details>

ddbeck:
Thanks for reporting this, @timstokman.

Related to this is @tari's PR https://github.com/mdn/browser-compat-data/pull/9351, which attempts to make this change (and a couple of others) for desktop Safari.

For several reasons, we don't track features in Safari TP (at least not yet). When there's a general release Safari version number to pin this to, we can update the data. Until then, this will have to wait.

timstokman:
Works fine on the iPad without using a TP version of safari, so I'm assuming it made its way into the regular release. Here: https://webkit.org/status/#feature-date-and-time-input-types it says that only the week and month input types are not fully supported, time is.

tari:
My understanding is that mobile Safari has supported these input types for a long time; the recent addition is for desktop Safari. The data already say support exists on iOS.

ddbeck:
Ah, it looks like I misunderstood the desktop/iOS distinction. Looking this over again, here's a summary of the situation:

| feature | iOS support | macOS support |
| --- | --- | --- |
| [`date`](https://github.com/mdn/browser-compat-data/blob/master/html/elements/input/date.json) | 5 | `false`, noted as "The input type is recognized, but there is no date-specific control." (shipping in a future version) |
| [`time`](https://developer.mozilla.org/docs/Web/HTML/Element/input/time) | `true` (missing version) | `false` (shipping in a future version) |
| [`datetime-local`](https://github.com/mdn/browser-compat-data/blob/master/html/elements/input/datetime-local.json) | `true` (missing version) | `false`, noted as "The input type is recognized, but there is no date-specific control." (shipping in a future version) |
[...]

PR #9351, opened 2021-03-06
Desktop Safari adds date/time input support in v14.1

tari:
[Webkit feature status](https://webkit.org/status/#feature-date-and-time-input-types) says `date`, `datetime-local`, and `time` input is supported [as of October 2020](https://bugs.webkit.org/attachment.cgi?id=412182&action=diff), but no release seems to have included that support yet. Technology preview releases have included support since [at least TP 114](https://webkit.org/blog/11300/release-notes-for-safari-technology-preview-114/) which should be included in the next version: 15. [CanIUse agrees](https://caniuse.com/input-datetime) that support is present in current tech previews, though lacking `week` and `month` support.

### DIFF
--- a/html/elements/input/date.json
+++ b/html/elements/input/date.json
@@ -33,8 +33,7 @@
                 "version_added": "11"
               },
               "safari": {
-                "version_added": false,
-                "notes": "The input type is recognized, but there is no date-specific control. See <a href='https://webkit.org/b/119175'>bug 119175</a>."
+                "version_added": "14.1"
               },
               "safari_ios": {
                 "version_added": "5"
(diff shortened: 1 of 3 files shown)


tari:
I see #6896 discusses issues around Safari TPs and policy (in the form of the data validation checks) says unreleased versions aren't valid, so it seems this will have to wait for an actual new release.

foolip:
I'm pretty sure the version released soon will be 14.1, 15 should be in the fall.

tari:
Updated version to 14.1, which per #9172 is being released this week.

lukewarlow:
This was addressed in https://github.com/mdn/browser-compat-data/pull/10129 which has been merged.

PR #9374, opened 2021-03-07
Fix compat notes for Firefox of permissions.request

Rob--W:
- Removed compat notes at Firefox for Android. The notes aren't relevant, e.g. sidebars were never supported on Android.
- Fix out-of-date claims about bugs in support on Firefox desktop.

References:
- support for permissions.request in popup: https://bugzil.la/1432083
- (lack of) support for permissions.request in sidebar: https://bugzil.la/1493396
- support for permissions.request from an embedded options page in about:addons: https://bugzil.la/1382953

### DIFF
--- a/webextensions/api/permissions.json
+++ b/webextensions/api/permissions.json
@@ -165,34 +165,17 @@
               "edge": {
                 "version_added": "79"
               },
-              "firefox": [
-                {
-                  "version_added": "55",
-                  "notes": [
-                    "The user will be prompted again for permissions that have been previously granted and then removed.",
-                    "It's not possible to request permissions from a popup or a sidebar document."
-                  ]
-                },
-                {
-                  "version_added": "56",
-                  "version_removed": "61",
-                  "notes": "It's not possible to request permissions from an options page that's embedded in about:addons. To request permissions from an options page, set the <code>open_in_tab</code> property in the <code>options_ui</code> manifest key, so the options page opens in its own tab."
-                }
-              ],
-              "firefox_android": [
-                {
-                  "version_added": "55",
-                  "notes": [
-                    "The user will be prompted again for permissions that have been previously granted and then removed.",
-                    "It's not possible to request permissions from a popup or a sidebar document."
-                  ]
-                },
-                {
-                  "version_added": "56",
-                  "version_removed": "61",
-                  "notes": "It's not possible to request permissions from an options page that's embedded in about:addons. To request permissions from an options page, set the <code>open_in_tab</code> property in the <code>options_ui</code> manifest key, so the options page opens in its own tab."
-                }
-              ],
+              "firefox": {
+                "version_added": "55",
+                "notes": [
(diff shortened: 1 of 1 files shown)


ddbeck (changes requested):
Thanks for this PR, @Rob--W! A couple of suggestions for you:

* Structuring this as separate support entries feels kinda heavy weight. They don't really represent large departures of behavior, which is usually how we decide between new support statements or not. How about something like this, instead?
  
  ```
  {
    "version_added": "55",
    "notes": [
      "Before version 75, permissions cannot be requested from popups (see bug…)."
      "Before version 61, permissions cannot be requested from options pages embedded in <code>about:addons</code> (see bug…).",
    ]
  }
  ```

* For the "out-of-date claims about bugs in support on Firefox desktop" — did that happen recently? If so, we probably ought to keep it around (e.g., "Before version X…"). If it's quite old (or completely unknown), then we can ignore it.

Rob--W:
> Structuring this as separate support entries feels kinda heavy weight. They don't really represent large departures of behavior, which is usually how we decide between new support statements or not. How about something like this, instead?

Done.

> For the "out-of-date claims about bugs in support on Firefox desktop" — did that happen recently? If so, we probably ought to keep it around (e.g., "Before version X…"). If it's quite old (or completely unknown), then we can ignore it.

"out-of-date" in the sense that a feature was documented to be unsupported while they were in fact supported. 75 is relatively recent, so I've added it. 61 is 3 year old, so if you'd like I could remove the note about that one.

webextensions/api/permissions.json
```diff
@@ -165,34 +165,17 @@
...
-                  "version_removed": "61",
-                  "notes": "It's not possible to request permissions from an options page that's embedded in about:addons. To request permissions from an options page, set the <code>open_in_tab</code> property in the <code>options_ui</code> manifest key, so the options page opens in its own tab."
-                }
-              ],
+              "firefox": {
+                "version_added": "55",
+                "notes": [
+                  "It's not possible to request permissions from a sidebar document (<a href='https://bugzil.la/1493396'>bug 1493396</a>).",
+                  "Before version 75, permissions cannot be requested from popup panels (see <a href='https://bugzil.la/1432083'>bug 1432083</a>).",
+                  "Before version 61, permissions cannot be requested from options pages embedded in about:addons."
```
ddbeck, line 173: Let's keep the note and make it consistent with the others. Generally, I'm inclined to report things that are true and historic, if it's obvious that it's historic. If it's unclear (or unverifiable) and historic, then I usually drop it.

```suggestion
                  "Before version 61, permissions cannot be requested from options pages embedded in <code>about:addons</code> (see <a href='https://bugzil.la/1382953'>bug 1382953</a>."
```

Rob--W: Done. I updated the patch manually since your suggestion was missing a closing parenthesis.

ddbeck: Good eye, thank you!

ddbeck (approved):
I'm really happy with the way this turned out. Thank you for the back and forth, @Rob--W! 🎉

PR #1616, opened 2018-03-18
Add compat data for DOMErrorHandler

bunnybooboo:
https://developer.mozilla.org/en-US/docs/Web/API/DOMErrorHandler

PR #9382, opened 2021-03-08
Remove DOMErrorHandler entry

foolip:
This was added in https://github.com/mdn/browser-compat-data/pull/1616.

The MDN pages for this do not exist, the only reference to this is:
https://developer.mozilla.org/en-US/docs/Web/API/Document_Object_Model#obsolete_dom_interfaces


api/DOMErrorHandler.json
```diff
@@ -1,100 +0,0 @@
...
-        "mdn_url": "https://developer.mozilla.org/docs/Web/API/DOMErrorHandler",
-        "support": {
-          "chrome": {
-            "version_added": null
-          },
-          "chrome_android": {
-            "version_added": null
-          },
-          "edge": {
-            "version_added": "≤18"
```
foolip, line 14: I have tested Edge 18 and there is no `window.DOMErrorHandler`. I didn't expect there to be either, given the `handleError` member this was probably once a callback interface, and those aren't exposed unless they have constants.

Elchi3 (approved):

PR #9383, opened 2021-03-08
Put navigator.permissions.request/requestAll/revoke behind a flag

foolip:
https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/modules/permissions/permissions.idl;l=14;drc=83361f47040ea6253ee1b528160a6af03309ab24

Mirror the Opera data.

Samsung Internet doesn't do flags, so mark that as false.


ddbeck (approved):
Thank you! 🎉

PR #9386, opened 2021-03-08
Remove api.Response.useFinalURL entry

foolip:
There's no trace of this in Chromium, Gecko or WebKit, with no
case-insensitive matches for "usefinalurl" at all. The closest thing is
that there was once a TODO for `finalURL` in Blink which was removed:
https://bugs.chromium.org/p/chromium/issues/detail?id=455112

This is from the initial wiki migration:
https://github.com/mdn/browser-compat-data/pull/940

The version numbers match the parent feature and many sibling features,
and was probably copied.


ddbeck (approved):
Thank you! I've opened https://github.com/mdn/content/pull/3031 to drop the corresponding page from MDN.

PR #9387, opened 2021-03-08
Fix the capitalization of two SVG API entries

foolip:
These were in older versions of SVG:
https://www.w3.org/TR/SVG11/types.html#InterfaceSVGTests
https://www.w3.org/TR/SVG11/linking.html#InterfaceSVGViewElement


ddbeck (approved):
Looks good. Thank you! 🎉

PR #9388, opened 2021-03-08
Remove the api.Window.layoutShift entry

foolip:
This was added in https://github.com/mdn/browser-compat-data/pull/6090
but there is no `layoutShift` global property, nor any "layoutshift"
event. There is a `LayoutShift` interface, but that's already covered in
LayoutShift.json.

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -3443,55 +3443,6 @@
           }
         }
       },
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Interesting! Thank you, @foolip.

PR #9392, opened 2021-03-08
api.Navigator.share - Note Firefox for Android's lack of text sharing support

### LINKED ISSUE #9252: api.Navigator.share - Firefox for Android doesn't support Text, either

The article notes that Firefox for Android does not support sharing files.

The article should also state that sharing Text is also not supported.

Added `navigator.share({ title: 'title', text: 'text' })` to a page server over HTTPS and clicked the button.
Only the title gets through, whether the destination is a file manager or a chat application.

<details>

* Query: `api.Navigator.share`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/share
* Report started: 2021-02-24T16:21:40.537Z

</details>

ddbeck:
Fixes #9252

### DIFF
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -2513,7 +2513,7 @@
             "firefox_android": {
               "version_added": "79",
               "partial_implementation": true,
-              "notes": "Firefox does not support sharing files."
+              "notes": "Firefox does not support sharing files or text."
             },
             "ie": {
               "version_added": false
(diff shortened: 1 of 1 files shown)


api/Navigator.json
```diff
@@ -2513,7 +2513,7 @@
             "firefox_android": {
               "version_added": "79",
               "partial_implementation": true,
-              "notes": "Firefox does not support sharing files."
+              "notes": "Firefox does not support sharing files or text."
```
saschanaz, line 2516: Should it explicitly say it's Android specific issue?

ddbeck: Good point. Updated. 👍

foolip (approved):

PR #9400, opened 2021-03-09
Remove 3 PasswordCredential attributes which were unshipped

foolip:
These were removed in Chromium M64:
https://storage.googleapis.com/chromium-find-releases-static/57b.html#57b2dc23b962feeadd2dcf60a399c9f38d2087e3

That removal reached stable more than 2 years ago, for example Samsung
Internet 9.0 with M67 was released on 2018-09-15.

### DIFF
--- a/api/PasswordCredential.json
+++ b/api/PasswordCredential.json
@@ -96,54 +96,6 @@
           }
         }
       },
-      "additionalData": {
-        "__compat": {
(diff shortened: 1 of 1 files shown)


foolip:
These were documented on MDN, so I've sent https://github.com/mdn/content/pull/2969 to delete that too.

Elchi3 (approved):

PR #9401, opened 2021-03-09
Remove the api.Window.OverconstrainedError entry

foolip:
There's already an api.OverconstrainedError entry with subfeatures. The
compat data for that either matches or is more precise than that being
removed here.

The reason this entry is here is probably because `OverconstrainedError`
was originally not defined using Web IDL but with some spec prose
decribing the interface object.

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -96,54 +96,6 @@
           }
         }
       },
-      "OverconstrainedError": {
-        "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/Window/OverconstrainedError",
-          "support": {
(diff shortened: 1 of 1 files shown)


api/Window.json
```diff
@@ -96,54 +96,6 @@
           }
         }
       },
-      "OverconstrainedError": {
-        "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/Window/OverconstrainedError",
```
foolip, line 101: This page does not exist. No cleanup on MDN needed AFACT.

Elchi3 (approved):

PR #9408, opened 2021-03-09
Clear-Site-Data header: secure context is standard compliant

hamishwillee:
As discussed in https://github.com/mdn/browser-compat-data/pull/8785#issuecomment-793650145 there is a strong implication that the `Clear-Site-Data` requiring a secure context is required by the standard. This flips the status so that standard is true.


ddbeck (approved):
Thank you! 🎉

PR #9415, opened 2021-03-10
simplify browser block definition

septs:
- [ ] Summarize your changes

### DIFF
--- a/schemas/browsers.schema.json
+++ b/schemas/browsers.schema.json
@@ -5,22 +5,26 @@
 
     "browsers": {
       "type": "object",
-      "properties": {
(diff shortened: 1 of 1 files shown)


ddbeck:
Hi @septs and thanks for opening this PR. While I appreciate what you're trying to achieve here, I'm not sure that it represents an overall gain for the file.

Seeing that this change works the way it does made me understand how `additionalProperties` works better, for which I'm quite grateful. I also see that this change does reduce some repetition in the file. But I think the change hides our intent with the file, too. It doesn't show as explicitly that browsers are all of a similar kind.

With that in mind, I'm going to close this PR. Thank you for putting in the effort and giving me a better understanding of JSON Schema.

PR #9424, opened 2021-03-11
Link Chromium/WebKit bugs for grid animations and subgrid

foolip:
Also remove the MDN URL pointing to the parent feature. When on that
page, following the link and ending up on the same page is a confusing
experience.


ddbeck (approved):
Thank you! 🎉

PR #9427, opened 2021-03-11
Fix isPointInPath Path Parameter support info for Safari

ramhugger:
This Pull Request is the continuation of and closes #9353.
I changed Safari isPointInPath Path Parameter support info from false to "7", as suggested by @foolip in the original Pull Request #9353.

### DIFF
--- a/api/CanvasRenderingContext2D.json
+++ b/api/CanvasRenderingContext2D.json
@@ -2320,7 +2320,7 @@
                 "version_added": "24"
               },
               "safari": {
-                "version_added": false
+                "version_added": "7"
               },
               "safari_ios": {
                 "version_added": null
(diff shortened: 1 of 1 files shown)


api/CanvasRenderingContext2D.json
```diff
@@ -2320,7 +2320,7 @@
                 "version_added": "24"
               },
               "safari": {
-                "version_added": false
+                "version_added": "7"
               },
               "safari_ios": {
                 "version_added": null
```
ddbeck, line 2326: Hi @ramhugger. Based on the discussion on the previous PR, I think you could apply this change here too, then this PR could fully replace the previous one:

```suggestion
                "version_added": "7"
```

ddbeck (changes requested):

ddbeck (approved):
Thank you! 🎉

ramhugger:
It's me to thank you @foolip and @ddbeck 🥳!

foolip:
@ramhugger your change prompted me to take a closer look at some of the other entries, and I ended up changing a bunch more to be "7" in https://github.com/mdn/browser-compat-data/pull/9433.

Issue #9423, opened 2021-03-11
Should we treat Safari 6.2 as 6.1 or a later version?


foolip:
In https://github.com/foolip/mdn-bcd-collector, we collect results from various browsers, and map those to browser releases in BCD, since BCD doesn't have every single release but just the more noteworthy ones.

Given the current logic I have, Safari 6.2 on OS X would map to 6.1. However, the UA string of the version on BrowserStack is this:
> Mozilla/5.0 (Macintosh; Intel Mac OS X 10_8_5) AppleWebKit/600.8.9 (KHTML, like Gecko) Version/6.2.8 Safari/537.85.17

That "537.85.17" is larger than the "537.71" claims is the WebKit version for Safari 7. I don't know what "600.8.9" means, though.

This risks leading to spurious updates changing "7" to "6.1" based on results from Safari 6.2.8.

cc @gsnedders

ddbeck:
@apple-web-evangelist and @jensimmons: based on our call earlier this week, I think this would be an example of a question that y'all are uniquely situated to help us with 😃

To add a little more context: I think this is typical of a Safari-related versioning question. We're not quite sure on what versions count as significant, or how to assign features to versions when we have to round things off to known-significant releases. (And, if I understand correctly, this one has a fun subplot of the WebKit version numbers we know of suggesting that Safari versions are sometimes non-consecutive.)

gsnedders:
Safari version is mostly irrelevant to web engine support, what matters is the WebKit version. They're frequently the same, but they aren't necessarily.

Without digging too deeply into the history and looking at purely public sources, it appears what happened is Safari 6.1 (macOS 10.7–8) shipped alongside Safari 7 (macOS 10.9 only) with the same (or roughly the same) WebKit version. Safari 6.2 (macOS 10.8) shipped alongside Safari 7.1 (macOS 10.9 only) with the same (or roughly the same) WebKit version. I can look deeper internally, but I doubt there's much more to say than that?

This presumably means we have features that shipped in Safari 6.2 and Safari 7.1, but that aren't in Safari 7.

foolip:
Thanks @gsnedders that really helps explain what's going on here.

I wonder how we should deal with this in BCD then. I'm a situation like this, in order to pretend there was a strictly linear progression, perhaps it's best to not have the Safari versions which were only for older version of OS X at the time?

foolip:
[...]

PR #9433, opened 2021-03-11
Fix descriptions and Safari versions for canvas Path2D overloads

foolip:
The type was incorrectly described as Path. Rather than changing the
IDs, just lowercase them, as this is much more typical in BCD, and the
name of the argument is also path, although this is not observable.

Safari version archeology:

5 of these were around when isPointInPath/isPointInStroke were added and
were most likely shipped together with Path2D:
https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/html/canvas/CanvasRenderingContext2D.idl?rev=165927

See however https://github.com/mdn/browser-compat-data/issues/9423 about
the challenge of distinguishing Safari 6.1 and 7. 7 is used because
that's the version used for Path2D itself.

drawFocusIfNeeded is different, it was added at WebKit trunk 538.35.0:
https://trac.webkit.org/changeset/168476/webkit
https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Configurations/Version.xcconfig?rev=168476

The Path2D overload was added at WebKit trunk 601.1.1:
https://trac.webkit.org/changeset/173652/webkit
https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Configurations/Version.xcconfig?rev=173652

Both are after Safari 7 was released, so that can't be right. The
likely versions are instead inferred from the WebKit trunk version.

### DIFF
--- a/api/CanvasRenderingContext2D.json
+++ b/api/CanvasRenderingContext2D.json
@@ -382,9 +382,9 @@
             "deprecated": false
           }
         },
-        "Path_parameter": {
+        "path_parameter": {
           "__compat": {
-            "description": "<code>Path</code> parameter",
+            "description": "<code>Path2D</code> parameter",
             "support": {
               "chrome": {
                 "version_added": "36"
@@ -414,7 +414,7 @@
                 "version_added": "7"
               },
               "safari_ios": {
-                "version_added": "8"
+                "version_added": "7"
               },
               "samsunginternet_android": {
                 "version_added": "3.0"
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Thanks, @foolip. These look like good changes to me.

> See however #9423 about the challenge of distinguishing Safari 6.1 and 7. 7 is used because that's the version used for Path2D itself.

Yeah, I need to think on that one a little more. Perhaps we need to write up a semi-complicated procedure for distinguishing those versions.

Issue #6585, opened 2020-08-28
Utilities for querying and updating the data as a set of files


foolip:
Background: BCD is organized as a set of JSON files split across many directories, which are [combined](https://github.com/mdn/browser-compat-data/blob/1df1d90ae2a15447af6bf89dea312a8c990a23a6/index.js#L24-L26) into a single object with all data for data consumers.

Problem: The mapping from files to a single object is not easily reversible. Code in this repository (linter) and elsewhere ([mdn-bcd-collector](https://github.com/foolip/mdn-bcd-collector), [mdn-confluence](https://github.com/mdittmer/mdn-confluence)) sometimes wants to work with the data as a whole, but then report on or update the files themselves.

Concrete use cases / examples (please edit to add):
- Updating the version at object path `javascript.builtins.Intl.ListFormat.formatToParts` and saving the change to file `javascript/builtins/intl/ListFormat.json` (note "Intl" vs. "intl") (https://github.com/mdn/browser-compat-data/issues/3617)
- More generally, make any change, removal or addition to the object, call `something.writeToFile()` and have that work.
- Ensuring that `api.DOMTokenList` is supported if `api.Element.classList` is and reporting it as an error on api/DOMTokenList.json if not. (https://github.com/mdn/browser-compat-data/issues/6571)
- Enumerate all features while applying path-based filters, which at least `npm run mirror` already does
- Semantic BCD diffs. For example, load two different sets of BCD JSON files, and be able to enumerate the features which have changed between them (e.g., for generating human-readable summaries in PRs)

Inspired by the discussion in https://github.com/mdn/browser-compat-data/issues/6571 and supersedes https://github.com/mdn/browser-compat-data/issues/3617.

foolip:
Other nice-to-haves that could follow:
- Iterating browsers releases in the right order
- Querying all features by their support status in different browsers, listing which files (and ideally lines) those are defined by
- Iteration and querying with filters like removing anything behind a flag or in a future release, to make "has it shipped" an easy question to answer
- Helpers to update compat data if needed, something like `feature.support.set('firefox', '54', options)` which would ensure that the data reflects that it's supported in version 54, but only overwrites if `options.remove` or something is set. Or `feature.support.addNote(...)` which doesn't clobber an existing note but instead appends.

ddbeck:
I added the "Semantic BCD diffs" use case to the list.

[...]

PR #9441, opened 2021-03-12
Add experimental utils

ddbeck:
This PR adds some experimental utilities for traversing compat data. These are not intended to be included in the package and won't (at least for now) constitute a public API.

It's kinda big and it's got some rough edges. I'll do a self-review to highlight a few things.

Related to #6585 

Originally from: https://github.com/ddbeck/bcd-utils


.npmignore
```diff
@@ -10,3 +10,5 @@ schemas
 CODE_OF_CONDUCT.md
 CONTRIBUTING.md
 GOVERNANCE.md
+utils
+scripts
```
ddbeck, line 14: Borrowed from #9257

utils/bcd.js
```diff
@@ -0,0 +1,17 @@
...
+  console.warn('Warning: loading BCD from alternate path');
+  requireId = process.env.BCD_UTILS_BCD_PACKAGE_PATH;
+}
+
+if (process.env.BCD_UTILS_REQUIRE_PACKAGE) {
+  // To facilitate further experimentation while also not exposing this as part
+  // of BCD's public API. With this, you can use `/utils` as a submodule.
+  console.warn('Warning: `require`-ing BCD');
+  requireId = '@mdn/browser-compat-data';
+}
```
ddbeck, line 13: Unless and until we expose this a public API, you'd have to copy the `utils` to use them independently of the BCD package. This ought to enable a slightly more ergonomic hack around that problem.

utils/iter-support.js
```diff
@@ -0,0 +1,10 @@
+function iterSupport(compat, browser) {
+  if (browser in compat.support) {
+    const data = compat.support[browser];
+    return Array.isArray(data) ? data : [data];
+  }
+
+  return [];
```
foolip, line 7: An entry with `version_added` null would make this even more like the real data.

utils/query.js
```diff
@@ -0,0 +1,26 @@
...
+ * @throws {ReferenceError} For invalid identifiers
+ */
+function query(path, data = bcd) {
+  const pathElements = path.split('.');
+  let lookup = data;
+  while (pathElements.length) {
+    const next = pathElements.shift();
+    lookup = lookup[next];
+    if (lookup === undefined) {
+      throw ReferenceError(
```
foolip, line 18: `new` keyword not required?

utils/query.test.js
```diff
@@ -0,0 +1,41 @@
+const assert = require('assert');
```
foolip, line 1: If you append `.strict` you can simply the calls named `assert.strictSomething`.

utils/visit.test.js
```diff
@@ -0,0 +1,73 @@
...
+
+    for (const miss in misses) {
+      assert.ok(!hits.has(miss));
+    }
+  });
+
+  it('visitorFn can break iteration', function () {
+    visit(
+      undefined,
+      () => true,
```
ddbeck, line 43: This highlights two ugly things about `visit()` right now:

- To walk the whole tree, you pass in `undefined` as the first argument. That's not exactly an obvious notation. Maybe it'd be friendlier if you passed in another symbol such as `visit.ROOT` or a catchall string (e.g., `*`).
- To visit every feature, you pass in a function that returns `true`. That's more obvious than the first argument's `undefined`, but still not super cryptic. Similar solutions are available here such as a `visit.ALL` or special casing truthy, non-function values, as in `visit(undefined, true, visitorFn)`.

Taken together, there might be an argument to be made for handling multiple signatures for `visit()`, though that gets complex fast:

- `visit(visitorFn)`  (starts at the root, no test — roughly equivalent to `walk()`) 
- `visit(testFn, visitorFn)` (starts at the root, visits those features that pass tests)
- `visit(str, testFn, visitorFn)` (explicit start point)
- `visit(str, visitorFn)` (explicit start point, no test)

utils/walk.js
```diff
@@ -0,0 +1,45 @@
+const bcd = require('./bcd');
+const { isBrowser, descendantKeys, joinPath } = require('./walkingUtils');
+const query = require('./query');
+
+function* lowLevelWalk(data = bcd, path, depth = Infinity) {
```
ddbeck, line 5: Why two walking generators? `walk()` wraps `lowLevelWalk()` but only yields `__compat: { … }` objects. `lowLevelWalk()` is the basis for other sorts of interesting walks, such as walking non-`__compat` namespaces or walking browsers, which have not been implemented yet.

utils/walk.test.js
```diff
@@ -0,0 +1,68 @@
...
+    ];
+
+    const steps = Array.from(lowLevelWalk(undefined, undefined, 1));
+    const paths = steps.map(step => step.path);
+    assert.strictEqual(steps.length, expectedPaths.length);
+    assert.deepStrictEqual(paths, expectedPaths);
+  });
+  it('visits every point in the tree', function () {
+    const paths = Array.from(lowLevelWalk()).map(step => step.path);
+    assert.ok(paths.length > 13000);
```
ddbeck, line 31: Strictly speaking, this doesn't _really_ test this function—just that its output is reasonable—but I feel like the tests on `walk()` do the heavy lifting here.

utils/walk.test.js
```diff
@@ -0,0 +1,68 @@
...
+    );
+    assert.strictEqual(results.length, 28);
+    assert.strictEqual(results[0].path, 'api.Notification');
+    assert.strictEqual(
+      results[results.length - 1].path,
+      'css.properties.color',
+    );
+  });
+
+  it('should yield every feature by default', function () {
```
ddbeck, line 60: This test (and the similar test of `visit`) are rather slow (>300ms and >50ms on my machine), which Mocha highlights. I think it's mostly the overhead of Mocha, but we could restrict this to some subtree if we want to speed things up.

foolip: We could also have a unchanging set of BCD-like JSON files for testing only, which could exercise things that are uncommon in BCD or likely to change over time. That would probably be blazing fast to test with.

One might still have one or two smoke tests involving the real data.

utils/iter-support.js
ddbeck: Ooh, that's a great idea. Done in 59c11d03.

utils/query.js
ddbeck: Apparently it works without it, but it does look off, doesn't it? Fixed with 8e094cf4.

utils/query.test.js
ddbeck: Oh wow, that's convenient! Fixed with 10032cb12df7815ce60537a2dbfdd4e3fab448c7.

.npmignore
foolip: Perfect! I see test/ is already excluded.

utils/bcd.js
foolip: `query` allows passing in the BCD copy to use, which solves this and is convenient for testing. If we do that consistently, do we still need this mechanism?

ddbeck: That's great idea. I'll flag the couple of places that need this.

utils/visit.js
```diff
@@ -0,0 +1,31 @@
+const bcd = require('./bcd');
+const query = require('./query');
+const { descendantKeys, joinPath, isFeature } = require('./walkingUtils');
+
+const BREAK = Symbol('break');
+const CONTINUE = Symbol('continue');
+
+function visit(entryPoint, testFn, visitorFn) {
```
ddbeck, line 8: Needs a data argument https://github.com/mdn/browser-compat-data/pull/9441/files#r593447071

utils/walk.js
```diff
@@ -0,0 +1,45 @@
...
+  }
+
+  if (depth > 0) {
+    for (const key of descendantKeys(data)) {
+      yield* lowLevelWalk(data[key], joinPath(path, key), depth - 1);
+    }
+  }
+}
+
+function* walk(entryPoints) {
```
ddbeck, line 22: Needs a data argument https://github.com/mdn/browser-compat-data/pull/9441/files#r593447071

utils/bcd.js
ddbeck: OK, I've added a `data` parameter to all of the utils where that made sense. This made it possible to drop this module entirely.

utils/walk.test.js
ddbeck: Testing with a unchanging set will be a lot easier now that we've got the consistent `data` parameter.

Is this something I should come up with as part of this PR?

utils/visit.test.js
ddbeck: I'm glad I wrote this out and had to add the `data` argument. I had a bit of an insight here: the visitor function is the only part that's required and the rest are options. I've refactored this to a new-and-improved signature:

`visit(visitor, options)`

or, fully expanded: `visit(visitor, { entryPoint, test, data })`

This makes it possible to be much more explicit about a `visit` call:

```js
visit(
  (path) => console.log(path),
  {
    entryPoint: 'css',
    test: (path) => path.includes('grid')
  }
)
```

utils/walk.js
```diff
@@ -0,0 +1,45 @@
+const bcd = require('..');
+const { isBrowser, descendantKeys, joinPath } = require('./walkingUtils');
+const query = require('./query');
+
+function* lowLevelWalk(data = bcd, path, depth = Infinity) {
+  if (path !== undefined) {
+    yield {
+      path,
+      data,
+      compat: data.__compat,
```
foolip, line 10: How about including only one of `compat` and `browser`, rather than having one be undefined? Trying this out and seeing `browser: undefined` I first wondered when `browser` would ever not be null, if there's a browser filter or something. Not so :)

foolip (approved):
I've tried using `walk('api')` to answer a quick question I had with success.

utils/walk.js
ddbeck: Oh, yep, that's a reasonable interpretation of the undefined values. I've pushed a change that sets `browser`, `compat`, or neither, but never both.

foolip (approved):
Let's do it!!

ddbeck:
Last week, I talked with Florian and he suggested that I document the experimental status somewhere. I've done that by adding [a README file to the `utils` folder](https://github.com/ddbeck/browser-compat-data/blob/0dd95a27a1fc8f9a97a5daf140b49f1699ab8064/utils/README.md). With that satisfied, I'm going to merge this.

Thank you for the excellent reviewing here, @foolip!

PR #9445, opened 2021-03-13
Mark <content> and <shadow> elements as removed in Chromium

foolip:
The chrome, chrome_android and edge fixes for the interfaces were made
by mdn-bcd-collector, the rest was inferred.

Commit which removed this, confirming M89:
https://storage.googleapis.com/chromium-find-releases-static/e9a.html#e9aef1ba859e5942b9e8f5396380008b4443f69d

There are no releases of Opera Android or Samsung Internet using M89
yet, so those must be left as they are.

### DIFF
--- a/api/HTMLContentElement.json
+++ b/api/HTMLContentElement.json
@@ -5,13 +5,16 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLContentElement",
         "support": {
           "chrome": {
-            "version_added": "35"
+            "version_added": "35",
+            "version_removed": "89"
           },
           "chrome_android": {
-            "version_added": "37"
+            "version_added": "37",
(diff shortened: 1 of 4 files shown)


api/HTMLContentElement.json
```diff
@@ -5,13 +5,16 @@
...
+            "version_added": "35",
+            "version_removed": "89"
           },
           "chrome_android": {
-            "version_added": "37"
+            "version_added": "37",
+            "version_removed": "89"
           },
           "edge": {
-            "version_added": "≤79"
```
foolip, line 14: Changing ≤79 to 79 was intentional, suggested by mdn-bcd-collector which means the interfaces weren't in Edge 18.

Elchi3 (approved):

PR #9455, opened 2021-03-15
Fix up FF-Android, opera, safari pointerevents

hamishwillee:
It was reported here that Safari supports pointerdown event: https://github.com/mdn/content/issues/3076. I did some testing and there are several other errors too. Fixed here.

Looking at this:
- Safari support added in 13. I have no iOS but can see it is added here: https://webkit.org/blog/10247/new-webkit-features-in-safari-13-1/
- Opera now depends on Blink. The matching version is 42.
- FF was on preference, which got set default true on desktop in v59. However according to this bug https://bugzilla.mozilla.org/show_bug.cgi?id=1507495 the preference was set true on Geckoview in FF70. However FF70 is NOT valid for android - first relevant one is FF79.

In support of the Opera and Safari changes you can see that there are matching versions for the GlobalEventHandlers "on" pointer events - https://github.com/mdn/browser-compat-data/blob/master/api/GlobalEventHandlers.json#L3094-L3105

@ddbeck One issue. If you look at the GlobalEventHandlers you can see that it marks support for FF Android in version 59: https://github.com/mdn/browser-compat-data/blob/master/api/GlobalEventHandlers.json#L3070
I think this is a typo - someone updated the "on" event handlers and decided that they should be the same for FF desktop and android. However I see no reason why you would do the events and "on" thingies at the same time, and that does not match the bug I refer to above: https://bugzilla.mozilla.org/show_bug.cgi?id=1507495 

Proposal is that I make the https://github.com/mdn/browser-compat-data/blob/master/api/GlobalEventHandlers.json#L3070 match the changes I've done here (ie v59 to v79).


foolip:
https://webkit.org/blog/10247/new-webkit-features-in-safari-13-1/ is about changes in Safari 13.1 / iOS+iPadOS 13.4, but all our data suggests that pointer events were added in 13. https://developer.apple.com/documentation/safari-release-notes/safari-13-release-notes are the release notes, and they also mention pointer events. Not sure why they were also mentioned later, but let's go with 13.

api/Document.json
```diff
@@ -8624,7 +8629,7 @@
             },
             "edge": [
               {
-                "version_added": "12"
+                "version_added": "79"
```
foolip, line 8632:
```suggestion
                "version_added": "12"
```

Pointer Events was proposed and shipped first by Microsoft Edge. 
```suggestion
                "version_added": "79"
```

foolip (approved):

api/Document.json
```diff
@@ -8629,7 +8629,7 @@
             },
             "edge": [
               {
-                "version_added": "79"
+                "version_added": "12"
```
hamishwillee: @foolip Yeah, but they were added in version 12 under the alternative name `mspointerdown` (see the section below). In version 79 when Edge when to Blink everything is compatible - ie the alternative_name is no longer correct. So what I did was add a separate entry.

Now you have changed this to 12, we have a confusing story. Did you miss what I was doing?

That seems "correct"

foolip: Edge wasn't mentioned in the description so I assumed this was a mistake. There are lots of entries that looks something like this:
```
            "edge": [
              {
                "version_added": "12"
              },
              {
                "version_added": "12",
                "version_removed": "79",
                "alternative_name": "onmspointerleave"
              }
            ],
```

That just means both the prefixed and unprefixed variants were supported at the same time. I haven't checked if it's correct, but it's a common situation to have such an overlap.

ddbeck:
This PR is nearly all of the work needed to fix the `_event` features for #9455. Do you know anything about the other three events defined in the pointer events spec? It'd be nice to know if they still need work (or can be fixed alongside the others).

The three are:

- `gotpointercapture_event`
- `lostpointercapture_event`
- `pointerrawupdate_event`

> @ddbeck One issue. If you look at the GlobalEventHandlers you can see that it marks support for FF Android in version 59: https://github.com/mdn/browser-compat-data/blob/master/api/GlobalEventHandlers.json#L3070
> I think this is a typo - someone updated the "on" event handlers and decided that they should be the same for FF desktop and android. However I see no reason why you would do the events and "on" thingies at the same time, and that does not match the bug I refer to above: https://bugzilla.mozilla.org/show_bug.cgi?id=1507495

OK, that makes sense. The data you've got in this PR seems much more likely to be true than the (probably mirrored) stuff for the `on` handlers. Is that something you wanted to include here? Or in a follow up?

api/Document.json
hamishwillee: Thanks @foolip. So basically the first case indicates that it is supported for all versions, and the second indicates that from 12 > 79 there was also an alternative name? Makes sense thank you.

api/GlobalEventHandlers.json
```diff
@@ -2899,7 +2899,7 @@
             ],
             "firefox_android": [
               {
-                "version_added": "59"
+                "version_added": "79"
```
hamishwillee, line 2902: Fixes all the `onpointerxx` variants behind `dom.w3c_pointer_events.enabled` from 59 to 79

api/HTMLElement.json
```diff
@@ -2451,16 +2451,21 @@
...
-              "version_added": "29",
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "dom.w3c_pointer_events.enabled",
-                  "value_to_set": "true"
-                }
-              ]
-            },
+            "firefox_android": [
```
hamishwillee, line 2454: This and below fix all the `onpointerxx` variants behind `dom.w3c_pointer_events.enabled` in HTMLElement to add FF 79 for Android

hamishwillee:
@ddbeck OK, so I 
- updated the mirror onpointerXXX events in the GlobalEventHandlers. 
- updated Opera and Safari to match versions where they said null and false for the pointer events.
- Updates HTMLElement too, because it had the preference to hide some features too.

Looking at the other ones
- gotpointercapture_event - https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/gotpointercapture_event
- lostpointercapture_event - https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/lostpointercapture_event 
- pointerrawupdate_event - not in docs, not in BCD but appears to rely on mouse. Ran this test http://wpt.live/pointerevents/pointerevent_pointerrawupdate.html and got unimplemented from chrome and FF current releases.

So the last one appears not to be implemented in any of the main browsers. 
- The other two have BCD and docs entries. Safari is missing on these and I can't test - but I suspect they would be implemented. 
- BCD indicates that they are not supported on FFAndroid. However I ran some tests (e.g. https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/lostpointercapture_event) and my own test code and these DO seem to work on Firefox for android.

So propose we probably should at least update the BCD for those first two for FFAndroid to "true" - OK? I can't find a version that indicates when these were enabled.

ddbeck:
OK, great. That's very helpful, @hamishwillee. I'm really happy with the changes so far. Here's the remaining things to fix up, I think:

## `pointerrawupdate`

Let's ignore this for now. Safari does support it, but it would be a new feature. Let's not confusing things by adding it in this PR.

## Pointer capture events for Safari

It looks like the pointer capture stuff did get implemented in Safari 13. The API (`setPointerCapture()` etc.) is already set to 13. A search of the WebKit source seems to indicate that the events were implemented too: https://trac.webkit.org/changeset/240634/webkit. So the following features' Safaris need to be set to 13:

- [ ] `Document.getpointercapture_event`
- [ ] `Document.lostpointercapture_event`
- [ ] `HTMLElement.getpointercapture_event`
- [ ] `HTMLElement.lostpointercapture_event`
- [ ] `GlobalEventHandlers.ongotpointercapture`
- [ ] `GlobalEventHandlers.onlostpointercapture`

## Pointer capture events for Firefox

Digging into bugzilla, it looks like Firefox features have similar issues with mirroring desktop to Android around `dom.w3c_pointer_events.enabled` too. Can we fix this while we're at it? That is to say, I think the following features show support in Firefox for Android 59 (or `false`) when they should be 79 (based on everything else I've seen here):

- [ ] `Document.getpointercapture_event`
- [ ] `Document.lostpointercapture_event`
- [ ] `HTMLElement.getpointercapture_event`
- [ ] `HTMLElement.lostpointercapture_event`
- [ ] `GlobalEventHandlers.ongotpointercapture`
- [ ] `GlobalEventHandlers.onlostpointercapture`

hamishwillee:
@ddbeck One day you're just going to merge one of these without any changes (at which point I'll know that you went mad, and got committed to a hospital :-) ).

I've made those changes (except for ongot/onlost-pointercapture Safari, which was already set to 13).

ddbeck (approved):
@hamishwillee Thank you!

> @ddbeck One day you're just going to merge one of these without any changes (at which point I'll know that you went mad, and got committed to a hospital :-) ).

Someday you're going to stumble on something that's easy to fix and get whiplash from how fast I merge it. 😆

foolip:
Let me know if you find any data in BCD that's quick and easy to sort out, I've been looking for a long time :)

hamishwillee:
For all my complaints there is a certain satisfaction to feeling like we "got it right". Especially for stuff like pointers which are pretty "core".

PR #9459, opened 2021-03-15
Delete StorageQuota.json

foolip:
This was added in https://github.com/mdn/browser-compat-data/pull/2106
with reference to https://www.chromestatus.com/feature/6218562888794112.

However, it looks like this part of the API was never enabled, and it was
removed from Chromium in 2017:
https://chromium.googlesource.com/chromium/src/+/5fb4064595996e5f0ae942d928e3bccc5209bca7

Note that the entry point, `navigator.storageQuota`, isn't in BCD.


foolip:
I've sent https://github.com/mdn/content/pull/3142 to clean up MDN.

Elchi3 (approved):

PR #9464, opened 2021-03-15
Remove entries representing indexed property getters

foolip:
These are curious, because the ability to access the members of these
Array-like interfaces is the whole point of them. There are many
interfaces like this, but only these few have entries for [].

These pages exist and will need to be removed:
https://developer.mozilla.org/docs/Web/API/DataTransferItemList/DataTransferItem
https://developer.mozilla.org/docs/Web/API/SourceBufferList/SourceBuffer
https://developer.mozilla.org/docs/Web/API/TrackDefaultList/TrackDefault

These pages do not exist:
https://developer.mozilla.org/docs/Web/API/CSSNumericArray/CSSNumericValue
https://developer.mozilla.org/docs/Web/API/CSSUnparsedValue/CSSUnparsedSegment

The interface pages do exist and show that support is largely identical
to the parent feature:
https://developer.mozilla.org/docs/Web/API/CSSNumericArray#Browser_compatibility
https://developer.mozilla.org/docs/Web/API/CSSUnparsedValue#Browser_compatibility
https://developer.mozilla.org/docs/Web/API/DataTransferItemList#Browser_compatibility
https://developer.mozilla.org/docs/Web/API/SourceBufferList#Browser_compatibility
https://developer.mozilla.org/docs/Web/API/TrackDefaultList#Browser_compatibility

The exception is SourceBufferList, but that must be an error in the data.

### DIFF
--- a/api/CSSNumericArray.json
+++ b/api/CSSNumericArray.json
@@ -48,54 +48,6 @@
           "deprecated": false
         }
       },
-      "CSSNumericValue": {
-        "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/CSSNumericArray/CSSNumericValue",
-          "support": {
-            "chrome": {
-              "version_added": "66"
-            },
-            "chrome_android": {
-              "version_added": "66"
-            },
-            "edge": {
-              "version_added": "79"
-            },
-            "firefox": {
-              "version_added": false
-            },
-            "firefox_android": {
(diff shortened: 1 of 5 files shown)


sideshowbarker (approved):

foolip:
I've sent https://github.com/mdn/content/pull/5819 to remove the pages mentioned in the description.

PR #9467, opened 2021-03-15
Update support for DataTransfer.items for Safari

### LINKED ISSUE #9200: api.DataTransfer.items - Safari information incorrect, Safari has full support

Browser compatibility > Safari 

Full support

Tested in a Safari browser.

Browser version: 14.0.2

http://whatismybrowser.com/w/HJ66Q9Q

<details>

* Query: `api.DataTransfer.items`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/DataTransfer/items
* Report started: 2021-02-20T09:50:56.808Z

</details>

ddbeck:
Was reported in #9200 as supported in Safari 14. Found https://trac.webkit.org/changeset/222376/webkit showing that it was enabled in Safari (it says something about iOS, but I don't think iOS actually fires the drag and drop events) circa 2017. [Tested](https://gist.github.com/ddbeck/9b05cc6915c565eba89317d3eef1c678) to confirm that it landed in Safari 11.1 and not earlier.

Fixes #9200.

### DIFF
--- a/api/DataTransfer.json
+++ b/api/DataTransfer.json
@@ -418,7 +418,7 @@
               "version_added": "11.1"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "11.3"
             },
             "samsunginternet_android": {
               "version_added": true
--- a/api/DataTransfer.json
+++ b/api/DataTransfer.json
@@ -415,7 +415,7 @@
               "version_added": true
             },
             "safari": {
-              "version_added": false
+              "version_added": "11.1"
             },
             "safari_ios": {
               "version_added": false


api/DataTransfer.json
```diff
@@ -415,7 +415,7 @@
               "version_added": true
             },
             "safari": {
-              "version_added": false
+              "version_added": "11.1"
```
foolip, line 418: This exact edit is suggested by mdn-bcd-collector too.

api/DataTransfer.json
```diff
@@ -415,7 +415,7 @@
...
             },
             "safari_ios": {
               "version_added": false
```
foolip, line 421:
```suggestion
              "version_added": "11.3"
```

mdn-bcd-collector tested false in 10.3 and true in 11.3. (No iOS 11 results available.)

ddbeck: A check in BrowserStack shows `false` for iOS 11. 👍

For some reason I had in my head that `DataTransfer` wasn't supported at all on iOS. I'd bet there are lots of wrong `false` values for iOS in this interface.

foolip (approved):

PR #9468, opened 2021-03-15
Update stop-opacity for Safari iOS and remove duplications (#9402)

Havegum:
# Summary of changes
* Updated Safari iOS' entry for `stop-opacity` to show missing support between versions 11 and 14.
* Removed duplication of data for `stop-opacity` in svg.attributes.presentation following this discussion: #9402

Details on testing and verification of support information can also be found in that issue.

Tests and lints passed with no issues.
Although, when running `npm install` the package-lock.json file was modified a bunch, which I since had to unstage before committing. Not sure what the right way to deal with that is, but either way updating dependencies is outside the scope of this PR

Related issue: #9402

### DIFF
--- a/svg/elements/stop.json
+++ b/svg/elements/stop.json
@@ -144,6 +144,7 @@
         },
         "stop-opacity": {
           "__compat": {
+            "mdn_url": "https://developer.mozilla.org/docs/Web/SVG/Attribute/stop-opacity",
             "support": {
               "chrome": {
                 "version_added": "1"
--- a/svg/elements/stop.json
+++ b/svg/elements/stop.json
@@ -172,9 +172,15 @@
               "safari": {
                 "version_added": "3"
               },
-              "safari_ios": {
-                "version_added": "3"
-              },
+              "safari_ios": [
+                {
+                  "version_added": "3",
+                  "version_removed": "11"
+                },
+                {
+                  "version_added": "14"
+                }
+              ],
               "samsunginternet_android": {
                 "version_added": "1.0"
               },
--- a/svg/attributes/presentation.json
+++ b/svg/attributes/presentation.json
@@ -2210,54 +2210,6 @@
             }
           }
         },
-        "stop-opacity": {
-          "__compat": {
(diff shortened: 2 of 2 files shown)


svg/elements/stop.json
```diff
@@ -172,9 +172,15 @@
               "safari": {
                 "version_added": "3"
               },
-              "safari_ios": {
-                "version_added": "3"
-              },
+              "safari_ios": [
```
ddbeck, line 175: Looks great, @Havegum!

One suggestion I didn't notice until now: can we add `"mdn_url": "https://developer.mozilla.org/docs/Web/SVG/Attribute/stop-opacity"` to this feature (just before the `"support"` key, outside the diff)

ddbeck (changes requested):

ddbeck:
Oh about this:

> Although, when running `npm install` the package-lock.json file was modified a bunch, which I since had to unstage before committing. Not sure what the right way to deal with that is, but either way updating dependencies is outside the scope of this PR

Without seeing the diff, I can't be certain, but I suspect you're running a newer version of Node and npm. Since we support the oldest supported LTS, it uses `"lockfileVersion": 1`, while newer versions try to upgrade to 2. I think the workaround is to use `npm ci` instead of `npm install`. Sorry about that!

Havegum:
I've added the "mdn_url"

`npm ci` did the trick, new commit passed linting and testing without messing with any package-files, thanks!

ddbeck (approved):
Looks great, @Havegum. Thank you! 🎉

svg/elements/stop.json
```diff
@@ -172,9 +173,15 @@
...
+                {
+                  "version_added": "3",
+                  "version_removed": "11"
```
foolip, line 179: @ddbeck did you mean to suggest putting this entry first? Now https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/stop-opacity#browser_compatibility makes it look like this isn't supported on iOS at all.

ddbeck: No, I did not. I've opened a PR to correct this: https://github.com/mdn/browser-compat-data/pull/10629

Issue #9457, opened 2021-03-11
Issue with "Spread syntax (...)": should it be rest in destructuring?


arturcarvalho:
MDN URL: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax

Hi there,

On the browser compatibility table for the spread syntax, it is mentioned "spread destructuring". Shouldn't it be "rest destructuring" or something along those lines? [Initial discussion](https://twitter.com/rauschma/status/1369932326172524544?s=20).

The error when the syntax is with the wrong order also hints at that:

<img width="395" alt="image" src="https://user-images.githubusercontent.com/2487710/110771930-5aaf3a00-825b-11eb-94fe-c8deadb6082a.png">

Thanks and have a nice day!

<details>
<summary>MDN Content page report details</summary>

* Folder: `en-us/web/javascript/reference/operators/spread_syntax`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Spread_syntax
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/javascript/reference/operators/spread_syntax/index.html
* Last commit: https://github.com/mdn/content/commit/3b80812a1d4f68e57f35fca763f50bf3e5dfe997
* Document last modified: 2021-02-19T19:57:02.000Z

</details>

ddbeck:
Interesting find. Thanks for reporting this, @arturcarvalho!

I think the problem here is that we've duplicated the data for this syntax:

- incorrectly as [`javascript.operators.spread.spread_in_destructuring`](https://github.com/mdn/browser-compat-data/blob/5e7ff0a7194088756dedebb561d776abaf7ed03d/javascript/operators/spread.json#L131)
- correctly (and in more detail) as [`javascript.operators.destructuring.rest_in_arrays`](https://github.com/mdn/browser-compat-data/blob/4ee5bb9468593a786527a4a658759428e1ec7304/javascript/operators/destructuring.json#L109) and [`javascript.operators.destructuring.rest_in_objects`](https://github.com/mdn/browser-compat-data/blob/4ee5bb9468593a786527a4a658759428e1ec7304/javascript/operators/destructuring.json#L171) ([as seen on MDN](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/Destructuring_assignment#browser_compatibility))

[...]

PR #9472, opened 2021-03-15
Remove spread in destructuring

arturcarvalho:
Hi there,

The goal of this PR is to remove "spread in destructuring" from the spread syntax compatibility table as mentioned in this [issue](https://github.com/mdn/browser-compat-data/issues/9457).

Supporting information:
On the object literal [proposal](https://github.com/tc39/proposal-object-rest-spread) it's mentioned:

> rest elements for array destructuring assignment
> rest properties for object destructuring assignment 

Destructuring is rest and not spread.

The error message when the rest element also hints at being rest and not spread destructuring:

image

Hope this helps.

### DIFF
--- a/javascript/operators/spread.json
+++ b/javascript/operators/spread.json
@@ -128,57 +128,6 @@
             }
           }
         },
-        "spread_in_destructuring": {
-          "__compat": {
-            "description": "Spread in destructuring",
-            "support": {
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Perfect. Thank you, @arturcarvalho! 🎉

PR #9475, opened 2021-03-16
Updates api.ResizeObserver in Firefox for Android v79

### LINKED ISSUE #9261: api.ResizeObserver - Show Full Support in Firefox for Android 85

Firefox for Android v85 supports all of the ResizeObserver API

Currently the data on MDN shows it's not supported.

I used the API online via Firefox for Android and checked for all of its methods using version 85.  I also successfully ran the code in strict mode without error or warning.

<details>

* Query: `api.ResizeObserver`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver
* Report started: 2021-02-25T01:00:52.592Z

</details>

lucalves:
Hi guys!

Regarding issue #9261 all the methods of the `api.ResizeObserver` feature are working on Firefox version for Android 85.

However, I was unable to perform a test on previous versions to see if there is a version before 85 that was already working.

Could someone tell me what would be the best way to test with older mobile browsers on a desktop?

Fixes #9261.

### DIFF
--- a/api/ResizeObserver.json
+++ b/api/ResizeObserver.json
@@ -17,7 +17,7 @@
             "version_added": "69"
           },
           "firefox_android": {
-            "version_added": "85"
+            "version_added": "79"
           },
           "ie": {
             "version_added": false
--- a/api/ResizeObserver.json
+++ b/api/ResizeObserver.json
@@ -65,7 +65,7 @@
               "version_added": "69"
             },
             "firefox_android": {
-              "version_added": false
+              "version_added": "79"
             },
             "ie": {
               "version_added": false
@@ -113,7 +113,7 @@
               "version_added": "69"
             },
             "firefox_android": {
-              "version_added": false
+              "version_added": "79"
             },
             "ie": {
               "version_added": false
@@ -161,7 +161,7 @@
               "version_added": "69"
             },
             "firefox_android": {
-              "version_added": false
+              "version_added": "79"
             },
             "ie": {
               "version_added": false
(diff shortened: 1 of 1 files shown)


foolip:
> Could someone tell me what would be the best way to test with older mobile browsers on a desktop?

I've never done this, but I know that @vinyldarkscratch has installed old APKs of various browsers on Android for testing with http://mdn-bcd-collector.appspot.com/, but I don't know if there's an archive of old Firefox versions somewhere.

In a case like this where the API probably doesn't have any special platform dependency, I would just assume it was shipped in the same time as on desktop.

api/ResizeObserver.json
```diff
@@ -17,7 +17,7 @@
             "version_added": "69"
           },
           "firefox_android": {
-            "version_added": false
+            "version_added": "85"
```
foolip, line 20:
```suggestion
            "version_added": "79"
```

This is a relatively safe assumption to make, simply mirroring the Firefox data. Note that there was no Firefox 69-78 on Android, this is the first after 68.

foolip (commented):
Can you also update the other subfeatures in this file which currently are marked as not supported? Assuming "79" for those too seems fine.

lucalves:
Of course, @foolip. I also made the change to the Pull Request title.

foolip:
@lucalves I've gone ahead and made the same change on the subfeatures to save us both the trouble of another review roundtrip :)

foolip (approved):

Issue #9183, opened 2021-02-19
SVGGeometryElement/SVGPathElement duplication of members


foolip:
There are 3 members that were originally in `SVGPathElement` which have been moved to `SVGGeometryElement`: `getPointAtLength`, `getTotalLength` and `pathLength`.

For Chromium, that move happened in M33: https://bugs.chromium.org/p/chromium/issues/detail?id=315802

BCD still has these 3 members on `SVGPathElement` as well as on `SVGGeometryElement`, and the data isn't entirely aligned.

Fixing this will run into https://github.com/mdn/browser-compat-data/issues/3463, because it will be necessary to say that some members of `SVGGeometryElement` were introduced earlier than `SVGGeometryElement` itself.

PR #9479, opened 2021-03-16
Merge duplicate SVGGeometryElement+SVGPathElement entries

foolip:
This was tested by checking for the 3 members on all 7 types of elements
which currently inherit from SVGGeometryElement:

http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=9076

```js
var tags = ['circle', 'ellipse', 'line', 'path', 'polygon', 'polyline', 'rect'];
var members = ['getPointAtLength', 'getTotalLength', 'pathLength'];
tags.forEach(function(localName) {
  var e = document.createElementNS('http://www.w3.org/2000/svg', localName);
  members.forEach(function(member) {
    w(localName + '.' + member + ': ' + typeof e[member]);
  });
});
```

In Chromium, getTotalLength and getPointAtLength were moved in M56,
and pathLength was moved in M57:
https://bugs.chromium.org/p/chromium/issues/detail?id=596043
https://storage.googleapis.com/chromium-find-releases-static/887.html#887bbf6c0edcabfa012099379c2d68a3dc4afd3f
https://storage.googleapis.com/chromium-find-releases-static/6fd.html#6fd1cbb3c5a636ee3830729e4b26ca1d5d78c8d6

In Gecko, the SVGGeometryElement was introduced in 53, but elements
other than <path> were made to inherit in 61:
https://bugzilla.mozilla.org/show_bug.cgi?id=1239100
https://bugzilla.mozilla.org/show_bug.cgi?id=1325320

In WebKit the change was made at trunk version 606.1.14:
https://trac.webkit.org/changeset/230829/webkit
https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Configurations/Version.xcconfig?rev=230829

In Edge 13 and IE 11, only path.getPointAtLength and path.getTotalLength
were supported, there was no support for pathLength.

In Opera 12.16, all 3 members were supported on path and nowhere else.

Fixes https://github.com/mdn/browser-compat-data/issues/9183.


api/SVGGeometryElement.json
```diff
@@ -11,13 +11,13 @@
             "version_added": "18"
           },
           "edge": {
-            "version_added": "≤79"
+            "version_added": "12"
```
foolip, line 14: In order to satisfy the lint it's necessary to say that `SVGGeometryElement` was introduced when some of its members were first introduced on `SVGPathElement`. I think this is the right thing to do and support for the interface itself isn't very interesting, but should I use `partial_implementation` for the range where the interface actually didn't exist?

api/SVGGeometryElement.json
```diff
@@ -247,90 +423,122 @@
...
-            }
-          },
-          "status": {
-            "experimental": false,
-            "standard_track": true,
-            "deprecated": false
-          }
-        }
-      },
-      "supportOtherThanPath": {
```
foolip, line 294: This "supportOtherThanPath" entry is removed because it's all represented on the individual entries now.

api/SVGGeometryElement.json
```diff
@@ -101,44 +189,132 @@
...
+              {
+                "version_added": "1",
+                "version_removed": "56",
+                "partial_implementation": true,
+                "notes": "Only supported for <a href='https://developer.mozilla.org/docs/Web/API/SVGPathElement'><code>SVGPathElement</code></a>, not all <code>SVGGeometryElement</code> objects."
+              }
+            ],
+            "chrome_android": [
+              {
+                "version_added": "18"
```
foolip, line 205:
```suggestion
                "version_added": "56"
```
oops

api/SVGGeometryElement.json
```diff
@@ -247,90 +423,122 @@
...
             "edge": {
-              "version_added": null
-            },
-            "firefox": {
-              "version_added": "61"
-            },
-            "firefox_android": {
-              "version_added": "61"
-            },
+              "version_added": "79"
```
foolip, line 449: This is not a mistake. `pathLength` wasn't supported in IE or EdgeHTML, confirmed by testing.

api/SVGGeometryElement.json
```diff
@@ -247,90 +423,122 @@
...
+              {
+                "version_added": "≤12.1",
+                "version_removed": "44",
+                "partial_implementation": true,
+                "notes": "Only supported for <a href='https://developer.mozilla.org/docs/Web/API/SVGPathElement'><code>SVGPathElement</code></a>, not all <code>SVGGeometryElement</code> objects."
+              }
+            ],
+            "opera_android": [
+              {
+                "version_added": "43"
```
foolip, line 489: This is not a mistake. The off-by-13 rule held up to Opera 42 on both desktop and mobile, then things got out of sync. Opera desktop 43 was Blink 56, but Android 43 skipped to 59.

api/SVGGeometryElement.json
ddbeck: Yes, I think this is the right thing to do. I think it's nice to have the symmetry between the `partial_implementation` for the subfeatures and the parent. It's a little messy, but the reality is messy too.

ddbeck (changes requested):
This looks good, @foolip. I think the only change needed here is to extend the partial implementation detail to the parent feature. Thank you!

api/SVGGeometryElement.json
foolip: OK, I'm going to split and use notes like the one on `BaseAudioContext`.

foolip: Gah, I shouldn't have look at that! It took https://github.com/mdn/browser-compat-data/pull/9513 + https://github.com/mdn/browser-compat-data/pull/9514 + https://github.com/mdn/browser-compat-data/pull/9515 + https://github.com/mdn/browser-compat-data/pull/9516 to arrive at something to copy.

api/SVGGeometryElement.json
```diff
@@ -4,42 +4,132 @@
...
+            {
+              "version_added": "12",
+              "version_removed": "79",
+              "partial_implementation": true,
+              "notes": "The <code>SVGGeometryElement</code> interface itself is not present, but some of its members are available on the <a href='https://developer.mozilla.org/docs/Web/API/SVGPathElement'><code>SVGPathElement</code></a> interface."
+            }
+          ],
+          "firefox": [
+            {
+              "version_added": "53"
```
foolip, line 42: This is when `SVGGeometryElement` itself was introduced in Firefox. It was only later that all interfaces were made to inherit from it. But doing that as 3 entries here does not seem like it would make the situation more clear to anyone. The subfeatures are what matter more here.

ddbeck: Great. I like this pragmatic approach. 👍

ddbeck (approved):
Thank you, @foolip! 🎉

PR #9480, opened 2021-03-16
Update SVGGeometryElement's isPointIn* Edge / Firefox data

foolip:
It's a coincidence that it's "79" for both.

Edge 79 is suggested by mdn-bcd-collector, because these weren't in Edge
18.

mdn-bcd-collector also confirms support in current Firefox for Android,
and then the Firefox data was mirrored, assuming concurrent shipping.


ddbeck (approved):
Thank you, @foolip!

PR #9489, opened 2021-03-17
adding color-mix()

rachelandrew:
CSS Color Level 5 has the `color-mix()` functional notation, which has been implemented behind a pref in Firefox 88 so I'll document it as part of the work I'm doing to improve the docs on these functions.

Spec: https://www.w3.org/TR/css-color-5/#colormix
Bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1695376

It is also in Safari TP: https://webkit.org/blog/11577/release-notes-for-safari-technology-preview-122/

I also fixed a couple of the URLs to other color functional notations which now have their own pages as part of this work.

### DIFF
--- a/css/types/color.json
+++ b/css/types/color.json
@@ -196,7 +196,7 @@
         },
         "color-mix": {
           "__compat": {
-            "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/color_value/color_mix()",
+            "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/color_value/color-mix()",
             "description": "<code>color-mix()</code>",
             "support": {
               "chrome": {
--- a/css/types/color.json
+++ b/css/types/color.json
@@ -197,7 +197,7 @@
         "color-mix": {
           "__compat": {
             "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/color_value/color_mix()",
-            "description": "<code>color_mix()</code>",
+            "description": "<code>color-mix()</code>",
             "support": {
               "chrome": {
                 "version_added": false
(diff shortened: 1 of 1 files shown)


css/types/color.json
```diff
@@ -194,6 +194,69 @@
             }
           }
         },
+        "color_mix": {
```
ddbeck, line 197:
```suggestion
        "color-mix": {
```

css/types/color.json
```diff
@@ -194,6 +194,69 @@
...
+          "__compat": {
+            "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/color_value/color_mix()",
+            "description": "<code>color_mix()</code>",
```
ddbeck, line 200:
```suggestion
            "description": "<code>color-mix()</code>",
```

css/types/color.json
```diff
@@ -194,6 +194,69 @@
             }
           }
         },
+        "color_mix": {
+          "__compat": {
+            "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/color_value/color_mix()",
```
ddbeck, line 199: I couldn't find an issue or PR for this on the content side of things. I'd guess that you're planning to use the actual name from the spec, but feel free to ignore this suggestion if not.

```suggestion
            "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/color_value/color-mix()",
```

ddbeck (changes requested):
A couple little fixes needed here. Thank you, @rachelandrew!

ddbeck (approved):
Thank you, @rachelandrew! 🎉

PR #9510, opened 2021-03-19
Testing shows SVG <set> works on FF desktop/android

hamishwillee:
Fixes https://github.com/mdn/content/issues/3231

There is a test example on this page for set and "to" attribute: https://developer.mozilla.org/en-US/docs/Web/SVG/Element/set s
This works on current FF desktop and android variants. Does not work on current IE11. I can't find any docs on when this was introduced in FF so I set to true.

### DIFF
--- a/svg/elements/set.json
+++ b/svg/elements/set.json
@@ -15,13 +15,13 @@
               "version_added": "≤79"
             },
             "firefox": {
-              "version_added": null
+              "version_added": true
             },
             "firefox_android": {
-              "version_added": null
+              "version_added": true
             },
             "ie": {
-              "version_added": null
+              "version_added": false
             },
             "opera": {
               "version_added": true
@@ -61,13 +61,13 @@
                 "version_added": "≤79"
               },
               "firefox": {
-                "version_added": null
+                "version_added": true
               },
               "firefox_android": {
-                "version_added": null
+                "version_added": true
               },
               "ie": {
-                "version_added": null
+                "version_added": false
               },
               "opera": {
                 "version_added": true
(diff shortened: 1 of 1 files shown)


foolip:
In the absence of any other evidence, you could match the versions in https://developer.mozilla.org/en-US/docs/Web/API/SVGSetElement#browser_compatibility

hamishwillee:
Thanks @foolip .  The relationship between the API and elements was news to me. Updated to match.

There is no obvious (to me) direct mapping between the `<set>` `to` attribute and `SVGSetElement` - we might assume that "to" versions  follow  `<set>` , or perhaps that it follows  [SVGAnimationElement.getSimpleDuration](https://developer.mozilla.org/en-US/docs/Web/API/SVGAnimationElement#browser_compatibility). 

For now I have just just used true and false, because even if `SVGAnimationElement.getSimpleDuration` does not map to `to`, the existence of such a method implies that `to` must have been implemented.

My _leaning_ is to make the `to` attribute follow `<set>` version 

Thoughts?

foolip:
@hamishwillee the relationship between elements and interfaces are usually clear from the name alone, but there are tricky cases too. For HTML, https://html.spec.whatwg.org/multipage/indices.html#element-interfaces shows the relationship, and for SVG you can find the element in https://svgwg.org/svg2-draft/eltindex.html and then click through and look for "DOM Interfaces".

In this case, it's spelled out in https://svgwg.org/specs/animations/#InterfaceSVGSetElement:
> The SVGSetElement interface corresponds to the ‘set’ element.

Since the `to` attribute is central to what `<set>` is for, I would just assume it was introduced at the same time as `<set>` itself.

hamishwillee:
Thanks very much. Updated, I think we're good to go now.

foolip (approved):

PR #9513, opened 2021-03-19
Mark parts of Web Audio API as shipping in Chrome 14 (not 10)

foolip:
This was also wrong on caniuse.com and possibly copied from there. The
evidence for Chrome 14 being correct is pretty strong:
https://github.com/Fyrd/caniuse/pull/5823

webkitAudioContext itself is already marked as shipping in Chrome 14 in
BCD, so the previous state was inconsistent.


Elchi3 (approved):
Thanks! 👍

PR #9514, opened 2021-03-19
Add entries for webkitOfflineAudioContext

foolip:
In addition to not representing prefixed support, much of the existing
data claimed support that was earlier than it really was.

Chrome versions for webkitOfflineAudioContext and OfflineAudioContext were
found with two mdn-bcd-collector tests:
https://mdn-bcd-collector.appspot.com/tests/api/webkitOfflineAudioContext
https://mdn-bcd-collector.appspot.com/tests/api/OfflineAudioContext/OfflineAudioContext

Other Chromium data was mirrored.

Safari 14 does not support the unprefixed constructor, that should be
coming in Safari 14.1.

Support for webkitOfflineAudioContext was confirmed in iOS 7, and so
assumed to have been introduced in iOS 6, matching webkitAudioContext.

### DIFF
--- a/api/OfflineAudioContext.json
+++ b/api/OfflineAudioContext.json
@@ -4,12 +4,26 @@
       "__compat": {
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/OfflineAudioContext",
         "support": {
-          "chrome": {
-            "version_added": "14"
-          },
-          "chrome_android": {
-            "version_added": "18"
-          },
+          "chrome": [
+            {
+              "version_added": "35"
+            },
+            {
+              "version_added": "25",
+              "version_removed": "57",
+              "prefix": "webkit"
+            }
+          ],
+          "chrome_android": [
+            {
+              "version_added": "35"
+            },
+            {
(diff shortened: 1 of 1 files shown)


api/OfflineAudioContext.json
```diff
@@ -52,14 +96,26 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/OfflineAudioContext/OfflineAudioContext",
           "description": "<code>OfflineAudioContext()</code> constructor",
           "support": {
-            "chrome": {
-              "version_added": "55",
-              "notes": "Before Chrome 59, the default values were not supported."
```
foolip, line 57: I couldn't figure out what these notes might be referring to so I removed them. Expecting to be asked by a reviewer to dig deeper, but taking my chances :)

ddbeck: I gave a quick look for something which might even hint as to what this could be about and found nothing. I think this is the right call. 👍

ddbeck (approved):
Thank you! 🎉

PR #9515, opened 2021-03-19
Mirror Web Audio API shipping in Firefox 25 to Android (was 26)

foolip:
The existing data is inconsistent, with AudioScheduledSourceNode and
already ScriptProcessorNode marked as shipping in Firefox Android 25.

What actually happened is hard to tell from release notes and bug:
https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/25
https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/26
https://bugzilla.mozilla.org/show_bug.cgi?id=885505

This is made consistent because when updating existing data and notes in
BaseAudioContext, one has to assume one or the other. The difference
between 25 and 26 no longer matters, so just say 25.

### DIFF
--- a/api/AnalyserNode.json
+++ b/api/AnalyserNode.json
@@ -17,7 +17,7 @@
             "version_added": "25"
           },
           "firefox_android": {
-            "version_added": "26"
+            "version_added": "25"
           },
           "ie": {
             "version_added": false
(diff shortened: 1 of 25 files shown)


Elchi3 (approved):
https://bugzilla.mozilla.org/show_bug.cgi?id=885505 looks like the macOS part and reading the discussion it seems it was uplifted in Fx25. 
Lets assume Android is the same, and yes, it doesn't matter much anymore if 25 or 26.

PR #9520, opened 2021-03-20
Use relative paths in markdown files

saschanaz:
Removes one blocker in #6292.

This also makes it much easier to navigate within IDEs.


ddbeck (approved):
Thank you, @saschanaz! I double checked that links in `README.md` would work correctly on npm with `@npmcorp/marky-markdown` too, so we're good to go here.

PR #9562, opened 2021-03-23
api.EventTarget.addEventListener.options - add signal option

### LINKED ISSUE #9555: api.EventTarget.addEventListener.options - Missing feature for `signal` option
MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener

The browser compatibility table.

Information about the signal option.

<details>

* Folder: `en-us/web/api/eventtarget/addeventlistener`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/EventTarget/addEventListener
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/api/eventtarget/addeventlistener/index.html
* Last commit: https://github.com/mdn/content/commit/f37c24de21200c64921064e536ea591d71df6cb3
* Document last modified: 2021-03-02T22:00:49.000Z

</details>

hamishwillee:
Fixes #9555
@ddbeck Comments inline. 

Further, I just called this subfeature `signal`. There is a bit of a mess of naming, and I guess that will impact ordering. I was wondering if perhaps we should rename according to a patter. So perhaps option_signal, option_passive, option_once. Then the passive option exceptiosn could be option_passive_default_etc. I will leave unchanged unless you decide renaming is sensible.

### DIFF
--- a/api/EventTarget.json
+++ b/api/EventTarget.json
@@ -533,19 +533,19 @@
               "description": "<code>signal</code> option",
               "support": {
                 "chrome": {
-                  "version_added": "89"
+                  "version_added": "90"
                 },
                 "chrome_android": {
-                  "version_added": "89"
+                  "version_added": "90"
                 },
                 "edge": {
-                  "version_added": "89"
+                  "version_added": "90"
                 },
                 "firefox": {
                   "version_added": "86"
                 },
                 "firefox_android": {
-                  "version_added": true
+                  "version_added": "86"
                 },
                 "ie": {
                   "version_added": false
@@ -554,7 +554,7 @@
                   "version_added": false
                 },
                 "opera": {
-                  "version_added": "74"
+                  "version_added": "76"
                 },
                 "opera_android": {
                   "version_added": false
@@ -569,7 +569,7 @@
                   "version_added": false
                 },
                 "webview_android": {
-                  "version_added": "89"
+                  "version_added": "90"
                 }
               },
               "status": {
--- a/api/EventTarget.json
+++ b/api/EventTarget.json
@@ -527,6 +527,57 @@
                 "deprecated": false
               }
             }
+          },
+          "signal": {
+            "__compat": {
+              "description": "<code>signal</code> option",
+              "support": {
+                "chrome": {
+                  "version_added": "89"
+                },
+                "chrome_android": {
+                  "version_added": "89"
+                },
+                "edge": {
+                  "version_added": "89"
+                },
+                "firefox": {
+                  "version_added": "86"
+                },
+                "firefox_android": {
+                  "version_added": true
+                },
+                "ie": {
+                  "version_added": false
+                },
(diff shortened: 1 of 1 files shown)


api/EventTarget.json
```diff
@@ -527,6 +527,57 @@
                 "deprecated": false
               }
             }
+          },
+          "signal": {
+            "__compat": {
+              "description": "<code>signal</code> option",
+              "support": {
+                "chrome": {
+                  "version_added": "89"
```
hamishwillee, line 536: @ddbeck FYI you indicated Chrome 90, my testing shows 89 works, but the fact Opera works on Opera 74 indicates maybe it went in on Chrome 88?

api/EventTarget.json
```diff
@@ -527,6 +527,57 @@
...
+                "firefox": {
+                  "version_added": "86"
+                },
+                "firefox_android": {
+                  "version_added": true
+                },
+                "ie": {
+                  "version_added": false
+                },
+                "nodejs": {
```
hamishwillee, line 553: Note, the spec predates the last release we have, so I have just put as false.

api/EventTarget.json
ddbeck: Nice catch. You must have experimental web features turned on. It was [added behind a preference in 88](https://storage.googleapis.com/chromium-find-releases-static/f0b.html#f0bbc3f3ab2bf91bea606fbc5fda751f5e60a0d3) and [enabled by default in 90](https://storage.googleapis.com/chromium-find-releases-static/a64.html#a641ee4fd048768763e23a6b4126c1990b64f86b).

You can enter them as two support statements (one with a flag added in 88 and removed in 90, then one without a flag added in 90), or you can ignore the flag and only report 90 (since, by the time this lands in BCD, it will be interesting to web developers for a couple of weeks at best).

ddbeck (changes requested):
Thank you, @hamishwillee. Suggestion for handling the 88 vs. 90 question in a line comment.

> Further, I just called this subfeature `signal`. There is a bit of a mess of naming, and I guess that will impact ordering. I was wondering if perhaps we should rename according to a patter. So perhaps option_signal, option_passive, option_once. Then the passive option exceptiosn could be option_passive_default_etc. I will leave unchanged unless you decide renaming is sensible.

Let's leave this for now. The structure here is a bit weird. It's probably the case that we ought to have `passive_true_touch` and `passive_true_wheel` as subfeatures of `passive`, or rename all of them, as you suggest. But really, we don't have a guideline for how these things are supposed to be and I'm not inclined to rearrange things without a guideline. I'll open a separate issue for that.

hamishwillee:
Thanks @ddbeck . I ended up ignoring the preference versions and just updating to v90 (and equivalent, if there was a matching release). Should be good to go.

ddbeck (approved):
Thank you! 🎉

PR #9575, opened 2021-03-24
Update Safari support info for unicode-bidi

apple-web-evangelist:
This particular case is https://developer.mozilla.org/en-US/docs/Web/CSS/unicode-bidi. 

Notably, `isolate-override` and `plaintext` are listed as unsupported. https://bugs.webkit.org/show_bug.cgi?id=50949 and http://trac.webkit.org/changeset/89864 indicate that `plaintext` was implemented in 2011. https://bugs.webkit.org/show_bug.cgi?id=73164 and http://trac.webkit.org/changeset/109806 indicate that `isolate-override` was implemented in 2012.

### DIFF
--- a/css/properties/unicode-bidi.json
+++ b/css/properties/unicode-bidi.json
@@ -119,16 +119,20 @@
               "opera_android": {
                 "version_added": "35"
               },
-              "safari": {
+              "safari": [{
+                "version_added": "11"
+              }, {
                 "prefix": "-webkit-",
                 "version_added": "6",
                 "notes": "Avoiding using <code>-webkit-isolate</code>. It can lock up older versions of Safari (up to version 9) and Chrome (up to version 47)."
-              },
-              "safari_ios": {
+              }],
+              "safari_ios": [{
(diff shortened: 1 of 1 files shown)


css/properties/unicode-bidi.json
```diff
@@ -119,16 +119,20 @@
...
-              "safari_ios": {
+              }],
+              "safari_ios": [{
+                "version_added": "11"
+              }, {
                 "prefix": "-webkit-",
                 "version_added": "6",
                 "notes": "Avoiding using <code>-webkit-isolate</code>. It can lock up older versions of Safari (up to version 9) and Chrome (up to version 47)."
-              },
+              }],
```
Sammygirl30, line 135: Hmm

Sammygirl30 (commented):
As

css/properties/unicode-bidi.json
apple-web-evangelist: Style corrected in #9576.

PR #9576, opened 2021-03-24
Update Safari support for unicode-bidi

apple-web-evangelist:
This particular case is https://developer.mozilla.org/en-US/docs/Web/CSS/unicode-bidi. 

Notably, `isolate-override` and `plaintext` are listed as unsupported. https://bugs.webkit.org/show_bug.cgi?id=50949 and http://trac.webkit.org/changeset/89864 indicate that `plaintext` was implemented in 2011. https://bugs.webkit.org/show_bug.cgi?id=73164 and http://trac.webkit.org/changeset/109806 indicate that `isolate-override` was implemented in 2012.

### DIFF
--- a/css/properties/unicode-bidi.json
+++ b/css/properties/unicode-bidi.json
@@ -119,16 +119,20 @@
               "opera_android": {
                 "version_added": "35"
               },
-              "safari": {
+              "safari": [{
+                "version_added": "11"
+              }, {
                 "prefix": "-webkit-",
                 "version_added": "6",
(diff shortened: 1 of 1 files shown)


ddbeck (commented):
Thanks for opening this PR, @apple-web-evangelist.

The original prefixed versions make sense here. But the Safari 11 unprefixing, did you have anything to substantiate those changes?

apple-web-evangelist:
Safari 11 deprecated the prefixes: https://trac.webkit.org/browser/webkit/releases/Apple/Safari%2011.0/WebCore/css/CSSProperties.json#L3499

ddbeck (approved):
Thanks for the extra detail, @apple-web-evangelist!

PR #9580, opened 2021-03-24
Add XPath/XSLT spec URLs

sideshowbarker:
Given that for XSLT and XPath, we have only two BCD features that are documented in MDN, maybe we need to question the value of having these in BCD at all; maybe we should just `git rm -r` the `xpath` and `xslt` directories from the repo?


Elchi3 (approved):
Merging this but please file an issue that discusses removal.

PR #9586, opened 2021-03-24
Add Permissions Policy (aka Feature Policy) spec URLs

sideshowbarker:


Elchi3 (commented):
Hm, should we wait here given the renaming?

sideshowbarker:
> Hm, should we wait here given the renaming?

I thought about that too — but it seems there’s no need to wait for the renaming, because the spec_url values aren’t going to change as a result of the renaming. That is, when we rename the BCD filenames and the MDN article titles, they’re still going be referencing the same spec URLs.

To be clear: The spec URLs in this patch all already point to the right sections in the relevant specs — and those spec sections have, with one exception, already been retitled to reference the Permissions Policy spec rather than Feature Policy.

So all the spec_url values in this patch already have some form of `#permissions-policy` anchors — with the one exception being https://w3c.github.io/accelerometer/#accelerometer-interface, because that spec has no Permissions Policy Integration section yet. So I raised https://github.com/w3c/accelerometer/pull/60 to add a Permissions Policy Integration section for that.

But I don’t think we need to block this PR on that PR getting merged first. Instead, after that PR gets merged, we can do a follow-up here to update our `spec_url` for it.

Elchi3 (approved):
Well okay, can't hurt to merge this I guess. Thanks for digging into this!

PR #1890, opened 2018-04-22
Fill in compat data for api.BaseAudioContext

lucian95:
[BaseAudioContext](https://developer.mozilla.org/docs/Web/API/BaseAudioContext)
**Question**: in [createPeriodicWave](https://developer.mozilla.org/docs/Web/API/BaseAudioContext/createPeriodicWave#Browser_compatibility) the data about Chrome on android and Webview is more complete. Can we use it to infer about the other sub-features? (I mean to infer that all sub-features and basic support had prefix at v33, then no prefix at v57). 
Also, I forgot to add `baseLatency` and `suspend()`. Will come back to it later.

PR #9590, opened 2021-03-24
Make Android data for BaseAudioContext more consistent

foolip:
The origin of "33" appears to be when Web Audio was unprefixed:
https://github.com/mdn/browser-compat-data/pull/1890

(It was actually unprefixed in Chromium 35.)

The data references at the time does not inspire confidence:
https://web.archive.org/web/20180421031716/https://developer.mozilla.org/en-US/docs/Web/API/BaseAudioContext#Browser_compatibility
https://web.archive.org/web/20180309013210/https://developer.mozilla.org/en-US/docs/Web/API/BaseAudioContext/createPeriodicWave#Browser_compatibility

Instead set support to "18" like most of the Web Audio API and mirror
that to Samsung Internet which appears to have been mirrored from this.
WebView Android is similar, but set it to ≤37 like most of the Web Audio
API. It might have been supported earlier, but it's hard to test for.

The important part of this is the data previously said a lot of this was
shipped in Chrome for Android 33, but that's inconsistent with all the
other Web Audio API data, which says 18.

### DIFF
--- a/api/BaseAudioContext.json
+++ b/api/BaseAudioContext.json
@@ -179,7 +179,7 @@
               "version_added": "14"
             },
             "chrome_android": {
-              "version_added": "33"
+              "version_added": "18"
             },
             "edge": {
               "version_added": "12"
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
There was so much dubious stuff in those old tables! Thanks for cleaning this up, @foolip! 🎉

Issue #6656, opened 2020-09-08
How to represent the introduction of AudioScheduledSourceNode?


foolip:
As [documented](https://developer.mozilla.org/en-US/docs/Web/API/AudioScheduledSourceNode), [`AudioScheduledSourceNode`](https://webaudio.github.io/web-audio-api/#AudioScheduledSourceNode) is an abstract interface with concrete inheriting interfaces being `AudioBufferSourceNode`, `OscillatorNode`, and `ConstantSourceNode`. Unfortunately, it originally didn't exist, but was added to the prototype chain during the evolution of the Web Audio API.

The concrete problem in BCD now is that the ended event is represented in multiple places:
- [`AudioScheduledSourceNode`](https://developer.mozilla.org/en-US/docs/Web/API/AudioScheduledSourceNode#Browser_compatibility)  has `ended` event, `onended`, `start` and `stop`
- [`AudioBufferSourceNode`](https://developer.mozilla.org/en-US/docs/Web/API/AudioBufferSourceNode#Browser_compatibility) has `onended` and `start`
- [`OscillatorNode`](https://developer.mozilla.org/en-US/docs/Web/API/OscillatorNode#Browser_compatibility) has `onended`, `start` and `stop`
- [`ConstantSourceNode`](https://developer.mozilla.org/en-US/docs/Web/API/ConstantSourceNode#Browser_compatibility) has `onended`, `start` and `stop`

The data for these differs, and there's plenty wrong with it. But what should it say?

From the point of view of a web developer, the introduction of `AudioScheduledSourceNode` doesn't matter, what matters is that one can call `start()`, `stop()` and listen for the `ended` event on the 3 concrete audio node types.

This is a specific instance of a more general problem, [How to represent changes/moves on the prototype chain](https://github.com/mdn/browser-compat-data/issues/3463). It's the opposite situation to [Data for localName/namespaceURI/prefix attributes on Attr/Element/Node is inconsistent/confusing](https://github.com/mdn/browser-compat-data/issues/6545), as here things have moved upwards in the prototype chain, not downwards.

foolip:
As for a proposed solution in this case, I'd have to say removing the entries from the 3 concrete node types. If we find that event and methods became available at different times on the concrete types, a note on the support data for AudioScheduledSourceNode could explain it. I hope this is now far enough in the past that it doesn't matter in practice.

cc @ddbeck @vinyldarkscratch

foolip:
[...]

PR #9599, opened 2021-03-25
Remove duplicate AudioScheduledSourceNode start/stop/onended entries

foolip:
AudioBufferSourceNode is as old as the Web Audio API itself, and
originally had these members. OscillatorNode and ConstantSourceNode were
added later and also had these members. Later, AudioScheduledSourceNode
was introduced and the members were moved up to that prototype. Only an
overload start() method remained on AudioBufferSourceNode.

Summary of changes:
 - Remove the members no longer in spec/implementations.
 - Mark the members of AudioScheduledSourceNode as supported since the
   earliest time they were supported on AudioBufferSourceNode, which is
   the earliest of the concrete interfaces.
 - Mark AudioScheduledSourceNode as partially supported from the
   versions AudioBufferSourceNode were introduced until the
   AudioScheduledSourceNode interface was added.

Introduction of AudioScheduledSourceNode in each browser/engine:

Chromium 57:
https://bugs.chromium.org/p/chromium/issues/detail?id=661207
https://storage.googleapis.com/chromium-find-releases-static/009.html#00971f38908388728f49cd5127b9c6c6761d035f

Firefox 53:
https://bugzilla.mozilla.org/show_bug.cgi?id=1324568
mozilla/gecko-dev@bd93c7b

WebKit trunk 610.1.20, which was Safari 14:
https://bugs.webkit.org/show_bug.cgi?id=214381
https://trac.webkit.org/changeset/264538/webkit

There was an AudioSourceNode parent interface in WebKit, but it never
had any members.

These tests were used to find or confirm some version numbers:
https://staging-dot-mdn-bcd-collector.appspot.com/tests/api/AudioScheduledSourceNode

(It uses a AudioBufferSourceNode, as the oldest concrete type now
inheriting from AudioScheduledSourceNode.)

For start/stop, Chrome 23 and 24 were tested to confirm.

WebKit commit for that at trunk version 537.12:
https://trac.webkit.org/changeset/129260/webkit
https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Configurations/Version.xcconfig?rev=129260

For onended, Chrome 29 and 30 were tested to confirm.

WebKit commit for that at trunk version 537.44:
https://trac.webkit.org/changeset/150905/webkit
https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Configurations/Version.xcconfig?rev=150905

That also introduced support for the event itself.

Edge 13 was tested to confirm support for onended/start/stop, and it was
assumed to also be in Edge 12 since most of the Web Audio API was.

Firefox 24 and 25 tested were tested to confirm Firefox data.

Fixes https://github.com/mdn/browser-compat-data/issues/6656.


foolip:
I think this is my longest commit message in BCD yet :)

api/AudioBufferSourceNode.json
```diff
@@ -467,10 +419,10 @@
               "version_added": "14"
             },
             "safari": {
-              "version_added": "6"
+              "version_added": "6.1"
```
foolip, line 422: These is reflecting when `AudioBufferSourceNode.play` was introduced, in WebKit trunk version 537.12.

api/AudioScheduledSourceNode.json
```diff
@@ -10,8 +10,9 @@
             },
             {
               "version_added": "14",
-              "version_removed": "58",
-              "alternative_name": "AudioSourceNode"
```
foolip, line 14: I don't think there's any point to us documenting that there was a `AudioSourceNode` parent interface, since it didn't have any members. Since it even existing at the same time as `AudioScheduledSourceNode` briefly, it wasn't even a matter of a straight rename.

api/AudioScheduledSourceNode.json
```diff
@@ -127,19 +159,19 @@
...
-              "version_added": "15"
+              "version_added": "17"
             },
             "opera_android": {
-              "version_added": "14"
+              "version_added": "18"
             },
             "safari": {
-              "version_added": "14"
+              "version_added": "7"
```
foolip, line 168: This reflects WebKit trunk version 537.44, see description.

api/ConstantSourceNode.json
```diff
@@ -112,66 +110,10 @@
...
-              "version_added": "56"
-            },
-            "edge": {
-              "version_added": "79"
-            },
-            "firefox": {
-              "version_added": "52",
-              "version_removed": "53",
-              "notes": "This property is still available, but via the inheritance of <code>AudioScheduledSourceNode</code>."
+              "version_added": "52"
```
foolip, line 113: The diff here looks weird, but it is just a removal of some entries, I didn't change `offset`. (The data for it is wrong though. Not gonna fix it here.)

Elchi3 (approved):
Thanks, this follows the pattern of using partial_implementation nicely for prototype moves. It would be nice if you could write a data guideline for this now that we've successfully used this in at least two scenarios.

I've reviewed this by comparing with the current IDL files and it makes sense structurally. For the all versions numbers I just glanced over them and I think they makes sense.

PR #9602, opened 2021-03-25
Update when initProgressEvent was removed

foolip:
The removal was in WebKit trunk version 535.10:
https://trac.webkit.org/changeset/100727/webkit
https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Configurations/Version.xcconfig?rev=100727

All versions were inferred from that. This is not a 100% reliable
method, but being off by one version doesn't matter here.

### DIFF
--- a/api/ProgressEvent.json
+++ b/api/ProgressEvent.json
@@ -106,7 +106,7 @@
             },
             "chrome_android": {
               "version_added": true,
-              "version_removed": true
+              "version_removed": "18"
             },
             "edge": {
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
Thanks, hopefully this will qualify as irrelevant at some point anyways given it is just in IE.

foolip:
I would have deleted it, but am counting 2 years from Edge 79, which hasn't passed yet :/

PR #9603, opened 2021-03-25
Mark BatteryManager as never shipped in Safari iOS

foolip:
The implementation was removed here:
https://trac.webkit.org/changeset/208300/webkit

Per https://bugs.webkit.org/show_bug.cgi?id=164213, "Apple never enabled
it on their ports."

### DIFF
--- a/api/BatteryManager.json
+++ b/api/BatteryManager.json
@@ -70,8 +70,7 @@
             "version_added": false
           },
           "safari_ios": {
-            "version_added": true,
-            "version_removed": true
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
Thanks! 
I have always been very skeptical of allowing added/removed: true at the same time.

foolip:
> I have always been very skeptical of allowing added/removed: true at the same time.

Indeed, that's why I found and fixed these. There are very few left now :)

PR #9604, opened 2021-03-25
Drop Permissions Policy spec URL for XHR

sideshowbarker:
See https://github.com/whatwg/xhr/pull/321:

> We should probably drop this section for now as it doesn't have multi-implementer interest.

Related: https://github.com/whatwg/xhr/pull/295

---

So, while the section in question is currently still in the XHR spec, for BCD purposes, we should just ignore it anyway (per the comment cited above from the spec editor).

### DIFF
--- a/http/headers/feature-policy.json
+++ b/http/headers/feature-policy.json
@@ -1394,7 +1394,6 @@
         "sync-xhr": {
           "__compat": {
             "mdn_url": "https://developer.mozilla.org/docs/Web/HTTP/Headers/Feature-Policy/sync-xhr",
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):

ddbeck:
Should `standard_track` be flipped to `false` for this too? Or is there a subtlety here that I'm missing?

sideshowbarker:
> Should `standard_track` be flipped to `false` for this too? Or is there a subtlety here that I'm missing?

No subtlety missed — I just overlooked it. Will raise a follow-up PR for it right now

sideshowbarker:
Raised https://github.com/mdn/browser-compat-data/pull/9605

Issue #8891, opened 2021-01-28
originAgentCluster enabled in Chrome 88, likely Chromium 90 for other browsers


foolip:
Filing an issue because this launch is more complicated, and it's likely we'll get it wrong in BCD.

https://chromium-review.googlesource.com/c/chromium/src/+/2655607/1#message-6015d41347247cf5d3852eaed265745756826202 explains the situation.

Finch, a frameworks for remotely enabling/disabling features in Chrome, is being used to enable the `originAgentCluster` feature in Chrome 88, although not from the very beginning of its launch. (I'm on Chrome 88 on macOS and don't have it enabled.)

Other Chromium-based browsers will likely get this in milestone 90, when the feature was flipped to stable in the source.

ddbeck:
How does the post-release enabling work, exactly? For BCD's purposes, the interesting thing is probably when a web developer can actually _rely_ on the feature.

foolip:
I don't know Finch well, but I believe it's originally a framework for experiments, so settings can be scaled between 0 and 100% of users. Those settings and percentages are published by some server that Chrome clients ping for settings. So in effect, a feature can be turned from 0 to 100 or the other way at any date, and after some time passes most users will have it. However, I don't know how long it takes to propagate, and I also don't know how close to 100% one actually gets.

@domenic do you know more, when is it reasonable to say that  web developers can rely on `originAgentCluster`?

domenic:
Probably by mid-next-week, web developers can rely on `originAgentCluster` being deployed to 100% of Chrome 88 users.

foolip:
Testing https://mdn-bcd-collector.appspot.com/tests/api/Window/originAgentCluster on BrowserStack now, it tests false on Chrome 88 and 89 (on Windows 10) but true for Chrome 90. I don't know if BrowserStack are doing anything special here, but  testing Chrome 89 on Sauce Labs it also doesn't look supported.

Down the line, everything is going to be simpler if we just say that this was shipped in Chromium 90 across the board. Or we could go with 88 with a note.

foolip:
Well, I've sent https://github.com/mdn/browser-compat-data/pull/9606 and https://github.com/mdn/browser-compat-data/pull/9607. With https://github.com/mdn/browser-compat-data/pull/9607 and this issue remaining open, I don't think we'll forget how to handle this. We just need to wait.

foolip:
This is now done, with the data being updated across a number of PRs.

PR #9606, opened 2021-03-25
Add originAgentCluster attribute (enabled in Chrome 88)

foolip:
As discussed in https://github.com/mdn/browser-compat-data/issues/8891,
this was enabled in Chrome 88 using Finch, and will reach other
Chromium-based browsers when they ship with Chromium 90.

https://mdn-bcd-collector.appspot.com/tests/api/Window/originAgentCluster
has been tested to confirm it's *not* enabled in Edge 89, but *is*
enabled in Chrome 89 on my own computer.

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -6021,7 +6021,7 @@
             }
           },
           "status": {
-            "experimental": false,
+            "experimental": true,
             "standard_track": true,
             "deprecated": false
           }
--- a/api/Window.json
+++ b/api/Window.json
@@ -6023,7 +6023,7 @@
           "status": {
             "experimental": false,
             "standard_track": true,
-            "deprecated": true
+            "deprecated": false
           }
         }
       },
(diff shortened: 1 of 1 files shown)


api/Window.json
```diff
@@ -5979,6 +5979,54 @@
...
+            },
+            "samsunginternet_android": {
+              "version_added": false
+            },
+            "webview_android": {
+              "version_added": false
+            }
+          },
+          "status": {
+            "experimental": false,
```
ddbeck, line 6024:
```suggestion
            "experimental": true,
```

api/Window.json
```diff
@@ -5979,6 +5979,54 @@
...
+            "standard_track": true,
+            "deprecated": true
```
ddbeck, line 6026:
```suggestion
            "deprecated": false
```

ddbeck (changes requested):
I think the statuses got a little mixed up, but apart from that, I'm happy with this approach (paired with #9607). 👍

PR #9607, opened 2021-03-25
Mark originAgentCluster as shipped in Chromium 90 for non-Chrome

foolip:
See https://github.com/mdn/browser-compat-data/issues/8891.


foolip:
This was closed because the target branch was deleted and I don't know how to retarget it, so I've opened https://github.com/mdn/browser-compat-data/pull/9666 instead.

PR #9608, opened 2021-03-25
Add Opera for Android 62

foolip:
That it's based on Chromium 87 was confirmed by looking at the UA string
on Opera installed on a real Android phone.


Elchi3 (approved):
Confirmed in https://forums.opera.com/topic/47270/opera-for-android-62/3

PR #9609, opened 2021-03-25
Update data for CreateOscillator and related APIs

foolip:
Chrome 20 initially confirmed/suggested by this mdn-bcd-collector test:
http://mdn-bcd-collector.appspot.com/tests/api/BaseAudioContext/createOscillator

This was implemented in WebKit trunk version 536.6:
https://trac.webkit.org/changeset/112938/webkit
https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Configurations/Version.xcconfig?rev=112938

That maps to Chrome 20 and Safari 6.

Oscillator and its detune, frequency and type members can also be
confirmed shipping in Chrome 20 with this test:
http://mdn-bcd-collector.appspot.com/tests/api/OscillatorNode

PeriodicWave was originally called WaveTable, and was renamed in WebKit
trunk version 538.1:
https://trac.webkit.org/changeset/151914/webkit
https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Configurations/Version.xcconfig?rev=151914

That was after the Blink fork, but mdn-bcd-collector confirms Chrome 30.
538.1 maps to Safari 8. The available browsers on BrowserStack make it
impossible to confirm this, so just assume 8 is correct.

Don't bother adding the alternative_name entries for WaveTable. There
has been a lot of renaming in the Web Audio API, especially early on,
and trying to capture it all will make the data messier for no real
benefit given how far in the past this is now.

### DIFF
--- a/api/BaseAudioContext.json
+++ b/api/BaseAudioContext.json
@@ -752,10 +752,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/BaseAudioContext/createOscillator",
           "support": {
             "chrome": {
-              "version_added": "14"
+              "version_added": "20"
             },
             "chrome_android": {
-              "version_added": "18"
+              "version_added": "25"
             },
             "edge": {
               "version_added": "12"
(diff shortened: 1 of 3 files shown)


ddbeck (approved):
Thank you! 🎉

PR #9610, opened 2021-03-25
Lint for version_added being the same as version_removed

foolip:
Whether the versions are null, boolean or strings, this doesn't
communicate a lot of information.

### DIFF
--- a/test/linter/test-versions.js
+++ b/test/linter/test-versions.js
@@ -107,7 +107,12 @@ function checkVersions(supportData, relPath, logger) {
             chalk`{red → {bold ${relPath}} - {bold version_removed: "${statement.version_removed}"} is {bold NOT} a valid version number for {bold ${browser}}\n    Valid {bold ${browser}} versions are: ${validBrowserVersionsString}}`,
           );
         }
-        if ('version_removed' in statement && 'version_added' in statement) {
+        if ('version_added' in statement && 'version_removed' in statement) {
+          if (statement.version_added === statement.version_removed) {
+            logger.error(
+              chalk`{red → {bold ${relPath}} - {bold version_added: "${statement.version_added}"} must not be the same as {bold version_removed} for {bold ${browser}}}`,
+            );
+          }
           if (
             typeof statement.version_added !== 'string' &&
             statement.version_added !== true


foolip:
Here's what was failing initially:
```
Problems in 7 files:
✖ api/Element.json
  Versions – 1 error:
  → api.Element.createShadowRoot - version_added: "true" must not be the same as version_removed for webview_android
✖ api/MutationObserver.json
  Versions – 2 errors:
  → api.MutationObserver - version_added: "≤37" must not be the same as version_removed for webview_android
  → api.MutationObserver.MutationObserver - version_added: "≤37" must not be the same as version_removed for webview_android
✖ api/Window.json
  Versions – 2 errors:
  → api.Window.convertPointFromNodeToPage - version_added: "true" must not be the same as version_removed for chrome
  → api.Window.convertPointFromPageToNode - version_added: "true" must not be the same as version_removed for chrome
✖ css/at-rules/media.json
  Versions – 4 errors:
  → css.at-rules.media.-webkit-animation - version_added: "≤37" must not be the same as version_removed for webview_android
  → css.at-rules.media.-webkit-transform-2d - version_added: "≤37" must not be the same as version_removed for webview_android
  → css.at-rules.media.-webkit-transform-3d - version_added: "≤37" must not be the same as version_removed for webview_android
  → css.at-rules.media.-webkit-transition - version_added: "≤37" must not be the same as version_removed for webview_android
✖ css/properties/-webkit-mask-attachment.json
  Versions – 1 error:
  → css.properties.-webkit-mask-attachment - version_added: "≤37" must not be the same as version_removed for webview_android
✖ css/properties/display.json
  Versions – 1 error:
  → css.properties.display.display-outside.run-in - version_added: "≤37" must not be the same as version_removed for webview_android
✖ http/headers/accept-charset.json
  Versions – 1 error:
  → http.headers.Accept-Charset - version_added: "true" must not be the same as version_removed for webview_android
```

ddbeck (approved):
I'm happy with this change, provided we fix the failures. Though some of those `≤37`→`≤37` webview features might be kinda challenging to unravel. 🤷

foolip:
I've sent PRs for the API issues, only CSS remains now.

foolip:
The remaining errors are now:

```
Problems in 3 files:
✖ api/Element.json
  Versions – 1 error:
  → api.Element.createShadowRoot - version_added: "true" must not be the same as version_removed for webview_android
✖ css/properties/display.json
  Versions – 1 error:
  → css.properties.display.display-outside.run-in - version_added: "≤37" must not be the same as version_removed for webview_android
✖ http/headers/accept-charset.json
  Versions – 1 error:
  → http.headers.Accept-Charset - version_added: "true" must not be the same as version_removed for webview_android
```

The first is https://github.com/mdn/browser-compat-data/pull/9611 but the two that remain are a bit trickier.

css.properties.display.display-outside.run-in can probably be removed, but it was in IE, so... not sure?

The `Accept-Charset` header feature would be a simple removal if it were an API, but I wonder if MDN intends to document HTTP headers a bit more exhaustively, not only covering what could currently be sent or interpreted by browsers?

foolip:
I've tested http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=9132 in IE 11 and Edge 13 to confirm it was supported in IE but not Edge.

ddbeck:
OK, so we're down to this one?

> The `Accept-Charset` header feature would be a simple removal if it were an API, but I wonder if MDN intends to document HTTP headers a bit more exhaustively, not only covering what could currently be sent or interpreted by browsers?

For headers broadly, this is a long-running question. For example: https://discourse.mozilla.org/t/bcd-and-want-digest-digest-http-headers/43340/2.

More narrowly, I think there's a bit of a content question about whether there should be a page for `Accept-Charset`. It's relevance is probably limited to it being [forbidden](https://fetch.spec.whatwg.org/#forbidden-header-name). I've started asking around to figure out if we can mostly scrub `Accept-Charset` from MDN and eliminate this page.

foolip:
Yep, that's the only one left, except for `webkitCreateShadowRoot` which is still in an open PR.

Thanks for asking around about removing this!

ddbeck:
I've opened https://github.com/mdn/content/pull/4250 to remove `Accept-Charset`. If it's approved (not a given, but I think I've made a good case), then I'll drop it from BCD too.

foolip:
All issues have been fixed now, so this is ready to go!

ddbeck (approved):
Thank you! 🎉

PR #9611, opened 2021-03-25
Update when webkitCreateShadowRoot was removed in Chromium

foolip:
These tests were used:
http://staging-dot-mdn-bcd-collector.appspot.com/tests/api/Element/createShadowRoot
http://staging-dot-mdn-bcd-collector.appspot.com/tests/api/Element/webkitCreateShadowRoot

Chrome 34, 35 and 36 were tested to confirm the versions and that the
two did indeed overlap in Chrome 35, it wasn't a straight rename.

The data was mirrored for other Chromium-based browsers, achieving the
real point of this, namely avoiding version_added==version_removed for
the prefixed entry for webview_android.

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -2553,7 +2553,7 @@
               },
               {
                 "version_added": "25",
-                "version_removed": true,
+                "version_removed": "36",
                 "prefix": "webkit"
               }
             ],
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Thank you! 🎉

(Almost there!)

PR #9621, opened 2021-03-26
Add spec URLs for SVG "text" element

sideshowbarker:


Elchi3 (changes requested):
There are many references here to attribute specs that don't look like they are relevant for the `<text>` element. For example, the link https://drafts.fxtf.org/filter-effects/#element-attrdef-fedropshadow-dy is for `<feDropShadow>` and not for `<text>`.

sideshowbarker:
> There are many references here to attribute specs that don't look like they are relevant for the `<text>` element. For example, the link [drafts.fxtf.org/filter-effects/#element-attrdef-fedropshadow-dy](https://drafts.fxtf.org/filter-effects/#element-attrdef-fedropshadow-dy) is for `<feDropShadow>` and not for `<text>`.

oops — thanks, removed now

Elchi3 (approved):

PR #9622, opened 2021-03-26
Updates api.Blob.text in Opera Browser

### LINKED ISSUE #9469: Issue with "Blob.text()" - Opera is supporting blob.text()
In here https://github.com/mdn/browser-compat-data/blob/master/api/Blob.json it shows that Opera is not supporting `Blob.text()`
I expected it to see that Opera is supporting this method
I have tested this code block in Opera's developer console and it works.

```
const blob = new Blob([JSON.stringify({hello: "world"}, null, 2)], {type : 'application/json'});
const text = await blob.text();
```
#### Can you link to any release notes, bugs, pull requests, or MDN pages related to this?
https://developer.mozilla.org/en-US/docs/Web/API/Blob/text
https://developer.mozilla.org/en-US/docs/Web/API/Blob

lucalves:
Hello everyone, everything good?

Regarding issue #9469, the `.text()` function of `api.Blob` is working on version 75 of Opera.

I carried out the consultation on the website https://mdn-bcd-collector.appspot.com/ and I had the result:

opera

And I tested it on Opera v75 mentioned in the problem above and received feedback:

blob

However, I was unable to perform a test on previous versions to see if there is a version before 75 that was already working.

Fixes #9469.

### DIFF
--- a/api/Blob.json
+++ b/api/Blob.json
@@ -347,7 +347,7 @@
               "version_added": false
             },
             "opera": {
-              "version_added": "75"
+              "version_added": "63"
             },
             "opera_android": {
               "version_added": "54"
--- a/api/Blob.json
+++ b/api/Blob.json
@@ -347,7 +347,7 @@
               "version_added": false
             },
             "opera": {
-              "version_added": false
+              "version_added": "75"
             },
             "opera_android": {
               "version_added": "54"


api/Blob.json
```diff
@@ -347,7 +347,7 @@
               "version_added": false
             },
             "opera": {
-              "version_added": false
+              "version_added": "75"
```
foolip, line 350:
```suggestion
              "version_added": "63"
```

I tested https://mdn-bcd-collector.appspot.com/tests/api/Blob/text in Opera 62 and 63 on BrowserStack to confirm this. This is what you get from `npm run mirror -- -m always opera api.Blob.text`, since Opera 63 is baed on Blink 76.

foolip (changes requested):

PR #9623, opened 2021-03-26
-moz-submit-invalid behind a flag

rachelandrew:
Working on https://github.com/mdn/content/issues/3446

This feature is now behind a flag. I wasn't sure if I should do a separate block with version_removed for the unflagged version. If so let me know and I'll update it.

### DIFF
--- a/css/selectors/-moz-submit-invalid.json
+++ b/css/selectors/-moz-submit-invalid.json
@@ -17,7 +17,7 @@
             },
             "firefox": {
               "version_added": "4",
-              "notes": "From Firefox 88 the feature has been placed behind a flag. <a href='https://bugzil.la/1693969'>Bug: 1693969</a>",
+              "notes": "From Firefox 88 the feature has been placed behind a flag. See <a href='https://bugzil.la/1693969'>bug 1693969</a>.",
               "flags": [
                 {
                   "type": "preference",
@@ -28,7 +28,7 @@
             },
             "firefox_android": {
               "version_added": "4",
-              "notes": "From Firefox 88 the feature has been placed behind a flag. <a href='https://bugzil.la/1693969'>Bug: 1693969</a>",
+              "notes": "From Firefox 88 the feature has been placed behind a flag. See <a href='https://bugzil.la/1693969'>bug 1693969</a>.",
               "flags": [
                 {
                   "type": "preference",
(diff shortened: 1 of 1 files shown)


ddbeck:
> I wasn't sure if I should do a separate block with version_removed for the unflagged version. If so let me know and I'll update it.

Yes, two support statements would be ideal. The first from version 88 with the flag data (i.e., an open-ended `"version_added": "88"`) and the second without the flag (i.e., from version 4 and closed-ended with `"version_removed": "88"`). Thank you!

rachelandrew:
@ddbeck I've made that change

css/selectors/-moz-submit-invalid.json
```diff
@@ -15,12 +15,40 @@
...
-            "firefox": {
-              "version_added": "4"
-            },
-            "firefox_android": {
-              "version_added": "4"
-            },
+            "firefox": [
+              {
+                "version_added": "88",
+                "notes": "From Firefox 88 the feature has been placed behind a flag. See <a href='https://bugzil.la/1694129'>bug 1694129</a>.",
```
ddbeck, line 21: Doing something little remarkable here, so leaving a note for posterity: ordinarily, I'd say to leave this note out (or to only have the `See bug` portion). But this is an odd case of something going behind a pref after being unpreffed, so I'll leave it as is.

css/selectors/-moz-submit-invalid.json
```diff
@@ -15,12 +15,40 @@
...
+                ]
+              },
+              {
+                "version_added": "4",
+                "version_removed": "88"
+              }
+            ],
+            "firefox_android": [
+              {
+                "version_added": "88",
```
ddbeck, line 37: Oh, I'm sorry, @rachelandrew but I just remembered something significant: I learned last week that Firefox for Android doesn't provide access to preferences in regular releases (see https://github.com/mdn/browser-compat-data/issues/9679). Can we remove this support statement?

If you still think we ought to provide the bug link, we could add a note to the other support statement saying something like, `From version 88, the feature has been withdrawn. See <a href='https://bugzil.la/1694129'>bug 1694129</a>.`

ddbeck (changes requested):

css/selectors/-moz-submit-invalid.json
chrisdavidmills: @rachelandrew hi! Can you finalize this one?

ddbeck (approved):
Thank you! 🎉

PR #9625, opened 2021-03-26
Update versions for FormData.append() filename argument support

foolip:
This was added in WebKit trunk version 535.7.0:
https://trac.webkit.org/changeset/97274/webkit
https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Configurations/Version.xcconfig?rev=97274

Versions are inferred from that, which isn't 100% reliable, but being
off by a version here is not a big issue.

### DIFF
--- a/api/FormData.json
+++ b/api/FormData.json
@@ -203,10 +203,10 @@
             "description": "<code>append</code> with filename",
             "support": {
               "chrome": {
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Thank you! 🎉

PR #9626, opened 2021-03-26
Clean up the data for navigator.getBattery()

foolip:
Summary of changes:
 - Remove navigator.battery, which has been removed since 2013/2016 in
   Chrome and Firefox respectively.
 - Treat BatteryManager as only supported since navigator.getBattery()
   was supported, making the data consistent but technically wrong.
 - Simplify Firefox support to only represent when it was on by default
   for web content. Flags before that and chrome-only after is removed.
 - Remove claims of prefixes on BatteryManager, the only prefixes here
   have been navigator.mozBattery and navigator.webkitBattery.
 - Don't claim it's deprecated, https://w3c.github.io/battery/ is still
   a "real" spec and Chromium hasn't deprecated it.
 - Scope the chargingTime/dischargingTime notes to just those entries,
   and treat it as fixed in Chromium 42 (see below).

The chargingTime/dischargingTime issue was fixed in Chromium 42:
https://bugs.chromium.org/p/chromium/issues/detail?id=401553
https://storage.googleapis.com/chromium-find-releases-static/13e.html#13e397e55501bee3903bfe46edfd56f2442c1f11

Data originally from https://github.com/mdn/browser-compat-data/pull/1102.
There was no evidence for or discussion on the webview_android version
being 40. All existing versions were treated as untrustworthy and data
was mirrored to all Chromium browsers.

### DIFF
--- a/api/BatteryManager.json
+++ b/api/BatteryManager.json
@@ -8,55 +8,19 @@
             "version_added": "38"
           },
           "chrome_android": {
-            "version_added": "38",
-            "notes": "Values for <a href='https://developer.mozilla.org/docs/Web/API/BatteryManager/chargingTime'><code>BatteryManager.chargingTime</code></a> and <a href='https://developer.mozilla.org/docs/Web/API/BatteryManager/dischargingTime'><code>BatteryManager.dischargingTime</code></a> are always equal to <code>Infinity</code>."
+            "version_added": "38"
           },
           "edge": {
             "version_added": "79"
           },
-          "firefox": [
-            {
-              "version_added": "16",
-              "version_removed": "52"
-            },
-            {
-              "version_added": "10",
-              "version_removed": "16",
-              "prefix": "moz"
-            },
-            {
-              "version_added": "52",
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "dom.battery.enabled"
-                }
-              ],
-              "notes": "Disabled by default in Firefox 10, but can be enabled setting the preference <code>dom.battery.enabled</code> to <code>true</code>. Starting with Firefox 11, <code>mozBattery</code> is enabled by default. The Battery API is currently supported on Android, Windows, and Linux with UPower installed. Support for macOS is available starting with Firefox 18. Firefox also provides support for the deprecated <a href='https://developer.mozilla.org/docs/Web/API/Navigator/battery'><code>navigator.battery</code></a>."
-            },
-            {
-              "version_added": "72",
-              "notes": "From this version onwards, <code>BatteryManager</code> is only available in chrome/privileged code."
-            }
-          ],
-          "firefox_android": [
-            {
-              "version_added": "10",
-              "version_removed": "16",
-              "prefix": "moz"
-            },
-            {
-              "version_added": "16",
-              "version_removed": "52",
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "dom.battery.enabled"
-                }
-              ],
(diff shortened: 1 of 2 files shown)


foolip:
Reviewers, there are two things here I expect some might disagree with and are worth flagging.

First, should `navigator.battery` and its prefixed forms instead be treated as aliases of `navigator.getBattery()` and merged into that entry? I didn't do this because it was changed from an attributes to an async method, and I'm not sure how to avoid the impression that there was a `navigator.battery()`. If we ever add types to BCD, this entry would be problematic.

Second, I nuked a lot of notes for Firefox. They don't seem relevant to anyone today, but I'm not sure, should some of it be preserved to help people understand why this was remove from Firefox?

api/BatteryManager.json
```diff
@@ -73,17 +37,16 @@
...
-            "notes": "Values for <a href='https://developer.mozilla.org/docs/Web/API/BatteryManager/chargingTime'><code>BatteryManager.chargingTime</code></a> and <a href='https://developer.mozilla.org/docs/Web/API/BatteryManager/dischargingTime'><code>BatteryManager.dischargingTime</code></a> are always equal to <code>Infinity</code>."
+            "version_added": "3.0"
           },
           "webview_android": {
-            "version_added": "40"
+            "version_added": "38"
           }
         },
         "status": {
           "experimental": false,
```
ddbeck, line 47: I'm fine with undeprecating this, but it's single-implementation now, so let's flip this? Probably ought to be done for this and its descendants.

```suggestion
          "experimental": true,
```

ddbeck (changes requested):
> First, should `navigator.battery` and its prefixed forms instead be treated as aliases of `navigator.getBattery()` and merged into that entry?

No, I think you've made the right choice here. I think it's fine to treat this as a materially different feature which was removed. It so happens that it was removed sufficiently long ago that it's now irrelevant. Let's put it in the 🗑️.

That said, would you mind opening a PR on mdn/content to redirect https://developer.mozilla.org/en-US/docs/Web/API/Navigator/battery to https://developer.mozilla.org/en-US/docs/Web/API/Navigator/getBattery?

> Second, I nuked a lot of notes for Firefox. They don't seem relevant to anyone today, but I'm not sure, should some of it be preserved to help people understand why this was remove from Firefox?

I think you made the right call here too. The prefixes are ancient history and would be removed if they were standalone features. The prefs would also be irrelevant by now. The chrome code stuff is off-topic and could also be removed legitimately. What's left makes sense. 👍

I do have one suggestion in a line comment, about the `experimental` status for all of this.

api/BatteryManager.json
foolip: I can't find where we had the discussion, but this is a perfect example of the case where I think calling single-implementation feature experimental seems inaccurate. There really is no experimenting going on, and web developers shouldn't expect the behavior to change. The spec probably won't change much either.

Web developers do need to be aware that the feature isn't supported for a large fraction of their users, but that's something else which we don't have a flag for in BCD, I think.

I'll set `experimental` to true if that's the current guidelines, but how should we go about changing that?

ddbeck: We never actually resolved https://github.com/mdn/browser-compat-data/issues/6905 with a real definition. It's more than just a guideline, because it also impacts the schema docs.

I think the interim approach has been, basically, experimental is a proxy for whether or not there's multiple implementations. I guess I don't care that much and can merge without it, but…

I do have an incomplete rough-draft of a guideline that captures more nuance like we discussed in #6905, but it's got some problems (I ought to open a PR to actually discuss that). To keep symmetry with the other removal guidelines, I've got something along the lines of it being OK to flip `experimental` to `false` for features which have not been introduced into a new browser in the last two years, which doesn't really help here anyway. Edge adds a wrinkle to a lot of stuff I hadn't really thought through when we last discussed the prospect of revising experimental (though I guess it's a wrinkle that would go away in January 2022).

ddbeck (approved):

api/BatteryManager.json
foolip: @ddbeck I'd be happy to review that, however unsatisfactory you find it. If I were to state my position shortly, it would be that old and experimental are incompatible. If something first shipped over two years ago and there isn't much activity on the spec or implementation side, then it's not experimental.

PR #9630, opened 2021-03-26
Mirror Android data for NodeFilter and TreeWalker APIs

foolip:
These APIs are a very old part of the platform and can be safely assumed
to have been in Chrome for Android and Samsung Internet based on their
support early in Chrome and Safari.


ddbeck (approved):
Looks reasonable to me. Thank you! 🎉

PR #9634, opened 2021-03-27
Remove the api.WorkerGlobalScope.console entry

foolip:
This duplicates the api.Console.worker_support entry, which is more
accurate.

http://mdn-bcd-collector.appspot.com/tests/api/Console/log was used to
confirm the following version_added:
 - Chrome 31
 - Firefox 29
 - IE 10 (so Edge not tested)
 - Opera 12.16
 - Safari 6.1 or 6.2 (in 6.2 but not in 6, 6.1 not available)

In the preceding versions, an exception like "console is not defined"
was thrown, it wasn't `console.log` that was somehow missing.

The WebView version was inferred from Chromium 31.

### DIFF
--- a/api/WorkerGlobalScope.json
+++ b/api/WorkerGlobalScope.json
@@ -47,56 +47,6 @@
           "deprecated": false
         }
       },
-      "console": {
-        "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/WorkerGlobalScope/console",
-          "support": {
-            "chrome": {
-              "version_added": "4"
-            },
-            "chrome_android": {
-              "version_added": true
-            },
-            "edge": {
-              "version_added": "12"
-            },
-            "firefox": {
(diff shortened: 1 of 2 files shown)


api/WorkerGlobalScope.json
```diff
@@ -47,56 +47,6 @@
...
-            },
-            "chrome_android": {
-              "version_added": true
-            },
-            "edge": {
-              "version_added": "12"
-            },
-            "firefox": {
-              "version_added": "29",
-              "notes": "Before Firefox 30, the console would be of type <code>WorkerConsole</code> instead of <code>Console</code>."
```
foolip, line 65: This is not an important note, the name of the interface doesn't matter when using `console.log`. By now, `console` is a namespace, so there's not even anything one can do with an interface object which will work in all browsers.

ddbeck (approved):
Thank you, @foolip! 🎉

PR #9637, opened 2021-03-27
Update Samsung Internet support for Promise.allSettled()

### LINKED ISSUE #9636: allSettled() - Should be supported by Samsung Internet
I have the feeling that the [support for Promise.allSettled](https://caniuse.com/mdn-javascript_builtins_promise_allsettled) is incorrect. Chrome has supported this since version 76, and since Samsung Internet is based on Chromium this should be supported since at least version 12.0.1.47 (according to [Wikipedia](https://en.wikipedia.org/wiki/Samsung_Internet#History))

jonkoops:
This PR updates the support for the `Promise.allSettled()` method in Samsung Internet. Samsung Internet version 12.0.1.47 is using Chromium version 79.3945, which has support for this method.

See the [release notes](https://developer.samsung.com/internet/release-note.html) of Samsung Internet for more information. Closes #9636

### DIFF
--- a/javascript/builtins/Promise.json
+++ b/javascript/builtins/Promise.json
@@ -203,7 +203,7 @@
                 "version_added": "13"
               },
               "samsunginternet_android": {
-                "version_added": false
+                "version_added": "12.0"
               },
               "webview_android": {
                 "version_added": "76"


foolip:
Thanks @jonkoops, and welcome to BCD! Samsung Internet 12.0 was based on Blink 79 and the previous 11.2 was based on Blink 11.2, so this seems correct.

foolip (approved):

PR #1487, opened 2018-03-17
Add compat data for FormData Web API

theCalibrius:

PR #9640, opened 2021-03-27
Remove the api.FormData.SupportForOf entry (dupe of @@iterator)

foolip:
These entries were added separately at different times:
- SupportForOf in https://github.com/mdn/browser-compat-data/pull/1487
- @@iterator in https://github.com/mdn/browser-compat-data/pull/8104

The data is the same (or missing) except for Edge and Safari.

The @@iterator entry is based on mdn-bcd-collector results, but
https://software.hixie.ch/utilities/js/live-dom-viewer/saved/9119 was
used to confirm support of for...of loops in Edge 18 (but not 17) and in
Safari 11.1 (but not 11).

In other words, no data needed updating on the @@iterator entry.

### DIFF
--- a/api/FormData.json
+++ b/api/FormData.json
@@ -99,54 +99,6 @@
           }
         }
       },
-      "SupportForOf": {
-        "__compat": {
-          "description": "Support of <code>for...of</code>",
-          "support": {
-            "chrome": {
-              "version_added": "50"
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Thank you! 🎉

PR #9650, opened 2021-03-29
browserSettings.ftpProtocolEnabled readonly in FF88

hamishwillee:
This adds a note from FF88 the setting is read only. This comes from https://bugzilla.mozilla.org/show_bug.cgi?id=1626365.

### DIFF
--- a/webextensions/api/browserSettings.json
+++ b/webextensions/api/browserSettings.json
@@ -114,7 +114,7 @@
               },
               "firefox": {
                 "version_added": "72",
-                "notes": "From Firefox version 88 this setting is read-only (<a href='https://bugzil.la/1363010'>bug 1626365</a>)."
+                "notes": "From version 88, this setting is read-only (see <a href='https://bugzil.la/1626365'>bug 1626365</a>)."
               },
               "firefox_android": {
                 "version_added": false
--- a/webextensions/api/browserSettings.json
+++ b/webextensions/api/browserSettings.json
@@ -114,7 +114,7 @@
               },
               "firefox": {
                 "version_added": "72",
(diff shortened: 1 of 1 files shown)


webextensions/api/browserSettings.json
```diff
@@ -113,7 +113,8 @@
                 "version_added": false
               },
               "firefox": {
-                "version_added": "72"
+                "version_added": "72",
+                "notes": "From Firefox version 88 this setting is read-only (<a href='https://bugzil.la/1363010'>bug 1626365</a>)."
```
ddbeck, line 117: A few things here:

- There's a mismatch here between the bug text number and URL. This is the cause of the CI failure. I'm assuming you meant the text, not the URL, but please double check this.
- Generally, prefer "version XY" or "Browser XY" over "Browser version XY".
- Other nitpicky things about commas and referring to bug links I would probably wouldn't have otherwise bothered with, but since I'm already here…

```suggestion
                "notes": "From version 88, this setting is read-only (see <a href='https://bugzil.la/1626365'>bug 1626365</a>)."
```

ddbeck (changes requested):
A few things to get this moving, @hamishwillee. Thank you!

hamishwillee:
Thanks @ddbeck ! Dunno how I managed to update the bug link text but not the URL. Sorry! I had Friday off, hence the delay to seeing the CI failure.

7751894020 (approved):

7751894020 (approved):

7751894020:
webextensions/api/browserSettings.json

ddbeck (approved):
Thank you! 🎉

PR #9652, opened 2021-03-29
Mark customized built-in element support as shipped in Chromium 67

foolip:
This was controlled by the CustomElementsBuiltin feature, enabled here:
https://bugs.chromium.org/p/chromium/issues/detail?id=648828
https://storage.googleapis.com/chromium-find-releases-static/4a2.html#4a2b6e0a71e6d8f664697210b5fb407281ba0941

The data was slightly inconsistent, off-by-one in both directions.


ddbeck (approved):
Thank you! 🎉

PR #9658, opened 2021-03-29
Mark bug w/ text-decoration-thickness fixed from Chromium 89

36degrees:
Following on from https://github.com/mdn/browser-compat-data/pull/7625.

[Chromium bug 1154537][1] – which meant that the `text-decoration-thickness` property did not work unless either `text-underline-offset` was set to something other than `auto` or `text-decoration-color` was set to something other than `currentColor` – has been fixed as of Chromium 89.

This has been verified to fix the issue (using https://codepen.io/36degrees/pen/JjRKqYv as a test case) in:

- Chrome 89
- Edge 89
- Opera 75, which is [based on Chromium 89][2]

I have not been able to verify it as fixed in Chrome or WebView on Android, but based on my understanding of how Chrome works I’d expect the issue to be resolved there as well.

[1]: https://crbug.com/1154537
[2]: https://blogs.opera.com/desktop/2021/01/opera-75-developer/

### Chrome 88
Chrome 88
### Chrome 89
Chrome 89

### Edge 88
Edge 88
### Edge 89
Edge 89

### Opera 74
Opera 74
### Opera 75
Opera 75

---

- [x] Summarize your changes

### DIFF
--- a/css/properties/text-decoration-thickness.json
+++ b/css/properties/text-decoration-thickness.json
@@ -6,18 +6,33 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/text-decoration-thickness",
           "spec_url": "https://drafts.csswg.org/css-text-decor-4/#text-decoration-width-property",
           "support": {
-            "chrome": {
-              "version_added": "87",
-              "notes": "The <code>text-decoration-thickness</code> property does not work unless either <code>text-underline-offset</code> is set to something other than <code>auto</code> or <code>text-decoration-color</code> is set to something other than <code>currentColor</code>. See <a href='https://crbug.com/1154537'>Chromium bug 1154537</a>"
-            },
-            "chrome_android": {
-              "version_added": "87",
-              "notes": "The <code>text-decoration-thickness</code> property does not work unless either <code>text-underline-offset</code> is set to something other than <code>auto</code> or <code>text-decoration-color</code> is set to something other than <code>currentColor</code>. See <a href='https://crbug.com/1154537'>Chromium bug 1154537</a>"
-            },
-            "edge": {
-              "version_added": "87",
-              "notes": "The <code>text-decoration-thickness</code> property does not work unless either <code>text-underline-offset</code> is set to something other than <code>auto</code> or <code>text-decoration-color</code> is set to something other than <code>currentColor</code>. See <a href='https://crbug.com/1154537'>Chromium bug 1154537</a>"
-            },
+            "chrome": [
+              {
+                "version_added": "87",
+                "notes": "The <code>text-decoration-thickness</code> property does not work unless either <code>text-underline-offset</code> is set to something other than <code>auto</code> or <code>text-decoration-color</code> is set to something other than <code>currentColor</code>. See <a href='https://crbug.com/1154537'>Chromium bug 1154537</a>"
+              },
+              {
+                "version_added": "89"
+              }
+            ],
+            "chrome_android": [
+              {
+                "version_added": "87",
+                "notes": "The <code>text-decoration-thickness</code> property does not work unless either <code>text-underline-offset</code> is set to something other than <code>auto</code> or <code>text-decoration-color</code> is set to something other than <code>currentColor</code>. See <a href='https://crbug.com/1154537'>Chromium bug 1154537</a>"
+              },
+              {
+                "version_added": "89"
+              }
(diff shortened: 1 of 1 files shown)


ddbeck:
Thanks for this PR, @36degrees!.

We'll need to make some changes to the way this is represented, but I'm not sure how exactly. I'll start with the substantive question:

I'm not expert on this feature, so I thought I'd turn this question to you. Do you have a sense of how severe a departure the previous behavior was from expectations? Is this something than many developers trying to use `text-decoration-thickness` were likely to encounter?

The reason I ask is because we have two ways of handling notes:

1. Creating separate support statements like you've done here, where the note is only attached to the relevant support statements.
2. Having one support statement, with some context-setting wording, such as "Before version 89…".

Whether we go with option 1 or 2 depends on the significance of the change. In other words, does it represent a discontinuity in support, or a limited variation? For example, for a web API, we might use option 1 (separate support statements) when the signature of a method changes, but use option 2 for a subtle change in the type of the value returned by a method.

36degrees:
> Do you have a sense of how severe a departure the previous behavior was from expectations? Is this something than many developers trying to use `text-decoration-thickness` were likely to encounter?

I think it's relatively severe, as if you tried to use `text-decoration-thickness` on its own, without also setting `text-underline-offset` or `text-underline-color`, it wouldn't do anything, and thus it'd appear like it wasn't supported at all in browsers with this bug.

Additionally, unless you want to use a different color for your text decoration, or to change the offset of the underline, it can make using `text-decoration-thickness` in the browsers with this bug more complicated.

> Whether we go with option 1 or 2 depends on the significance of the change.

I'm possibly thinking too much of how the data from here is used on caniuse.com, but I'd expect the note to only be associated with the affected versions – especially as these are an evergreen browser, and so before long this bug will be ~irrelevant for most users unless they were specifically looking at Chrome 87/88.

css/properties/text-decoration-thickness.json
```diff
@@ -6,18 +6,33 @@
...
-              "version_added": "87",
-              "notes": "The <code>text-decoration-thickness</code> property does not work unless either <code>text-underline-offset</code> is set to something other than <code>auto</code> or <code>text-decoration-color</code> is set to something other than <code>currentColor</code>. See <a href='https://crbug.com/1154537'>Chromium bug 1154537</a>"
-            },
-            "edge": {
-              "version_added": "87",
-              "notes": "The <code>text-decoration-thickness</code> property does not work unless either <code>text-underline-offset</code> is set to something other than <code>auto</code> or <code>text-decoration-color</code> is set to something other than <code>currentColor</code>. See <a href='https://crbug.com/1154537'>Chromium bug 1154537</a>"
-            },
+            "chrome": [
+              {
+                "version_added": "87",
```
ddbeck, line 11: Since this note is not relevant beyond version 87, then let's not let this note be open-ended.

```suggestion
                "version_added": "87",
                "version_removed": "89",
```

css/properties/text-decoration-thickness.json
```diff
@@ -6,18 +6,33 @@
...
+                "notes": "The <code>text-decoration-thickness</code> property does not work unless either <code>text-underline-offset</code> is set to something other than <code>auto</code> or <code>text-decoration-color</code> is set to something other than <code>currentColor</code>. See <a href='https://crbug.com/1154537'>Chromium bug 1154537</a>"
+              },
+              {
```
ddbeck, line 14: Sort this support statement to the beginning of the array (we go with most recent and relevant first)

ddbeck (changes requested):
Thanks for the extra info, @36degrees.

In that case, let's let this stand as two entries. To merge this, we do need to make an adjustment to the now-historic support statement and sort the most recent statement first. Thank you for your help on this!

36degrees:
@ddbeck thanks for the prompt review – that all makes sense. I believe I've addressed your comments, and I'll try and remember for the future!

36degrees:
Would it make sense to mark the versions with the bug as a `partial_implementation`? 🤔

ddbeck:
> Would it make sense to mark the versions with the bug as a `partial_implementation`? 🤔

Yeah, that would probably also be a good addition.

36degrees:
@ddbeck done!

ddbeck (approved):
Looks great. Thank you, @36degrees! 🎉

PR #9661, opened 2021-03-29
Remove prefix for element.requestFullscreen() in Presto

foolip:
The Fullscreen API was actually shipped unprefixed in Presto.

Tested with http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=9122 in Opera 12.1.


ddbeck (approved):
Thank you! 🎉

PR #9664, opened 2021-03-29
Mark WebKitMutationObserver as still supported

foolip:
https://trac.webkit.org/changeset/143386/webkit added support for
MutationObserver in addition to WebKitMutationObserver, but support for
the latter was never removed.

Confirmed supported in Chrome 89 and Safari 14.

### DIFF
--- a/api/MutationObserver.json
+++ b/api/MutationObserver.json
@@ -10,7 +10,6 @@
             },
             {
               "version_added": "18",
-              "version_removed": "26",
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Oh this is neat. I was somewhat skeptical that there would be much material impact to contemporary developers by fixing the same-added/removed versions the ≤ values. This proves me wrong! Thank you! 🎉

PR #9667, opened 2021-03-29
Update -webkit-mask-attachment removal versions

foolip:
This was removed in WebKit trunk version 537.20:
https://trac.webkit.org/changeset/136080/webkit
https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Configurations/Version.xcconfig?rev=136080

The Safari versions can be determined with fairly high confidence based
on this. When this was added in WebView Android is a little bit more
uncertain, but doesn't really matter since this entry will now be
eligible for removal as an obsolete feature.

### DIFF
--- a/css/properties/-webkit-mask-attachment.json
+++ b/css/properties/-webkit-mask-attachment.json
@@ -32,18 +32,20 @@
               "version_added": false
             },
             "safari": {
-              "version_added": "4"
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Makes sense. Thank you! 🎉

PR #9670, opened 2021-03-30
Presentation API mark version removal

hamishwillee:
This adds version removal information for Firefox on the [Presentation API](https://developer.mozilla.org/en-US/docs/Web/API/Presentation_API). 
- FF88 removes the API for Android: https://bugzilla.mozilla.org/show_bug.cgi?id=1697680
- FF61 removed it for desktop https://bugzilla.mozilla.org/show_bug.cgi?id=1371346

Associated docs tracking is in https://github.com/mdn/content/issues/3461

### DIFF
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -1896,6 +1896,7 @@
             },
             "firefox": {
               "version_added": "51",
+              "version_removed": "61",
               "flags": [
                 {
                   "type": "preference",
@@ -1906,6 +1907,7 @@
             },
             "firefox_android": {
               "version_added": "51",
+              "version_removed": "88",
               "flags": [
                 {
                   "type": "preference",
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -1907,7 +1907,7 @@
             },
             "firefox_android": {
               "version_added": "51",
-              "version_removed": "88",
+              "version_removed": "79",
               "flags": [
                 {
                   "type": "preference",
--- a/api/Presentation.json
+++ b/api/Presentation.json
@@ -15,6 +15,7 @@
           },
           "firefox": {
             "version_added": "51",
+            "version_removed": "61",
             "flags": [
               {
                 "type": "preference",
@@ -25,6 +26,7 @@
           },
           "firefox_android": {
             "version_added": "51",
+            "version_removed": "88",
             "flags": [
               {
                 "type": "preference",
@@ -76,6 +78,7 @@
             },
             "firefox": {
               "version_added": "51",
+              "version_removed": "61",
               "flags": [
                 {
                   "type": "preference",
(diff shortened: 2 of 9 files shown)


hamishwillee:
Actually according to https://bugzilla.mozilla.org/show_bug.cgi?id=1697680#c14 this never really worked on Fenix. Did that come in after version 51? If before we could just say `false` from 51 for Android

ddbeck (changes requested):
The desktop changes look good to me. 👍

> Actually according to https://bugzilla.mozilla.org/show_bug.cgi?id=1697680#c14 this never really worked on Fenix. Did that come in after version 51? If before we could just say `false` from 51 for Android

For Android, I think we should show them as removed in version 79. As far as I can tell, the main release channel for Fenix doesn't ship access to `about:config`. I've opened #9679 to discuss this problem generally, but in this case, we can do the right thing and mark them as removed in 79 instead of 88.

hamishwillee:
Thanks very much! Changed.

ddbeck (approved):
Thank you, @hamishwillee!

(See, that one wasn't too bad. 😅)

hamishwillee:
@ddbeck YOu say that, but there is now a [later comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1697680#c16).

> BTW, I doubt Firefox (and Fenix) ever really exposed navigator.presentation per my mozregression test and the current Fenix 87 behavior?

If this is correct then we should put `false` for Android, all parts (since you access presentation API from `navigator.presentation`. Do you want me to follow that up (try test it). My temptation is to leave it alone - won't be long before the old versions are more or less irrelevant.

ddbeck:
@hamishwillee Oh no, I investigated it and I have regrets. With https://mdn-bcd-collector.appspot.com/tests/api/Navigator/presentation and BrowserStack, I checked this with and without the flag in several versions. With the flag on (and only on), the API exists in Firefox 51 (desktop) and it doesn't stop existing until Firefox 88. 😬

hamishwillee:
@ddbeck Arrggg, so "FF61 removed it for desktop https://bugzilla.mozilla.org/show_bug.cgi?id=1371346" is incorrect? Sorry, I thought it stated removal in that version :-(

The action from me being to change the FF desktop cases where it says 61 removal to FF88 right?

ddbeck:
@hamishwillee 

> Sorry, I thought it stated removal in that version :-(

No, it definitely said that. I don't think this is your fault. I think _something_ was removed, but perhaps only landed in a particular branch, got reverted, or some other selective thing happened. Weird things happen sometimes, I guess. 🤷

> The action from me being to change the FF desktop cases where it says 61 removal to FF88 right?

That's correct. Thank you!

PR #9671, opened 2021-03-30
Drop object[typemustmatch] and typeMustMatch IDL attribute

sideshowbarker:
This change deletes `typeMustMatch` from `api/HTMLObjectElement.json` and deletes `typemustmatch` from `html/elements/object.json`.

The history of `typeMustMatch`/`typemustmatch` is that it was added to the spec in 2011 in https://github.com/whatwg/html/commit/4030e71 but never got implemented across browsers and never got adopted by web developers.  So https://github.com/whatwg/html/pull/4590 dropped it from the spec in 2019, and it's now just a footnote in the “Non-conforming features” section at https://html.spec.whatwg.org/obsolete.html#attr-object-typemustmatch

So there's statistically near-zero content on the web that’s using `typeMustMatch`/`typemustmatch`, and there’s no value to continue tracking support data for it in BCD.

Related MDN content change: https://github.com/mdn/content/pull/3655

### DIFF
--- a/api/HTMLObjectElement.json
+++ b/api/HTMLObjectElement.json
@@ -1007,56 +1007,6 @@
           }
         }
       },
-      "typeMustMatch": {
-        "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLObjectElement/typeMustMatch",
-          "support": {
-            "chrome": {
-              "version_added": false
-            },
-            "chrome_android": {
-              "version_added": false
-            },
-            "edge": {
-              "version_added": false
-            },
-            "firefox": {
-              "version_added": "27",
-              "version_removed": "68"
-            },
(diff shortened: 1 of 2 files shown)


foolip (approved):

ddbeck:
A nitpicky process note:

This did not actually meet the usual criteria for removal under [our existing guideline](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-features). I'm not actually going to do anything about this because it seems silly to revert when it will satisfy the guideline in three months and the docs have already been removed. I think the occasional exception is fine, but I think it would be nice to call attention to exceptions made and document why it has happened.

foolip:
Oops, I didn't notice that Firefox 68 isn't old enough to meet the guidelines. I'll do better next time!

PR #9677, opened 2021-03-30
Add missing entries for Chrome for Android

foolip:
Each of these was found supported in Chrome for Android 88 using
mdn-bcd-collector, and manually edited to say 18 based on the earlier
support in Chrome and the fact that these APIs have no platform
dependency.


ddbeck (approved):
Looks good. Thank you! 🎉

(It has been **_0_** days since I learned about a little-used HTML element.)

PR #9678, opened 2021-03-30
Add missing entries for Firefox for Android

foolip:
BroadcastChannel and IntersectionObserver were found support in Firefox
for Android 86 using mdn-bcd-collector, but the exact version can't be
determined without exhaustive results from all Firefox for Android
versions. Instead mirror the desktop data, which is likely correct.

Also remove flags to simplify as allowed by this guideline:

### DIFF
--- a/api/BroadcastChannel.json
+++ b/api/BroadcastChannel.json
@@ -16,6 +16,9 @@
           "firefox": {
             "version_added": "38"
           },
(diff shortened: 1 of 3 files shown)


ddbeck (approved):
Looks reasonable to me. Thank you! 🎉

Issue #6070, opened 2020-05-04
font-variant-caps works in Safari and iOS Safari


ryanditjia:
[Data ref](https://github.com/mdn/browser-compat-data/blob/master/css/properties/font-variant-caps.json)
`font-variant-caps: small-caps` and `font-variant-caps: all-small-caps` work on my:

- Safari (macOS 10.15.4)
- iOS Safari (13.4.1)

And I seem to remember it working from even earlier versions.

ddbeck:
Fixed by #9789. 🎉

PR #9680, opened 2021-03-30
Update Safari support for font-variant-caps

jceipek:
# Summary
Safari has supported `font-variant-caps` for around 6 years. I updated the compatibility table accordingly.
This is my first time updating a browser support compatibility table, so I'd appreciate someone double-checking my findings.

# Data
This patch: https://trac.webkit.org/changeset/190192/webkit/trunk/Source/WebCore/css/CSSComputedStyleDeclaration.cpp
seems to be when these properties were added to WebKit:

- font-variant-ligatures
- font-variant-position
- font-variant-caps
- font-variant-numeric
- font-variant-alternates
- font-variant-east-asian

Per https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Configurations/Version.xcconfig?rev=190192 it appears that this is from webkit version 602.1.5.0.

According to the release notes, Mac Safari 9.1 added `font-variant-caps`: https://developer.apple.com/library/archive/releasenotes/General/WhatsNewInSafari/Articles/Safari_9_1.html#//apple_ref/doc/uid/TP40014305-CH10-SW1

Wikipedia's Safari version table (https://en.wikipedia.org/wiki/Safari_version_history#Safari_9) claims that Mac Safari 9.1 corresponds to webkit version 601.5.17 and iOS Safari 10 corresponds to WebKit versions 602.1.38 - 603.2.4. Note that the patch webkit version seems to be newer than the ones in the version table.

However, https://trac.webkit.org/browser/webkit/tags/Safari-601.5.7/Source/WebCore/css/CSSComputedStyleDeclaration.cpp?rev=194199 from Webkit version 601.5.7 does include the changes, as does https://trac.webkit.org/browser/webkit/tags/Safari-602.1.38/Source/WebCore/css/CSSComputedStyleDeclaration.cpp?rev=202385. Therefore, I'm reasonably confident in the versions I included in the PR.

# Related
https://github.com/mdn/browser-compat-data/issues/6070

### DIFF
--- a/css/properties/font-variant-caps.json
+++ b/css/properties/font-variant-caps.json
@@ -57,10 +57,10 @@
               "version_added": "41"
             },
             "safari": {
-              "version_added": false
+              "version_added": "9.1"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "10"
             },
             "samsunginternet_android": {
               "version_added": "6.0"


foolip:
Tests:
https://mdn-bcd-collector.appspot.com/tests/css/properties/-webkit-font-variant-ligatures
https://mdn-bcd-collector.appspot.com/tests/css/properties/font-variant-ligatures
https://mdn-bcd-collector.appspot.com/tests/css/properties/font-variant-ligatures
https://mdn-bcd-collector.appspot.com/tests/css/properties/font-variant-ligatures
https://mdn-bcd-collector.appspot.com/tests/css/properties/font-variant-ligatures

foolip:
Tests:
https://mdn-bcd-collector.appspot.com/tests/css/properties/-webkit-font-variant-ligatures
https://mdn-bcd-collector.appspot.com/tests/css/properties/font-variant-ligatures
https://mdn-bcd-collector.appspot.com/tests/css/properties/font-variant-position
https://mdn-bcd-collector.appspot.com/tests/css/properties/font-variant-caps
https://mdn-bcd-collector.appspot.com/tests/css/properties/font-variant-numeric
https://mdn-bcd-collector.appspot.com/tests/css/properties/font-variant-numeric

foolip:
Tests:
https://mdn-bcd-collector.appspot.com/tests/css/properties/-webkit-font-variant-ligatures
https://mdn-bcd-collector.appspot.com/tests/css/properties/font-variant-ligatures
https://mdn-bcd-collector.appspot.com/tests/css/properties/font-variant-position
https://mdn-bcd-collector.appspot.com/tests/css/properties/font-variant-caps
https://mdn-bcd-collector.appspot.com/tests/css/properties/font-variant-numeric
https://mdn-bcd-collector.appspot.com/tests/css/properties/font-variant-alternates
https://mdn-bcd-collector.appspot.com/tests/css/properties/font-variant-east-asian

The linked patched removes `-webkit-font-variant-ligatures`, but I've confirmed all of the unprefixed properties supported in Safari 9.1 on BrowserStack and Sauce Labs, and that `-webkit-font-variant-ligatures` was supported in Safari 8 but not in 9.1. Unfortunately I don't have access to Safari 9.0, but I think we can trust the release notes here.

For mapping WebKit trunk versions to Safari releases I usually use [`safari.json`](https://github.com/mdn/browser-compat-data/blob/main/browsers/safari.json) and [`safari_ios.json`](https://github.com/mdn/browser-compat-data/blob/main/browsers/safari_ios.json). However, in this case that would suggest Safari 10, since Safari 9.1 was based on 601.5.17 according to `safari.json`, and I see Safari 9.1.2 is based on 601.7.7 on Sauce Labs.

However, mapping trunk versions to releases isn't entirely reliable, this must be a case of backporting or similar. Safari 9.1 is correct for macOS, I'm convinced.

css/properties/font-variant-caps.json
```diff
@@ -57,10 +57,10 @@
               "version_added": "41"
             },
             "safari": {
-              "version_added": false
+              "version_added": "9.1"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "10"
```
foolip, line 63:
```suggestion
              "version_added": "9.3"
```

This is what I'd expect since the point releases of Safari and iOS go together even if the numbers are different. This has the same release date and WebKit version according to `safari_ios.json`.

But more importantly, I could confirm this using https://mdn-bcd-collector.appspot.com/tests/css/properties/font-variant-caps on the iOS 9.3 simulator on BrowserStack.

foolip (changes requested):

foolip:
@jceipek thanks for sending this PR, and welcome to BCD! It's a very well-researched contribution you have here, thank you!

Since it's supported by the same evidence you've gathered, can you make the same change for the other properties?

ddbeck:
Hi @jceipek, I was wondering, do you plan to return to this PR? There's a conflict resolved/rebase needed and some changes suggested above.

ddbeck:
Thanks for this PR, @jceipek. I've followed up with #10520, which incorporates your changes. It was just merged. 🎉

PR #9683, opened 2021-03-31
Remove old flags for CSS Grid

foolip:

The only irregular edit here is for the repeat() features of
grid-template-columns and grid-template-rows. There, the Firefox for
Android data was out of sync with Firefox, and was made to match.

### DIFF
--- a/css/properties/grid-template-columns.json
+++ b/css/properties/grid-template-columns.json
@@ -125,7 +125,7 @@
             "description": "<code>fit-content()</code>",
             "support": {
               "chrome": {
-                "version_added": "29"
+                "version_added": "57"
               },
               "chrome_android": {
                 "version_added": "57"
@@ -134,10 +134,10 @@
                 "version_added": "16"
               },
               "firefox": {
-                "version_added": "51"
+                "version_added": "52"
               },
               "firefox_android": {
-                "version_added": "51"
+                "version_added": "52"
               },
               "ie": {
                 "version_added": false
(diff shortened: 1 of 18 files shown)


css/properties/grid-template-columns.json
```diff
@@ -410,66 +248,34 @@
...
-                      "name": "layout.css.grid.enabled",
-                      "value_to_set": "true"
-                    }
-                  ]
                 }
               ],
               "firefox_android": [
                 {
-                  "version_added": "52"
+                  "version_added": "76"
```
foolip, line 255: Here's the first irregular change to look extra carefully at.

css/properties/grid-template-rows.json
```diff
@@ -466,66 +304,34 @@
...
-                      "name": "layout.css.grid.enabled",
-                      "value_to_set": "true"
-                    }
-                  ]
                 }
               ],
               "firefox_android": [
                 {
-                  "version_added": "52"
+                  "version_added": "76"
```
foolip, line 311: Here's the other irregular change to look extra carefully at.

css/properties/grid-template-columns.json
vinyldarkscratch: Looking at the data, this seems like a reasonable change to me.  The commit history indicates that changes were made to Firefox desktop that should have been made to FxA as well, but weren't.  I can't think of a reason that the data should vary between the two.

css/properties/grid-template-rows.json
vinyldarkscratch: Looks like the same case here, so we're all good!

vinyldarkscratch (approved):
Looks all good to me, thanks!

PR #9684, opened 2021-03-31
Remove old flags for CSS Flexbox

foolip:


vinyldarkscratch (approved):
LGTM, thanks for the PR!  I only saw one change that seemed a bit odd at first (the changes to the version numbers of a feature without flag data), but I see that it was to fix version consistency.

PR #9685, opened 2021-03-31
Remove old flags for Firefox CSS features

foolip:

### DIFF
--- a/css/properties/inline-size.json
+++ b/css/properties/inline-size.json
@@ -15,38 +15,12 @@
             "edge": {
               "version_added": "79"
             },
-            "firefox": [
-              {
-                "version_added": "41"
-              },
-              {
-                "version_added": "38",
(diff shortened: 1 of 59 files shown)


foolip:
Reviewers, each flag is in a separate commit, which could make it easier to verify the correctness of this.

css/properties/inline-size.json
```diff
@@ -90,11 +64,11 @@
               },
               "firefox": {
                 "prefix": "-moz-",
-                "version_added": "38"
+                "version_added": "41"
```
foolip, line 67: This change is necessary to pass the lint, and was verified by testing https://software.hixie.ch/utilities/js/live-dom-viewer/?saved=9139 in Firefox 40 and 41.

vinyldarkscratch (approved):
All looking good to me, thanks for the PR!

PR #9695, opened 2021-03-31
:user-valid and :user-invalid updates

rachelandrew:
This PR:

- adds `:user-valid`
- removes `:-moz-ui-valid`
- updates Firefox data for `:user-invalid` and fixes the alternate name (it's a pseudo-class not pseudo-element, so one colon.)

Part of https://github.com/mdn/content/issues/3455


ddbeck (approved):
This is great. Thank you! 🎉

PR #9703, opened 2021-04-01
Add Edge versions for HTMLSelectElement API

vinyldarkscratch:
This PR adds real values for Microsoft Edge for the `HTMLSelectElement` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.0).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/HTMLSelectElement

### DIFF
--- a/api/HTMLSelectElement.json
+++ b/api/HTMLSelectElement.json
@@ -263,7 +263,7 @@
               "version_added": "18"
             },
             "edge": {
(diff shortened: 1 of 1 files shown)


foolip (approved):
Makes sense since these were in IE5.

PR #9713, opened 2021-04-01
Remove irrelevant flags in Firefox for AnimationTimeline API

vinyldarkscratch:
This PR removes irrelevant flag data for Firefox and Firefox Android for the `AnimationTimeline` API as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-flag-data).

### DIFF
--- a/api/AnimationTimeline.json
+++ b/api/AnimationTimeline.json
@@ -26,17 +26,6 @@
                   "value_to_set": "true"
                 }
               ]
-            },
-            {
-              "version_added": "48",
-              "version_removed": "63",
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "dom.animations-api.core.enabled",
-                  "value_to_set": "true"
-                }
-              ]
             }
           ],
           "firefox_android": [
@@ -49,17 +38,6 @@
                 {
                   "type": "preference",
                   "name": "dom.animations-api.timelines.enabled",
-                  "value_to_set": "true"
-                }
-              ]
-            },
-            {
-              "version_added": "48",
-              "version_removed": "63",
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "dom.animations-api.core.enabled",
                   "value_to_set": "true"
                 }
               ]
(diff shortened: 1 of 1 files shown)


foolip:
@vinyldarkscratch have you fixed the bug in the script that I speculated about in https://github.com/mdn/browser-compat-data/pull/9193 or do I need to carefully check the dates of the remaining data here?

vinyldarkscratch:
I haven't modified the script per the comments in #9193.  For the time being, I'm double-checking the versions and their dates, and discarding changes that don't adhere to the guidelines.

foolip:
I see, thanks! In this case since some flags are still left, removing just the oldest statements for a feature doesn't seem very worthwhile. It is allowed per the guideline though.

@ddbeck WDYT?

ddbeck:
I feel like I'm missing a key thing here: why?

I mean, I _can_ referee the guidelines, but I might be able to do a bit more than that, if I can understand the motivation for this removal. What's the outcome you want from this?

The flag-removal guideline came about in response to a specific goal, to make it easier to reconcile conflicting or messy data and all the earliest removals were in that context. This one's not clearly connected to any context that I know of, so whether or not this is a good idea feels a bit contingent.

vinyldarkscratch:
The reasoning for these PRs is simple: consolidation and simplification.  From what I understand, the policy came about as a conflict resolution to aid with contributions.  However, from the initial discussion regarding the policy (as well as my own opinion), such flag data isn't really worth maintaining.  The increased effort to maintain and verify the flag-based data (especially since it can't really be automated, or if it can, not efficiently), with the lack of maintenance interests from [certain browser vendors](https://github.com/mdn/browser-compat-data/issues/3318#issuecomment-456441565), and the seemingly low consumer interest, it just seems like a recipe that adds up to scrapping the data.  (And I realize the irony of me talking about scrapping old data, considering my reputation of testing in old Internet Explorer 1, 2, 3, etc. 😛)

An added benefit to this flag removal project is catching some data inaccuracies and inconsistencies.  For example, if I hadn't gone through with my flag removal script, we wouldn't have found oddities with Firefox for the [WebVR APIs](https://github.com/mdn/browser-compat-data/issues/9790), nor would we have found inconsistencies in the (CSS Grid data)[https://github.com/mdn/browser-compat-data/pull/9683] due to modifying Firefox Desktop but not Firefox Android.

ddbeck (approved):
OK, good enough for me. Thank you!

PR #9721, opened 2021-04-02
Remove irrelevant flags in Firefox for api.Document.*fullscreen

vinyldarkscratch:
This PR removes irrelevant flag data for Firefox and Firefox Android for the fullscreen-related members of the `Document` API as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-flag-data).

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -5301,17 +5301,6 @@
                 "version_added": "64"
               },
               {
-                "version_added": "49",
-                "version_removed": "65",
-                "flags": [
-                  {
-                    "type": "preference",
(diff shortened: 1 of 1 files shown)


foolip (changes requested):
Can you search for "full-screen-api.unprefix.enabled" and clean up the other cases at the same time?

foolip (approved):
There are additional files with this same flag, but one roundtrip is enough, that can be fixed separately :)

PR #9724, opened 2021-04-03
Remove irrelevant flags in Firefox for api.Document.pointer_events

vinyldarkscratch:
This PR removes irrelevant flag data for Firefox and Firefox Android for the pointer events of the `Document` API as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-flag-data).

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -8551,36 +8551,12 @@
                 "alternative_name": "mspointercancel"
               }
             ],
-            "firefox": [
-              {
-                "version_added": "59"
-              },
-              {
-                "version_added": "29",
-                "flags": [
-                  {
-                    "type": "preference",
-                    "name": "dom.w3c_pointer_events.enabled",
-                    "value_to_set": "true"
-                  }
-                ]
-              }
-            ],
-            "firefox_android": [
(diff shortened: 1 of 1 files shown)


lucalves (approved):

api/Document.json
```diff
@@ -8551,36 +8551,12 @@
...
-                    "value_to_set": "true"
-                  }
-                ]
-              }
-            ],
+            "firefox": {
+              "version_added": "59"
+            },
+            "firefox_android": {
+              "version_added": "79"
```
foolip, line 8558: When was Firefox for Android 79 released? I think this is not old enough.

vinyldarkscratch: Ah yeah, forgot to consider FxA's version -- do you think it makes more sense to keep only the desktop changes, or close this PR altogether and come back to it another time?

foolip: First I would check if the Android data is actually correct, since the difference is so big. If it is I'd leave this cleanup for later, doing it all at once.

Issue #8908, opened 2021-01-30
javascript.builtins.String.replace - IE5.0 doesn't seem to support a function as a second argument


unterwulf:
#### What information was incorrect, unhelpful, or incomplete?
IE has full support of javascript.builtins.String.replace since IE4

#### What did you expect to see?
I haven't checked IE4, but at least IE5.00.2314.1003 under Windows 95 doesn't take a function as a second parameter. Instead it always interpret it as a string, even when there are no quotes.

#### Did you test this? If so, how?
`alert('abc'.replace(/./g, function(c) { return '%' + c.charCodeAt(0).toString(16); }));`
##### I expected to see:
%61%62%63
##### I actually saw:
function(c) { return '%' + c.charCodeAt(0).toString(16); }))function(c) { return '%' + c.charCodeAt(0).toString(16); }))function(c) { return '%' + c.charCodeAt(0).toString(16); }))'

<details>
<summary>MDN page report details</summary>

* Query: `javascript.builtins.String.replace`
* MDN URL: https://developer.mozilla.org/en-us/docs/Web/JavaScript/Reference/Global_Objects/String/replace
* Report started: 2021-01-30T23:15:24.269Z

</details>

ddbeck:
Thanks for reporting this issue, @unterwulf.

I'm only able to test as far back as IE 6 (via BrowserStack) and it does work there, but I'll take your word for it that this fails in IE 5. To fix this in the compat data, we'd need to set the IE value to `"≤6"` in the [`replace` feature in `String.json`](https://github.com/mdn/browser-compat-data/blob/master/javascript/builtins/String.json#L1807-L1858).

Alternatively, if you're able to test this is in IE 5.5, we could set a more specific value. If it works in IE 5.5, we can use `"5.5"`; if it doesn't, then we can use `"6"`. If you are able to test this, I'd appreciate it if you could update this issue with the results (whether you can do this next part or not).

I'd welcome a PR from you (or another contributor) making this change to the data. Thank you!

unterwulf:
Tested the same JS code on IE 5.50.4134.0600 under Win 95 4.00.950 B and it works as expected. So I guess 5.5 is the minimal IE version that fully supports this method.

Sorry, I can't create a PR as I'm completely stranger in this project, just using these webpages as a reference doing retro websites.

ddbeck:
Fixed by https://github.com/mdn/browser-compat-data/pull/9725. Thank you, @unterwulf!

PR #9725, opened 2021-04-03
String.prototype.replace is not fully supported until IE5.5 (#8908)

unterwulf:
This patch refines IE support of String.prototype.replace. The method was found not to be fully supported until IE5.5.

### DIFF
--- a/javascript/builtins/String.json
+++ b/javascript/builtins/String.json
@@ -1824,9 +1824,17 @@
               "firefox_android": {
                 "version_added": "4"
               },
-              "ie": {
-                "version_added": "4"
-              },
+              "ie": [
+                {
+                  "version_added": "5.5"
+                },
+                {
+                  "version_added": "4",
+                  "version_removed": "5.5",
+                  "partial_implementation": true,
+                  "notes": "A replacement function as second argument is not supported."
+                }
+              ],
               "nodejs": {
                 "version_added": "0.10.0"
               },


javascript/builtins/String.json
```diff
@@ -1824,9 +1824,16 @@
...
               },
-              "ie": {
-                "version_added": "4"
-              },
+              "ie": [
+                {
+                  "version_added": "5.5"
+                },
+                {
+                  "version_added": "4",
```
ddbeck, line 1832: The note here doesn't apply to later versions, so we can close this support statement when the next one begins.

```suggestion
                  "version_added": "4",
                  "version_removed": "5.5",
```

javascript/builtins/String.json
```diff
@@ -1824,9 +1824,16 @@
...
+                  "partial_implementation": true,
+                  "notes": "No support for a function as a second parameter before version 5.5."
```
ddbeck, line 1834: This rewords the note to avoid restating the version number (and to use a style more like other notes in the repo).

```suggestion
                  "notes": "A replacement function as second argument is not supported."
```

ddbeck (changes requested):
Excellent work, @unterwulf. A couple of small suggestions, but this is a great first PR to the project. Welcome!

javascript/builtins/String.json
unterwulf: Agree. Fixed.

ddbeck (approved):
Perfect. Thank you! 🎉

javascript/builtins/String.json
unterwulf: I used other notes sections in this file for inspiration and they do restate version numbers :-) But I agree, your wording is much better. Fixed.

PR #9729, opened 2021-04-04
Remove irrelevant flags in Firefox for CSSPseudoElement API

vinyldarkscratch:
This PR removes irrelevant flag data for Firefox and Firefox Android for the `CSSPseudoElement` API as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-flag-data).


foolip (commented):
This is also just removing some of the flag data. I think it would be better to remove it in one go when possible.

PR #9733, opened 2021-04-04
Remove irrelevant flags in Firefox for api.Document.timeline

vinyldarkscratch:
This PR removes irrelevant flag data for Firefox and Firefox Android for the `timeline` member of the `Document` API as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-flag-data).

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -10275,36 +10275,12 @@
             "edge": {
               "version_added": "12"
             },
-            "firefox": [
-              {
-                "version_added": "48"
-              },
-              {
-                "version_added": "47",
-                "version_removed": "48",
-                "flags": [
-                  {
-                    "type": "preference",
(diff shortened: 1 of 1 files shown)


api/Document.json
```diff
@@ -10605,17 +10533,6 @@
                     "value_to_set": "true"
                   }
                 ]
-              },
-              {
-                "version_added": "32",
-                "version_removed": "63",
-                "flags": [
```
foolip, line 10612: I'm not so sure about removing only part of the flag data for an entry, can you leave this in until it can all be removed?

foolip (changes requested):

foolip (approved):

PR #9737, opened 2021-04-04
Remove irrelevant flags in Chromium for FeaturePolicy API

vinyldarkscratch:
This PR removes irrelevant flag data for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `FeaturePolicy` API as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-flag-data).

### DIFF
--- a/api/FeaturePolicy.json
+++ b/api/FeaturePolicy.json
@@ -18,18 +18,6 @@
                   "value_to_set": "Enabled"
                 }
               ]
-            },
-            {
-              "version_added": "69",
-              "version_removed": "73",
-              "alternative_name": "Policy",
-              "flags": [
-                {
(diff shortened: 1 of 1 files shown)


foolip (commented):
This is also just removing some of the flag data. I think it would be better to remove it in one go when possible.

But I'm also pretty sure this data is wrong, that it's the wrong flag. Probably a copy-paste mistake.

vinyldarkscratch:
Did some checking, and it appears that the flag is actually correct.  Though since there's newer flag data (which is just barely out of the two year mark), I'll close this PR and follow up with a more "complete" PR later on.

PR #9742, opened 2021-04-04
Remove irrelevant flags in Firefox for HMDVRDevice API

vinyldarkscratch:
This PR removes irrelevant flag data for Firefox and Firefox Android for the `HMDVRDevice` API as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-flag-data).

### DIFF
--- a/api/HMDVRDevice.json
+++ b/api/HMDVRDevice.json
@@ -24,23 +24,10 @@
             ],
             "notes": "The support for this feature is currently disabled by default in Firefox. To enable WebVR support in Firefox Nightly/Developer Edition, you can go to <code>about:config</code> and enable the <code>dom.vr*</code> prefs. A better option however is to install the <a href='http://www.mozvr.com/downloads/webvr-addon-0.1.0.xpi'>WebVR Enabler Add-on</a>, which does this for you and sets up other necessary parts of the <a href='https://developer.mozilla.org/docs/Web/API/WebVR_API/WebVR_environment_setup'>environment</a>"
           },
-          "firefox_android": [
-            {
-              "version_added": "44",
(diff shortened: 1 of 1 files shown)


api/HMDVRDevice.json
```diff
@@ -24,23 +24,10 @@
...
-                  "type": "preference",
-                  "name": "dom.vr*"
-                }
-              ],
-              "notes": "The support for this feature is currently disabled by default in Firefox. To enable WebVR support in Firefox Nightly/Developer Edition, you can go to <code>about:config</code> and enable the <code>dom.vr*</code> prefs. A better option however is to install the <a href='http://www.mozvr.com/downloads/webvr-addon-0.1.0.xpi'>WebVR Enabler Add-on</a>, which does this for you and sets up other necessary parts of the <a href='https://developer.mozilla.org/docs/Web/API/WebVR_API/WebVR_environment_setup'>environment</a>."
-            }
-          ],
+          "firefox_android": {
+            "version_added": "44",
+            "notes": "The <code>dom.vr*</code> prefs are enabled by default at this point, in Nightly/Aurora editions."
```
foolip, line 29: A note like this can be removed.

foolip (commented):
Is the data really correct, that this only ever shipped in Firefox for Android?

api/HMDVRDevice.json
vinyldarkscratch: Ack, what this note is saying is that the flag wasn't actually removed.  In fact, there shouldn't really be a separate support statement in the first place -- if the flags are only enabled by default in nightly and beta builds, but disabled in stable, consumers still have to toggle the flag.

I'll open up an issue about this.

PR #9743, opened 2021-04-04
Remove irrelevant flags in Firefox for HTMLIFrameElement API

vinyldarkscratch:
This PR removes irrelevant flag data for Firefox and Firefox Android for the `HTMLIFrameElement` API as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-flag-data).

### DIFF
--- a/api/HTMLIFrameElement.json
+++ b/api/HTMLIFrameElement.json
@@ -429,18 +429,6 @@
                     "value_to_set": "Enabled"
                   }
                 ]
-              },
-              {
-                "version_added": "69",
-                "version_removed": "73",
-                "alternative_name": "policy",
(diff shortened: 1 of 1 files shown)


foolip:
Can you fold this into https://github.com/mdn/browser-compat-data/pull/9737? Same issue applies.

vinyldarkscratch:
Closing for the same reason as #9737:

> Did some checking, and it appears that the flag is actually correct. Though since there's newer flag data (which is just barely out of the two year mark), I'll close this PR and follow up with a more "complete" PR later on.

PR #9751, opened 2021-04-04
Remove irrelevant flags in all browsers for Headers API

vinyldarkscratch:
This PR removes irrelevant flag data for all browsers for the `Headers` API as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-flag-data).

Paired with #9916 and #9917.


api/Headers.json
```diff
@@ -4,85 +4,30 @@
...
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "dom.fetch.enabled"
-                }
-              ]
-            }
-          ],
+          "firefox": {
+            "version_added": "39"
```
foolip, line 17: I doubt this data is correct, that `fetch()` would have shipped earlier in desktop than on Android. This change makes the difference more obvious. Are you able to test for when this was shipped in Firefox for Android?

Looks like the data came from wiki migration: https://github.com/mdn/browser-compat-data/pull/929

api/Headers.json
```diff
@@ -107,85 +52,30 @@
...
-                    "name": "dom.fetch.enabled"
-                  }
-                ]
-              }
-            ],
+            "firefox": {
+              "version_added": "39"
+            },
             "firefox_android": {
               "version_added": false
```
foolip, line 68: This seems implausible.

api/Headers.json
```diff
@@ -854,7 +515,7 @@
...
             },
             "opera": {
-              "version_added": "28"
+              "version_added": "29"
```
foolip, line 518: Right, Opera 29 is based on Chromium 42, this seems to have been wrong.

foolip (approved):
There seems to be a lot wrong with the Firefox for Android data here. I don't think it should be fixed in this PR, but it could be clearer if it's fixed before this PR. I'm approving since either way would be fine, but not merging since I'm not sure.

api/Headers.json
vinyldarkscratch: I'll have to see if I can run a Firefox Android instance that old, but for now, I've opened up #9880.

ddbeck:
> I'm approving since either way would be fine, but not merging since I'm not sure.

@vinyldarkscratch @foolip what's the plan here? Are changes planned or should this get merged?

foolip:
Over to @vinyldarkscratch to decide on the order.

vinyldarkscratch:
I'm not quite ready to test old Firefox Android versions (there's more steps I need to perform to get my test environment ready), so I'd like to get this PR merged first if that's alright!

PR #9775, opened 2021-04-06
Add protocol_handlers subfeature ftp

hamishwillee:
As per https://bugzilla.mozilla.org/show_bug.cgi?id=1626365#c10 FF88
- adds ftp to the list of protocol_handlers. Associated MDN page https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/manifest.json/protocol_handlers.
- removes any action if you set FTP in the proxy settings (i.e. you can still set it but it has not effect)

This is part of FF88 removing support for FTP (doc tracking in https://github.com/mdn/content/issues/3462)

### DIFF
--- a/webextensions/api/proxy.json
+++ b/webextensions/api/proxy.json
@@ -499,7 +499,7 @@
               "firefox_android": [
                 {
                   "version_added": "56",
-                  "version_removed": "71"
+                  "version_removed": "79"
                 },
                 {
                   "alternative_name": "registerProxyScript",
@@ -565,7 +565,7 @@
               },
               "firefox_android": {
                 "version_added": "56",
-                "version_removed": "71"
+                "version_removed": "79"
               },
               "opera": {
                 "version_added": false
--- a/webextensions/api/proxy.json
+++ b/webextensions/api/proxy.json
@@ -489,11 +489,12 @@
               "firefox": [
                 {
                   "version_added": "56",
-                  "version_removed": "71"
+                  "version_removed": "79"
                 },
                 {
                   "alternative_name": "registerProxyScript",
-                  "version_added": "55"
+                  "version_added": "55",
+                  "version_removed": "71"
                 }
               ],
               "firefox_android": [
@@ -503,7 +504,8 @@
                 },
                 {
                   "alternative_name": "registerProxyScript",
-                  "version_added": "55"
+                  "version_added": "55",
+                  "version_removed": "79"
                 }
               ],
               "opera": {
@@ -531,11 +533,8 @@
                 "version_added": false
               },
               "firefox": {
-                "notes": [
-                  "In version 59, this property was listed as <code>proxyConfig</code> in the <code>browserSettings</code> namespace, but it had a bug that made it mostly unusable.",
-                  "From version 88 the <code>ftp</code> setting has no effect, because FTP is not longer supported (see <a href='https://bugzil.la/1626365'>bug 1626365</a>)."
-                ],
-                "version_added": "60"
+                "version_added": "60",
+                "notes": "From version 88, the <code>ftp</code> setting has no effect because FTP is no longer supported (see <a href='https://bugzil.la/1626365'>bug 1626365</a>)."
               },
               "firefox_android": {
                 "version_added": false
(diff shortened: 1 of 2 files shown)


webextensions/api/proxy.json
```diff
@@ -498,7 +498,8 @@
               ],
               "firefox_android": [
                 {
-                  "version_added": "56"
+                  "version_added": "56",
+                  "version_removed": "71"
```
hamishwillee, line 502: Version removed markup was put in for FF desktop in version 71 back in #4496. The associated issues doesn't explicitly talk about the removal in android, but it seems very unreasonable to assume that it was not removed. So I think it was an omission. Added here to register, and below in unregister.

webextensions/api/proxy.json
```diff
@@ -560,7 +564,8 @@
                 "version_removed": "71"
               },
               "firefox_android": {
-                "version_added": "56"
+                "version_added": "56",
+                "version_removed": "79"
```
hamishwillee, line 568: Actually, note is 79 (first version after 71 for FF Android)

webextensions/api/proxy.json
hamishwillee: Changed to 79 (first valid release)

webextensions/api/proxy.json
```diff
@@ -530,7 +531,10 @@
                 "version_added": false
               },
               "firefox": {
-                "notes": "In version 59, this property was listed as <code>proxyConfig</code> in the <code>browserSettings</code> namespace, but it had a bug that made it mostly unusable.",
+                "notes": [
+                  "In version 59, this property was listed as <code>proxyConfig</code> in the <code>browserSettings</code> namespace, but it had a bug that made it mostly unusable.",
+                  "From version 88 the <code>ftp</code> setting has no effect, because FTP is not longer supported (see <a href='https://bugzil.la/1626365'>bug 1626365</a>)."
```
ddbeck, line 536: The old note is a bit of a nuisance, really (it covers a version not listed in the range). How about we get rid of it and sub in your note? We can even put it after the "version_added" field, which is conventional, though not strictly required.

(There's also a couple of tiny copy edits in the note itself)

```
"version_added": "60",
"notes": "From version 88, the <code>ftp</code> setting has no effect because FTP is no longer supported (see <a href='https://bugzil.la/1626365'>bug 1626365</a>)."

File: webextensions/api/proxy.json [hunk 4]
```diff
@@ -498,7 +498,8 @@
               ],
               "firefox_android": [
                 {
-                  "version_added": "56"
+                  "version_added": "56",
+                  "version_removed": "79"
                 },
                 {
                   "alternative_name": "registerProxyScript",
```
ddbeck, line 505: It struck me as… unlikely that the alternative_name would persist while the rest of the feature went away. Sure enough, `registerProxyScript` was removed in Firefox 71: https://bugzilla.mozilla.org/show_bug.cgi?id=1443259#c11

Can we close out this support statement for both Firefox desktop (71) and Android (79)?

ddbeck (changes requested):
A couple little suggestions for you. Thanks, @hamishwillee!

webextensions/api/proxy.json
hamishwillee: Makes sense!

webextensions/api/proxy.json
hamishwillee: Thanks. For android this is version 79 that is first version after v71. I also updated the firefox desktop  case above. Needs to be checked.

hamishwillee:
Should be good now I hope.

webextensions/api/proxy.json
```diff
@@ -489,20 +489,23 @@
               "firefox": [
                 {
                   "version_added": "56",
-                  "version_removed": "71"
+                  "version_removed": "79"
```
ddbeck, line 492: I think this got over corrected. It was right to switch Android to 79, but desktop should stay at 71

```suggestion
                  "version_removed": "71"
```

ddbeck (changes requested):
One last thing! Thank you, @hamishwillee!

webextensions/api/proxy.json
hamishwillee: Thanks @ddbeck. This is what we call in the trade, a screwup. Thanks for catching it.

ddbeck (approved):
Looks good. Thank you! 🎉

Issue #9711, opened 2021-04-01
css.properties.font-variant-caps - seems to be supported in Safari


vasilisvg:
In the Browser compatibility tables it says that `font-variant-caps` is not supported by Safari. Yet both in *Safari 14.0.3 on MacOS 11.2.3* and on *Safari on iOS 14.4.2* it seems to work fine.

<details>
<summary>MDN page report details</summary>

* Query: `css.properties.font-variant-caps`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/CSS/font-variant-caps
* Report started: 2021-04-01T15:27:22.129Z

</details>

ddbeck:
Thanks for reporting this, @vasilisvg!

I looked into this and you're correct. It's been supported since Safari 9.1 on the Mac (and Safari 9.3 on iOS), according to [the release notes](https://developer.apple.com/library/archive/releasenotes/General/WhatsNewInSafari/Articles/Safari_9_1.html#//apple_ref/doc/uid/TP40014305-CH10-SW14). A quick test in BrowserStack confirmed this.

I'd welcome a PR (from you or another contributor) that updates the Safari data in `font-variant-caps.json`:

https://github.com/mdn/browser-compat-data/blob/c820ce274991dd4bff1560a507ef75dcab213f02/css/properties/font-variant-caps.json#L59-L64

lucalves:
Hello @vasilisvg , I opened #9789 to solve this.

ddbeck:
Fixed by #9789. 🎉

PR #9789, opened 2021-04-07
Includes support for font-variant-caps in Safari

lucalves:
According to the [release notes](https://developer.apple.com/library/archive/releasenotes/General/WhatsNewInSafari/Articles/Safari_9_1.html#//apple_ref/doc/uid/TP40014305-CH10-SW14), it is supported since version 9.1 on Safari and version 9.3 on Safari iOS.

Fixes **#9711**.

### DIFF
--- a/css/properties/font-variant-caps.json
+++ b/css/properties/font-variant-caps.json
@@ -57,10 +57,10 @@
               "version_added": "41"
             },
             "safari": {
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Looks good. Thank you! 🎉

PR #9792, opened 2021-04-07
Mark most -ms-prefixed CSS properties as removed in Edge 79

foolip:
These were confirmed supported in Edge 18 but not supported in Edge 80
on BrowserStack:
https://mdn-bcd-collector.appspot.com/tests/css/properties/-ms-grid-columns
https://mdn-bcd-collector.appspot.com/tests/css/properties/-ms-grid-rows
https://mdn-bcd-collector.appspot.com/tests/css/properties/-ms-text-combine-horizontal

Edge 79 (not available on BrowserStack) was assumed to match.

Note that -ms-high-contrast-adjust remains in Chromium-based Edge and is
not updated here.

### DIFF
--- a/css/properties/grid-auto-columns.json
+++ b/css/properties/grid-auto-columns.json
@@ -40,6 +40,7 @@
               },
               {
                 "version_added": "12",
+                "version_removed": "79",
                 "alternative_name": "-ms-grid-columns"
               }
             ],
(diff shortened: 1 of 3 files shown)


ddbeck (approved):
> Edge 79 (not available on BrowserStack) was assumed to match.

It is available on Sauce Labs and I confirmed it, too. 👍

Thank you, @foolip.

PR #9805, opened 2021-04-09
Fix presentation API removal back to FF88

hamishwillee:
Fixes up error introduced in #9670 (testing shows feature was removed FF88 while issue shows FF61 - based on a bug report).

FYI @ddbeck


ddbeck (approved):
Thank you! 🎉

PR #9807, opened 2021-04-09
Remove Edge "Enable service workers" flag

vinyldarkscratch:
This PR removes the data for Edge's "Enable service workers" flag, as it is now considered irrelevant.  Inspired by #9755.

### DIFF
--- a/api/PushManager.json
+++ b/api/PushManager.json
@@ -10,20 +10,9 @@
           "chrome_android": {
             "version_added": "42"
           },
-          "edge": [
-            {
-              "version_added": "16",
-              "flags": [
-                {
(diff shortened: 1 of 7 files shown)


api/PushManager.json
```diff
@@ -10,20 +10,9 @@
           "chrome_android": {
             "version_added": "42"
           },
-          "edge": [
-            {
-              "version_added": "16",
-              "flags": [
```
foolip, line 16: The flagged entry was first here, so this will improve what people see on MDN by default, yay 😃

foolip (approved):
Thank you!

PR #9810, opened 2021-04-09
Remove irrelevant "dom.animations-api.core.enabled" flag

vinyldarkscratch:
This PR removes the now-irrelevant `dom.animations-api.core.enabled` flag from API data.


foolip:
I don't think that removing parts of the flag data is really improving the data or its maintainability. I'd prefer to leave it in until the entry can be simplified to contain no flag data at all.

vinyldarkscratch:
I was thinking that since this particular flag had been removed from `AnimationTimeline` already (#9713), we should remove it from the other files containing data for it as well.  Happy to wait until we can remove more flags, though.

foolip:
OK, I'm a sucker for consistency :)

api/CSSPseudoElement.json
```diff
@@ -33,39 +33,17 @@
...
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "dom.animations-api.core.enabled"
-                }
-              ]
             }
           ],
+          "firefox_android": {
+            "version_added": "63",
```
foolip, line 39: Is the Android data just out of sync with desktop here? Why no "dom.css_pseudo_element.enabled" entry?

api/CSSPseudoElement.json
```diff
@@ -246,39 +202,17 @@
...
-                "version_removed": "63",
-                "flags": [
-                  {
-                    "type": "preference",
-                    "name": "dom.animations-api.core.enabled"
-                  }
-                ]
               }
             ],
+            "firefox_android": {
```
foolip, line 207: "dom.css_pseudo_element.enabled" entry missing here too?

foolip (commented):
In addition to missing "dom.css_pseudo_element.enabled" entried for Firefox for Android, there's also the issue of Firefox for Android dropping support for flags entirely recently.

api/CSSPseudoElement.json
vinyldarkscratch: The changes to add the new flag are in #5784.  My guess is it's just a desync between Firefox and Fenix, but of course, since Fenix doesn't support flags at all anymore, we should probably add a `"version_removed": "79"` statement?

foolip: Yes, but just putting an end to the flags statement wouldn't be correct if the data is out of sync. Rather there should be another (newer) entry that ends at 79.

vinyldarkscratch: Sorry, I realize that I poorly worded my reply (I'm horrible at explanations, haha)... 
 When I said "desync", what I was really trying to say was that data was added for Firefox Desktop, but not for Firefox Android, as Fenix was still in development and there were no FxA releases.  Since FxA doesn't have any flags any longer, `dom.css_pseudo_element.enabled` was never introduced to begin with.

foolip: If Firefox for Android 79 is when flags were discontinued, then there would be a 75-79 entry for "dom.css_pseudo_element.enabled", right?

vinyldarkscratch: There was no Firefox Android releases between 68 and 79 as a part of the development of the new Fenix edition, so Firefox Android 75-78 never existed.

foolip: Aha, right!

foolip: So then I guess a `"version_removed": "79"` line is all that's missing, as you suggested.

api/Element.json
```diff
@@ -710,39 +710,17 @@
...
-                  "flags": [
-                    {
-                      "type": "preference",
-                      "name": "dom.animations-api.core.enabled"
-                    }
-                  ]
                 }
               ],
+              "firefox_android": {
+                "version_added": "63",
```
foolip, line 716: Missing `version_removed` here.

api/Element.json
```diff
@@ -907,39 +885,17 @@
...
```
foolip, line 891: Missing `version_removed` here.

api/Document.json
```diff
@@ -10626,17 +10626,6 @@
                     "value_to_set": "true"
```
foolip, line 10626: Missing `version_removed` in this entry, but I can't comment on the appropriate line.

api/DocumentTimeline.json
```diff
@@ -27,17 +27,6 @@
...
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "dom.animations-api.core.enabled",
-                  "value_to_set": "true"
-                }
-              ]
             }
           ],
           "firefox_android": [
```
foolip, line 32: Missing `version_removed` in this entry, but I can't comment on the appropriate line.

api/DocumentTimeline.json
```diff
@@ -122,17 +100,6 @@
...
-                "flags": [
-                  {
-                    "type": "preference",
-                    "name": "dom.animations-api.core.enabled",
-                    "value_to_set": "true"
-                  }
-                ]
               }
             ],
             "firefox_android": [
```
foolip, line 105: Missing `version_removed` in this entry, but I can't comment on the appropriate line.

api/Element.json
```diff
@@ -3476,54 +3432,6 @@
...
-                "notes": "Does not automatically flush pending style changes and does not support the <code>subtree</code> option.",
-                "flags": [
-                  {
-                    "type": "preference",
-                    "name": "dom.animations-api.core.enabled"
-                  }
-                ]
               }
             ],
             "firefox_android": [
```
foolip, line 3437: Missing `version_removed` in this entry, but I can't comment on the appropriate line.

vinyldarkscratch:
Added `version_removed` in all the other locations!  I also added a few to Firefox Desktop in the process (if we're here, we might as well, right?).

foolip (approved):

PR #9812, opened 2021-04-09
updates backdrop-filter for Opera

rachelandrew:
Raised in https://github.com/mdn/content/issues/3059

Opera has support for `backdrop-filter`. Assumed Opera 63 from Chromium version and also tested that which showed support.

Fixes #9812


foolip (approved):
Confirmed by testing http://mdn-bcd-collector.appspot.com/tests/css/properties/backdrop-filter in Opera 62 and 63.

PR #9821, opened 2021-04-11
Remove extend dependency, becoming a zero-dependency package

jonathantneal:
This PR replaces the lone `extend` dependency with an explicit function, possibly improving the readability <sup>[a]</sup> of `index.js`.

<sup>a:</sup> this package only extends plain objects, and it extends them in a unique way, where any two nested objects are merged into one new object, rather than continually merging into the former. This unique strategy seems useful to have in the open.

### DIFF
--- a/index.js
+++ b/index.js
@@ -1,7 +1,6 @@
 'use strict';
 const fs = require('fs');
 const path = require('path');
-const extend = require('extend');
 
 function load() {
   // Recursively load one or more directories passed as arguments.
@@ -23,7 +22,7 @@ function load() {
 
     // The JSON data is independent of the actual file
     // hierarchy, so it is essential to extend "deeply".
-    result = extend(true, result, extra);
+    result = extend(result, extra);
   }
 
   for (dir of arguments) {
@@ -32,6 +31,29 @@ function load() {
   }
 
   return result;
+}
+
+function extend(a, b) {
+  // iterate over all direct and inherited enumerable properties
+  for (const name in b) {
+    let newValue = b[name];
+
+    // check if the new value is an object and its property exists on the former
+    if (typeof newValue === 'object' && newValue !== null && name in a) {
+      const oldValue = a[name];
+
+      // check if the former value is also an object
+      if (typeof oldValue === 'object' && oldValue !== null) {
+        // update the new value as a new object extended by the former and later
+        newValue = extend(extend({}, oldValue), newValue);
+      }
+    }
+
+    // update the object with the new value
+    a[name] = newValue;
+  }
+
+  return a;
 }
 
 module.exports = load(
--- a/index.js
+++ b/index.js
@@ -22,7 +22,7 @@ function load() {
 
     // The JSON data is independent of the actual file
     // hierarchy, so it is essential to extend "deeply".
-    result = extend(result, extra);
+    extend(result, extra);
   }
 
   for (dir of arguments) {
@@ -42,11 +42,11 @@ function extend(target, source) {
     throw new Error('Both target and source must be plain objects');
   }
 
-  // iterate over all enumerable properties
+  // iterate over own enumerable properties
   for (const [key, value] of Object.entries(source)) {
     // recursively extend if target has the same key, otherwise just assign
     if (Object.prototype.hasOwnProperty.call(target, key)) {
-      target[key] = extend(extend({}, target[key]), value);
+      extend(target[key], value);
     } else {
       target[key] = value;
     }
--- a/index.js
+++ b/index.js
@@ -51,8 +51,6 @@ function extend(target, source) {
       target[key] = value;
     }
   }
-
-  return target;
 }
 
 module.exports = load(
--- a/index.js
+++ b/index.js
@@ -33,27 +33,26 @@ function load() {
   return result;
 }
 
-function extend(a, b) {
-  // iterate over all direct and inherited enumerable properties
-  for (const name in b) {
-    let newValue = b[name];
+function isPlainObject(v) {
+  return typeof v === 'object' && v !== null && !Array.isArray(v);
+}
 
-    // check if the new value is an object and its property exists on the former
-    if (typeof newValue === 'object' && newValue !== null && name in a) {
-      const oldValue = a[name];
-
-      // check if the former value is also an object
-      if (typeof oldValue === 'object' && oldValue !== null) {
-        // update the new value as a new object extended by the former and later
-        newValue = extend(extend({}, oldValue), newValue);
-      }
-    }
-
-    // update the object with the new value
-    a[name] = newValue;
+function extend(target, source) {
+  if (!isPlainObject(target) || !isPlainObject(source)) {
+    throw new Error('Both target and source must be plain objects');
   }
 
-  return a;
+  // iterate over all enumerable properties
+  for (const [key, value] of Object.entries(source)) {
+    // recursively extend if target has the same key, otherwise just assign
+    if (Object.prototype.hasOwnProperty.call(target, key)) {
+      target[key] = extend(extend({}, target[key]), value);
+    } else {
+      target[key] = value;
+    }
+  }
+
+  return target;
 }
 
 module.exports = load(
--- a/package.json
+++ b/package.json
@@ -4,9 +4,6 @@
   "description": "Browser compatibility data provided by MDN Web Docs",
   "main": "index.js",
   "types": "index.d.ts",
-  "dependencies": {
-    "extend": "3.0.2"
-  },
   "engines": {
     "node": ">=10.0.0"
   },
--- a/package-lock.json
+++ b/package-lock.json
@@ -466,11 +466,6 @@
       "integrity": "sha1-G2HAViGQqN/2rjuyzwIAyhMLhtQ=",
       "dev": true
     },
-    "extend": {
-      "version": "3.0.2",
-      "resolved": "https://registry.npmjs.org/extend/-/extend-3.0.2.tgz",
-      "integrity": "sha512-fjquC59cD7CyW6urNXK0FBufkZcoiGG80wTuPujX590cB5Ttln20E2UB4S/WARVqhXffZl2LNgS+gQdPIIim/g=="
-    },
     "fast-deep-equal": {
       "version": "3.1.3",
       "resolved": "https://registry.npmjs.org/fast-deep-equal/-/fast-deep-equal-3.1.3.tgz",


foolip:
I can confirm that this produces exactly the same object (compared after `JSON.stringify`) as before the change.

It would be great if the new `extend` could be tested, but to do that would require breaking it out into a helper that's not excluded by `.npmignore`, so I'll leave it to @ddbeck to say if that's worthwhile.

package.json
```diff
@@ -4,9 +4,6 @@
   "description": "Browser compatibility data provided by MDN Web Docs",
   "main": "index.js",
   "types": "index.d.ts",
-  "dependencies": {
```
foolip, line 7: `package-lock.json` also needs to be updated.

index.js
```diff
@@ -34,6 +33,29 @@ function load() {
   return result;
 }
 
+function extend(a, b) {
+  // iterate over all direct and inherited enumerable properties
+  for (const name in b) {
+    let newValue = b[name];
+
+    // check if the new value is an object and its property exists on the former
```
foolip, line 41: This merging logic is more lenient than BCD needs, and would allow for weird errors in the BCD data which would go unnoticed unless there's a lint for it. This might be a good opportunity to simplify this to make more assumptions about the data, throwing exceptions if they're not met.

I think that any time a property exists on both `a` and `b`, both values must be regular objects (not arrays, for example), otherwise something is wrong.

index.js
```diff
@@ -34,6 +33,29 @@ function load() {
...
+    if (typeof newValue === 'object' && newValue !== null && name in a) {
+      const oldValue = a[name];
+
+      // check if the former value is also an object
+      if (typeof oldValue === 'object' && oldValue !== null) {
+        // update the new value as a new object extended by the former and later
+        newValue = extend(extend({}, oldValue), newValue);
```
foolip, line 48: Are two `extend` calls needed here? Since `extend` modifies `a` in-place, I think this could be just `extend(oldValue, newValue)`, assuming the `a[name] = newValue` is done in an else branch and not unconditionallly.

foolip (commented):
Some ideas, but ultimately these are suggestions, what's in this PR seems to already work just fine.

index.js
foolip: Doing this will reveal that there is some broken structure in BCD, the first one I run into is "bcd.css.selectors.not_match_link" defined in multiple files by accident.

foolip: I've fixed that in https://github.com/mdn/browser-compat-data/pull/9842.

index.js
jonathantneal: Yes, I agree that this is **peculiar**, which is a reason I think it belongs _in_ the codebase and not abstracted away.

From my PR, with emphasis added:

> ... it extends them in a unique way, where any two nested objects are merged into one **new** object

The (jQuery) extend method mutates a root object, but it does not mutate any of its nested objects.

If this (new) extend method were simplified so that the nested objects were also mutated, then many tests would fail.

From those failures I inferred that this **particular** functionality is by design.

index.js
jonathantneal: > This might be a good opportunity to simplify this to make more assumptions about the data, throwing exceptions if they're not met.

Yes. I’d like to help here, if I can. 😄

This PR aims to provide a simplified method that would not throw new errors, and then also honor the ** peculiar** bits from the (jQuery) extend dependency implementation; versus something like a deep `Object.assign`.

I see two options, and defer to your preference. One option is to "_do the most_", and improve the usefulness of this function and then bundle any resulting fixes that are surfaced. The other option is to "_do the least_", and approve or improve this function to replace the existing dependency, with its **peculiar** bits and leniencies out in the open.

foolip: With that fixed, here's a less forgiving variant:

```js
function isPlainObject(v) {
  return typeof v === 'object' && v !== null && !Array.isArray(v);
}

function extend(target, source) {
  if (!isPlainObject(target) || !isPlainObject(source)) {
    throw new Error('Both target and source must be plain objects');
  }
  // iterate over all enumerable properties
  for (const [key, value] of Object.entries(source)) {
    // recursively extend if target has the same key, otherwise just assign
    if (Object.prototype.hasOwnProperty.call(target, key)) {
      extend(target[key], value);
    } else {
      target[key] = value;
    }
  }

  return target;
}
```

Feel free to use as much or as little of that as you like.

foolip: Sorry, I commented before I read your comment. A deep `Object.assign` is indeed what I think we should use here, something that throws if used with arrays in particular. `Object.assign({}, null)` doesn't throw, but that's a case we should also disallow to catch more errors I think.

index.js
foolip: I defer to @ddbeck, but I'd say that as long as the resulting object is identical to before, various peculiarities of `extend` that we don't depend on need not be preserved, indeed it's better if they don't work :)

index.js
```diff
@@ -34,6 +33,29 @@ function load() {
   return result;
 }
 
+function extend(a, b) {
+  // iterate over all direct and inherited enumerable properties
```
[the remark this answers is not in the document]
jonathantneal: Thank you for all of these thoughtful comments and questions. I also agree so much!

This leniency was done purposefully, because that’s how it behaved in the old (jQuery) extend method. From the [`extend` package documentation](https://www.npmjs.com/package/extend):

> ... properties inherited from the object's prototype will be copied over.

This is why I added the explicit comment. At least now it’s out in the open? 😅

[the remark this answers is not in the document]
foolip: Yes, very good to shine a light on it. It made me nervous since in BCD we have properties like "toString" and "hasOwnProperty" where `p in {}` would return true. It seems like we don't have any accidents because of it, but dealing only with enumerable own properties seems preferable, or really matching whatever the behavior of `Object.assign` is. (I'm still not sure about the case where there's an enumerable own property on the one side and a non-enumerable own property on the other. We can't have that in BCD, though.)

package.json
jonathantneal: Yes, I can do this! Doing so may also upgrade the [lockfileVersion](https://docs.npmjs.com/cli/v7/configuring-npm/package-lock-json#lockfileversion). Is that all right? I may not be able to avoid this with `npm`.

Sorry for all these "thing a causes thing b" complications. I was attempting to touch as little as possible. 😄

ddbeck:
@jonathantneal Thank you for opening this PR. This is pretty compelling (and reminds me that I've neglected [some other dependent-friendly PRs](https://github.com/mdn/browser-compat-data/issues/7374) 😬 ). I'm definitely drawn to things that are more explicit (and less peculiar). I'll respond to some of the line-comments shortly.

But first, to respond to @foolip's top-level question:

> It would be great if the new `extend` could be tested, but to do that would require breaking it out into a helper that's not excluded by `.npmignore`, so I'll leave it to @ddbeck to say if that's worthwhile.

On this point, could we go a half-step and check-in the tests now, but not permanently wire them up? There are a couple of PRs to compile all the JSON down to a single file to shrink the package size (see https://github.com/mdn/browser-compat-data/issues/7374). If we did that after this PR, we could simplify `index.js` further and sidestep this whole `.npmignore` thing.

index.js
ddbeck: > I defer to @ddbeck, but I'd say that as long as the resulting object is identical to before, various peculiarities of `extend` that we don't depend on need not be preserved, indeed it's better if they don't work :)

I agree with this. I'm fine with making this less weird, if the resulting object is the same or it breaks in ways that reveal genuine data problems (as in #9842).

I think `extend` was originally adopted when BCD was quite immature and it was convenient for experimentation, particularly when the schema went through some major changes early on (before my time—@Elchi3 can speak to this history better than I can). I don't think we've ever actually benefited from any of `extend'`s oddities.

So I guess what I'm saying here is: feel free to make this less permissive, but let's make sure to do a full accounting of any differences in the resulting object before we merge.

package.json
foolip: I have a version of `npm` that doesn't cause any other changes to `package-lock.json`, the resulting diff is actually trivial if you want to apply it manually:
```diff
--- a/package-lock.json
+++ b/package-lock.json
@@ -466,11 +466,6 @@
       "integrity": "sha1-G2HAViGQqN/2rjuyzwIAyhMLhtQ=",
       "dev": true
     },
-    "extend": {
-      "version": "3.0.2",
-      "resolved": "https://registry.npmjs.org/extend/-/extend-3.0.2.tgz",
-      "integrity": "sha512-fjquC59cD7CyW6urNXK0FBufkZcoiGG80wTuPujX590cB5Ttln20E2UB4S/WARVqhXffZl2LNgS+gQdPIIim/g=="
-    },
     "fast-deep-equal": {
       "version": "3.1.3",
       "resolved": "https://registry.npmjs.org/fast-deep-equal/-/fast-deep-equal-3.1.3.tgz",
```

foolip:
I love https://github.com/mdn/browser-compat-data/issues/7374 + https://github.com/mdn/browser-compat-data/pull/7398, I'd had the same thought but didn't know work had been done already. If we go that path, the code merging JSON files would be used at build time only and would be easily tested.

bershanskiy:
I love this! Replacing over 23kB of unpacked `node_modules` with just 20 lines of code and less weirdness is great.

index.js
foolip: As it is, the resulting objects are identical before and after this change, and also if https://github.com/mdn/browser-compat-data/pull/9842 is merged, my shorter and stricter `extend` produces the exact same result.

This should be confirmed once more before merging, of course.

jonathantneal:
I’ll be around in an hour to implement the requests.

package.json
jonathantneal: 🙏 Thank you.

Done in https://github.com/mdn/browser-compat-data/pull/9821/commits/a00eb8a8f64d57e6eea8b4c96380056b0471f55e

index.js
jonathantneal: Hi, @foolip! After rebasing with the latest `main` branch and updating the `package-lock.json` file, everything continues to run as expected.

However, if I update the `extend` function with the _less forgiving variant_ then the `lint` command fails.

The error is rather catastrophic on a terminal as the linter attempts to whole compat objects as being invalid, but it starts like this:

```
✖ api/ANGLE_instanced_arrays.json
  JSON Schema – 1 error:
```
  
I suspect the deep mutation of `target` causes this error, and _something else_ expects deep objects not to change.

I also tried updating the function to _always_ return a new plain object, and the lint command failed again. This time, I suspect the missing shallow mutation of `target` causes this error! Haha. 

This place in the middle works, and is done in https://github.com/mdn/browser-compat-data/pull/9821/commits/bde56ca3b3b8040b17f4f9c8a1d97a96c6ff008b 🎉

```js
function isPlainObject(v) {
  return typeof v === 'object' && v !== null && !Array.isArray(v);
}

function extend(target, source) {
  if (!isPlainObject(target) || !isPlainObject(source)) {
    throw new Error('Both target and source must be plain objects');
  }

  // iterate over all enumerable properties
  for (const [key, value] of Object.entries(source)) {
    // recursively extend if target has the same key, otherwise just assign
    if (Object.prototype.hasOwnProperty.call(target, key)) {
      target[key] = extend(extend({}, target[key]), value);
    } else {
      target[key] = value;
    }
  }

  return target;
}
```

Of course, I’m glad to continue working with you on this, if you’d like to remove this last **peculiar** bit. 😄

index.js
```diff
@@ -23,7 +22,7 @@ function load() {
 
     // The JSON data is independent of the actual file
     // hierarchy, so it is essential to extend "deeply".
-    result = extend(true, result, extra);
+    result = extend(result, extra);
```
foolip, line 25:
```suggestion
    extend(result, extra);
```

Might as well ignore the return value, since `extend` is now more like `Object.assign`.

index.js
```diff
@@ -34,6 +33,28 @@ function load() {
...
+function isPlainObject(v) {
+  return typeof v === 'object' && v !== null && !Array.isArray(v);
+}
+
+function extend(target, source) {
+  if (!isPlainObject(target) || !isPlainObject(source)) {
+    throw new Error('Both target and source must be plain objects');
+  }
+
+  // iterate over all enumerable properties
```
foolip, line 45:
```suggestion
  // iterate over own enumerable properties
```

index.js
```diff
@@ -34,6 +33,28 @@ function load() {
...
+  for (const [key, value] of Object.entries(source)) {
+    // recursively extend if target has the same key, otherwise just assign
+    if (Object.prototype.hasOwnProperty.call(target, key)) {
+      target[key] = extend(extend({}, target[key]), value);
```
foolip, line 49:
```suggestion
      extend(target[key], value);
```

Leaning into the fact that `extend` modifies in-place.

index.js
```diff
@@ -34,6 +33,28 @@ function load() {
...
+    } else {
+      target[key] = value;
+    }
+  }
+
+  return target;
```
foolip, line 55: The return statement won't be needed.

foolip (approved):
I have confirmed again that this results in exactly the same object (if written to disk as JSON) as before. Suggested some additional simplification, but it's not super important.

I'll leave the final review of this to @ddbeck or @Elchi3.

ddbeck:
Hi @jonathantneal, were you planning to come back to this PR?

foolip:
@ddbeck I've applied my own suggestions now and think this is good to go without @jonathantneal making any further changes, if you want to give it a review.

foolip:
I've also confirmed that the resulting data (serialized as JSON) is still exactly the same with this change, comparing commit 176d4ed9bf017de1a07d738a4b7f6c594a19b968 and this PR merged into the same commit.

jonathantneal:
Thanks, @foolip! 👍

foolip:
Ping @ddbeck for review. If this goes stale, the work to validate that it doesn't change anything will have to be done again.

ddbeck (approved):
Thank you! 🎉

ddbeck:
Actually, that wasn't enough: excellent work here, @jonathantneal and @foolip! I'm really pleased with what this achieves and the data problems it should prevent in the future. Well done!

PR #9836, opened 2021-04-13
Drop mdn_url for the appcache "manifest" attribute

sideshowbarker:
https://github.com/mdn/content/commit/80f1176 removed the https://developer.mozilla.org/docs/Web/HTML/Using_the_application_cache article from MDN.

(That article isn’t 404 yet because there’s a lag between when the source gets removed from the repo and when it goes away in production.)

### DIFF
--- a/html/elements/html.json
+++ b/html/elements/html.json
@@ -51,7 +51,6 @@
         },
         "manifest": {
           "__compat": {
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
Looks good to me

Compat data updates, see https://github.com/mdn/browser-compat-data/issues/7480

PR #9839, opened 2021-04-13
Update Chromium version for ping content attributes

foolip:
Chrome 12 is inferred from WebKit trunk version 534.27.0 when the IDL
attribute was added:
https://github.com/mdn/browser-compat-data/pull/9470#issuecomment-817091562

That came after the content attribute, but that behavior was behind a
preference, so this could still be slightly wrong.

However, support for the network requests was confirmed in Chrome 15, so
at most this could be wrong in a range between Chrome 12-15.

### DIFF
--- a/html/elements/a.json
+++ b/html/elements/a.json
@@ -451,10 +451,10 @@
           "__compat": {
             "support": {
               "chrome": {
-                "version_added": true
+                "version_added": "12"
               },
               "chrome_android": {
-                "version_added": true
+                "version_added": "18"
               },
               "edge": {
                 "version_added": "17"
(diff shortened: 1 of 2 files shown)


foolip:
> However, support for the network requests was confirmed in Chrome 15, so
> at most this could be wrong in a range between Chrome 12-15.

Hmm, actually the behavior could have been in Chrome 11 and the IDL attribute in 12, but who cares :)

ddbeck (approved):
Looks great (and I endorse accepting the possibility of error for Chrome 11-15). Thank you, @foolip!

PR #9841, opened 2021-04-13
Add opera support for :focus-visible

lukewarlow:
Updated :focus-visible support in Opera and Opera Android.

Opera Desktop v72 is based on Chromium 86 which is where support was added. https://blogs.opera.com/desktop/changelog-for-72/#b3779.0

Opera Android v61 is also based on Chromium 86. https://forums.opera.com/post/235956


ddbeck (approved):
Looks good. Thank you! 🎉

PR #9842, opened 2021-04-13
Fix the nesting of not_match_link features

foolip:
These were accidentally both defining css.selectors.not_match_link,
merged together by `extend`.

This removes css.selectors.not_match_link and adds
css.selectors.any-link.not_match_link + css.selectors.link.not_match_link.

Note that a correctly nested css.selectors.visited.not_match_link
already existed.


ddbeck (approved):
Looks good. Thank you! 🎉

PR #9853, opened 2021-04-14
Add more Bluetooth* and BarcodeDetector spec URLs

sideshowbarker:
These are some additional spec URLs that didn’t get added in the first pass for the B part of the API tree because the MDN articles are relatively recent.


Elchi3 (approved):
Yay new docs! 👍

PR #9857, opened 2021-04-14
Chrome 91 support for at-rule counter-style

rachelandrew:
Chrome 91 adds support for most of `@counter-style`, so I've updated BCD. Also clarified that no browser (including Firefox) implements `<image>` for the `symbols` descriptor.

https://chromestatus.com/features/5692693659254784

### DIFF
--- a/css/at-rules/counter-style.json
+++ b/css/at-rules/counter-style.json
@@ -8,10 +8,10 @@
           "spec_url": "https://drafts.csswg.org/css-counter-styles/#the-counter-style-rule",
           "support": {
             "chrome": {
-              "version_added": false
+              "version_added": "91"
             },
             "chrome_android": {
-              "version_added": false
+              "version_added": "91"
             },
             "edge": {
               "version_added": false
@@ -41,7 +41,7 @@
               "version_added": false
             },
             "webview_android": {
-              "version_added": false
+              "version_added": "91"
             }
           },
           "status": {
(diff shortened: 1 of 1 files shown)


css/at-rules/counter-style.json
```diff
@@ -457,19 +457,23 @@
             "description": "<code>symbols</code>",
             "support": {
               "chrome": {
-                "version_added": false
+                "version_added": "91",
+                "notes": "Does not support <code>&lt;image&gt;</code> as a value for the <code>symbols</code> descriptor."
```
ddbeck, line 461: It's a bit worrying when there's a repeating note. It often suggests that the data needs to change too. Since the spec says this ought to be supported, perhaps we should set `partial_implementation` as well?

ddbeck (changes requested):
One idea for you in a line comment, @rachelandrew. Otherwise, looks good. Thank you!

ddbeck:
Hi @rachelandrew, do you plan to return to this PR?

css/at-rules/counter-style.json
rachelandrew: I'm not sure I understand, which data needs updating?

rachelandrew:
@ddbeck yes, sorry, I'm not sure I understand the comment.

css/at-rules/counter-style.json
ddbeck: Sorry for the confusion. I'm suggesting either one of:

1. Set `"partial_implementation": true` for every entry with "Does not support <code>&lt;image&gt;</code> as a value for the <code>symbols</code> descriptor".
2. Make a subfeature to capture support for `<image>` and drop the notes.

I prefer the first option, in this case.

ddbeck (approved):

PR #9860, opened 2021-04-14
Edge and Opera data border-radius logical props

rachelandrew:
Edge and Opera (desktop) now support the logical version of the border-radius longhands. I've tested these.


ddbeck (approved):
Thank you! 🎉

PR #9868, opened 2021-04-14
Update for Samsung Internet 14.0 (and one at 14.2)

AdaRoseCannon:
This updates the data for Samsung Internet, it uses the mirror script to generate the changes.

### DIFF
--- a/api/AbortPaymentEvent.json
+++ b/api/AbortPaymentEvent.json
@@ -35,7 +35,7 @@
             "version_added": false
           },
           "samsunginternet_android": {
-            "version_added": true
+            "version_added": "10.0"
           },
           "webview_android": {
             "version_added": false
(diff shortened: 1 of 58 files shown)


vinyldarkscratch:
Thanks @AdaRoseCannon!  Giving this PR a quick skim, there's some interesting changes coming from the mirror script that should be looked at a little more carefully.  While the first changes, which set `true` to `10.0`, are fine, I'm also noticing instances where the original version number is changed (i.e. `11.0` to `12.0`).

Can we revert changes to data that isn't `false` -> `14.0` (or the browser file, of course) so we can get this PR merged quicker?

AdaRoseCannon:
Sorry for the slow response i've had a work blocker. Getting back to it now.

AdaRoseCannon:
I fixed the 11.0 to 12.0 error, nice spot.

The ones which are changing to 14.2 are because it was a mistake we had decided to not support the prefers color scheme whilst we worked on our own implementation.

I also updated the browser data again since 14.0 just became stable

vinyldarkscratch (approved):
Alright, thanks for the updates!  I trust the data, and will merge this right away!

PR #5795, opened 2020-03-07
refactor: Avoid `require(…)` of JSON files

ExE-Boss:
This avoids having to go through **Node’s** `require(…)` resolution algorithm.

It will also make eventual migration of the test files to [**ECMAScript Modules**](https://nodejs.org/api/esm.html) easier.

PR #9874, opened 2021-04-15
Load individual JSON files with fs instead of require()'ing

foolip:
This is slightly faster, probably mainly because it avoids the module
cache, which will hold onto the parsed result of each file individually.

The time to load was measured 10 times with and without flushing the
disk cache (using `purge` on macOS).

Before: 1.397-1.510s uncached, 0.641-0.695 cached

After: 1.163-1.397s uncached, 0.500-0.555s cached

### DIFF
--- a/index.js
+++ b/index.js
@@ -17,7 +17,7 @@ function load() {
       extra = load(fp);
     } else if (path.extname(fp) === '.json') {
       try {
-        extra = require(fp);
+        extra = JSON.parse(fs.readFileSync(fp));
       } catch (e) {}
     }
 


ddbeck:
This was proposed a while back and I didn't go for it: https://github.com/mdn/browser-compat-data/pull/5795 (mostly because I wasn't ready to commit to the motive, not that the diff was really a problem, looking back). There's a bunch of places where we use `require` instead of `JSON.parse()`. Is this something we should address all at once?

foolip:
I don't think it's really a problem to use `require()` if no difference can be measured or if speed just doesn't matter.

Do the other places already `require('fs')`? If so I could replace them all for consistency.

WDYT @ddbeck?

ddbeck (approved):
> Do the other places already `require('fs')`?

I checked. Not all of them, no. It's a bit fussy to do this everywhere, so let's just do this here for now. Thank you!

PR #9875, opened 2021-04-15
Don't let compare-versions cover up schema errors

ddbeck:
I'm not completely sure sure why (the code we had before was hard to understand) but the improved code in #5576 prevents the friendlier error messages from showing when you enter an unsanctioned version string. This PR catches a likely error thrown by `compare-versions`.

Right now, you get something like

```
Problems in 1 file:
✖ css/selectors/-webkit-details-marker.json
Error: Invalid argument not valid semver ('test' received)
    <ugly traceback here>
```

Instead of the previous:

```
Problems in 1 file:
✖ css/selectors/-webkit-details-marker.json
  Versions – 9 errors:
  → css.selectors.-webkit-details-marker - version_added: "test" is NOT a valid version number for chrome
    Valid chrome versions are: <a bunch of valid versions>
```

Since I think the latter is nicer to look at, I've added a `try … catch` to allow this to happen again.

### DIFF
--- a/test/linter/test-versions.js
+++ b/test/linter/test-versions.js
@@ -52,13 +52,19 @@ function isValidVersion(browserIdentifier, version) {
  * @returns {boolean}
  */
 function removedAfterAdded(statement) {
-  return compareVersions.compare(
-    statement.version_added.startsWith('≤')
-      ? '0' // 0 was chosen as it's a number lower than any possible browser version
-      : statement.version_added,
-    statement.version_removed.replace('≤', ''),
-    '>=',
-  );
+  try {
+    return compareVersions.compare(
+      statement.version_added.startsWith('≤')
+        ? '0' // 0 was chosen as it's a number lower than any possible browser version
+        : statement.version_added,
+      statement.version_removed.replace('≤', ''),
+      '>=',
+    );
+  } catch (err) {
+    if (err.message === 'Invalid argument not valid semver') {
+      return null;
+    }
+  }
 }
 
 /**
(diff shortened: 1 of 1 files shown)


vinyldarkscratch:
Huh, not sure why this issue wasn't coming up before, but I can understand why it is.  Thanks for writing a fix!

If we want to avoid try-catch blocks, one possibility could be to take advantage of `compare-versions`' [`validate`](https://www.npmjs.com/package/compare-versions#validate-version-numbers) function, AKA something along the lines of:

```js
function removedAfterAdded(statement) {
  if (!(
    compareVersions.validate(statement.version_added.replace('≤', '')) && 
    compareVersions.validate(statement.version_removed.replace('≤', ''))
  )) {
    return null;
  }

  return compareVersions.compare(
    statement.version_added.startsWith('≤')
      ? '0' // 0 was chosen as it's a number lower than any possible browser version
      : statement.version_added,
    statement.version_removed.replace('≤', ''),
    '>=',
  );
}
```


WDYT?

vinyldarkscratch (approved):
Loving it!  Thanks, @ddbeck, looking great to me!  (I'm not merging this due to my peer role, unless you'd like me to!)

ddbeck:
Oh, so I discovered a funny thing: this only fixes some cases of the error bubbling up (there's another condition where we attempt `compare()` with potentially invalid version strings). I'll merge this, since it fixes one of them, but gosh the linter needs a real refactoring plan at some point.

PR #9896, opened 2021-04-17
Drop the ToC from docs/data-guidelines.md

sideshowbarker:
Per https://github.blog/changelog/2021-04-13-table-of-contents-support-in-markdown-files/ the GitHub UI now provides a button for displaying a generated ToC for all markdown files — making our in-document ToC in the docs/data-guidelines.md file somewhat redundant. So this change removes it.


ddbeck (approved):
It's not completely redundant—it's nice to have outside of GitHub itself—but it's so fussy to update the ToC. 👍

PR #9904, opened 2021-04-18
Update Safari Service Worker Client/WindowClient support data

tgecho:
Updates the browser compatibility data to show Safari mostly supports the Service Worker Client/WindowClient APIs.

It appears that support was included with the initial release of Service Workers support in macOS Safari 11.1 and iOS Safari 11.3 (Changeset: https://trac.webkit.org/changeset/224880/webkit)

Based on this line in one of the modified files, it does seem that this changeset was included in 11.1: https://trac.webkit.org/browser/webkit/releases/Apple/Safari%2011.1/WebCore/workers/service/ServiceWorker.cpp#L38

iOS Safari 11.3: https://trac.webkit.org/browser/webkit/releases/Apple/iOS%2011.3/WebCore/workers/service/ServiceWorker.cpp#L38

### Testing

I was able to verify that it at least works in Safari 14.0.3 (macOS 11.2.3). Here's a screenshot of the dev tools showing a `WindowClient` object returned by `clients.matchAll` in the service worker.

Screen Shot 2021-04-18 at 10 40 02 AM

I was also able to send a message back to the client tab with `client.postMessage(...)`.

**Not marked as supported:** `.focus(...)` and `.navigate(...)` on WindowClient throw a "not implemented" exception. Everything else appears to work as expected.

I have not had a chance to dig up an older version (or test in iOS) to verify exactly when this started working. If that's a blocker I'll see what I can come up with.

### DIFF
--- a/api/Client.json
+++ b/api/Client.json
@@ -31,10 +31,10 @@
             "version_added": "27"
           },
           "safari": {
-            "version_added": false
+            "version_added": "11.1"
           },
           "safari_ios": {
-            "version_added": false
+            "version_added": "11.3"
           },
           "samsunginternet_android": {
             "version_added": "4.0"
@@ -80,10 +80,10 @@
               "version_added": "30"
             },
             "safari": {
-              "version_added": false
+              "version_added": "11.1"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "11.3"
             },
             "samsunginternet_android": {
               "version_added": "4.0"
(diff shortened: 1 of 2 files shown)


ddbeck (approved):
Thank you for this update, @tgecho! It's very well substantiated.

As final confirmation, I double-checked the results (via https://mdn-bcd-collector.appspot.com/tests/api/WindowClient and https://mdn-bcd-collector.appspot.com/tests/api/Client with BrowserStack) and got the same version numbers.

(The collector did report that `focus` was in the prototype, but it doesn't actually try to call it, so I'll assume that it doesn't actually work here.)

Thank you again! 🎉

PR #9913, opened 2021-04-19
Remove irrelevant Chromium flag data for FeaturePolicy (and fix Opera Android data)

vinyldarkscratch:
This PR removes irrelevant flag data for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `FeaturePolicy` API, along with other APIs containing a `featurePolicy` member (`Document`, `HTMLIFrameElement`...) as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-flag-data).

Additionally, this PR fixes an issue with the Opera Android data, where it was indicated `FeaturePolicy` was only available behind a flag.  Running a [quick test](https://mdn-bcd-collector.appspot.com/tests/api/Document/featurePolicy) on the almost-latest Opera Android showed normal support.  I'm certain it was introduced at the same time as Chrome Android, so I set the version number to the equivalent version.

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -4861,62 +4861,12 @@
         "__compat": {
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Document/featurePolicy",
           "support": {
-            "chrome": [
-              {
-                "version_added": "74"
-              },
-              {
-                "version_added": "73",
-                "version_removed": "74",
-                "flags": [
-                  {
-                    "type": "preference",
-                    "name": "#enable-experimental-productivity-features",
-                    "value_to_set": "Enabled"
-                  }
-                ]
-              },
-              {
-                "version_added": "69",
-                "version_removed": "73",
-                "alternative_name": "policy",
-                "flags": [
-                  {
-                    "type": "preference",
-                    "name": "#enable-experimental-productivity-features",
-                    "value_to_set": "Enabled"
-                  }
-                ]
-              }
-            ],
-            "chrome_android": [
-              {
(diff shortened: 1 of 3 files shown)


api/Document.json
```diff
@@ -4861,62 +4861,12 @@
...
-                  {
-                    "type": "preference",
-                    "name": "#enable-experimental-productivity-features",
-                    "value_to_set": "Enabled"
-                  }
-                ]
-              }
-            ],
+            "chrome": {
+              "version_added": "74"
```
foolip, line 4865: Chrome 74 was released on 2019-04-23, so it'll be another 3 days before this can be merged. But more importantly, it suggests a bug in the code determining if a flag can be removed yet.

vinyldarkscratch: I haven't yet fixed the script to check the version of the previous (AKA, most recent) statement and determine removal based upon that (vs. the version number from just the flag data).  Since it's only a few days before Chrome 74 reaches the two-year mark, I figured it'd be fine to include those removals.

foolip: Let's merge this after tomorrow's release to avoid having the discussion :)

ddbeck:
This is closely related to https://github.com/mdn/browser-compat-data/pull/9707

vinyldarkscratch:
I double-checked the data to ensure that the APIs were enabled by default in Chrome 74 (rather than earlier versions), and am happy to report that the data is accurate!

PR #9914, opened 2021-04-19
Mark originAgentCluster as shipped in Edge and WebView 90

foolip:
See https://github.com/mdn/browser-compat-data/issues/8891.


ddbeck (approved):
Thank you! 🎉

PR #9916, opened 2021-04-19
Remove irrelevant all browsers flag data for Request API

vinyldarkscratch:
This PR removes irrelevant flag data for all browsers for the `Request` API as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-flag-data).

Paired with #9751 and #9917.


api/Request.json
```diff
@@ -902,75 +692,31 @@
...
-                "flags": [
-                  {
-                    "type": "preference",
-                    "name": "Experimental Web Platform Features"
-                  }
-                ]
-              }
-            ],
+            "chrome": {
+              "version_added": "42"
```
foolip, line 696: Was this really removed from Chrome for Android but continues to ship in Chrome desktop?

api/Request.json
```diff
@@ -1139,75 +885,31 @@
...
```
foolip, line 889: The below Chrome for Android removal is also suspect here.

api/Request.json
```diff
@@ -1377,20 +1079,9 @@
...
```
foolip, line 1083: Chrome for Android being `false` below is suspect.

api/Request.json
```diff
@@ -97,75 +53,31 @@
           "description": "<code>Request()</code> constructor",
           "support": {
             "chrome": {
-              "version_added": "41",
+              "version_added": "42",
```
foolip, line 56: Good, removing flags reveals this kind of error in the data. I presume the lint was failing and that's how you noticed?

foolip (commented):
Some pre-existing issue with the data here...

api/Request.json
vinyldarkscratch: Fixed up the data in #10070!

api/Request.json
vinyldarkscratch: You guessed it!  This seems to be pretty common among both flagged and prefixed data.

foolip: With a prefixed parent feature it should still usually be correct, since interface and member are rarely (never?) both prefixed, but yeah there are similar issues.

vinyldarkscratch:
Now that the Chrome Android data's been fixed up, I believe this is ready for re-review!

api/Request.json
```diff
@@ -571,74 +449,30 @@
...
-                    "name": "Experimental Web Platform Features"
-                  }
-                ]
-              }
-            ],
+            "chrome": {
+              "version_added": "42"
+            },
             "chrome_android": {
               "version_added": false
```
foolip, line 456: I don't believe that this isn't supported on Chrome for Android, but this PR doesn't do anything with the entry, so OK to leave it for later.

foolip (approved):

api/Request.json
vinyldarkscratch: Submitted a PR to correct this!  #10565

PR #9925, opened 2021-04-19
Remove irrelevant Firefox flag data for StorageManager API (and fix FxA data)

vinyldarkscratch:
This PR removes irrelevant flag data for Firefox and Firefox Android for the `StorageManager` API as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-flag-data).

Additionally, this PR resolves issues with the Firefox Android data.  It was determined that this API is indeed supported in Firefox Android, and since the same bugs were linked for both Firefox Desktop and Android, I feel it's safe to say that it was supported in the same version.

### DIFF
--- a/api/StorageManager.json
+++ b/api/StorageManager.json
@@ -13,23 +13,9 @@
           "edge": {
             "version_added": "≤79"
           },
-          "firefox": [
-            {
-              "version_added": "57"
-            },
-            {
-              "version_added": "51",
-              "version_removed": "57",
-              "notes": "See <a href='https://bugzil.la/1304966'>bug 1304966</a> and <a href='https://bugzil.la/1399038'>bug 1399038</a>.",
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "dom.storageManager.enabled",
-                  "value_to_set": "true"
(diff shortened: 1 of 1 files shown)


api/StorageManager.json
```diff
@@ -13,23 +13,9 @@
...
-                }
-              ]
-            }
-          ],
+          "firefox": {
+            "version_added": "57"
+          },
           "firefox_android": {
             "version_added": "51",
             "notes": "See <a href='https://bugzil.la/1304966'>bug 1304966</a> and <a href='https://bugzil.la/1399038'>bug 1399038</a>.",
```
foolip, line 21: These bugs are the same ones as you removed above, and are closed. Was this really not shipped on Android? If it isn't, another bug needs to be linked.

vinyldarkscratch: This is shown as supported in Firefox Android, and since the same bugs as Firefox Desktop are linked, it was most likely enabled at the same time as well.  Fixing up the data now!

foolip (approved):

PR #9931, opened 2021-04-19
Remove irrelevant "layout.css.prefixes.webkit" flag

vinyldarkscratch:
This PR removes irrelevant flag data for `layout.css.prefixes.webkit` as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-flag-data).

### DIFF
--- a/css/properties/user-select.json
+++ b/css/properties/user-select.json
@@ -55,17 +55,6 @@
               {
                 "prefix": "-moz-",
                 "version_added": "4"
-              },
-              {
-                "prefix": "-webkit-",
-                "version_added": "44",
-                "flags": [
-                  {
-                    "type": "preference",
-                    "name": "layout.css.prefixes.webkit",
-                    "value_to_set": "true"
-                  }
-                ]
               }
             ],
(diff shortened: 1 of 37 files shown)


css/properties/user-select.json
```diff
@@ -45,17 +45,6 @@
...
-                "flags": [
-                  {
-                    "type": "preference",
-                    "name": "layout.css.prefixes.webkit",
-                    "value_to_set": "true"
-                  }
-                ]
               }
             ],
             "firefox_android": [
```
ddbeck, line 50: This one leaves Firefox for Android untouched. Some reason?

ddbeck (changes requested):

css/properties/user-select.json
vinyldarkscratch: Might've been me trying not to include the FxA changes that aren't ready yet.  Fixed!

ddbeck (approved):

PR #9932, opened 2021-04-19
Update Firefox data for `supports` CSS at-rule

vinyldarkscratch:
This PR updates the Firefox data for the `supports` at-rule by doing the following:
- Removes the now-irrelevant flag that enables the feature
- Fixes the Firefox Android data to say that it is now supported in Fenix (confirmed by testing via https://mdn-bcd-collector.appspot.com)

### DIFF
--- a/css/at-rules/supports.json
+++ b/css/at-rules/supports.json
@@ -66,12 +66,16 @@
               "edge": {
                 "version_added": "83"
               },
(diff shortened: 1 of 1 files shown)


foolip (commented):
How about just fixing the data error for now, and removing the flag data at the same time when allowed?

PR #9933, opened 2021-04-19
Adopt new definition and guideline for experimental status

### LINKED ISSUE #6905: Improve documentation of experimental status

ddbeck:
This PR:

- Proposes a new, hopefully clearer definition of experimental in the schema docs, for consumers. This is less an attempt to change the definition, so much as to describe what it _actually_ is. It has an additional benefit of sidestepping standards questions.
- Proposes a guideline for setting and changing the `experimental` status, for contributors. This is an attempt to get consensus on how we set the status as PR authors and reviewers.
- Reformats the other status definitions in the schema docs for symmetry with the new definition.

That said, this is probably not complete. See my line comment on sunsetting experimental features. There's a wrinkle that I don't like and I'm not sure how to resolve it.

Fixes #6905.

### DIFF
--- a/docs/data-guidelines.md
+++ b/docs/data-guidelines.md
@@ -280,3 +280,23 @@ Members of this mixin are available to `HTMLAnchorElement` and `HTMLAreaElement`
    ```
 
 This guideline was proposed in [#8929](https://github.com/mdn/browser-compat-data/issues/8929), based in part on previous discussion in [#472](https://github.com/mdn/browser-compat-data/issues/472).
+
+## Choosing an experimental status
+
+Generally, when a feature is supported by one and only one browser engine, set `experimental` to `true`. When a feature is supported by two or more engines, then set `experimental` to `false`. Some exceptions apply, however, for long-standing features and features behind flags and prefixes.
+
+If a feature is supported behind flags only, no matter how many engines, then set `experimental` to `true`.
+
+If a feature is supported behind incompatible prefixes only (such as `-webkit-` in one engine and `-moz-` in another), no matter how many engines support the feature overall, then set `experimental` to `true`. If two or more engines support a feature behind a common prefix (such as `-webkit-` only), then set `experimental` to `false`.
+
+If a feature has been supported by one and only one engine without major changes for two or more years (relative to the most-recent browser release introducing support for the feature, or the most-recent browser release since the last major change) , then `experimental` may be set to `false`.
+
+| Example                                                                     | Experimental |
+| --------------------------------------------------------------------------- | ------------ |
+| An API supported in Chrome and Firefox, without flags or prefixes.          | No           |
+| A CSS property supported in Chrome and Firefox, with the `-webkit-` prefix. | No           |
+| An HTTP header supported in Chrome and Firefox, behind flags.               | Yes          |
+| A CSS value supported in Safari, released last week.                        | Yes          |
+| An API supported in Firefox, released three years ago.                      | No           |
+
+This guideline was proposed in [#6905](https://github.com/mdn/browser-compat-data/issues/6905) and adopted in [#TK]().
--- a/schemas/compat-data-schema.md
+++ b/schemas/compat-data-schema.md
@@ -447,10 +447,12 @@ A `matches` object contains hints to help automatically detect whether source co
 The mandatory status property contains information about stability of the feature. It is
 an object named `status` and has three mandatory properties:
 
-- `experimental`: a `boolean` value that indicates this functionality is
-  intended to be an addition to the Web platform. Some features are added to
-  conduct tests. Set to `false`, it means the functionality is mature, and no
-  significant incompatible changes are expected in the future.
+- `experimental`: a `boolean` value.
+
+  If `experimental` is `true`, it means that Web developers should experiment with this feature and provide feedback to browser vendors and standards authors about this feature. It also means that Web developers _should not_ rely on the feature's continued existence in its current (or potentially any) form in future browser releases.
+
+  If `experimental` is `false`, it means the functionality is mature and no significant changes are expected in the future.
+
 - `standard_track`: a `boolean` value indicating if the feature is part of an
   active specification or specification process.
 - `deprecated`: a `boolean` value that indicates if the feature is no longer recommended.
--- a/schemas/compat-data-schema.md
+++ b/schemas/compat-data-schema.md
@@ -453,10 +453,13 @@ an object named `status` and has three mandatory properties:
 
   If `experimental` is `false`, it means the functionality is mature and no significant changes are expected in the future.
 
-- `standard_track`: a `boolean` value indicating if the feature is part of an
-  active specification or specification process.
-- `deprecated`: a `boolean` value that indicates if the feature is no longer recommended.
-  It might be removed in the future or might only be kept for compatibility purposes. Avoid using this functionality.
+- `standard_track`: a `boolean` value.
+
+  If `standard_track` is `true`, then the feature is part of an active specification or specification process.
+
+- `deprecated`: a `boolean` value.
+
+  If `deprecated` is `true`, then the feature is no longer recommended. It might be removed in the future or might only be kept for compatibility purposes. Avoid using this functionality.
 
 ```json
 "__compat": {


docs/data-guidelines.md
```diff
@@ -280,3 +280,23 @@ Members of this mixin are available to `HTMLAnchorElement` and `HTMLAreaElement`
    ```
 
 This guideline was proposed in [#8929](https://github.com/mdn/browser-compat-data/issues/8929), based in part on previous discussion in [#472](https://github.com/mdn/browser-compat-data/issues/472).
+
+## Choosing an experimental status
+
+Generally, when a feature is supported by one and only one browser engine, set `experimental` to `true`. When a feature is supported by two or more engines, then set `experimental` to `false`. Some exceptions apply, however, for long-standing features and features behind flags and prefixes.
+
+If a feature is supported behind flags only, no matter how many engines, then set `experimental` to `true`.
```
ddbeck, line 288: Note that I _carefully_ avoided the word "implementation" throughout this guideline. We have proposals to handle stuff in development with things `implementation_url`. It seems to me that if it's not on by default, it doesn't count.

docs/data-guidelines.md
```diff
@@ -280,3 +280,23 @@ Members of this mixin are available to `HTMLAnchorElement` and `HTMLAreaElement`
...
+
+If a feature is supported behind incompatible prefixes only (such as `-webkit-` in one engine and `-moz-` in another), no matter how many engines support the feature overall, then set `experimental` to `true`. If two or more engines support a feature behind a common prefix (such as `-webkit-` only), then set `experimental` to `false`.
+
+If a feature has been supported by one and only one engine without major changes for two or more years (relative to the most-recent browser release introducing support for the feature, or the most-recent browser release since the last major change) , then `experimental` may be set to `false`.
```
ddbeck, line 292: OK, this is the sunsetting part of the guideline. It started out as this:

> If a feature has been supported by one and only one engine for two or more years without major changes, then set `experimental` to `false`.

But I had a bit of trouble with it. We have release dates for browser versions, not engines. So the new text contains quite the parenthetical to address that. It's trying to capture when features achieve stasis, if not consensus. It's still awkward for reviews and authors (you have to look at all the release dates to figure out what the right value is), but it's answerable and might accommodate oddities (like Edge supporting a thing before Chrome) that going with least-recent release wouldn't.

docs/data-guidelines.md
```diff
@@ -280,3 +280,23 @@ Members of this mixin are available to `HTMLAnchorElement` and `HTMLAreaElement`
...
 
 This guideline was proposed in [#8929](https://github.com/mdn/browser-compat-data/issues/8929), based in part on previous discussion in [#472](https://github.com/mdn/browser-compat-data/issues/472).
+
+## Choosing an experimental status
+
+Generally, when a feature is supported by one and only one browser engine, set `experimental` to `true`. When a feature is supported by two or more engines, then set `experimental` to `false`. Some exceptions apply, however, for long-standing features and features behind flags and prefixes.
+
+If a feature is supported behind flags only, no matter how many engines, then set `experimental` to `true`.
+
+If a feature is supported behind incompatible prefixes only (such as `-webkit-` in one engine and `-moz-` in another), no matter how many engines support the feature overall, then set `experimental` to `true`. If two or more engines support a feature behind a common prefix (such as `-webkit-` only), then set `experimental` to `false`.
```
foolip, line 290: This exception makes sense to me, well carved!

docs/data-guidelines.md
```diff
@@ -280,3 +280,23 @@ Members of this mixin are available to `HTMLAnchorElement` and `HTMLAreaElement`
...
+
+If a feature has been supported by one and only one engine without major changes for two or more years (relative to the most-recent browser release introducing support for the feature, or the most-recent browser release since the last major change) , then `experimental` may be set to `false`.
+
+| Example                                                                     | Experimental |
+| --------------------------------------------------------------------------- | ------------ |
+| An API supported in Chrome and Firefox, without flags or prefixes.          | No           |
+| A CSS property supported in Chrome and Firefox, with the `-webkit-` prefix. | No           |
+| An HTTP header supported in Chrome and Firefox, behind flags.               | Yes          |
+| A CSS value supported in Safari, released last week.                        | Yes          |
```
foolip, line 299:
```suggestion
| A CSS property supported in Safari, released last week.                        | Yes          |
```

(Or "property value", but I guess you were going for a simple example.)

ddbeck: I saw this after opening the PR and thought to myself, who's going to notice? 😂

(Fixed.)

docs/data-guidelines.md
foolip: OK, this is the good stuff that we need to get right.

I think your original attempt was pretty good, but I see the problem. Throughout this guideline, one could understand "engine" as "the set of browsers using the same engine" and it mostly works. The question is which date to use here. I think it should actually be least-recent, because there is a high variability in when a Chromium version rolls out to all of Chrome (+Android), Edge, Opera (+Android), Samsung Internet and WebView. I think Opera for Android or Samsung Internet will tend to be the last.

I would propose:

> Otherwise, if a feature has been supported by some browser for two or more years without major changes and there is no support behind a flag in some other browser with major differences from the first, then set `experimental` to `false`.

The "Otherwise" is to not have to also carve out the preceding cases by saying "one and only one" in some way.

docs/data-guidelines.md
```diff
@@ -280,3 +280,26 @@ Members of this mixin are available to `HTMLAnchorElement` and `HTMLAreaElement`
...
+
+## Choosing an experimental status
+
+Generally, when a feature is supported by one and only one browser engine, set `experimental` to `true`. When a feature is supported by two or more engines, then set `experimental` to `false`. Some exceptions apply, however, for long-standing features and features behind flags and prefixes.
+
+If a feature is supported behind flags only, no matter how many engines, then set `experimental` to `true`.
+
+If a feature is supported behind incompatible prefixes only (such as `-webkit-` in one engine and `-moz-` in another), no matter how many engines support the feature overall, then set `experimental` to `true`. If two or more engines support a feature behind a common prefix (such as `-webkit-` only), then set `experimental` to `false`.
+
+A single-engine feature's `experimental` status may expire and switch to `false` when the following conditions are met:
```
ddbeck, line 292: OK, this picks up where @foolip left off at https://github.com/mdn/browser-compat-data/pull/9933#discussion_r616103899

> I would propose:
> 
> > Otherwise, if a feature has been supported by some browser for two or more years without major changes and there is no support behind a flag in some other browser with major differences from the first, then set `experimental` to `false`.

I thought this was a really clever way to sidestep my concern about intra-engine releases getting ahead (or behind) each other. I took this idea and rephrased it, because I think it's a little bit tough to parse as a single sentence.

What do you think?

foolip: I like it, I don't have any nits, looks good as-is.

docs/data-guidelines.md
jpmedley: Completely agree where Chrome is concerned. This is especially true in light of more recent use of our flag mechanism, which is now [explicitly earlier than an origin trial](https://www.chromium.org/blink/launching-features), which are not official parts of the web platform.

docs/data-guidelines.md
```diff
@@ -280,3 +280,26 @@ Members of this mixin are available to `HTMLAnchorElement` and `HTMLAreaElement`
...
+
+- The feature has been supported by default and without major changes by some browser for two or more years.
+- If any other browser engine supports the feature behind a flag, then the behaviors are mutually compatible.
+
+| Example                                                                     | Experimental |
+| --------------------------------------------------------------------------- | ------------ |
+| An API supported in Chrome and Firefox, without flags or prefixes.          | No           |
+| A CSS property supported in Chrome and Firefox, with the `-webkit-` prefix. | No           |
+| An HTTP header supported in Chrome and Firefox, behind flags.               | Yes          |
+| A CSS property value supported in Safari, released last week.               | Yes          |
```
jpmedley, line 302: What if the feature is in a W3C recommendation and it just happens that Safari reached the finish line first?

ddbeck: If Safari crosses the finish line first, then it's still experimental until at least one engine catches up.

Part of the goal here is to clearly distinguish between experimental and standard track. Under this revised approach, experimental can be set independently of the standard track value (that is to say, the two values don't depend on each other—you don't even have to _know_ what the standards status is to set the experimental status). This makes it more clear what experimental actually adds to our data set (a marker of consensus, or failing that, stasis) and lets us deal with features that are unchanging, but not standardized.

jpmedley: I get all that. It just seems misapplied here. If a feature is a full W3C recommendation, it's not a matter of if the second browser will implement. It's a matter of when. Knowing that fact is useful to developers. This information can help developers make other vendors aware that they should prioritize something. Here's how this would play out for Chrome.

1. Safari implements a full recommendation.
2. Developers see that Chrome hasn't implemented it yet, but they want it.
3.  They do one of three things: star an existing ticket on crbug.com, open a ticket on crbug.com, or they star the feature's entry on chromestatus.com.
4. Chrome management sees a surge in developer interest and decides to prioritize it over The Obscure API. (Developer interest isn't the only thing we consider, but we do prioritize developer interest if we see a lot of it.)

ddbeck: > It just seems misapplied here.

First, a question: do you mean that the guideline and definition _as written_ shouldn't result in this outcome? Or that you'd like to see different text that leads to a different outcome?

Second, what I want from this PR is to encourage the scenario you describe. Ideally, the experimental flag should be a signal to web developers to tell browser vendors and spec authors what they think of an emerging feature, including saying "I want this feature." I want to upstream the new experimental definition to MDN itself too, so this signal is stronger.

jpmedley: You overall heuristic seems misapplied to the case of a completely speced feature with one implementation, where 'speced feature' means one in a full recommendation.

>what I want from this PR is to encourage the scenario you describe

The scenarios for a speced and unspeced feature are different.

* For features in candidate recommendations, or wicg specs, or whatever, the signal vendors need is should the experimentation continue?
* For features that are full recommendations, the signal vendors need is should we prioritize implementing feature X over feature Y? We have to do both because we agreed they should be part of the platform. Which do we want to do first?

Web developers would treat these differently as well.

* For truly experimental features, I may not want to use the feature yet, or my employer may not let me use the feature in production. My shop doesn't let me use an open source polyfill. (Yes, that's a thing. I worked in one. There were liability concerns.) My shop's approved framework says they'll support it with a polyfill, but it's not coming for six months. Perhaps I'm overly cautious because I previously embraced an experimental feature and it got dropped. (Remember WebVR?)
* For full recommendations, the most important concern with using is when will the other vendors catch up and will my framework support it in the meantime.

foolip: @jpmedley I'm not really following here. How good a spec here is actually very hard to judge, and browser implementers will not trust the status the spec claims for itself, whether that's "Unofficial Draft", "Living Standard", or "Recommendation".

I don't think we should use the spec's own claim about how stable it is as input to the "experimental" setting here, but base it entirely on implementation status, like @ddbeck is suggesting in this guideline.

Elchi3 (approved):

FiyaasAhmedKhan:
ddbeck:6905-new-experimental-definition

PR #1271, opened 2018-03-05
Update display: contents compat table

gpgekko:
Using the following sources:
 - https://caniuse.com/#feat=css-display-contents
 - https://www.chromestatus.com/feature/5663606012116992
 - https://bugs.webkit.org/show_bug.cgi?id=157477

PR #9940, opened 2021-04-19
Remove irrelevant all browsers flag data for display CSS property

vinyldarkscratch:
This PR removes irrelevant flag data for all browsers for the `display` CSS property as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-flag-data).

### DIFF
--- a/css/properties/display.json
+++ b/css/properties/display.json
@@ -52,91 +52,30 @@
         "contents": {
           "__compat": {
             "support": {
-              "chrome": [
-                {
-                  "version_added": "65"
-                },
-                {
-                  "version_added": "58",
-                  "version_removed": "65",
-                  "flags": [
-                    {
-                      "type": "preference",
-                      "name": "Enable experimental Web Platform features"
-                    }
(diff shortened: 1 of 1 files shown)


css/properties/display.json
```diff
@@ -52,91 +52,30 @@
...
-                      "value_to_set": "true"
-                    }
-                  ]
-                }
-              ],
+              "firefox": {
+                "version_added": "37"
+              },
               "firefox_android": {
                 "version_added": "57"
```
foolip, line 105: Is the difference here real, or actually a bug in the data? It doesn't seem very plausible to me.

vinyldarkscratch: The data originally came from #1271.  By the looks of it, the data was sourced from CanIUse, which doesn't retain data for Firefox Android other than the current version.  I doubt the data is correct, fixing!

foolip (approved):

PR #9942, opened 2021-04-19
Remove irrelevant "layout.css.masking.enabled" flag

vinyldarkscratch:
This PR removes irrelevant flag data for `layout.css.masking.enabled` as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-flag-data).

### DIFF
--- a/css/properties/mask-position.json
+++ b/css/properties/mask-position.json
@@ -80,7 +80,8 @@
                 "version_removed": "70"
               },
               "firefox_android": {
-                "version_added": "53"
+                "version_added": "53",
+                "version_removed": "79"
               },
               "ie": {
                 "version_added": false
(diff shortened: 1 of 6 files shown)


css/properties/mask-position.json
```diff
@@ -101,39 +75,13 @@
...
-                  ]
-                }
-              ],
+              "firefox": {
+                "version_added": "53",
+                "version_removed": "70"
+              },
+              "firefox_android": {
+                "version_added": "53"
+              },
```
foolip, line 84: This does not look plausible.

foolip (approved):

PR #9946, opened 2021-04-19
Remove irrelevant "layout.css.shape-outside.enabled" flag

vinyldarkscratch:
This PR removes irrelevant flag data for `layout.css.shape-outside.enabled` as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-flag-data).

### DIFF
--- a/css/properties/shape-outside.json
+++ b/css/properties/shape-outside.json
@@ -15,38 +15,12 @@
             "edge": {
               "version_added": "79"
             },
-            "firefox": [
-              {
-                "version_added": "62"
-              },
-              {
-                "version_added": "61",
-                "version_removed": "62",
-                "flags": [
-                  {
-                    "type": "preference",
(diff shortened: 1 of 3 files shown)


css/properties/shape-outside.json
```diff
@@ -467,39 +311,13 @@
...
-                    }
-                  ]
-                }
-              ],
+              "firefox": {
+                "version_added": "62",
+                "version_removed": "70"
+              },
+              "firefox_android": {
+                "version_added": "62"
```
foolip, line 319: What's going on here? Was this really removed on desktop but remains on Android?

vinyldarkscratch: Probably another issue due to the pause on FxA while the new version was in development.  Fixing!

foolip (approved):

PR #9948, opened 2021-04-19
Remove irrelevant Firefox flag data for text-combine-upright CSS property

vinyldarkscratch:
This PR removes irrelevant flag data for Firefox and Firefox Android for the `text-combine-upright` CSS property as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-flag-data).

### DIFF
--- a/css/properties/text-combine-upright.json
+++ b/css/properties/text-combine-upright.json
@@ -47,16 +47,6 @@
               {
                 "version_added": "48",
                 "notes": "Before version 48, Firefox did not implement layout support for tate-chū-yoko."
-              },
-              {
-                "version_added": "41",
-                "flags": [
-                  {
-                    "type": "preference",
-                    "name": "layout.css.text-combine-upright.enabled",
-                    "value_to_set": "true"
-                  }
-                ]
               }
             ],
             "firefox_android": [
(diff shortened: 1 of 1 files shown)


css/properties/text-combine-upright.json
```diff
@@ -47,16 +47,6 @@
               {
                 "version_added": "48",
                 "notes": "Before version 48, Firefox did not implement layout support for tate-chū-yoko."
```
foolip, line 49: This note end up being pointless with the flag data removed.

foolip (changes requested):

foolip (commented):
Looking at this again, I think the entries for 81 should also be removed. If being animatable is a bug it's the earliest entry that should have a note and `partial_implementation`, but I think it's more like the feature was too supported.

Since the current behavior is more restricted than the historical, pointing it out isn't really helpful. (When the feature has gotten bigger in some sense, it's different.)

But I'd want @ddbeck to weigh in here.

ddbeck:
@foolip I think it makes sense to have a note because it's possible that some web developers have relied on the buggy behavior. But I'd probably fold it into one support statement: added at 48, with a note saying, "Before version 81…".

(Though this note will be increasingly uninteresting as time goes on—might be nice to figure out an irrelevant note guideline someday.)

foolip (approved):

foolip:
Oops, lint is failing.

foolip (approved):

PR #9950, opened 2021-04-19
Remove irrelevant Firefox flag data for touch-action CSS property

vinyldarkscratch:
This PR removes irrelevant flag data for Firefox and Firefox Android for the `touch-action` CSS property as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-flag-data).

### DIFF
--- a/css/properties/touch-action.json
+++ b/css/properties/touch-action.json
@@ -18,38 +18,13 @@
             "edge": {
               "version_added": "12"
             },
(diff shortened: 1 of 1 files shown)


foolip (approved):
The existing notes here aren't perfect but removing flag data is an improvement :)

PR #9953, opened 2021-04-19
Remove irrelevant Firefox flag data for will-change CSS property

vinyldarkscratch:
This PR removes irrelevant flag data for Firefox and Firefox Android for the `will-change` CSS property as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-flag-data).

### DIFF
--- a/css/properties/will-change.json
+++ b/css/properties/will-change.json
@@ -15,40 +15,14 @@
             "edge": {
               "version_added": "79"
             },
-            "firefox": [
-              {
-                "version_added": "36",
-                "notes": "Before Firefox 43, Firefox supported <code>will-change: will-change</code>, which is invalid by the specification."
-              },
-              {
-                "version_added": "31",
-                "version_removed": "36",
-                "flags": [
-                  {
-                    "type": "preference",
-                    "name": "layout.css.will-change.enabled",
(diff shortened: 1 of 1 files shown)


css/properties/will-change.json
```diff
@@ -15,40 +15,14 @@
...
-                    "type": "preference",
-                    "name": "layout.css.will-change.enabled",
-                    "value_to_set": "true"
-                  }
-                ]
-              }
-            ],
+            "firefox": {
+              "version_added": "36",
+              "notes": "Before Firefox 43, Firefox supported <code>will-change: will-change</code>, which is invalid by the specification."
```
foolip, line 20: I don't think there's any value in keeping this note. @ddbeck?

ddbeck: I agree. This is very obscure. Based on [the bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1195884) (which is the only evidence I can find for `will-change: will-change`, apart from old translations of MDN pages), I'm not convinced this ever mattered to anyone who wasn't a spec author or implementer.

ddbeck (approved):
Thank you! 🎉

PR #9983, opened 2021-04-20
Add Edge versions for DocumentTimeline API

vinyldarkscratch:
This PR adds real values for Microsoft Edge for the `DocumentTimeline` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.1).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/API/DocumentTimeline

### DIFF
--- a/api/DocumentTimeline.json
+++ b/api/DocumentTimeline.json
@@ -12,7 +12,7 @@
             "version_added": "54"
           },
           "edge": {
-            "version_added": "≤79"
+            "version_added": "80"
           },
           "firefox": [
             {
@@ -87,7 +87,7 @@
               "version_added": "61"
             },
             "edge": {
-              "version_added": "79"
+              "version_added": "80"
             },
             "firefox": [
               {


api/DocumentTimeline.json
```diff
@@ -12,7 +12,7 @@
             "version_added": "54"
           },
           "edge": {
-            "version_added": "≤79"
+            "version_added": "80"
```
foolip, line 15: This is very surprising, can you confirm that in Edge 79 the `DocumentTimeline` constructor wasn't there, but `document.timeline` was? It just seems weird that this would have been disabled for just one release, when then change in Chrome happened way back in 54....

vinyldarkscratch: Apparently I initially made a typo, and meant to put "84" as the version.  However, upon digging deeper into the data, I found that the API wasn't around in Chrome 54 either, but rather, it was introduced in Chrome 84.  It appears we need to correct __all__ of our Chrome data accordingly -- I've opened #9999 (nice!) to fix all of the Chromium data (and supersede this PR).

PR #9984, opened 2021-04-20
Add opera support for :where()

lukewarlow:
Updated :where() support in Opera.

Opera Desktop v74 is based on Chromium 88 which is where support was added.


ddbeck (approved):
Looks good. Thank you! 🎉

PR #9986, opened 2021-04-20
Add opera support for aspect-ratio

lukewarlow:
Updated aspect-ratio support in Opera.

Opera Desktop v74 is based on Chromium 88 which is where support was added.


ddbeck (approved):
Thank you! 🎉

PR #9988, opened 2021-04-20
Fix data for badging API.

jpmedley:
I had a email exchange with one of the Chrome engineers who worked on this. 

>It shipped in 83. That was on Windows and Mac. It has never shipped on Linux, Android or Chrome OS (due to lack of support for badging for native apps on those platforms). It is coming to Chrome OS in 91.

### DIFF
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -2406,6 +2406,7 @@
             },
             "edge": {
               "version_added": "83",
+              "partial_implementation": true,
               "notes": "Windows and Mac only."
             },
             "firefox": {
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -386,10 +386,11 @@
         "__compat": {
           "support": {
             "chrome": {
-              "version_added": "81"
+              "version_added": "83",
+              "notes": "Windows and Mac only."
             },
             "chrome_android": {
-              "version_added": "81"
+              "version_added": false
             },
             "edge": {
               "version_added": "81"
@@ -419,7 +420,7 @@
               "version_added": "13.0"
             },
             "webview_android": {
-              "version_added": "81"
+              "version_added": false
             }
           },
           "status": {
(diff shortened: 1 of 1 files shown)


ddbeck:
Two questions on this:

1. Have you tried this? Is the API not exposed on the unsupported platforms? Or is it only non-functional?
2. Do you have any idea how this impacts the other Chromium browsers? The Edge, Opera Android, and Samsung Internet data looks a bit suspect, in light of this change.

ddbeck:
Hi @jpmedley, do you plan to come back to this PR?

jpmedley:
I have not tried it. I don't see the need to since the engineer who implemented told me it was not supported.

I don't have any idea how this impacts other Chromium browsers. If the choice is between having one value that's flat out wrong and one that's suspect, I choose suspect.

api/Navigator.json
```diff
@@ -2395,13 +2396,15 @@
         "__compat": {
           "support": {
             "chrome": {
-              "version_added": "81"
+              "version_added": "83",
```
ddbeck, line 2399:
```suggestion
              "version_added": "83",
              "partial_implementation": true,
```

api/Navigator.json
```diff
@@ -2395,13 +2396,15 @@
...
+              "notes": "Windows and Mac only."
             },
             "chrome_android": {
-              "version_added": "81"
+              "version_added": false
             },
             "edge": {
-              "version_added": "81"
+              "version_added": "83",
```
ddbeck, line 2406:
```suggestion
              "version_added": "83",
              "partial_implementation": true,
```

api/Navigator.json
```diff
@@ -386,10 +386,11 @@
         "__compat": {
           "support": {
             "chrome": {
-              "version_added": "81"
+              "version_added": "83",
```
ddbeck, line 389:
```suggestion
              "version_added": "83",
              "partial_implementation": true,
```

ddbeck (changes requested):
Thanks, Joe. A few of the entries need `"partial_implementation": true` per https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#operating-system-limitations-imply-partial_implementation. Then this will be ready.

jpmedley:
Done.

ddbeck (approved):
Thank you! 🎉

PR #1023, opened 2018-02-11
Adds compat data for DocumentTimeline

bennypowers:

PR #9999, opened 2021-04-21
Update Chromium versions for DocumentTimeline API

vinyldarkscratch:
This PR updates and corrects the real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `DocumentTimeline` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.1).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/DocumentTimeline

Additional context:
While updating Edge's data to remove ranged versions, I noticed an inconsistency between the Chrome data and Edge.  After further review, I found that the data for this API was inaccurate, and that the API wasn't introduced before Chrome 84.  The original data seems to have come from #1023, which was a part of the wiki migration.

My best guess is that when the data was first added, the user that added it had a flag enabled which enabled the feature.  Since we're already working to remove flag data, I didn't bother to try to track down the flag.

### DIFF
--- a/api/DocumentTimeline.json
+++ b/api/DocumentTimeline.json
@@ -6,13 +6,13 @@
         "spec_url": "https://drafts.csswg.org/web-animations/#the-documenttimeline-interface",
         "support": {
           "chrome": {
-            "version_added": "54"
+            "version_added": "84"
           },
           "chrome_android": {
-            "version_added": "54"
+            "version_added": "84"
           },
           "edge": {
-            "version_added": "≤79"
+            "version_added": "84"
           },
           "firefox": [
             {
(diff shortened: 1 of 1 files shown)


foolip:
9999!

foolip:
Who will get 10000?

foolip:
@vinyldarkscratch note a typo in the URL https://mdn-bcd-collector.appspot.com/tests/API/DocumentTimeline, if that's from a template you might want to update it.

foolip:
I've tested these in Chrome 83 and 84:
https://mdn-bcd-collector.appspot.com/tests/api/DocumentTimeline
https://mdn-bcd-collector.appspot.com/tests/api/Document/timeline

Both go from false to true, which makes much more sense than the previous data.

vinyldarkscratch:
Thanks for the catch!  Accidentally told the URL to use the "pretty category name" rather than the proper namespace.  😝

foolip (approved):
This now matches api.Document.timeline in https://github.com/mdn/browser-compat-data/blob/main/api/Document.json 👍

PR #10005, opened 2021-04-21
"transform-origin" is partially supported in Safari #9834

maxlk:
`transform-origin` SVG presentation attribute is only partially supported in Safari.
It doesn't properly work in desktop Safari versions 12.1.2, 13.1.2, 14.0.3, and iOS Safari 12, 13, 14 (didn't test in earlier versions).
`transform-origin` SVG attribute value is applied only to transformations defined in `transform` CSS property, but not in `transform` SVG attribute.
WebKit [bug 201854](https://bugs.webkit.org/show_bug.cgi?id=201854)

### DIFF
--- a/svg/attributes/presentation.json
+++ b/svg/attributes/presentation.json
@@ -2954,12 +2954,12 @@
               "safari": {
                 "version_added": true,
                 "partial_implementation": true,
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
In the hopes of pinning this to a version number, I tried the test case from the WebKit bug in Safaris before 12.1 and it broke even more spectacularly. I'm not sure if that's anything to do with this bug, however, and I'd rather get the note in without.

Thank you! 🎉

sideshowbarker:
@maxlk Really nice work on this — thanks much for having caught the problem and followed through on  getting it documented; I think this can help save other developers a lot of time and frustration

PR #10013, opened 2021-04-21
Add `::-webkit-details-marker`

### LINKED ISSUE #9869: Missing data: `::-webkit-details-marker`
According to this note, `::-webkit-details-marker` exists:

https://github.com/mdn/browser-compat-data/blob/cd6fbfee6af5265b60f950a56fb60a423b00493d/html/elements/summary.json#L58

But we don't actually have a feature entry for this. Rather than reporting on its status via a note on an HTML element, we ought to create a new feature (presumably `css.selectors.-webkit.details.marker`) and populate it with some useful data.

Inspired by #9809.

ddbeck:
This PR fixes #9869.

The values here are either from testing (or mirroring from test browsers), with the exception of one point: support for this appearing in Chrome 12. I couldn't actually test this that far back. From testing, I could only show ≤15, but based on Safari and the `<details>` element appearing in Chrome 12, I'd guess that this landed at some point between 12 and 15. I figured it didn't matter that much where specifically, since it's on its way out of Chrome.

### DIFF
--- a/css/selectors/-webkit-details-marker.json
+++ b/css/selectors/-webkit-details-marker.json
@@ -1,0 +1,59 @@
+{
+  "css": {
+    "selectors": {
+      "-webkit-details-marker": {
+        "__compat": {
+          "description": "<code>::-webkit-details-marker</code>",
+          "support": {
+            "chrome": {
+              "version_added": "12",
+              "version_removed": "89"
+            },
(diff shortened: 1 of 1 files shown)


vinyldarkscratch (approved):
LGTM, thanks!

PR #10070, opened 2021-04-24
Fix mobile Chromium data for Request API

vinyldarkscratch:
While working on #9916, a number of interesting issues were found with the Chrome Android (and related browsers).  This PR aims to correct them, based upon the results manually from the mdn-bcd-collector project.


ddbeck (approved):
Thank you, @vinyldarkscratch! 🎉

PR #10082, opened 2021-04-26
FF89 removal DeviceProximityEvent and Window.ondeviceproximityevent

hamishwillee:
- FF89 docs tracking: https://github.com/mdn/content/issues/4308
- Firefox bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1699707 

Essentially this event has been removed from firefox in FF89 (which was the last thing that supported it). Further, previously this was present behind a preference, so the Android removal is effectively FF79 - see https://github.com/mdn/browser-compat-data/issues/9679

For MDN I have removed the attribute docs (max, min, value) so these urls have also been removed from BCD

### DIFF
--- a/api/DeviceProximityEvent.json
+++ b/api/DeviceProximityEvent.json
@@ -41,7 +41,7 @@
                   "value_to_set": "true"
                 }
               ],
-              "version_removed": "89",
+              "version_removed": "79",
               "notes": "See <a href='https://bugzil.la/1462308'>bug 1462308</a>, <a href='https://bugzil.la/1699707'>bug 1699707</a>."
             },
             {
--- a/api/DeviceProximityEvent.json
+++ b/api/DeviceProximityEvent.json
@@ -23,7 +23,8 @@
                   "value_to_set": "true"
                 }
               ],
-              "notes": "See <a href='https://bugzil.la/1462308'>bug 1462308</a>."
+              "version_removed": "89",
+              "notes": "See <a href='https://bugzil.la/1462308'>bug 1462308</a>, <a href='https://bugzil.la/1699707'>bug 1699707</a>."
             },
             {
               "version_added": true,
@@ -40,7 +41,8 @@
                   "value_to_set": "true"
                 }
               ],
-              "notes": "See <a href='https://bugzil.la/1462308'>bug 1462308</a>."
+              "version_removed": "89",
+              "notes": "See <a href='https://bugzil.la/1462308'>bug 1462308</a>, <a href='https://bugzil.la/1699707'>bug 1699707</a>."
             },
             {
               "version_added": "15",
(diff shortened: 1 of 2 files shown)


api/Window.json
```diff
@@ -4625,12 +4625,42 @@
             "edge": {
               "version_added": false
             },
-            "firefox": {
-              "version_added": "15"
-            },
-            "firefox_android": {
-              "version_added": "15"
-            },
+            "firefox": [
```
hamishwillee, line 4628: FYI, this matches the windows event handler versioning to the event itself - ie puts it behind the preference. To me this seems like it was an omissions, since the handler isn't usable without the event.

api/Window.json
```diff
@@ -4625,12 +4625,42 @@
...
+                    "type": "preference",
+                    "name": "device.sensors.proximity.enabled",
+                    "value_to_set": "true"
+                  }
+                ],
+                "version_removed": "89",
+                "notes": "See <a href='https://bugzil.la/1462308'>bug 1462308</a>, <a href='https://bugzil.la/1699707'>bug 1699707</a>."
+              },
+              {
+                "version_added": true,
```
ddbeck, line 4642:
```suggestion
                "version_added": "15",
```

api/DeviceProximityEvent.json
```diff
@@ -23,7 +23,8 @@
                   "value_to_set": "true"
                 }
               ],
-              "notes": "See <a href='https://bugzil.la/1462308'>bug 1462308</a>."
+              "version_removed": "89",
+              "notes": "See <a href='https://bugzil.la/1462308'>bug 1462308</a>, <a href='https://bugzil.la/1699707'>bug 1699707</a>."
```
ddbeck, line 27: Same story as before, with the notes. Don't add a new link or drop them entirely. 🚮

api/DeviceProximityEvent.json
```diff
@@ -98,7 +99,8 @@
                     "value_to_set": "true"
                   }
                 ],
-                "notes": "See <a href='https://bugzil.la/1462308'>bug 1462308</a>."
+                "version_removed": "89",
+                "notes": "See <a href='https://bugzil.la/1462308'>bug 1462308</a>, <a href='https://bugzil.la/1699707'>bug 1699707</a>."
               },
               {
                 "version_added": true,
```
ddbeck, line 106:
```suggestion
                "version_added": "15",
```

api/DeviceProximityEvent.json
```diff
@@ -174,7 +176,8 @@
...
```
ddbeck, line 183:
```suggestion
                "version_added": "15",
```

api/DeviceProximityEvent.json
```diff
@@ -250,7 +253,8 @@
...
```
ddbeck, line 260:
```suggestion
                "version_added": "15",
```

ddbeck (changes requested):
Several suggestions here. Thank you, @hamishwillee!

api/DeviceProximityEvent.json
hamishwillee: All notes dropped. They aren't very useful.

hamishwillee:
@ddbeck Thanks for the advice. All makes lots of sense. Done.

ddbeck (approved):
Looks good. Thank you! 🎉

PR #10087, opened 2021-04-26
Update `html.elements.img.aspect_ratio_computed_from_attributes` to include information about Safari's implementation

sulliops:
Safari's implementation of `aspect_ratio_computed_from_attributes` currently does not work as intended with lazy-loaded images, per https://bugs.webkit.org/show_bug.cgi?id=224197. A fix has been committed and should be rolled out in a future release of Safari.

### DIFF
--- a/html/elements/img.json
+++ b/html/elements/img.json
@@ -187,12 +187,12 @@
               "safari": {
                 "version_added": "14",
                 "partial_implementation": true,
-                "notes": "Safari's implementation does not yet work properly with lazy-loaded images. See: https://bugs.webkit.org/show_bug.cgi?id=224197"
+                "notes": "Safari's implementation does not yet work properly with lazy-loaded images. See: https://webkit.org/b/224197"
               },
               "safari_ios": {
                 "version_added": "14",
                 "partial_implementation": true,
-                "notes": "Safari's implementation does not yet work properly with lazy-loaded images. See: https://bugs.webkit.org/show_bug.cgi?id=224197"
+                "notes": "Safari's implementation does not yet work properly with lazy-loaded images. See: https://webkit.org/b/224197"
               },
               "samsunginternet_android": {
                 "version_added": "12.0"
--- a/html/elements/img.json
+++ b/html/elements/img.json
@@ -185,10 +185,14 @@
                 "version_added": "57"
               },
               "safari": {
-                "version_added": "14"
+                "version_added": "14",
+                "partial_implementation": true,
+                "notes": "Safari's implementation does not yet work properly with lazy-loaded images. See: https://bugs.webkit.org/show_bug.cgi?id=224197"
               },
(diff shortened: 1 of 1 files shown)


html/elements/img.json
```diff
@@ -185,10 +185,14 @@
                 "version_added": "57"
               },
               "safari": {
-                "version_added": "14"
+                "version_added": "14",
+                "partial_implementation": true,
+                "notes": "Safari's implementation does not yet work properly with lazy-loaded images. See: https://webkit.org/b/224197"
```
ddbeck, line 190: I misunderstood this note at first to mean that it somehow prevents lazy loading. How about we revise it somewhat to make it clear that this affects layout. What do you think of this?

```suggestion
                "notes": "Safari doesn't preserve space for images without a valid <code>src</code>, which may disrupt layouts that rely on lazy loading (see <a href='https://webkit.org/b/224197'>bug 224197</a>)."
```

html/elements/img.json
```diff
@@ -185,10 +185,14 @@
...
               },
               "safari_ios": {
-                "version_added": "14"
+                "version_added": "14",
+                "partial_implementation": true,
+                "notes": "Safari's implementation does not yet work properly with lazy-loaded images. See: https://webkit.org/b/224197"
```
ddbeck, line 195:
```suggestion
                "notes": "Safari doesn't preserve space for images without a valid <code>src</code>, which may disrupt layouts that rely on lazy loading (see <a href='https://webkit.org/b/224197'>bug 224197</a>)."
```

ddbeck (changes requested):
Thanks for this PR, @sulliops. I see it's your first on the project—welcome!

I've got a suggested rewrite for the note and I'd like your opinion on it.

sulliops:
> I've got a suggested rewrite for the note and I'd like your opinion on it.

Good suggestion! No further changes on my end. I preemptively deleted my patch repo, but I believe you can commit your suggestions on your end.

By the way, how would I go about listing the change if the Safari `version_added` were `TP` or Technical Preview? On a different commit, the checks failed because `TP` is not recognized.

ddbeck:
@sulliops! Thank you. The missing fork made applying the changes a little weird (and it's possible I used the wrong command line incantation), but ee9cbf002d8d0c1cf4f5004484ba5e44bbe5ebfd applies your PR to the main branch. 🎉

> By the way, how would I go about listing the change if the Safari `version_added` were `TP` or Technical Preview? On a different commit, the checks failed because `TP` is not recognized.

We don't have a way to do this yet. We're trying to work out a method for several browsers at once (Safari TP, Chrome Canary, Firefox Nightly). In the mean time, feel free to open issues to track the things our schema doesn't yet represent (we can flag it as blocked until we have a solution, or the feature ships).

PR #10088, opened 2021-04-26
The image-set type function is in Firefox 89

rachelandrew:
Working on https://github.com/mdn/content/issues/4306

The `type()` function is supported in `image-set()` as from Firefox 89. Updated the note.

### DIFF
--- a/css/types/image.json
+++ b/css/types/image.json
@@ -2054,7 +2054,7 @@
               "firefox": {
                 "version_added": "88",
                 "notes": [
-                  "Prior to Firefox 89, the <code>type()</code> function was not supported as an argument to <code>image-set()</code>.",
+                  "Before Firefox 89, the <code>type()</code> function is not supported as an argument to <code>image-set()</code>.",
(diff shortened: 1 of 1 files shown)


css/types/image.json
```diff
@@ -2054,7 +2054,7 @@
               "firefox": {
                 "version_added": "88",
                 "notes": [
-                  "The <code>type()</code> function is not supported as an argument to <code>image-set()</code>. See <a href='https://bugzil.la/1695404'>bug 1695404</a>.",
+                  "Prior to Firefox 89, the <code>type()</code> function was not supported as an argument to <code>image-set()</code>.",
```
ddbeck, line 2057: A couple style suggestions (I _still_ need to properly document them somewhere):

```suggestion
                  "Before Firefox 89, the <code>type()</code> function is not supported as an argument to <code>image-set()</code>.",
```

Also, can we put this note after the next one, since the next one is now applicable to more versions?

ddbeck (changes requested):
A couple of little suggestions here. Thank you!

ddbeck (approved):
Thank you! 🎉

PR #10099, opened 2021-04-26
Fix Samsung Internet release dates

atjn:
The release date of Samsung Internet 14.0 and 13.2 are wrong. At first, I thought they were accidentally set to `2020` instead of `2021`, but the months also seem to be wrong. The new dates are set based on the Google Play release date and the date on [Samsung's own website](https://developer.samsung.com/internet/release-note.html).

CC @AdaRoseCannon (who added these dates in #9868), just in case you know something I don't :)

### DIFF
--- a/browsers/samsunginternet_android.json
+++ b/browsers/samsunginternet_android.json
@@ -184,7 +184,7 @@
           "engine_version": "83"
         },
         "13.2": {
-          "release_date": "2020-11-26",
+          "release_date": "2021-01-20",
           "status": "retired",
           "engine": "Blink",
           "engine_version": "83"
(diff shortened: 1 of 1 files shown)


AdaRoseCannon:
Those were the dates the Beta version had those versions available.

I don't mind one way or the other. So if you want to standardise on the Stable release that is fine too.

atjn:
> Those were the dates the Beta version had those versions available.

Oh, that makes perfect sense. I probably should've specified that the real problem is that, according to the dataset right now, v14.0 was released before v11.2, and v13.2 was released before v13.0. That seems to be because the other dates are based on the stable release date, so I think it makes sense to base these on it as well.

AdaRoseCannon:
That works for me I am okay with this being merged.

ddbeck (approved):
Thank you, @atjn and @AdaRoseCannon! 🎉

PR #10103, opened 2021-04-27
FF89 removal UserProximityEvent and Window.onuserproximity

hamishwillee:
This API was previously under a preference and was supposed to be removed (along with `DeviceProximityEvent`) in FF89 - see https://bugzilla.mozilla.org/show_bug.cgi?id=1699707#c15. Prior to that it was behind a preference, which effectively means that it was removed in FF79 on Android (see #9679).

However while `DeviceProximityEvent` was actually removed this was not, because it is still useful for KaiOS (only). However as per the [comment](https://bugzilla.mozilla.org/show_bug.cgi?id=1699707#c15) the INTENT is that this is off for Android and that a user will not turn it on.

So what I have done is make this match `DeviceProximityEvent` - it is shown as being removed in v89 for desktop and v79 for android. I also "rationalised" the windows event handler to match the versionining, because without the pref set you won't get an event. 

I do not mention KaiOS because we have no Firefox for KaiOS option.

Docs tracking for this is in https://github.com/mdn/content/issues/4308

### DIFF
--- a/api/UserProximityEvent.json
+++ b/api/UserProximityEvent.json
@@ -23,11 +23,10 @@
                   "name": "device.sensors.proximity.enabled",
                   "value_to_set": "true"
                 }
-              ],
-              "notes": "See <a href='https://bugzil.la/1462308'>bug 1462308</a>, <a href='https://bugzil.la/1699707'>bug 1699707</a>."
+              ]
             },
             {
-              "version_added": true,
+              "version_added": "15",
               "version_removed": "61"
             }
           ],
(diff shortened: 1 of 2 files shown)


ddbeck (changes requested):
A couple of suggestions for you, @hamishwillee. Thanks!

hamishwillee:
Thanks @ddbeck - fixed as per your suggestions/explanations. All makes sense.

Gulyas89 (approved):

ddbeck (approved):
Thank you! 🎉

PR #10104, opened 2021-04-27
FF89 removal DeviceLightEvent and Window.ondevicelight

hamishwillee:
This is just like #10103 (removal of UserProximityEvent). As discussed in https://bugzilla.mozilla.org/show_bug.cgi?id=1701824#c5 the API is intended to be removed in FF89. There has been a preference and there is still a preference, however the intent is that this not be re-enabled on android. It is ONLY present because KaiOS still uses this.

I have documented so that android is shown as removing support in FF79 (https://github.com/mdn/browser-compat-data/issues/9679) and desktop goes in FF89. I don't talk about KaiOS. The window versions have been matched to the event versions for firefox. This makes sense because if the event can't fire unless the preference is enabled, then IMO the API has the same versioning as the event even if it was not explicitly disabled in the same versions.

I also removed the docs links for the sub-feature, as will be removing that file from MDN.

Docs tracking for this is https://github.com/mdn/content/issues/4308

### DIFF
--- a/api/DeviceLightEvent.json
+++ b/api/DeviceLightEvent.json
@@ -24,8 +24,7 @@
                   "name": "device.sensors.ambientLight.enabled",
                   "value_to_set": "true"
                 }
-              ],
-              "notes": "See <a href='https://bugzil.la/1462308'>bug 1462308</a>, <a href='https://bugzil.la/1701824'>bug 1701824</a>."
+              ]
             },
             {
               "version_added": "22",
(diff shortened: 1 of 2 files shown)


api/DeviceLightEvent.json
```diff
@@ -17,14 +17,15 @@
...
+              "version_removed": "89",
               "flags": [
                 {
                   "type": "preference",
                   "name": "device.sensors.ambientLight.enabled",
                   "value_to_set": "true"
                 }
               ],
-              "notes": "See <a href='https://bugzil.la/1462308'>bug 1462308</a>."
+              "notes": "See <a href='https://bugzil.la/1462308'>bug 1462308</a>, <a href='https://bugzil.la/1701824'>bug 1701824</a>."
```
ddbeck, line 28: Like the other PR, I'd advise not including the new link or dropping this note outright.

ddbeck (changes requested):
My review here is a bit like the other PR. Thanks, @hamishwillee!

hamishwillee:
Thanks! Fixed as you suggested.

PR #10105, opened 2021-04-27
Add safari support for flex gap

manuelmeister:
Added safari version with flex gap support. I tested it using the following Testcase on Safari 14.1 on macOS 10.15.7 and on iOS 14.5.

Testcase: https://codepen.io/manuelmeister/pen/zYNbVGK?editors=1100
Release Notes: https://developer.apple.com/documentation/safari-release-notes/safari-14_1-release_notes
TP Release Notes: https://developer.apple.com/safari/technology-preview/release-notes/#r115
Bug: https://bugs.webkit.org/show_bug.cgi?id=206767

### DIFF
--- a/css/properties/column-gap.json
+++ b/css/properties/column-gap.json
@@ -37,10 +37,10 @@
                 "version_added": false
               },
               "safari": {
-                "version_added": false
+                "version_added": "14.1"
               },
               "safari_ios": {
-                "version_added": false
+                "version_added": "14.5"
               },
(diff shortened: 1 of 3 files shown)


ddbeck:
Thanks for opening this, @manuelmeister. I expect that #10129, which includes this change plus many others, will be merged shortly. I'm going to close this PR, but I wanted to thank you for the independent validation. 👍

PR #10121, opened 2021-04-27
Safari class support update

robpalme:
This updates the public/private field support for class in latest Safari and Safari on iOS.

All sources are linked here:  https://github.com/Fyrd/caniuse/pull/5858

PR #10129, opened 2021-04-27
Updates for Safari 14.1 on macOS and Safari on iOS 14.5

apple-web-evangelist:
This is a large update of features supported with the release of Safari 14.1 on macOS and Safari on iOS 14.5. The changes have been compiled using mdn-results-collector, a thorough review of changes by the WebKit engineering team at Apple, manual tests to verify results, and engineer submissions.

### DIFF
--- a/api/AudioParam.json
+++ b/api/AudioParam.json
@@ -80,7 +80,7 @@
               "version_added": false
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "14.5"
             },
             "samsunginternet_android": {
               "version_added": "10.0"
@@ -126,10 +126,10 @@
               "version_added": "43"
             },
             "safari": {
-              "version_added": false
+              "version_added": "14.1"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "14.5"
             },
             "samsunginternet_android": {
               "version_added": "7.0"
--- a/api/Blob.json
+++ b/api/Blob.json
@@ -134,7 +134,7 @@
               "version_added": "14"
             },
             "safari_ios": {
-              "version_added": "14"
+              "version_added": "14.5"
             },
             "samsunginternet_android": {
               "version_added": "12.0"
@@ -310,10 +310,10 @@
               "version_added": "54"
             },
             "safari": {
-              "version_added": false
+              "version_added": "14.1"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "14.5"
             },
             "samsunginternet_android": {
               "version_added": "12.0"
@@ -363,7 +363,7 @@
               "version_added": false
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "14.5"
             },
             "samsunginternet_android": {
               "version_added": "12.0"
--- a/api/HTMLAreaElement.json
+++ b/api/HTMLAreaElement.json
@@ -317,10 +317,10 @@
               "version_added": "41"
             },
             "safari": {
-              "version_added": false
+              "version_added": "14.1"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "14.5"
             },
             "samsunginternet_android": {
               "version_added": "7.2"
--- a/api/KHR_parallel_shader_compile.json
+++ b/api/KHR_parallel_shader_compile.json
@@ -29,10 +29,10 @@
             "version_added": null
           },
           "safari": {
-            "version_added": false
+            "version_added": "14.1"
           },
           "safari_ios": {
-            "version_added": false
+            "version_added": "14.5"
           },
           "samsunginternet_android": {
             "version_added": "12.0"
--- a/api/MediaMetadata.json
+++ b/api/MediaMetadata.json
@@ -44,10 +44,10 @@
             "version_added": false
           },
           "safari": {
-            "version_added": false
+            "version_added": "14.1"
           },
           "safari_ios": {
-            "version_added": false
+            "version_added": "14.5"
           },
           "samsunginternet_android": {
             "version_added": "7.0"
@@ -93,10 +93,10 @@
               "version_added": false
             },
             "safari": {
-              "version_added": false
+              "version_added": "14.1"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "14.5"
             },
             "samsunginternet_android": {
               "version_added": "7.0"
@@ -142,10 +142,10 @@
               "version_added": false
             },
             "safari": {
-              "version_added": false
+              "version_added": "14.1"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "14.5"
             },
             "samsunginternet_android": {
               "version_added": "7.0"
@@ -191,10 +191,10 @@
               "version_added": false
             },
             "safari": {
-              "version_added": false
+              "version_added": "14.1"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "14.5"
             },
             "samsunginternet_android": {
               "version_added": "7.0"
@@ -240,10 +240,10 @@
               "version_added": false
             },
             "safari": {
-              "version_added": false
+              "version_added": "14.1"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "14.5"
             },
             "samsunginternet_android": {
               "version_added": "7.0"
@@ -289,10 +289,10 @@
               "version_added": false
             },
             "safari": {
-              "version_added": false
+              "version_added": "14.1"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "14.5"
             },
             "samsunginternet_android": {
               "version_added": "7.0"
--- a/api/MediaRecorder.json
+++ b/api/MediaRecorder.json
@@ -31,12 +31,26 @@
           "opera_android": {
             "version_added": "36"
           },
-          "safari": {
-            "version_added": "14"
-          },
-          "safari_ios": {
-            "version_added": "14"
-          },
+          "safari": [
+            {
+              "version_added": "14.1"
+            },
+            {
+              "version_added": "14.0.2",
+              "partial_implementation": true,
+              "notes": "Does not include support of pause and resume."
+            }
+          ],
+          "safari_ios": [
+            {
+              "version_added": "14.5"
+            },
+            {
+              "version_added": "14.3",
+              "partial_implementation": true,
+              "notes": "Does not include support of pause and resume."
+            }
+          ],
           "samsunginternet_android": {
             "version_added": "5.0"
           },
@@ -81,10 +95,10 @@
               "version_added": "36"
             },
             "safari": {
-              "version_added": "14"
+              "version_added": "14.0.2"
             },
             "safari_ios": {
-              "version_added": "14"
+              "version_added": "14.3"
             },
             "samsunginternet_android": {
               "version_added": "5.0"
@@ -128,10 +142,10 @@
                 "version_added": "36"
               },
               "safari": {
-                "version_added": "14"
+                "version_added": "14.0.2"
               },
               "safari_ios": {
-                "version_added": "14"
+                "version_added": "14.3"
               },
               "samsunginternet_android": {
                 "version_added": "5.0"
@@ -178,10 +192,10 @@
               "version_added": "36"
             },
             "safari": {
-              "version_added": false
+              "version_added": "14.0.2"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "14.3"
             },
             "samsunginternet_android": {
               "version_added": "5.0"
@@ -228,10 +242,10 @@
               "version_added": "36"
             },
             "safari": {
-              "version_added": "14"
+              "version_added": "14.0.2"
             },
             "safari_ios": {
-              "version_added": "14"
+              "version_added": "14.3"
             },
             "samsunginternet_android": {
               "version_added": "5.0"
@@ -331,10 +345,10 @@
               "version_added": "36"
             },
             "safari": {
-              "version_added": "14"
+              "version_added": "14.0.2"
             },
             "safari_ios": {
-              "version_added": "14"
+              "version_added": "14.3"
             },
             "samsunginternet_android": {
               "version_added": "5.0"
@@ -397,10 +411,10 @@
               "version_added": "36"
             },
             "safari": {
-              "version_added": "14"
+              "version_added": "14.0.2"
             },
             "safari_ios": {
-              "version_added": "14"
+              "version_added": "14.3"
             },
             "samsunginternet_android": {
               "version_added": "5.0"
@@ -454,10 +468,10 @@
               "version_added": "36"
             },
             "safari": {
-              "version_added": "14"
+              "version_added": "14.0.2"
             },
             "safari_ios": {
-              "version_added": "14"
+              "version_added": "14.3"
             },
             "samsunginternet_android": {
               "version_added": "5.0"
@@ -503,10 +517,10 @@
               "version_added": "36"
             },
             "safari": {
-              "version_added": "14"
+              "version_added": "14.0.2"
             },
             "safari_ios": {
-              "version_added": "14"
+              "version_added": "14.3"
             },
             "samsunginternet_android": {
               "version_added": "5.0"
@@ -650,10 +664,10 @@
               "version_added": "36"
             },
             "safari": {
-              "version_added": "14"
+              "version_added": "14.0.2"
             },
             "safari_ios": {
-              "version_added": "14"
+              "version_added": "14.3"
             },
             "samsunginternet_android": {
               "version_added": "5.0"
@@ -699,10 +713,10 @@
               "version_added": "36"
             },
             "safari": {
-              "version_added": "14"
+              "version_added": "14.0.2"
             },
             "safari_ios": {
-              "version_added": "14"
+              "version_added": "14.3"
             },
             "samsunginternet_android": {
               "version_added": "5.0"
@@ -797,10 +811,10 @@
               "version_added": "36"
             },
             "safari": {
-              "version_added": false
+              "version_added": "14.1"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "14.5"
             },
             "samsunginternet_android": {
               "version_added": "5.0"
@@ -846,10 +860,10 @@
               "version_added": "36"
             },
             "safari": {
-              "version_added": "14"
+              "version_added": "14.0.2"
             },
             "safari_ios": {
-              "version_added": "14"
+              "version_added": "14.3"
             },
             "samsunginternet_android": {
               "version_added": "5.0"
@@ -895,10 +909,10 @@
               "version_added": "36"
             },
             "safari": {
-              "version_added": false
+              "version_added": "14.1"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "14.5"
             },
             "samsunginternet_android": {
               "version_added": "5.0"
@@ -944,10 +958,10 @@
               "version_added": "36"
             },
             "safari": {
-              "version_added": "14"
+              "version_added": "14.0.2"
             },
             "safari_ios": {
-              "version_added": "14"
+              "version_added": "14.3"
             },
             "samsunginternet_android": {
               "version_added": "5.0"
@@ -1009,10 +1023,10 @@
               "version_added": "36"
             },
             "safari": {
-              "version_added": "14"
+              "version_added": "14.0.2"
             },
             "safari_ios": {
-              "version_added": "14"
+              "version_added": "14.3"
             },
             "samsunginternet_android": {
               "version_added": "5.0"
@@ -1066,10 +1080,10 @@
               "version_added": "36"
             },
             "safari": {
-              "version_added": "14"
+              "version_added": "14.0.2"
             },
             "safari_ios": {
-              "version_added": "14"
+              "version_added": "14.3"
             },
             "samsunginternet_android": {
               "version_added": "5.0"
@@ -1131,10 +1145,10 @@
               "version_added": "36"
             },
             "safari": {
-              "version_added": "14"
+              "version_added": "14.0.2"
             },
             "safari_ios": {
-              "version_added": "14"
+              "version_added": "14.3"
             },
             "samsunginternet_android": {
               "version_added": "5.0"
@@ -1180,10 +1194,10 @@
               "version_added": "36"
             },
             "safari": {
-              "version_added": false
+              "version_added": "14.1"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "14.5"
             },
             "samsunginternet_android": {
               "version_added": "5.0"
--- a/api/OfflineAudioContext.json
+++ b/api/OfflineAudioContext.json
@@ -56,14 +56,26 @@
               "prefix": "webkit"
             }
           ],
-          "safari": {
-            "version_added": "6",
-            "prefix": "webkit"
-          },
-          "safari_ios": {
-            "version_added": "6",
-            "prefix": "webkit"
-          },
+          "safari": [
+            {
+              "version_added": "14.1"
+            },
+            {
+              "version_added": "6",
+              "version_removed": "14.1",
+              "prefix": "webkit"
+            }
+          ],
+          "safari_ios": [
+            {
+              "version_added": "14.5"
+            },
+            {
+              "version_added": "6",
+              "version_removed": "14.5",
+              "prefix": "webkit"
+            }
+          ],
           "samsunginternet_android": [
             {
               "version_added": "3.0"
@@ -148,14 +160,26 @@
                 "prefix": "webkit"
               }
             ],
-            "safari": {
-              "version_added": "6",
-              "prefix": "webkit"
-            },
-            "safari_ios": {
-              "version_added": "6",
-              "prefix": "webkit"
-            },
+            "safari": [
+              {
+                "version_added": "14.1"
+              },
+              {
+                "version_added": "6",
+                "version_removed": "14.1",
+                "prefix": "webkit"
+              }
+            ],
+            "safari_ios": [
+              {
+                "version_added": "14.5"
+              },
+              {
+                "version_added": "6",
+                "version_removed": "14.5",
+                "prefix": "webkit"
+              }
+            ],
             "samsunginternet_android": [
               {
                 "version_added": "3.0"
@@ -212,10 +236,10 @@
                 "version_added": "46"
               },
               "safari": {
-                "version_added": false
+                "version_added": "14.1"
               },
               "safari_ios": {
-                "version_added": false
+                "version_added": "14.5"
               },
               "samsunginternet_android": {
                 "version_added": "8.0"
@@ -310,10 +334,10 @@
               "version_added": "41"
             },
             "safari": {
-              "version_added": false
+              "version_added": "14.1"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "14.5"
             },
             "samsunginternet_android": {
               "version_added": "5.0"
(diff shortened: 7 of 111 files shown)


KenG-pro (approved):

css/properties/gap.json
```diff
@@ -33,10 +33,10 @@
                 "version_added": false
               },
               "safari": {
-                "version_added": false
+                "version_added": "14.1"
```
foolip, line 36: Yes! Thanks to @svillar for [closing the gap in flexbox](https://blogs.igalia.com/svillar/2020/10/01/closing-the-gap-in-flexbox/) 🎉

api/OfflineAudioContext.json
```diff
@@ -453,12 +477,26 @@
...
+                "version_removed": "14.1",
+                "prefix": "webkit"
+              }
+            ],
+            "safari_ios": [
+              {
+                "version_added": "14.5"
+              },
+              {
+                "version_added": "≤10.3",
```
foolip, line 495:
```suggestion
                "version_added": "6",
```

Confirming the data for `OfflineAudioContext` proved difficult, but this is consistent with the current claim that `webkitOfflineAudioContext` was introduced in iOS 6. Using http://mdn-bcd-collector.appspot.com/tests/api/webkitOfflineAudioContext I'm actually able to confirm that *wasn't* the case, and that it must have shipped either in 6.1 / 7.0 or 6.2 / 7.1. Unfortunately I don't have access to 6.1 / 7.0 and this runs into the complications in https://github.com/mdn/browser-compat-data/issues/9423, but just matching the base feature seems OK here. If that data is updated later, this lint would fail and this would have to change too.

browsers/safari.json
```diff
@@ -173,6 +173,19 @@
           "status": "current",
           "engine": "WebKit",
           "engine_version": "610.1.28"
+        },
+        "14.0.2": {
```
foolip, line 177: See https://github.com/mdn/browser-compat-data/pull/8838 for a previous discussion on how to handle `MediaRecorder` and iOS 14.3. The conclusion then was to treat it as a minor version and live with a little bit of inaccuracy in the data.

But if the Safari teams considers these versions significant and would like to add them, that's a decision that should be revisited. cc @ddbeck @Elchi3

foolip (commented):
I've reviewed some of the trickier bits, I'll come back to review all the other changes after upgrading to macOS 11.3 :)

api/PannerNode.json
```diff
@@ -28,9 +28,16 @@
...
-          "safari": {
-            "version_added": "6"
-          },
+          "safari": [
+            {
+              "version_added": "14.1"
+            },
+            {
+              "version_added": "6",
+              "version_removed": "14.1",
```
foolip, line 37: Confirmed with https://mdn-bcd-collector.appspot.com/tests/api/webkitAudioPannerNode

api/PannerNode.json
```diff
@@ -28,9 +28,16 @@
...
+              "prefix": "webkit"
+            }
+          ],
           "safari_ios": {
             "version_added": "6"
```
foolip, line 42: The same changes as above are needed here.

api/PannerNode.json
```diff
@@ -79,7 +86,7 @@
               "version_added": "42"
             },
             "safari": {
-              "version_added": false
+              "version_added": "14.1"
             },
             "safari_ios": {
               "version_added": false
```
foolip, line 92:
```suggestion
              "version_added": "14.5"
```

Confirmed on iOS 14.5.

api/PerformanceNavigation.json
```diff
@@ -29,10 +29,10 @@
             "version_added": "14"
           },
           "safari": {
-            "version_added": false
+            "version_added": "11"
```
foolip, line 32:
```suggestion
            "version_added": "8"
```

I assume this change was needed because the lint would otherwise fail, but it looks like `performance.navigation` and all members except `toJSON` were supported since Safari 8, confirmed by testing 7.1 and 8 on https://mdn-bcd-collector.appspot.com/tests/api/PerformanceNavigation.

This matches the data in https://developer.mozilla.org/en-US/docs/Web/API/Performance/navigation#browser_compatibility.

It's not necessary for this PR, but it would be nice to also update

foolip, line 32:
```suggestion
            "version_added": "8"
```

I assume this change was needed because the lint would otherwise fail, but it looks like `performance.navigation` and all members except `toJSON` were supported since Safari 8, confirmed by testing 7.1 and 8 on https://mdn-bcd-collector.appspot.com/tests/api/PerformanceNavigation.

This matches the data in https://developer.mozilla.org/en-US/docs/Web/API/Performance/navigation#browser_compatibility.

It's not necessary for this PR, but it would be nice to also update the `redirectCount` and `type` entries to "8".

foolip:
Oops, I submitted that review too early by accident. The rest is coming in a second batch :)

api/PerformanceNavigation.json
```diff
@@ -29,10 +29,10 @@
...
           },
           "safari_ios": {
-            "version_added": false
+            "version_added": "11"
```
foolip, line 35:
```suggestion
            "version_added": "9"
```

This matches https://developer.mozilla.org/en-US/docs/Web/API/Performance/navigation#browser_compatibility and it looks like that data is correct, it shouldn't be iOS 8. Testing on iOS 8 and 9 on BrowserStack.

The same change can be made for `redirectCount` and `type`.

api/RTCPeerConnection.json
```diff
@@ -2715,7 +2715,7 @@
               "version_added": false
```
foolip, line 2715:
```suggestion
              "version_added": "14.1"
```

I can confirm support using https://mdn-bcd-collector.appspot.com/tests/api/RTCPeerConnection/onicecandidateerror in Safari 14.1 and iOS 14.5.

It was added in https://trac.webkit.org/changeset/270101/webkit which is so recent that I'm not going to check further back.

api/RTCPeerConnectionIceErrorEvent.json
```diff
@@ -32,7 +32,7 @@
             "version_added": false
```
foolip, line 32:
```suggestion
            "version_added": "14.1"
```

Confirmed with https://mdn-bcd-collector.appspot.com/tests/api/RTCPeerConnectionIceErrorEvent, and this was also part of https://trac.webkit.org/changeset/270101/webkit.

api/RTCPeerConnectionIceErrorEvent.json
```diff
@@ -79,7 +79,7 @@
               "version_added": false
```
foolip, line 79:
```suggestion
              "version_added": "14.1"
```

api/RTCPeerConnectionIceErrorEvent.json
```diff
@@ -127,7 +127,7 @@
...
```
foolip, line 127:
```suggestion
              "version_added": "14.1"
```

api/RTCPeerConnectionIceErrorEvent.json
```diff
@@ -223,7 +223,7 @@
...
```
foolip, line 223:
```suggestion
              "version_added": "14.1"
```

api/SpeechRecognition.json
```diff
@@ -35,10 +35,12 @@
             "version_added": false
           },
           "safari": {
-            "version_added": false
+            "prefix": "webkit",
```
foolip, line 38: Prefixed-only support confirmed by https://mdn-bcd-collector.appspot.com/tests/api/webkitSpeechRecognition + https://mdn-bcd-collector.appspot.com/tests/api/SpeechRecognition.

The latter test didn't try using `webkitSpeechRecognition` if present, so I've fixed that in https://github.com/foolip/mdn-bcd-collector/pull/1130 and can confirm support for all members except `grammars` and `serviceURI`, which seems right.

api/StereoPannerNode.json
```diff
@@ -29,7 +29,7 @@
             "version_added": "28"
           },
           "safari": {
-            "version_added": false
+            "version_added": "14.1"
           },
           "safari_ios": {
             "version_added": false
```
foolip, line 35:
```suggestion
            "version_added": "14.5"
```

Manually confirmed.

api/StereoPannerNode.json
```diff
@@ -79,7 +79,7 @@
               "version_added": "42"
             },
             "safari": {
-              "version_added": false
+              "version_added": "14.1"
             },
             "safari_ios": {
               "version_added": false
```
foolip, line 85:
```suggestion
              "version_added": "14.5"
```

api/StereoPannerNode.json
```diff
@@ -129,7 +129,7 @@
               "version_added": "28"
             },
             "safari": {
-              "version_added": false
+              "version_added": "14.1"
             },
             "safari_ios": {
               "version_added": false
```
foolip, line 135:
```suggestion
              "version_added": "14.5"
```

api/VTTCue.json
```diff
@@ -413,10 +413,10 @@
               "version_added": false
             },
             "safari": {
-              "version_added": false
+              "version_added": "14.1"
```
foolip, line 416: This is a little surprising since `VTTRegion` shipped much earlier, but seems correct, it's a recent fix by @frankolivier in https://trac.webkit.org/changeset/270738/webkit.

api/_mixins/ParentNode__Document.json
```diff
@@ -431,7 +431,7 @@
               "version_added": "14"
             },
             "safari_ios": {
-              "version_added": "14"
+              "version_added": "14.5"
```
foolip, line 434:
```suggestion
              "version_added": "14"
```

The original data here seems correct, I can confirm support using https://mdn-bcd-collector.appspot.com/tests/api/Document/replaceChildren in Safari 14.0.1 on iOS 14.2.1 on BrowserStack.

Is there a bug in a script that might have caused this?

api/_mixins/ParentNode__DocumentFragment.json
```diff
@@ -431,7 +431,7 @@
               "version_added": "14"
             },
             "safari_ios": {
-              "version_added": "14"
+              "version_added": "14.5"
```
foolip, line 434:
```suggestion
              "version_added": "14"
```

As above, but tested with https://mdn-bcd-collector.appspot.com/tests/api/DocumentFragment/replaceChildren

api/_mixins/ParentNode__Element.json
```diff
@@ -438,7 +438,7 @@
               "version_added": "14"
             },
             "safari_ios": {
-              "version_added": "14"
+              "version_added": "14.5"
```
foolip, line 441:
```suggestion
              "version_added": "14"
```

As above, but tested with https://mdn-bcd-collector.appspot.com/tests/api/Element/replaceChildren

foolip (commented):
OK, I've gotten through all of api/ now. Taking a break, because this was very time consuming :)

I anyone else wants to review css/ and later in detail, that would be great.

foolip:
OK, with this many review comments I realize it's going to be hard to make sense of. To summarize the state of the API data, there are a bunch of minor errors that I've suggested fixes for, and one interesting decision to be made about whether to have "14.0.2" and "14.3" as their own versions.

@apple-web-evangelist some of my comments are just confirming support or noting something interesting. Feel free to resolve those when you've read them to get them out of the way. Also note that GitHub will unhelpfully hide comments when there are lots of them, so somewhere in the middle there will be "Load more..." links to click.

ddbeck:
> I anyone else wants to review css/ and later in detail, that would be great.

@foolip Thank you so much for taking on a big chunk of the work here. Makes this PR feel a lot smaller. 😄 I'll take a look at the CSS (and the interesting decision stuff) more closely later today. 👍

json-derulo:
`Intl.DisplayNames` was also added with Safari 14.1. I've created a separate PR for this (#10130), but maybe it makes sense to add it to this PR instead.

jonathandavis:
This is all great feedback. I'm working through incorporating the suggested changes based on your findings @foolip. Thanks for the additional verification.

@json-derulo Also a good catch! I will incorporate that change as part of this PR as well.

api/MediaMetadata.json
```diff
@@ -44,10 +44,10 @@
             "version_added": false
           },
           "safari": {
-            "version_added": false
+            "version_added": "14.1"
```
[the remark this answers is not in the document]
jonathandavis: We will track this internally to ensure it is updated.

robpalme:
This looks good.  I had earlier created [a PR for JavaScript classes](https://github.com/mdn/browser-compat-data/pull/10121) which also included updates to "static" and "static fields" sub-features which are currently missing from this PR.

jonathandavis:
It also appears that updates for `Intl.NumberFormat` and `Intl.DateTimeFormat` were missed, along with new support for `Intl.ListFormat` and `Intl.Segmenter`. 

There appears no existing data file tracking `Intl.Segmenter`.

apple-web-evangelist:
> This is all great feedback. I'm working through incorporating the suggested changes based on your findings @foolip. Thanks for the additional verification.
> 
> @json-derulo Also a good catch! I will incorporate that change as part of this PR as well.

In addition, there is also data missing for `Intl.DateTimeFormat` and `Intl.NumberFormat` updates. Support for `Intl.ListFormat` and `Intl.Segmenter` are also available in these releases.

In the case of `Intl.Segmenter` there is a page, but currently no data file. Is there existing/formal process for requesting or generating a new file?

css/properties/scroll-margin-bottom.json
```diff
@@ -30,18 +30,28 @@
...
-            },
+            "safari": [
+              {
+                "version_added": "14.1"
+              },
+              {
+                "version_added": "11",
+                "alternative_name": "scroll-snap-margin-bottom",
+                "partial_implementation": true,
+                "notes": "Scroll margin is not applied for scrolls to fragment target or <code>scrollIntoView()</code>, see <a href='https://webkit.org/b/189265'>bug 189265</a>."
```
ddbeck, line 41: It's not strictly required to be addressed, but this data raises some interesting questions:

Does the bug noted apply to standard name? Or are they separate implementations?

The linked bug is fixed, so the answers here might help decide the disposition of the note and partial implementation, depending on whether (and, if so, when) the fixed shipped.

(It looks like this applies to all the `scroll-margin-*` features.)

api/MediaMetadata.json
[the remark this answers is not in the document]
apple-web-evangelist: We will track this internally to ensure it is updated.

javascript/classes.json
```diff
@@ -585,16 +585,26 @@
...
-              "version_added": "14",
-              "partial_implementation": true,
-              "notes": "Doesn't support public static fields. See WebKit bug <a href='https://webkit.org/b/194095'>194095</a>."
-            },
+            "safari": [
+              {
+                "version_added": "14.1"
+              },
+              {
+                "version_added": "14",
```
ddbeck, line 593:
```suggestion
                "version_added": "14",
                "version_removed": "14.1"
```

javascript/classes.json
```diff
@@ -585,16 +585,26 @@
...
+                "partial_implementation": true,
+                "notes": "Doesn't support public static fields. See WebKit bug <a href='https://webkit.org/b/194095'>194095</a>."
+              }
+            ],
+            "safari_ios": [
+              {
+                "version_added": "14.5"
+              },
+              {
+                "version_added": "14",
```
ddbeck, line 603:
```suggestion
                "version_added": "14",
                "version_removed": "14.1"
```

webextensions/manifest/background.json
```diff
@@ -79,8 +79,7 @@
                 "version_added": true
               },
               "safari": {
-                "notes": "Only persistent pages are supported.",
-                "partial_implementation": true,
+                "notes": "Support for non-persistent pages was added in 14.1.",
```
ddbeck, line 82: It doesn't always happen, but I strive to have notes framed such that the note is about the exceptional or historic case, not the general case, so notes about historic behavior recede from importance.

```suggestion
                "notes": "Before 14.1, only persistent pages are supported",
```

browsers/safari.json
apple-web-evangelist: We are opting for fudging this data and using "14" in cases where `14.0.2` or `safari_ios` shows `14.3`.

ddbeck (changes requested):

api/HTMLAreaElement.json
```diff
@@ -317,10 +317,10 @@
               "version_added": "41"
             },
             "safari": {
-              "version_added": false
+              "version_added": "14.1"
```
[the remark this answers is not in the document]
apple-web-evangelist: We will need to address the HTML/SVG elements "referrerpolicy" entries in a follow-up. We will track this internally to ensure we follow-up here.

apple-web-evangelist:
Performing another pass to address @ddbeck's comments.

api/MediaMetadata.json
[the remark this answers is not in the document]
apple-web-evangelist: I will submit another update to revert these entries to "false" for now. Our internal tracking will update this when appropriate.

lukewarlow:
With regards to the new date/time input support. You don't appear to have changed the time input to show it's supported as of Safari 14.1.

apple-web-evangelist:
A follow-up commit will include it. Thank you for pointing it out that it didn't make it into our the data.

css/properties/aspect-ratio.json
```diff
@@ -69,7 +69,7 @@
               "version_added": false
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "14.5"
```
lukewarlow, line 72: The aspect ratio experimental feature is disabled for me on both macOS and iOS. Is this not the case for others?

foolip: https://mdn-bcd-collector.appspot.com/tests/css/properties/aspect-ratio shows support for me on iOS 14.5, but feature detection can be broken.

Some of the basic test cases in https://wpt.live/css/css-sizing/aspect-ratio/ aren't passing, so I think most likely the property is recognized but the actual behavior is gated by a flag.

apple-web-evangelist: I verified that this is just wrong.

api/Blob.json
```diff
@@ -310,10 +310,10 @@
               "version_added": "54"
             },
             "safari": {
-              "version_added": false
+              "version_added": "14.1"
```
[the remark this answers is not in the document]
lukewarlow: Is it worth adding a note that while it did ship in 14.1, it shipped with a bug that causes it to crash if the Blob is too big? https://bugs.webkit.org/show_bug.cgi?id=220776#c8

api/KHR_parallel_shader_compile.json
```diff
@@ -29,10 +29,10 @@
             "version_added": null
           },
           "safari": {
-            "version_added": false
+            "version_added": "14.1"
           },
           "safari_ios": {
-            "version_added": false
+            "version_added": "14.5"
```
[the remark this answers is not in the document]
lukewarlow: That link says that Chrome fails the test, which contradicts the existing MDN browser support. So it could be a false positive for Safari?

html/elements/input/time.json
```diff
@@ -33,7 +33,7 @@
                 "version_added": true
               },
               "safari": {
-                "version_added": false,
+                "version_added": "14.1",
                 "notes": "See <a href='https://webkit.org/b/200416'>bug 200416</a>."
```
lukewarlow, line 37: This note isn't relevant anymore is it?

apple-web-evangelist: Addressed in 774f3c0.

css/properties/aspect-ratio.json
apple-web-evangelist: Addressed in 90183cb.

webextensions/manifest/background.json
apple-web-evangelist: Addressed in 90183cb.

KenG-pro (approved):

api/MediaRecorder.json
```diff
@@ -131,7 +131,7 @@
                 "version_added": "14"
               },
               "safari_ios": {
-                "version_added": "14"
+                "version_added": "14.3"
```
foolip, line 134:
```suggestion
                "version_added": "14"
```

This is the only remaining occurrence of "14.3", all others changed to "14", so I presume the intention in https://github.com/mdn/browser-compat-data/pull/10129#discussion_r622381020 was to treat "14.3" as "14" for `safari_os`.

browsers/safari_ios.json
```diff
@@ -158,8 +158,18 @@
           "engine": "WebKit",
           "engine_version": "610.1.28"
         },
+        "14.3": {
```
foolip, line 161: This "14.3" entry can be removed if it ends up not being used as a version anywhere in the data.

If I've misunderstood and you do want "14.3" for iOS but not "14.0.2" for macOS, then that seems OK too, but a lot more changes would be needed to get the data right.

api/AudioParam.json
```diff
@@ -80,7 +80,7 @@
               "version_added": false
```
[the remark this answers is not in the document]
foolip: This suggestion is still relevant, this was a case of only `safari_ios` being updated.

api/Blob.json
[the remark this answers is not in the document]
foolip: @lukewarlow this PR is getting really big, so I'd suggest sending a separate PR with your suggested note text, so that can be reviewed in isolation. (Creating that PR now would result in conflicts if this is merged, but they'll be easy to resolve.)

api/HTMLAreaElement.json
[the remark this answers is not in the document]
foolip: Thanks @apple-web-evangelist!

api/KHR_parallel_shader_compile.json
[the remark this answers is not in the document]
foolip: There is lots of incorrect data in BCD, and also https://mdn-bcd-collector.appspot.com/ has some false positives, so investigation is needed in every case. Support for WebGL extensions can also depend on the hardware which makes it trickier.

This *could* be a false positive for Safari, but the extension object is returned on a real iPhone I tested with, so I see no reason to question it. At least I have higher trust in this data than the Chrome data, which I haven't tested myself.

api/MediaRecorder.json
```diff
@@ -797,10 +811,10 @@
               "version_added": "36"
             },
             "safari": {
-              "version_added": false
+              "version_added": "14.1"
```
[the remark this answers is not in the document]
foolip: Fixed by 2d299ed4221d256727581bc0f362e113a4e9647d.

javascript/builtins/intl/NumberFormat.json
```diff
@@ -195,16 +195,26 @@
...
-                    "version_added": "10",
-                    "partial_implementation": true,
-                    "notes": "Doesn't support <code>currencyDisplay: 'narrowSymbol'</code>."
-                  },
+                  "safari": [
+                    {
+                      "version_added": "14.1"
+                    },
+                    {
+                      "version_added": "10",
```
ddbeck, line 203: If the 14.1 implementation supports `currencyDisplay: 'narrowSymbol'`, then…

```suggestion
                      "version_added": "10",
                      "version_removed": "14.1",
```

(otherwise, we need some more information about what the 14.1 support statement means)

css/properties/aspect-ratio.json
foolip: And I've updated mdn-bcd-collector in a way to avoid the WebKit bug leading to this false positive in https://github.com/foolip/mdn-bcd-collector/pull/1132.

webextensions/manifest/chrome_url_overrides.json
```diff
@@ -95,7 +95,8 @@
                 "version_added": false
               },
               "safari": {
-                "version_added": false
+                "version_added": "14.1",
+                "notes": "The user may select a specific extension to override the new tab or window page in Safari General Preferences."
```
ddbeck, line 99: If I understand this note, then developers can set `newtab`, but it won't take effect unless the user chooses the extension in Safari's preferences. If that's correct, we should make that relationship more explicit in the note. Something like:

```suggestion
                "notes": "An extension can define a custom new tab or window page, but it does not take effect until the user chooses the extension to override the page in Safari's <em>General</em> preferences."
```

(I based the formatting of the general preferences text based on my machine, but I'm not running the latest version. If _Safari General Preferences_ is the name of the window or tab, then please revise that bit)

ddbeck (changes requested):

json-derulo:
Closes #10153

api/MediaRecorder.json
apple-web-evangelist: Addressed in a086511.

browsers/safari_ios.json
apple-web-evangelist: Also addressed in a086511.

javascript/builtins/intl/NumberFormat.json
apple-web-evangelist: Addressed in d766019.

webextensions/manifest/chrome_url_overrides.json
apple-web-evangelist: Addressed in 2037915.

api/AudioParam.json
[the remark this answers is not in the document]
apple-web-evangelist: Addressed in 6b7b6d3.

foolip (approved):
api/ looks good to me now!

css/properties/scroll-margin-bottom.json
ddbeck: @apple-web-evangelist After _a lot_ of scrolling, I think this is the only actionable unresolved comment standing. Did you plan on tackling this? And do you have any other changes to this PR planned?

apple-web-evangelist: I see what you did there. 😉

The bug *is* fixed and shipped with 14.1/14.5. I added the additional support for the other `scroll-margin-*` entries in 90183cb. I believe that means we should add the `version_removed` entry in `scroll-margin-bottom`. Is that the last detail to address?

apple-web-evangelist: Addressed in fbf2d54.

ddbeck: Great, the bug being fixed simplifies things. Do I understand correctly that `scroll-snap-*` were all removed? If so, then I think you need to apply the `version_removed` change to the other `scroll-snap-*` statements and then we're done.

(what I wrote before I saw your latest change)

---
If `scroll-snap-*` still exist (that is, the alternative names are still supported), then the notes starting with "Scroll margin is not applied" need to qualify the versions they apply to (i.e., find and replace your way to "Before version 14.1/14.5, scroll margin is…") and then we're done.

If `scroll-snap-*` have been removed, then add a `version_removed` line to those support statements.

---

Either way, I believe that's the last of it. If you can get to that shortly, I can include this in today's release.

apple-web-evangelist: The `scroll-snap-*` exist as an alias, so the behavior is corrected.

apple-web-evangelist: In other words, they still work, and the "Scroll margin is not applied…" does not, in fact, apply to those cases.

ddbeck: Ah, in that case, I'd suggest undoing fbf2d54 and changing the notes to start with, "Before version 14.1, scroll margin is" (or 14.5, for the iOS entries). Otherwise, it looks like the aliases have gone away.

apple-web-evangelist: Will do!

apple-web-evangelist: Addressed in 20dfeee.

ddbeck: Thanks, @apple-web-evangelist. Do you want to apply that change to the text like this in the other `scroll-snap` entries too? (that is, `scroll-margin{,-top,-left,-right}.json`)? It's the same bug.

apple-web-evangelist: Ahh, I missed that. I will correct it quickly.

apple-web-evangelist: Addressed all of it with 8fc447d.

ddbeck: Great. Marking this resolved. 👍

ddbeck (approved):
I'm satisfied that all of the comments here that can be resolved have been. Thank you, @apple-web-evangelist!

PR #10140, opened 2021-04-28
<a> download attribute supported from iOS Safari 13.0

Hallot:
The "download" attribute on `<a>` elements has been supported on iOS Safari since 13.0.  
See https://bugs.webkit.org/show_bug.cgi?id=167341

### DIFF
--- a/html/elements/a.json
+++ b/html/elements/a.json
@@ -192,7 +192,7 @@
                 "version_added": "10.1"
               },
               "safari_ios": {
-                "version_added": false
+                "version_added": "13"
               },
               "samsunginternet_android": {
                 "version_added": "1.0",


html/elements/a.json
```diff
@@ -540,7 +540,7 @@
                 "version_added": "11.1"
               },
               "safari_ios": {
-                "version_added": false
+                "version_added": "13.0"
```
ddbeck, line 543: Omit the `.0` for major versions

```suggestion
                "version_added": "13"
```

ddbeck (changes requested):
Thanks for this PR, @Hallot. One suggestion in a line comment (to pass CI), then we can get this merged. Also, I see it's your first PR to the project—welcome!

html/elements/a.json
Hallot: Thanks, it was the wrong line as well, not sure how I messed that up.  
I have now properly changed the download attribute line.

ddbeck (approved):
Thank you, @Hallot (especially for catching that I issue I didn't)!

PR #10127, opened 2021-04-27
Moz outline radius removal

mirunacurtean:
A checklist to help your pull request get merged faster:
- [x] Summarize your changes
Added version removed for property and notes regarding the circumstances that led to that removal. 
- [x] Link to related issues or pull requests, if any
Follow up of https://github.com/mdn/content/issues/3453

PR #10151, opened 2021-04-28
Fix `css.properties.-moz-outline-radius*` problems and remove `yarn.lock`

ddbeck:
This fixes a few problems introduced by #10127 (which was merged prematurely):

- Removes `yarn.lock`
- Resolves discrepancies between the `-moz-outline-radius*` notes and Firefox and Fenix's `version_removed` values (I tested it and it was in fact removed in 88, not 87, on desktop)
- Does some light copy editing on the notes for `-moz-outline-radius*` data generally


ddbeck:
@mirunacurtean This is the follow up to your PR. I went ahead and corrected the issues with the PR here. If you haven't already, please see https://github.com/mdn/browser-compat-data/pull/10127#issuecomment-828388050.

mirunacurtean:
@ddbeck Thanks. In the conversations with Chris leading up to this change we thought setting the removed version as 87 was a way to indicate it was already out on FF version 88, but perhaps it's better to keep it 88 since it might create misunderstandings. 
I have no idea why the yarn.lock file had to be included in the push, but git would not give me the go ahead without adding it so it's just as good to me that it's been removed. 

The other changes look good to me, should I go ahead and merge it then?

ddbeck:
> 87 was a way to indicate it was already out on FF version 88, but perhaps it's better to keep it 88 since it might create misunderstandings.

I tested it and it was definitely still working in 87, so I think the change here is best. (It doesn't help that `version_removed` is a bit confusing too: it's really saying, "the version in which this feature stops working".)

But yes, if you approve the changes here, I'll say that's good enough to merge (one of the privileges of being an "Owner" I guess).

Elchi3 (approved):

PR #10157, opened 2021-04-28
Mark AbstractRange supported in Chrome 90

sideshowbarker:
https://source.chromium.org/chromium/chromium/src/+/b95a3cd5c2cc3b47769022fbafd68440315df924 added support for AbstractRange in Blink.

Per https://storage.googleapis.com/chromium-find-releases-static/b95.html#b95a3cd that puts it in Chrome 90.

And per https://wpt.fyi/results/dom/idlharness.window.html%3Fexclude%3DNode, Chrome is passing the relevant WPT tests for AbstractRange support.

### DIFF
--- a/api/AbstractRange.json
+++ b/api/AbstractRange.json
@@ -65,10 +65,15 @@
             "chrome_android": {
               "version_added": "90"
             },
-            "edge": {
-              "version_added": "18",
-              "version_removed": "79"
-            },
+            "edge": [
+              {
+                "version_added": "90"
+              },
+              {
+                "version_added": "18",
+                "version_removed": "79"
+              }
+            ],
             "firefox": {
               "version_added": "69"
             },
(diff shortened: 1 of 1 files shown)


foolip (approved):
Confirmed by testing https://mdn-bcd-collector.appspot.com/tests/api/AbstractRange in Chrome 89+90.

Also confirmed the same with Edge 89+90. This PR is correct, but @sideshowbarker do you want to add the Edge entries for this being back?

sideshowbarker:
> Also confirmed the same with Edge 89+90. This PR is correct, but @sideshowbarker do you want to add the Edge entries for this being back?

Yup — thanks, just now pushed a commit which adds that

foolip:
@sideshowbarker all of the subfeatures also need the Edge 90 fix, right?

sideshowbarker:
> @sideshowbarker all of the subfeatures also need the Edge 90 fix, right?

oops yeah — will add those too

PR #10197, opened 2021-05-02
Update Chromium versions for BluetoothRemoteGATTService API

vinyldarkscratch:
This PR updates and corrects the real values for Chrome-based browsers for the `BluetoothRemoteGATTService` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.1).  Additionally, this simplifies the child data for the API, removing the redundant multiple statements and derived flag data.  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/BluetoothRemoteGATTService


api/BluetoothRemoteGATTService.json
```diff
@@ -114,7 +114,7 @@
               "version_added": false
             },
             "edge": {
-              "version_added": false
+              "version_added": "79"
```
foolip, line 117: Chrome above says false, is that right? Can you update all Chromium browsers together?

vinyldarkscratch: Confirmed this is in Chrome as well, just updated all the other Chromium browsers!

api/BluetoothRemoteGATTService.json
```diff
@@ -125,27 +121,9 @@
             "ie": {
               "version_added": false
             },
-            "opera": [
```
foolip, line 128: I agree with simplifying all of this repeated OS and flag information since it's already on the parent feature. But it does mean that information won't be visible on a page like https://developer.mozilla.org/en-US/docs/Web/API/BluetoothRemoteGATTService/getCharacteristic#browser_compatibility.

@ddbeck how do you think we should deal with cases like this?

foolip (approved):
Looks good, but I'd also like @ddbeck to take a look at the question of deduplicating.

Note that searching for "macOS only" finds a bunch more cases like this, and probably it's best to delete all of that in a single PR, separate from fixing the data?

api/BluetoothRemoteGATTService.json
ddbeck: There's some disagreement on this point. See https://github.com/mdn/browser-compat-data/issues/6509 for the broader issue.

My approach to flags has been to only represent the flag on the feature that requires the flag, not its descendants (unless the flags differ). For example, a CSS property behind a flag doesn't need to duplicate the flag data on subfeatures for property values. This does lead to a poorer presentation for individual interface member pages, however. In those cases I don't think it's too bad: for `SomeFlaggedInterface.method()`, you'd get an informative `SomeFlaggedInterface is not defined` error.

But since there's no consensus, I'm happy to allow anything that looks reasonable. If you think it'd be preferable to have flags inherit, then go for it, but I'm not going to ask for it myself.

api/BluetoothRemoteGATTService.json
```diff
@@ -32,7 +32,7 @@
           "edge": [
             {
               "version_added": "79",
-              "notes": "macOS only."
+              "notes": "macOS and Windows 10."
```
ddbeck, line 35: Two things here:

- Could we make this change consistent with the other Chromiums?
- Can we revise these notes to be more explicit? Something like "Supported on Windows 10 and later and macOS." or "Supported by default only on Windows 10 and macOS." (whatever the case is).

ddbeck (dismissed):

foolip:
The "macOS only.", "Linux and versions of Windows earlier than 10.", and "Windows 10." notes are very widespread, although inconsistently applied, and it would be great to  change them all at once to not create confusion for future contributors trying to update parts of this data.

Some of the data also indicates unconditional support, like api.BluetoothRemoteGATTCharacteristic.writeValueWithResponse. It would be good to decide what level of OS support we should consider complete and consistently either drop notes or add them where missing.

ddbeck:
@foolip Yeah, I understand this stuff is widespread. We did adopt a guideline to try to encourage new data to be more consistent when it comes to OS limitations: https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#operating-system-limitations-imply-partial_implementation. We could revise it to be more assertive about what note is needed (e.g., "You must have a note that reads, …").

foolip:
Right, so it should be `partial_implementation` in a lot of cases here, but when does it stop being `partial_implementation`? Chrome is supported on Windows, macOS, Linux and ChromeOS. The simple answer would be to say that it's `partial_implementation` if something isn't supported on any of those platforms. What feels a bit weird about that is that Linux support in particular won't matter much to web developers, since it's not a very significant part of the user share.

Edimalius (approved):
vinyldarkscratch:api/BluetoothRemoteGATTService/edge-corrections

ddbeck:
> The simple answer would be to say that it's `partial_implementation` if something isn't supported on any of those platforms. What feels a bit weird about that is that Linux support in particular won't matter much to web developers, since it's not a very significant part of the user share.

My glib answer to that is: well, if a browser claims to support Linux, then excluding some features from Linux seems like a partial implementation to me.

But I get where you're coming from. I'd accept a revision of the guidelines to say something like, if it's not supported on either of the top two desktop operating systems (citing some source which names Windows and macOS), then the `partial_implementation` flag is required; otherwise, only the note is required.

foolip:
> My glib answer to that is: well, if a browser claims to support Linux, then excluding some features from Linux seems like a partial implementation to me.

Let's go with that simpler solution for now. @vinyldarkscratch do you feel like wrangling the Bluetooth into that shape? Alternatively, updating data in a way that matches the current structure and filing an issue about the Bluetooth data would be OK with me.

vinyldarkscratch:
This commit should do the trick?

vinyldarkscratch:
Hey @foolip, could you give this one a re-review when you are able?

api/BluetoothRemoteGATTService.json
```diff
@@ -62,22 +56,19 @@
           "opera": [
             {
               "version_added": "43",
-              "notes": "macOS only."
+              "partial_implementation": true,
+              "notes": "Before Before Opera 57, this feature was only supported in macOS.  In Before Opera 57, support was added for Windows 10. Linux support is not enabled by default."
```
foolip, line 60:
```suggestion
              "notes": "Before Opera 57, this feature was only supported in macOS.  In Opera 57, support was added for Windows 10. Linux support is not enabled by default."
```

api/BluetoothRemoteGATTService.json
foolip: I'll go along with removing the flags on subfeatures here. I don't know what I think should be the guideline :)

foolip (approved):

foolip:
I've filed https://github.com/mdn/browser-compat-data/issues/11532 about making the remaining data consistent with this.

PR #10198, opened 2021-05-02
Add Edge versions for FetchEvent API

vinyldarkscratch:
This PR adds real values for Microsoft Edge for the `FetchEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.0).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/FetchEvent

### DIFF
--- a/api/FetchEvent.json
+++ b/api/FetchEvent.json
@@ -12,7 +12,7 @@
             "version_added": "40"
           },
           "edge": {
-            "version_added": "≤18"
+            "version_added": "17"
           },
           "firefox": {
             "version_added": "44",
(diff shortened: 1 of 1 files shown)


foolip:
Some "≤79" remain after this change, don't the tests cover those members?

vinyldarkscratch:
They either don't cover those members (at least, in the collector version I used, which is slightly out of date I'll admit), or they're returning a value/throwing an error that results in us deciding unknown support.  Either way, I plan to look into them with more detail in a future PR.

foolip (approved):

PR #10201, opened 2021-05-02
Add Edge versions for HTMLMediaElement API

vinyldarkscratch:
This PR adds real values for Microsoft Edge for the `HTMLMediaElement` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.1).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/HTMLMediaElement.json
+++ b/api/HTMLMediaElement.json
@@ -688,7 +688,7 @@
               "version_added": "58"
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "79"
             },
             "firefox": {
               "version_added": false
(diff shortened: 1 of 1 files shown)


api/HTMLMediaElement.json
```diff
@@ -1479,7 +1479,7 @@
               "version_added": null
             },
             "edge": {
-              "version_added": null
+              "version_added": false
```
foolip, line 1482: This `initialTime` entry really should be removed. I'll send a PR.

foolip (approved):

PR #10202, opened 2021-05-02
Add Edge versions for api.IDBTransaction.objectStoreNames

vinyldarkscratch:
This PR adds real values for Microsoft Edge for the `objectStoreNames` member of the `IDBTransaction` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.1).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/IDBTransaction/objectStoreNames

### DIFF
--- a/api/IDBTransaction.json
+++ b/api/IDBTransaction.json
@@ -640,7 +640,7 @@
               "version_added": "48"
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "79"
             },
             "firefox": {
               "version_added": true


foolip:
The test here involves only the prototype, but other parts of this API are available in Edge 12. What manual checking of this have you done, to ensure this isn't a problem of the interface object being missing and then being added?

vinyldarkscratch:
Running the example on the [MDN web docs page](https://developer.mozilla.org/en-US/docs/Web/API/IDBTransaction#examples) (that is, https://mdn.github.io/to-do-notifications/) and checking the `IDBTransaction` object logged to the console shows that `objectStoreNames` is not present in the transaction instance either (it's only in the `IDBDatabase`).

foolip:
Excellent, thanks for testing that!

foolip (approved):

PR #10220, opened 2021-05-02
Update Edge versions for MediaMetadata API

vinyldarkscratch:
This PR updates and corrects the real values for Microsoft Edge for the `MediaMetadata` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.1).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/MediaMetadata

### DIFF
--- a/api/MediaMetadata.json
+++ b/api/MediaMetadata.json
@@ -12,7 +12,7 @@
             "version_added": "57"
           },
           "edge": {
-            "version_added": false
+            "version_added": "79"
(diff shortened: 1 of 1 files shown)


foolip (approved):
Matches https://developer.mozilla.org/en-US/docs/Web/API/MediaSession

PR #10221, opened 2021-05-02
Add Edge versions for MouseEvent API

vinyldarkscratch:
This PR adds real values for Microsoft Edge for the `MouseEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.1).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/MouseEvent.json
+++ b/api/MouseEvent.json
@@ -65,7 +65,7 @@
               "version_added": "47"
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "12"
             },
             "firefox": {
               "version_added": "11"


api/MouseEvent.json
```diff
@@ -65,7 +65,7 @@
               "version_added": "47"
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "12"
```
foolip, line 68: I confirmed that `new MouseEvent('bla')` works in Edge 15.

I also confirmed http://mdn-bcd-collector.appspot.com/tests/api/MouseEvent fails in IE11 with "Object doesn't support this action" so no IE update needed. Good!

foolip (approved):

PR #10225, opened 2021-05-03
Remove irrelevant Notation API

vinyldarkscratch:
This PR removes the Notation API data, as per the [removal of irrelevant features guideline](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-features).  Since this was removed back in Safari 9 and no other browsers have shown signs of support, this API qualifies as an irrelevant feature.


foolip:
Do you want to send a PR to delete this file entirely instead, under the irrelevant feature guideline?

foolip (approved):
Approving but not merging since https://developer.mozilla.org/en-US/docs/Web/API/Notation will also be removing. It's best to merge this at the same time as an MDN PR is merged.

Elchi3 (approved):
I opened an MDN PR for this. Thanks for finding and cleaning this! I'm a fan of getting rid of irrelevant entries :)

PR #10226, opened 2021-05-03
Add Edge versions for NotificationEvent API

vinyldarkscratch:
This PR adds real values for Microsoft Edge for the `NotificationEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.1).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/NotificationEvent

### DIFF
--- a/api/NotificationEvent.json
+++ b/api/NotificationEvent.json
@@ -12,7 +12,7 @@
             "version_added": "42"
           },
           "edge": {
-            "version_added": "≤18"
+            "version_added": "17"
           },
           "firefox": {
             "version_added": "44",
@@ -63,7 +63,7 @@
               "version_added": "42"
             },
             "edge": {
-              "version_added": "≤18"
+              "version_added": "17"
             },
             "firefox": {
               "version_added": "44",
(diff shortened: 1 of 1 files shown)


foolip:
https://developer.mozilla.org/en-US/docs/Web/API/Notification was available earlier, is this event not a core part of the API or why did it ship later?

foolip:
So `NotificationEvent` is `[Exposed=ServiceWorker]` and should then probably coincide when with Edge got service worker support. But https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerGlobalScope#browser_compatibility has Edge 79 everywhere.

@vinyldarkscratch can you look into what part of the data is wrong? It can't all be right :)

vinyldarkscratch:
It looks like `ServiceWorkerGlobalScope` actually has `≤79`, but after crunching the collector's results, it seems to be indicating that support is included in Edge 17 (which matches up with the rest of our data).  I hadn't submitted a PR for updates to the global scope quite yet, as it included changes that I wanted to look into more.

TL;DR: Edge 17 seems accurate!

foolip:
Great, thanks for checking @vinyldarkscratch!

foolip (approved):

PR #10232, opened 2021-05-03
Add Edge versions for OfflineAudioCompletionEvent API

vinyldarkscratch:
This PR adds real values for Microsoft Edge for the `OfflineAudioCompletionEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.1).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/OfflineAudioCompletionEvent

### DIFF
--- a/api/OfflineAudioCompletionEvent.json
+++ b/api/OfflineAudioCompletionEvent.json
@@ -63,7 +63,7 @@
               "notes": "Before Chrome 59, the default values were not supported."
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "79"
             },
             "firefox": {
               "version_added": "53"


foolip (approved):
This is the `OfflineAudioCompletionEvent` constructor. I tried http://mdn-bcd-collector.appspot.com/tests/api/OfflineAudioCompletionEvent/OfflineAudioCompletionEvent in Edge 18 and it fails with "Function expected", which indeed seems like no constructor support. (None of the arguments are functions.)

I tried `new OfflineAudioCompletionEvent('bla', { renderedBuffer: anAudioBufferInstance })` and it still fails with "Function expected".

PR #10233, opened 2021-05-03
Add Edge versions for OfflineAudioContext API

vinyldarkscratch:
This PR adds real values for Microsoft Edge for the `OfflineAudioContext` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.1).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/OfflineAudioContext

### DIFF
--- a/api/OfflineAudioContext.json
+++ b/api/OfflineAudioContext.json
@@ -131,7 +131,7 @@
               }
             ],
             "edge": {
-              "version_added": "≤79"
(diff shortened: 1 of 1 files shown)


foolip (approved):
OK, so `suspend()` and `resume()` came later than the feature itself. That plausible.

PR #10235, opened 2021-05-03
Add Edge versions for OverconstrainedErrorEvent API

vinyldarkscratch:
This PR adds real values for Microsoft Edge for the `OverconstrainedErrorEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.1).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/OverconstrainedErrorEvent

### DIFF
--- a/api/OverconstrainedErrorEvent.json
+++ b/api/OverconstrainedErrorEvent.json
@@ -11,7 +11,7 @@
             "version_added": null
           },
           "edge": {
-            "version_added": null
+            "version_added": false
           },
           "firefox": {
             "version_added": null
@@ -58,7 +58,7 @@
               "version_added": null
             },
             "edge": {
-              "version_added": null
+              "version_added": false
             },
             "firefox": {
               "version_added": null


foolip (commented):
Perhaps more of this file can be updated at once, at least all Chromium browsers?

I think this was once supported, so more digging is needed.

foolip (commented):
I'd expect this to match https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamTrack/onoverconstrained#browser_compatibility which says Edge 12. Can you check that out?

vinyldarkscratch:
I can't really find much information on this particular API...  It looks like this file was added by #1422, and at some point, it _was_ in the spec (but since then it's been removed, it looks like).  This seems to be a little bit of a rabbit hole to go down eventually...just not before I head to bed.  😛

foolip:
This API should most likely match https://mdn-bcd-collector.appspot.com/tests/api/MediaStreamTrack/onoverconstrained, and while that test doesn't work in Edge, I've confirmed that `'onoverconstrained' in MediaStreamTrack.prototype` is true.

I think this is probably a case of the event interface not being exposed but the event actually being supported.

vinyldarkscratch:
I started looking into this again whilst tackling trying to obtain the versions for Chrome as well.  Looking around, I'm unable to find anything that suggests this event handler was actually implemented in any spec or browser -- even the page for `onoverconstrained` talks about a `MediaStreamErrorEvent`, which I can't find any data on either.

I was able to trace the addition of this file back to #1422, which seemed to be a use of webidl2mdnbcd to convert the [spec](https://w3c.github.io/mediacapture-main/getusermedia.html) to BCD data.  My best guess is that it was performed while the spec was in an early stage, and as such included IDL that was since abandoned.  I'm attempting to track this down now, but issue/PR history for the spec's repository isn't helping much.

vinyldarkscratch:
Ah, yep, this was in a previous version of the spec as suspected (https://web.archive.org/web/20190731084828/https://w3c.github.io/mediacapture-main/).  It appears that it was initially called `MediaStreamErrorEvent`, but then was [renamed](https://github.com/w3c/mediacapture-main/pull/215) to `OverconstrainedErrorEvent`.  From there, it just seems like it suddenly vanished with no announcement or mention.

vinyldarkscratch:
Closing in favor of #10738, which includes Chrome updates as well.

PR #3312, opened 2019-01-18
Add missing Edge 16 data based on Confluence

foolip:
Produced by `npm run confluence -- --fill-only --browsers=edge` with
local changes applied to only include changes where the new version
was 16.

PR #10239, opened 2021-05-03
Add Edge versions for PerformanceNavigationTiming API

vinyldarkscratch:
This PR adds real values for Microsoft Edge for the `PerformanceNavigationTiming` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.1).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/PerformanceNavigationTiming

### DIFF
--- a/api/PerformanceNavigationTiming.json
+++ b/api/PerformanceNavigationTiming.json
@@ -405,7 +405,7 @@
               "version_added": "57"
             },
             "edge": {
-              "version_added": "≤18"
+              "version_added": "16"
             },
             "firefox": {
               "version_added": "58"


foolip (approved):

foolip:
Hmm, wait, this is `toJSON` which is prone to false positives due to inheritance. Have you checked where on the prototype chain this appeared in Edge 16?

vinyldarkscratch:
This seems to match the data for `PerformanceEntry.toJSON`, however I tracked down the commits to find that data came from Confluence (#3312), so I wonder if it's accurate as well.  If anything though, I'd imagine it might be supported _earlier_?

foolip:
The way to test with would be `Object.prototype.hasOwnProperty.call(PerformanceNavigationTiming.prototype, 'toJSON')`. I just tested it, and it's false in Edge 15 *and* still false in Edge 16 and 18. It's true in Edge 79 however. So it looks like the data was wrong, and still is.

It is *possible* that a `toJSON` on an inherited interface actually does serialize the properties of `PerformanceNavigationTiming`, which would be wrong per Web IDL, but actually work for web developers.

Testing with a `PerformanceNavigationTiming` is needed to figure this one out.

foolip:
One can get an instance as `entry = performance.getEntriesByType("navigation")[0]`. And it turns out that `entry.toJSON()` actually *does* include members of `PerformanceNavigationTiming`, even though the `toJSON` is in the wrong place on the prototype chain.

OK then, I guess this data is correct.

But it does mean that tests for `toJSON` support cannot be trusted, one really needs an instance, to call `instance.toJSON()` and see if the expected members are there. That's what it means for `toJSON()` to be meaningfully supported.

PR #10254, opened 2021-05-03
Add Edge versions for PresentationRequest API

vinyldarkscratch:
This PR adds real values for Microsoft Edge for the `PresentationRequest` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.1).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/PresentationRequest

### DIFF
--- a/api/PresentationRequest.json
+++ b/api/PresentationRequest.json
@@ -12,7 +12,7 @@
             "version_added": "48"
           },
           "edge": {
-            "version_added": "≤79"
+            "version_added": "79"
           },
           "firefox": {
             "version_added": "51",
@@ -77,7 +77,7 @@
               "version_added": "48"
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "79"
             },
             "firefox": {
               "version_added": "51",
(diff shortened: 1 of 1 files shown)


api/PresentationRequest.json
```diff
@@ -451,7 +451,7 @@
               "version_added": "48"
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": false
```
foolip, line 454: Is it really correct that Chrome has this but not Edge?

foolip: https://mdn-bcd-collector.appspot.com/tests/api/PresentationRequest/startWithDevice tests false for me in Chrome as well, so no, it's not correct.

@vinyldarkscratch do you want to back this change out and fix the `startWithDevice` entry as one PR?

vinyldarkscratch: Good catch!  I created #10493 to take care of all the Chrome data.

foolip (approved):

PR #10255, opened 2021-05-03
Add Edge versions for ProgressEvent API

vinyldarkscratch:
This PR adds real values for Microsoft Edge for the `ProgressEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.1).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/ProgressEvent

### DIFF
--- a/api/ProgressEvent.json
+++ b/api/ProgressEvent.json
@@ -61,7 +61,7 @@
               "version_added": true
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "14"
             },
             "firefox": {
               "version_added": "22"


foolip:
Can you check what was going on in Edge 12 and 13? Was the interface object exposed but the constructor throws an exception?

Detection of constructor support isn't necessary 100% so double checking is best :)

vinyldarkscratch:
In Edge 13, a `TypeError` occurs, indicating "Function expected" -- I'm assuming it's the same for Edge 12.

foolip:
Thanks @vinyldarkscratch, "Function expected" is pretty typical of something not being supported as a constructor at all.

foolip (approved):

PR #10280, opened 2021-05-04
Mark SpeechGrammar() constructor non-standard

sideshowbarker:
https://wicg.github.io/speech-api/#speechgrammar defines no constructor for the `SpeechGrammar` object. So this change marks it standard_track:false.

Related MDN change: https://github.com/mdn/content/pull/4707


foolip (approved):
Indeed, I have in backlog to remove this constructor in Chromium.

Issue #9635, opened 2021-03-27
css.properties.overflow-x - clip value is now supported in Chrome 90


raphaelsaunier:
#### What information was incorrect, unhelpful, or incomplete?

the `clip` value for `overflow(-x|-y)` properties is now supported in Chrome. c.f. https://www.chromestatus.com/feature/5638444178997248

#### What did you expect to see?

A green box indicating that the value is supported in Chrome 90+

#### Did you test this? If so, how?

By setting an `overflow-x: clip` value on a relatively positioned div with an absolutely positioned child element that overflows the parent div.

<details>
<summary>MDN page report details</summary>

* Query: `css.properties.overflow-x`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/CSS/overflow-x
* Report started: 2021-03-27T15:03:47.578Z

</details>

ddbeck:
Thanks for reporting this, @raphaelsaunier. You're right, we need to reconcile the `clip` subfeature data for `overflow`, `overflow-x`, and `overflow-y`. Thanks to the details you provided, this was quick to fix and https://github.com/mdn/browser-compat-data/pull/10281 is now open to resolve this.

jculver050:
Close

PR #10281, opened 2021-05-04
Reconcile `css.properties.overflow-{x,y}.clip` with `css.properties.overflow.clip`

ddbeck:
This PR reconciles `css.properties.overflow-x.clip` and `css.properties.overflow-y.clip` with `css.properties.overflow.clip`, which ought to be the same for Chrome. While I was here, I mirrored to Opera.

Original source: https://www.chromestatus.com/feature/5638444178997248
    
Fixes https://github.com/mdn/browser-compat-data/issues/9635.

### DIFF
--- a/css/properties/overflow-x.json
+++ b/css/properties/overflow-x.json
@@ -92,7 +92,7 @@
                 "version_added": false
               },
               "opera": {
-                "version_added": false
+                "version_added": "76"
               },
               "opera_android": {
                 "version_added": false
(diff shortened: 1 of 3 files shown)


ddbeck:
@Elchi3 Would you mind looking at this? It's been stalled for sometime and it's not a particularly big change.

Elchi3 (approved):
Thanks Daniel, makes sense to me! 👍

PR #10285, opened 2021-05-05
Mark OfflineAudioContext's resume() shipped in Safari 14.1

foolip:
Confirmed with https://mdn-bcd-collector.appspot.com/tests/api/OfflineAudioContext in Safari 14.1.

This was missed in https://github.com/mdn/browser-compat-data/pull/10129.

### DIFF
--- a/api/OfflineAudioContext.json
+++ b/api/OfflineAudioContext.json
@@ -436,10 +436,10 @@
               "version_added": "36"
             },
             "safari": {
-              "version_added": false
+              "version_added": "14.1"
             },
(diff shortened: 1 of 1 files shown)


foolip:
@apple-web-evangelist FYI, this stood out as suspicious when looking at https://developer.mozilla.org/en-US/docs/Web/API/OfflineAudioContext#browser_compatibility while reviewing another PR.

HTML-CODDER:
great

ddbeck (approved):
Looks good! 🎉

PR #10286, opened 2021-05-05
Mark SourceBuffer.p.removeAsync non-standard

sideshowbarker:
Similar to the case of `SourceBuffer.p.appendBufferAsync` in https://github.com/mdn/browser-compat-data/pull/10279, `SourceBuffer.p.removeAsync` isn’t part of any spec and at https://developer.mozilla.org/en-US/docs/Web/API/SourceBuffer/removeAsync is already marked non-standard in MDN. So this change marks it standard_track:false in BCD.

### DIFF
--- a/api/SourceBuffer.json
+++ b/api/SourceBuffer.json
@@ -925,7 +925,7 @@
           },
           "status": {
             "experimental": false,
-            "standard_track": true,
+            "standard_track": false,
             "deprecated": false
           }
         }


foolip (approved):
This is still in https://github.com/mozilla/gecko-dev/blob/e08a9aa5c5dfb77e8847ab7af3f753a9c3e551f8/dom/webidl/SourceBuffer.webidl#L61, but I question the value of having this in the data. But it's not standard, this PR is correct.

sideshowbarker:
> This is still in [mozilla/gecko-dev@`e08a9aa`/dom/webidl/SourceBuffer.webidl#L61](https://github.com/mozilla/gecko-dev/blob/e08a9aa5c5dfb77e8847ab7af3f753a9c3e551f8/dom/webidl/SourceBuffer.webidl#L61), but I question the value of having this in the data.

Me too. But it’s documented at https://developer.mozilla.org/en-US/docs/Web/API/SourceBuffer/removeAsync and we have an mdn_url for it in BCD. So as long as we do have it included in BCD, we should at least have it flagged correctly (which this patch has now done). But I’d personally be happy with just dropping it completely from BCD.

PR #10288, opened 2021-05-05
    Remove StaticRange.p.toRange

sideshowbarker:
`StaticRange.p.toRange` isn’t part of any spec, was only ever implemented in Chrome — but way back at Chrome 60. So this change removes it, per https://github.com/mdn/browser-compat-data/pull/10288#issuecomment-832610787

Related MDN change: https://github.com/mdn/content/pull/4723

### DIFF
--- a/api/StaticRange.json
+++ b/api/StaticRange.json
@@ -336,55 +336,6 @@
             "deprecated": false
           }
         }
-      },
-      "toRange": {
-        "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/StaticRange/toRange",
-          "description": "<code>toRange()</code>",
-          "support": {
-            "chrome": {
-              "version_added": "60"
-            },
-            "chrome_android": {
-              "version_added": "60"
-            },
-            "edge": {
-              "version_added": "18"
-            },
-            "firefox": {
-              "version_added": false
-            },
-            "firefox_android": {
-              "version_added": false
-            },
-            "ie": {
-              "version_added": false
-            },
-            "opera": {
-              "version_added": "47"
-            },
-            "opera_android": {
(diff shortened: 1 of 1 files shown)


foolip:
Per https://mdn-bcd-collector.appspot.com/tests/api/StaticRange/toRange this isn't actually in Chrome. Looks like it was removed in https://chromium.googlesource.com/chromium/src/+/e2b8278dd5aaf5a6ec415aa5e7dd828f17b644f6 and this entry should actually be removed entirely.

sideshowbarker:
> Per [mdn-bcd-collector.appspot.com/tests/api/StaticRange/toRange](https://mdn-bcd-collector.appspot.com/tests/api/StaticRange/toRange) this isn't actually in Chrome. Looks like it was removed in [chromium.googlesource.com/chromium/src/+/e2b8278dd5aaf5a6ec415aa5e7dd828f17b644f6](https://chromium.googlesource.com/chromium/src/+/e2b8278dd5aaf5a6ec415aa5e7dd828f17b644f6) and this entry should actually be removed entirely.

OK, I’ve amended the commit to remove it — as well as updating the PR description accordingly.

foolip (approved):
https://developer.mozilla.org/en-US/docs/Web/API/StaticRange/toRange also needs to be removed, so approving but not merging, so this could go together with a MDN PR.

sideshowbarker:
> [developer.mozilla.org/en-US/docs/Web/API/StaticRange/toRange](https://developer.mozilla.org/en-US/docs/Web/API/StaticRange/toRange) also needs to be removed, so approving but not merging, so this could go together with a MDN PR.

OK, I’ve amended the https://github.com/mdn/content/pull/4723 change to remove it completely from MDN.

PR #10307, opened 2021-05-06
Add Edge versions for CloseEvent API constructor

vinyldarkscratch:
This PR adds real values for Microsoft Edge for the `CloseEvent` API's constructor, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.1).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/CloseEvent

### DIFF
--- a/api/CloseEvent.json
+++ b/api/CloseEvent.json
@@ -65,7 +65,7 @@
               "version_added": null
             },
             "edge": {
-              "version_added": null
+              "version_added": "14"
             },
             "firefox": {
               "version_added": "8"


foolip (commented):
Are you sure this was in Edge 79 too? Filling in the Chrome data as well would increase my confidence that this should just be a single statement :)

vinyldarkscratch:
Yep, this is present in Edgeium as well!  I'll write a follow-up PR to update Chrome as well.

foolip:
This is the event interface for https://mdn-bcd-collector.appspot.com/tests/api/WebSocket/onclose, and https://developer.mozilla.org/en-US/docs/Web/API/WebSocket/close#browser_compatibility says that was supported since Edge 12.

It could be that a "close" event was fired but with the regular `Event` interface, or it was a `CloseEvent` but the interface object wasn't exposed. To be sure, I'm afraid that a test that case that causes the event to fire is necessary. http://wpt.live/websockets/Close-1000-reason.any.html is such a test, but unfortunately it doesn't seem to work on Edge 15. But maybe it can be modified and testing on Edge 12.

vinyldarkscratch:
Apologies for the long delay in a response.  This PR is actually updating the data for the event's constructor, rather than the entire event.

foolip (approved):

PR #10309, opened 2021-05-06
Update Chromium versions for PushSubscriptionChangeEvent API

vinyldarkscratch:
This PR adds real values for Chromium for the `PushSubscriptionChangeEvent` API and the corresponding events, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.1).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/PushSubscriptionChangeEvent


foolip (changes requested):
The data around this event seems to be pretty wrong. This is the event interface for "pushsubscriptionchange" events, and there are "onpushsubscriptionchange" and "pushsubscriptionchange_event" entries in BCD, but their data appears to be wrong. How about a PR that updates those two, so it's clearer if the result is consistent or not?

vinyldarkscratch:
Updated!

foolip (approved):
Data makes sense for Chromium now! It's still contradictory for Safari, if you'd like to send another PR for that.

PR #10310, opened 2021-05-06
Add Edge versions for PushSubscriptionOptions API

vinyldarkscratch:
This PR adds real values for Microsoft Edge for the `PushSubscriptionOptions` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.1).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/PushSubscriptionOptions.json
+++ b/api/PushSubscriptionOptions.json
@@ -11,7 +11,7 @@
             "version_added": true
           },
           "edge": {
-            "version_added": "17"
+            "version_added": "16"
           },
           "firefox": {
             "version_added": true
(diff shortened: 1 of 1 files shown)


foolip (changes requested):
This doesn't match https://developer.mozilla.org/en-US/docs/Web/API/PushSubscription/options#browser_compatibility, which is an object of this type, so something must be wrong.

vinyldarkscratch:
Probably a case of an unexposed interface.  The "16" from `api.PushSubscription.options` came from #1518, so I'm a little more trusting of its data.  Correcting now.

foolip (approved):

PR #10314, opened 2021-05-06
Add Edge versions for RTCDataChannelEvent API

vinyldarkscratch:
This PR adds real values for Microsoft Edge for the `RTCDataChannelEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.1).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/RTCDataChannelEvent

### DIFF
--- a/api/RTCDataChannelEvent.json
+++ b/api/RTCDataChannelEvent.json
@@ -12,7 +12,7 @@
             "version_added": "28"
           },
           "edge": {
-            "version_added": "≤79"
+            "version_added": "79"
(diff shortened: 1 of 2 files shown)


foolip (changes requested):
Can you also update the "datachannel_event" entry? That has to be wrong if this is correct.

foolip (approved):

PR #10318, opened 2021-05-06
Set Edge to false for unsupported SVG API features

vinyldarkscratch:
This PR changes Microsoft Edge from `null` -> `false` for various SVG APIs, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.1).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/SVG*

### DIFF
--- a/api/SVGURIReference.json
+++ b/api/SVGURIReference.json
@@ -12,7 +12,7 @@
             "version_added": null
           },
           "edge": {
-            "version_added": null
+            "version_added": false
           },
           "firefox": {
             "version_added": null
--- a/api/SVGURIReference.json
+++ b/api/SVGURIReference.json
@@ -12,7 +12,7 @@
             "version_added": null
           },
           "edge": {
-            "version_added": false
+            "version_added": null
           },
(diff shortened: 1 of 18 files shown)


api/SVGURIReference.json
```diff
@@ -12,7 +12,7 @@
             "version_added": null
           },
           "edge": {
-            "version_added": null
+            "version_added": false
```
foolip, line 15:
```suggestion
            "version_added": null
```

`SVGURIReference` is part of the problem in https://github.com/mdn/browser-compat-data/issues/7752 and should be removed from BCD/MDN, but it's not technically true that it's not supported, because the `href` attributes is supported on various interfaces. Reverting this back to null so that it's an all-null entry and less to wonder about when it's removed.

foolip (approved):

foolip:
I've sent https://github.com/mdn/browser-compat-data/pull/10522 about `SVGURIReference`.

PR #10319, opened 2021-05-06
Add Edge versions for SVG APIs

vinyldarkscratch:
This PR adds real values for Microsoft Edge for various SVG APIs, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.1).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/SVG*

### DIFF
--- a/api/SVGFEDropShadowElement.json
+++ b/api/SVGFEDropShadowElement.json
@@ -12,7 +12,7 @@
             "version_added": "18"
           },
           "edge": {
-            "version_added": "≤79"
+            "version_added": "79"
           },
           "firefox": {
             "version_added": "30"
--- a/api/SVGPointList.json
+++ b/api/SVGPointList.json
@@ -291,7 +291,7 @@
               "version_added": true
             },
             "edge": {
-              "version_added": true
+              "version_added": "79"
             },
             "firefox": {
               "version_added": "5"
--- a/api/SVGStringList.json
+++ b/api/SVGStringList.json
@@ -295,7 +295,7 @@
               "version_added": true
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "79"
             },
             "firefox": {
               "version_added": "13"
--- a/api/SVGTransformList.json
+++ b/api/SVGTransformList.json
@@ -58,7 +58,7 @@
               "version_added": true
             },
             "edge": {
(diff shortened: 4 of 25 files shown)


api/SVGFEDropShadowElement.json
```diff
@@ -12,7 +12,7 @@
             "version_added": "18"
           },
           "edge": {
-            "version_added": "≤79"
+            "version_added": "79"
```
foolip, line 15: Matches https://developer.mozilla.org/en-US/docs/Web/SVG/Element/feDropShadow#browser_compatibility, good!

api/SVGPointList.json
```diff
@@ -291,7 +291,7 @@
               "version_added": true
             },
             "edge": {
-              "version_added": true
+              "version_added": "79"
```
foolip, line 294: This is a `length` property, and at first I though it's unlikely that `SVGPointList` was supported without this, but there's also `numberOfItems` which was supported from Edge 12. http://mdn-bcd-collector.appspot.com/tests/api/SVGPointList/length looks correct to me too.

api/SVGStringList.json
```diff
@@ -295,7 +295,7 @@
               "version_added": true
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "79"
```
foolip, line 298: Another `length` property, same story.

api/SVGTransformList.json
```diff
@@ -58,7 +58,7 @@
               "version_added": true
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "79"
```
foolip, line 61: `length` property, but lots of the members of `SVGTransformList` are missing, notably including `numberOfItems`. I think this is still correct though.

api/SVGViewElement.json
```diff
@@ -60,7 +60,8 @@
               "version_removed": "56"
             },
             "edge": {
-              "version_added": null
+              "version_added": "12",
+              "version_removed": "79"
```
foolip, line 64: Nice this makes it clearer that `viewTarget` should probably be removed.

foolip (approved):

PR #10331, opened 2021-05-06
submit.json: Escape <button> HTML in Firefox notes

Ashenfactory:
The footnotes for Firefox version 1 and 4 contain unescaped "<" and ">" marks, leading to the notes containing an actual button element.


ddbeck (approved):
Nice catch, @Ashenfactory. Thanks for this PR. Also, I noticed it's your first to the project—welcome! 🎉

PR #10336, opened 2021-05-06
Drop the api.TransitionEvent.animationName feature

sideshowbarker:
https://drafts.csswg.org/css-transitions/#interface-transitionevent has no member named `animationName`. Instead it has a `propertyName` member, which we already include in BCD.

Related MDN change: https://github.com/mdn/content/pull/4760

### DIFF
--- a/api/TransitionEvent.json
+++ b/api/TransitionEvent.json
@@ -124,58 +124,10 @@
           }
         }
       },
-      "animationName": {
-        "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/TransitionEvent/animationName",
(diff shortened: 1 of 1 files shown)


foolip (approved):
I can't find any trace of this in http://web-confluence.appspot.com/#!/catalog?releases=%5B%22Edge_18.17763_Windows_10.0%22,%22Safari_14.0_OSX_10.15.6%22,%22Chrome_87.0.4280.66_Windows_10.0%22,%22Firefox_84.0_Windows_10.0%22%5D&q=%22animationName%22 so this has probably never been shipping anywhere.

PR #10337, opened 2021-05-06
Add Safari 14.1 support for ::file-selector-button

lukewarlow:
Add Safari 14.1 and Safari on iOS 14.5 support for the `::file-selector-button` pseudo element.

Tested on Safari 14.1 (16611.1.21.161.6) on macOS Big Sur 11.3.1

### DIFF
--- a/css/selectors/file-selector-button.json
+++ b/css/selectors/file-selector-button.json
@@ -62,7 +62,7 @@
             ],
             "safari": [
               {
-                "version_added": false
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Thank you! 🎉

I couldn't directly test this to confirm, but seems likely based on https://bugs.webkit.org/show_bug.cgi?id=219836.

PR #10346, opened 2021-05-06
Document scroll-margin bug of firefox and firefox android

dns2utf8:
Hi all

This is my first PR to the data repo. Hope this PR is okay like this.
I tried to solve my issue https://github.com/mdn/content/issues/4770

I did build a minimal example that demonstrates the bug with both stable firefoxes:
https://data.estada.ch/firefox_scroll-margin_bug/

The corresponding bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1708303

Cheers,
Stefan

### DIFF
--- a/css/properties/scroll-margin.json
+++ b/css/properties/scroll-margin.json
@@ -21,6 +21,7 @@
               },
               {
                 "version_added": "68",
+                "version_removed": "90",
                 "partial_implementation": true,
                 "notes": "The <code>scroll-margin</code> property can cause an element's visibility to be incorrectly calculated for <code>element.focus()</code>. See <a href='https://bugzil.la/1708303'>bug 1708303</a>."
               }
--- a/css/properties/scroll-margin.json
+++ b/css/properties/scroll-margin.json
@@ -31,6 +31,7 @@
               },
               {
                 "version_added": "68",
+                "version_removed": "90",
                 "partial_implementation": true,
                 "notes": "The <code>scroll-margin</code> property can cause an element's visibility to be incorrectly calculated for <code>element.focus()</code>. See <a href='https://bugzil.la/1708303'>bug 1708303</a>."
               }
--- a/css/properties/scroll-margin.json
+++ b/css/properties/scroll-margin.json
@@ -22,7 +22,7 @@
               {
                 "version_added": "68",
                   "partial_implementation": true,
-                  "notes": "Before version 90, scroll margin is not calculated correctly on <code>element.focus()</code>, see <a href='https://bugzilla.mozilla.org/show_bug.cgi?id=1708303'>bug 1708303</a>."
+                  "notes": "Before version 90, the <code>scroll-margin</code> property can cause an element's visibility to be incorrectly calculated for <code>element.focus()</code>. See <a href='https://bugzil.la/1708303'>bug 1708303</a>."
               }
             ],
             "firefox_android": [
--- a/css/properties/scroll-margin.json
+++ b/css/properties/scroll-margin.json
@@ -32,7 +32,7 @@
               {
                 "version_added": "68",
                   "partial_implementation": true,
-                  "notes": "Before version 90, scroll margin is not calculated correctly on <code>element.focus()</code>, see <a href='https://bugzilla.mozilla.org/show_bug.cgi?id=1708303'>bug 1708303</a>."
+                  "notes": "Before version 90, the <code>scroll-margin</code> property can cause an element's visibility to be incorrectly calculated for <code>element.focus()</code>. See <a href='https://bugzil.la/1708303'>bug 1708303</a>."
               }
             ],
             "ie": {
--- a/css/properties/scroll-margin.json
+++ b/css/properties/scroll-margin.json
@@ -15,12 +15,26 @@
             "edge": {
               "version_added": "79"
             },
-            "firefox": {
-              "version_added": "68"
-            },
-            "firefox_android": {
-              "version_added": "68"
-            },
+            "firefox": [
+              {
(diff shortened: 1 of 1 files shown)


css/properties/scroll-margin.json
```diff
@@ -15,12 +15,26 @@
...
-              "version_added": "68"
-            },
+            "firefox": [
+              {
+                "version_added": "90"
+              },
+              {
+                "version_added": "68",
+                  "partial_implementation": true,
+                  "notes": "Before version 90, scroll margin is not calculated correctly on <code>element.focus()</code>, see <a href='https://bugzilla.mozilla.org/show_bug.cgi?id=1708303'>bug 1708303</a>."
```
ddbeck, line 25: A few suggestions to this note text:

- I've tried expanding the summary of the issue, to minimize the likelihood that a reader will need to read the full bug to understand what's going on here
- I've shortened the URL, to partly fix the CI failure

```suggestion
                  "notes": "Before version 90, the <code>scroll-margin</code> property can cause an element's visibility to be incorrectly calculated for <code>element.focus()</code>. See <a href='https://bugzil.la/1708303'>bug 1708303</a>."
```

That said, there's one or thing which is unresolved by my suggestion. Either we should have either:

- one support statement with `"version_added": "68"`, no `partial_implementation`, and the text "Before version 90", or
- two support statements without the text "Before version 90"

I'm inclined to go with one support statement because, if I understand correctly, this issue with `scroll-margin` only affects this one API call.

css/properties/scroll-margin.json
```diff
@@ -15,12 +15,26 @@
...
+              }
+            ],
+            "firefox_android": [
+              {
+                "version_added": "90"
+              },
+              {
+                "version_added": "68",
+                  "partial_implementation": true,
+                  "notes": "Before version 90, scroll margin is not calculated correctly on <code>element.focus()</code>, see <a href='https://bugzilla.mozilla.org/show_bug.cgi?id=1708303'>bug 1708303</a>."
```
ddbeck, line 35: Whatever we go with in the preceding case should be reflected in the data for Firefox for Android too.

css/properties/scroll-margin.json
```diff
@@ -15,12 +15,26 @@
...
-            "firefox_android": {
-              "version_added": "68"
-            },
+            "firefox": [
+              {
+                "version_added": "90"
+              },
+              {
+                "version_added": "68",
+                  "partial_implementation": true,
```
ddbeck, line 24: The indentation is off here. The easiest way to fix this is by running `npm run fix`, committing the resulting changes to the file, and pushing to your branch.

ddbeck (changes requested):
Thanks for this PR, @dns2utf8. Also I see that it's your first to the project—welcome!

I've made some suggestions in comments below. Thanks for your help!

css/properties/scroll-margin.json
dns2utf8: Thanks! :)

css/properties/scroll-margin.json
dns2utf8: Thank you for the fix.

I would go with two support statements because the fixed version don't have an annotation like we have in the safari and safari_ios entries.
Also, it keeps the UI clean on applications like https://caniuse.com/?search=scroll-margin

dns2utf8:
Thanks for the feedback :smiley:  I incorporated it.
Should I squash the commits before the merge?

css/properties/scroll-margin.json
```diff
@@ -15,12 +15,26 @@
...
-            },
-            "firefox_android": {
-              "version_added": "68"
-            },
+            "firefox": [
+              {
+                "version_added": "90"
+              },
+              {
+                "version_added": "68",
```
ddbeck, line 23: If we're going with two support statements, then this one needs to be closed out with (sorry, should've mentioned this before):

```suggestion
                "version_added": "68",
                "version_removed": "90",
```

css/properties/scroll-margin.json
```diff
@@ -15,12 +15,26 @@
...
+                "partial_implementation": true,
+                "notes": "The <code>scroll-margin</code> property can cause an element's visibility to be incorrectly calculated for <code>element.focus()</code>. See <a href='https://bugzil.la/1708303'>bug 1708303</a>."
+              }
+            ],
+            "firefox_android": [
+              {
+                "version_added": "90"
+              },
+              {
+                "version_added": "68",
```
ddbeck, line 33:
```suggestion
                "version_added": "68",
                "version_removed": "90",
```

ddbeck (changes requested):
Oops, I missed a detail on my last review. Suggestions for you below.

And you don't need to squash—we usually squash and merge on this repo anyway. Thank you!

dns2utf8:
Oh, I was under the impression that the removed was implicit.
Now it is included with the firefoxes. Should I include it with the safaris as well?

ddbeck:
> Should I include it with the safaris as well?

The Safari entries are a slightly different case. For Safari, there's two concurrent support statements: one for the standard property name and another showing support for the non-standard name, which still exists alongside the standard name.

By comparison, in Firefox, the current diff records two _consecutive_ support statements: one for versions 68-90 (with partial support) and one for version 90+ (with full support).

Does that still make sense?

dns2utf8:
Ahh, makes sense. Do you think this PR is ready for merge?

ddbeck (approved):
Yes, let's do it. Thank you, @dns2utf8! 🎉

dns2utf8:
Cool! Very happy right now 🎉 
Thank you for the reviews!

Issue #10224, opened 2021-05-03
Fixing up doc vs BCD deprecation


hamishwillee:
I've been through MDN and BCD attempting to match up deprecation info on web apis (theory being that ideally they should ideally match completely, but certainly not differ. MDN now matches BCD, where BCD has an entry.

- The first section below are cases where BCD has a different entry/inconsistency. I presume we will want to fix these. 
- The second section is cases where there is no BCD entry. Wondering whether you think this is worth documenting in BCD to make everything sync up.

FYI @ddbeck 

### Marked as deprecated in MDN but not on BCD

`Document.domain` - in discussion #10348
`SVGStylable` - under investigation on MDN https://github.com/mdn/content/pull/4782
`SVGTransformable` - #10354

#10356 - fixes
`Window: vrdisplayblur event` - marked as not deprecated (`onvrdisplayblur` marked as deprecated).
`Window: vrdisplayconnect event`  - marked as not deprecated (`onvrdisplayconnect ` marked as deprecated).
`Window: vrdisplayactivate event` -  marked as not deprecated (`onvrdisplayactivate` marked as deprecated).
`Window: vrdisplaydeactivate event`-  marked as not deprecated (`on` version marked as deprecated).
`Window: vrdisplaydisconnect event` -  marked as not deprecated (`on` version marked as deprecated).
`Window: vrdisplayfocus event` - marked as not deprecated (`on` version marked as deprecated).
`Window: vrdisplaypointerrestricted` event - marked as not deprecated (`on` version marked as deprecated).
`Window: vrdisplaypointerunrestricted` event - marked as not deprecated (`on` version marked as deprecated).
`Window: vrdisplaypresentchange event` - marked as not deprecated (`on` version marked as deprecated).
`Window: appinstalled event` - Not documented. `onappinstalled` does exist and is marked as deprecated

`WorkerGlobalScope.close()` - Not documented. `onclose` does exist and is marked as deprecated - see #10416

### Deprecated subfeatures where BCD has no record

UIEvent.cancelBubble 
- there is Event.cancelBubble and this has note that says is only supported in UIEvent on Firefox
- This is non-spec and I think perhaps should be deleted everywhere. - https://github.com/mdn/content/issues/4859

Still to look at:
- [CanvasRenderingContext2D.addHitRegion()](https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/addHitRegion) - Entry for CanvasRenderingContext2D  only
[...]

PR #10348, opened 2021-05-07
Document.domain SETTER is deprecated

hamishwillee:
The `Document.domain` setter is deprecated, as indicated in [MDN docs](https://developer.mozilla.org/en-US/docs/Web/API/Document/domain#setter) and the specs:
image

There is no way to mark just the setter deprecated, and I don't know how I might add a note to clarify this. So I have marked the whole item as deprecated. This PR is therefore obviously "for discussion."

This came out of #10224

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -3605,7 +3605,7 @@
           "status": {
             "experimental": false,
             "standard_track": true,
-            "deprecated": false
+            "deprecated": true
           }
         }
       },


foolip:
I'm skeptical about deprecating in BCD if this might lead to tooling warning against using this API, even if it's only read and not set. I'm not actually aware of any JS lint tooling that uses BCD, but it's something that could be built.

Maybe noting this in MDN is sufficient?

hamishwillee:
It would be "better" if it can be somehow recorded in BCD because it should be the canonical source of true on compatibility issues. Where I'd like to see it is as a note on deprecation or even at the "support" level. 

If that is not something that you want to support then I'd agree having it just on MDN is probably better than stating the "whole" use of the API is deprecated.

Really I just created this so the BCD team can tell me the right way to handle this :-)

foolip:
I will leave it to @ddbeck to give a more authoritative opinion :)

When the scope of an API has been reduced in the standard I think it's not super useful to point out the previously bigger shape of the API, because it's impossible to write code that works in current browsers which doesn't work in older browsers. A part of the API that doesn't exist anymore, and if it were a separate entry would be removed from BCD as irrelevant.

The other way around is more important to capture, where there are ways of using the API that won't work in slightly older browsers which are still in use.

ddbeck:
I must be a glass-half-full person because if half of something is deprecated…

OK, two major things here. First, the theoretical bit. BCD should be the canonical place to record relevant compatibility data, that is true. But the extent to which we record status information is in service to understanding the compatibility data we have. If a feature is supported in some browsers but not others, you kinda need the context provided by the statuses to understand what that means, whether a feature is emerging (or failing to), inert, or going away.

As a practical matter, we've got the problem of only half the behavior represented by the feature being deprecated. If we were serious about showing the deprecation status, we'd create getter and setter subfeatures and set their deprecation status accordingly. But I don't think we should do that, at least not yet. As far as I can tell, the _only_ difference between support for the setter and getter is the deprecation status; no browser has actually started the process of withdrawing support for the setter. Which is to say, there's no actual compatibility story to tell here, just a spec author's aspiration.

Perhaps someday, the spec will be revised to make `domain` read-only and a browser will actually introduce some different behavior. When that day comes, we'll have a concrete divergence in behavior between browsers to talk about, at which point the deprecation status will be relevant again. It will be way more specific and actionable and much clearer about what we're supposed with the data at that point.

So I'm in favor of deferring this, at least. Does that make sense?

foolip:
That make perfect sense. I didn't do any digging and assumed the setter had actually been removed from all browsers, when the opposite is true :)

hamishwillee:
Thanks guys, that all makes perfect sense.

PR #10351, opened 2021-05-07
Add Edge versions for Touch API

vinyldarkscratch:
This PR adds real values for Microsoft Edge for the `Touch` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.1).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Touch

### DIFF
--- a/api/Touch.json
+++ b/api/Touch.json
@@ -11,7 +11,7 @@
             "version_added": true
           },
           "edge": {
-            "version_added": "≤18"
+            "version_added": "79"
           },
           "firefox": [
             {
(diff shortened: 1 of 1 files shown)


foolip:
These APIs might be exposed or not depending on the hardware, I think this has to be tested on a laptop with a touchscreen to have confidence.

vinyldarkscratch:
Putting this one on the backburner for now since it doesn't contribute to decreasing `null`/`true` values.

PR #10353, opened 2021-05-07
Make JSON schema allow WebGL extension spec URLs

sideshowbarker:
We’re at the point where we need to add spec URLs for 61 WebGL features from https://www.khronos.org/registry/webgl/extensions/ WebGL extension specs.

But all https://www.khronos.org/registry/webgl/extensions/ specs lack any ID/anchors at all, and therefore don’t conform to the JSON schema requirement in the compat-data.schema.json file for `spec_url` values to have a fragment identifier. So this change updates the schema to allow any URLs matching the following pattern:

```
^https://www.khronos.org/registry/webgl/extensions/[^/]+/
```

Otherwise, without this change, CI fails on all those URLs, because doesn’t find the required fragment IDs in them.

https://github.com/mdn/browser-compat-data/pull/10358 is the related data change.


ddbeck (approved):
Thank you! 🎉

PR #10354, opened 2021-05-07
Delete SVGTransformable

hamishwillee:
[SVGTransformable](https://developer.mozilla.org/en-US/docs/Web/API/SVGTransformable) on MDN is marked as deprecated, but is not marked as deprecated on BCD. 

Looking into this, in https://www.w3.org/TR/SVG/changes.html#types it says"
image

This is not even present in the spec URL for this BCD entry https://svgwg.org/svg2-draft/types.html#InterfaceSVGGraphicsElement

Upshot, I think this is deprecated and have marked it as such. Is there anything else I need to do for this?


ddbeck:
I can't find any evidence this was ever supported, by any browser. Can we remove this instead? Is the content still needed for something?

hamishwillee:
@ddbeck 
Well, the BCD says it was supported on pre Chromium edge. If you're saying this is not true or that you don't care about that, then I personally am very happy to delete this and delete the doc on MDN. PLease confirm :-)

ddbeck:
Yeah, to clarify: I tested it on EdgeHTML and it didn't work. It's definitely not in any contemporary browser. And searching the webkit/gecko/chromium histories didn't turn up anything. Let's go ahead with the deletion instead. 👍

hamishwillee:
@ddbeck Thanks. I've deleted the SVGTransformable

ddbeck (approved):
Thank you! 🎉

PR #10356, opened 2021-05-07
Window event deprecate

hamishwillee:
This changes a number of windows events to match the deprecation status of their on_event handler, as discussed in #10224.

This includes Window event for vrdisplayblur vrdisplayconnect vrdisplayactivate vrdisplaydeactivate vrdisplaydisconnect vrdisplayfocus vrdisplaypointerrestricted vrdisplaypointerunrestricted vrdisplaypresentchange 

It also creates a deprecated appinstalled event based on onappinstalled (when this merges, need to change https://developer.mozilla.org/en-US/docs/Web/API/Window/appinstalled_event to point to this new event).

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -517,6 +517,7 @@
       },
       "appinstalled_event": {
         "__compat": {
+          "description": "<code>appinstalled</code> event",
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Window/appinstalled_event",
           "support": {
             "chrome": {
--- a/api/Window.json
+++ b/api/Window.json
@@ -10555,7 +10555,7 @@
           "status": {
             "experimental": true,
             "standard_track": true,
-            "deprecated": false
+            "deprecated": true
           }
         }
       },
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Thank you! 🎉

api/Window.json
```diff
@@ -515,6 +515,69 @@
           }
         }
       },
+      "appinstalled_event": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/Window/appinstalled_event",
```
ddbeck, line 520: Oh, wait, I spoke to soon. This is failing CI due to a lack of description.

ddbeck (changes requested):

api/Window.json
hamishwillee: @ddbeck Arrrg. Fixed and passes `yarn test`. Thanks!

ddbeck (approved):
Thank you! 🎉

PR #10359, opened 2021-05-07
Deprecate all vrdisplay* events

sideshowbarker:
The WebVR spec at https://immersive-web.github.io/webvr/spec/1.1/ is deprecated (in favor of WebXR), and all features from it are marked deprecated — except for the `vrdisplay*` events defined in the spec. So this change marks all those `vrdisplay*` events deprecated:true.

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -10211,7 +10211,7 @@
           "status": {
             "experimental": true,
             "standard_track": true,
-            "deprecated": false
+            "deprecated": true
           }
         }
       },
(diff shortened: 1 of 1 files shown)


ddbeck:
I think this might be a duplicate of https://github.com/mdn/browser-compat-data/pull/10356. Perhaps you could double check that one to confirm they touch the same features?

sideshowbarker:
> I think this might be a duplicate of #10356. Perhaps you could double check that one to confirm they touch the same features?

Thanks — yes, I’ll go ahead and close this PR; I rebased this PR branch against the #10356 branch, and that branch includes all the changes from this one.

PR #10361, opened 2021-05-07
Add Edge versions for WindowClient API

vinyldarkscratch:
This PR adds real values for Microsoft Edge for the `WindowClient` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.1).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/WindowClient

### DIFF
--- a/api/WindowClient.json
+++ b/api/WindowClient.json
@@ -11,7 +11,7 @@
             "version_added": "42"
           },
           "edge": {
-            "version_added": "≤18"
+            "version_added": "17"
           },
           "firefox": {
             "version_added": "44",
@@ -59,7 +59,7 @@
               "version_added": null
             },
             "edge": {
-              "version_added": null
+              "version_added": false
             },
             "firefox": {
               "version_added": null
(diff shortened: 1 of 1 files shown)


api/WindowClient.json
```diff
@@ -59,7 +59,7 @@
               "version_added": null
             },
             "edge": {
-              "version_added": null
+              "version_added": false
```
foolip, line 62: Does any browser support `ancestorOrigins` or should this entry be removed entirely?

vinyldarkscratch: It doesn't seem like any browsers support this, and there's no MDN web docs page for it.  Removing this entry sounds like the best way to go about it!

foolip: Do you want to send a removal PR instead then?

vinyldarkscratch: It's already been done and merged, actually!  Florian reviewed and approved #10415.

foolip: Oh wait, that wasn't the only change :)

foolip (approved):

PR #10371, opened 2021-05-07
Mark css.properties.background-image.image-rect non-standard

sideshowbarker:
https://drafts.csswg.org/css-backgrounds/ defines no `image-rect` value for the `background-image` property. The mdn_url in BCD points to https://developer.mozilla.org/en-US/docs/Web/CSS/-moz-image-rect, which has a **Non-standard** banner.

### DIFF
--- a/css/properties/background-image.json
+++ b/css/properties/background-image.json
@@ -209,7 +209,7 @@
             },
             "status": {
               "experimental": false,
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Looks good. Thank you! 🎉

PR #10372, opened 2021-05-07
Add WebGLRenderingContext spec URLs

sideshowbarker:
Heads-up: This one adds 160 spec URLs. And in general the WebGL spec lacks anchors for individual members, so (with some exceptions) the URLs point to spec subsection headings. Also, there are a number of features with multiple spec URLs — due to the fact that the WebGL2 spec and some extension specs add members to objects or extend behavior of some.

### DIFF
--- a/api/WebGLRenderingContext.json
+++ b/api/WebGLRenderingContext.json
@@ -3,6 +3,7 @@
     "WebGLRenderingContext": {
       "__compat": {
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/WebGLRenderingContext",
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
Looks good to me. I spot checked quite a few links but didn't verify all chapter numbers. 

On MDN, it will also display the chapter numbers, but we can see what to do about that in the future (Dom already suggested to use webref to get better section links).

The MDN pages (which I authored btw) also reference the OpenGL specs, but looking at them now I think I'm fine with not linking to OpenGL anymore.

Great work, thanks!

PR #10379, opened 2021-05-07
Add IE versions for ErrorEvent API

vinyldarkscratch:
This PR adds real values for Internet Explorer and Edge for the `ErrorEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.2).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/ErrorEvent

### DIFF
--- a/api/ErrorEvent.json
+++ b/api/ErrorEvent.json
@@ -21,7 +21,7 @@
             "version_added": true
           },
           "ie": {
-            "version_added": true
+            "version_added": "10"
           },
           "opera": {
             "version_added": "11"
@@ -69,7 +69,7 @@
               "version_added": null
             },
             "ie": {
-              "version_added": null
+              "version_added": false
             },
             "opera": {
               "version_added": true
@@ -117,7 +117,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": null
+              "version_added": "10"
             },
             "opera": {
               "version_added": true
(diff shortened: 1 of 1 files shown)


api/ErrorEvent.json
```diff
@@ -21,7 +21,7 @@
             "version_added": true
           },
           "ie": {
-            "version_added": true
+            "version_added": "10"
```
foolip, line 24: This is the event interface for "error" events fired at the window object for errors in script parsing/execution. https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onerror#browser_compatibility says that `onerror` was supported since IE9, so this needs double checking.

vinyldarkscratch: `onerror` seems to simply return a string, not an `ErrorEvent` instance, so I'm not sure how to test this one...

foolip: That string is probably the event handler code serialized. When the event is actually fired, what does the event object look like?

Making a working demo in a modern browser might be a good first step, so you're sure the problem is with IE if it still doesn't work.

vinyldarkscratch: Using the same test code as in #10844 (AKA, `addEventListener`), I've determined that the error event handler throws a regular event in IE 9, but starts throwing an `ErrorEvent` in IE 10, so this checks out!

foolip: Excellent, thanks for testing!

foolip (approved):

PR #10380, opened 2021-05-07
Add IE versions for HTMLInputElement API

vinyldarkscratch:
This PR adds real values for Internet Explorer and Edge for the `HTMLInputElement` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.2).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/HTMLInputElement.json
+++ b/api/HTMLInputElement.json
@@ -780,7 +780,7 @@
               "version_added": "4"
             },
             "ie": {
-              "version_added": true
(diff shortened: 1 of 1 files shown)


foolip (approved):
This matches the api.HTMLButtonElement.formEnctype and related "formenctype" entry for the content attributes, so is probably correct.

Note that a "formenctype" entry in input.json is missing.

PR #10381, opened 2021-05-07
Add IE versions for KeyboardEvent API

vinyldarkscratch:
This PR adds real values for Internet Explorer and Edge for the `KeyboardEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.2).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/KeyboardEvent

### DIFF
--- a/api/KeyboardEvent.json
+++ b/api/KeyboardEvent.json
@@ -167,7 +167,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": true
+              "version_added": "9"
             },
             "opera": {
               "version_added": true
@@ -314,7 +314,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": true
+              "version_added": "9"
             },
             "opera": {
               "version_added": true
@@ -1581,7 +1581,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": true
+              "version_added": "9"
             },
             "opera": {
               "version_added": true
@@ -1679,7 +1679,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": true
+              "version_added": "9"
             },
             "opera": {
               "version_added": true


foolip:
This seems suspicious. Keyboard events of some kind must have been supported since IE 4 and maybe earlier. Can you do some digging?

vinyldarkscratch:
Running a little test code (see below), it isn't until IE 9 that I see the `event` argument to `onkeydown` being an event, let alone a `KeyboardEvent`.  In IE 8 and below, it is `undefined`.

```js
document.onkeydown = function(event) {
  alert(event);
}
```

foolip:
@vinyldarkscratch if you omit the `event` argument, is `alert(event)` still undefined? IE4 and later had `window.event` (per BCD) and maybe the argument just shadows that.

Alternatively, you could try the `addEventListener`. It looks like the event is fired, just need to see what the event object looks like...

vinyldarkscratch:
It appears that in IE 10 and below, `window.event` is some sort of proprietary `MSEventObj` (but then a regular `Event` in IE 8), of which I can't get any of the `KeyboardEvent` properties.  Additionally, it seems that the `keydown` event is only supported in IE 9 and above, despite the fact I'm also using `attachEvent` for IE 8 and lower, and that `onkeydown` is present in earlier IE versions.

foolip:
Thanks @vinyldarkscratch! If we have a lot of event data to work through, and if some big events rewrite was in IE9, maybe we should go with a <=IE9 range and not bother testing earlier. Or just <=IE11 would be fine honestly.

vinyldarkscratch:
Is there anything I need to do to get this PR ready for merging?

foolip:
I had suggested a <=9 range, but given that I just merged the MouseEvent PR this seems very plausible too. Nobody will care if it's right anyway, who's supporting IE8 now?

foolip (approved):

PR #10382, opened 2021-05-07
Add IE versions for MouseEvent API

vinyldarkscratch:
This PR adds real values for Internet Explorer and Edge for the `MouseEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.2).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/MouseEvent.json
+++ b/api/MouseEvent.json
@@ -235,7 +235,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": true
+              "version_added": "9"
             },
             "opera": {
               "version_added": true
@@ -577,7 +577,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": true
+              "version_added": "9"
             },
             "opera": {
               "version_added": true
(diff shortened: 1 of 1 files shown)


foolip:
Like in https://github.com/mdn/browser-compat-data/pull/10381 this seems later than plausible.

vinyldarkscratch:
Looks like this is the same case as #10381: the event handler's first argument is `undefined` until IE 9.

```js
document.onclick = function(event) {
  alert(event);
}
```

foolip:
Suggesting the same `window.event` or `addEventListener` workaround as there :)

vinyldarkscratch:
Looks like it's the _exact_ same case as in #10381 -- `window.event` is a proprietary event object with none of the properties for `MouseEvent`, `addEventListener` (`attachEvent`) isn't firing anything in IE 8 or lower, etc.

foolip (approved):

foolip:
Thanks for confirming! Things were weirder than I remember before IE 9 :)

PR #10386, opened 2021-05-07
Add IE versions for PerformanceEntry API

vinyldarkscratch:
This PR adds real values for Internet Explorer and Edge for the `PerformanceEntry` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.2).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/PerformanceEntry.json
+++ b/api/PerformanceEntry.json
@@ -33,7 +33,7 @@
             "version_added": "25"
           },
           "ie": {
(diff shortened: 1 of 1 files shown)


foolip (approved):
Looks very plausible in light of https://developer.mozilla.org/en-US/docs/Web/API/Performance#browser_compatibility

PR #10387, opened 2021-05-07
Add IE versions for Point API

vinyldarkscratch:
This PR adds real values for Internet Explorer and Edge for the `Point` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.2).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Point

### DIFF
--- a/api/Point.json
+++ b/api/Point.json
@@ -20,7 +20,7 @@
             "version_added": false
           },
           "ie": {
-            "version_added": null
+            "version_added": false
           },
           "opera": {
             "version_added": null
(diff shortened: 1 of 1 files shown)


foolip:
I think we should instead delete this entry by folding it into `DOMPoint` as an alias where appropriate. I suspect the Edge data is wrong BTW, so needs double checking.

vinyldarkscratch:
This PR is no longer needed since the data has been set to `false` in another PR.  Closing.

PR #10389, opened 2021-05-07
Add IE versions for SVGExternalResourcesRequired API

vinyldarkscratch:
This PR adds real values for Internet Explorer and Edge for the `SVGExternalResourcesRequired` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.2).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/SVGExternalResourcesRequired

### DIFF
--- a/api/SVGExternalResourcesRequired.json
+++ b/api/SVGExternalResourcesRequired.json
@@ -21,7 +21,7 @@
             "version_added": false
           },
           "ie": {
-            "version_added": null
+            "version_added": false
           },
           "opera": {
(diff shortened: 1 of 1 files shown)


foolip:
This interface is gone in SVG 2, but I think it was originally mixin-like, see https://github.com/mdn/browser-compat-data/issues/7752.

A PR to remove the entry entirely would be great.

PR #10392, opened 2021-05-07
Add IE versions for SVGStylable API

vinyldarkscratch:
This PR adds real values for Internet Explorer and Edge for the `SVGStylable` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.2).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/SVGStylable.json
+++ b/api/SVGStylable.json
@@ -23,7 +23,7 @@
             "version_removed": "20"
           },
           "ie": {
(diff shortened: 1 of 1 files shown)


foolip:
I've sent https://github.com/mdn/browser-compat-data/pull/10536 to remove this entry instead.

PR #10397, opened 2021-05-07
Add IE versions for TrackEvent API

vinyldarkscratch:
This PR adds real values for Internet Explorer and Edge for the `TrackEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.2).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/TrackEvent

### DIFF
--- a/api/TrackEvent.json
+++ b/api/TrackEvent.json
@@ -21,7 +21,7 @@
             "version_added": null
           },
           "ie": {
-            "version_added": null
+            "version_added": "10"
           },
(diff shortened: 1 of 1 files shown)


foolip (approved):
Matches https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/textTracks#browser_compatibility

PR #10398, opened 2021-05-07
Add IE versions for TransitionEvent API

vinyldarkscratch:
This PR adds real values for Internet Explorer and Edge for the `TransitionEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.2).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/TransitionEvent

### DIFF
--- a/api/TransitionEvent.json
+++ b/api/TransitionEvent.json
@@ -253,7 +253,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": null
+              "version_added": "10"
             },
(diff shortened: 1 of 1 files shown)


foolip (approved):
Matches the rest of https://developer.mozilla.org/en-US/docs/Web/API/TransitionEvent#browser_compatibility

PR #10401, opened 2021-05-07
Add IE versions for WorkerNavigator API

vinyldarkscratch:
This PR adds real values for Internet Explorer for the `WorkerNavigator` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.2).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/WorkerNavigator.json
+++ b/api/WorkerNavigator.json
@@ -20,7 +20,7 @@
             "version_added": "4"
           },
           "ie": {
-            "version_added": null
+            "version_added": "10"
           },
           "opera": {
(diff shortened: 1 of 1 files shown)


foolip (commented):
If this right then Edge should also be 12, can you update?

vinyldarkscratch:
I've submitted an earlier PR which includes updates to the Edge versions (see #10363).  I can combine these two together if desired?

foolip:
Ah, no need, I'll just review and merge them at the same time.

PR #10409, opened 2021-05-08
Mark "URL scheme-aware (schemeful) SameSite cookies" supported in Chrome 89

### LINKED ISSUE #10271: http.headers.Set-Cookie - Scheme-ful SameSite changed in Chrome 89
For Chrome, the URL scheme-aware ("schemeful") feature is listed as "disabled, hidden behind flag".

This feature was enabled by default in Chrome 89. The table should read "Full Support" as of 89.

Source: https://www.chromestatus.com/feature/5096179480133632

<details>

* Query: `http.headers.Set-Cookie`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Set-Cookie/SameSite
* Report started: 2021-05-04T02:18:01.796Z

</details>

soren121:
Fixes #10271. 

URL scheme-aware (schemeful) SameSite support was enabled by default in Chrome 89 and Chrome for Android 89, per Chrome Platform Status: https://www.chromestatus.com/feature/5096179480133632

I wasn't sure if I should keep the support statements for Chrome 86 when it was available via a flag, but it seems like that's unnecessary. Please let me know if I'm wrong and I'll fix it.

### DIFF
--- a/http/headers/set-cookie.json
+++ b/http/headers/set-cookie.json
@@ -418,26 +418,36 @@
             "__compat": {
               "description": "URL scheme-aware (\"schemeful\")",
               "support": {
-                "chrome": {
-                  "version_added": "86",
-                  "flags": [
-                    {
-                      "type": "preference",
-                      "name": "#schemeful-same-site",
-                      "value_to_set": "Enabled"
-                    }
-                  ]
-                },
-                "chrome_android": {
-                  "version_added": "86",
-                  "flags": [
-                    {
-                      "type": "preference",
-                      "name": "#schemeful-same-site",
-                      "value_to_set": "Enabled"
-                    }
-                  ]
-                },
+                "chrome": [
+                  {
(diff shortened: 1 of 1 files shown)


ddbeck (changes requested):
Thanks for opening this PR, @soren121! Also, I see that it's your first on the project—welcome!

I would prefer it if you would reinstate the flag data. You can have an array of two support objects, the first one without the flag, the second one with the flag. We have [a guideline for removing old flag data](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-flag-data) (typically, two years after the flag has been enabled by default), so it's too soon to remove it now. Thank you again!

soren121:
Ah, excellent! I was looking for that sort of guidance re: flags, but wasn't able to find it earlier. I've added the flag statements back in.

PR #10416, opened 2021-05-10
WorkerGlobalScope - add close() and close_event to match onclose()

hamishwillee:
This is part of fixing #10224

`WorkerGlobalScope` has the (deprecated) `onclose()` event handler property, but not the associated method `close()` and event `close_event`. This copies the event hanlder for the other two items. This then ensures that BCD matches MDN for this property.


ddbeck (approved):
Thank you! 🎉

PR #10418, opened 2021-05-10
Mark several Permission Policy features non-standard

sideshowbarker:
There is no spec for the following policy-controlled features:

* layout-animations
* legacy-image-formats
* oversized-images
* unoptimized-images
* unsized-media

So this change marks them all standard_track:false.

Related MDN change: https://github.com/mdn/content/pull/4872
Related MDN change: https://github.com/mdn/content/pull/4899

### DIFF
--- a/http/headers/feature-policy.json
+++ b/http/headers/feature-policy.json
@@ -859,7 +859,7 @@
             },
             "status": {
               "experimental": true,
-              "standard_track": true,
+              "standard_track": false,
               "deprecated": false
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Thanks, @sideshowbarker. This is consistent with what I learned in the process of https://github.com/mdn/browser-compat-data/pull/9707. (I expect to clean out the all-`false`, non-standard features soon.)

PR #10420, opened 2021-05-10
Update BCD to reflect that top level await is enabled starting with FF 89

mirunacurtean:
- [x] Summarize your changes

Updated BCD table to reflect: 
- [top level awai](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Operators/await)t is enabled starting with FF 89. 
- Update experimental feature status to False
- update FF and FF for Android columns to show that the feature was added in FF version 89 and removed previous notes. 

https://bugzilla.mozilla.org/show_bug.cgi?id=1681046

Fixes https://github.com/mdn/content/issues/4313


yarn.lock
```diff
@@ -0,0 +1,903 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
ddbeck, line 1: Delete me 🗑️

javascript/operators/await.json
```diff
@@ -79,24 +79,10 @@
...
               "firefox": {
-                "version_added": false,
-                "notes": "Available only in nightly builds. See <a href='https://bugzil.la/1681046'>bug 1681046</a>.",
-                "flags": [
-                  {
-                    "type": "preference",
-                    "name": "javascript.options.experimental.top_level_await"
-                  }
-                ]
+                "version_added": "89"
```
ddbeck, line 82: Nice cleanup! Ordinarily, I'd suggest keeping old prefs and adding new support statements, but this is a weird one with Nightly-only.

ddbeck (changes requested):
Looks good, @mirunacurtean. Just need the `yarn.lock` file removed. Thank you!

yarn.lock
mirunacurtean: Seems like the file is autogenerated and it gets pushed along with all the other changes despite not being staged. 
I will pay attention to deleting it before making any commits from now on.

ddbeck: As an alternative, you can ignore the file on your local clone. To do that, add `yarn.lock` on a new line in `.git/info/exclude` (i.e., from the root of the repository, run `echo 'yarn.lock' >> .git/info/exclude`).

ddbeck (approved):
Thank you! 🎉

PR #10423, opened 2021-05-10
Modify PerformanceEventTiming table to show it is enabled on FF 89

mirunacurtean:
- [x] Summarize your changes

- removed experimental status
- Updated FF and FF for Android columns to show that feature has been enabled in 89th version.  

[We have PerformanceEventTiming enabled in Nightly for over a month now...](https://bugzilla.mozilla.org/show_bug.cgi?id=1701029#c0)

Fixes https://github.com/mdn/content/issues/4309


ddbeck:
Hey @mirunacurtean, you've updated the main interface here, but none of its members. Is that right? Or is there additional work to be done on the subfeatures here?

mirunacurtean:
> Hey @mirunacurtean, you've updated the main interface here, but none of its members. Is that right? Or is there additional work to be done on the subfeatures here?

@ddbeck  Yes, I think so; if you're referring to only the PerformanceEventTiming being affected by these changes and yes it was only what was in the UI that should've been modified.  If I misunderstand something, we can chat over it on slack.

mirunacurtean:
@ddbeck This is now ready for another review. I've updated all the other subfeature rows and deleted the yarn file. 
So that the consistency is maintained I've also added an MDN URL to the .target attribute, even if there is no page created for it yet.

ddbeck (approved):
Looks great, @mirunacurtean. Thank you! 🎉

PR #10441, opened 2021-05-11
Fix Safari data for text-decoration-skip CSS property

### LINKED ISSUE #6211: BCD for text-decoration-skip
In https://developer.mozilla.org/en-US/docs/Web/CSS/text-decoration-skip, the BCD says that Safari only supports the none and skip, but skip is not a value.

vinyldarkscratch:
This PR resolves a small issue with the notes for Safari.  It mentions that the property only supports the `none` and `skip` values, yet `skip` is not a value, either in Safari or the spec.  I believe the note author meant `auto`, which is included in the spec.

Fixes #6211.

### DIFF
--- a/css/properties/text-decoration-skip.json
+++ b/css/properties/text-decoration-skip.json
@@ -41,23 +41,23 @@
             "safari": [
               {
                 "version_added": "12.1",
-                "notes": "Only supports the <code>none</code> and <code>skip</code> values; all other values behave like those two values."
+                "notes": "Only supports the <code>none</code> and <code>auto</code> values; all other values behave like those two values."
               },
               {
                 "version_added": "8",
                 "prefix": "-webkit-",
-                "notes": "Only supports the <code>none</code> and <code>skip</code> values; all other values behave like those two values."
+                "notes": "Only supports the <code>none</code> and <code>auto</code> values; all other values behave like those two values."
               }
             ],
             "safari_ios": [
               {
                 "version_added": "12.2",
-                "notes": "Only supports the <code>none</code> and <code>skip</code> values; all other values behave like those two values."
+                "notes": "Only supports the <code>none</code> and <code>auto</code> values; all other values behave like those two values."
               },
               {
                 "version_added": "8",
                 "prefix": "-webkit-",
-                "notes": "Only supports the <code>none</code> and <code>skip</code> values; all other values behave like those two values."
+                "notes": "Only supports the <code>none</code> and <code>auto</code> values; all other values behave like those two values."
               }
             ],
             "samsunginternet_android": {
--- a/css/properties/text-decoration-skip.json
+++ b/css/properties/text-decoration-skip.json
@@ -41,23 +41,23 @@
             "safari": [
               {
                 "version_added": "12.1",
-                "notes": "Only supports the <code>none</code> and <code>auto</code> values; all other values behave like those two values."
+                "notes": "Supports only <code>none</code>, <code>auto</code>, and <code>objects</code> values."
               },
               {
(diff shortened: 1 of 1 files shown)


css/properties/text-decoration-skip.json
```diff
@@ -41,23 +41,23 @@
             "safari": [
               {
                 "version_added": "12.1",
-                "notes": "Only supports the <code>none</code> and <code>skip</code> values; all other values behave like those two values."
+                "notes": "Only supports the <code>none</code> and <code>auto</code> values; all other values behave like those two values."
```
ddbeck, line 44: OK, this was a bit of a time waster, but I wondered what this note actually meant. All other values behave _like_ those two values? What?

Turns out, this data is pretty rough. It's from the old wiki tables and it shows. I tested this and found that an additional value is supported and the version numbers are wrong. I think the note ought to read:

```suggestion
                "notes": "Supports only <code>none</code>, <code>auto</code>, and <code>objects</code> values."
```

Also the versions supported are wrong for the prefixed version. The prefixed version appears to have shipped in 6.1 or 6.2, so based on the discussion in https://github.com/mdn/browser-compat-data/issues/9423, I'd set that to 7.

css/properties/text-decoration-skip.json
```diff
@@ -41,23 +41,23 @@
...
               },
               {
                 "version_added": "8",
```
ddbeck, line 47:
```suggestion
                "version_added": "7",
```

css/properties/text-decoration-skip.json
```diff
@@ -41,23 +41,23 @@
...
                 "prefix": "-webkit-",
-                "notes": "Only supports the <code>none</code> and <code>skip</code> values; all other values behave like those two values."
+                "notes": "Only supports the <code>none</code> and <code>auto</code> values; all other values behave like those two values."
```
ddbeck, line 49:
```suggestion
                "notes": "Supports only <code>none</code>, <code>auto</code>, and <code>objects</code> values."
```

css/properties/text-decoration-skip.json
```diff
@@ -41,23 +41,23 @@
...
               }
             ],
             "safari_ios": [
               {
                 "version_added": "12.2",
-                "notes": "Only supports the <code>none</code> and <code>skip</code> values; all other values behave like those two values."
+                "notes": "Only supports the <code>none</code> and <code>auto</code> values; all other values behave like those two values."
```
ddbeck, line 55:
```suggestion
                "notes": "Supports only <code>none</code>, <code>auto</code>, and <code>objects</code> values."
```

css/properties/text-decoration-skip.json
```diff
@@ -41,23 +41,23 @@
...
               },
               {
                 "version_added": "8",
```
ddbeck, line 58:
```suggestion
                "version_added": "7",
```

css/properties/text-decoration-skip.json
```diff
@@ -41,23 +41,23 @@
...
                 "prefix": "-webkit-",
-                "notes": "Only supports the <code>none</code> and <code>skip</code> values; all other values behave like those two values."
+                "notes": "Only supports the <code>none</code> and <code>auto</code> values; all other values behave like those two values."
```
ddbeck, line 60:
```suggestion
                "notes": "Supports only <code>none</code>, <code>auto</code>, and <code>objects</code> values."
```

ddbeck (changes requested):

ddbeck (approved):
Thank you! 🎉

PR #10442, opened 2021-05-11
Include CanIUse Embed as a project using the data

vinyldarkscratch:
Since the inclusion of MDN BCD content directly in CanIUse, projects that are using CanIUse data are also able to use BCD data, including the CanIUse embed by @ireade.  Ire has also made some tweaks so that any BCD data will display with a blue border (vs. an orange one) and state the data comes from BCD.  This PR adds a link back to their tool in the "projects using the data" section of the Readme, to help users interested in embedding BCD find a tool much quicker (and as a thanks to the tool's author).


foolip (approved):
I didn't know about this project, so linking it is useful! I sent https://github.com/ireade/caniuse-embed/pull/28 to tweak how BCD is linked from those embeds.

ddbeck (approved):
Looks good. Thank you! 🎉

PR #10449, opened 2021-05-12
Remove window.returnValue

sideshowbarker:
All support values for the feature are either “false” or “null”, it’s not defined in any current spec, and it has no MDN article (the mdn_url https://developer.mozilla.org/docs/Web/API/Window/returnValue is 404).  So this change removes it from BCD.

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -7471,54 +7471,6 @@
           }
         }
       },
-      "returnValue": {
-        "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/Window/returnValue",
-          "support": {
-            "chrome": {
-              "version_added": false
-            },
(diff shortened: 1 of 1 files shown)


foolip (approved):
This must have been a conflation with https://developer.mozilla.org/en-US/docs/Web/API/Event/returnValue at some point.

sideshowbarker:
> This must have been a conflation with [developer.mozilla.org/en-US/docs/Web/API/Event/returnValue](https://developer.mozilla.org/en-US/docs/Web/API/Event/returnValue) at some point.

Either that or some confusion related to `showModalDialog`. See https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/platform-apis/ms534371(v=vs.85)

PR #10458, opened 2021-05-12
Fix the order of v0.0.5x in release notes

vinyldarkscratch:
When going through the release notes document, I noticed that the order of some of the releases was reversed -- oops!  This PR fixes the sections of release notes for v0.0.5x to maintain reverse-chronological order throughout the file.


ddbeck (approved):
That'll teach me to trust the ordering of the GitHub releases API. Thanks, @vinyldarkscratch!

PR #10484, opened 2021-05-14
Spec URLs for entries/keys/values/forEach: reference WebIDL

sideshowbarker:
This change makes the data for entries/keys/values/forEach members `iterable`-annotated interfaces use the spec URL https://heycam.github.io/webidl/#idl-iterable

---

Without this change, entries/keys/values/forEach members of these interfaces either have no spec URLs at all, or were set to URL values for the anchors/fragment IDs of the interfaces themselves. But a developer following a link from MDN to the anchor for an interface is likely to be confused, because they won’t see anything there about entries/keys/values/forEach members.

But if instead direct developers to https://heycam.github.io/webidl/#idl-iterable, they will see something useful there.

### DIFF
--- a/api/CSSTransformValue.json
+++ b/api/CSSTransformValue.json
@@ -100,7 +100,7 @@
       "entries": {
         "__compat": {
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/CSSTransformValue/entries",
-          "spec_url": "https://drafts.css-houdini.org/css-typed-om/#transformvalue-objects",
+          "spec_url": "https://heycam.github.io/webidl/#idl-iterable",
           "support": {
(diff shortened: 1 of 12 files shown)


Elchi3 (approved):
Makes sense to me. 
Related is https://github.com/mdn/browser-compat-data/issues/6367 which hopefully makes this more consistent. I wonder if every API really needs to document entries/forEach/key/values etc. but we will see what we'll agree on.

sideshowbarker:
> I wonder if every API really needs to document entries/forEach/key/values etc. but we will see what we'll agree on.

Yeah this may be case where we could figure out if there’s some redundancy we could reduce.

PR #10493, opened 2021-05-15
api.PresentationRequest.startWithDevice is not supported in Chrome

vinyldarkscratch:
While working on #10254, @foolip noticed that Chrome didn't support the `startWithDevice` feature either, leaving question as to whether the data was accurate.  Doing some more digging into the issue, I found that Chrome never actually supported this property to begin with.  Considering the original data came from the wiki migration, I question its accuracy.


foolip (approved):
I can confirm there is no `startWithDevice` method in Chromium source.

PR #10515, opened 2021-05-18
Remove unimplemented SVG color-profile & solidcolor elements

sideshowbarker:
The SVG `color-profile` and `solidcolor` elements have never been implemented in any browser engines, so this change removes the data for the elements as well as the data for their associated APIs. Related: https://github.com/mdn/browser-compat-data/issues/6768


nutza978:
@apple-web-evangelist

sideshowbarker:
@apple-web-evangelist On the off chance that @nutza978 isn’t just some junk account posting random comment spam and that instead there’s some question about whether these proposed removals would be dropping data for anything actually implemented in WebKit: I checked he WebKit sources before raising this, and confirmed there’s no current code in WebKit for SVGColorProfileElement or SVGSolidcolorElement — and no corresponding code for the elements.

We have a related issue at https://github.com/mdn/browser-compat-data/issues/6768 which initially included a proposal to drop data from BCD for 17 SVG elements/APIs due to them not being implemented in any engines. But from subsequent investigation and testing it turned out that 15 of those had actually been implemented in WebKit (and only WebKit).

SVGColorProfileElement and SVGSolidcolorElement are the only ones from that initial https://github.com/mdn/browser-compat-data/issues/6768 list that are in fact completely unimplemented.

saschanaz:
> On the off chance that @nutza978 isn’t just some junk account

I highly suspect it is per #10532 🤔

ddbeck (approved):
This seems fine. Thank you! 🎉 

(The spam accounts thing is getting to be quite annoying. I'm seriously considering turning on the temporary interaction limits for new users.)

ddbeck:
Opened https://github.com/mdn/content/pull/5147 for corresponding content removal.

sideshowbarker:
> (The spam accounts thing is getting to be quite annoying. I'm seriously considering turning on the temporary interaction limits for new users.)

We could also consider blocking the offending accounts. That’d at least prevent repeat offenses (such at the case with the account that posted to this issue). But that said, I think we can’t do blocking per-repo but instead need to block accounts per-org. So that’d mean we’d need somebody with admin perms for the @mdn org to add the accounts to the org-wide block list.

https://github.com/organizations/mdn/settings/user_blocks is where the UI for it would be (for somebody with the right perms)

ddbeck:
@sideshowbarker Yes, we could use this more aggressively. I've asked for repeat offenders to be blocked from time to time, but I don't have the privileges to do it myself. I'll look into this more.

PR #10522, opened 2021-05-19
Remove api.SVGURIReference entry

foolip:
This is a mixin in SVG 2:
https://svgwg.org/svg2-draft/types.html#InterfaceSVGURIReference

However, the href member isn't represented in BCD, so this entry doesn't
really convey anything at all.

Part of https://github.com/mdn/browser-compat-data/issues/7752.


Elchi3 (approved):
Thanks! 🎉

PR #10523, opened 2021-05-19
Add Android WebView engine versions for 1-1.6

vinyldarkscratch:
This PR adds the WebKit versions for Android WebView for 1.0 through 1.6.  These were determined based upon the user agent strings from corresponding Android emulators (totally not foreshadowing here >:D).


foolip (approved):
Awesome, thank you!

PR #10529, opened 2021-05-19
Update to node 16 and fs.rmdir to fs.rm because with the flag recursi…

dns2utf8:
#### Summary
Update to node 16 and `fs.rmdir` to `fs.rm` because with the flag recursive the call has been deprecated.

Node 16 will become the LTS on 2021-10-26 so the release will be around and maintained until 2024 atleast.

I ran the following commands and everything worked:
```bash
npm install
npm test
npm run release-build
```

#### Related issues

Builds on #9877

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -17,7 +17,7 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-node@v2
       with:
-        node-version: 14.x
+        node-version: 16.x
     - run: npm install
     - run: npm test
       env:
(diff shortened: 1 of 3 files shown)


ddbeck:
Thanks for this PR, @dns2utf8.

> Node 16 will become the LTS on 2021-10-26 so the release will be around and maintained until 2024 atleast.

The consensus on this project has been to use the Active LTS release (see https://github.com/mdn/browser-compat-data/pull/10511 for the most recent citation of this, though I'm certain we've discussed it before). I'm happy to adopt Node 16 in October, but not until then.

I take it the main objective here to suppress a deprecation warning? If so, I _think_ this might work with only the `rm` change under Node 14, but I'd welcome a test to find out. Otherwise, I think the only route is to close this and revisit it in late October.

ddbeck:
Since there's been no follow up on this, I think it's time to close this PR. Thanks for the proposal, @dns2utf8, and feel free to revisit this question, if you have something to add.

PR #10534, opened 2021-05-19
Update Safari data for Blob.slice

apple-web-evangelist:
#### Summary
Provides Safari prefixed/unprefixed versions for `blob.slice`

Reviewed WebKit code available from https://opensource.apple.com


foolip (approved):
Thanks!

PR #10537, opened 2021-05-20
Drop SVGStylable as unimplemented

sideshowbarker:
The `SVGStylable` interface as such has had no shipping implementation for years. The current BCD data indicates it was previously in Firefox but removed years ago, and was (maybe) part of (pre-Blink) Edge. It was also apparently once part of Webkit, but in 2013 was removed in https://trac.webkit.org/changeset/140265/webkit, with change message:

> SVG2 changes element inheritance so all SVGElements are stylable. As a first-step towards a cleaner class hierarchy, this patch moves SVGStylable into SVGStyledElement.

https://developer.mozilla.org/en-US/docs/Web/API/SVGStylable already has it marked as deprecated.

Related: https://github.com/mdn/browser-compat-data/issues/6768


sideshowbarker:
Duplicate of https://github.com/mdn/browser-compat-data/pull/10536

PR #10542, opened 2021-05-20
-webkit-overflow-scrolling no longer supported in iOS 13+

fabb:
#### Summary
-webkit-overflow-scrolling is no longer supported on iOS 13+

Tested on iPhone with iOS 14 + Safari 14.

https://developer.apple.com/documentation/safari-release-notes/safari-13-release-notes
"Added support for one-finger accelerated scrolling to all frames and overflow:scroll elements eliminating the need to set-webkit-overflow-scrolling: touch."

### DIFF
--- a/css/properties/background-attachment.json
+++ b/css/properties/background-attachment.json
@@ -77,11 +77,21 @@
               "opera_android": {
                 "version_added": "14"
               },
-              "safari": {
-                "version_added": false
-              },
+              "safari": [
+                {
+                  "version_added": "14",
+                  "partial_implementation": true,
+                  "notes": "<code>local</code> is recognized but has no effect due to [a bug](https://bugs.webkit.org/show_bug.cgi?id=219324)."
+                },
+                {
+                  "version_added": "3.1",
+                  "version_removed": "14"
+                }
+              ],
               "safari_ios": {
-                "version_added": "4.2"
+                "version_added": "5",
+                "partial_implementation": true,
+                "notes": "<code>local</code> is recognized but has no effect."
               },
               "samsunginternet_android": {
                 "version_added": "1.0"
@@ -125,12 +135,29 @@
               "opera_android": {
                 "version_added": "14"
               },
-              "safari": {
-                "version_added": false
-              },
-              "safari_ios": {
-                "version_added": "4.2"
-              },
+              "safari": [
+                {
+                  "version_added": "13",
+                  "partial_implementation": true,
+                  "notes": "<code>local</code> is recognized but has no effect due to [a bug](https://bugs.webkit.org/show_bug.cgi?id=219324)."
+                },
+                {
+                  "version_added": "5",
+                  "version_removed": "13"
+                }
+              ],
+              "safari_ios": [
+                {
+                  "version_added": "13",
+                  "partial_implementation": true,
+                  "notes": "<code>local</code> is recognized but has no effect due to [a bug](https://bugs.webkit.org/show_bug.cgi?id=219324)."
+                },
+                {
+                  "version_added": "4.2",
+                  "version_removed": "13",
+                  "notes": "<code>local</code> is only supported when <code>-webkit-overflow-scrolling: touch</code> is NOT used."
+                }
+              ],
               "samsunginternet_android": {
                 "version_added": "1.0"
               },
--- a/css/properties/background-attachment.json
+++ b/css/properties/background-attachment.json
@@ -81,7 +81,7 @@
                 {
                   "version_added": "14",
                   "partial_implementation": true,
-                  "notes": "<code>local</code> is recognized but has no effect due to [a bug](https://bugs.webkit.org/show_bug.cgi?id=219324)."
+                  "notes": "<code>local</code> is recognized but has no effect due to <a href='https://webkit.org/b/219324'>a bug</a>."
                 },
                 {
                   "version_added": "3.1",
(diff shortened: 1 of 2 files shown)


css/properties/background-attachment.json
```diff
@@ -80,7 +80,8 @@
                 "version_added": "5"
               },
               "safari_ios": {
-                "version_added": "4.2"
+                "version_added": "4.2",
+                "version_removed": "13"
```
ddbeck, line 84: This will need to become an array of support statements. Something like:

```
[
  {
    "version_added": "13",
    "notes": "<code>local</code> is recognized but has no effect."
  },
  {
    "version_added": "4.2",
    "version_removed": "13"
  }
]
```

ddbeck (changes requested):
Thanks for this PR, @fabb. Some interesting details here; see a line comment for the actual change that needs to happen here.

## `-webkit-overflow-scrolling`

For posterity, I wanted to note that this change would appear to go against a data guideline, but for good reason.

Safari still seems to continue to "support" the property in the sense that it's a valid property name (e.g., the inspector doesn't flag it and `CSS.supports()` returns `true` for it), but it doesn't _do_ anything. Ordinarily, this would mean [setting `partial_implementation`](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#non-functional-defined-names-imply-partial_implementation). But since no other browser supports this and it doesn't seem to have a standard equivalent to interact poorly with in the cascade, I think it's safe to go with the `version_removed` here.

## `background-attachment: local`

Like above, this appears to be ineffectual rather than unsupported per se. Since this is a standard valid value on a standard property, I think this is much more likely to end up affecting someone's cascade. Let's follow the guideline in this case. See my line comment for details.

fabb:
Ah, I created the PR based off the old `master` branch, not the current `main` branch 🤦 
In the mean time #10653 and #9254 happened, so I will rebase and make my changes upon these states. That PR changed Safari support to `"version_added": false` which is not correct, I will change it accordingly as suggested in the code comment above by @ddbeck.

fabb:
Note that in the caniuse sibling-PR it was agreed that Safari 13/14 should be marked as unsupported, not as partial as in this PR - should we do it consistently across the two repos?
https://github.com/Fyrd/caniuse/pull/5911

fabb:
I updated the information based on my own tests in Safari 13/14 on macOS and iOS + on the information in the caniuse repo: https://github.com/Fyrd/caniuse/blob/main/features-json/background-attachment.json.

As said above, there the newest version is marked as "unsupported" while here it's marked as "partial support", and maybe we should straighten this out.

css/properties/background-attachment.json
```diff
@@ -77,11 +77,21 @@
...
                 "version_added": "14"
               },
-              "safari": {
-                "version_added": false
-              },
+              "safari": [
+                {
+                  "version_added": "14",
+                  "partial_implementation": true,
+                  "notes": "<code>local</code> is recognized but has no effect due to [a bug](https://bugs.webkit.org/show_bug.cgi?id=219324)."
```
ddbeck, line 84: This is failing CI. For reasons I only vaguely remember, we use specific formats of bug URLs. See the CI output for details: https://github.com/mdn/browser-compat-data/pull/10542/checks?check_run_id=2756698276#step:5:2338

Also, we don't use Markdown in the notes field. You'll have to use an old fashioned `<a href='…'>` link, I'm afraid.

css/properties/background-attachment.json
```diff
@@ -77,11 +77,21 @@
...
+                {
+                  "version_added": "3.1",
+                  "version_removed": "14"
+                }
+              ],
               "safari_ios": {
-                "version_added": "4.2"
+                "version_added": "5",
+                "partial_implementation": true,
+                "notes": "<code>local</code> is recognized but has no effect."
```
ddbeck, line 94: I'd suggest using the same note text for both Safari and Safari for iOS (since it's possible that BCD consumers may present them separately). Ideally, notes should stand alone.

css/properties/background-attachment.json
```diff
@@ -125,12 +135,29 @@
...
+              "safari_ios": [
+                {
+                  "version_added": "13",
+                  "partial_implementation": true,
+                  "notes": "<code>local</code> is recognized but has no effect due to [a bug](https://bugs.webkit.org/show_bug.cgi?id=219324)."
+                },
+                {
+                  "version_added": "4.2",
+                  "version_removed": "13",
+                  "notes": "<code>local</code> is only supported when <code>-webkit-overflow-scrolling: touch</code> is NOT used."
```
ddbeck, line 158: For symmetry with the other notes, I'd propose new text like this:

```suggestion
                  "notes": "If <code>-webkit-overflow-scrolling: touch</code> is set, then <code>local</code> has no effect."
```

ddbeck (changes requested):
Thanks for the updates and additional research, @fabb. A few suggestions for you in line comments.

css/properties/background-attachment.json
fabb: afaik `local` never worked in iOS safari unlike macOS safari, and the webkit bug ticket only mentions macOS, so i'm not sure if i really should link the bug ticket here for iOS.

css/properties/background-attachment.json
fabb: done ✅

css/properties/background-attachment.json
fabb: done ✅

css/properties/background-attachment.json
ddbeck: Ah, OK. This is fine then. Thank you!

ddbeck (approved):
Looks good. Thank you! 🎉

PR #10560, opened 2021-05-21
Remove notes about support for SVG 1.1

foolip:
These notes came with the wiki migration:
https://github.com/mdn/browser-compat-data/pull/1585
https://github.com/mdn/browser-compat-data/pull/1587
https://github.com/mdn/browser-compat-data/pull/1588

Without knowing the difference between these interfaces in SVG 1.1 and
2, there's no way to interpret these notes. Any differences, if
uncovered, should be captured as much more concrete notes.

### DIFF
--- a/api/SVGPolygonElement.json
+++ b/api/SVGPolygonElement.json
@@ -12,20 +12,16 @@
             "version_added": "18"
           },
           "edge": {
-            "version_added": "12",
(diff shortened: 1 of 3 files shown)


ddbeck (approved):
I agree with this. Thank you! 👍

PR #10561, opened 2021-05-21
Fix order and versions of gutter properties for Safari

foolip:
This change happened in WebKit trunk version 606.1.4:
https://trac.webkit.org/changeset/228095/webkit
https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Configurations/Version.xcconfig?rev=228095

From the commit message:
> grid-column-gap => column-gap
> grid-row-gap => row-gap
> grid-gap => gap

See https://github.com/mdn/yari/issues/3839 about the order.

### DIFF
--- a/css/properties/column-gap.json
+++ b/css/properties/column-gap.json
@@ -134,7 +134,7 @@
               ],
               "safari": [
                 {
-                  "version_added": "12.1"
+                  "version_added": "12"
                 },
                 {
                   "version_added": "10.1",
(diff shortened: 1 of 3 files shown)


foolip:
This change happened in WebKit trunk version 606.1.4:
https://trac.webkit.org/changeset/228095/webkit
https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Configurations/Version.xcconfig?rev=228095

That is indeed Safari 12 / iOS 12, but I noticed some of the data says 12.1, so I'll fix that.

ddbeck (approved):
Thank you! 🎉

PR #10565, opened 2021-05-21
Update Chromium versions for api.Request.clone

vinyldarkscratch:
This PR fixes the Chrome Android data for `api.Request.clone`.  Running a quick test in the latest Chrome Android reveals there is indeed support, and I highly doubt that it would be implemented later than Chrome Desktop, so this PR sets it to match the version.

### DIFF
--- a/api/Request.json
+++ b/api/Request.json
@@ -453,7 +453,7 @@
               "version_added": "42"
             },
             "chrome_android": {
-              "version_added": false
+              "version_added": "42"
             },
             "edge": {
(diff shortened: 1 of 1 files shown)


foolip:
Is this based on testing or mirroring? Is it also supported in Samsung Internet and WebView?

vinyldarkscratch:
Oh what the heck, where did my PR description go?  Sorry about that, fixing now and providing details!

foolip (approved):

PR #10568, opened 2021-05-21
Remove WebView for Android 1.1, 1.6 and 4.2

vinyldarkscratch:
This PR is one part of work towards #10524.  I've recently found that Android WebView 1.1 and 1.6 have the same engine versions as 1.0 and 1.5 respectively.  Additionally, none of our data was set to either of these WebView versions, further leading towards a decision to remove it from the data.  I then found the same scenario with Android WebView 4.1 and 4.2, where 4.2 is unused and has the same engine version as 4.1.

This PR removes the data for WebView Android 1.1, 1.6, and 4.2 as they are unused in any of our data and have the exact same engine versions as their preceeding versions.

### DIFF
--- a/browsers/webview_android.json
+++ b/browsers/webview_android.json
@@ -10,23 +10,9 @@
           "engine": "WebKit",
           "engine_version": "523.12"
         },
-        "1.1": {
-          "release_date": "2009-02-09",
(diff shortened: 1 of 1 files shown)


foolip (approved):

Elchi3 (approved):
Thanks, it was me who added these. I didn't know better at the time :)
Some APIs might require a higher Android versions to work but since no feature is using these releases, let's merge it.

PR #10576, opened 2021-05-22
Remove Android WebView 2.3

vinyldarkscratch:
This PR removes Android WebView 2.3 from our data (as a part of #10524).  Since the WebKit versions match for both WebView 2.2 and 2.3, it's highly doubtful that there would be anything added in 2.3 that isn't present in 2.2.  Additionally, this PR changes everywhere where WebView is set to 2.3 and replaces the value with 2.2 in its place.

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -4843,7 +4843,7 @@
               "version_added": "1.0"
             },
             "webview_android": {
-              "version_added": "2.3"
+              "version_added": "2.2"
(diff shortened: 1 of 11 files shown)


foolip (approved):

Elchi3 (approved):
Looks good to me. None of the features changed look like they would typically require a higher Android version to work.

PR #10577, opened 2021-05-22
Remove Android WebView 4.1

vinyldarkscratch:
This PR removes Android WebView 4.1 from our data (as a part of #10524). Since the WebKit versions match for both WebView 4.1 and 4, it's highly doubtful that there would be anything added in 4.1 that isn't present in 4. Additionally, this PR changes everywhere where WebView is set to 4.1 and replaces the value with 4 in its place.


lucalves (approved):

foolip (approved):

jonjarry (commented):
[Spam]

Elchi3 (approved):
Thanks! Same here, none of these features look like they would typically require a higher Android version. Engine support seems the important factor here.

PR #10578, opened 2021-05-22
Add CSS functions folder

### LINKED ISSUE #10563: CSS folder structure - add functions folder
Ok, so the [CSS folder](https://github.com/mdn/browser-compat-data/tree/main/css) had 4 sub-folders:

* **at-rules**
* **properties**
* **selectors**
* **types** - includes both data-types and functions

CSS has over 100 functions and only ~25 data-types. In October 2020 we added **CSS functions** page (see https://github.com/mdn/sprints/issues/3838 and https://github.com/mdn/sprints/issues/3848).

We didn't do the separation in browser compatibility repository. Currently all the browser compatibility json files for CSS functions are inside the **types** folder. We need to create a new **functions** folder, move the functions-related JSON files to the new folder.

Also, we need to update the function related content pages, to change the **browser-compat** tag. For example, in [calc()](https://developer.mozilla.org/en-US/docs/Web/CSS/calc()) page we need to replace `browser-compat: css.types.calc` with `browser-compat: css.functions.calc`

@chrisdavidmills @rachelandrew

ramiy:
Fixes #10563


ddbeck:
We decided against making this change; see https://github.com/mdn/browser-compat-data/issues/10563#issuecomment-854657625. Thanks again for your understanding!

PR #10588, opened 2021-05-23
Change Android WebView from 1 to ≤37 for API A-G

vinyldarkscratch:

### DIFF
--- a/api/AnimationEvent.json
+++ b/api/AnimationEvent.json
@@ -108,7 +108,7 @@
               "version_added": "43"
             },
             {
-              "version_added": "1",
+              "version_added": "≤37",
               "version_removed": "70",
               "prefix": "WebKit"
             }
--- a/api/CanvasRenderingContext2D.json
+++ b/api/CanvasRenderingContext2D.json
@@ -599,7 +599,7 @@
               "version_added": "1.0"
             },
             "webview_android": {
-              "version_added": "1"
+              "version_added": "≤37"
             }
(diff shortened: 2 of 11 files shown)


foolip:
A bunch of these APIs probably were in WebView 1, rather than changing to a range and later back to an exact version, can we skip the middle step and only update data verified to be incorrect?

vinyldarkscratch:
I still need to configure Android emulators for versions between 2.0 and 4.4, so I'm not quite ready to set exact WebView versions for these just yet.  All I can verify is that these weren't present in Android WebView 1.0 (or 1.1 for that matter).

With that in mind, though, I could turn this into a breakdown draft PR to then follow-up on in the future?

foolip:
Aha, so you're sure this data is wrong, it's just not clear what exactly is right? Could the mirroring script be improved and run again for this data as a good approximation?

vinyldarkscratch:
I've just opened #10612 to discuss updates to the mirroring script!  (I'm personally thinking about making said updates after #6582...whenever I actually get around to finishing it up...)

foolip:
@jpmedley what do you think, should we change all of these versions?

foolip:
I just discussed this with @vinyldarkscratch. Since the existing data is confirmed to be wrong, and the new data also doesn't contradict the old data, making this change seems like an improvement, a more honest reflection of what we know.

foolip (approved):

PR #10589, opened 2021-05-23
Change Android WebView from 1 to ≤37 for API H-M

vinyldarkscratch:

### DIFF
--- a/api/MessageEvent.json
+++ b/api/MessageEvent.json
@@ -42,7 +42,7 @@
             "version_added": "1.0"
           },
           "webview_android": {
-            "version_added": "1"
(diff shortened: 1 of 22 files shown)


api/MessageEvent.json
```diff
@@ -94,7 +94,7 @@
               "version_added": "1.0"
             },
             "webview_android": {
-              "version_added": "37"
```
foolip, line 97: This is an outlier change in this PR, where the original data was "37" instead of "1". But it seems correct to me, I don't believe this was introduced in WebView 37 specifically.

foolip (approved):

PR #10614, opened 2021-05-24
Safari 14 is not "current" any longer

vinyldarkscratch:
This PR fixes a data issue where Safari 14 was left as `current` when Safari 14.1 released.  Caught by the linter introduced in #6167.


ddbeck (approved):
Thank you! 🎉

PR #10620, opened 2021-05-24
Update BluetoothRemoteGATTService members where no support data exists

saschanaz:
This is a subset of #10197 and partially replaces #10581.

#### Summary

Updates members filled only with `false`.

This assumes #10197 is correct 😉


foolip:
https://mdn-bcd-collector.appspot.com results from Windows set the Chrome version to 70, so I assume this has the same platform dependency as all of Web Bluetooth. Like in https://github.com/mdn/browser-compat-data/pull/10197 it's a bit hard to draw the line for how much cleanup of the structure to suggest, but I would approve simply matching data for other entries exactly, and treating cleanup of it all as a later problem.

vinyldarkscratch:
Since #10197 has been merged, the diff for this PR is now at 0, so we can close this PR.

PR #10621, opened 2021-05-24
Remove notes about nightly build from PaymentAddress

saschanaz:
Those are [not expected to be shipped](https://bugzilla.mozilla.org/show_bug.cgi?id=1709761) since [the interface is removed from the spec](https://github.com/w3c/payment-request/pull/955).

#### Summary

Removing notes.


ddbeck:
@saschanaz Merging https://github.com/mdn/browser-compat-data/pull/10581 created a conflict here, so rebase or merge is needed. Thank you!

ddbeck (approved):
Thank you! 🎉

PR #10622, opened 2021-05-24
Update data for paymentmethodchange event

saschanaz:
Copied from onpaymentmethodchange.

I think there should be a linter for this 🤔

#### Summary

A simple data copy.

### DIFF
--- a/api/PaymentRequest.json
+++ b/api/PaymentRequest.json
@@ -516,13 +516,13 @@
           "description": "<code>paymentmethodchange</code> event",
           "support": {
             "chrome": {
-              "version_added": false
+              "version_added": "76"
             },
             "chrome_android": {
-              "version_added": false
+              "version_added": true
             },
             "edge": {
-              "version_added": false
+              "version_added": "79"
             },
             "firefox": {
               "version_added": false,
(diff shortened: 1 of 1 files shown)


foolip:
> I think there should be a linter for this 🤔

Yeah, there's a lot of obviously wrong data that could be caught be something pretty simple here. The trouble is there's no hard rule, neither kind of entry implies the other. See https://github.com/mdn/browser-compat-data/issues/7545.

Some kind of grouping concept would be useful I think, where the expectation is that data should match but there's a way to make it not match if that's really what happened.

Edimalius:
#7545 <https://github.com/mdn/browser-compat-data/issues/7545>

foolip (approved):
Running mdn-bcd-collector doesn't update the `onpaymentmethodchange` entry this is based on, so it's probably mostly right.

Edimalius:
---------- Forwarded message ---------
From: exaelarras ***@***.***>
Date: Tue, Jun 1, 2021, 2:21 PM
Subject: [mdn/browser-compat-data] html.elements.menu - <PUT TITLEhttps://
docs.google.com/document/d/10696as0pWGV58nIGiq8nPZTwVDUtcW2or0orH35yoXo/edit?usp=drivesdk&disco=AAAAMsbAJRk
HERE> (#10689)
To: mdn/browser-compat-data ***@***.***>
Cc: Subscribed ***@***.***>


What information was incorrect, unhelpful, or incomplete? What did you
expect to see? Did you test this? If so, how? MDN page report details

   - Query: html.elements.menu
   - MDN URL: https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menu
   - Report started: 2021-06-01T06:55:07.067Z

—
You are receiving this because you are subscribed to this thread.
Reply to this email directly, view it on GitHub
<https://github.com/mdn/browser-compat-data/issues/10689>, or unsubscribe

PR #10632, opened 2021-05-25
Add Opera 75, Opera Android 63 and Edge 89 support for top level await

lukewarlow:
#### Summary

This PR adds Opera 75, Opera Android 63 and Edge 89 (which are all based on Chromium 89) support for await at the top level of a module.

Tested in Opera 75 and Edge 92 (which are the version I have access to)

### DIFF
--- a/javascript/operators/await.json
+++ b/javascript/operators/await.json
@@ -109,6 +109,7 @@
                 },
                 {
                   "version_added": "73",
+                  "version_removed": "75",
                   "flags": [
                     {
                       "type": "preference",
--- a/javascript/operators/await.json
+++ b/javascript/operators/await.json
@@ -76,7 +76,7 @@
                 "version_added": "89"
               },
               "edge": {
-                "version_added": false
+                "version_added": "89"
               },
               "firefox": {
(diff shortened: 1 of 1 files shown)


lukewarlow:
If https://github.com/mdn/browser-compat-data/pull/10334 is merged first this MR will need updating (or a follow up MR) to set Safari version to preview, as top level await has been supported since TP122

javascript/operators/await.json
```diff
@@ -103,17 +103,22 @@
...
-                    "name": "#enable-javascript-harmony"
-                  }
-                ]
-              },
+              "opera": [
+                {
+                  "version_added": "75"
+                },
+                {
+                  "version_added": "73",
```
ddbeck, line 111: I don't think this can be switched off now, so…

```suggestion
                  "version_added": "73",
                  "version_removed": "75",
```

ddbeck (changes requested):
One small suggestion in a line comment. Thank you, @lukewarlow!

ddbeck (approved):
Looks good. Thank you! 🎉

PR #10633, opened 2021-05-25
Fix release date for Safari 14.1

lukewarlow:
Release date for Safari 14.1 was listed as 2020 not 2021. This fixes that.


ddbeck (approved):
Thanks, @lukewarlow!

PR #10634, opened 2021-05-25
Update aspect-ratio.json

aburymichael:
#### Summary

Safari TP release 123 enables aspect ratio CSS property by default.

https://bugs.webkit.org/show_bug.cgi?id=223117

### DIFF
--- a/css/properties/aspect-ratio.json
+++ b/css/properties/aspect-ratio.json
@@ -69,7 +69,7 @@
               "version_added": false
             },
             "safari": {
-              "version_added": false
+              "version_added": "TP"
             },
             "safari_ios": {
               "version_added": false


ddbeck:
Thanks for this PR and welcome to the project, @aburymichael.

This PR is blocked (by https://github.com/mdn/browser-compat-data/pull/10334) and will fail CI because we don't yet have a way of tracking Safari TP and similar in-development versions of browsers. We're working on it, but this will be on hold until then.

aburymichael:
Thanks for the feedback @ddbeck.

My apologies, I had seen "TP" in a couple json files whilst learning the structure - granted I was looking for "TP" specifically as I noticed that [caniuse.com had "support unknown"](https://caniuse.com/mdn-css_properties_aspect-ratio) for the TP browser when in fact it is supported - assumed this was the place to do so. Will wait on #10334 to be resolved, thanks!

ddbeck:
Though #10334 is now resolved, this specific change was resolved with a specific version number by https://github.com/mdn/browser-compat-data/pull/10898. Thank you for starting this PR, @aburymichael, even though I have to close it without merging.

PR #10635, opened 2021-05-25
Update `text-indent` `each-line` and `hanging` keywords for Safari TP

aburymichael:
#### Summary
Added TP for Safari as Release 124 resolves `each-line` and `hanging` keywords for `text-indent` property

https://bugs.webkit.org/show_bug.cgi?id=223851

### DIFF
--- a/css/properties/text-indent.json
+++ b/css/properties/text-indent.json
@@ -78,7 +78,7 @@
                 "version_added": false
               },
               "safari": {
-                "version_added": false
+                "version_added": "TP"
               },
               "safari_ios": {
                 "version_added": false
@@ -126,7 +126,7 @@
                 "version_added": false
               },
               "safari": {
-                "version_added": false
+                "version_added": "TP"
               },
               "safari_ios": {
                 "version_added": false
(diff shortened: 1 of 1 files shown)


ddbeck:
Like your other PR, this is blocked (by https://github.com/mdn/browser-compat-data/pull/10334) and will fail CI because we don't yet have a way of tracking Safari TP and similar in-development versions of browsers. We're working on it, but this will be on hold until then.

css/properties/text-indent.json
```diff
@@ -78,7 +78,7 @@
                 "version_added": false
               },
               "safari": {
-                "version_added": false
+                "version_added": "TP"
```
ddbeck, line 81:
```suggestion
                "version_added": "preview"
```

css/properties/text-indent.json
```diff
@@ -126,7 +126,7 @@
...
```
ddbeck, line 129:
```suggestion
                "version_added": "preview"
```

ddbeck (changes requested):
Hi @aburymichael, since https://github.com/mdn/browser-compat-data/pull/10334 has been merged, we can support Safari TP with the new `"preview"` value. I've left line comments to demonstrate the change. Thank you for your patience on this!

ddbeck (approved):
Thank you! 🎉

PR #5472, opened 2020-01-10
[Chrome] Bug 1041095: Track `[attr=val s]` implementation

ExE-Boss:
**Chromium** has&nbsp;[bug&nbsp;1041095](https://bugs.chromium.org/p/chromium/issues/detail?id=1041095)&nbsp;open to&nbsp;implement&nbsp;this.

---

review?(@jpmedley)

PR #10641, opened 2021-05-26
[Chrome] Bug 1041095: Implement `[attr=val s]`

ExE-Boss:
#### Summary

This is now implemented in **Chromium 90**.

- <https://crbug.com/1041095>
- <https://github.com/chromium/chromium/commit/0e98f25e95f15b5cd36ad86d6e1d12872e4c29ed>
- <https://storage.googleapis.com/chromium-find-releases-static/0e9.html#0e98f25e95f15b5cd36ad86d6e1d12872e4c29ed>

#### Related issues

- <https://github.com/mdn/browser-compat-data/pull/5472>

### DIFF
--- a/css/selectors/attribute.json
+++ b/css/selectors/attribute.json
@@ -103,16 +103,31 @@
             "description": "Case-sensitive modifier (<code>s</code>)",
             "support": {
               "chrome": {
-                "version_added": false,
-                "notes": "See <a href='https://crbug.com/1041095'>bug 1041095</a>."
+                "version_added": "90",
+                "flags": [
+                  {
+                    "type": "preference",
+                    "name": "CSSCaseSensitiveSelector"
+                  }
+                ]
               },
               "chrome_android": {
-                "version_added": false,
-                "notes": "See <a href='https://crbug.com/1041095'>bug 1041095</a>."
+                "version_added": "90",
+                "flags": [
+                  {
+                    "type": "preference",
+                    "name": "CSSCaseSensitiveSelector"
+                  }
+                ]
               },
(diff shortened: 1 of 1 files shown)


ddbeck (changes requested):
Thanks for this PR, @ExE-Boss.

The linked bug/commit says:

> This support is before a ship so it's behind the flag.

Is there something else showing it was enabled by default? Or do you want to update this to show which flag needs to be flipped on?

css/selectors/attribute.json
```diff
@@ -103,16 +103,31 @@
             "description": "Case-sensitive modifier (<code>s</code>)",
             "support": {
               "chrome": {
-                "version_added": false,
-                "notes": "See <a href='https://crbug.com/1041095'>bug 1041095</a>."
+                "version_added": "90",
+                "flags": [
+                  {
+                    "type": "preference",
+                    "name": "CSSCaseSensitiveSelector"
```
ddbeck, line 110: I hadn't seen a flag named like this before and didn't find it in the UI. Turns out, this is the feature name in [`runtime_enabled_features.json5`, where it's a "test"](https://github.com/chromium/chromium/blob/master/third_party/blink/renderer/platform/runtime_enabled_features.json5#L479). As far as I can tell, such features can't be enabled in regular Chrome releases. I don't think BCD has a way to represent this yet (perhaps someday with https://github.com/mdn/browser-compat-data/pull/7396). I'd welcome a return of this PR, when/if Chrome supports this in a way that web developers can actually try out. Thanks, @ExE-Boss!

PR #10643, opened 2021-05-26
Adding Temporal API

meyerweb:
#### Summary

Addition of the Temporal API to BCD.


vinyldarkscratch:
This is currently blocked by #10619, where we're deciding whether or not to allow for all-false features within BCD.  As such, I'll be marking this as "not ready" for the time being.

meyerweb:
> This is currently blocked by #10619, where we're deciding whether or not to allow for all-false features within BCD. As such, I'll be marking this as "not ready" for the time being.

Fair enough.  Thanks for the update!

meyerweb:
Temporal support is landing in Safari Technology Preview as of STP 133, so this is no longer all-false!  If there’s a way to get this back on the “review and merge” train, let’s do it.

meyerweb:
> @meyerweb To get this moving again, we'd need to have the supported Safari features set to `"version_added": "preview"` (assuming it's not actually shipping yet). Do you want to start setting those? Maybe you know of some test suite, so you can quickly determine what, if any, parts of the API are unsupported (otherwise, we can talk about using the BCD collector to figure it out).

I do.  Remind me again, is there anything I need to say other than `"version_added": "preview"`?  It’s been too long since i was in BCD source.  (Pointers to documentation most welcome!)

queengooborg:
If it's enabled by default in Safari TP releases, then `"version_added": "preview"` is all that you need!

meyerweb:
Updated the information to what my testing shows (using https://people.igalia.com/emeyer/js/tests/).

queengooborg:
Thanks for the quick updates!

Last thing: I noticed that there are a number of added features marked as all-false.  We should probably remove them until a decision has been made of whether to keep them or not, but I'll defer to @ddbeck on that one.

meyerweb:
> Last thing: I noticed that there are a number of added features marked as all-false. We should probably remove them until a decision has been made of whether to keep them or not, but I'll defer to @ddbeck on that one.

Yes, portions of the Temporal API have yet to be implemented.  My feeling is that if we have parts of the API, then we should have all of it, even if it’s all-false.  That way readers get a clear view of what’s been done, and what hasn’t, instead of seeing that parts are available for testing out but not having any insight into the status of the rest of the API.  But that’s my personal take and may run afoul of the current state of MDN policy.

queengooborg (approved):
Thank you for waiting, and thank you for your hard work on this PR!  Although we're still deciding whether we want to keep/introduce all-false features in BCD, the general consensus is currently to disallow features without a valid spec URL.  You make a great point, in that if we're adding part of the API, we should add all of it.

This PR is looking good to me, thank you!

meyerweb:
Thank you!

PR #10645, opened 2021-05-27
Support multiple positional arguments to mirror scripts

foolip:
#### Summary

Make it possible to run the mirror script like `npm run mirror -- firefox_android api/_mixins/SVGFilterPrimitiveStandardAttributes__SVGFE*` by supporting multiple positional arguments.

No tests!

### DIFF
--- a/scripts/mirror.js
+++ b/scripts/mirror.js
@@ -17,7 +17,7 @@ const compareVersions = require('compare-versions');
 const browsers = require('..').browsers;
 
 const { argv } = require('yargs').command(
-  '$0 <browser> [feature_or_file]',
+  '$0 <browser> [feature_or_path..]',
   'Mirror values onto a specified browser if "version_added" is true/null, based upon its parent or a specified source',
   yargs => {
     yargs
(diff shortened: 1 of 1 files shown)


foolip:
@ddbeck or @Elchi3 for review?

vinyldarkscratch (approved):
Oh yeah, much nicer-looking and much more sensible!

scripts/mirror.js
```diff
@@ -17,18 +17,28 @@ const compareVersions = require('compare-versions');
...
+          'html',
+          'http',
+          'svg',
+          'javascript',
+          'mathml',
+          'webdriver',
+          'webextensions',
+        ],
       })
       .option('source', {
```
ddbeck, line 43: Doesn't quite reach within the diff, but would you mind adding a couple of usage examples (with [`.example()`](http://yargs.js.org/docs/#api-reference-examplecmd-desc)) toward the end of the yargs stuff? Two or three would be great (e.g., with a feature, with a path, with more than one of something).

ddbeck (changes requested):

queengooborg:
I'm going to go ahead and merge this PR now, but Daniel's suggestion is a great one, so I'll follow up on that shortly!

PR #10648, opened 2021-05-27
Mark many SVG APIs as supported in Opera 12.1 or earlier

foolip:
This is based on https://staging-dot-mdn-bcd-collector.appspot.com/
results from Opera 12.6 on Windows 7 only. No data for Opera 15 or 17
was reconfirmed but assumed to be correct.

Data for Android is simply mirrored with search/replace.

### DIFF
--- a/api/SVGAnimatedBoolean.json
+++ b/api/SVGAnimatedBoolean.json
@@ -24,10 +24,10 @@
             "version_added": "9"
           },
           "opera": {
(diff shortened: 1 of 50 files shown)


Elchi3 (approved):
I remember we added opera versions 15/14 to a lot of data because these were the first chromium based versions. It makes sense this was in an unknown Presto version already (≤12.1). Thanks! 👍

PR #10564, opened 2021-05-21
Mark SVGGraphicsElement members as supported since ancient times

foolip:
http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=9339 was
used to test for support of the interface itself. Results:
 - Chrome 30 (not in 29)
 - Not in IE11
 - Edge 13 => Edge 12 assumed
 - Firefox 20 (not in 19)
 - Not in Opera 12.16
 - Safari 6.2 (not in 6.0) => Safari 6.1 assumed

http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=9295 was
used to test for support for the members on concrete interfaces:
```html
<!DOCTYPE html>
<script>
var members = ['transform', 'getBBox', 'getCTM', 'getScreenCTM'];
var instances = {
  SVGAElement: document.createElementNS('http://www.w3.org/2000/svg', 'a'),
  SVGCircleElement: document.createElementNS('http://www.w3.org/2000/svg', 'circle'),
  SVGDefsElement: document.createElementNS('http://www.w3.org/2000/svg', 'defs'),
  SVGEllipseElement: document.createElementNS('http://www.w3.org/2000/svg', 'ellipse'),
  SVGForeignObjectElement: document.createElementNS('http://www.w3.org/2000/svg', 'foreignObject'),
  SVGGElement: document.createElementNS('http://www.w3.org/2000/svg', 'g'),
  SVGImageElement: document.createElementNS('http://www.w3.org/2000/svg', 'image'),
  SVGLineElement: document.createElementNS('http://www.w3.org/2000/svg', 'line'),
  SVGPathElement: document.createElementNS('http://www.w3.org/2000/svg', 'path'),
  SVGPolygonElement: document.createElementNS('http://www.w3.org/2000/svg', 'polygon'),
  SVGPolylineElement: document.createElementNS('http://www.w3.org/2000/svg', 'polyline'),
  SVGRectElement: document.createElementNS('http://www.w3.org/2000/svg', 'rect'),
  SVGSVGElement: document.createElementNS('http://www.w3.org/2000/svg', 'svg'),
  SVGSwitchElement: document.createElementNS('http://www.w3.org/2000/svg', 'switch'),
  SVGSymbolElement: document.createElementNS('http://www.w3.org/2000/svg', 'symbol'),
  SVGUseElement: document.createElementNS('http://www.w3.org/2000/svg', 'use'),
};
for (var iface in instances) {
  var instance = instances[iface];
[...]

PR #10653, opened 2021-05-27
Change Android WebView from 1 to ≤37 for API N-T

vinyldarkscratch:


api/SVGTRefElement.json
```diff
@@ -44,7 +44,7 @@
             "version_removed": "2.0"
           },
           "webview_android": {
-            "version_added": "1",
+            "version_added": "≤37",
```
foolip, line 47: Due to https://github.com/mdn/yari/issues/3238 this is going to end up looking broken on MDN. This was in Safari 1, so I'm guessing it was enabled with the rest of SVG in some early version of Android WebView? Can you identify that version to avoid this looking broken?

api/SVGAnimateColorElement.json
```diff
@@ -44,7 +44,7 @@
             "version_removed": "2.0"
           },
           "webview_android": {
-            "version_added": "1",
+            "version_added": "≤37",
```
foolip, line 47: This and other cases with `"version_removed": "37"` will end up looking broken on MDN due to https://github.com/mdn/yari/issues/3238. It's not so broken that I think it has to change, but if you can find the right version for this now, that would be better.

foolip (changes requested):

foolip:
For the SVG stuff, can you run `npm run traverse webview_android all 1 | grep -i svg` to ensure nothing is left? https://github.com/mdn/browser-compat-data/pull/10564 just changed some data that I think will need to be updated too.

api/SVGTRefElement.json
vinyldarkscratch: Unfortunately I don't have access to many old Android versions at the moment, only 1.0, 1.5, and 4.4, so I can't determine a specific version yet...

foolip:
But also, some of these entries can probably be removed to sidestep the whole problem.

foolip:
I've sent https://github.com/mdn/browser-compat-data/pull/10714 for one piece that can be dropped.

foolip:
Remaining lint errors if applying https://github.com/mdn/browser-compat-data/pull/10710:
```
Problems in 4 files:
✖ api/SVGAnimateColorElement.json
  Versions – 1 error:
  → api.SVGAnimateColorElement - version_removed: "37" must be greater than version_added: "≤37"
✖ api/SVGGraphicsElement.json
  Versions – 1 error:
  → api.SVGGraphicsElement - version_removed: "4.4" must be greater than version_added: "≤37"
✖ api/SVGSVGElement.json
  Versions – 2 errors:
  → api.SVGSVGElement.contentScriptType - version_removed: "37" must be greater than version_added: "≤37"
  → api.SVGSVGElement.contentStyleType - version_removed: "37" must be greater than version_added: "≤37"
✖ api/SVGTRefElement.json
  Versions – 1 error:
  → api.SVGTRefElement - version_removed: "4.4.3" must be greater than version_added: "≤37"
```

Since these have all been removed, I'd suggest leaving added as "1" and filing an issue about getting the right initial version for SVG in WebView, which involves updating both these remaining 1s and a lot of the new ≤37s.

vinyldarkscratch:
I've opened #10790 to discuss this, and will revert the linter-conflicting changes for the time being, until I can find a way to test those other WebView versions!

foolip (approved):

PR #10658, opened 2021-05-28
Added compat info for lab(), lch(), color()

LeaVerou:
#### Summary

I used `"TP"` for features that are in Safari TP and have not yet shipped. I also asked [here](https://github.com/mdn/browser-compat-data/issues/2991#issuecomment-850355950).
Note that Safari 10 only supported color(display-p3). The other predefined color spaces were added recently and are still in TP. We can iterate on breaking this down in future PRs.

[Testcase for `color(display-p3)`](https://result.dabblet.com/gist/f491f94dba0af1dfccffa24c46e770e5/7e79347a06ba4256297cc593dd24845d79efd0dd)

https://webkit.org/blog/11555/release-notes-for-safari-technology-preview-121/
https://webkit.org/blog/11548/release-notes-for-safari-technology-preview-120/
https://webkit.org/blog/7477/new-web-features-in-safari-10-1/#:~:text=css%20wide-gamut%20colors

### DIFF
--- a/css/types/color.json
+++ b/css/types/color.json
@@ -594,7 +594,7 @@
                 "version_added": "10.1"
               },
               "safari_ios": {
-                "version_added": "10"
+                "version_added": "10.3"
               },
               "samsunginternet_android": {
                 "version_added": false
--- a/css/types/color.json
+++ b/css/types/color.json
@@ -196,7 +196,7 @@
                 "version_added": false
               },
               "safari": {
-                "version_added": "TP"
+                "version_added": "preview"
               },
               "safari_ios": {
                 "version_added": false
@@ -441,7 +441,7 @@
                 "version_added": false
               },
               "safari": {
-                "version_added": "TP"
+                "version_added": "preview"
               },
               "safari_ios": {
                 "version_added": false
@@ -491,7 +491,7 @@
                 "version_added": false
               },
               "safari": {
-                "version_added": "TP"
+                "version_added": "preview"
               },
               "safari_ios": {
                 "version_added": false
--- a/css/types/color.json
+++ b/css/types/color.json
@@ -196,7 +196,7 @@
                 "version_added": false
               },
               "safari": {
-                "version_added": false
+                "version_added": "TP"
               },
               "safari_ios": {
                 "version_added": false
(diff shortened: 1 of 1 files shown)


lukewarlow:
With regards to the "TP" designation, https://github.com/mdn/browser-compat-data/pull/10334 will be of interest to you. From what I gather the exact value to use has not be decided but might be "preview". However, for now I believe the correct value would be false as it's not supported by a stable Safari.

LeaVerou:
Thank you. I would not want these to be merged as `false`, as this gives developers an erroneous idea about the state of these features. Therefore, I converted the PR to a draft for now and will monitor the PR you linked to.

css/types/color.json
```diff
@@ -196,7 +196,7 @@
                 "version_added": false
               },
               "safari": {
-                "version_added": false
+                "version_added": "TP"
```
foolip, line 199: In the new proposal this is going to be "preview"

css/types/color.json
```diff
@@ -460,6 +560,56 @@
...
+                "version_added": false
+              },
+              "opera_android": {
+                "version_added": false
+              },
+              "safari": {
+                "version_added": "10.1"
+              },
+              "safari_ios": {
+                "version_added": "10"
```
foolip, line 597: The corresponding iOS version is 10.3. Or was this actually shipped in different versions on macOS and iOS? If you have a test page for this I or another reviewer could verify the exact versions using BrowserStack.

LeaVerou: I have linked to a test page above, and I used Browserstack to test in Safari 10.

css/types/color.json
LeaVerou: Just changed it to `"preview"`!

css/types/color.json
foolip: Thanks, I see [the test](https://result.dabblet.com/gist/f491f94dba0af1dfccffa24c46e770e5/7e79347a06ba4256297cc593dd24845d79efd0dd). BrowserStack has only Safari 10.1 for macOS, but Sauce Labs has both, and with that I could verify the test is red in 10 but green in 10.1. One cause for doubt is that it's not running the same macOS version, but it's probably right.

Verifying the iOS version is harder. ([For Safari for iOS, use the iOS version number, not the Safari version number or WebKit version number.](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#safari-for-ios-versioning)) Neither BrowserStack nor Sauce have iOS 10.x other than 10.3.

https://webkit.org/blog/10042/wide-gamut-color-in-css-with-display-p3/ links to https://trac.webkit.org/changeset/207442/webkit which landed *after* the release of iOS 10, so it seems 10.3 should be used then:

```suggestion
                "version_added": "10.3"
```

ddbeck (changes requested):
Hi @LeaVerou, since https://github.com/mdn/browser-compat-data/pull/10334 has been merged, we can support Safari TP with the new `"preview"` value. Some merge conflicts have appeared however. If you could rebase and resolve the conflicts, this ought to be able to progress. Thank you!

LeaVerou:
I see that meanwhile, Safari 15 beta got released including this functionality, so someone else has added it. Exciting! Will go ahead and close this PR.

ddbeck:
Thanks for this PR and tidying up here, @LeaVerou!

PR #10662, opened 2021-05-28
Added Edge support for ::backdrop

lukewarlow:
#### Summary

Added Edge support for the standard ::backdrop pseudo element as of version 79 (when Edge migrated to chromium).

Also marked ms prefix as not supported from version 79.

Chrome has supported this since version 37.


ddbeck (approved):
Looks good. Thank you! 🎉

PR #10663, opened 2021-05-28
Added Opera Android support for ::target-text

lukewarlow:
#### Summary

With the release of Opera Android 63, it's now based on Chromium 89 which added support for the `::target-text` pseudo element.

Tested on Opera Android 63.3.3216 with this demo: https://mdn.github.io/css-examples/target-text/index.html#:~:text=From%20the%20foregoing%20remarks%20we%20may%20gather%20an%20idea%20of%20the%20importance

### DIFF
--- a/css/selectors/target-text.json
+++ b/css/selectors/target-text.json
@@ -29,7 +29,7 @@
               "version_added": "75"
             },
             "opera_android": {
-              "version_added": false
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Makes sense. Thank you! 🎉

PR #10679, opened 2021-05-31
Firefox 90 supports HTTP Fetch Sec-Fetch- headers

hamishwillee:
Firefox 90 adds support for the Fetch metadata request headers in https://bugzilla.mozilla.org/show_bug.cgi?id=1695911

This PR adds the version support for the four headers for FF90 for desktop and android. As this is now supported by 2 major browsers I have also changed to `experimental: false` (is that OK??)

Associated docs for this will be tracked in https://github.com/mdn/content/issues/5375

### DIFF
--- a/http/headers/sec-fetch-dest.json
+++ b/http/headers/sec-fetch-dest.json
@@ -16,10 +16,10 @@
               "version_added": "80"
             },
             "firefox": {
-              "version_added": false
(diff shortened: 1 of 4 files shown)


ddbeck (approved):
Looks great. Thank you! 🎉

Issue #10682, opened 2021-05-31
"Only supported for HTMLDocument" notes inconsistently applied


foolip:
https://github.com/mdn/browser-compat-data/issues/6839 fixed a bunch of entries for things that have been moved from `HTMLDocument` to `Document`, but there are still lots of APIs which were once on `HTMLDocument` which haven't gotten this treatment.

https://software.hixie.ch/utilities/js/live-dom-viewer/?saved=9330 is a test for all the entries that are currently in Document.json. As an example `document.head` seems to have been moved from `HTMLDocument` to `Document` in Firefox 61.

This is fairly well suited to systematic testing, but https://mdn-bcd-collector.appspot.com/ doesn't generate the variant tests that would be needed to do this. It's also doubtful that it's worth automating, since it would be fixed once and then not need any further testing.

caugner:
For what it's worth, here's a slightly updated version of the Live DOM Viewer test with ✅ and ❌ instead of "true" and "false" to make it more easy to spot the differences.

caugner:
This is the summary for Chrome: https://github.com/caugner/document-members/blob/main/issues/chrome.txt

Here's what Chrome version fixed which members:

| Member | Since | Until |
|--------|--------|--------|
| `hasFocus` | 17 | 34 |
| `alinkColor` | 1 | 64 |
| `all` | 1 | 64 | 
| `bgColor` | 1 | 64 | 
| `captureEvents` | 1 | 64 | 
| `clear` | 1 | 64 | 
| `close` | 1 | 45 | 
| `compatMode` | 1 | 8 | 
| `designMode` | 1 | 36 | 
| `dir` | 1 | 36 | 
| `embeds` | 1 | 36 | 
| `fgColor` | 1 | 64 | 
| `hasFocus` | 17 | 34 | 
| `height` | 1, 19 | 17, 31 |
| `linkColor` | 1 | 64 | 
| `open` | 1 | 45 | 
| `plugins` | 1 | 36 | 
| `releaseEvents` | 1 | 64 | 
| `scripts` | 1 | 36 | 
| `vlinkColor` | 1 | 64 | 
| `width` | 1, 19 | 17, 31 |
| `write` | 1 | 45 | 
| `writeln` | 1 | 45 |

caugner:
Here's a better overview for Chrome, Firefox and Safari: https://github.com/caugner/document-members/blob/main/SUMMARY.md

PR #10683, opened 2021-05-31
Make Firefox notes around documeny.body consistent with other notes

foolip:
Fixes https://github.com/mdn/browser-compat-data/issues/6839.

https://github.com/mdn/browser-compat-data/issues/10682 tracks using
this structure consistently where there's currently just a simple
statement.


ddbeck (approved):
Thank you! 🎉

PR #10687, opened 2021-05-31
Mirror Opera Android and Samsung Internet support for flex gap

foolip:
This is incorrectly making support seem less complete than it is:
https://developer.mozilla.org/en-US/docs/Web/CSS/gap#support_in_flex_layout


vinyldarkscratch (approved):
Looking good to me!

PR #10690, opened 2021-06-01
Improve XSLTProcessor notes

foolip:
A space after &lt;/code> was missing, but while in the area...

Test for whether non-string parameters are supported:
http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=9342

Test for whether transformTo* throws or return null on error:
http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=9344

Testing revealed the following:
- Safari 14.1 only supports string values and return null on error.
  This matches Chrome, as expected since this started out in WebKit.
- Edge 18 only supports string values and return null on error.
- Opera 12.16 only supports string values, but throws on error.

This unfortunately requires more notes for Opera in order to capture
what happened.

### DIFF
--- a/api/XSLTProcessor.json
+++ b/api/XSLTProcessor.json
@@ -108,7 +108,8 @@
               "notes": "Chrome only supports string values."
             },
             "edge": {
-              "version_added": "12"
+              "version_added": "12",
+              "notes": "Edge only supports string values."
             },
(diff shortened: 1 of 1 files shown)


vinyldarkscratch (approved):
Looks good to me, thanks!

PR #10693, opened 2021-06-01
Add exact versions for insertRule optional index

foolip:
http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=9347 was
used to confirm it was optional in the following browsers:
 - Chrome ≤15
 - Edge ≤13 (not optional in IE11)
 - Firefox 55 (not optional in 54)
 - Opera 15 (not optional in 12.16)
 - Safari ≤4

Assumed/mirrored data:
 - Edge 12
 - Firefox for Android 55
 - Opera for Android 14

### DIFF
--- a/api/CSSStyleSheet.json
+++ b/api/CSSStyleSheet.json
@@ -305,22 +305,22 @@
                 "version_added": "18"
               },
               "edge": {
-                "version_added": "≤79"
+                "version_added": "12"
               },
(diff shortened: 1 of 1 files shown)


vinyldarkscratch (approved):
Confirmed the behavior is the same for both Edge 12 and 13.  Thanks!

PR #10694, opened 2021-06-01
Add exact versions for toggle force argument

foolip:
Follow-up to https://github.com/mdn/browser-compat-data/pull/8482, using
http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=8808 to
confirm support in Edge 13 and Opera 15, and a lack of support in Opera
12.16. Edge 12 is assumed but not confirmed with the test.

### DIFF
--- a/api/DOMTokenList.json
+++ b/api/DOMTokenList.json
@@ -790,7 +790,7 @@
                 "version_added": "25"
               },
               "edge": {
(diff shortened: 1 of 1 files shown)


vinyldarkscratch (approved):
Confirmed log output is the same in Edge 12, thanks!

PR #10695, opened 2021-06-01
Rename entry for FormData.append's filename argument

foolip:
This is more consistent with force_argument and similar entries for
DOMTokenList. Importantly, including an underscore makes it clear that
it's not an exposed name of an API.


api/FormData.json
```diff
@@ -153,9 +153,9 @@
             "deprecated": false
           }
         },
-        "AppendWithFilename": {
+        "filename_argument": {
```
ddbeck, line 156: Nice! This is the kind of thing that I was trying to tackle with this new guideline: https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#parameters-and-parameter-object-features. Want to change this to use "parameter" instead of "argument" instead?

ddbeck (changes requested):

api/FormData.json
foolip: Since you ask I have to say I don't quite *want* to use "parameter", since the "argument" terminology is [baked into the platform itself](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/arguments), but consistency always wins and I see there are more foo_parameter than foo_argument entries in BCD. So I'll fall in line :)

foolip: And pushing the rest of the data into line as well: https://github.com/mdn/browser-compat-data/pull/10700

ddbeck: Yeah, "want" was probably the wrong word (I also thought it should've been argument, https://github.com/mdn/browser-compat-data/pull/10509, but MDN uses parameter far more often). Thank you!

PR #10696, opened 2021-06-01
Replace Opera ≤15 versions with something more precise

foolip:
http://mdn-bcd-collector.appspot.com/tests/api/SVGFECompositeElement
shows support for the interface corresponding to SVG's <feComposite>
element.

http://mdn-bcd-collector.appspot.com/tests/css/properties/-webkit-tap-highlight-color
shows no support in Presto, but it is in Opera 15.

Tested on Opera 12.16 + 15 on Windows 7.

### DIFF
--- a/css/properties/-webkit-tap-highlight-color.json
+++ b/css/properties/-webkit-tap-highlight-color.json
@@ -24,10 +24,10 @@
               "version_added": false
             },
             "opera": {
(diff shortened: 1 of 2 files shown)


vinyldarkscratch (approved):
Thanks for the PR, looks good to me!

PR #10699, opened 2021-06-01
Set Firefox 89 as stable

lukewarlow:
#### Summary

Set Firefox and Firefox Android 89 as stable release.

https://www.mozilla.org/en-US/firefox/89.0/releasenotes/

https://www.mozilla.org/en-US/firefox/android/89.0/releasenotes/


ddbeck (approved):
Thank you! 🎉

PR #10700, opened 2021-06-01
Rename argument to parameter in context of methods/constructors

foolip:
See https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#parameters-and-parameter-object-features

### DIFF
--- a/api/RTCPeerConnection.json
+++ b/api/RTCPeerConnection.json
@@ -1726,9 +1726,9 @@
             "deprecated": false
           }
         },
-        "MediaStreamTrack_parameter": {
+        "selector_parameter": {
           "__compat": {
-            "description": "Optional <code>MediaStreamTrack</code> parameter",
+            "description": "<code>selector</code> parameter",
             "support": {
               "chrome": {
                 "version_added": "67"
--- a/api/RTCPeerConnection.json
+++ b/api/RTCPeerConnection.json
@@ -1726,9 +1726,9 @@
             "deprecated": false
           }
         },
-        "MediaStreamTrack_argument": {
+        "MediaStreamTrack_parameter": {
           "__compat": {
-            "description": "Optional <code>MediaStreamTrack</code> argument",
+            "description": "Optional <code>MediaStreamTrack</code> parameter",
             "support": {
               "chrome": {
                 "version_added": "67"
--- a/api/URL.json
+++ b/api/URL.json
@@ -224,7 +224,7 @@
             "deprecated": false
           }
         },
-        "no_MediaStream_parameter": {
+        "no_MediaStream_argument": {
           "__compat": {
             "description": "No longer accepts <code>MediaStream</code> object",
             "support": {
--- a/api/URL.json
+++ b/api/URL.json
@@ -224,7 +224,7 @@
             "deprecated": false
           }
         },
-        "no_MediaStream_argument": {
+        "no_MediaStream_parameter": {
           "__compat": {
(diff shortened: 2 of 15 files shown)


api/RTCPeerConnection.json
```diff
@@ -1726,9 +1726,9 @@
             "deprecated": false
           }
         },
-        "MediaStreamTrack_argument": {
+        "MediaStreamTrack_parameter": {
```
foolip, line 1729: I still don't like this name but don't know what it should be per https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#parameters-and-parameter-object-features

*All* arguments are unnamed, so read pedantically this should just be `first_parameter`. But if not for that, maybe `selector_parameter` would be better, since that's what the parameter is called by convention.

api/URL.json
```diff
@@ -224,7 +224,7 @@
             "deprecated": false
           }
         },
-        "no_MediaStream_argument": {
+        "no_MediaStream_parameter": {
```
foolip, line 227: I don't approve of this name either but don't know what to suggest. It's about the lack of support for a specific type as the first parameter, not support for the parameter itself.

I think it'd probably make sense to invert this and treat it as a deprecated feature which has been removed, to put it on a path to eventual removal from BCD.

api/RTCPeerConnection.json
foolip: Arguments are unnamed in the sense that the names cannot be observed by web developers. If the arguments were renamed in specs and implementations to random strings, everything would continue to work the same for web developers.

Dictionary members however do have names, which have to appear at the call sites. Compare:
```js
let ev = new Event("hey", {bubbles:true}); // bubbles is a name!
document.dispatchEvent(ev); // the first argument's name "event" appears nowhere
```

ddbeck: Ah, OK. I understand now. I guess my take here is that we should always go for a name, if at all possible, in the description. Names may not be strictly observable, but they're not completely hidden from developers either (e.g., when I type `new Event(`, VS Code shows me a tooltip for the `type` argument, presumably derived from the IDL).

If you want to name the feature keys by the ordinal though, I might be convinced of that. We could start a new discussion on the guideline, if you like. I'm willing to entertain some level of pedantry. 😄

foolip: No, I also think we should use the spec names for arguments, just noting that a pedantic reading of the guideline didn't work :)

ddbeck: OK, so want to rename this to `selector_parameter` instead and revise the description to match?

api/URL.json
ddbeck: > I think it'd probably make sense to invert this and treat it as a deprecated feature which has been removed, to put it on a path to eventual removal from BCD.

Yes, this sounds like the right approach to me and what I would've suggested myself. (If it's too fussy to deal with in this PR, then I'd suggest leaving it out of this PR entirely and opening an issue to track the idea.)

ddbeck (changes requested):

api/RTCPeerConnection.json
foolip: Yes!

api/URL.json
foolip: I'll revert this.

ddbeck (approved):
Thank you! 🎉

PR #10708, opened 2021-06-02
Add Chromium versions for SVGAltGlyphDefElement API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `SVGAltGlyphDefElement` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.2).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/SVGAltGlyphDefElement

### DIFF
--- a/api/SVGAltGlyphDefElement.json
+++ b/api/SVGAltGlyphDefElement.json
@@ -26,7 +26,7 @@
             "version_added": false
           },
           "opera": {
-            "version_added": "≤15",
+            "version_added": "15",
             "version_removed": "27"
           },
           "opera_android": {
--- a/api/SVGAltGlyphDefElement.json
+++ b/api/SVGAltGlyphDefElement.json
@@ -30,7 +30,7 @@
             "version_removed": "27"
           },
           "opera_android": {
-            "version_added": "≤14",
+            "version_added": "14",
             "version_removed": "27"
           },
           "safari": {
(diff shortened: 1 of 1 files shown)


api/SVGAltGlyphDefElement.json
```diff
@@ -24,10 +26,12 @@
             "version_added": false
           },
           "opera": {
-            "version_added": null
+            "version_added": "≤15",
```
foolip, line 29:
```suggestion
            "version_added": "15",
```

I confirmed no support in Opera 12.16 using http://mdn-bcd-collector.appspot.com/tests/api/SVGAltGlyphDefElement

I'm actually not 100% sure this means no support of the `<altGlyphDef>` element, but the interface isn't there. Actually neither is `SVGAltGlyphElement`. There's always something more wrong :)

api/SVGAltGlyphDefElement.json
```diff
@@ -24,10 +26,12 @@
...
+            "version_removed": "27"
           },
           "opera_android": {
-            "version_added": null
+            "version_added": "≤14",
```
foolip, line 33:
```suggestion
            "version_added": "14",
```

foolip (changes requested):

foolip (approved):

PR #10712, opened 2021-06-02
Add Chromium versions for api.ServiceWorkerGlobalScope

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `ServiceWorkerGlobalScope` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.0.2).  Results are manually confirmed for accuracy.

Additionally, this does a little bit of flag cleanup to simplify the data.

### DIFF
--- a/api/ServiceWorkerGlobalScope.json
+++ b/api/ServiceWorkerGlobalScope.json
@@ -694,40 +694,14 @@
         "__compat": {
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/ServiceWorkerGlobalScope/oncanmakepayment",
           "support": {
-            "chrome": [
-              {
-                "version_added": "70"
-              },
-              {
-                "version_added": "61",
-                "flags": [
-                  {
-                    "type": "preference",
-                    "name": "#service-worker-payment-apps",
-                    "value_to_set": "Enabled"
-                  }
-                ]
-              }
-            ],
+            "chrome": {
+              "version_added": "70"
+            },
             "chrome_android": {
-              "version_added": "61",
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "#service-worker-payment-apps",
-                  "value_to_set": "Enabled"
-                }
-              ]
+              "version_added": "70"
             },
             "edge": {
-              "version_added": "≤79",
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "#service-worker-payment-apps",
-                  "value_to_set": "Enabled"
-                }
-              ]
+              "version_added": "79"
             },
             "firefox": {
               "version_added": null
@@ -739,10 +713,10 @@
               "version_added": false
             },
             "opera": {
-              "version_added": null
+              "version_added": "57"
             },
             "opera_android": {
-              "version_added": null
+              "version_added": "49"
             },
             "safari": {
               "version_added": null
@@ -751,7 +725,7 @@
               "version_added": null
             },
             "samsunginternet_android": {
-              "version_added": false
+              "version_added": "10.0"
             },
             "webview_android": {
               "version_added": false
--- a/api/ServiceWorkerGlobalScope.json
+++ b/api/ServiceWorkerGlobalScope.json
@@ -1161,40 +1161,14 @@
         "__compat": {
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/ServiceWorkerGlobalScope/onpaymentrequest",
           "support": {
-            "chrome": [
-              {
-                "version_added": "70"
-              },
-              {
-                "version_added": "57",
-                "flags": [
-                  {
-                    "type": "preference",
-                    "name": "#service-worker-payment-apps",
-                    "value_to_set": "Enabled"
-                  }
-                ]
-              }
-            ],
+            "chrome": {
+              "version_added": "70"
+            },
             "chrome_android": {
-              "version_added": "57",
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "#service-worker-payment-apps",
-                  "value_to_set": "Enabled"
-                }
-              ]
+              "version_added": "70"
             },
             "edge": {
-              "version_added": "≤79",
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "#service-worker-payment-apps",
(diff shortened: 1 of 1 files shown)


api/ServiceWorkerGlobalScope.json
```diff
@@ -404,16 +404,21 @@
...
-                {
-                  "type": "preference",
-                  "name": "#service-worker-payment-apps",
-                  "value_to_set": "Enabled"
-                }
-              ]
-            },
+            "chrome": [
+              {
+                "version_added": "70"
```
foolip, line 409: This matches `AbortPaymentEvent`, good!

Can you update the commit message to reflect this is updating more than the `onmessageerror` entry?

api/ServiceWorkerGlobalScope.json
```diff
@@ -404,16 +404,21 @@
...
+              },
+              {
+                "version_added": "61",
+                "flags": [
```
foolip, line 413: Chrome 70 is old enough (2018-10-16) that this flag entry can be removed now.

api/ServiceWorkerGlobalScope.json
```diff
@@ -404,16 +404,21 @@
...
+                  {
+                    "type": "preference",
+                    "name": "#service-worker-payment-apps",
+                    "value_to_set": "Enabled"
+                  }
+                ]
+              }
+            ],
             "chrome_android": {
```
foolip, line 422: Hasn't this shipped in Chrome for Android too?

api/ServiceWorkerGlobalScope.json
```diff
@@ -1079,7 +1089,7 @@
           "spec_url": "https://notifications.spec.whatwg.org/#dom-serviceworkerglobalscope-onnotificationclick",
           "support": {
             "chrome": {
-              "version_added": "40"
+              "version_added": "42"
```
foolip, line 1092: I would suggest making this change separately. A small change in the version number requires a closer look to see if the old data was actually correct and a note is missing. (Also Chrome for Android is left unchanged below.)

api/ServiceWorkerGlobalScope.json
```diff
@@ -1177,16 +1187,21 @@
...
-                }
-              ]
-            },
+            "chrome": [
+              {
+                "version_added": "70"
+              },
+              {
+                "version_added": "57",
+                "flags": [
```
foolip, line 1196: Can be removed.

api/ServiceWorkerGlobalScope.json
```diff
@@ -1345,7 +1360,7 @@
           "spec_url": "https://w3c.github.io/push-api/#dom-serviceworkerglobalscope-onpushsubscriptionchange",
           "support": {
             "chrome": {
-              "version_added": "40"
+              "version_added": false
```
foolip, line 1363: I would split out this false&rarr;real change too.

foolip (changes requested):

api/ServiceWorkerGlobalScope.json
vinyldarkscratch: It has, yep -- correcting!

api/ServiceWorkerGlobalScope.json
```diff
@@ -715,16 +694,21 @@
...
-                }
-              ]
-            },
+            "chrome": [
+              {
+                "version_added": "70"
+              },
+              {
+                "version_added": "61",
+                "flags": [
```
foolip, line 703: This flag entry can be removed.

api/ServiceWorkerGlobalScope.json
```diff
@@ -1295,7 +1258,7 @@
           "spec_url": "https://w3c.github.io/push-api/#dom-serviceworkerglobalscope-onpush",
           "support": {
             "chrome": {
-              "version_added": "40"
+              "version_added": "42"
```
foolip, line 1261: This now doesn't match Chrome for Android. Was it really shipped at different times on different platforms?

foolip (changes requested):

api/ServiceWorkerGlobalScope.json
vinyldarkscratch: I doubt this was added at different times.  The data came from the collector, so I'm thinking that it's possible Chrome 40 is incorrect.  I'd like to await a follow-up PR to look into this more.

foolip: Would you mind splitting out the changes that make versions out of sync? I can look into them while reviewing, but if we merge it I think it's likely it will be left inconsistent as we have no burndown list of issues like this.

vinyldarkscratch: Sure thing!  I've opened #10793 for this change.

api/ServiceWorkerGlobalScope.json
```diff
@@ -1129,7 +1087,7 @@
           "spec_url": "https://notifications.spec.whatwg.org/#dom-serviceworkerglobalscope-onnotificationclose",
           "support": {
             "chrome": {
-              "version_added": "40"
+              "version_added": "50"
```
foolip, line 1090: Can you split out this change too?

api/ServiceWorkerGlobalScope.json
```diff
@@ -1596,7 +1533,7 @@
           "spec_url": "https://w3c.github.io/ServiceWorker/#service-worker-global-scope-registration",
           "support": {
             "chrome": {
-              "version_added": "40"
+              "version_added": "42"
```
foolip, line 1536: And this.

api/ServiceWorkerGlobalScope.json
```diff
@@ -1693,7 +1630,7 @@
           "spec_url": "https://w3c.github.io/ServiceWorker/#service-worker-global-scope-skipwaiting",
           "support": {
             "chrome": {
-              "version_added": "40"
+              "version_added": "41"
```
foolip, line 1633: And this.

foolip (changes requested):

foolip (approved):

PR #10715, opened 2021-06-02
Add Chromium versions for WebGLVertexArrayObjectOES API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `WebGLVertexArrayObjectOES` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (pending PR).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/WebGLVertexArrayObjectOES

### DIFF
--- a/api/WebGLVertexArrayObjectOES.json
+++ b/api/WebGLVertexArrayObjectOES.json
@@ -6,10 +6,10 @@
         "spec_url": "https://www.khronos.org/registry/webgl/specs/latest/2.0/#3.6",
         "support": {
           "chrome": {
-            "version_added": null
+            "version_added": false
           },
           "chrome_android": {
-            "version_added": null
+            "version_added": false
           },
           "edge": {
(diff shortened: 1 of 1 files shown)


foolip:
This is actually in Chromium:
https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/webgl/webgl_vertex_array_object_oes.idl;drc=f5eccec110f06a006f4ec2b249b3c4b92e7dac35

I think the data should probably match `OES_vertex_array_object` but I don't know that for a fact.

vinyldarkscratch:
Ack, I thought we had a custom test set up for this API.  Revising now!

foolip (approved):
Matches `OES_vertex_array_object`, makes sense!

PR #10718, opened 2021-06-02
Add Chromium versions for WorkerNavigator API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `WorkerNavigator` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.4).  Results are manually confirmed for accuracy.

### DIFF
--- a/api/WorkerNavigator.json
+++ b/api/WorkerNavigator.json
@@ -6,10 +6,10 @@
         "spec_url": "https://html.spec.whatwg.org/multipage/workers.html#the-workernavigator-object",
         "support": {
           "chrome": {
-            "version_added": "31"
+            "version_added": "1"
           },
           "chrome_android": {
-            "version_added": "31"
+            "version_added": true
           },
           "edge": {
             "version_added": "12"
@@ -24,10 +24,10 @@
             "version_added": "10"
           },
           "opera": {
-            "version_added": "18"
+            "version_added": true
           },
           "opera_android": {
-            "version_added": "18"
+            "version_added": true
           },
           "safari": {
             "version_added": true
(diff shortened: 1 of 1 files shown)


api/WorkerNavigator.json
```diff
@@ -6,10 +6,10 @@
         "spec_url": "https://html.spec.whatwg.org/multipage/workers.html#the-workernavigator-object",
         "support": {
           "chrome": {
-            "version_added": "1"
+            "version_added": "31"
```
foolip, line 9: This does not seem plausible to me, and seems like it's probably a test issue. I've sent https://github.com/foolip/mdn-bcd-collector/pull/1194 to use an instance.

I would expect `api.WorkerNavigator` data to match `api.Worker`, which it didn't before but it was much closer.

api/WorkerNavigator.json
```diff
@@ -202,10 +202,10 @@
           "spec_url": "https://w3c.github.io/ServiceWorker/#navigator-serviceworker",
           "support": {
             "chrome": {
-              "version_added": null
+              "version_added": false
```
foolip, line 205: This part of the change looks good, I can't find `serviceWorker` on `WorkerNavigator` in Chromium.

foolip (changes requested):

api/WorkerNavigator.json
vinyldarkscratch: I'll go ahead and revert this for now as I find the data questionable, though I believe Chrome 1 is also incorrect (considering worker support was added in Chrome 4).  Something to revisit later, though!

foolip (approved):

PR #10721, opened 2021-06-02
Demix GeometryUtils to Document/Element/Text.getBoxQuads

foolip:
Nothing of GeometryUtils has shipped in any browser yet. Preserve just
the minimum of what already had flag data, which is getBoxQuads.

CSSPseudoElement includes GeometryUtils in the spec, but not in Gecko
source, so it is omitted.


foolip:
@vinyldarkscratch FYI

vinyldarkscratch:
Thanks @foolip!  Based upon the [spec](https://drafts.csswg.org/cssom-view/#the-geometryutils-interface), it looks like `GeometryUtils` is also included on `CSSPseudoElement`?

foolip:
> Thanks @foolip! Based upon the [spec](https://drafts.csswg.org/cssom-view/#the-geometryutils-interface), it looks like `GeometryUtils` is also included on `CSSPseudoElement`?

Yes, but to quote myself:

> CSSPseudoElement includes GeometryUtils in the spec, but not in Gecko source, so it is omitted.

The entry would have only false values.

vinyldarkscratch:
Ah, apologies, didn't read that part until just now -- looks good to me!

vinyldarkscratch (approved):

foolip:
MDN cleanup in https://github.com/mdn/content/pull/5579.

PR #10722, opened 2021-06-02
Add Chromium versions for api.Headers.lexicographical_sorting

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `lexicographical_sorting` member of the `Headers` API, based upon commit history and date.

Commit: https://storage.googleapis.com/chromium-find-releases-static/abb.html#abbafbc0bacfca5beada1e3d1a5299569e0a5262 / https://storage.googleapis.com/chromium-find-releases-static/da6.html#da6e2412d9e3547954e1196c3b8d523be7090df6

### DIFF
--- a/api/Headers.json
+++ b/api/Headers.json
@@ -844,13 +844,13 @@
           "description": "Lexicographical sorting, and values from duplicate header names combined when iterated.",
           "support": {
             "chrome": {
-              "version_added": null
+              "version_added": "57"
             },
             "chrome_android": {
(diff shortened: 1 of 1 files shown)


foolip (approved):
Those two commits match with impressive accuracy to the description of this entry. A bit surprising when it must have been created for the Firefox entry first, but I guess these changes naturally go together in implementation. Thanks!

PR #10738, opened 2021-06-03
Add/correct Chromium data for OverconstrainedErrorEvent (and related)

vinyldarkscratch:
This PR updates the Chromium data for the `OverconstrainedErrorEvent` API, as well as the `onoverconstrained` event handler of the `MediaStreamTrack` API, setting both to `false`.

From what I have gathered, this event API was implemented in an older version of the [Media Capture and Streams spec](https://w3c.github.io/mediacapture-main/), before suddenly vanishing (no, seriously, there's little to no logs indicating it was removed).  Based upon what I can determine, the event handler is the _only_ location where it was in use -- so, in my search through the source code, rather than try to find the event API, I searched for the event handler.  Looking through the commits, both reading their titles and viewing the exact file, I have been able to determine that Chrome has never supported this event handler (which in turn, means the event itself isn't supported either).

Additionally, I've run the following test code in a few different Chrome versions.  All have reported `false`:

```js
var mediaStream = navigator.mediaDevices.getUserMedia({ audio: true });
mediaStream.then(function (ms) {
	return ms.getAudioTracks()[0];
}).then(function (instance) {
	alert('onoverconstrained' in instance);
});
```

### DIFF
--- a/api/MediaStreamTrack.json
+++ b/api/MediaStreamTrack.json
@@ -936,10 +936,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/MediaStreamTrack/onoverconstrained",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": false
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": false
             },
             "edge": {
               "version_added": "12"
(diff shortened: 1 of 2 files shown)


foolip:
The spec removal was in https://github.com/w3c/mediacapture-main/pull/576, found with `git log -S OverconstrainedErrorEvent` in a checkout of the spec repo.

foolip (approved):
Excellent, thank you!

PR #10740, opened 2021-06-03
Add mdn link for SVGGraphicsElement.getBBox

teoli2003:
Add mdn link for SVGGraphicsElement.getBBox


Elchi3 (approved):
Good catch, thank you! 👍

PR #10741, opened 2021-06-03
Add Chromium versions for api.ExtendableEvent.waitUntil.async_waitUntil

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `waitUntil.async_waitUntil` member of the `ExtendableEvent` API, based upon commit history.

Reasoning: Return type is void and has never changed to any other return type
Related line: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/service_worker/extendable_event.idl;l=36;bpv=0;bpt=0

### DIFF
--- a/api/ExtendableEvent.json
+++ b/api/ExtendableEvent.json
@@ -150,13 +150,13 @@
         },
         "async_waitUntil": {
           "__compat": {
-            "description": "Asynchronous <code>waitUntil</code>",
+            "description": "<code>waitUntil</code> may be called asynchronously",
             "support": {
               "chrome": {
-                "version_added": false
+                "version_added": "60"
               },
               "chrome_android": {
-                "version_added": false
+                "version_added": "60"
               },
               "edge": {
                 "version_added": "17"
@@ -171,10 +171,10 @@
                 "version_added": false
               },
               "opera": {
-                "version_added": false
+                "version_added": "47"
               },
               "opera_android": {
-                "version_added": false
+                "version_added": "44"
               },
               "safari": {
                 "version_added": false
@@ -183,10 +183,10 @@
                 "version_added": false
               },
               "samsunginternet_android": {
-                "version_added": false
+                "version_added": "8.0"
               },
               "webview_android": {
-                "version_added": false
+                "version_added": "60"
               }
             },
             "status": {
--- a/api/ExtendableEvent.json
+++ b/api/ExtendableEvent.json
@@ -153,10 +153,10 @@
             "description": "Asynchronous <code>waitUntil</code>",
             "support": {
               "chrome": {
-                "version_added": null
+                "version_added": false
               },
               "chrome_android": {
-                "version_added": null
+                "version_added": false
               },
               "edge": {
                 "version_added": "17"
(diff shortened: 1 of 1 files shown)


foolip:
Since the return type is undefined in https://w3c.github.io/ServiceWorker/#extendableevent I think this can't be about the return type, but maybe about what kinds of values one can pass in? I can't tell from the description really. Looking up the context where this entry was added seems like a good next step.

vinyldarkscratch:
Unfortunately looking at the commit history isn't too much of a help, so this was my best guess on what it is.  Considering that Firefox has a different version number than the main feature, we can probably base it upon whatever change was made there.

foolip:
Yes exactly, what evidence was originally given for the Firefox version? That should lead to some clue about what this feature is.

vinyldarkscratch:
Argh, the Firefox version came from the wiki tables, so there's no history there...  The WebIDL doesn't seem to be giving any clear indication what the feature means, and what's more, I see [no commits](https://hg.mozilla.org/mozilla-central/log/tip/dom/webidl/ExtendableEvent.webidl) that would've landed in Firefox 53 for the WebIDL.  Not really finding anything in the source code either, but that's probably because it's a mess and Searchfox doesn't show the logs before a file rename.

I'm still confused on what this feature really means, and to be honest, I'm leaning towards just scrapping it.  Not quite to that conclusion though.

foolip:
I'm pretty sure this about https://bugzilla.mozilla.org/show_bug.cgi?id=1263304, if the wiki edit history was available it could be confirmed for certain if it was @chrisdavidmills making the edits around the time of https://bugzilla.mozilla.org/show_bug.cgi?id=1263304#c36.

foolip:
Assuming that's correct, I followed links like this:
- https://bugzilla.mozilla.org/show_bug.cgi?id=1263304
- https://github.com/w3c/ServiceWorker/issues/771
- https://bugs.chromium.org/p/chromium/issues/detail?id=601095
- https://bugs.chromium.org/p/chromium/issues/detail?id=621440
- https://chromium.googlesource.com/chromium/src.git/+/eaf348dc9864d8073fe01472a91c33f21e3ec4b6
- https://storage.googleapis.com/chromium-find-releases-static/eaf.html#eaf348dc9864d8073fe01472a91c33f21e3ec4b6

So the answer is Chromium 60.

foolip:
I sort of think we should delete this entry though or turn it into notes. I fully expect the data for Safari is wrong too.

vinyldarkscratch:
Thanks for tracking this down!  I couldn't find those commits for Chromium, but then again I was also laser-focused on the IDL, thinking it was some IDL-based change.  Since it's not a Firefox-only change and an actual spec change, I think it makes sense to keep it as a subfeature, but the description isn't too helpful -- I've corrected it a bit.

Safari data for `api.ExtendableEvent.waitUntil` is most certainly incorrect as I have just determined, though I think that's another problem for another PR.

api/ExtendableEvent.json
```diff
@@ -150,13 +150,13 @@
         },
         "async_waitUntil": {
           "__compat": {
-            "description": "Asynchronous <code>waitUntil</code>",
+            "description": "<code>waitUntil</code> may be called asynchronously",
```
foolip, line 153: Much better!

foolip (approved):

foolip:
I'm fixing the Safari data in https://github.com/mdn/browser-compat-data/pull/10758.

PR #10742, opened 2021-06-03
adding spec and MDN URLs.

rachelandrew:
I'm documenting OTPCredential. We already have compat data, this PR updates to add the MDN URLs which will exist shortly, and the spec URLs:

Spec: https://wicg.github.io/web-otp/#OTPCredential


Elchi3 (approved):
Nice!! 👍

PR #10743, opened 2021-06-03
Copy data from api.PaymentRequest.onmerchantvalidation to its event

vinyldarkscratch:
This PR copies the data from `api.PaymentRequest.onmerchantvalidation` to `api.PaymentRequest.merchantvalidation_event`.  Looking through Chrome's source, I don't see any indication that the event is supported, and I highly doubt that the event handler would be missing if the event itself were implemented.

### DIFF
--- a/api/PaymentRequest.json
+++ b/api/PaymentRequest.json
@@ -263,13 +263,13 @@
           "description": "<code>merchantvalidation</code> event",
           "support": {
             "chrome": {
-              "version_added": null
+              "version_added": false
             },
             "chrome_android": {
-              "version_added": null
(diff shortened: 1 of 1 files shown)


foolip (approved):

foolip:
> I highly doubt that the event handler would be missing if the event itself were implemented

This has happened a lot actually, but one can spot it with `git grep -l merchantvalidation` in Chromium source. In this case that leads to the discovery that Chromium has a "PaymentRequestMerchantValidationEvent" flag which is not enabled in stable yet. So while flag data could be added here, I'm happy to merge it as false.

PR #10745, opened 2021-06-03
Remove irrelevant MediaStreamTrack entries

foolip:
These are all false, null or removed, with the exception of Edge 12 for
readonly, but that's gone in Chromium-based Edge.

### DIFF
--- a/api/MediaStreamTrack.json
+++ b/api/MediaStreamTrack.json
@@ -440,57 +440,6 @@
           }
         }
       },
-      "getSources": {
-        "__compat": {
-          "support": {
-            "chrome": {
-              "version_added": true,
-              "version_removed": "56"
-            },
-            "chrome_android": {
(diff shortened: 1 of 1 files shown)


api/MediaStreamTrack.json
```diff
@@ -440,57 +440,6 @@
...
-              "version_removed": "43"
-            },
-            "safari": {
-              "version_added": false
-            },
-            "safari_ios": {
-              "version_added": false
-            },
-            "samsunginternet_android": {
-              "version_added": "6.0"
```
foolip, line 480: I didn't mention this in the commit message, but it must be wrong.

foolip:
https://github.com/mdn/content/pull/5629 cleans up MDN to match.

vinyldarkscratch (approved):
Thanks for performing this cleanup, looks good to me!

PR #10748, opened 2021-06-03
Ensure that extend's second argument is not undefined

foolip:
Most importantly, be explicit about skipping things that aren't
directories or *.json files. Also skip extend() for invalid JSON and
explain why that's actually OK.


ddbeck (approved):
I like these early returns! It's much clearer about intent. 👍

Elchi3 (approved):

PR #10751, opened 2021-06-03
Update padding-inline.json

mattbrundage:
Chromium Edge support


foolip (approved):
Thanks @mattbrundage!

PR #10758, opened 2021-06-04
Update data for Service Worker APIs shipped in Safari 11.1 / iOS 11.3

foolip:
The Safari 11.1 data comes from mdn-bcd-collector 3.1.4 results for
Safari 11 and 11.1. The iOS 11.3 data was confirmed by older collector
results for just iOS 11.3:
https://github.com/foolip/mdn-bcd-results/pull/1815

Other than the tests being wrong, the only room for error is that some
of this could have been shipped earlier in iOS, but in context that is
very unlikely and not worth verifying further.

Finally, the async_waitUntil is based on WebKit source, where the
current comment in ExtendableEvent::waitUntil makes it clear it's
supported, and the latest change to it was in trunk version 605.1.13
which predates the Safari 11.1 version:
https://trac.webkit.org/changeset/224584/webkit#file19
https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Configurations/Version.xcconfig?rev=224584

### DIFF
--- a/api/ExtendableEvent.json
+++ b/api/ExtendableEvent.json
@@ -31,10 +31,10 @@
             "version_added": "24"
           },
           "safari": {
-            "version_added": false
+            "version_added": "11.1"
           },
           "safari_ios": {
-            "version_added": false
+            "version_added": "11.3"
           },
           "samsunginternet_android": {
             "version_added": "4.0"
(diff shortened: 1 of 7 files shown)


foolip:
@gsnedders FYI this is a typical example of working from WebKit source to a very plausible Safari release, where just a somewhat accurate `engine_version` was enough pin it down.

vinyldarkscratch (approved):
Looking good to me, thank you!

PR #10762, opened 2021-06-04
Add all browsers versions for Point API

vinyldarkscratch:
This PR adds real values for all browsers for the `Point` API, based upon manual testing.

Test Code Used: `new WebKitPoint(); new MSPoint(); new OPoint(); new Point(); // AKA, trying all the different prefix variants in each browser`


foolip (approved):
I think we should combine this feature with `DOMPoint`, but this makes it clear what the support of `WebKitPoint` and `Point` has looked like, making it easier to do that merging.

PR #10763, opened 2021-06-04
Add Chromium versions for PushSubscriptionChangeEvent API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `PushSubscriptionChangeEvent` API, based upon IDL source code.  Running the mdn-bcd-collector reveals no support by default, and there are no flags to enable support (as far as I can tell).  It appears that this is implemented and supported by a runtime flag, which we aren't tracking.

Source: https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/push_messaging/push_subscription_change_event.idl

### DIFF
--- a/api/PushSubscriptionChangeEvent.json
+++ b/api/PushSubscriptionChangeEvent.json
@@ -5,13 +5,13 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/PushSubscriptionChangeEvent",
         "support": {
           "chrome": {
-            "version_added": null
+            "version_added": false
           },
           "chrome_android": {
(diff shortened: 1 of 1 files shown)


foolip (approved):
I can confirm, the "PushMessagingSubscriptionChange" feature is marked experimental:
https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/platform/runtime_enabled_features.json5;l=1693;drc=1b3a56de114a90840edbb2a5f42b41408c563e37

PR #10764, opened 2021-06-04
Update Chromium versions for ReadableStreamDefaultController API

vinyldarkscratch:
This PR updates and corrects the real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `ReadableStreamDefaultController` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.4).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/ReadableStreamDefaultController

### DIFF
--- a/api/ReadableStreamDefaultController.json
+++ b/api/ReadableStreamDefaultController.json
@@ -6,13 +6,13 @@
         "spec_url": "https://streams.spec.whatwg.org/#rs-default-controller-class",
         "support": {
           "chrome": {
-            "version_added": "52"
+            "version_added": "89"
           },
           "chrome_android": {
-            "version_added": "52"
+            "version_added": "89"
           },
           "edge": {
-            "version_added": "≤79"
+            "version_added": "89"
           },
           "firefox": {
             "version_added": "65"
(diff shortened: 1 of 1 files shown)


vinyldarkscratch:
I have no idea why this was originally set to Chrome 52, though it appears those changes came from the original wiki table.  It wasn't until [Chrome 74](https://storage.googleapis.com/chromium-find-releases-static/dac.html#dac55f7b3ba2471df01f6bfa414f7d73423254fc) that the IDL files were even implemented, let alone exposed.  The results from the collector match [ChromeStatus](https://www.chromestatus.com/feature/5711333280448512) as well, so I'm trusting our own results more.

api/ReadableStreamDefaultController.json
```diff
@@ -36,10 +36,10 @@
             "version_added": false
           },
           "samsunginternet_android": {
-            "version_added": "6.0"
+            "version_added": false
```
foolip, line 39: Why false here?

vinyldarkscratch: The latest Samsung Internet release is based on Chrome 87.

foolip: Aha, thanks!

foolip (approved):

PR #10766, opened 2021-06-04
Add Chromium versions for RTCIdentityProviderGlobalScope API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `RTCIdentityProviderGlobalScope` API, based upon commit history and date.

Commit: No IDL in the source code, other than the web tests


foolip (approved):
Looks like this also isn't in Gecko source, so a candidate for removal when all the false values have been filled out :)

PR #10769, opened 2021-06-04
Chrome doesn't implement strict MIME checks for Workers

vinyldarkscratch:
This PR sets Chromium to false for `api.[Shared]Worker.[Shared]Worker.mime_checks`.  Looking through the code, I can't find anything that suggests that Chrome checks the MIME type and enforces `text/javascript`.  Additionally, [wpt.fyi](https://wpt.fyi/results/workers/Worker_script_mimetype.htm?label=master&product=chrome%5Bexperimental%5D&product=chrome%5Bstable%5D&aligned) indicates that in the latest stable Chrome, it does not.

### DIFF
--- a/api/SharedWorker.json
+++ b/api/SharedWorker.json
@@ -110,13 +110,13 @@
             "description": "Strict MIME type checks for shared worker scripts",
             "support": {
               "chrome": {
-                "version_added": null
+                "version_added": false
               },
               "chrome_android": {
                 "version_added": false
               },
               "edge": {
-                "version_added": null
+                "version_added": false
               },
               "firefox": {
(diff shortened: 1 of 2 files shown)


foolip:
Given your wpt.fyi link shows different results for stable and dev, this must have changed recently. Or so I thought.

Via bug search I found my way to https://bugs.chromium.org/p/chromium/issues/detail?id=794548, but it turns out a "StrictMimeTypesForWorkers" feature was added which is still experimental. Still, if you could link to crbug.com/794548 in a note, that would be great for updating this data in the future.

foolip (approved):

PR #10770, opened 2021-06-04
Add Chromium versions for api.SourceBuffer.trackDefaults

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `trackDefaults` member of the `SourceBuffer` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.4).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/SourceBuffer/trackDefaults


foolip (approved):
Yep, this is behind a "MediaSourceExperimental" feature in Chromium which is still experimental.

PR #10775, opened 2021-06-04
Add Chromium/Safari versions for api.Window.postMessage.transfer_argument_support

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) and Safari (Desktop and iOS/iPadOS) for the `postMessage.transfer_argument_support` member of the `Window` API, based upon commit history and date.

Commit: https://source.chromium.org/chromium/chromium/src/+/d0cf776d6ff039396acb08430e64567eb311c08a (by commit date of September 3, 2009 for Chrome 4)

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -6872,10 +6872,10 @@
             "description": "<code>transfer</code> argument",
             "support": {
               "chrome": {
-                "version_added": "38"
+                "version_added": "4"
               },
               "chrome_android": {
-                "version_added": "38"
+                "version_added": "18"
               },
               "edge": {
                 "version_added": "12"
@@ -6890,10 +6890,10 @@
                 "version_added": true
               },
               "opera": {
-                "version_added": "25"
+                "version_added": "≤15"
               },
               "opera_android": {
-                "version_added": "25"
+                "version_added": "≤14"
               },
               "safari": {
                 "version_added": null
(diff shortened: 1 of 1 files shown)


138166234:
1381662341mohamadsalmanpor

در تاریخ جمعه ۴ ژوئن ۲۰۲۱،‏ ۶:۵۰ بعدازظهر Queen Vinyl Darkscratch <
***@***.***> نوشت:

foolip:
I'm afraid that https://source.chromium.org/chromium/chromium/src/+/38b7b40abe5848dc30fdba9496bfc6282795ca6d is just refactoring, not the commit that introduced the behavior.

foolip:
I think as step 1 we'd have to figure out which of these two this entry refers to:
- `postMessage('', '*', [stuff])`
- `postMessage('', { targetOrigin: '*', transfer: [stuff] })`

vinyldarkscratch:
I'm pretty certain this is referring to the first of those two, based upon the wording used (particularly the use of the word "argument").  Looking at this a little closer, I see why I was confused about which commit implemented the functionality, since it was called `messagePorts` before.  Fixing now!

foolip:
The commit you found is this one in WebKit:
https://trac.webkit.org/changeset/48025/webkit
https://trac.webkit.org/browser/webkit/trunk/WebCore/Configurations/Version.xcconfig?rev=48025 (532.0.0)

Based on that I think we can also fill in Safari 5 / iOS 4.

foolip (approved):

PR #10776, opened 2021-06-04
Firefox 90 aliases webkit-image-set() to image-set()

rachelandrew:
Working on https://github.com/mdn/content/issues/5379

Adds a note that `webkit-image-set()` has been aliased to `image-set()`.

### DIFF
--- a/css/types/image.json
+++ b/css/types/image.json
@@ -1930,21 +1930,32 @@
                 "prefix": "-webkit-",
                 "version_added": "79"
               },
-              "firefox": {
-                "version_added": "88",
-                "notes": [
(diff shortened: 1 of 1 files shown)


foolip (commented):
Can this be done as a separate entry with `"prefix": "-webkit-"` instead? And I presume the same change applies to Firefox on Android?

rachelandrew:
@foolip it seemed more like a note type thing, but I'm not sure. @ddbeck do you have an opinion on what is best here?

foolip:
I've gone for entries for any prefixes or alternative names, but I haven't touched the CSS data much. I'll defer to @ddbeck.

ddbeck:
@rachelandrew I agree with @foolip here. This note reads to me that it would be a second support statement (after the first), with `"version_added": 90`, a prefix, and no note needed. I revised the PR title to reflect that. Thank you!

foolip (approved):

PR #10782, opened 2021-06-05
Remove Element's currentStyle and runtimeStyle as irrelevant

foolip:
https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-features

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -2537,56 +2537,6 @@
           }
         }
       },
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
I'm going ahead with these, since the content removals have been merged. Thank you! 🚀

PR #10784, opened 2021-06-05
Add WebView versions for api.XRSessionMode.immersive-ar

vinyldarkscratch:
This PR adds real values for WebView Android for the `immersive-ar` member of the `XRSessionMode` API, based upon manual testing.

Test Code Used: `Copied data from Chrome Android`

### DIFF
--- a/api/XRSessionMode.json
+++ b/api/XRSessionMode.json
@@ -87,7 +87,7 @@
               "version_added": "12.1"
             },
             "webview_android": {
-              "version_added": null
+              "version_added": "81"
             }
           },
           "status": {


foolip (commented):
If https://developer.mozilla.org/en-US/docs/Web/API/Navigator/xr is right XR isn't supported on WebView. Can you check which is right?

vinyldarkscratch:
XR is indeed unsupported on WebView, however this is an interface that slipped through the cracks and wasn't disabled, so I'm going off of that.

foolip:
`XRSessionMode` is an enum in spec and implementation:
https://immersive-web.github.io/webxr/#xrsessionmode-enum
https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/xr/xr_session.idl;l=7;drc=d78a26c4267a178f6ceefb599478692907cd6206

An enum is basically a subset of a string, where only certain strings are allowed and others cause `TypeError` to be thrown. So what it means for `XRSessionMode` is really for these strings to work as arguments for one of the two methods here:
https://immersive-web.github.io/webxr/#xrsystem-interface

If those methods aren't supported in WebView, then this enum can't be either.

Aside: I'm skeptical about having enums in BCD at all, but that's for https://github.com/mdn/browser-compat-data/issues/6810.

vinyldarkscratch:
My mistake, I thought I had tested this in the collector as well. I've followed this up with #10926 to set this all to `false`.

PR #10787, opened 2021-06-06
Add Chromium versions for SourceBuffer API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `SourceBuffer` API, based upon information in a tracking bug.

Tracking Bug: https://www.chromestatus.com/feature/5696773133697024

### DIFF
--- a/api/SourceBuffer.json
+++ b/api/SourceBuffer.json
@@ -626,6 +626,9 @@
             "chrome": {
               "version_added": "53"
             },
+            "chrome_android": {
+              "version_added": "53"
+            },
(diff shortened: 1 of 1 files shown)


vinyldarkscratch:
Some additional context: the Chrome data was updated in #1949 to change the version number to 53.  However, the Opera and WebView data weren't updated in that PR.  This is intended to fix that data, along with adding the Chrome Android and Samsung Internet data.

foolip (approved):

PR #10789, opened 2021-06-06
Add Chromium/Safari versions for api.Element.mousewheel_event

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) and Safari for the `mousewheel_event` member of the `Element` API by mirroring the data.

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -5413,10 +5413,10 @@
           "description": "<code>mousewheel</code> event",
           "support": {
             "chrome": {
-              "version_added": "31"
+              "version_added": "1"
             },
             "chrome_android": {
-              "version_added": "31"
+              "version_added": "18"
             },
             "edge": {
               "version_added": "≤79"
@@ -5431,19 +5431,19 @@
               "version_added": null
             },
             "opera": {
-              "version_added": "18"
+              "version_added": "≤15"
             },
             "opera_android": {
-              "version_added": "18"
+              "version_added": "≤14"
             },
             "safari": {
-              "version_added": null
+              "version_added": "3"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "1"
             },
             "samsunginternet_android": {
-              "version_added": "2.0"
+              "version_added": "1.0"
             },
             "webview_android": {
               "version_added": "4.4.3"
--- a/api/Element.json
+++ b/api/Element.json
@@ -5446,7 +5446,7 @@
               "version_added": "1.0"
             },
             "webview_android": {
-              "version_added": "4.4.3"
+              "version_added": "1"
             }
           },
           "status": {
--- a/api/Element.json
+++ b/api/Element.json
@@ -5416,7 +5416,7 @@
               "version_added": "31"
             },
             "chrome_android": {
-              "version_added": null
+              "version_added": "31"
             },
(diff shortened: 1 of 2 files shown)


api/Element.json
```diff
@@ -5416,7 +5416,7 @@
               "version_added": "31"
             },
             "chrome_android": {
-              "version_added": null
+              "version_added": "31"
```
foolip, line 5419: Can the mousewheel event actually be fired on Chrome for Android? I imagine this has to depend on mouse support in Android, and that didn't always exist?

vinyldarkscratch: I'm not sure exactly when Android supported a mouse, but I found this article that [talks about their mouse support in Android 4.0](https://kelbin.medium.com/mouse-support-in-android-b87f0c95b78d).  Considering the time when Android 4.0 was released (Q4 2011) and the time Chrome 31 was released (Q4 2013), this seems reasonable that the event would be supported at the same time as it was implemented in Chrome.

foolip: Good digging, that makes this very plausible indeed.

foolip:
But now I really doubt the data being mirrored here. http://mdn-bcd-collector.appspot.com/tests/api/HTMLElement/onmousewheel tests true in Chrome 30. Can you investigate?

vinyldarkscratch:
Looks like the data was indeed incorrect, this event was supported all the way back in Chrome 1.  I've updated the data accordingly.

api/Element.json
```diff
@@ -5431,22 +5431,22 @@
               "version_added": null
             },
             "opera": {
-              "version_added": null
+              "version_added": "≤15"
```
foolip, line 5434: I'm positive this will go back far in Presto too, but a range for now means we'll come back to it.

api/Element.json
```diff
@@ -5431,22 +5431,22 @@
...
             },
             "opera_android": {
-              "version_added": null
+              "version_added": "≤14"
             },
             "safari": {
-              "version_added": null
+              "version_added": "3"
```
foolip, line 5440: Is this based on Chrome 1 timeline, or testing in Safari?

api/Element.json
```diff
@@ -5431,22 +5431,22 @@
...
-              "version_added": null
+              "version_added": "1"
             },
             "samsunginternet_android": {
-              "version_added": null
+              "version_added": "1.0"
             },
             "webview_android": {
-              "version_added": null
+              "version_added": "4.4.3"
```
foolip, line 5449: Maybe <=37 if this isn't based on testing?

foolip (changes requested):

api/Element.json
vinyldarkscratch: This was based upon testing in Safari!

api/Element.json
vinyldarkscratch: Whoops, meant to change this to 1!  (I haven't tested it, but I'm pretty confident this would be in Android 1 based upon the engine versions in relation to Safari 3.)

api/Element.json
foolip: Great, thanks!

foolip (approved):

PR #10792, opened 2021-06-06
Add IE/Edge versions for api.CanvasRenderingContext2D.drawImage

vinyldarkscratch:
This PR adds real values for Internet Explorer and Edge for the `drawImage` member of the `CanvasRenderingContext2D` API, based upon a combination of manual testing, and the time of implementation in other browsers, including Firefox and Edge.

### DIFF
--- a/api/CanvasRenderingContext2D.json
+++ b/api/CanvasRenderingContext2D.json
@@ -1137,7 +1137,7 @@
                 "version_added": "42"
               },
               "ie": {
-                "version_added": null
+                "version_added": false
               },
               "opera": {
                 "version_added": "17"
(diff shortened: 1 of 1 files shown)


foolip:
> I've set Internet Explorer to false based upon the time of implementation in other browsers, including Firefox and Edge.

I think I don't understand, is Firefox really relevant here?

vinyldarkscratch:
What I mean is that considering how recent the implementation has been in other browsers, I find it highly unlikely that these features were around in IE 11 or before.

foolip:
Two of the Edge versions here are ≤79, can we pin those down first? If it's actually in Edge 12, then this may well be in IE too.

foolip (approved):

PR #10793, opened 2021-06-06
Update Chromium versions for api.ServiceWorkerGlobalScope.onpush

vinyldarkscratch:
This PR updates and corrects the real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `onpush` member of the `ServiceWorkerGlobalScope` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.4).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/ServiceWorkerGlobalScope/onpush

### DIFF
--- a/api/ServiceWorkerGlobalScope.json
+++ b/api/ServiceWorkerGlobalScope.json
@@ -1298,7 +1298,7 @@
               "version_added": "42"
             },
             "chrome_android": {
-              "version_added": "40"
+              "version_added": "42"
             },
             "edge": {
               "version_added": "≤79"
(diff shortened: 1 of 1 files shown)


foolip (changes requested):
This matches https://developer.mozilla.org/en-US/docs/Web/API/PushEvent and seems plausible. Can you update Android and mirrored data too?

foolip (approved):

Issue #7354, opened 2020-11-13
Opera for Android appears to not support flags


ddbeck:
Like Samsung Internet (#4753), Opera for Android appears to expose a very limited set of flags—not the full complement like Chrome for Android does. We should probably remove and bar flag data in `"opera_android"` entries.

vinyldarkscratch:
Can confirm, viewing `opera://flags` reveal only one flag on Android (as opposed to the near match of entries that Chrome has).

![screenshot-1605292332952](https://user-images.githubusercontent.com/5179191/99108000-8e26d400-259b-11eb-88f8-a84d5173923e.jpg)
![Screenshot 2020-11-13 103442](https://user-images.githubusercontent.com/5179191/99108256-f1186b00-259b-11eb-8451-2435b6fdb28c.png)

jculver050:
Close

PR #10795, opened 2021-06-06
Update browser support for inset css properties

lukewarlow:
#### Summary

Updated browser support for the inset css properties.

Chrome 87 added support, this MR just mirrors this to other chromium based browsers.

### DIFF
--- a/css/properties/inset-block-end.json
+++ b/css/properties/inset-block-end.json
@@ -36,16 +36,21 @@
                 ]
               }
             ],
-            "edge": {
-              "version_added": "79",
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "enable-experimental-web-platform-features",
-                  "value_to_set": "enabled"
-                }
-              ]
-            },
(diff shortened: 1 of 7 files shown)


css/properties/inset-block-end.json
```diff
@@ -84,24 +89,29 @@
...
-                  "value_to_set": "enabled"
-                }
-              ]
-            },
+            "opera_android": [
+              {
+                "version_added": "62"
+              },
+              {
+                "version_added": "48",
```
ddbeck, line 97: Opera on Android doesn't really support flags (we can't formally block them yet; see https://github.com/mdn/browser-compat-data/issues/7354, https://github.com/mdn/browser-compat-data/issues/10616). Would you mind dropping the flagged support statements from all the Opera for Android changes in this PR?

ddbeck (changes requested):

ddbeck (approved):
Looks good. Thank you! 🎉

PR #10797, opened 2021-06-07
Add IE versions for api.GlobalEventHandlers.onanimationcancel

vinyldarkscratch:
This PR adds real values for Internet Explorer for the `onanimationcancel` member of the `GlobalEventHandlers` API by mirroring the data.

### DIFF
--- a/api/GlobalEventHandlers.json
+++ b/api/GlobalEventHandlers.json
@@ -118,7 +118,7 @@
               "version_added": "54"
             },
             "ie": {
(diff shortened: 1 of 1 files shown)


foolip (approved):
Even though https://developer.mozilla.org/en-US/docs/Web/CSS/animation#browser_compatibility says IE10 supports the `animation` CSS property, this is consistent with all other data about the `animationcancel` event.

PR #10799, opened 2021-06-07
Add Chromium versions for DOMPointInit API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `DOMPointInit` API.  This dictionary was added into Chrome a little after DOMPoint was implemented but before it was exposed by default, so tI'm confident to set this to the same versions as DOMPoint.  See https://source.chromium.org/chromium/chromium/src/+/a23c0523692fe0ddf105a1cfbf63ca02af555c31 for the exact commit.

### DIFF
--- a/api/DOMPointInit.json
+++ b/api/DOMPointInit.json
@@ -6,13 +6,13 @@
         "spec_url": "https://drafts.fxtf.org/geometry/#dictdef-dompointinit",
         "support": {
           "chrome": {
-            "version_added": null
+            "version_added": "61"
           },
           "chrome_android": {
-            "version_added": null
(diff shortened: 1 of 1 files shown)


foolip:
https://source.chromium.org/chromium/chromium/src/+/a23c0523692fe0ddf105a1cfbf63ca02af555c31 was refactoring to use Web IDL for this dictionary, but even before it all the observable behavior of `DOMPointInit` was there. But in any case, if it happened before this shipped, then there's no doubt.

foolip (approved):

PR #10800, opened 2021-06-07
Add all browsers versions for api.Window.open.once_per_event

vinyldarkscratch:
This PR adds real values for all browsers for the `open.once_per_event` member of the `Window` API, based upon manual testing.

Test Code Used:
```js
document.getElementById('hello').onclick = function(e) {
  window.open('https://www.google.com');
  window.open('https://developer.mozilla.org');
}

// Based results off of whether one or both tabs opened
```

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -5800,13 +5800,13 @@
             "description": "One <code>Window.open()</code> call per event",
             "support": {
               "chrome": {
-                "version_added": null
+                "version_added": "23"
               },
               "chrome_android": {
-                "version_added": null
+                "version_added": "25"
               },
               "edge": {
-                "version_added": null
+                "version_added": "12"
               },
               "firefox": {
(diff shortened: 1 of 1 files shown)


api/Window.json
```diff
@@ -5815,25 +5815,25 @@
...
-                "version_added": null
+                "version_added": "≤12.1"
               },
               "opera_android": {
-                "version_added": null
+                "version_added": "≤12.1"
               },
               "safari": {
-                "version_added": null
+                "version_added": false
```
foolip, line 5827: I'm surprised but can confirm that Safari 14.1 on macOS will open both.

foolip (approved):
Excellent testing, thank you!

foolip:
@vinyldarkscratch just a note that the code formatting in the commit message was a bit weird and I fixed it up manually when merging.

PR #10809, opened 2021-06-07
Add IE versions for RequestDestination API

vinyldarkscratch:
This PR adds real values for Internet Explorer for the `RequestDestination` API by mirroring the data from Edge (AKA, IE can't be truthy if Edge is greater than 12).

### DIFF
--- a/api/RequestDestination.json
+++ b/api/RequestDestination.json
@@ -21,7 +21,7 @@
             "version_added": "61"
           },
           "ie": {
-            "version_added": null
+            "version_added": false
           },
           "opera": {
             "version_added": "52"


foolip (approved):
This looks right to me, but I think the description might mislead future readers. The mirroring script wasn't used here was it? IE is a source browser, it can't be mirrored...

vinyldarkscratch:
Description updated!

foolip:
Ah, I get it. While I'm sure there's some exception to this rule, that's a solid rule of thumb.

foolip (approved):
This is an enum, but it matches the data for https://developer.mozilla.org/en-US/docs/Web/API/Request/destination#browser_compatibility where the enum is used.

PR #10811, opened 2021-06-07
Add IE versions for SourceBuffer API

vinyldarkscratch:
This PR adds real values for Internet Explorer for the `SourceBuffer` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.4).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/SourceBuffer

### DIFF
--- a/api/SourceBuffer.json
+++ b/api/SourceBuffer.json
@@ -262,7 +262,7 @@
               "version_added": false
             },
             "ie": {
-              "version_added": null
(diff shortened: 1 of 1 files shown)


foolip (approved):
I've gone ahead and added the same "Only works on Windows 8+." note as all the other entries have here.

PR #10815, opened 2021-06-07
Add IE/Edge versions for message/messageerror events

vinyldarkscratch:
This PR adds real values for Internet Explorer and Edge for the `onmessage` and `onmessageerror` members of various APIs, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.4).  Results are manually confirmed for accuracy.

Tests Used:
https://mdn-bcd-collector.appspot.com/tests/api/Window/onmessage
https://mdn-bcd-collector.appspot.com/tests/api/Window/onmessageerror

### DIFF
--- a/api/WindowEventHandlers.json
+++ b/api/WindowEventHandlers.json
@@ -362,7 +362,7 @@
               "version_added": "60"
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "12"
             },
             "firefox": {
               "version_added": null
@@ -371,7 +371,7 @@
               "version_added": null
             },
             "ie": {
-              "version_added": null
+              "version_added": "8"
             },
             "opera": {
               "version_added": "47"
@@ -411,7 +411,7 @@
               "version_added": "60"
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "18"
             },
             "firefox": {
               "version_added": "57"
@@ -420,7 +420,7 @@
               "version_added": "57"
             },
             "ie": {
-              "version_added": null
+              "version_added": false
             },
             "opera": {
               "version_added": "47"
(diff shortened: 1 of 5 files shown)


api/WindowEventHandlers.json
```diff
@@ -411,7 +411,7 @@
               "version_added": "60"
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "18"
```
foolip, line 414: This is updating "onmessageerror", which isn't mentioned in the description.

api/WindowEventHandlers.json
```diff
@@ -467,7 +467,7 @@
                 "version_added": "57"
               },
               "ie": {
-                "version_added": null
+                "version_added": false
```
foolip, line 470: Hah, this is bizarre, a "worker_support" entry on `WindowEventHandlers`. Sorting this out will be fun, but not in this PR I suggest since this is also about the "messageerror" event.

foolip (changes requested):

foolip:
Was https://mdn-bcd-collector.appspot.com/tests/api/Window/onmessage the test used here, or others too?

api/WindowEventHandlers.json
```diff
@@ -371,7 +371,7 @@
               "version_added": null
             },
             "ie": {
-              "version_added": null
+              "version_added": "8"
```
foolip, line 374: Matches https://developer.mozilla.org/en-US/docs/Web/API/Window/postMessage#browser_compatibility, so makes sense. Looks like it was a partial implementation back then, but anyway.

foolip (approved):
This data all makes sense to me. The "message" event should be as old as `postMessage()` and the "messageerror" event is a recent addition for dealing with serialization failures of some sort.

PR #10816, opened 2021-06-07
Add IE/Edge versions for api.WorkerGlobalScope.importScripts.mime_checks

vinyldarkscratch:
This PR adds real values for Internet Explorer and Edge for the `importScripts.mime_checks` member of the `WorkerGlobalScope` API.  Based upon when this security check was implemented in other browsers, it doesn't seem likely that this would be around in Trident/EdgeHTML.

### DIFF
--- a/api/WorkerGlobalScope.json
+++ b/api/WorkerGlobalScope.json
@@ -255,7 +255,7 @@
                 "version_added": "71"
               },
               "edge": {
-                "version_added": "≤79"
+                "version_added": "79"
               },
               "firefox": {
                 "version_added": "67"
@@ -264,7 +264,7 @@
                 "version_added": "67"
               },
               "ie": {
-                "version_added": null
+                "version_added": false
               },
               "opera": {
                 "version_added": "58"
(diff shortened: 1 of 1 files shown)


api/WorkerGlobalScope.json
```diff
@@ -404,7 +404,7 @@
               "version_added": "40"
             },
             "edge": {
-              "version_added": "17"
+              "version_added": "12"
```
foolip, line 407: This and all that follows is about something other than the PR description. Should be reverted?

foolip (changes requested):

foolip:
> Based upon when this security check was implemented in other browsers, it doesn't seem likely that this would be around in Trident/EdgeHTML.

I wouldn't bet on that, actually. This isn't really a new feature, there's been differences in whether `Content-Type` is trusted or if there's sniffing based on the response body in many parts of the platform, and without testing I'm not confident that IE/Edge didn't have the strict behavior here.

vinyldarkscratch:
I found a test from our friends at WPT that tests this very feature: http://wpt.live/workers/importscripts_mime.any.worker.html

I ran it in Edge 18 and found that the majority of the tests don't throw (specifically, the ones where it's confirming non-script types are NOT supported), so that test backs up this PR.

foolip:
Excellent, good detective work!

foolip (approved):

PR #10821, opened 2021-06-07
Update IE versions for api.Document.transition_events

vinyldarkscratch:
This PR updates real values for Internet Explorer for the `transition` events members of the `Document` API.  These events aren't supported in Edge.

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -10969,7 +10969,7 @@
               "version_added": "53"
             },
             "ie": {
-              "version_added": null
+              "version_added": false
             },
             "opera": {
               "version_added": null
@@ -11049,7 +11049,7 @@
               "version_added": "51"
             },
             "ie": {
-              "version_added": null
+              "version_added": false
             },
             "opera": {
               "version_added": null
@@ -11099,7 +11099,7 @@
               "version_added": "53"
             },
             "ie": {
-              "version_added": null
+              "version_added": false
             },
             "opera": {
               "version_added": null
@@ -11179,7 +11179,7 @@
               "version_added": "53"
             },
             "ie": {
-              "version_added": null
+              "version_added": false
             },
             "opera": {
               "version_added": null
--- a/api/Document.json
+++ b/api/Document.json
@@ -11040,7 +11040,7 @@
               "version_added": false
             },
             "edge": {
-              "version_added": false
+              "version_added": "12"
             },
             "firefox": {
               "version_added": "51"
@@ -11049,7 +11049,7 @@
               "version_added": "51"
             },
             "ie": {
-              "version_added": false
+              "version_added": "10"
             },
             "opera": {
               "version_added": null
@@ -11090,7 +11090,7 @@
               "version_added": false
             },
             "edge": {
-              "version_added": false
+              "version_added": "12"
             },
             "firefox": {
               "version_added": "53"
@@ -11099,7 +11099,7 @@
               "version_added": "53"
             },
             "ie": {
-              "version_added": false
+              "version_added": "10"
             },
             "opera": {
               "version_added": null
@@ -11170,7 +11170,7 @@
               "version_added": false
             },
             "edge": {
-              "version_added": false
+              "version_added": "12"
             },
             "firefox": {
               "version_added": "53"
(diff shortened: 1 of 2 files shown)


foolip (approved):

foolip:
A nit about the description:

> These events aren't supported in Edge, and they're too new to be in Internet Explorer.

These events are related to the CSS `transition` property, which actually isn't too new, that was in IE10:
https://developer.mozilla.org/en-US/docs/Web/CSS/transition#browser_compatibility

But the transitioncancel events seem to have been a later addition. Maybe it was added to the spec after IE11?

foolip:
Yep, https://github.com/w3c/csswg-drafts/commit/b18ba175d182128ca2d879555f8a17361483cbaa was in May 2015, so between IE 11 and Edge 12 release dates. It also introduced the transitionstart events.

@vinyldarkscratch perhaps you can turn this into a PR updating transitionstart too, and making the onfoo and foo_event entries match?

vinyldarkscratch:
The `transitionstart` event is already getting updated though?  There also aren't any event handlers for these events (`onfoo`).

foolip:
OK, so once again questioning everything leads to the existing data being wrong. I took the example in https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/transitionstart_event and turned it into this:
https://software.hixie.ch/utilities/js/live-dom-viewer/?saved=9408

Running that in IE11 shows the events are fired. It's just the event handler attributes that aren't supported, so some kind of notes will be needed.

The onfoo attributes for these events are in GlobalEventHandlers.json.

Do you want to update all of this in this PR, or split it up a bit? Either way is fine with me.

vinyldarkscratch:
Running that test in IE 10 reveals that those transition events _are_ supported, fixed!

foolip (approved):
Yay!

foolip:
So before this is merged I think we also need to add notes to explain that the `ontransition*` properties came later, but it's possible to listen for the events using `addEventListener`. And maybe double check that the `ontransition*` data is in fact correct so we don't just add to lies and confusion :)

vinyldarkscratch:
The event handlers are indeed not supported in IE.  Adding notes to describe as such now!

api/Document.json
```diff
@@ -11049,7 +11049,8 @@
               "version_added": "51"
             },
             "ie": {
-              "version_added": "10"
+              "version_added": "10",
+              "notes": "This event's handler attribute (<code>ontransitionend</code>) is not supported; however, the event itself is supported since IE 10. The event can be listened to via <code>document.addEventListener('transitionend', function() {});</code>."
```
foolip, line 11053: Can you match the wording of the visibilitychange_event entry, or change it to match this if you think it's better like this?

foolip (changes requested):

foolip:
Can you also add notes to the ontransition* entries that presumably are just false now? There should be existing note wording to copy elsewhere.

api/GlobalEventHandlers.json
```diff
@@ -4975,7 +4976,8 @@
               "version_added": "53"
             },
             "ie": {
-              "version_added": false
+              "version_added": false,
+              "notes": "The <code>ontransitionrun</code> attribute is not supported in IE. To listen to this event, use <code>document.addEventListener('transitionend', function() {});</code>."
```
foolip, line 4980:
```suggestion
              "notes": "The <code>ontransitionrun</code> attribute is not supported in IE. To listen to this event, use <code>document.addEventListener('transitionrun', function() {});</code>."
```

api/GlobalEventHandlers.json
```diff
@@ -5037,7 +5039,8 @@
               "version_added": "53"
             },
             "ie": {
-              "version_added": false
+              "version_added": false,
+              "notes": "The <code>ontransitionstart</code> attribute is not supported in IE. To listen to this event, use <code>document.addEventListener('transitionend', function() {});</code>."
```
foolip, line 5043:
```suggestion
              "notes": "The <code>ontransitionstart</code> attribute is not supported in IE. To listen to this event, use <code>document.addEventListener('transitionstart', function() {});</code>."
```

foolip (approved):

PR #10825, opened 2021-06-07
Add Edge versions for SubmitEvent API

vinyldarkscratch:
This PR adds real values for Microsoft Edge for the `SubmitEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.4).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/SubmitEvent

### DIFF
--- a/api/SubmitEvent.json
+++ b/api/SubmitEvent.json
@@ -12,7 +12,7 @@
             "version_added": "81"
           },
           "edge": {
-            "version_added": null
+            "version_added": "81"
           },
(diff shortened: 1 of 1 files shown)


foolip (approved):
This was introduced in the spec in https://github.com/whatwg/html/pull/4984, so makes sense that this can't have been in Edge 18 or earlier.

PR #10837, opened 2021-06-07
Add IE versions for StorageEvent API

vinyldarkscratch:
This PR adds real values for Internet Explorer for the `StorageEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.4).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/StorageEvent

### DIFF
--- a/api/StorageEvent.json
+++ b/api/StorageEvent.json
@@ -21,7 +21,7 @@
             "version_added": true
           },
           "ie": {
-            "version_added": null
+            "version_added": "9"
           },
           "opera": {
             "version_added": true
@@ -69,7 +69,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": null
+              "version_added": false
             },
             "opera": {
               "version_added": true
@@ -117,7 +117,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": null
+              "version_added": "9"
             },
             "opera": {
               "version_added": true
@@ -165,7 +165,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": null
+              "version_added": "9"
             },
             "opera": {
               "version_added": true
@@ -213,7 +213,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": null
+              "version_added": "9"
             },
             "opera": {
               "version_added": true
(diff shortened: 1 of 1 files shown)


foolip (commented):
https://caniuse.com/namevalue-storage and https://developer.mozilla.org/en-US/docs/Web/API/Window/localStorage#browser_compatibility  say IE8 for the API that would fire this event.

Can you try making the event fire somehow to see if this is a question of a missing interface object?

vinyldarkscratch:
In #10183, the `onstorage` event handler and the `storage` event were set to IE 9, so I'd say this makes sense.  (I'm having troubles setting up a test case too...)

foolip:
I'm too stubborn/curious for my own good. I found https://docs.microsoft.com/en-us/previous-versions/windows/internet-explorer/ie-developer/platform-apis/cc289326(v=vs.85) which says IE8, and there's enough clues in there to arrive at this test case:
```html
<!DOCTYPE html>
<button id=b1>setItem</button>
<button id=b2>clear</button>
<div id=log></div>
<script>
alert(document.onstorage)
document.onstorage = function() {
  alert(event.key + ' was updated');
}; 
document.getElementById('b1').onclick = function() {
  localStorage.setItem('key', 'value');
  alert('set done');
};
document.getElementById('b2').onclick = function() {
  localStorage.clear();
  alert('clear done');
};
</script>
```

Loading that in *2* tabs in IE shows that:
- `document.onstorage` is supported
- The event is fired in the other tab when clicking "set"
- The event that's fired doesn't have a `key` property, but it does have a `url` property. It seems seems to be just an `Event` though, I guess in IE8 they just added properties to the event instances or something like that.

IE8 doesn't support `addEventListener`, but I also couldn't get this to work with `attachEvent`. Maybe I did it wrong and `document.attachEvent('storage', ...)` works, but it doesn't really matter so not checked further.

So how should one summarize this situation? In the end, since the event is fired on `document` instead of [the global object as per spec](https://html.spec.whatwg.org/multipage/webstorage.html#concept-storage-broadcast) and it doesn't have the same properties, I think we should say that today's notion of the event was not supported, even partially, in IE8. So IE9 is correct.

foolip (approved):

PR #10840, opened 2021-06-07
Add Firefox versions for DeviceMotionEventAcceleration API

vinyldarkscratch:
This PR adds real values for Firefox and Firefox Android for the `DeviceMotionEventAcceleration` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.4).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/DeviceMotionEventAcceleration

### DIFF
--- a/api/DeviceMotionEventAcceleration.json
+++ b/api/DeviceMotionEventAcceleration.json
@@ -23,10 +23,14 @@
             "version_removed": "79"
           },
           "firefox": {
-            "version_added": false
+            "version_added": "6",
+            "partial_implementation": true,
+            "notes": "Doesn't expose the <code>DeviceMotionEventAcceleration</code> interface itself."
           },
           "firefox_android": {
-            "version_added": false
+            "version_added": "6",
+            "partial_implementation": true,
+            "notes": "Doesn't expose the <code>DeviceMotionEventAcceleration</code> interface itself."
           },
           "ie": {
             "version_added": false
@@ -88,10 +92,10 @@
               "version_added": false
             },
             "firefox": {
-              "version_added": false
+              "version_added": null
             },
             "firefox_android": {
-              "version_added": false
+              "version_added": null
             },
             "ie": {
               "version_added": false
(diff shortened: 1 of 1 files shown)


foolip (changes requested):
I'm pretty sure this is a non-exposed interface, that the data should match https://developer.mozilla.org/en-US/docs/Web/API/DeviceMotionEvent#browser_compatibility and there should be notes like the ones for Safari here.

vinyldarkscratch:
Yeah, this checks out, looks like this is not exposed as per the [IDL](https://searchfox.org/mozilla-central/source/dom/webidl/DeviceMotionEvent.webidl#9).  I don't really know to fire an event and test this on older Firefox versions, but I feel it's safe to mirror the data from `api.DeviceMotionEvent.acceleration`.

foolip:
Yes, if you trust the `acceleration` property data, then mirroring it seems like a pretty safe bet.

vinyldarkscratch:
I double-checked the `acceleration` property and confirmed the accuracy of the data, so I'd say this is ready to go!

api/DeviceMotionEventAcceleration.json
```diff
@@ -23,10 +23,14 @@
             "version_removed": "79"
           },
           "firefox": {
-            "version_added": null
+            "version_added": "6",
+            "partial_implementation": true,
+            "notes": "Doesn't expose the <code>DeviceMotionEventAcceleration</code> interface itself."
```
foolip, line 28: This wording is different from some other similar cases. Can you search for "expose" and "interface" in notes to see if you can find the most common wording to copy?

api/DeviceMotionEventAcceleration.json
```diff
@@ -23,10 +23,14 @@
...
           },
           "firefox_android": {
-            "version_added": null
+            "version_added": "6",
+            "partial_implementation": true,
+            "notes": "Doesn't expose the <code>DeviceMotionEventAcceleration</code> interface itself."
```
foolip, line 33: Let's have the note only on the interface entry itself, not on any of the members.

foolip (changes requested):

api/DeviceMotionEventAcceleration.json
vinyldarkscratch: I was copying the notes from Safari, which seems to have the same behavior.  It doesn't seem like there's many of those notes around...

foolip: OK, let's go with this, I'm sure I'll run into the notes I'm thinking of at some point and make them consistent if needed.

foolip (approved):

PR #10841, opened 2021-06-07
Update Firefox versions for BlobEvent API

vinyldarkscratch:
This PR updates and corrects the real values for Firefox and Firefox Android for the `BlobEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.4).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/BlobEvent

### DIFF
--- a/api/BlobEvent.json
+++ b/api/BlobEvent.json
@@ -15,10 +15,10 @@
             "version_added": "79"
           },
           "firefox": {
-            "version_added": "22"
+            "version_added": "21"
           },
           "firefox_android": {
-            "version_added": "22"
+            "version_added": "21"
           },
           "ie": {
             "version_added": false
@@ -64,10 +64,10 @@
               "version_added": "79"
             },
             "firefox": {
-              "version_added": "22"
+              "version_added": "21"
             },
             "firefox_android": {
-              "version_added": "22"
+              "version_added": "21"
             },
             "ie": {
               "version_added": false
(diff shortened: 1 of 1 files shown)


foolip (commented):
https://developer.mozilla.org/en-US/docs/Web/API/MediaRecorder#browser_compatibility is the API that fires this event if I'm not mistaken.

What's up with this event being exposed before that? If it was just an exposed interface but it could never be fired, then maybe a `partial_implementation` entry would be appropriate?

vinyldarkscratch:
I'm not certain, it seems like this API is a part of the spec for the media recorder, so it doesn't make too much sense this would be implemented prior...though it is.  What do you think we should do?

foolip:
Can you double check the entry point for `MediaRecorder` to confirm that data is correct?

If so this was probably just accidentally exposed too early and I say let's pretend it wasn't and match the other data.

vinyldarkscratch:
The data for `MediaRecorder` is correct, so this is a case of early exposure, it seems.  Updating now!

api/BlobEvent.json
```diff
@@ -162,10 +162,10 @@
               "version_added": "79"
             },
             "firefox": {
-              "version_added": null
+              "version_added": false
```
foolip, line 165: Indeed no `timecode`:
https://github.com/mozilla/gecko-dev/blob/81c32a2ea5605c5cb22bd02d28c362c140b5cfb4/dom/webidl/BlobEvent.webidl#L11-L15

foolip (approved):

PR #10843, opened 2021-06-07
Add Firefox versions for DeviceOrientationEvent API

vinyldarkscratch:
This PR adds real values for Firefox and Firefox Android for the `DeviceOrientationEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.4).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/DeviceOrientationEvent

### DIFF
--- a/api/DeviceOrientationEvent.json
+++ b/api/DeviceOrientationEvent.json
@@ -70,10 +70,10 @@
               "version_added": "14"
             },
             "firefox": {
-              "version_added": null
+              "version_added": "17"
             },
             "firefox_android": {
-              "version_added": null
+              "version_added": "17"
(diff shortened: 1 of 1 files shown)


foolip (approved):
OK, so this is about the constructor, not the whole interface. I've confirmed with https://mdn-bcd-collector.appspot.com/tests/api/DeviceOrientationEvent/DeviceOrientationEvent in Firefox 16 and 17, the failure in Firefox 16 is "DeviceOrientationEvent is not a constructor", pretty clear :)

PR #10844, opened 2021-06-07
Add Firefox versions for ErrorEvent API

vinyldarkscratch:
This PR adds real values for Firefox and Firefox Android for the `ErrorEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.4).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/ErrorEvent

### DIFF
--- a/api/ErrorEvent.json
+++ b/api/ErrorEvent.json
@@ -15,10 +15,10 @@
             "version_added": "12"
           },
           "firefox": {
-            "version_added": "27"
+            "version_added": "30"
           },
           "firefox_android": {
-            "version_added": "27"
+            "version_added": "30"
           },
           "ie": {
             "version_added": true
@@ -63,10 +63,10 @@
               "version_added": "14"
             },
             "firefox": {
-              "version_added": "27"
+              "version_added": "30"
             },
             "firefox_android": {
-              "version_added": "27"
+              "version_added": "30"
             },
             "ie": {
               "version_added": null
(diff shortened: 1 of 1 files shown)


foolip:
I'm pretty sure the error event must be older than this, but the tests are just for the interface object / prototype.

Perhaps a custom test that triggers the event with a JavaScript parse error? (Need not be added to the collector, just for manual testing.)

vinyldarkscratch:
Interestingly enough, this seems to be the opposite case: this API was exposed before it was implemented into anything.  By using `window.addEventListener('error', function(e) {...});`, it appears that it wasn't until Firefox 30 that an error would create an `ErrorEvent` in contrast to just an `Event`.

foolip:
Oh events are a delight, it would take us forever to test and get everything right 😭

So what's the first version when an event using this interface can be fired? I say let's use that version in the data.

foolip:
@vinyldarkscratch what is https://github.com/mdn/browser-compat-data/pull/10844/commits/7091dcfe735c192ca1998c39b6ea2e2936676d91 based on?

foolip:
Hmm wait, it's https://github.com/mdn/browser-compat-data/pull/10844#issuecomment-860053472.

So we should follow https://github.com/mdn/browser-compat-data/pull/10379 and use the version when `ErrorEvent` was first used. That's different to whether the `ErrorEvent` interface object is exposed, but that doesn't seem to be the thing that changed here.

foolip (approved):

PR #10849, opened 2021-06-07
Add release notes for Safari 14.1 and Safari 15 beta.

lukewarlow:
#### Summary

I've added the missing release notes for Safari 14.1 (they've only been made available recently) and also added a beta entry for the just announced Safari 15 with relevant release notes.

https://developer.apple.com/documentation/safari-release-notes/safari-14_1-release-notes

https://developer.apple.com/documentation/safari-release-notes/safari-15-beta-release-notes

### DIFF
--- a/browsers/safari.json
+++ b/browsers/safari.json
@@ -176,9 +176,15 @@
         },
         "14.1": {
           "release_date": "2021-04-26",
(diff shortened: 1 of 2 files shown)


ddbeck (approved):
We'll probably have to sub in a non-beta URL when Safari 15 actually ships, but this looks good for now. Thank you, @lukewarlow!

PR #10851, opened 2021-06-07
Add Safari 15 compat data

gsouquet:
#### Summary

Following WWDC here are the updated compat data of features I could match in the repo

Here is a list of things to mark as done if we deem it ok to not include them as part of this pull request

#### Questions

- ~~Where to include predefined color spaces using the color() syntax?~~ should probably be tackled as part of #10658
- ~~What about `lab()`, `lch()`, `hwb()`? There are currently no references to those features in the dataset~~ should probably be tackled as part of #10658
- `Error.cause` is inexistant in MDN at the moment, should I create a new file for it even though the page does not exist yet?
- There are currently no differences between private data members and private class methods on MDN they are documented under the same dataset
- Shall I mark the `WebShare` enhancements as a note under the `Navigator.share()` entry?

### DIFF
--- a/api/WebGL2RenderingContext.json
+++ b/api/WebGL2RenderingContext.json
@@ -30,10 +30,10 @@
             "version_added": "43"
           },
           "safari": {
-            "version_added": false
+            "version_added": "15"
           },
           "safari_ios": {
-            "version_added": false
+            "version_added": "15"
           },
           "samsunginternet_android": {
             "version_added": "7.0"
--- a/css/properties/aspect-ratio.json
+++ b/css/properties/aspect-ratio.json
@@ -69,10 +69,10 @@
               "version_added": false
             },
             "safari": {
-              "version_added": false
+              "version_added": "15"
             },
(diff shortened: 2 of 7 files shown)


138166234:
please write by language pershian.

1381662341mohamadsalmanpor

در تاریخ سه‌شنبه ۸ ژوئن ۲۰۲۱،‏ ۲:۲۲ Germain ***@***.***> نوشت:

lukewarlow:
The colour functions such as `lab()` are added in this pr: https://github.com/mdn/browser-compat-data/pull/10658

lukewarlow:
It would be good to differentiate between WebShare level 1 and level 2 somehow. Maybe adding the separate properties of share data, or at least separate `files` out, as that's the difference between level 1 and 2? I'm not sure how those kind of spec enhancements are generally handled.

lukewarlow:
#10849 has been merged

gsouquet:
Superseded by #10898

Issue #9833, opened 2021-04-12
javascript.classes.private_class_fields - Rename and build out


meyerweb:
Currently, the only information about private class features (see https://github.com/tc39/proposal-private-methods#private-methods-and-fields) is a single row of data labeled `private_class_fields`.  I propose to replace it with `private_class_features` and give a row each to fields and methods, in both static and instance variants.  So the structure would be:

```
javascript
   classes
      private_class_features
         _compat
         instance_fields
         instance_methods
         static_fields
         static_methods
```

I would also handle converting any MDN documentation referencing `private_class_fields` to instead refer to `private_class_features`.  My current assumption is that the actual update process would be:

1. Add the above `private_class_features` structure to the BCD in a PR
1. Update MDN pages to point to it
1. Remove the `private_class_fields` structure from the BCD in a second PR

I’m opening this issue to invite discussion of the proposal, so any adjustments can be made ahead of submitting actual pull requests.

jculver050:
Close

hamishwillee:
Why? Is it because you expect private features to continue to be extended, and to deliver across many specs, an many different browser releases?

Note, #10854 is currently modifying this because the private features appear in FF90, so if we're going to, a good time to think about touching this.

meyerweb:
Because different aspects of private class features have been added to browsers at different times (e.g. Chrome, which added private fields in 74 and private methods in 84).  The distinction between static and instance seems to be important 
to JS documentation and its readers, so I preserved it here.

I don’t know if there will be feature extensions or not.

ddbeck:
Just saw this thanks to https://github.com/mdn/browser-compat-data/pull/10854. Since private class fields and methods were introduced separately in more than one browser, it definitely makes sense to have features for each. How to actually structure this is a bit tougher.

My inclination is to flatten everything:

```
javascript
   classes
      private_instance_fields
      private_instance_methods
      private_static_fields
      private_static_methods
```

or, with some more depth:

```
javascript
   classes
      private_fields*
        instance_fields
        static_fields
      private_methods*
        instance_methods
        static_methods
```

\* we'd have to do something tricky here, like `partial_implementation` for support for only one of the subfeatures being supported

The main thing I want to avoid here is the ambiguity created by the proposed `private_class_features`, which, given our schema, would only be able to mean "support for one or more of private class/instance fields/methods, but we won't tell you which one unless you actually read all of the subfeatures".

meyerweb:
[...]

PR #10854, opened 2021-06-08
Javascript private class fields supported from FF90

hamishwillee:
FF90 adds support for [private class fields](https://bugzilla.mozilla.org/show_bug.cgi?id=1708235) and [private class methods](https://bugzilla.mozilla.org/show_bug.cgi?id=1708236). 

Previously this was only supported on Nightly build and behind a preference. Because the preference was never available in release build (AFAK - or at least documented that way) I have removed the old preference info altogether from BCD as "not relevant".

Docs tracking for this is in https://github.com/mdn/content/issues/5386

As an aside, might be a good time to consider whether we want to do the work suggested in #9833

### DIFF
--- a/javascript/classes.json
+++ b/javascript/classes.json
@@ -489,36 +489,10 @@
               "version_added": "79"
             },
             "firefox": {
-              "version_added": false,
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Looks good. Thank you! 🎉

PR #10864, opened 2021-06-08
Add Firefox versions for PushSubscriptionOptions API

vinyldarkscratch:
This PR adds real values for Firefox and Firefox Android for the `PushSubscriptionOptions` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.4).

### DIFF
--- a/api/PushSubscriptionOptions.json
+++ b/api/PushSubscriptionOptions.json
@@ -15,10 +15,10 @@
             "version_added": "16"
           },
           "firefox": {
-            "version_added": true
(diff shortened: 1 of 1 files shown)


foolip (changes requested):
This doesn't match https://developer.mozilla.org/en-US/docs/Web/API/PushSubscription/options#browser_compatibility, this would fail a consistency check if we had one :)

foolip (approved):

PR #10867, opened 2021-06-08
Add Firefox versions for RTCPeerConnectionIceEvent API

vinyldarkscratch:
This PR adds real values for Firefox and Firefox Android for the `RTCPeerConnectionIceEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.4).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/RTCPeerConnectionIceEvent

### DIFF
--- a/api/RTCPeerConnectionIceEvent.json
+++ b/api/RTCPeerConnectionIceEvent.json
@@ -29,10 +29,10 @@
             "version_added": "15"
           },
           "firefox": {
-            "version_added": true
+            "version_added": "24"
           },
           "firefox_android": {
-            "version_added": true
+            "version_added": "24"
           },
           "ie": {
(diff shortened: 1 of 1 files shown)


foolip (changes requested):
https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection#browser_compatibility says the related APIs were introduced in Firefox 22. Which is right?

vinyldarkscratch:
Yeah, just confirmed that was the case: the `RTCPeerConnection` API was implemented in Firefox 22, but this event in 24.

foolip:
Alright, such things can happen :)

foolip (approved):

PR #10875, opened 2021-06-08
Add Firefox versions for TrackEvent API

vinyldarkscratch:
This PR adds real values for Firefox and Firefox Android for the `TrackEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.4).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/TrackEvent

### DIFF
--- a/api/TrackEvent.json
+++ b/api/TrackEvent.json
@@ -15,10 +15,10 @@
             "version_added": "12"
           },
           "firefox": {
-            "version_added": null
+            "version_added": "27"
           },
           "firefox_android": {
-            "version_added": null
+            "version_added": "27"
           },
           "ie": {
             "version_added": "10"
@@ -63,10 +63,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": null
+              "version_added": "27"
             },
             "firefox_android": {
-              "version_added": null
+              "version_added": "27"
             },
             "ie": {
               "version_added": "10"
--- a/api/TrackEvent.json
+++ b/api/TrackEvent.json
@@ -15,10 +15,10 @@
             "version_added": "12"
           },
           "firefox": {
-            "version_added": "27"
+            "version_added": "31"
           },
           "firefox_android": {
-            "version_added": "27"
+            "version_added": "31"
           },
           "ie": {
             "version_added": "10"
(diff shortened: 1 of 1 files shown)


foolip:
This should match https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/textTracks#browser_compatibility I think, or what was this event used for before that was available?

vinyldarkscratch:
Seems to be exposed early for no particular reason, I can't find anything that uses this event other than the linked property.  It _is_ exposed, however.

foolip:
So this was probably accidentally missing a runtime flag in the IDL and exposed on its own, which is a pretty common mishap.

It happened for `MediaMetadata` in Safari 14.1 most recently. Then we just pretended it wasn't exposed.

I say let's make the data match `textTracks` and add this to a collector ignore list if needed later.

saschanaz:
https://bugzilla.mozilla.org/show_bug.cgi?id=833385 says they implemented and exposed under the same flag 🤔

saschanaz:
https://bugzilla.mozilla.org/show_bug.cgi?id=887978 says texttrack is enabled on Firefox 28 while TrackEvent was added in 27. Not sure why BCD says 23 👀

foolip:
@saschanaz BCD is at least a little bit wrong about a large proportion of the data because it started out as just manually maintained tables in the wiki. https://mdn-bcd-collector.appspot.com/ disagrees with lots of the data, which we should eventually work through, but we're focusing on missing data now...

Prediction: randomly sampling BCD entries and trying to find something wrong would find something wrong with >20% of entries.

foolip:
I've confirmed http://mdn-bcd-collector.appspot.com/tests/api/HTMLMediaElement/textTracks went from false to true in Firefox 31, and `npm run traverse firefox all 31 | grep -i track` shows all the related APIs being enabled then, nothing in 27 or 28. So I trust this is correct now.

foolip (approved):

saschanaz:
Ah yes, the comment says it was backed out and re-enabled in 31. Sorry for the confusion.

foolip:
No worries, digging up bugs is always helpful!

Issue #10533, opened 2021-05-19
Create script to diff aligned browser releases


gsnedders:
It would be nice to get a summary of places where there's a data mismatch between two aligned browser releases (e.g., between Safari 13 and Safari for iOS 13).

As an example of differences:

```
gsnedders@gsnedders-milk browser-compat-data % npm run traverse safari all 13 > safari.txt
gsnedders@gsnedders-milk browser-compat-data % npm run traverse safari_ios all 13 > safari_ios.txt
gsnedders@gsnedders-milk browser-compat-data % comm -3 safari.txt safari_ios.txt
> node scripts/traverse "safari" "all" "13"

	> node scripts/traverse "safari_ios" "all" "13"
	
	api.HTMLMediaElement.disableRemotePlayback
api.MediaDevices.getDisplayMedia
	api.SourceBuffer
	api.SourceBuffer.audioTracks
	api.SourceBuffer.textTracks
	api.SourceBuffer.videoTracks
	api.TextTrack.sourceBuffer
css.at-rules.media.display-mode
	css.at-rules.media.prefers-color-scheme
	css.at-rules.media.prefers-color-scheme.no-preference
	css.properties.clip-path.path
css.properties.touch-action
css.properties.touch-action.manipulation
	html.elements.a.download
136
	142
```

One could imagine a more complex UI to compare `safari` and `safari_ios` outputting a table like:

```
| feature | safari | safari_ios |
| --- | --- | --- | --- | --- |
| api.HTMLMediaElement.disableRemotePlayback | 13.1 | 13 |
| api.MediaDevices.getDisplayMedia | 13 | false |
| api.SourceBuffer | 8 | 13* |
| api.SourceBuffer.abort | 8 | false |
| api.SourceBuffer.appendStream | null | false |
| api.SourceBuffer.audioTracks | 8 | 13* |
```

This would require figuring out aligned browser releases (though hopefully the engine version would suffice here? though we will run into #8854 in the WebKit case) to give the full output.

There are certainly cases when we expect to these to differ, but the majority of differences are likely data bugs.

@foolip also mentioned the potential of using such functionality to compare not only Chrome and Chrome for Android, but also Chrome for Android and Samsung Internet (cc @AdaRoseCannon).

foolip:
I think this is an excellent idea, and would use such a script myself a lot. Given powerful enough capabilities, one could:
- Check for things in Chrome 90 but not in Edge 90 or Opera 76 (uses Chromium 90)
[...]

PR #10876, opened 2021-06-08
Hack traverse script to print CSV instead

foolip:
For https://github.com/mdn/browser-compat-data/issues/10533


foolip:
With this, run `node scripts/traverse.js > bcd.csv` to produce CSV which can be imported into stuff.

PR #10887, opened 2021-06-09
updating Chrome support for @counter-style API

rachelandrew:
The `@counter-style` rule shipped in Chrome 91, this updates the API data.

https://www.chromestatus.com/feature/5692693659254784

### DIFF
--- a/api/CSSCounterStyleRule.json
+++ b/api/CSSCounterStyleRule.json
@@ -6,13 +6,13 @@
         "spec_url": "https://drafts.csswg.org/css-counter-styles/#the-csscounterstylerule-interface",
         "support": {
           "chrome": {
-            "version_added": false
+            "version_added": "91"
           },
           "chrome_android": {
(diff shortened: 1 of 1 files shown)


api/CSSCounterStyleRule.json
```diff
@@ -287,13 +287,13 @@
         "__compat": {
           "support": {
             "chrome": {
-              "version_added": false
+              "version_added": "91"
```
foolip, line 290: The following `webview_android` change is missing here. I can make that edit though.

foolip (approved):
The same edits (for `chrome`) are suggested by mdn-bcd-collector using data from Chrome 90 and 91, so I'm pretty confident this is right.

foolip:
I see https://github.com/mdn/browser-compat-data/blob/main/css/at-rules/counter-style.json was already updated, so this is good to go.

PR #10888, opened 2021-06-09
Adds Edge support for @counter-style

rachelandrew:
Edge supports `@counter-style` as of version 91. Already had Chrome data so this just brings Edge up to date.


ddbeck (approved):
Thank you! 🎉

PR #10889, opened 2021-06-09
Add scripts for tracking demixes (and other interesting feature diffs)

ddbeck:
This PR adds two new scripts to the repo: `enumerate-features` and `diff-features`. Ostensibly, these two scripts make it easier for me to write release notes by helping me find out what features have been added or removed in a given commit, but these may also demonstrate the basis for additional automation and reporting on BCD. But it's also a step toward https://github.com/mdn/browser-compat-data/issues/10533.

The first script here is `enumerate-features`, which is like a lower-level plumbing alternative to `traverse`. Using the newish `walk()` util, `enumerate-features` writes out a JSON file listing every single feature in BCD (or, if specified, an alternate path to the package—this is is going to be significant later on).

The second script, `diff-features`, builds on the first. It runs `enumerate-features` against two different commits, then compares the set of features in each enumeration. This quickly produces a list of added or removed features. Because of the way we've implemented `walk()`, it even works with historic revisions, before we implemented `enumerate-features` itself.

The `diff-features` script could be a model for doing the tasks discussed in https://github.com/mdn/browser-compat-data/issues/10533. Instead of diffing between commits, we could have a script that runs `enumerate-features` for one set of criteria (e.g., "Supported in Safari 13") and again for another (e.g. "Supported in Safari for iOS 13"), then reports the symmetric difference.

A couple additional notes:

- Because it writes out to a JSON file, `enumerate-features` has a bunch of potential uses for workflow automation. For example, we could automatically count changes in the overall number of features for release notes…
- …or [save hashes of compat objects to detect when features have been modified](https://github.com/ddbeck/browser-compat-data/compare/inventory-paths...ddbeck:more-diffing). That seems to require an additional dependency, so I didn't include it in this PR.

### DIFF
--- a/scripts/diff-features.js
+++ b/scripts/diff-features.js
@@ -46,9 +46,7 @@ function enumerateFeatures(ref = 'HEAD') {
 
     return JSON.parse(fs.readFileSync('.features.json', { encoding: 'utf-8' }));
   } finally {
-    if (ref !== 'HEAD') {
-      execSync(`git worktree remove ${worktree}`);
-    }
+    execSync(`git worktree remove ${worktree}`);
   }
 }
 
--- a/scripts/diff-features.js
+++ b/scripts/diff-features.js
@@ -1,0 +1,95 @@
+const { execSync } = require('child_process');
+const fs = require('fs');
+
+const yargs = require('yargs');
+
+function main({ ref1, ref2, format }) {
+  let refA, refB;
+
+  if (ref1 === undefined && ref2 === undefined) {
+    // No refs: compare HEAD to parent commit
+    refA = 'HEAD^';
+    refB = 'HEAD';
+  } else if (ref2 === undefined) {
+    // One ref: compare ref to parent of ref
+    refB = `${ref1}`;
+    refA = `${ref1}^`;
+  } else {
+    // Two refs: compare ref2 to ref1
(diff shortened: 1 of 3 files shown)


scripts/diff-features.js
```diff
@@ -0,0 +1,97 @@
+const { execSync } = require('child_process');
+const fs = require('fs');
+
+const yargs = require('yargs');
+
+function main({ ref1, ref2, format }) {
```
ddbeck, line 6: It occurs to me now that we could decompose this further, to help other use cases better by separating the checkout-and-enumeration stuff from the feature set diffing stuff. Perhaps this script ought to take two JSON files arguments, instead of refs, then you could do something like:

```shell
$ node ./scripts/diff-features.js <(command-that-enumerates-a-side) <(command-that-enumerates-b-side)
```

(The checkout stuff could move to `enumerate-features`, as an option.)

ddbeck:
@Elchi3 Would you be interested in reviewing this at some point? It's not a critical thing, but it's a convenient tool for me for writing release notes. Not having to have this checked out separately from the rest of the repo would be nice.

Elchi3 (approved):
Looks good to me, Daniel! 
`enumerate-features` seems really useful, I like how our "private" API grows. I think there is no risk in just landing this.

PR #10898, opened 2021-06-09
Safari 15 beta updates

apple-web-evangelist:
#### Summary

Updates for Safari 15 beta from the Safari/WebKit teams at Apple.

This information has been compiled from Safari/WebKit engineers at Apple, combined with the results of the BCD collector where suspicious data was manually reviewed and corrected when necessary.

For more details see: [Safari 15 Beta Release Notes](https://developer.apple.com/documentation/safari-release-notes/safari-15-beta-release-notes)


tomayac:
@apple-web-evangelist One more thing to flip from `false` to `"15"` is Ecmascript module support for workers, here's the deeplink: https://github.com/mdn/browser-compat-data/blob/14781adc20f2c045f2a0886570f8fe00455d4771/api/Worker.json#L177

foolip:
I can take a look at the API data here at some point. I've filed https://github.com/mdn/yari/issues/4005 about a problem I anticipate. (Didn't know there was already so much Firefox beta data, or I would have suggested this earlier.)

html/elements/meta.json
```diff
@@ -675,10 +675,12 @@
                   "version_added": false
                 },
                 "safari": {
-                  "version_added": false
+                  "version_added": "15",
+                  "notes": "Safari 15 beta supports a media attribute that recognizes (prefers-color-scheme: light) or (prefers-color-scheme: dark)."
```
ddbeck, line 679: A few things on these notes, in order of seriousness:

1. I don't understand this note. The way this is phrased makes it seem quite indirect. I don't understand how it relates to the feature. A media attribute? Which one?
2. Since _only_ version 15 of Safari is supported, there's no need to mention the status (which is also captured in the release data) nor the version number. This helps prevent notes from going stale.
3. You can use a subset of HTML in notes, including `<code>`.

```suggestion
                  "notes": "Safari supports a media attribute that recognizes <code>prefers-color-scheme: light</code> or <code>prefers-color-scheme: dark</code>."
```

html/elements/meta.json
```diff
@@ -675,10 +675,12 @@
...
                 },
                 "safari_ios": {
-                  "version_added": false
+                  "version_added": "15",
+                  "notes": "Safari 15 beta supports a media attribute that recognizes (prefers-color-scheme: light) or (prefers-color-scheme: dark) on iPadOS 15."
```
ddbeck, line 683: And a bonus one:

4. Where possible, front-load exceptions, so readers can tell if it's relevant to them without reading to the very end.

```suggestion
                  "notes": "On iPadOS, Safari supports a media attribute that recognizes <code>prefers-color-scheme: light</code> or <code>prefers-color-scheme: dark</code>."
```

javascript/builtins/Error.json
```diff
@@ -105,6 +105,56 @@
             }
           }
         },
+        "cause": {
```
ddbeck, line 108: What is this feature? There's no description, URL, or spec URL. I don't know how to review this without some context.

api/WebGLVertexArrayObjectOES.json
```diff
@@ -30,7 +30,7 @@
             "version_added": "14"
           },
           "safari": {
-            "version_added": false
+            "version_added": "15"
```
ddbeck, line 33: This is the only change I've spotted so far without an iOS pairing. Is this intentional?

ddbeck (changes requested):
Thanks for this PR, @apple-web-evangelist. I spotted a few issues and logged a few questions in line comments.

(I was looking for structural issues mostly, more or less trusting any `false` → `"15"` change. I didn't think through the API changes too closely, so there might be things implied by these changes I haven't noticed.)

css/properties/scroll-margin-block-end.json
```diff
@@ -31,10 +31,10 @@
               "version_added": "48"
             },
             "safari": {
-              "version_added": "14.1"
+              "version_added": "15"
```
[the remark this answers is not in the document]
gsnedders: Suspect the data came from the results collector, and `CSS.supports` and/or `CSSStyleDeclaration` incorrectly including it (due to a variety of bugs in WebKit where properties behind flags sometimes show up).

html/elements/meta.json
lukewarlow, line 683: If this is referring to the new "theme-color" meta tag support, it isn't iPad OS only, it's also support on iOS.

Source: https://developer.apple.com/documentation/safari-release-notes/safari-15-beta-release-notes#:~:text=Added%20support%20for%20the%20theme-color%20meta%20tag%20to%20change%20the%20tab%20bar%20background%20and%20over-scroll%20area%20in%20macOS%20and%20iPadOS%2C%20and%20the%20status%20bar%20in%20iOS

css/properties/scroll-margin-block-end.json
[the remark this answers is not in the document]
apple-web-evangelist: I verified that the originally submitted data for Safari 14.1 and Safari on iOS 14.5 is correct. I will revert these.

css/types/color.json
```diff
@@ -151,6 +151,58 @@
             }
           }
         },
+        "color": {
+          "__compat": {
+            "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/color_value/color()",
+            "spec_url": "https://drafts.csswg.org/css-color/#color-function",
+            "description": "<code>color()</code>",
```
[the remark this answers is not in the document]
apple-web-evangelist: It was easy to match the label style for Lab, LCH, and HWB as those could easily be identified in the spec. A similar title label from the spec was not abundantly clear for `color()`. In my follow-up, I plan to update to use: `"Profiled color values <code>color()</code>"` for the description.

css/types/color.json
```diff
@@ -151,6 +151,58 @@
...
+              },
+              "opera": {
+                "version_added": false
+              },
+              "opera_android": {
+                "version_added": false
+              },
+              "safari": {
+                "version_added": "15",
+                "notes": "Supports <code>srgb</code>, <code>display-p3</code>, <code>a98-rgb</code>, <code>prophoto-rgb</code>, <code>rec2020</code>, and <code>xyz</code> predefined color spaces."
```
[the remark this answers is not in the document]
apple-web-evangelist: I will drop these.

html/elements/meta.json
apple-web-evangelist: Perhaps the better alternative proposal is to add a new entry for `media` showing Safari 15 support with a status of `experimental` and `standard_track`. This note could be attached there and rewritten to describe support for these specific cases.

api/WebGLVertexArrayObjectOES.json
apple-web-evangelist: This is not intentional and will be reverted.

javascript/builtins/Error.json
gsnedders: https://github.com/tc39/proposal-error-cause

html/elements/meta.json
apple-web-evangelist: I verified that this is indeed correct.

Teimtem:
Suttirat Ampha

ในวันที่ ศ. 11 มิ.ย. 2021 23:08 Jon Davis ***@***.***>
เขียนว่า:

html/elements/meta.json
apple-web-evangelist: It's probably best to move `media` attribute support to a new entry.

html/elements/meta.json
apple-web-evangelist: Ditto above a new entry here.

html/elements/meta.json
apple-web-evangelist: We're going to hold off updating `media` attribute in a follow-up PR.

css/types/color.json
```diff
@@ -151,6 +151,56 @@
...
+              },
+              "opera_android": {
+                "version_added": false
+              },
+              "safari": {
+                "version_added": "15"
+              },
+              "safari_ios": {
+                "version_added": "15"
+              },
```
Schweinepriester, line 189: Hasn't

> Specifying profiled colors: the color() function

been supported since Safari 10.1?

Per
- https://github.com/Fyrd/caniuse/issues/3134
- https://webkit.org/blog/7477/new-web-features-in-safari-10-1/#:~:text=css%20wide-gamut%20colors

Or did something change or the like…?

apple-web-evangelist: Yes, support did change. While `color()` function support was initially added in Safari 10.1, the only supported color profile at the time was `display-p3`. Now all of the predefined color profiles are supported.

This can be updated to reflect that.

Schweinepriester: I see!

For <https://caniuse.com/css-color-function> my first thought is to downgrade Safari <= 14 to partial with a note like "only supports `display-p3`" and start full support with 15.

But I'll observe how it will be handled for MDN here :)

css/types/color.json
```diff
@@ -151,6 +151,56 @@
             }
           }
         },
+        "color": {
+          "__compat": {
+            "description": "Profiled color values <code>color()</code>",
```
ddbeck, line 156:
```suggestion
            "description": "Profiled color values (<code>color()</code>)",
```

css/types/color.json
ddbeck: > While `color()` function support was initially added in Safari 10.1, the only supported color profile at the time was `display-p3`.

If this was the case, then you'll need to do two support statements. One from 15 onwards, without any notes, and another from 10.1 to 15 with a note saying that only `display-p3` is supported.

javascript/builtins/Error.json
```diff
@@ -105,6 +105,58 @@
...
+          "__compat": {
+            "description": "<code>Error.cause</code> represents additional contextual information about an error for the purpose of chaining errors.",
```
ddbeck, line 110: We can omit this description, so it'll show up like other properties of `Error`.

I'd also expect a corresponding subfeature of `Error()`, showing whether `options.cause` is supported.

```suggestion
```

ddbeck (changes requested):
A couple more, mostly small suggestions. Thank you!

ddbeck:
@apple-web-evangelist looking this over again, I think there's [one unresolved comment](https://github.com/mdn/browser-compat-data/pull/10898#discussion_r649578173) and the conflict in `api/Worker.json` to resolve. Getting quite close to merging this!

foolip (approved):
All changes in api/ look good to me, just one question about the very last one.

api/Worker.json
```diff
@@ -171,10 +171,10 @@
                 "version_added": "57"
               },
               "safari": {
-                "version_added": "14.1"
+                "version_added": "15"
```
foolip, line 174: Seems like I didn't submit my comment here. Is the original 14.1 wrong here, where did that data come from?

apple-web-evangelist: Yes. It is wrong. I don't know where it came from.

foolip: Thanks for confirming, no need to track it down if you're confident it's wrong.

apple-web-evangelist: I verified it was wrong on Safari 14.1 using wpt.live tests. The change appears to have come from this: https://github.com/mdn/browser-compat-data/pull/10749 That PR cites a blogger discussing it, but the blogger discussed it in the context of Safari Technology Preview, not the Safari 14.1 release.

foolip: Oh, I merged that, oops! @tomayac FYI that this turned out to be wrong.

tomayac: Oops, sorry. I was under the impression the TP version numbers could be used for this. Honest mistake.

foolip: Ah, and I didn't realize it was intended to mean STP :)

Soon we will have "preview" as an allowed string for things that are only in STP (and Chrome Canary or Firefox Nightly).

javascript/builtins/Error.json
```diff
@@ -105,6 +105,57 @@
             }
           }
         },
+        "cause": {
```
ddbeck, line 108: @apple-web-evangelist sorry, I missed this previously. This nesting is off by one level. It's at `javascript.builtins.Error.options_cause_parameter` instead of `javascript.builtins.Error.Error.options_cause_parameter`.

(that is, it's a sibling of the constructor data, rather than a descendant of it)

ddbeck (approved):
Looks good. If there's anything else to do, let's get to it in a follow up PR. Thank you! 🎉

danburzo:
I stumbled upon the MDN page for `color()` and saw the note about new support in Safari 15. It seems, however, that `srgb` was supported in addition to `display-p3` since Safari 10.1. 

Confirmed with this test case in BrowserStack:

```js
<!DOCTYPE html>
<html lang='en'>
<head>
	<meta charset="utf-8">
	<title></title>
</head>
<body style='background: color(srgb 0.9 0.25 0.3)'>
</body>
</html>
```

ddbeck:
@danburzo 

> Shall I submit a PR?

Please do (or open an issue)! It's easy for follow ups on merged PRs to get forgotten (like I almost did here).

PR #10906, opened 2021-06-10
Add Safari versions for AudioNode API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `AudioNode` API by mirroring the data.


foolip (approved):
https://caniuse.com/audio-api agrees

PR #10908, opened 2021-06-10
Update Safari versions for DOMParser API

vinyldarkscratch:
This PR updates the Safari versions for the `DOMParser` API based upon commit history (see below) and manual testing.

### DIFF
--- a/api/DOMParser.json
+++ b/api/DOMParser.json
@@ -33,7 +33,7 @@
             "version_added": "3.2"
           },
           "safari_ios": {
-            "version_added": "1.3"
+            "version_added": "2"
           },
           "samsunginternet_android": {
             "version_added": "1.0"
@@ -82,7 +82,7 @@
               "version_added": "3.2"
             },
             "safari_ios": {
-              "version_added": "1.3"
+              "version_added": "2"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
@@ -131,7 +131,7 @@
               "version_added": "3.2"
             },
             "safari_ios": {
-              "version_added": "1.3"
+              "version_added": "2"
             },
             "samsunginternet_android": {
               "version_added": true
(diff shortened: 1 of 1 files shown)


foolip:
I don't have high trust in the accuracy of mirroring Safari to iOS, so I found where this was implemented instead:
https://trac.webkit.org/changeset/9339/webkit

The `Version.xcconfig` file didn't exist then, but I could bisect it to being introduced in Safari 1.3:
https://trac.webkit.org/browser/webkit/branches/old/Safari-1-2-branch/WebCore/khtml/ecma/kjs_window.cpp (no `DOMParser`)
https://trac.webkit.org/browser/webkit/branches/old/Safari-1-3-branch/WebCore/khtml/ecma/kjs_window.cpp (yes `DOMParser`)

Given that, it should also be in iOS 1.

foolip:
Figuring out the support for HTML, SVG and XML separately will take some extra work, but let's start there.

api/DOMParser.json
```diff
@@ -223,10 +223,10 @@
                 "version_added": true
               },
               "safari": {
-                "version_added": "3.2"
+                "version_added": "3"
```
foolip, line 226: This is the SVG support, did you track down a commit or where is the version from?

api/DOMParser.json
```diff
@@ -271,10 +271,10 @@
                 "version_added": true
               },
               "safari": {
-                "version_added": "3.2"
+                "version_added": "1.3"
```
foolip, line 274: This is the XML support, and that was from the start, correct!

api/DOMParser.json
vinyldarkscratch: This one's admittedly a guess, I figure that `DOMParser` support would have been around for SVGs when SVG support was first introduced in Safari.  I've just checked the code, however, and it seems to suggest what I assumed.

https://trac.webkit.org/browser/webkit/tags/old/Safari-5522.11/WebCore/dom/DOMImplementation.cpp#L227
https://trac.webkit.org/browser/webkit/tags/old/Safari-5522.11/WebCore/xml/DOMParser.cpp#L33

foolip: That looks like the right code for this, thanks for verifying!

foolip (approved):

PR #10909, opened 2021-06-10
Update Safari versions for Geolocation APIs

vinyldarkscratch:
This PR updates and corrects the real values for Safari (Desktop and iOS/iPadOS) for the `GeolocationCoordinates` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.4).

Tests Used:
https://mdn-bcd-collector.appspot.com/tests/api/Geolocation
https://mdn-bcd-collector.appspot.com/tests/api/GeolocationCoordinates
https://mdn-bcd-collector.appspot.com/tests/api/GeolocationPosition

### DIFF
--- a/api/GeolocationCoordinates.json
+++ b/api/GeolocationCoordinates.json
@@ -78,7 +78,8 @@
             },
             {
               "alternative_name": "Coordinates",
-              "version_added": "5"
+              "version_added": "5",
+              "version_removed": "13.1"
             }
           ],
           "safari_ios": [
@@ -87,7 +88,8 @@
             },
             {
               "alternative_name": "Coordinates",
-              "version_added": "4.2"
+              "version_added": "4.2",
+              "version_removed": "13.4"
             }
           ],
           "samsunginternet_android": {
--- a/api/GeolocationCoordinates.json
+++ b/api/GeolocationCoordinates.json
@@ -88,7 +88,7 @@
             },
             {
               "alternative_name": "Coordinates",
-              "version_added": "4.2",
+              "version_added": "≤3",
               "version_removed": "13.4"
             }
           ],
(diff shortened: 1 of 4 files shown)


api/GeolocationCoordinates.json
```diff
@@ -72,14 +72,24 @@
...
-            "alternative_name": "Coordinates",
-            "version_added": "4.2"
-          },
+          "safari": [
+            {
+              "version_added": "13.1"
+            },
+            {
+              "alternative_name": "Coordinates",
+              "version_added": "5"
```
foolip, line 81: This matches api.GeolocationPosition.coords, good!

But should there also be a `version_removed`?

api/GeolocationCoordinates.json
```diff
@@ -146,7 +156,7 @@
               "version_added": "5"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "5"
```
foolip, line 159: The above edit says 4.2, and api.Navigator.geolocation does too. But https://caniuse.com/geolocation suggests this was shipped on iOS before macOS, which actually makes a lot of sense for this API. Perhaps leave the iOS data alone, and do a separate PR to make it consistent?

foolip (changes requested):

api/GeolocationCoordinates.json
vinyldarkscratch: Added based upon testing the alt. name's existence in Safari 13.1!

api/GeolocationCoordinates.json
foolip: For the avoidance of doubt: I'm waiting for this before reviewing again.

vinyldarkscratch: It does seem like support on iOS has been around a lot longer per testing in BrowserStack on iOS 3.  I'll set this to `≤3` for iOS accordingly, and update the other Geolocation APIs as well.

foolip: Wohoo!

foolip (approved):

PR #10910, opened 2021-06-10
Update Safari data for api.HTMLCanvasElement.getContext.webgl_context

vinyldarkscratch:
This PR updates the data for the `getContext.webgl_context` member of the `HTMLCanvasElement` API.  By running a line of test code (`document.createElement('canvas').getContext('webgl')`) in a few different Safari versions, I found out that Safari included support for the regular `webgl` value to `getContext` along with `experimental-webgl` in Safari 8.  Additionally, this updates the Safari iOS data to reflect the same version stated in `WebGLRenderingContext`.

### DIFF
--- a/api/HTMLCanvasElement.json
+++ b/api/HTMLCanvasElement.json
@@ -428,9 +428,15 @@
                   "alternative_name": "experimental-webgl"
                 }
               ],
-              "safari_ios": {
-                "version_added": "8"
-              },
+              "safari_ios": [
+                {
+                  "version_added": "8"
+                },
+                {
+                  "version_added": "8",
+                  "alternative_name": "experimental-webgl"
+                }
+              ],
               "samsunginternet_android": [
                 {
                   "version_added": "2.0"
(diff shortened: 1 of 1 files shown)


foolip (changes requested):
Per https://caniuse.com/webgl WebGL didn't ship at the same time on desktop and mobile, and I'd guess that's right, although I wouldn't blindly trust the specific versions claimed there.

Can you check what the various WebGL data we have currently says, and if it all needs updating perhaps?

vinyldarkscratch:
This should now be ready to go!  I did some further digging and found that Safari introduced the standard `webgl` value in Safari 8 (though has kept `experimental-webgl` as well).

foolip:
Did this involve testing iOS too? When WebGL first shipped in iOS is my main point of uncertainty.

vinyldarkscratch:
I just performed the testing in Safari iOS 7 and 8, and can confirm it's accurate!

foolip:
Excellent! 🙏

foolip (approved):

foolip:
I wonder if there's more WebGL data for iOS that should be bumped to 8?

PR #10911, opened 2021-06-10
Add Safari iOS versions for api.IDBCursor.worker_support

vinyldarkscratch:
This PR adds real values for Safari iOS/iPadOS for the `worker_support` member of the `IDBCursor` API by mirroring the data.

### DIFF
--- a/api/IDBCursor.json
+++ b/api/IDBCursor.json
@@ -676,7 +676,7 @@
               "version_added": "7"
             },
             "safari_ios": {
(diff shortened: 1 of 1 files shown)


foolip (approved):
Matches the parent feature and https://developer.mozilla.org/en-US/docs/Web/API/Worker#browser_compatibility says workers were available since iOS 4, so I don't see any reason to dig deeper.

PR #10912, opened 2021-06-10
Correct Safari versions for MediaStreamAudioSourceOptions API

vinyldarkscratch:
This PR corrects values for Safari for the `MediaStreamAudioSourceOptions` API by mirroring the data from `MediaStreamAudioSourceNode`.

### DIFF
--- a/api/MediaStreamAudioSourceOptions.json
+++ b/api/MediaStreamAudioSourceOptions.json
@@ -30,10 +30,10 @@
             "version_added": "14"
           },
           "safari": {
-            "version_added": "11"
+            "version_added": "14.1"
           },
           "safari_ios": {
-            "version_added": "11"
+            "version_added": "14.5"
           },
           "samsunginternet_android": {
             "version_added": "6.0"
--- a/api/MediaStreamAudioSourceOptions.json
+++ b/api/MediaStreamAudioSourceOptions.json
@@ -78,10 +78,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": "14.1"
             },
             "safari_ios": {
-              "version_added": null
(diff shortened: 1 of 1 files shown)


foolip (changes requested):
The existing data for Safari doesn't make sense together with https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamAudioSourceNode#browser_compatibility, something must be wrong.

vinyldarkscratch:
Looks like the "6" came from the original wiki tables, which I corrected for `MediaStreamAudioSourceNode` in #7935.  Seems like this was left untouched.  Fixing!

foolip:
Taking another look. This dictionary is only used in the `MediaStreamAudioSourceNode` constructor, so actually it should match https://developer.mozilla.org/en-US/docs/Web/API/MediaStreamAudioSourceNode/MediaStreamAudioSourceNode#browser_compatibility.

vinyldarkscratch:
Looking through the [commit history](https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Modules/webaudio/MediaStreamAudioSourceNode.idl?rev=150924#L1), I also see that the `mediaStream` argument was around since this API was first implemented, so I'd say we can fill in the data for it as well!

foolip (approved):

foolip:
Merged, but just a note that this is the evidence that the `mediaStream` dictionary member was supported from the beginning:
https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Modules/webaudio/MediaStreamAudioSourceOptions.idl?rev=265342

What you found was the `MediaStreamAudioSourceNode` interface, and it's totally possible for the interface to have an attribute without a corresponding member in the constructor init dict.

PR #10913, opened 2021-06-10
Add Safari versions for Navigator API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `Navigator` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.4).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Navigator

### DIFF
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -774,13 +774,11 @@
             },
             "safari": {
               "version_added": "5.1",
-              "version_removed": "7",
-              "notes": "Safari 7.1.3+ uses <code>window.doNotTrack</code> rather than <code>navigator.doNotTrack</code>."
+              "version_removed": "7"
             },
             "safari_ios": {
               "version_added": "5.1",
-              "version_removed": "7",
-              "notes": "Safari 7.1.3+ uses <code>window.doNotTrack</code> rather than <code>navigator.doNotTrack</code>."
+              "version_removed": "7"
             },
             "samsunginternet_android": {
               "version_added": true
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -314,10 +314,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": false
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": false
             },
             "samsunginternet_android": {
               "version_added": false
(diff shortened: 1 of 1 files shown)


api/Navigator.json
```diff
@@ -364,10 +364,10 @@
               "version_added": false
             },
             "safari": {
-              "version_added": false
+              "version_added": "14"
```
foolip, line 367: Tried calling `navigator.canShare()` in Safari 14.1, it returns false, so I can confirm it's shipped. Interesting!

api/Navigator.json
```diff
@@ -778,7 +778,9 @@
               "notes": "Safari 7.1.3+ uses <code>window.doNotTrack</code> rather than <code>navigator.doNotTrack</code>."
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "5.1",
+              "version_removed": "7",
+              "notes": "Safari 7.1.3+ uses <code>window.doNotTrack</code> rather than <code>navigator.doNotTrack</code>."
```
foolip, line 783: I've tested `window.doNotTrack` on Safari 14.1 on macOS and it's not there. This and the previous note needs revising.

api/Navigator.json
```diff
@@ -2011,10 +2013,10 @@
               "version_removed": "14"
             },
             "safari": {
-              "version_added": null
+              "version_added": false
```
foolip, line 2016: No `registerContentHandler`, makes sense.

api/Navigator.json
```diff
@@ -314,10 +314,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": false
```
foolip, line 317: No `buildID`, makes sense.

foolip (changes requested):

api/Navigator.json
foolip: I've tested with https://software.hixie.ch/utilities/js/live-dom-viewer/?saved=9402 and `window.doNotTrack` was gone by Safari 9.1, so I think we should just delete this feature-in-a-note as irrelevant.

foolip (approved):

PR #10914, opened 2021-06-10
Add Safari iOS versions for VR* API

vinyldarkscratch:
This PR adds real values for Safari iOS/iPadOS for the `VR*` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.4).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/VR*

### DIFF
--- a/api/VRPose.json
+++ b/api/VRPose.json
@@ -46,8 +46,7 @@
             "version_added": null
           },
           "safari": {
-            "version_added": "11.1",
-            "version_removed": "14"
+            "version_added": false
           },
           "safari_ios": {
             "version_added": false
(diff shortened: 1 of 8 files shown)


api/VRPose.json
```diff
@@ -46,6 +46,10 @@
             "version_added": null
           },
           "safari": {
+            "version_added": "11.1",
```
foolip, line 49: Can you separate out this change and just add safari_ios in this PR? I'm suspicious of this and would want to look closer at what else was shipped and unshipped along with it.

foolip (changes requested):

foolip (approved):

PR #10915, opened 2021-06-10
Add Safari versions for CanMakePaymentEvent API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `CanMakePaymentEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.4).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/CanMakePaymentEvent

### DIFF
--- a/api/CanMakePaymentEvent.json
+++ b/api/CanMakePaymentEvent.json
@@ -50,10 +50,10 @@
             "version_added": false
           },
           "safari": {
-            "version_added": null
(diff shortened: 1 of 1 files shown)


foolip (approved):
No relevant "CanMakePaymentEvent" on WebKit source so this seems right.

PR #10916, opened 2021-06-10
Add Safari versions for CompositionEvent API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `CompositionEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.4).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/CompositionEvent

### DIFF
--- a/api/CompositionEvent.json
+++ b/api/CompositionEvent.json
@@ -30,10 +30,10 @@
             "version_added": true
           },
           "safari": {
-            "version_added": true
+            "version_added": "6"
           },
           "safari_ios": {
-            "version_added": true
+            "version_added": "6"
           },
           "samsunginternet_android": {
             "version_added": true
@@ -79,10 +79,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": true
+              "version_added": "6.1"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "7"
             },
             "samsunginternet_android": {
               "version_added": true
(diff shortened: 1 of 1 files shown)


foolip:
Implemented here:
https://trac.webkit.org/changeset/50968/webkit
https://trac.webkit.org/browser/webkit/trunk/WebCore/dom/CompositionEvent.idl?rev=50968
https://trac.webkit.org/browser/webkit/trunk/WebCore/Configurations/Version.xcconfig?rev=50968

I'm not going to trust the WebKit trunk version, however, I think this required changes in JSC which didn't happen until later.

The constructor came later:
https://trac.webkit.org/changeset/141028/webkit
https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Configurations/Version.xcconfig?rev=141028

So anyway, I think there's a problem still and I'll comment inline.

api/CompositionEvent.json
```diff
@@ -128,10 +128,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "6.1"
```
foolip, line 131: The `data` attribute was there from the start, so something seems to be wrong here. I would suspect that the interface object wasn't exposed initially, but then when was the event first supported?

If that could be answered, data for the 3 composition* events could also be filled out.

vinyldarkscratch:
I just checked and found the event was around in the source code for [Safari 5](https://trac.webkit.org/browser/webkit/tags/Safari-533.16/WebCore/dom/CompositionEvent.idl?rev=278926).  I've updated the version numbers accordingly.

foolip (approved):

PR #10917, opened 2021-06-10
Add Safari versions for DedicatedWorkerGlobalScope API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `DedicatedWorkerGlobalScope` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.4).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/DedicatedWorkerGlobalScope

### DIFF
--- a/api/DedicatedWorkerGlobalScope.json
+++ b/api/DedicatedWorkerGlobalScope.json
@@ -178,10 +178,10 @@
               "version_added": "47"
             },
             "safari": {
-              "version_added": null
+              "version_added": false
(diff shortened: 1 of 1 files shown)


foolip (approved):
This is correct, but while confirming it I found something unfortunate, as usual:
https://bugs.webkit.org/show_bug.cgi?id=171216#c2

PR #10918, opened 2021-06-10
add URLs to RemotePlayback

rachelandrew:
I'm documenting RemotePlayback today, this PR adds the MDN and spec URLs to the BCD.


foolip (approved):
Thanks @rachelandrew!

PR #10919, opened 2021-06-10
Adds spec and MDN URLs for ElementInternals

rachelandrew:
Another API I will be working on, adds the spec and MDN URLs.


Elchi3 (approved):
🎉

PR #10920, opened 2021-06-10
Add Safari versions for IDB* API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `IDB*` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.4).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/IDB*


api/IDBCursorWithValue.json
```diff
@@ -52,10 +52,10 @@
             "version_added": "14"
           },
           "safari": {
-            "version_added": "7"
+            "version_added": "6.1"
```
foolip, line 55: I can confirm that this and other bits of IndexedDB were exposed in Safari 6.1 using BrowserStack, but I think we should change everything listed by `npm run traverse safari api 7 | grep -iE 'idb|indexeddb'` in one go if we're going to do it, and that would require care to not change things that actually didn't ship in the initial release. https://caniuse.com/indexeddb says something pretty different here, with notes about very buggy behavior.

Can you separate all changes that decrease the version number, bump any added "6.1" to "7" and  file an issue for this particular mess?

foolip (changes requested):

api/IDBCursorWithValue.json
vinyldarkscratch: By running the traverse script, it appears the only IDB entries still marked as "7" is the `IDBCursor` (which I might have accidentally undone the changes for; running the collector in Safari 6.1 confirms support) and various events (which the collector doesn't update anyways).  But for Safari iOS, support was not added until iOS 8, as I have just confirmed, so I'll be reverting the Safari updates.

foolip:
@vinyldarkscratch there are conflicts now, I'll hold off with review until pinged.

vinyldarkscratch:
@foolip Ping!

api/IDBCursor.json
```diff
@@ -673,10 +673,10 @@
               "version_added": "14"
             },
             "safari": {
-              "version_added": "7"
+              "version_added": "10.1"
```
foolip, line 676: Is it really the case that worker support for `IDBCursorWithValue` was in Safari 10, but `IDBCursor` came in 10.1? That's very strange given that `IDBCursorWithValue` inherits from `IDBCursor`...

api/IDBEnvironment.json
```diff
@@ -44,10 +44,10 @@
             "version_added": "14"
           },
           "safari": {
-            "version_added": "7"
+            "version_added": false
```
foolip, line 47: Indeed, no trace of "IDBEnvironment" in WebKit source.

api/IDBObjectStore.json
```diff
@@ -966,10 +966,10 @@
               "version_added": "14"
             },
             "safari": {
-              "version_added": "7"
+              "version_added": "10.1"
```
foolip, line 969: This is `openKeyCursor` and is later than the rest. Is this based on mdn-bcd-collector results?

api/IDBTransaction.json
```diff
@@ -661,10 +661,10 @@
               "version_added": "35"
             },
             "safari": {
-              "version_added": true
+              "version_added": "10.1"
```
foolip, line 664: This is `objectStoreNames`, confirmed manually?

api/IDBVersionChangeEvent.json
```diff
@@ -44,10 +44,10 @@
             "version_added": true
           },
           "safari": {
-            "version_added": "7"
+            "version_added": "6.1"
           },
           "safari_ios": {
-            "version_added": true
+            "version_added": "7"
```
foolip, line 50: Is this "7" is mistake? Are any other IDB* APIs left in iOS < 8 after this change?

foolip (commented):
With all the batch changes I asked more annoying questions about anything that didn't fit the pattern :)

api/IDBCursor.json
vinyldarkscratch: This is just updating its availability in workers!  The API is still marked as supported since Safari 6.1.

api/IDBObjectStore.json
vinyldarkscratch: Yeah, this is pretty much all from the collector results, aside from ensuring iOS stays marked as "8" and not earlier.

api/IDBVersionChangeEvent.json
vinyldarkscratch: It is -- correcting now!

foolip (approved):
I've confirmed with some `npm run traverse safari_ios api $version` calls that there's no iOS data <8 now.

foolip:
Hmm, now I'm having second thoughts in light of https://github.com/mdn/browser-compat-data/issues/9423, where we're contemplating just removing 6.1 and 6.2 as releases. If we do, we'd change these 6.1 back to 7 again.

So... uh... merge away, regret it later.

api/IDBCursor.json
foolip: Yes, but all the other worker support entries say Safari 10. It seems implausible that just this would be exposed later. Maybe a mishap with the interface object exposure, but the functionality itself, wasn't it there in 10?

vinyldarkscratch: Ah, gotcha -- yeah, this is actually in Safari 10 so I feel this might have been a typo.  Fixed!

foolip:
OK, so final deal. I sent https://github.com/mdn/browser-compat-data/pull/11145 and think we should go with the 6.1&rarr;7 rewrite in https://github.com/mdn/browser-compat-data/issues/9423#issuecomment-809630301.

Can you back out all the 7&rarr;6.1 changes here?

api/IDBObjectStore.json
foolip: This doesn't match https://developer.mozilla.org/en-US/docs/Web/API/IDBIndex/openKeyCursor, can you check if the data for that is also wrong?

foolip:
This all also matches the api.WindowOrWorkerGlobalScope.indexedDB.worker_support entry:
https://github.com/mdn/browser-compat-data/blob/fa9e3ca597235933a76fe8fb8ae48a57f1ece062/api/WindowOrWorkerGlobalScope.json#L901-L906

api/IDBObjectStore.json
foolip: OK, I've compared https://mdn-bcd-collector.appspot.com/tests/api/IDBObjectStore and https://mdn-bcd-collector.appspot.com/tests/api/IDBIndex in Safari 10 and 10.1, and there's a lot more than this that came in 10.1.

How about backing this out and making all of those changes as a single PR? I'm not sure I understand why the collector isn't updating all of the data, but it looks like api.IDBIndex.openKeyCursor was in Safari 8 and just temporarily gone. Something weird happened at any rate, which is better to split out.

api/IDBTransaction.json
foolip: Let's back this out too, I'm no longer trusting of the 10.1 versions.

foolip:
@vinyldarkscratch this is now only blocked on backing out some 10.1 / 10.3 versions that seems complicated.

foolip (approved):

PR #10922, opened 2021-06-10
Add Safari versions for DeviceOrientationEvent API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `DeviceOrientationEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.4).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/DeviceOrientationEvent

### DIFF
--- a/api/DeviceOrientationEvent.json
+++ b/api/DeviceOrientationEvent.json
@@ -34,7 +34,7 @@
             "version_added": "14"
           },
           "safari": {
-            "version_added": "5"
+            "version_added": false
           },
           "safari_ios": {
             "version_added": "4.2"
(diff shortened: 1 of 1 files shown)


api/DeviceOrientationEvent.json
```diff
@@ -134,10 +134,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": null
+              "version_added": false
             },
             "safari_ios": {
-              "version_added": "4.2"
+              "version_added": false
```
foolip, line 140: Yep, no `absolute` here:
https://github.com/WebKit/WebKit/blob/d8f9f3ed1bab5802f7e5ec29b22e186dca54d331/Source/WebCore/dom/DeviceOrientationEvent.idl#L38-L49

foolip (approved):

PR #10923, opened 2021-06-10
Update spec URLs for Class Fields

sideshowbarker:
The Class Fields proposal has been merged into the ES spec itself. So this change updates the spec URLs for the class-fields features to their locations in the ES spec.


Elchi3 (approved):
Thanks!

Ideally, I was hoping that w3c/browser-specs would drop proposals when they are merged into the main spec (and our tests would fail with a new browser-specs release then) but I see this is tracked at https://github.com/w3c/browser-specs/issues/307. 

So, thanks for taking care manually!

PR #10924, opened 2021-06-10
Correct data for crypto.subtle and crypto.webkitSubtle in WebKit

foolip:
version_added was wrong for the unprefixed variant, and webkitSubtle has
been removed. The two overlapped for one released.

Based on testing https://mdn-bcd-collector.appspot.com/tests/api/Crypto
in Safari 10.1, 11 and 11.1 on Sauce Labs.

### DIFF
--- a/api/Crypto.json
+++ b/api/Crypto.json
@@ -140,19 +140,21 @@
             },
             "safari": [
               {
-                "version_added": "10.1"
+                "version_added": "11"
               },
               {
                 "version_added": "7",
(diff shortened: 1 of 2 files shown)


foolip:
Since this was shipped in Safari 7, I also went back and checked that the Chrome data was correct, and it is. The unprefixed `crypto.subtle` shipped in 37, and before that nothing.

vinyldarkscratch (approved):
Nice catch, @foolip!  Thanks for the PR!

PR #10925, opened 2021-06-10
Adds reporting API and other reports

rachelandrew:
Making a PR I need to get some feedback on this. 

The Reporting API is one of the things we're documenting over at Google and it was missing BCD.

This includes the Reporting API interfaces themselves, but also some reports which are now in separate specs (those Interfaces that are documented on MDN already refer to the old Reporting API spec which included them).

So this includes: 

- `Report`
- `ReportBody` 

Which are part of the Reporting API: https://w3c.github.io/reporting/#interface-reporting-observer
Chrome Status for the Reporting API: https://www.chromestatus.com/feature/4672626140119040

I believe that the original reports that were part of the Reporting API:

- `CrashReportBody`
- `DeprecationReportBody`
- `InterventionReportBody`

were shipped at the same time, but I'm having trouble confirming.

I also have `PermissionsPolicyViolationReportBody` from the Permissions Policy Spec: https://w3c.github.io/webappsec-permissions-policy/#permissionspolicyviolationreportbody

Chrome Status for the header/that spec although I know this has all recently changed from Feature Policy https://www.chromestatus.com/feature/5745992911552512 

`CSPViolationReportBody` from the Content Security Policy spec: https://www.w3.org/TR/CSP3/#cspviolationreportbody

Then there are two interfaces which seem to exist in Chrome IDLs but I can't find a spec for, so I haven't included them in this PR so far.

- `CoopViolationReportBody` which I only find mentioned in an explainer though it appears to have shipped in Chrome: https://www.chromestatus.com/feature/5755687994916864
- `DocumentPolicyViolationReportBody` which I can't find anywhere.

I did however find this issue: https://github.com/w3c/reporting/issues/216 

So that's where I got to, and I figured it might be easier to open this then ping relevant people to help me fill in the blanks.

### DIFF
--- a/api/CSPViolationReportBody.json
+++ b/api/CSPViolationReportBody.json
@@ -6,10 +6,10 @@
         "spec_url": "https://www.w3.org/TR/CSP3/#cspviolationreportbody",
         "support": {
           "chrome": {
-            "version_added": "69"
+            "version_added": "74"
           },
           "chrome_android": {
-            "version_added": "69"
+            "version_added": "74"
           },
           "edge": {
             "version_added": "79"
@@ -39,7 +39,7 @@
             "version_added": "10.0"
           },
           "webview_android": {
-            "version_added": "69"
+            "version_added": "74"
           }
         },
         "status": {
@@ -54,10 +54,10 @@
           "spec_url": "https://www.w3.org/TR/CSP3/#dom-cspviolationreportbody-blockedurl",
           "support": {
             "chrome": {
-              "version_added": "69"
+              "version_added": "74"
             },
             "chrome_android": {
-              "version_added": "69"
+              "version_added": "74"
             },
             "edge": {
               "version_added": "79"
@@ -87,7 +87,7 @@
               "version_added": "10.0"
             },
             "webview_android": {
-              "version_added": "69"
+              "version_added": "74"
             }
           },
           "status": {
@@ -103,10 +103,10 @@
           "spec_url": "https://www.w3.org/TR/CSP3/#dom-cspviolationreportbody-columnnumber",
           "support": {
             "chrome": {
-              "version_added": "69"
+              "version_added": "74"
             },
             "chrome_android": {
-              "version_added": "69"
+              "version_added": "74"
             },
             "edge": {
               "version_added": "79"
@@ -136,7 +136,7 @@
               "version_added": "10.0"
             },
             "webview_android": {
-              "version_added": "69"
+              "version_added": "74"
             }
           },
           "status": {
@@ -152,10 +152,10 @@
           "spec_url": "https://www.w3.org/TR/CSP3/#dom-cspviolationreportbody-disposition",
           "support": {
             "chrome": {
-              "version_added": "69"
+              "version_added": "74"
             },
             "chrome_android": {
-              "version_added": "69"
+              "version_added": "74"
             },
             "edge": {
               "version_added": "79"
@@ -185,7 +185,7 @@
               "version_added": "10.0"
             },
             "webview_android": {
-              "version_added": "69"
+              "version_added": "74"
             }
           },
           "status": {
@@ -201,10 +201,10 @@
           "spec_url": "https://www.w3.org/TR/CSP3/#dom-cspviolationreportbody-documenturl",
           "support": {
             "chrome": {
-              "version_added": "69"
+              "version_added": "74"
             },
             "chrome_android": {
-              "version_added": "69"
+              "version_added": "74"
             },
             "edge": {
               "version_added": "79"
@@ -234,7 +234,7 @@
               "version_added": "10.0"
             },
             "webview_android": {
-              "version_added": "69"
+              "version_added": "74"
             }
           },
           "status": {
@@ -250,10 +250,10 @@
           "spec_url": "https://www.w3.org/TR/CSP3/#dom-cspviolationreportbody-effectivedirective",
           "support": {
             "chrome": {
-              "version_added": "69"
+              "version_added": "74"
             },
             "chrome_android": {
-              "version_added": "69"
+              "version_added": "74"
             },
             "edge": {
               "version_added": "79"
@@ -283,7 +283,7 @@
               "version_added": "10.0"
             },
             "webview_android": {
-              "version_added": "69"
+              "version_added": "74"
             }
           },
           "status": {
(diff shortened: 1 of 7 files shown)


jpmedley:
@foolip Would you mind reviewing this instead? I think you're more tuned into BCD issues these days than I am.

api/ReportBody.json
```diff
@@ -0,0 +1,102 @@
+{
+  "api": {
+    "Report": {
```
foolip, line 3:
```suggestion
    "ReportBody": {
```

api/ReportBody.json
```diff
@@ -0,0 +1,102 @@
...
+        },
+        "status": {
+          "experimental": false,
+          "standard_track": true,
+          "deprecated": false
+        }
+      },
+      "toJSON": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/Report/toJSON",
```
foolip, line 53:
```suggestion
          "mdn_url": "https://developer.mozilla.org/docs/Web/API/ReportBody/toJSON",
```

Except, I don't think this page will ever be created, so should `mdn_url` just be omitted here? I know @sideshowbarker feels pretty strongly that we shouldn't include broken URLs. That's tricky to guarantee in general, but in this case it's easy to predict and avoid.

api/CSPViolationReportBody.json
```diff
@@ -0,0 +1,641 @@
+{
+  "api": {
+    "CSPViolationReportBody": {
+      "__compat": {
+        "mdn_url": "https://developer.mozilla.org/docs/Web/API/CSPViolationReportBody",
+        "spec_url": "https://www.w3.org/TR/CSP3/#cspviolationreportbody",
+        "support": {
+          "chrome": {
+            "version_added": "69"
```
foolip, line 9: Testing in https://mdn-bcd-collector.appspot.com/tests/api/CSPViolationReportBody suggests this isn't shipped yet.

But that's because none of these interface objects are exposed. Without them, one would have to dig through Chromium source to figure out when each one shipped.

foolip:
@clelland this is a PR to add data for Report and *ReportBody APIs to BCD, for MDN compat tables. Due to https://crbug.com/1201273 it's hard to figure out when each part was shipped. Perhaps you know which runtime flags were involved here and when they were flipped?

maudnals:
Hi, I'm a web DevRel and have been working on this API.
Sorry for joining this thread just now!
@clelland has all details on this, but a few bits below, to the best of my knowledge:

### Demo
I build a report generation demo page and in particular this [endpoint](https://reports-endpoint.glitch.me/) where reports of various types are sent and their structure can be visualized. The [endpoint](https://reports-endpoint.glitch.me/) page provides instructions on how to generate the reports.
Run it with Chrome Canary and use the CLI flag `--short-reporting-delay` to get reports immediately.
Attached is a screenshot of a few reports.

### Supported reports
Reports that I've seen (empirically) generated and properly sent in Chrome 92+ are: 
- CSP Level 3 violation
- Deprecation
- COOP violation
- COEP violation
- Document-policy violation reports
- Crash reports
- Intervention reports

Permissions policy violation reports were generated too but AFAIK these are only experimental (source: @clelland).

### (probably not helpful) Chrome versions

`Reporting-Endpoints` is supported starting from 91 or 92 (@clelland ?)
However note that the full feature set for the new Reporting API v1 isn't present in Chrome yet.
IIUC this shouldn't impact the version from which the reports themselves were generated for each policy, so I don't think this piece of info is helpful here.


image

maudnals:
@rachelandrew Does the demo help answer the initial question? If not, what would help unblock this?
Additionally: Safari now supports Reporting API (via `report-to`) for COOP ([bug](https://bugs.webkit.org/show_bug.cgi?id=230046)).

maudnals:
Noted! 

@clelland, would you please help clarify when each part of this has shipped, so that this can be documented on MDN?

@rachelandrew, one question (this may not be in scope for this PR, so let me know if there's a better place to discuss this!): 
Chrome 96 will ship the Reporting API v1; it's configured via the `Reporting-Endpoints` header as opposed to the legacy `Report-To` header. The Reporting API v1 sends exactly the same report format and content type as v0. 
Will it be possible to document v1 on MDN? Should this live on the [Reporting API](https://developer.mozilla.org/en-US/docs/Web/API/Reporting_API) page?
- [Overview of the differences between v1 and v0](https://deploy-preview-5580--web-dev-staging.netlify.app/reporting-api-migration/#differences-between-v0-and-v1)
- [v0 spec](https://www.w3.org/TR/reporting/)
- [v1 spec](https://w3c.github.io/reporting/)
- [bug](https://bugs.chromium.org/p/chromium/issues/detail?id=1156814)

rachelandrew:
- ReportingObserver looks correct in BCD.
- I think generally this PR represents reality.
- the `document-policy` and other headers will need documenting here: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers
- Chrome has `DocumentPolicyViolationReportBody` but I don't understand from the above how that should be documented. Looking at the discussion should it be the same as https://w3c.github.io/webappsec-permissions-policy/#permissionspolicyviolationreportbody ? 

As far as the differences between v0 and v1, then yes that's something that could go onto MDN. It should also be represented in BCD, probably by including the availability of the `Reporting-Endpoints` header. Given that v0 is not going to be supported cross-browser, perhaps the thing to do is to essentially make v1 the primary version and v0 as an earlier version. Though I'd  need to get some advice from the MDN folks.

I think probably the best thing to do right now is to make sure this PR, which represents things that are in specs is accurate and see if we can get it merged, then look at dealing with the other bits in separate PRs.

Also, just a note that I'm happy to help get this over the line but it was part of work I was doing as a contractor, and I'm now a FT Googler so apologies for slowness of responses as I'm not spending 20 hours a week putting things on MDN right now. :)

api/CSPViolationReportBody.json
queengooborg: It appears that this API had actually landed in Chrome 74 as per https://storage.googleapis.com/chromium-find-releases-static/88d.html#88dcf893609cf4916ee33aef1f7658f18f13f760.  @rachelandrew, do you mind updating this?

rachelandrew: updated

queengooborg (approved):
Thank you, this is LGTM!

PR #10927, opened 2021-06-10
Update Safari versions for VRPose API

vinyldarkscratch:
This PR updates and corrects the real values for Safari (Desktop and iOS/iPadOS) for the `VRPose` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.4).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/VRPose

### DIFF
--- a/api/VRPose.json
+++ b/api/VRPose.json
@@ -46,7 +46,8 @@
             "version_added": null
           },
           "safari": {
-            "version_added": false
+            "version_added": "11.1",
+            "version_removed": "14"
           },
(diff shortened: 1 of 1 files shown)


foolip:
Any idea what happened here? Why was just this interface exposed and removed? If it was just the lone interface object with no way to actually get an instance, I think it should still be set to false, like `MediaMetadata` in https://github.com/mdn/browser-compat-data/pull/10129#discussion_r622026120

PR #10934, opened 2021-06-10
Note onmessageerror attributes in Safari as not really supported

foolip:
See https://bugs.webkit.org/show_bug.cgi?id=171216#c2

Part of an unfortunate series of events called the web platform.


vinyldarkscratch (approved):
Yikes...  Great find, @foolip, the wording looks perfect to me and will most definitely be useful to web developers!

Issue #10938, opened 2021-06-10
Firefox implements ResizeObserverEntry.contentBoxSize as an object, not an array


wbamberg:
See https://github.com/mdn/content/issues/3276, and the Firefox issue at https://bugzilla.mozilla.org/show_bug.cgi?id=1689645.

It looks like [`ResizeObserverEntry.contentBoxSize`](https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserverEntry/contentBoxSize) should be an array according to the spec, and is implemented as such in Chrome. But in Firefox it's an object.

It would be helpful to have a note in BCD capturing this.

PR #10939, opened 2021-06-10
Note Firefox implementation of contentBoxSize

wbamberg:
Fixes https://github.com/mdn/browser-compat-data/issues/10938 . 

I've tested this using the MDN example at https://developer.mozilla.org/en-US/docs/Web/API/ResizeObserver#examples, and it's also filed as a Firefox bug at https://bugzilla.mozilla.org/show_bug.cgi?id=1689645.

This update is part of https://github.com/mdn/content/issues/3276.

I don't know what the Firefox for Android situation is so I've left it.

### DIFF
--- a/api/ResizeObserverEntry.json
+++ b/api/ResizeObserverEntry.json
@@ -112,7 +112,8 @@
               "version_added": "84"
             },
             "firefox": {
-              "version_added": "69"
+              "version_added": "69",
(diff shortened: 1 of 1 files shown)


Elchi3 (changes requested):
Seems like this is significant enough to mark the Firefox support as partial?

Elchi3 (approved):

PR #10951, opened 2021-06-11
HTTP spec urls for rtt, downlink, ect headers 

hamishwillee:
I'm getting ready to do the docs for these HTTP headers. This adds spec urls. All three are in https://wicg.github.io/netinfo/


Elchi3 (approved):
Thanks! 👍

Issue #10734, opened 2021-06-03
BCD loader crashes when mixin feature included in both files


queengooborg:
When working on #10724, I ran into an issue where the `extend()` function in `index.js` crashes, throwing an error that `"Both target and source must be plain objects"`.  After doing some research, I was able to determine two causes for this error.  The first is due to invisible files created by the system (which I've addressed in #10733).  The second is a little more troublesome, which is why I'm opening this issue.

The reason behind this boils down to two separate files containing the exact same feature.  While demixing `LinkStyle`, I hadn't noticed that the feature it exposes on `HTMLStyleElement`was also documented in the dedicated API's JSON, so when creating the new mixin file, both `api/HTMLStyleElement.json` and `api/_mixins/LinkStyle__HTMLStyleElement.json` expose `api.HTMLStyleElement.sheet`.  The `load()` function loads the data from the first file, and then when it gets to the second file, notices that we already have a `sheet` member inside of `api.HTMLStyleElement`, so it [performs a recursive `extend()`](https://github.com/mdn/browser-compat-data/blob/71a3bca95a9cf385d2d73365c1cc26f8e93d0f5c/index.js#L48) through the feature.  That is, until it gets to the `mdn_url` member, which it attempts to perform a recursive `extend()` upon, but since it throws an error if either variable isn't a plain object (and `mdn_url`'s value is a string), the whole function crashes.

Now of course, we wouldn't want to have duplicate data anyways, and should probably have some sort of linter or otherwise to prevent it.  However, an error message such as the one we're getting is just plain unhelpful in figuring out what exactly is going on.

foolip:
Throwing errors when anything other than an object appears in two places was a deliberate change, see the discussion in https://github.com/mdn/browser-compat-data/pull/9821#discussion_r612582315.

However, the error it throws isn't very obvious I guess.

PR #10954, opened 2021-06-11
Refactor index.js to allow for better error messages

foolip:
In order to print a useful error message, we need to know both which
file is being loaded, and the JSON path at which the problem occurred.

As a not-strictly-needed improvement, change how directories are
traversed to only create a single result object and extend that with
each file in order, avoiding intermediate result objects in the previous
recursive load() calls.

Fixes https://github.com/mdn/browser-compat-data/issues/10734.

### DIFF
--- a/index.js
+++ b/index.js
@@ -2,57 +2,69 @@
 const fs = require('fs');
 const path = require('path');
 
-function load() {
-  // Recursively load one or more directories passed as arguments.
-  let dir,
-    result = {};
+function load(...dirs) {
+  const result = {};
 
-  function processFilename(fn) {
-    const fp = path.join(dir, fn);
-    let extra;
+  // Traverse directories depth-first by maintaining a stack. Reverse the given
+  // dirs so that we pop the first directory first. All subdirectories of that
+  // will be pushed to the stack and popped before the next top-level directory
+  // is popped. Directory paths on the stack are always relative to __dirname.
+  const stack = dirs.reverse();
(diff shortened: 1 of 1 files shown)


foolip:
This sort of works if you'd like to review, @vinyldarkscratch, but I'm leaving it as draft.

The big remaining question is that this changes the order of things in the resulting JSON, despite my attempt to traverse directories in the same order. @ddbeck do you think we should be concerned about the key order in the JSON? I'd like to suggest that we actually sort all keys alphabetically, like what `sort_keys=True` does in Python. Then it would be easier to tell that this change doesn't have any impact on the result.

ddbeck:
Oh, interesting! My inclination is to not mess with the order (I know nobody should _rely_ on the order, and yet…), but I don't know that it's a real difference.

I started down a different path to fix this yesterday: https://github.com/ddbeck/browser-compat-data/compare/main...ddbeck:friendlier-extend-errors. My version breaks the tests (I guess because I've created a circular dependency) and I wasn't sure how to fix that just yet.

foolip:
So the challenge in not messing with the order is that it depends on details about how the current implementation traverses, and it's not really a sensible order to my eye. For example, all mixin data will end up appended as the last keys of their target interface.

But I don't feel strongly, I might abandon this PR if the error messages are improved in some other way.

q805:
可拉取請求合併嗎？謝謝

PR #10970, opened 2021-06-12
Add missing spec_url in CSS* apis

teoli2003:
Added a few missing spec_url on some old CSSOM* apis.


teoli2003:
I'm closing this. I will try to do it by having a different text directly on MDN Web docs. That way `spec_url` will stay only for document on the standard track.

PR #10976, opened 2021-06-12
Add APIs for custom state pseudo class

foolip:
Spec: https://wicg.github.io/custom-state-pseudo-class/.

Most of the API surface comes from the `setlike<DOMString>` declaration,
which can't be linked directly, so no `spec_url` for those.

The flag was enabled for Chromium 90 here:
https://storage.googleapis.com/chromium-find-releases-static/78f.html#78fbd97dbf148a448be60f038701037ac6cb9151

This hasn't reached Opera for Android or Samsung Internet yet.

### DIFF
--- a/api/CustomStateSet.json
+++ b/api/CustomStateSet.json
@@ -1,18 +1,17 @@
 {
   "api": {
-    "ElementInternals": {
+    "CustomStateSet": {
       "__compat": {
-        "mdn_url": "https://developer.mozilla.org/docs/Web/API/ElementInternals",
-        "spec_url": "https://html.spec.whatwg.org/multipage/custom-elements.html#the-elementinternals-interface",
+        "spec_url": "https://wicg.github.io/custom-state-pseudo-class/#customstateset",
         "support": {
           "chrome": {
-            "version_added": "77"
+            "version_added": "90"
           },
           "chrome_android": {
-            "version_added": "77"
+            "version_added": "90"
           },
           "edge": {
-            "version_added": "79"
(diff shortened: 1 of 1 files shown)


lucalves (approved):

ddbeck:
> Most of the API surface comes from the `setlike<DOMString>` declaration,
> which can't be linked directly, so no `spec_url` for those.

@Elchi3 or @sideshowbarker: is this consistent with what y'all have been doing with spec_urls?

sideshowbarker:
> > Most of the API surface comes from the `setlike<DOMString>` declaration,
> > which can't be linked directly, so no `spec_url` for those.
> 
> @Elchi3 or @sideshowbarker: is this consistent with what y'all have been doing with spec_urls?

For interfaces with `maplike` declarations, we do consistently have a `spec_url` for those, which is the same URL in every case: https://heycam.github.io/webidl/#idl-maplike

(and for interfaces with `iterable` declarations, we do consistently have a `spec_url` for those, which is the same URL in every case: https://heycam.github.io/webidl/#idl-iterable)

So it seems like what would be consistent for interfaces with a `setlike` declarations is, we add a `spec_url` with a link to https://heycam.github.io/webidl/#idl-setlike.

foolip:
See https://github.com/mdn/browser-compat-data/issues/11426 about linking to Web IDL in `spec_url`. I'm doing it here anyway to follow current style and make it easy to find them all for fixing.

sideshowbarker (approved):

PR #10980, opened 2021-06-13
Add spec_url to FileSystemDirectoryReader[\.*]?

teoli2003:
Add spec_url link to all entries (2) related to FileSystemDirectoryReader.


Elchi3 (commented):
This spec is also in browser-specs 👍 Is standards track true these days? 

The MDN page has hefty warnings here but now we're adding a spec... What is correct?

teoli2003:
I have not the answer, but the File* pages on MDN are a mess, between competing specs over the years, vendor divergences, and renaming during the lifetime of spec and beta (or not!) implementation. It may be a project of its own to clean this out.

Elchi3 (approved):
Okay, let's merge this, the spec pointers seem to be correct

PR #10981, opened 2021-06-13
Add spec_url  and mdn_url to FileReader constructor

teoli2003:
FileReader() was the only entry on this interface without mdn_url and spec_url. This PR fixes it.

### DIFF
--- a/api/FileReader.json
+++ b/api/FileReader.json
@@ -52,6 +52,8 @@
       "FileReader": {
         "__compat": {
           "description": "<code>FileReader()</code> constructor",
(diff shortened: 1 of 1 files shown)


api/FileReader.json
```diff
@@ -52,6 +52,8 @@
       "FileReader": {
         "__compat": {
           "description": "<code>FileReader()</code> constructor",
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/FileReader/FileReader",
+          "spec_url": "https://w3c.github.io/FileAPI/#filereaderConstrctr",
```
foolip, line 56: Confirmed this is right even though it totally looks like a typo.

foolip (approved):

PR #10983, opened 2021-06-13
Add spec_url  and mdn_url to FetchEvent.preloadResponse

teoli2003:
Add spec_url  and mdn_url to FetchEvent.preloadResponse

### DIFF
--- a/api/FetchEvent.json
+++ b/api/FetchEvent.json
@@ -299,6 +299,8 @@
       },
       "preloadResponse": {
         "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/FetchEvent/PreloadResponse",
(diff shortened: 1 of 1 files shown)


api/FetchEvent.json
```diff
@@ -299,6 +299,8 @@
       },
       "preloadResponse": {
         "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/FetchEvent/PreloadResponse",
```
foolip, line 302: The capitalization here is wrong, does it need to be renamed/redirected on MDN?

teoli2003: Yes, I think so. I'll open a PR there (tomorrow). Once fixed I'll commit the requested change here.

foolip (approved):

PR #10985, opened 2021-06-13
Add all browsers versions for api.SVGElement.*_event

vinyldarkscratch:
This PR adds real values for all browsers for the events of the `SVGElement` API.  The data for theese events was determined by a combination of copying data from the event handlers (from `GlobalEventHandlerers`) and manual testing (for `load_event` and `unload_event` especially).

### DIFF
--- a/api/SVGElement.json
+++ b/api/SVGElement.json
@@ -358,11 +358,13 @@
             },
             "firefox": {
               "version_added": "4",
-              "alternative_name": "SVGLoad"
+              "alternative_name": "SVGLoad",
+              "notes": "See <a href='https://bugzill.la/620002'>bug 620002</a> for implementation status of the standard <code>load</code> event."
             },
             "firefox_android": {
               "version_added": "4",
-              "alternative_name": "SVGLoad"
+              "alternative_name": "SVGLoad",
+              "notes": "See <a href='https://bugzill.la/620002'>bug 620002</a> for implementation status of the standard <code>load</code> event."
             },
             "ie": {
               "version_added": "9"
--- a/api/SVGElement.json
+++ b/api/SVGElement.json
@@ -55,40 +55,40 @@
           "spec_url": "https://dom.spec.whatwg.org/#eventdef-abortsignal-abort",
           "support": {
             "chrome": {
-              "version_added": "1"
+              "version_added": null
             },
             "chrome_android": {
-              "version_added": "18"
+              "version_added": null
             },
             "edge": {
-              "version_added": "12"
+              "version_added": null
             },
             "firefox": {
-              "version_added": "9"
+              "version_added": null
             },
             "firefox_android": {
-              "version_added": "9"
+              "version_added": null
             },
             "ie": {
-              "version_added": "9"
+              "version_added": null
             },
             "opera": {
-              "version_added": "≤12.1"
+              "version_added": null
             },
             "opera_android": {
-              "version_added": "≤12.1"
+              "version_added": null
             },
             "safari": {
-              "version_added": "3"
+              "version_added": null
             },
             "safari_ios": {
-              "version_added": "1"
+              "version_added": null
             },
             "samsunginternet_android": {
-              "version_added": "1.0"
+              "version_added": null
             },
             "webview_android": {
-              "version_added": "≤37"
+              "version_added": null
             }
           },
           "status": {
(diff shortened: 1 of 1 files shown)


foolip:
> copying data from the event handlers (from `GlobalEventHandlerers`)

I have very low confidence in the existing data here since `GlobalEventHandlers` is a mixin, and also because the existence or non-existence of event handler properties has been all over the place historically, and mishaps continue to happen now as well.

I'll review the diff and see where we need to look closer.

api/SVGElement.json
```diff
@@ -55,40 +55,40 @@
           "spec_url": "https://dom.spec.whatwg.org/#eventdef-abortsignal-abort",
```
foolip, line 55: You didn't add this, but this spec URL is wrong, and furthermore I can't find that this event would ever be fired.

As far as I can tell, `onabort` is on `SVGElement` only because it's in `GlobalEventHandlers`, and nothing would ever fire such an event on any `SVGElement`, the only type of `Element` were it's fired is `HTMLMediaElement`.

So I think we should remove this entry and https://developer.mozilla.org/en-US/docs/Web/API/SVGElement/abort_event.

api/SVGElement.json
```diff
@@ -201,40 +201,40 @@
           "spec_url": "https://svgwg.org/svg2-draft/interact.html#ErrorEvent",
```
foolip, line 201: This should be possible to test with a [structurally external element](https://svgwg.org/svg2-draft/struct.html#TermStructurallyExternalElement), and `<image>` is probably the best since it's always been supported:
```svg
<svg width="200" height="200" xmlns="http://www.w3.org/2000/svg">
  <script>
  // <![CDATA[
  document.addEventListener('error', function(event) {
    alert(event);
  }, true);
  // ]]>
  </script>
  <image href="404.png" height="200" width="200"/>
</svg>
```

Can you confirm the data with a test like that?

api/SVGElement.json
```diff
@@ -348,42 +348,42 @@
...
+              "version_added": "18"
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "12"
             },
             "firefox": {
-              "version_added": true,
+              "version_added": "4",
               "alternative_name": "SVGLoad"
```
foolip, line 361: Does Firefox still not fire a "load" event? I expect this is wrong, that in fact "SVGLoad" is no longer being fired and there's just "load". A test for this can probably also be written using `<image>`.

api/SVGElement.json
```diff
@@ -675,40 +675,40 @@
           "spec_url": "https://drafts.csswg.org/cssom-view/#eventdef-window-resize",
```
foolip, line 675: I think we should remove this entry, the actual event is the global "resize" event represented by api.Window.resize_event. I can't find any "resize" event being dispatched in Chromium's SVG code, at least, and nothing in the spec.

api/SVGElement.json
```diff
@@ -725,40 +725,40 @@
           "spec_url": "https://drafts.csswg.org/cssom-view/#eventdef-document-scroll",
```
foolip, line 725: Let's remove this entry too.

api/SVGElement.json
```diff
@@ -775,40 +775,40 @@
           "spec_url": "https://svgwg.org/svg2-draft/interact.html#UnloadEvent",
```
foolip, line 775: And this too :)

foolip (changes requested):

api/SVGElement.json
vinyldarkscratch: Removing this and the other three events in #11071!

api/SVGElement.json
vinyldarkscratch: Surprisingly enough, it doesn't appear to, at least if we added the event listener to the `<svg>` or `<image>` elements rather than the document.

foolip: But it does fire a "SVGLoad" event?

vinyldarkscratch: Precisely, from Firefox 4 to present.

foolip: Huh. Can you link https://bugzilla.mozilla.org/show_bug.cgi?id=620002 in a note so it's easy to figure this out the next time around?

vinyldarkscratch: Thanks for finding the bug, added!

138166234:
82398981

1381662341mohamadsalmanpor

در تاریخ پنجشنبه ۱۷ ژوئن ۲۰۲۱،‏ ۱۳:۱۱ Queen Vinyl Darkscratch <
***@***.***> نوشت:

api/SVGElement.json
```diff
@@ -348,42 +348,44 @@
...
             "edge": {
-              "version_added": "≤79"
+              "version_added": "12"
             },
             "firefox": {
-              "version_added": true,
-              "alternative_name": "SVGLoad"
+              "version_added": "4",
+              "alternative_name": "SVGLoad",
+              "notes": "See <a href='https://bugzill.la/620002'>bug 620002</a> for implementation status of the standard <code>load</code> event."
```
foolip, line 362:
```suggestion
              "notes": "See <a href='https://bugzil.la/620002'>bug 620002</a> for implementation status of the standard <code>load</code> event."
```

api/SVGElement.json
```diff
@@ -348,42 +348,44 @@
...
             },
             "firefox_android": {
-              "version_added": true,
-              "alternative_name": "SVGLoad"
+              "version_added": "4",
+              "alternative_name": "SVGLoad",
+              "notes": "See <a href='https://bugzill.la/620002'>bug 620002</a> for implementation status of the standard <code>load</code> event."
```
foolip, line 367:
```suggestion
              "notes": "See <a href='https://bugzil.la/620002'>bug 620002</a> for implementation status of the standard <code>load</code> event."
```

foolip (approved):
Bugs links didn't work, otherwise 👍

PR #10988, opened 2021-06-13
Add spec_url  and mdn_url to HTMLImageElement.loading

teoli2003:
HTMLImageElement.loading had neither a spec_url nor a mdn_url entries. This PR fixes it.

### DIFF
--- a/api/HTMLImageElement.json
+++ b/api/HTMLImageElement.json
@@ -688,6 +688,8 @@
       },
       "loading": {
         "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLImageElement/loading",
+          "spec_url": "https://html.spec.whatwg.org/multipage/embedded-content.html#attr-img-loading",
           "support": {
             "chrome": {
               "version_added": "77"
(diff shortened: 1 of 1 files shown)


api/HTMLImageElement.json
```diff
@@ -688,6 +688,8 @@
       },
       "loading": {
         "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLImageElement/loading",
+          "spec_url": "https://html.spec.whatwg.org/multipage/embedded-content.html#attr-img-loading",
```
foolip, line 692: This URL is for the content attribute. https://html.spec.whatwg.org/multipage/embedded-content.html#dom-img-loading is for the IDL attribute and what should be used here.

foolip (changes requested):

api/HTMLImageElement.json
teoli2003: Oh. Good catch. I copied (removing redirections) the spec link that was on the page.

Fix in next commit.

Elchi3 (approved):

foolip:
Thanks @teoli2003! I expect there's a lot of conflation between content attributes and IDL attributes in existing spec URLs, and in fact there isn't always something linkable for both in more handwavy specs, so it's easy to understand why it's happened.

Rant: If all content attributes had always been correctly reflected as IDL attributes, no web developer would ever need to think of them as separate things, much like CSS properties are just one thing &ndash; nobody needs to worry about the difference between `background-color: red` in CSS and `style.backgroundColor = 'red'` in JS.

PR #10989, opened 2021-06-13
Add spec_url  and mdn_url to HTMLInputElement.select

teoli2003:
The select property was missing, though being there since more than a decade. Fixed.

### DIFF
--- a/api/HTMLInputElement.json
+++ b/api/HTMLInputElement.json
@@ -1864,6 +1864,8 @@
       },
       "select": {
         "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLInputElement/select",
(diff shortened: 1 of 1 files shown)


api/HTMLInputElement.json
```diff
@@ -1864,6 +1864,8 @@
       },
       "select": {
         "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLInputElement/select",
+          "spec_url": "https://html.spec.whatwg.org/multipage/form-control-infrastructure.html#dom-textarea%2Finput-select",
```
foolip, line 1868: URL is unusual with the percent encoded bit but it checks out.

foolip (approved):

PR #10999, opened 2021-06-14
Add Safari versions for AudioNodeOptions API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `AudioNodeOptions` API.  The data is copied from api.AnalyserNode.AnalyserNode, as it is for all the other browsers.

### DIFF
--- a/api/AudioNodeOptions.json
+++ b/api/AudioNodeOptions.json
@@ -30,10 +30,10 @@
             "version_added": "42"
           },
           "safari": {
-            "version_added": null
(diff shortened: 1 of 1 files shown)


foolip:
> The data is copied from api.AnalyserNode.AnalyserNode, as it is for all the other browsers.

That makes sense, because this dictionaries or ones inheriting from it was introduced together with audio node constructors.

foolip (approved):

PR #11001, opened 2021-06-14
Add Safari versions for BeforeUnloadEvent API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `BeforeUnloadEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.4).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/BeforeUnloadEvent

### DIFF
--- a/api/BeforeUnloadEvent.json
+++ b/api/BeforeUnloadEvent.json
@@ -77,10 +77,10 @@
               "version_added": "44"
             },
             "safari": {
-              "version_added": "11"
+              "version_added": null
             },
             "safari_ios": {
-              "version_added": "11"
+              "version_added": null
             },
             "samsunginternet_android": {
               "version_added": "8.0"
--- a/api/BeforeUnloadEvent.json
+++ b/api/BeforeUnloadEvent.json
@@ -30,10 +30,10 @@
             "version_added": "18"
           },
           "safari": {
-            "version_added": true
+            "version_added": "6.1"
           },
           "safari_ios": {
-            "version_added": true
+            "version_added": "7"
           },
           "samsunginternet_android": {
             "version_added": "3.0"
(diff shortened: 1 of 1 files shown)


api/BeforeUnloadEvent.json
```diff
@@ -77,10 +77,10 @@
               "version_added": "44"
             },
             "safari": {
-              "version_added": null
+              "version_added": "11"
```
foolip, line 80: This must have involved manual testing or research, can you update the description with what you did?

api/BeforeUnloadEvent.json
```diff
@@ -30,10 +30,10 @@
             "version_added": "18"
           },
           "safari": {
-            "version_added": true
+            "version_added": "6.1"
```
foolip, line 33: Basic support for the event goes back to https://trac.webkit.org/changeset/11259/webkit which was probably Safari 3.

The interface (this entry) was added here:
https://trac.webkit.org/changeset/155367/webkit
https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Configurations/Version.xcconfig?rev=155367

That would map to Safari 8, did https://mdn-bcd-collector.appspot.com/tests/api/BeforeUnloadEvent really test true in Safari 6.1?

api/BeforeUnloadEvent.json
vinyldarkscratch: This one was tested based upon the same test as #11002.  Since we have an active discussion there regarding what this means, I'll split this off into its own PR.

api/BeforeUnloadEvent.json
vinyldarkscratch: Yeah, this is testing `true` in Safari 6.1,  The test is pretty simple too, just `"BeforeUnloadEvent" in self`.

foolip: That's interesting. But I trust testing over source code, maybe it lived in a patch outside of WebKit or something for a while.

foolip (approved):

PR #11002, opened 2021-06-14
Add versions for api.BeforeUnloadEvent.user_interaction

vinyldarkscratch:
This PR adds real values for Firefox and Firefox Android for the `user_interaction` member of the `BeforeUnloadEvent` API, based upon manual testing.

Test Code Used: `Used the example from the wiki page, and then reloaded the page to see if a dialog would come up.  No interactable elements were provided on the page.`

### DIFF
--- a/api/BeforeUnloadEvent.json
+++ b/api/BeforeUnloadEvent.json
@@ -77,10 +77,10 @@
               "version_added": "44"
             },
             "safari": {
-              "version_added": null
+              "version_added": "11"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "11"
             },
             "samsunginternet_android": {
               "version_added": "8.0"
--- a/api/BeforeUnloadEvent.json
+++ b/api/BeforeUnloadEvent.json
@@ -62,10 +62,10 @@
               "version_added": "18"
             },
             "firefox": {
-              "version_added": "61"
+              "version_added": "44"
             },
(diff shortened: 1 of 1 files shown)


foolip:
How did you reload the page? I'm pretty sure this is about not firing the event when the navigation was triggered by script without any user interaction. So a test would have to setTimeout for >3 seconds and then navigate the page.

vinyldarkscratch:
I just hit the reload button after the page loaded.  When I was writing the test in Chrome, I had reloaded the page as well, and it had thrown an error saying "blocked because there was no user interaction" (or something along those lines).

foolip:
Can you share the test you used? Here's what I've now tested:
```html
<script>
window.addEventListener("beforeunload", function(event) {
  event.returnValue = "\o/";
  event.preventDefault();
});
setTimeout(function() {
  location.reload();
}, 2500);
</script>
```

I bisected my way to a difference between Firefox 43 and 44 which I think is what this entry is about &ndash; if one never clicks anything in the page, the reload happens without prompting.

foolip:
And with that same test I can observe the difference between Chrome 59 and 60, so I think it's the behavior in question here.

vinyldarkscratch:
Hmm, now I'm not sure why I got Firefox 61 before, since the test code you posted is the same as what I used...  Correcting now!

foolip (approved):

PR #11003, opened 2021-06-14
Add Edge versions for api.BeforeUnloadEvent.user_interaction

vinyldarkscratch:
This PR adds real values for Microsoft Edge for the `user_interaction` member of the `BeforeUnloadEvent` API, based upon manual testing.

Test Code Used: `Used the example from the wiki page, and then reloaded the page to see if a dialog would come up.  No interactable elements were provided on the page.`

### DIFF
--- a/api/BeforeUnloadEvent.json
+++ b/api/BeforeUnloadEvent.json
@@ -59,7 +59,7 @@
               "version_added": "60"
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "18"
             },
             "firefox": {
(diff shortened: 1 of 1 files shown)


foolip:
I'm treating this as blocked on https://github.com/mdn/browser-compat-data/pull/11002, where we need to figure out what the entry actually means and how to test it.

vinyldarkscratch:
I'm merging this into #11002 along with an update to Safari, so we can review and merge it all in one go!

PR #11004, opened 2021-06-14
Add Safari versions for EffectTiming API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `EffectTiming` API.  The data was created based upon a combination of mirroring from `api.Element.animate`, as well as viewing the source for the file itself on WebKit Trac: https://trac.webkit.org/browser/webkit/tags/Safari-609.2.1/Source/WebCore/animation/EffectTiming.idl

### DIFF
--- a/api/EffectTiming.json
+++ b/api/EffectTiming.json
@@ -44,10 +44,10 @@
             "version_added": true
           },
           "safari": {
-            "version_added": null
+            "version_added": "13.1"
           },
           "safari_ios": {
-            "version_added": null
+            "version_added": "13.4"
           },
           "samsunginternet_android": [
             {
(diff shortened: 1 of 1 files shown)


api/EffectTiming.json
```diff
@@ -253,10 +253,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": false
+              "version_added": "13.1"
```
foolip, line 256: `easing` has a subfeature "jump- keywords for steps()" which is left false. Should it be?

vinyldarkscratch: I've updated it by copying the data from the same feature within the CSS data.

foolip (approved):

PR #11006, opened 2021-06-14
Add Safari versions for InputDeviceInfo API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `InputDeviceInfo` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.4).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/InputDeviceInfo

### DIFF
--- a/api/InputDeviceInfo.json
+++ b/api/InputDeviceInfo.json
@@ -30,10 +30,10 @@
             "version_added": true
           },
           "safari": {
-            "version_added": null
+            "version_added": false
           },
           "safari_ios": {
-            "version_added": null
+            "version_added": false
           },
           "samsunginternet_android": {
             "version_added": true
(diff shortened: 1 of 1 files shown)


vinyldarkscratch:
As an additional sanity check, I modified the test for `MediaDeviceInfo` to log all the devices and see if there were any inputs that were an `InputDeviceInfo` object instead (which the example on the MDN page indicates is how to test for support).  They weren't.  We don't have a custom test for this API, however `MediaDeviceInfo` is exposed by default in the latest Safari, so I'm confident we won't need to create a custom test for it.

foolip (approved):
Seems right, WebKit has `Promise<sequence<MediaDeviceInfo>> enumerateDevices()` in its IDL and its `getCapabilities()` methods aren't the one on `InputDeviceInfo`, they're on `MediaStreamTrack`, `RTCRtpReceiver` and RTCRtpSender`.

PR #11007, opened 2021-06-14
Add Safari versions for MediaCapabilitiesInfo API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `MediaCapabilitiesInfo` API, based upon commit history and date.

Commit: https://trac.webkit.org/browser/webkit/tags/Safari-608.2.11/Source/WebCore/platform/MediaCapabilitiesInfo.h

### DIFF
--- a/api/MediaCapabilitiesInfo.json
+++ b/api/MediaCapabilitiesInfo.json
@@ -30,10 +30,10 @@
             "version_added": null
           },
           "safari": {
-            "version_added": null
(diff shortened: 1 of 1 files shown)


foolip (approved):
This is used in https://developer.mozilla.org/en-US/docs/Web/API/Navigator/mediaCapabilities#browser_compatibility and the data matches.

PR #11010, opened 2021-06-14
Add Safari versions for MediaKeySystemConfiguration API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `MediaKeySystemConfiguration` API, based upon commit history and date.

Commit: https://trac.webkit.org/browser/webkit/tags/Safari-607.1.40.3.1/Source/WebCore/Modules/encryptedmedia/MediaKeySystemConfiguration.idl

### DIFF
--- a/api/MediaKeySystemConfiguration.json
+++ b/api/MediaKeySystemConfiguration.json
@@ -30,10 +30,10 @@
             "version_added": null
           },
           "safari": {
-            "version_added": null
+            "version_added": "13.1"
           },
           "safari_ios": {
-            "version_added": null
+            "version_added": "13.4"
           },
           "samsunginternet_android": {
             "version_added": "4.0"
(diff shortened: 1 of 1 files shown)


api/MediaKeySystemConfiguration.json
```diff
@@ -30,10 +30,10 @@
             "version_added": null
           },
           "safari": {
-            "version_added": null
+            "version_added": "13.1"
```
foolip, line 33: https://developer.mozilla.org/en-US/docs/Web/API/MediaKeySystemAccess/getConfiguration#browser_compatibility says Safari 12, which if true would mean that it returned something else than `MediaKeySystemConfiguration` in Safari 12 and 12.1. I suspect one is wrong, can you check?

vinyldarkscratch: I was looking at the wrong file, admittedly -- I've corrected it now!

foolip (approved):

PR #11011, opened 2021-06-14
Add Safari versions for MediaStreamConstraints API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `MediaStreamConstraints` API, based upon manual testing.

Test Code Used: Tested each property in Safari using `navigator.mediaDevices.getUserMedia()`

### DIFF
--- a/api/MediaStreamConstraints.json
+++ b/api/MediaStreamConstraints.json
@@ -30,10 +30,10 @@
             "version_added": true
           },
           "safari": {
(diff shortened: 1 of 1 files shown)


foolip (approved):
Matches https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getUserMedia#browser_compatibility

PR #11030, opened 2021-06-15
Chromium browsers have arbitrary caps on rtt and downlink

hamishwillee:
Chrome adds arbitrary limits on the downlink and rtt client hints. Discovered this in https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation/downlink where you can see this note:
image

The link takes you to here: https://source.chromium.org/chromium/chromium/src/+/master:third_party/blink/renderer/platform/network/network_state_notifier.cc;l=460;drc=49bf35554c123bbc44a0ef52675144eba2dd7bbc;bpv=1;bpt=1?originalUrl=https:%2F%2Fcs.chromium.org%2F where you can see `kMaxDownlinkKbps = 10.0 * 1000;`. A search finds https://chromium-review.googlesource.com/c/chromium/src/+/1002232/8/chrome/browser/client_hints/client_hints.cc#121 , which shows `static const double kMaxRttMsec = 3.0 * 1000;`

I think BCD is right place for this. Not sure if there is a better in-source link we can provide as a bug reference though (I am concerned that if Chrome changed this we might not hear about it).

If this goes in, that note in https://developer.mozilla.org/en-US/docs/Web/API/NetworkInformation/downlink should be removed.

### DIFF
--- a/api/NetworkInformation.json
+++ b/api/NetworkInformation.json
@@ -55,7 +55,7 @@
           "support": {
             "chrome": {
               "version_added": "61",
-              "notes": "The value is capped to 10Mbps as an anti-fingerprinting measure. This limit is not present in the specification."
+              "notes": "The value is never greater than 10 Mbps, as a non-standard anti-fingerprinting measure."
             },
             "chrome_android": {
               "version_added": "38",
--- a/api/NetworkInformation.json
+++ b/api/NetworkInformation.json
@@ -59,11 +59,11 @@
             },
             "chrome_android": {
               "version_added": "38",
-              "notes": "The value is capped to 10Mbps as an anti-fingerprinting measure. This limit is not present in the specification."
+              "notes": "The value is never greater than 10 Mbps, as a non-standard anti-fingerprinting measure."
             },
             "edge": {
               "version_added": "79",
-              "notes": "The value is capped to 10Mbps as an anti-fingerprinting measure. This limit is not present in the specification."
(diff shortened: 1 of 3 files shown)


api/NetworkInformation.json
```diff
@@ -54,13 +54,16 @@
           "spec_url": "https://wicg.github.io/netinfo/#dom-networkinformation-downlink",
           "support": {
             "chrome": {
-              "version_added": "61"
+              "version_added": "61",
+              "notes": "The value is capped to 10Mbps as an anti-fingerprinting measure. This limit is not present in the specification."
```
ddbeck, line 58: style: I wasn't sure "capped" was particularly explicit (and potentially unclear to newcomers to English). While I was here, I tried giving that unit some room to breath and collapsed the second sentence into "non-standard".

What do you think?

```suggestion
              "notes": "The value is never greater than 10 Mbps, as a non-standard anti-fingerprinting measure."
```

ddbeck (changes requested):
I agree that BCD is the right place for this. Style suggestions for the notes, in a line comment. Thanks, @hamishwillee!

api/NetworkInformation.json
hamishwillee: Thanks - it is certainly more succinct. I will roll out to the other cases.

ddbeck (approved):
Thank you!

PR #11033, opened 2021-06-15
Add PaymentMethodChangeEvent constructor

teoli2003:
Adding the PaymentMethodChangeEvent constructor that was missing.

Sources:
1. It is in standard: https://w3c.github.io/payment-request/#dom-paymentmethodchangeevent-constructor
2. It is on MDN: https://developer.mozilla.org/en-US/docs/Web/API/PaymentMethodChangeEvent/PaymentMethodChangeEvent
3. It is implemented in Chromium: https://chromium.googlesource.com/chromium/src/+blame/03a4cbda6cf358d742f059aae5c73852aa38cd3d/third_party/blink/renderer/modules/payments/payment_method_change_event.idl (line 10) and the blame function tel us this idl has never been edited, so it has been done at the same time as the initial creation.

### DIFF
--- a/api/PaymentMethodChangeEvent.json
+++ b/api/PaymentMethodChangeEvent.json
@@ -66,12 +66,10 @@
               "version_added": "79"
             },
             "firefox": {
-              "version_added": false,
-              "notes": "Available only in nightly builds. Requires <code>dom.payments.request.enabled</code> to be set to <code>true</code> and the comma-delineated list in <code>dom.payments.request.supportedRegions</code> to contain one or more of the supported 2-character ISO locales, currently US and CA."
+              "version_added": false
             },
             "firefox_android": {
-              "version_added": false,
-              "notes": "Available only in nightly builds. Requires <code>dom.payments.request.enabled</code> to be set to <code>true</code> and the comma-delineated list in <code>dom.payments.request.supportedRegions</code> to contain one or more of the supported 2-character ISO locales, currently US and CA."
+              "version_added": false
             },
             "ie": {
               "version_added": false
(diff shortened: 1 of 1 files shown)


api/PaymentMethodChangeEvent.json
```diff
@@ -50,6 +50,58 @@
...
+              "version_added": false
+            },
+            "opera": {
+              "version_added": "63"
+            },
+            "opera_android": {
+              "version_added": "54"
+            },
+            "safari": {
+              "version_added": "12.1"
```
foolip, line 86: The description gave evidence for Chromium, is this also based on source inspection, or testing? https://mdn-bcd-collector.appspot.com/tests/api/PaymentMethodChangeEvent confirms it's currently supported at least.

(Note that mdn-bcd-collector could probably fill in most of this data, avoiding the need to look at source code.)

api/PaymentMethodChangeEvent.json
```diff
@@ -50,6 +50,58 @@
...
+            },
+            "chrome_android": {
+              "version_added": "76"
+            },
+            "edge": {
+              "version_added": "79"
+            },
+            "firefox": {
+              "version_added": false,
+              "notes": "Available only in nightly builds. Requires <code>dom.payments.request.enabled</code> to be set to <code>true</code> and the comma-delineated list in <code>dom.payments.request.supportedRegions</code> to contain one or more of the supported 2-character ISO locales, currently US and CA."
```
foolip, line 70: Repeating the notes from the parent feature doesn't seem that helpful, and isn't done for the other entries.

api/PaymentMethodChangeEvent.json
foolip: Never mind, I've confirmed this with https://github.com/foolip/mdn-bcd-collector#updating-bcd-using-the-results and let's assume iOS matches.

api/PaymentMethodChangeEvent.json
teoli2003: Fixed in next commit.

api/PaymentMethodChangeEvent.json
teoli2003: So, kept as it i.

foolip (approved):

PR #11046, opened 2021-06-16
Add Safari iOS versions for api.OfflineAudioContext.oncomplete/complete_event

vinyldarkscratch:
This PR adds real values for Safari iOS/iPadOS for the `oncomplete/complete_event` member of the `OfflineAudioContext` API by mirroring the data.

### DIFF
--- a/api/OfflineAudioContext.json
+++ b/api/OfflineAudioContext.json
@@ -292,7 +292,7 @@
               "version_added": "6"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "6"
             },
             "samsunginternet_android": {
               "version_added": "1.5"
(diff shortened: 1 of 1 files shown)


api/OfflineAudioContext.json
```diff
@@ -292,7 +292,7 @@
               "version_added": "6"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "6"
```
foolip, line 295: I tried https://mdn-bcd-collector.appspot.com/tests/api/webkitOfflineAudioContext in iOS 6 (BrowserStack simulator) and it looks like it's not supported, while https://mdn-bcd-collector.appspot.com/tests/api/webkitAudioContext is.

But it is in iOS 7 on a real device.

It seems like maybe `OfflineAudioContext` came a release later, but this is consistent with the existing data, and the difference is minor, so I'm going to look away and say it's fine.

foolip (approved):
👀

PR #11051, opened 2021-06-16
Add Safari iOS versions for api.RTCPeerConnection.*_event

vinyldarkscratch:
This PR adds real values for Safari iOS/iPadOS for the event members of the `RTCPeerConnection` API by mirroring the data.

### DIFF
--- a/api/RTCPeerConnection.json
+++ b/api/RTCPeerConnection.json
@@ -332,7 +332,7 @@
               "version_added": false
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": false
             },
             "samsunginternet_android": {
               "version_added": "6.0"
@@ -1093,7 +1093,7 @@
               "version_added": "11"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "11"
             },
             "samsunginternet_android": {
               "version_added": "6.0"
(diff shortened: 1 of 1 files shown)


api/RTCPeerConnection.json
```diff
@@ -1093,7 +1093,7 @@
               "version_added": "11"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "11"
```
foolip, line 1096: Matches https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel#browser_compatibility

api/RTCPeerConnection.json
```diff
@@ -1862,7 +1862,7 @@
...
```
foolip, line 1865: Matches https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidate#browser_compatibility

api/RTCPeerConnection.json
```diff
@@ -4453,7 +4453,7 @@
...
```
foolip, line 4456: Matches https://developer.mozilla.org/en-US/docs/Web/API/RTCTrackEvent#browser_compatibility

foolip (approved):

PR #11052, opened 2021-06-16
Add Safari iOS versions for api.TextTrack.cuechange_event

vinyldarkscratch:
This PR adds real values for Safari iOS/iPadOS for the `cuechange_event` member of the `TextTrack` API by mirroring the data.


foolip (approved):
Confirmed the event is in WebKit's `EventNames.h`, let's figure out the versions later.

PR #11057, opened 2021-06-16
Remove RTCSessionDescriptionCallback

vinyldarkscratch:
This PR removes RTCSessionDescriptionCallback from BCD.  We shouldn't be recording callbacks like APIs in BCD.


vinyldarkscratch:
The content PR has been merged, so this is now ready to go!

foolip (approved):

PR #11063, opened 2021-06-16
Add missing urls for Trusted* (and toJSON() entries)

teoli2003:
Added the mdn_url and spec_url to `TrustedHTML`, `TrustedScript`, `TrustedScriptURL`, `TrustedTypePolicy`, and `TrustedTypePolicyFactor`(both on the interface and the children).

Added the `TrustedHTML.toJSON()`, `TrustedScript.toJSON()`, and `TrustedScriptURL.toJSON()` compat data. They are only in Canary, so I put a note there.

### DIFF
--- a/api/TrustedHTML.json
+++ b/api/TrustedHTML.json
@@ -2,6 +2,8 @@
   "api": {
     "TrustedHTML": {
       "__compat": {
+        "mdn_url": "https://developer.mozilla.org/docs/Web/API/TrustedHTML",
+        "spec_url": "https://w3c.github.io/webappsec-trusted-types/dist/spec/#trusted-html",
         "support": {
           "chrome": {
             "version_added": "83"
@@ -46,8 +48,59 @@
           "deprecated": false
         }
       },
+      "toJSON": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/TrustedHTML/toJSON",
+          "spec_url": "https://w3c.github.io/webappsec-trusted-types/dist/spec/#dom-trustedhtml-tojson",
+          "support": {
+            "chrome": {
+              "version_added": "83"
+            },
+            "chrome_android": {
+              "version_added": "83"
+            },
+            "edge": {
+              "version_added": "83"
+            },
+            "firefox": {
+              "version_added": false
+            },
+            "firefox_android": {
+              "version_added": false
+            },
+            "ie": {
+              "version_added": false
+            },
+            "opera": {
+              "version_added": "69"
+            },
+            "opera_android": {
+              "version_added": "59"
+            },
+            "safari": {
+              "version_added": false
+            },
+            "safari_ios": {
+              "version_added": false
+            },
+            "samsunginternet_android": {
+              "version_added": "13.0"
+            },
+            "webview_android": {
+              "version_added": "83"
+            }
+          },
+          "status": {
+            "experimental": false,
+            "standard_track": true,
+            "deprecated": false
+          }
+        }
+      },
       "toString": {
         "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/TrustedHTML/toString",
+          "spec_url": "https://w3c.github.io/webappsec-trusted-types/dist/spec/#trustedhtml-stringification-behavior",
           "support": {
             "chrome": {
               "version_added": "83"
--- a/api/TrustedHTML.json
+++ b/api/TrustedHTML.json
@@ -54,13 +54,14 @@
           "spec_url": "https://w3c.github.io/webappsec-trusted-types/dist/spec/#dom-trustedhtml-tojson",
           "support": {
             "chrome": {
-              "version_added": "83"
+              "version_added": false,
+              "notes": "Currently Chrome Canary only"
             },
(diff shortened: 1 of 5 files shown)


teoli2003:
For the record, these, including the `toJSON()` methods, have been added to MDN web docs in mdn/content#2978

api/TrustedHTML.json
```diff
@@ -46,8 +48,60 @@
           "deprecated": false
         }
       },
+      "toJSON": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/TrustedHTML/toJSON",
```
foolip, line 53: Wow, https://developer.mozilla.org/docs/Web/API/TrustedHTML/toJSON actually exists :)

Commented on something at https://github.com/mdn/content/pull/2978#pullrequestreview-686087283...

api/TrustedHTML.json
```diff
@@ -46,8 +48,60 @@
...
+          "spec_url": "https://w3c.github.io/webappsec-trusted-types/dist/spec/#dom-trustedhtml-tojson",
+          "support": {
+            "chrome": {
+              "version_added": false,
+              "notes": "Currently Chrome Canary only"
```
foolip, line 58: Very few things are enabled in Chrome Canary only in the same way that Firefox does, unless perhaps we count all of the "experimental web platform features" flag in this bucket.

Anyway, how did you determine this is in Canary only? Judging by https://mdn-bcd-collector.appspot.com/tests/api/TrustedHTML/toJSON it's already shipped to stable in Chrome 90, and that happened in this commit:
https://storage.googleapis.com/chromium-find-releases-static/0db.html#0dbce1408181f002902cba22fa275f658e8d9c18

foolip (changes requested):

138166234:
#3060

1381662341mohamadsalmanpor

در تاریخ پنجشنبه ۱۷ ژوئن ۲۰۲۱،‏ ۱۴:۱۳ Philip Jägenstedt <
***@***.***> نوشت:

api/TrustedHTML.json
teoli2003: We have a few toJSON and toString (stringifier) pages. Like for the iterable pages, they all have a close structure that make them fairly quick to create.

api/TrustedHTML.json
teoli2003: I trusted @RachelAndrew comment in  mdn/content#297, and it could have changed in the meantime!
Should we put 90?

foolip: Yeah, I'd use Chrome 90 and then `npm run mirror` to update the rest.

teoli2003: I tried to do it. My first run wiith `npm run mirror`, so I hope I did well. Fix in next commit

api/TrustedHTML.json
```diff
@@ -46,8 +48,59 @@
...
+          "spec_url": "https://w3c.github.io/webappsec-trusted-types/dist/spec/#dom-trustedhtml-tojson",
+          "support": {
+            "chrome": {
+              "version_added": "90"
+            },
+            "chrome_android": {
+              "version_added": "90"
+            },
+            "edge": {
+              "version_added": null
```
foolip, line 63:
```suggestion
              "version_added": "90"
```

Just mirroring Edge is typically safe for new APIs, but also confirmed by testing Edge 89 and 90 on https://mdn-bcd-collector.appspot.com/tests/api/TrustedHTML/toJSON.

api/TrustedScript.json
```diff
@@ -46,8 +48,59 @@
...
+          "spec_url": "https://w3c.github.io/webappsec-trusted-types/dist/spec/#dom-trustedscript-tojson",
+          "support": {
+            "chrome": {
+              "version_added": "90"
+            },
+            "chrome_android": {
+              "version_added": "90"
+            },
+            "edge": {
+              "version_added": false
```
foolip, line 63:
```suggestion
              "version_added": "90"
```

Confirmed using https://mdn-bcd-collector.appspot.com/tests/api/TrustedScript/toJSON

foolip (approved):
Missing some Edge data, but I'll apply my own suggestions and merge.

api/TrustedHTML.json
foolip: IMHO, for them to be useful, they would have to say which attributes are actually serialized in the resulting JSON, since that's often defined by prose in specs and isn't really guessable from looking at https://developer.mozilla.org/en-US/docs/Web/API/TrustedHTML, for example.

api/TrustedHTML.json
foolip: I didn't try running it myself, but `npm run mirror edge` should have mirrored the Edge data, but you have to run it once for each file you care about so it's very tedious. https://github.com/mdn/browser-compat-data/pull/10645 will make it less tedious.

teoli2003: I tried, but I think it mirrored it to false as it took IE as source. Maybe I did it wrong, though.

foolip: Ah right, https://github.com/mdn/browser-compat-data/pull/6582 hasn't been merged yet :)

PR #11064, opened 2021-06-16
Add missing spec_url & mdn_url to TextTrack*.json

teoli2003:
A few spec_url and mdn_url were missing.

### DIFF
--- a/api/TextTrackList.json
+++ b/api/TextTrackList.json
@@ -248,6 +248,8 @@
       },
       "onaddtrack": {
         "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/TextTrackList/onaddtrack",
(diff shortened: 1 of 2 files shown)


api/TextTrackList.json
```diff
@@ -342,6 +346,8 @@
       },
       "onremovetrack": {
         "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/TextTrackList/onchange",
```
foolip, line 349: This URL is wrong.

foolip (changes requested):

api/TextTrackList.json
teoli2003: oops. Fixed in next commit

Elchi3 (approved):

PR #11069, opened 2021-06-16
Mark HID entries as standards track

foolip:
Matches https://github.com/mdn/browser-compat-data/pull/11040.

### DIFF
--- a/api/HID.json
+++ b/api/HID.json
@@ -44,7 +44,7 @@
         },
         "status": {
           "experimental": true,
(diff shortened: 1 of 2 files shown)


foolip:
Per https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#choosing-an-experimental-status we should leave experimental as true.

Elchi3 (approved):
Thanks!

PR #11071, opened 2021-06-17
Remove unused SVGElement events

vinyldarkscratch:
This PR removes four events from the SVGElement API: abort, resize, scroll, and unload.  None of these events seem to be in the spec specific for SVGElement, and the only data we can find is for the global Window or for GlobalEventHandlers.


foolip (approved):

foolip:
https://github.com/mdn/browser-compat-data/pull/10985#discussion_r650825873 has a bit more background on why we're removing these.

vinyldarkscratch:
The content PR has been merged, so this is now ready to go!

PR #11076, opened 2021-06-17
Add Safari real values for Payment Request dictionaries

vinyldarkscratch:
This PR adds real values to Safari for the various Payment Request spec's dictionaries, based upon the source code at various Safari releases.  Additionally, this fixes the bad data from #11053, where I marked it as `false` even though there was actually support (turns out I was looking in the wrong place).  Data is as follows:

    - Support initially enabled by default in Safari 11.1 per CanIUse
    - https://caniuse.com/?search=payment

Safari 11.1:

    api.PaymentCurrencyAmount - https://trac.webkit.org/browser/webkit/tags/Safari-605.1.33.1.4/Source/WebCore/Modules/paymentrequest/PaymentCurrencyAmount.idl
    api.PaymentDetailsBase - https://trac.webkit.org/browser/webkit/tags/Safari-605.1.33.1.4/Source/WebCore/Modules/paymentrequest/PaymentDetailsBase.idl
    api.PaymentDetailsInit - https://trac.webkit.org/browser/webkit/tags/Safari-605.1.33.1.4/Source/WebCore/Modules/paymentrequest/PaymentDetailsInit.idl
    api.PaymentDetailsUpdate - https://trac.webkit.org/browser/webkit/tags/Safari-605.1.33.1.4/Source/WebCore/Modules/paymentrequest/PaymentDetailsUpdate.idl
    api.PaymentItem - https://trac.webkit.org/browser/webkit/tags/Safari-605.1.33.1.4/Source/WebCore/Modules/paymentrequest/PaymentItem.idl

Safari 12.1:

    api.PayerErrors - https://trac.webkit.org/browser/webkit/tags/Safari-607.1.40.3.1/Source/WebCore/Modules/paymentrequest/PayerErrorFields.idl
    api.PaymentDetailsUpdate.shippingAddressErrors - https://trac.webkit.org/browser/webkit/tags/Safari-607.1.40.3.1/Source/WebCore/Modules/paymentrequest/PaymentDetailsUpdate.idl

### DIFF
--- a/api/PaymentItem.json
+++ b/api/PaymentItem.json
@@ -70,10 +70,10 @@
             "version_added": false
           },
           "safari": {
-            "version_added": null
+            "version_added": "11.1"
           },
           "safari_ios": {
-            "version_added": null
+            "version_added": "11.3"
           },
           "samsunginternet_android": {
             "version_added": false
@@ -145,10 +145,10 @@
               "version_added": false
             },
             "safari": {
-              "version_added": null
+              "version_added": "11.1"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "11.3"
             },
             "samsunginternet_android": {
               "version_added": false
@@ -221,10 +221,10 @@
               "version_added": false
             },
             "safari": {
-              "version_added": null
+              "version_added": "11.1"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "11.3"
             },
             "samsunginternet_android": {
               "version_added": false
@@ -297,10 +297,10 @@
               "version_added": false
             },
             "safari": {
-              "version_added": null
+              "version_added": "11.1"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "11.3"
             },
             "samsunginternet_android": {
               "version_added": false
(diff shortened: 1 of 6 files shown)


foolip:
The caniuse.com entry is https://caniuse.com/payment-request. It's important to double check because the seemingly similar https://caniuse.com/mdn-api_paymentrequest is from BCD, and the URL https://caniuse.com/?search=payment doesn't reveal which kind of entry we're looking at.

api/PaymentItem.json
```diff
@@ -297,10 +297,10 @@
               "version_added": false
             },
             "safari": {
-              "version_added": null
+              "version_added": "11.1"
```
foolip, line 300: I randomly picked this single change to double check. It's been there since https://github.com/WebKit/WebKit/commit/386771f29a48831236a4427dc0854f8d2413e589 so this seems correct in that sense.

But I can't actually find any code in WebKit that *uses* this dictionary member. However, since there's OS integration for this feature, the code for that might not be inside of WebKit. So let's assume this is correct until we get bug reports about it actually not working.

foolip (approved):

PR #11078, opened 2021-06-17
Add Safari versions for api.Document.createElementNS.options

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `createElementNS.options` member of the `Document` API, based upon commit history and date.

Commit: https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/dom/Document.idl#L65

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -2282,10 +2282,10 @@
                 "notes": "For backwards compatibility, the <code>options</code> argument can be an object or a string with the custom element tag name, although the string version is deprecated."
               },
               "safari": {
-                "version_added": null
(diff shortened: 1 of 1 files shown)


foolip (approved):
Confirmed by https://github.com/WebKit/WebKit/blob/f71f9834b7c0d90d17e05c3479182a6122984fe9/Source/WebCore/dom/Document.idl#L65

foolip:
@vinyldarkscratch just a note that I converted the URL to something that will remain stable over time.

PR #11081, opened 2021-06-17
Invert meaning of URL.createObjectURL(mediaStream) entry

foolip:
This puts it on track for removal as irrelevant:
https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-features

The WebKit removal was in trunk version 604.1.16:
https://trac.webkit.org/changeset/214951/webkit
https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Configurations/Version.xcconfig?rev=214951

### DIFF
--- a/api/URL.json
+++ b/api/URL.json
@@ -224,56 +224,65 @@
             "deprecated": false
           }
         },
-        "no_MediaStream_argument": {
+        "MediaStream_support": {
           "__compat": {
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Nice! Thank you! 🎉

PR #11084, opened 2021-06-17
Add Safari versions for api.Element.attachShadow.delegatesFocus

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `attachShadow.delegatesFocus` member of the `Element` API, based upon commit history and date.

Commit: https://trac.webkit.org/browser/webkit/tags/Safari-609.1.20.111.8/Source/WebCore/dom/Element.idl#L154

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -1183,10 +1183,10 @@
                 "version_added": null
               },
               "safari": {
-                "version_added": null
+                "version_added": "13.1"
               },
               "safari_ios": {
-                "version_added": null
+                "version_added": "13.4"
               },
               "samsunginternet_android": {
                 "version_added": true


foolip (commented):
Can you also fix the entry name and description to match https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#parameters-and-parameter-object-features?

It would be `init_delegatesFocus_parameter` and `<code>init.delegatesFocus</code> parameter`.

https://developer.mozilla.org/en-US/docs/Web/API/Element/attachShadow#syntax would also need updating.

api/Element.json
```diff
@@ -1183,10 +1183,10 @@
                 "version_added": null
               },
               "safari": {
-                "version_added": null
+                "version_added": "13.1"
```
foolip, line 1186: I can confirm my method arrives at the same answer:
https://trac.webkit.org/changeset/251043/webkit
https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Configurations/Version.xcconfig?rev=251043

foolip (approved):

foolip:
Actually, let's merge this and deal with the other change separately.

PR #11085, opened 2021-06-17
Add Safari versions for api.Element.requestFullscreen.options_parameter

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `requestFullscreen.options_parameter` member of the `Element` API, based upon commit history and date.

Commit: https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/dom/Element%2BFullscreen.idl#L33

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -6548,10 +6548,10 @@
                 "version_added": null
               },
               "safari": {
-                "version_added": null
+                "version_added": false
               },
               "safari_ios": {
-                "version_added": null
+                "version_added": false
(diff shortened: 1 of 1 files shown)


foolip (approved):

foolip:
Just a nit that https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/dom/Element%2BFullscreen.idl#L33 is not a commit, but I didn't bother updating the commit message.

PR #11086, opened 2021-06-17
Add Safari versions for api.HTMLCanvasElement.getContext.failIfMajorPerformanceCaveat

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `getContext.failIfMajorPerformanceCaveat` member of the `HTMLCanvasElement` API, based upon commit history and date.

Commit: https://trac.webkit.org/changeset/204618/webkit / https://trac.webkit.org/browser/webkit/tags/Safari-603.2.1/Source/WebCore/html/canvas/WebGLContextAttributes.idl

### DIFF
--- a/api/HTMLCanvasElement.json
+++ b/api/HTMLCanvasElement.json
@@ -284,10 +284,10 @@
                 "version_added": null
               },
               "safari": {
-                "version_added": null
+                "version_added": "10.1"
               },
               "safari_ios": {
-                "version_added": null
+                "version_added": "10.3"
               },
               "samsunginternet_android": {
                 "version_added": false


foolip:
@vinyldarkscratch a lot wrong with the existing structure here. Do you wanna update this all to conform with https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#parameters-and-parameter-object-features more? I'm pretty sure that will reveal even more that's wrong here.

api/HTMLCanvasElement.json
```diff
@@ -284,10 +284,10 @@
                 "version_added": null
               },
               "safari": {
-                "version_added": null
+                "version_added": "10.1"
```
foolip, line 287: https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Configurations/Version.xcconfig?rev=204618 agrees 👍

foolip (approved):

PR #11095, opened 2021-06-17
Add Safari versions for FontFace API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `FontFace` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.4).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/FontFace

### DIFF
--- a/api/FontFace.json
+++ b/api/FontFace.json
@@ -926,10 +926,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": false
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": false
             },
             "samsunginternet_android": {
               "version_added": "10.0"


foolip (approved):
This seems right, but it's coming: https://github.com/WebKit/WebKit/commit/792fde69d188dc5b10a7a09372bee06931dd9bac

foolip:
@sideshowbarker this is a case where `implementation_url` would have helped ensure that this laters gets updated, if you want more evidence that the idea is useful.

sideshowbarker:
> @sideshowbarker this is a case where `implementation_url` would have helped ensure that this laters gets updated, if you want more evidence that the idea is useful.

Cool, yes, having evidence like this gives me motivation to get back to finishing up the PR for that.

PR #11098, opened 2021-06-17
Add Safari versions for PerformanceEntry API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `PerformanceEntry` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.4).

### DIFF
--- a/api/PerformanceEntry.json
+++ b/api/PerformanceEntry.json
@@ -380,10 +380,10 @@
               "version_added": "46"
             },
             "safari": {
-              "version_added": null
+              "version_added": "11.1"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "11.3"
             },
             "samsunginternet_android": {
               "version_added": "8.0"
--- a/api/PerformanceEntry.json
+++ b/api/PerformanceEntry.json
@@ -380,10 +380,10 @@
               "version_added": "46"
             },
             "safari": {
-              "version_added": "11.1"
+              "version_added": "11"
             },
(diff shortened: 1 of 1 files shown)


foolip:
https://mdn-bcd-collector.appspot.com/tests/api/PerformanceEntry is actually a broken test since it tries to use `window.performance` in a worker, did you really use that test?

https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Configurations/Version.xcconfig?rev=211594 suggests this should be Safari 11, confirming or disconfirming with testing would be great.

Aside: If you were using BrowserStack to test, I'd recommend using Sauce for all Safari testing, since they make it relatively easy to test both x.0 and x.1 releases.

foolip:
I've sent https://github.com/foolip/mdn-bcd-collector/pull/1238 to fix the tests.

vinyldarkscratch:
I was basing it off of the `toJSON` property, and based it upon whether `PerformanceEntry` was exposed in the worker.  That part hasn't changed since the collector update, so I'm assuming I made a typo on my notepad.  Fixed the data!

foolip (approved):

foolip:
> I was basing it off of the `toJSON` property, and based it upon whether `PerformanceEntry` was exposed in the worker

I did spot that single test being green, that explains it then. Still it's better to have the fixed tests, `toJSON` could have been added after worker exposure.

PR #11099, opened 2021-06-17
Add Safari versions for PerformanceObserver API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `PerformanceObserver` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.4).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/PerformanceObserver

### DIFF
--- a/api/PerformanceObserver.json
+++ b/api/PerformanceObserver.json
@@ -351,10 +351,10 @@
               "version_added": "46"
             },
             "safari": {
-              "version_added": null
+              "version_added": "11.1"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "11.3"
             },
             "samsunginternet_android": {
               "version_added": "8.0"
--- a/api/PerformanceObserver.json
+++ b/api/PerformanceObserver.json
@@ -351,10 +351,10 @@
               "version_added": "46"
             },
             "safari": {
-              "version_added": "11.1"
+              "version_added": "11"
             },
             "safari_ios": {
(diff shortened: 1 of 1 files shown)


api/PerformanceObserver.json
```diff
@@ -351,10 +351,10 @@
               "version_added": "46"
             },
             "safari": {
-              "version_added": null
+              "version_added": "11.1"
```
foolip, line 354:
```suggestion
              "version_added": "11"
```

api/PerformanceObserver.json
```diff
@@ -351,10 +351,10 @@
...
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "11.3"
```
foolip, line 357:
```suggestion
              "version_added": "11"
```

foolip (changes requested):
Most of https://developer.mozilla.org/en-US/docs/Web/API/PerformanceObserver#browser_compatibility says 11 so I found this a bit implausible.

Tried to confirm with https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Configurations/Version.xcconfig?rev=211594, but that points to 11.

I've tested https://mdn-bcd-collector.appspot.com/tests/api/PerformanceObserver in 11 on Sauce, it's supported.

foolip (approved):

PR #11100, opened 2021-06-17
Remove non-standard api.Event.getPreventDefault

vinyldarkscratch:
This PR removes the non-standard `getPreventDefault` from the Event API.  This appears to be a Firefox-only feature which has been removed over two years ago.

### DIFF
--- a/api/Event.json
+++ b/api/Event.json
@@ -653,61 +653,6 @@
           }
         }
       },
-      "getPreventDefault": {
-        "__compat": {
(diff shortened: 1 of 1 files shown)


vinyldarkscratch:
Apparently the documentation for `getPreventDefault` has already been removed!

ddbeck:
Similar issue with content here as in https://github.com/mdn/browser-compat-data/pull/11101#issuecomment-872386312

vinyldarkscratch:
@sideshowbarker merged the content PR for this, so this is now ready to go!

PR #11101, opened 2021-06-17
Remove non-standard api.Event.preventBubble/preventCapture

vinyldarkscratch:
This PR removes the non-standard `preventBubble` and `preventCapture` members of the Event API.  These features seem to have only been implemented in Firefox, and have been removed over two years ago.


vinyldarkscratch:
Looks like the documentation for these two methods have already been removed from MDN!

ddbeck:
Not quite, this page stills need scrubbing: https://developer.mozilla.org/en-US/docs/Web/API/Event#deprecated_methods

(And I haven't checked for any lingering cross references from elsewhere)

vinyldarkscratch:
@sideshowbarker merged the content PR for this, so this is now ready to go!

PR #11108, opened 2021-06-17
Add Safari versions for SVGZoomAndPan API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `SVGZoomAndPan` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.4).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/SVGZoomAndPan

### DIFF
--- a/api/SVGZoomAndPan.json
+++ b/api/SVGZoomAndPan.json
@@ -31,10 +31,10 @@
             "version_added": "≤12.1"
           },
           "safari": {
-            "version_added": null
+            "version_added": false
           },
           "safari_ios": {
-            "version_added": null
+            "version_added": false
           },
           "samsunginternet_android": {
             "version_added": true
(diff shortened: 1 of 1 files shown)


foolip:
I don't think we can trust the test for this, I'm pretty sure this is in WebKit. Can you search the source?

vinyldarkscratch:
Argh, this is a mixin...

vinyldarkscratch:
Tracked this down and found that it is indeed around in Safari 3, see https://trac.webkit.org/browser/webkit/tags/old/Safari-522.11.5b/WebCore/ksvg2/svg/SVGZoomAndPan.idl.  It had later become an official mixin, but it probably just wasn't exposed before.  Fixed!

foolip (approved):
We should probably find a way to just remove this entry, but these versions are correct if we interpret it as a mixin. It's gone from the spec though, so it isn't really anything now :)

PR #11109, opened 2021-06-17
Remove redundant viewTarget member of SVGViewElement API

vinyldarkscratch:
This PR removes the `viewTarget` member of the SVGViewElement API.  This member doesn't have any MDN doc page, is non-standard, and has been removed from browsers for over two years.

### DIFF
--- a/api/SVGViewElement.json
+++ b/api/SVGViewElement.json
@@ -47,62 +47,6 @@
           "standard_track": true,
           "deprecated": false
         }
-      },
-      "viewTarget": {
-        "__compat": {
-          "support": {
-            "chrome": {
-              "version_added": true,
-              "version_removed": "56"
-            },
-            "chrome_android": {
-              "version_added": true,
-              "version_removed": "56"
-            },
-            "edge": {
-              "version_added": "12",
-              "version_removed": "79"
-            },
-            "firefox": {
-              "version_added": true,
-              "version_removed": "61"
-            },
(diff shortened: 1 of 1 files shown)


api/SVGViewElement.json
```diff
@@ -47,62 +47,6 @@
...
-              "version_added": true,
-              "version_removed": "56"
-            },
-            "chrome_android": {
-              "version_added": true,
-              "version_removed": "56"
-            },
-            "edge": {
-              "version_added": "12",
-              "version_removed": "79"
```
foolip, line 64: This is unfortunately less than two years ago.

ddbeck: @vinyldarkscratch are you going to repurpose this PR to some other end or should we close it?

vinyldarkscratch: Right, legacy Edge still has a half a year to go...  Unfortunately I don't see a way to repurpose this PR, so unless we just want to skip ahead and remove it early, we'll have to reopen this PR another time.

foolip: Or actually, maybe a single hardcoded date would do it, then we could move it forward as much as the data allows at any time.

vinyldarkscratch: I was thinking about adding a linter, and I think I still have the code around.  The linter is based upon the current date, which can of course cause issues.  A hard-coded date makes much more sense!

PR #11110, opened 2021-06-17
api/SourceBuffer/safari

vinyldarkscratch:
- Remove redundant viewTarget member of SVGViewElement API
- Add Safari versions for SourceBuffer API


foolip:
The PR title here is the branch name I think :)

PR #11111, opened 2021-06-17
Add Safari versions for Document API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `Document` API.  The data was derived from their handlers (AKA `on*`).

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -8790,10 +8790,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": "10.1"
+              "version_added": null
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": null
             },
             "samsunginternet_android": [
               {
@@ -8868,10 +8868,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": "10.1"
+              "version_added": null
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": null
             },
             "samsunginternet_android": [
               {
--- a/api/Document.json
+++ b/api/Document.json
@@ -5425,10 +5425,12 @@
               "version_added": "32"
             },
             "safari": {
-              "version_added": null
+              "version_added": "5.1",
+              "alternative_name": "webkitfullscreenchange"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "5.1",
+              "alternative_name": "webkitfullscreenchange"
             },
             "samsunginternet_android": {
               "version_added": "5.0"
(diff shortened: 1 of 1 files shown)


api/Document.json
```diff
@@ -5425,10 +5425,12 @@
               "version_added": "32"
             },
             "safari": {
-              "version_added": null
+              "version_added": "5.1",
```
foolip, line 5428: This doesn't match https://developer.mozilla.org/en-US/docs/Web/API/Element/requestFullScreen#browser_compatibility, the event can't have been fired before the API was supported.

api/Document.json
```diff
@@ -8786,10 +8790,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": "10.1"
```
foolip, line 8793: https://developer.mozilla.org/en-US/docs/Web/API/Pointer_Lock_API#browser_compatibility says 10.1 for the unprefixed API, but also that there was a webkit-prefixed one. Is that right, and did that fire a prefixed event that should be added here?

api/Document.json
```diff
@@ -8864,10 +8868,10 @@
...
```
foolip, line 8871: Ditto.

foolip (changes requested):

api/Document.json
vinyldarkscratch: It appears that `api.Element.requestFullScreen` is incorrect, then.  I just launched Safari 5.1 on BrowserStack and ran `document.forms[0].webkitRequestFullScreen`, and the result came up as a function.  I'll submit a PR to correct it.

api/Document.json
vinyldarkscratch: I took a closer look into this one, and found that the data we have for `api.Element.requestPointerLock` is inaccurate.  There was no prefixed version for the function before Safari 10.1, or even in 10.1 itself.  Tested in SauceLabs in Safari 10.0, 10.1, and 9.0.

I'll write a PR to fix up its data.

api/Document.json
foolip: Thanks! That's https://github.com/mdn/browser-compat-data/pull/11162, would you mind folding those changes into that PR for easier review as a group?

api/Document.json
foolip: Would you mind folding these changes into https://github.com/mdn/browser-compat-data/pull/11164 for easier review?

api/Document.json
vinyldarkscratch: I've merged the changes from this PR into that one!  This PR is now no longer needed.

PR #11118, opened 2021-06-17
Add Firefox versions for MediaSettingsRange API

vinyldarkscratch:
This PR adds real values for Firefox and Firefox Android for the `MediaSettingsRange` API, based upon information in a tracking bug.

Tracking Bug: https://bugzil.la/1655715

### DIFF
--- a/api/MediaSettingsRange.json
+++ b/api/MediaSettingsRange.json
@@ -15,10 +15,10 @@
             "version_added": "≤79"
           },
           "firefox": {
-            "version_added": null
+            "version_added": "81"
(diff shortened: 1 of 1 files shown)


foolip:
https://bugzilla.mozilla.org/show_bug.cgi?id=1655715 isn't a tracking bug for the feature, it's just an automatically filed bug for syncing tests with https://github.com/web-platform-tests/wpt.

Searching Gecko source shows this isn't supported, so it should be all false.

foolip (approved):

PR #11119, opened 2021-06-17
Add Firefox versions for RTCConfiguration API

vinyldarkscratch:
This PR adds real values for Firefox and Firefox Android for the `RTCConfiguration` API, based upon information in a tracking bug.

Tracking Bug: https://bugzil.la/835712, https://bugzil.la/1112692, https://bugzil.la/1172785, https://bugzil.la/1187775, https://bugzil.la/942367

### DIFF
--- a/api/RTCConfiguration.json
+++ b/api/RTCConfiguration.json
@@ -15,10 +15,10 @@
             "version_added": "≤79"
           },
           "firefox": {
-            "version_added": null
+            "version_added": "21"
           },
           "firefox_android": {
-            "version_added": null
+            "version_added": "21"
           },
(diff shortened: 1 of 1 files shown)


foolip (approved):

foolip:
Hmm, I didn't review this too carefully. This data is not compatible with https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection#browser_compatibility which claims `RTCPeerConnection` was shipped in Firefox 22.

foolip:
I've filed https://github.com/mdn/browser-compat-data/issues/11136.

PR #11124, opened 2021-06-17
Add Firefox versions for ServiceWorkerGlobalScope API

vinyldarkscratch:
This PR adds real values for Firefox and Firefox Android for the `ServiceWorkerGlobalScope` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.4).

### DIFF
--- a/api/ServiceWorkerGlobalScope.json
+++ b/api/ServiceWorkerGlobalScope.json
@@ -414,10 +414,10 @@
               "version_added": "≤79"
             },
             "firefox": {
-              "version_added": null
+              "version_added": false
             },
             "firefox_android": {
-              "version_added": null
+              "version_added": false
             },
             "ie": {
               "version_added": false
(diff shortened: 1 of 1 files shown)


api/ServiceWorkerGlobalScope.json
```diff
@@ -997,10 +997,10 @@
               "version_added": "81"
             },
             "firefox": {
-              "version_added": null
+              "version_added": "65"
```
foolip, line 1000: This is suspect, https://developer.mozilla.org/en-US/docs/Web/API/Window/messageerror_event#browser_compatibility is much earlier. Can you investigate?

vinyldarkscratch: I did a quick search and found the corresponding bug for this, which confirms Firefox 65: https://bugzilla.mozilla.org/show_bug.cgi?id=1399446

foolip: I see, so it really was implemented separately and later. I checked https://hg.mozilla.org/mozilla-central/rev/988f40d0fd2c and it also includes actually firing the event, it wasn't just the `onmessageerror` property being added.

foolip (approved):

PR #11126, opened 2021-06-17
Add Firefox versions for RTCIdentityProviderGlobalScope API

vinyldarkscratch:
This PR adds real values for Firefox and Firefox Android for the `RTCIdentityProviderGlobalScope` API.  There is nothing in Firefox's IDL for this interface.


foolip (approved):
We should remove this entirely.

PR #11127, opened 2021-06-17
Add Firefox versions for RTCIdentityProviderRegistrar API

vinyldarkscratch:
This PR adds real values for Firefox and Firefox Android for the `RTCIdentityProviderRegistrar` API, based upon information in a tracking bug.

Tracking Bug: https://bugzil.la/975144

### DIFF
--- a/api/RTCIdentityProviderRegistrar.json
+++ b/api/RTCIdentityProviderRegistrar.json
@@ -14,10 +14,10 @@
             "version_added": false
           },
           "firefox": {
-            "version_added": "38"
+            "version_added": false
           },
           "firefox_android": {
-            "version_added": "38"
+            "version_added": false
           },
           "ie": {
             "version_added": false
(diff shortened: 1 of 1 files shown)


foolip:
It's here in Gecko: https://github.com/mozilla/gecko-dev/blob/f9c007f93321f02889f7495048c6b364ae71997c/dom/webidl/RTCIdentityProvider.webidl#L9-L11

But it's *only* there, so how could these objects become web exposed?

In the spec it's used in https://w3c.github.io/webrtc-identity/identity.html#dom-rtcidentityproviderglobalscope-rtcidentityprovider, so probably this should be removed entirely since `RTCIdentityProviderGlobalScope` itself isn't supported.

vinyldarkscratch:
Yeah, this seems odd to me, probably leftover code from a previous attempt to implement it, but it was abandoned for one reason or another.  (That's my guess, at least.)  I'll set this to `false`!

foolip (approved):

PR #11130, opened 2021-06-17
Link WebKit bug for all messageerror event entries

foolip:


vinyldarkscratch (approved):
Thanks for finding and linking the bug, this is looking good to me!

PR #11138, opened 2021-06-18
Update Safari versions for Media Extensions APIs

vinyldarkscratch:
This PR adds real values for Safari iOS/iPadOS for the Media Extensions APIs, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.4).  These APIs were not supported before iOS 13, and a lot was set to `false` when it wasn't supposed to be.

Note: only the `MediaSource` API was actually disabled on iPhones, however to get a SourceBuffer object in the first place, you need the `MediaSource` API.  I've simply marked everything as non-exposed on iPhones since they can't be used.

Tests Used:
https://mdn-bcd-collector.appspot.com/tests/api/MediaSource
https://mdn-bcd-collector.appspot.com/tests/api/SourceBuffer
https://mdn-bcd-collector.appspot.com/tests/api/SourceBufferList
https://mdn-bcd-collector.appspot.com/tests/api/AudioTrack/sourceBuffer
https://mdn-bcd-collector.appspot.com/tests/api/TextTrack/sourceBuffer
https://mdn-bcd-collector.appspot.com/tests/api/VideoTrack/sourceBuffer

### DIFF
--- a/api/AudioTrack.json
+++ b/api/AudioTrack.json
@@ -697,7 +697,9 @@
               "version_added": "6.1"
             },
             "safari_ios": {
-              "version_added": "7"
+              "version_added": "13",
+              "partial_implementation": true,
+              "notes": "Exposed in Mobile Safari on iPad but not on iPhone."
             },
             "samsunginternet_android": {
               "version_added": false
--- a/api/SourceBuffer.json
+++ b/api/SourceBuffer.json
@@ -120,7 +120,9 @@
               "version_added": "8"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "13",
+              "partial_implementation": true,
+              "notes": "Exposed in Mobile Safari on iPad but not on iPhone."
             },
             "samsunginternet_android": {
               "version_added": "2.0"
@@ -170,7 +172,9 @@
               "version_added": "8"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "13",
+              "partial_implementation": true,
+              "notes": "Exposed in Mobile Safari on iPad but not on iPhone."
             },
             "samsunginternet_android": {
               "version_added": "3.0"
@@ -325,7 +329,9 @@
               "version_added": "8"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "13",
+              "partial_implementation": true,
+              "notes": "Exposed in Mobile Safari on iPad but not on iPhone."
             },
             "samsunginternet_android": {
               "version_added": "3.0"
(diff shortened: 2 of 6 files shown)


api/SourceBuffer.json
```diff
@@ -563,10 +563,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": "12.1"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "13"
```
foolip, line 569: The other entries have comments saying "Exposed in Mobile Safari on iPad but not on iPhone."

Testing https://mdn-bcd-collector.appspot.com/tests/api/SourceBuffer on an iPad shows a lot more of the API supported, would you mind updating more of it? Or file an issue and leave this false, it's just one of many incorrectly false entries here.

foolip (changes requested):

api/SourceBuffer.json
vinyldarkscratch: Huh, interestingly enough, this API seems to be supported on iPhone too, so I don't know why that note was added.  I'll update this PR accordingly.

foolip: Does it actually work on the iPhone, or is the interface just accidentally exposed but useless? https://caniuse.com/mediasource also talks about it being supported only on iPad, so a working demo of it on iPhone would be required to convince me that's wrong :)

vinyldarkscratch: Alright, so I did some more digging around and found that an API this depends on, `MediaSource`, seems to be supported only on iPad.  A `MediaSource` object is required to get a `SourceBuffer` object (see `api.MediaSource.addSourceBuffer`), and the entire API is marked as unsupported on Safari iOS.

vinyldarkscratch: Updates have been made!

api/AudioTrack.json
```diff
@@ -697,7 +697,9 @@
               "version_added": "6.1"
             },
             "safari_ios": {
-              "version_added": "7"
```
foolip, line 700: Are you sure about this? `AudioTrack` is separate from MSE and I believe shipped much earlier, I suspect the existing data is correct.

api/TextTrack.json
```diff
@@ -663,7 +663,9 @@
               "version_added": "8"
             },
             "safari_ios": {
-              "version_added": "13"
```
foolip, line 666: This and `VideoTrack` also aren't part of MSE.

foolip (changes requested):

api/AudioTrack.json
vinyldarkscratch: This particular property is a part of the spec for MSE.  I've confirmed that there's no support before Safari iOS 13 in BrowserStack.

foolip: Oh, this is the `sourceBuffer` property! I didn’t expand the diff context 🙄

foolip (approved):

PR #11139, opened 2021-06-18
Add Safari versions for api.WebSocket.worker_support

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `worker_support` member of the `WebSocket` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.4).

### DIFF
--- a/api/WebSocket.json
+++ b/api/WebSocket.json
@@ -1047,10 +1047,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": "5"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "5.1"
             },
             "samsunginternet_android": {
               "version_added": true


api/WebSocket.json
```diff
@@ -1047,10 +1047,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": "5"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "5.1"
```
foolip, line 1053: Is this based on testing both iOS 5 and 5.1? After trying to fill in https://gist.github.com/foolip/6872cf5350dccf6224fce984fb73fba1 I have quite low confidence in, well, most Safari versions between 3 and 9.

vinyldarkscratch: This is not, rather it's based upon mirroring, and when worker support was first added.

foolip: I see. Mirroring would give 4.2, but workers were supported later. This is consistent with https://developer.mozilla.org/en-US/docs/Web/API/Worker#browser_compatibility and https://caniuse.com/webworkers

foolip (approved):

PR #11140, opened 2021-06-18
Add WheelEvent.wheelDelta, wheelDeltaX, wheelDeltaY

hamishwillee:
https://bugzilla.mozilla.org/show_bug.cgi?id=1708829 added (deprecated) properties to Wheelevent: wheelDelta, wheelDeltaX, wheelDeltaY

There is some confusion here as I have outlined in https://github.com/mdn/content/pull/5966#issuecomment-861887078

I know that the bug linked above added these properties to [WheelEvent](https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent). However the properties are already shown as present in BCD and MDN under [MouseWheelEvent](https://developer.mozilla.org/en-US/docs/Web/API/MouseWheelEvent). @ddbeck indicates these are documented in the wrong place.

What I have done is copied the `MouseWheelEvent` properties across and then added support for FF90. I had to mark Safari as `false` for properties to match the `WheelEvent`. 

@ddbeck 
- Do we remove the BCD for MouseWheelEvent? 
- What about for the MDN docs - just delete all the content except for the warnings not to use and have a note in the properties "This inherits properties from WheelEvent"? 
- As per https://github.com/mdn/content/pull/5966#issuecomment-861887078 I have not noted the compatibility change in the WheelEvent that caused this bug change. There has been a change in behaviour, but it was a change between two allowed  "compatible" outputs.

### DIFF
--- a/api/WheelEvent.json
+++ b/api/WheelEvent.json
@@ -349,10 +349,10 @@
         "__compat": {
           "support": {
             "chrome": {
-              "version_added": "43"
+              "version_added": true
             },
             "chrome_android": {
-              "version_added": "43"
+              "version_added": true
             },
             "edge": {
               "version_added": "12"
@@ -367,22 +367,22 @@
               "version_added": false
             },
             "opera": {
-              "version_added": "30"
+              "version_added": true
             },
             "opera_android": {
-              "version_added": "30"
+              "version_added": true
             },
             "safari": {
-              "version_added": "7"
+              "version_added": true
             },
             "safari_ios": {
               "version_added": false
             },
             "samsunginternet_android": {
-              "version_added": "4.0"
+              "version_added": true
             },
             "webview_android": {
-              "version_added": "43"
+              "version_added": true
             }
           },
           "status": {
@@ -396,10 +396,10 @@
         "__compat": {
           "support": {
             "chrome": {
-              "version_added": "43"
+              "version_added": true
             },
             "chrome_android": {
-              "version_added": "43"
+              "version_added": true
             },
             "edge": {
               "version_added": "12"
@@ -414,22 +414,22 @@
               "version_added": false
             },
             "opera": {
-              "version_added": "30"
+              "version_added": true
             },
             "opera_android": {
-              "version_added": "30"
+              "version_added": true
             },
             "safari": {
-              "version_added": "7"
+              "version_added": true
             },
             "safari_ios": {
               "version_added": false
             },
             "samsunginternet_android": {
-              "version_added": "4.0"
+              "version_added": true
             },
             "webview_android": {
-              "version_added": "43"
+              "version_added": true
             }
           },
           "status": {
@@ -443,10 +443,10 @@
         "__compat": {
           "support": {
             "chrome": {
-              "version_added": "43"
+              "version_added": true
             },
             "chrome_android": {
-              "version_added": "43"
+              "version_added": true
             },
             "edge": {
               "version_added": "12"
@@ -461,22 +461,22 @@
               "version_added": false
             },
             "opera": {
-              "version_added": "30"
+              "version_added": true
             },
             "opera_android": {
-              "version_added": "30"
+              "version_added": true
             },
             "safari": {
-              "version_added": "7"
+              "version_added": true
             },
             "safari_ios": {
               "version_added": false
             },
             "samsunginternet_android": {
-              "version_added": "4.0"
+              "version_added": true
             },
             "webview_android": {
-              "version_added": "43"
+              "version_added": true
             }
           },
           "status": {
--- a/api/WheelEvent.json
+++ b/api/WheelEvent.json
@@ -344,6 +344,151 @@
             "deprecated": false
           }
         }
+      },
+      "wheelDelta": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/WheelEvent",
+          "support": {
+            "chrome": {
+              "version_added": false
+            },
+            "chrome_android": {
+              "version_added": false
+            },
+            "edge": {
+              "version_added": "12",
+              "version_removed": "79"
+            },
+            "firefox": {
+              "version_added": "90"
+            },
+            "firefox_android": {
+              "version_added": "90"
+            },
+            "ie": {
+              "version_added": "9"
+            },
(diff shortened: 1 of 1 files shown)


ddbeck:
Commentary and answers to questions:

> There is some confusion here as I have outlined in [mdn/content#5966 (comment)](https://github.com/mdn/content/pull/5966#issuecomment-861887078)

No, it was just _my_ confusion. I think the content is fine, actually.

> I know that the bug linked above added these properties to [WheelEvent](https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent). However the properties are already shown as present in BCD and MDN under [MouseWheelEvent](https://developer.mozilla.org/en-US/docs/Web/API/MouseWheelEvent). @ddbeck indicates these are documented in the wrong place.

I was wrong about this. It's fine how things are.
 
> * Do we remove the BCD for MouseWheelEvent?

No. Well, _maybe_. I think it's been removed from all the browsers (it's not exposed in Firefox, Chrome, or Safari), but it depends on how recently. Needs testing to figure that part out.

> * What about for the MDN docs - just delete all the content except for the warnings not to use and have a note in the properties "This inherits properties from WheelEvent"?

I don't know that it does actually. I'd leave it be, especially if complete removal is a likely outcome.

> * As per [mdn/content#5966 (comment)](https://github.com/mdn/content/pull/5966#issuecomment-861887078) I have not noted the compatibility change in the WheelEvent that caused this bug change. There has been a change in behaviour, but it was a change between two allowed  "compatible" outputs.

Yes, this makes sense. 👍

ddbeck (changes requested):
A few broad changes to suggest here:

- Don't link to parent features. Either link to a non-existent page for the feature specifically, or don't link at all. In this case, since there's probably not a lot of interest in the non-standard properties, don't bother with `mdn_url`s.
- The data here is pretty suspect. I did some brief testing:
  - Chrome (etc.) supports these from version 43 (or equivalent)
  - Edge 12+
  - Safari on macOS: 7 (didn't figure it out for iOS, but mirroring probably works here)
  - IE: no support (at least, not on this interface)

foolip:
> Chrome (etc.) supports these from version 43

See https://github.com/mdn/browser-compat-data/issues/7843 for why you should always doubt the results when testing of attributes seems to point to Chrome 43.

Of course Chrome 43 included some new attributes, but care is needed to determine that.

Something similar happened in some version of Safari too.

api/WheelEvent.json
```diff
@@ -344,6 +344,151 @@
...
+            },
+            "chrome_android": {
+              "version_added": false
+            },
+            "edge": {
+              "version_added": "12",
+              "version_removed": "79"
+            },
+            "firefox": {
+              "version_added": "90"
```
foolip, line 363: The reason this is being added to Firefox is that it's some other engine and there was a web compat problem, yet all the other data is false. I'm suspicious 🤨

hamishwillee: @foolip You'd be right to be suspicious :-). Everything about this starting from the original bug fix feels a bit dodgy. 

As I said, I copied the `MouseWheelEvent` properties across and then added support for FF90. My assumptions here are:
- BCD should match mousewheel 
- We don't expect or want anyone to use these on any browser. So given they are deprecated, marking them as false seems like not such a "bad thing" to be wrong on.

In other words, this is a hack, but the effort of proving the versions seems like a waste of time.

What would you like me to do?

hamishwillee: @ddbeck (& @foolip )

1. Removed the `mdn_url`s. Not going to create pages for these deprecated properties, so no point linking to something that does not exist.
2. I have matched these to version you suggested (based on chrome version 43. Did not mirror support for Safari on iOS as top level says "false".

Now I appreciate that there are likely issues with v43 matching, but at least we know that users will not expect earlier versions, and since we don't want people to use these anyway, is that good enough?

If not, then what versions should we use? You might reasonably expect that since these are legacy properties they would match the original introduction of the API. So Chrome 31, Webview after 37, Safari after 6.1.

foolip: > I copied the `MouseWheelEvent` properties across

This does not seems like the right starting point to me. The `WheelEvent` and `MouseWheelEvent` events correspond to the "wheel" and "mousewheel" events, and which of those is supported has been a long-standing compat issue.

Confirming the versions with https://staging-dot-mdn-bcd-collector.appspot.com/tests/api/WheelEvent should be easy once https://github.com/foolip/mdn-bcd-collector/pull/1247 is merged.

foolip: OK, so with http://staging-dot-mdn-bcd-collector.appspot.com/tests/api/WheelEvent I can confirm that support of `WheelEvent` itself and these added properties goes back to at least Chrome 16, the earliest available on BrowserStack. It's probably older than that, maybe Chrome 1.

I would suggest setting everything except Edge, IE and Firefox to true, so that I and @vinyldarkscratch will come back to it as part of https://github.com/mdn/browser-compat-data/issues/6369. This will require testing on older versions of Chrome and Safari than I have.

api/WheelEvent.json
```diff
@@ -344,6 +344,147 @@
...
+            "samsunginternet_android": {
+              "version_added": "4.0"
+            },
+            "webview_android": {
+              "version_added": "43"
+            }
+          },
+          "status": {
+            "experimental": false,
+            "standard_track": false,
```
foolip, line 390: Unfortunately true, but see https://github.com/w3c/uievents/issues/305 about standardizing it.

api/WheelEvent.json
hamishwillee: @foolip Thanks for the advice. Done! Ready for re-review.

ddbeck (approved):
Thank you! 🎉

(And thank you for catching that detail with Chrome 43, @foolip.)

ddbeck:
@vinyldarkscratch I didn't write a release note for this one, but let me know if I've missed something here that made you think one was required.

vinyldarkscratch:
Based upon the guidelines now added in #11195, we wouldn't need a release note for this.  I feel that if we're adding new entries, however, that it should be considered something worthwhile to mention?

ddbeck:
@vinyldarkscratch Want to propose that as a change to the new label documentation? 😃

PR #11144, opened 2021-06-18
HTTP DPR header is secure context

hamishwillee:
As part of [docs work](https://github.com/mdn/content/pull/6097) on [DPR HTTP header](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/DPR) I noted that this has secure context header. Did further research and notes seem to indicate all client hints should require secure context.

The problem is I can't find a current spec for DPR to verify this. But it does appear to be well understood - e.g. see 
 comment in https://developers.google.com/web/fundamentals/performance/optimizing-content-efficiency/client-hints

Anyway, this adds the secure_context_required to DPR. It is by way of "testing the water". If you approve this case I'll add it to the other ones :-)

### DIFF
--- a/http/headers/dpr.json
+++ b/http/headers/dpr.json
@@ -47,6 +47,54 @@
             "standard_track": false,
             "deprecated": true
           }
+        },
+        "secure_context_required": {
+          "__compat": {
+            "description": "Secure context required",
+            "support": {
+              "chrome": {
+                "version_added": "46"
+              },
+              "chrome_android": {
+                "version_added": "46"
+              },
+              "edge": {
+                "version_added": "≤79"
+              },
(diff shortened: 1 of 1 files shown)


ddbeck:
> The problem is I can't find a current spec for DPR to verify this.

If this is only going to be mirrored from the parent feature, then I'm not inclined to add this data. It's sort of synthetic and, unless there's a difference between the secure context requirement and the feature itself, it doesn't really add anything.

Now, if you want to test this and find out that there's discrepancies…

hamishwillee:
@ddbeck One day you're going to get distracted, and I'll be able to add data rather than just "(in)compatibility data. Until then. Closing, because you're right :-).

PR #11145, opened 2021-06-18
Update the release date of Safari 6.1

foolip:
The 2013-06-11 date was originally added here:
https://github.com/mdn/browser-compat-data/pull/595

Source for 2013-10-22:
https://en.wikipedia.org/wiki/Safari_version_history#Safari_6

This also makes more sense in light of the relationship between Safari 6
and 7.1, see https://github.com/mdn/browser-compat-data/issues/9423#issuecomment-796880969.

### DIFF
--- a/browsers/safari.json
+++ b/browsers/safari.json
@@ -78,7 +78,7 @@
           "engine_version": "536.25"
         },
         "6.1": {
-          "release_date": "2013-06-11",
+          "release_date": "2013-10-22",
           "status": "retired",
           "engine": "WebKit",
           "engine_version": "537.43"


foolip:
I'd like to remove this entry, but fixing the date first to make it more obviously a good idea.

Elchi3 (approved):
In 2017, my source was the Wikipedia article, too, so that's where 2013-06-11 came from:
https://en.wikipedia.org/w/index.php?title=Safari_version_history&oldid=803745129

Given Wikipedia was updated, I guess we should trust that and update as well.

foolip:
Thanks @Elchi3, good to know where the old data came from :)

PR #11150, opened 2021-06-18
Fix RTCPeerConnection

teoli2003:
Added two missing events: `onicegatheringstatechange_event` and `icecandidateerror_event`(I took their compat info form `onicegatheringstatechange` and `onicecadidateerror` that were both existing.
Added mdn and spec urls for `RTCPeerConnection.getTranceivers`.


foolip:
> I took their compat info form `onicegatheringstatechange` and `onicecadidateerror` that were both existing

@teoli2003 this is often correct, but depressingly often it's not. See https://github.com/mdn/browser-compat-data/issues/7545 for more complexities than you're probably interested in regarding events.

My advice when adding `onfoo` and `foo_event` entries would be:
- First work out when the event can be fired and write a test for that, then use BrowserStack to find out when it was first supported
- Then check if the `onfoo` property was added in the same version
- If it wasn't, add notes about how only one of the methods of listening for the event work for some range of versions, and cry a little for web developers

Digging through implementation source is often faster than the testing approach, but has many pitfalls of its own. In particular, it's hard to be confident whether some code path firing an event is behind a flag or not.

api/RTCPeerConnection.json
```diff
@@ -1878,6 +1880,56 @@
...
         }
       },
+      "icecandidateerror_event": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/RTCPeerConnection/icecandidateerror_event",
+          "spec_url": "https://w3c.github.io/webrtc-pc/#event-icecandidateerror",
+          "description": "<code>icecandidateerror</code> event",
+          "support": {
+            "chrome": {
+              "version_added": false
```
foolip, line 1890: The existing data for `onicecandidateerror` is wrong. This event was implemented in Chromium 77:
https://storage.googleapis.com/chromium-find-releases-static/7d9.html#7d952183f841e90ef48c36c6f29ab896e7175934
https://www.chromestatus.com/feature/6602864602382336

The event also has an interface, and https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnectionIceErrorEvent#browser_compatibility already looks correct. Can you copy that data?

Then, can you mirror this to the other Chromium-based browsers?

api/RTCPeerConnection.json
```diff
@@ -1878,6 +1880,56 @@
...
+            },
+            "edge": {
+              "version_added": false
+            },
+            "firefox": {
+              "version_added": null
+            },
+            "firefox_android": {
+              "version_added": null
+            },
```
foolip, line 1903:
```suggestion
            "firefox": {
              "version_added": false,
              "notes": "See <a href='https://bugzil.la/1561441'>bug 1561441</a>"
            },
            "firefox_android": {
              "version_added": false,
              "notes": "See <a href='https://bugzil.la/1561441'>bug 1561441</a>"
            },
```

api/RTCPeerConnection.json
```diff
@@ -1878,6 +1880,56 @@
...
+              "version_added": false
+            },
+            "opera": {
+              "version_added": null
+            },
+            "opera_android": {
+              "version_added": null
+            },
+            "safari": {
+              "version_added": "14.1"
```
foolip, line 1914: Seems correct, https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Configurations/Version.xcconfig?rev=270101 points to this version too, and this matches existing data.

api/RTCPeerConnection.json
```diff
@@ -2054,6 +2106,56 @@
...
         }
       },
+      "icegatheringstatechange_event": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/RTCPeerConnection/icegatheringstatechange_event",
+          "spec_url": "https://w3c.github.io/webrtc-pc/#event-icegatheringstatechange",
+          "description": "<code>icegatheringstatechange</code> event",
+          "support": {
+            "chrome": {
+              "version_added": "59"
```
foolip, line 2116: Seems correct:
https://www.chromestatus.com/feature/5738839242964992
https://storage.googleapis.com/chromium-find-releases-static/0a7.html#0a734fd400a160bdc1ea1e2eaff4bc44299d523a

api/RTCPeerConnection.json
```diff
@@ -2054,6 +2106,56 @@
...
+            },
+            "edge": {
+              "version_added": "15"
+            },
+            "firefox": {
+              "version_added": "22"
+            },
+            "firefox_android": {
+              "version_added": "44"
```
foolip, line 2128: The Firefox + Firefox for Android data in https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection#browser_compatibility looks pretty implausible, suggesting that the unprefixed `RTCPeerConnection` was available on desktop since 22, but that on Android it was prefixed as `mozRTCPeerConnection` from 22 to 43, and unprefixed in 44.

That's probably wrong, and the likely error is being spread around more here.

https://mdn-bcd-collector.appspot.com/tests/api/mozRTCPeerConnection can be used to figure out when the prefixed support was added and removed, if you have the appetite for it.

api/RTCPeerConnection.json
```diff
@@ -2054,6 +2106,56 @@
...
+            },
+            "ie": {
+              "version_added": false
+            },
+            "opera": {
+              "version_added": "37"
```
foolip, line 2134:
```suggestion
              "version_added": "46"
```

Based on the Chromium version, unless you have other evidence for Opera.

api/RTCPeerConnection.json
```diff
@@ -2054,6 +2106,56 @@
...
+            },
+            "opera_android": {
+              "version_added": "37"
```
foolip, line 2137:
```suggestion
              "version_added": "37"
```
```suggestion
              "version_added": "43"
```

api/RTCPeerConnection.json
```diff
@@ -2054,6 +2106,56 @@
...
+            },
+            "safari": {
+              "version_added": "11"
```
foolip, line 2140: Matches https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/onicegatheringstatechange#browser_compatibility and couldn't be any earlier, so seems right.

foolip (changes requested):

foolip:
I've also sent https://github.com/foolip/mdn-bcd-collector/pull/1246 to improve the collector's ability to detect early support of `RTCPeerConnection` members.

foolip:
@teoli2003 would you like to give this a final push? I'm happy to review again :)

api/RTCPeerConnection.json
teoli2003: The unprefixing happened in Firefox 44 in the code: https://bugzilla.mozilla.org/show_bug.cgi?id=1155923. So RTCPeerConnection on Fx 22 is impossible.

Note that the removal of the unprefixed version has not happened yet: https://bugzilla.mozilla.org/show_bug.cgi?id=1531812.

I'm adding a commit reflecting this.

teoli2003:
(Thanks for reminding me of this PR, lost in the list of those waiting review)

foolip (approved):
There's still lots of Firefox Android data claiming 44 where the Firefox data says something else. This adds one such error, but I'll send a follow-up to fix them all.

teoli2003:
Thanks!

PR #11152, opened 2021-06-18
Remove entries for cloneNode() and importNode() deep parameters

foolip:
These entries represent not the parameter itself, but whether it's
optional and defaulting to false. Since that is the current spec'd
behavior, these entries aren't irrelevant in the "removed for two years"
sense, but if they were inverted to mean "defaults to true" they would
be.

This issue is already documented on both MDN pages:
https://developer.mozilla.org/en-US/docs/Web/API/Node/cloneNode
https://developer.mozilla.org/en-US/docs/Web/API/Document/importNode


foolip:
@ddbeck removed as requested in https://github.com/mdn/browser-compat-data/pull/11080. I think this is fine, but it's borderline removing accurate data that in a way we usually don't.

Not sure if the MDN pages should also be purged of this historical tidbit?

ddbeck:
Sorry if it seemed like I was suggesting something unorthodox. Here's what I was thinking:

The "`deep` optional" feature could also be represented as a feature (of sorts) where deep is required. _That_ feature has been removed from all browsers, so we could drop it as irrelevant. This was the thing that inspired my comment.

The "<code>deep</code> defaults to <code>false</code>" feature is similar in that we could represent it as "<code>deep</code> defaults to <code>true</code>". This does seem less friendly to wholesale removal because it seems that only one browser had a difference worth mentioning. Perhaps a better way to handle this would be to put a note on the parent feature (for Firefox) noting that, for some versions, it had the wrong default value.

foolip:
Capturing this in a note seems OK, but then I'd want to test all ancient browsers since most likely the note applies to more than Firefox.

This would be a useless note to show to anyone today, would a range for just that note be acceptable, even without `partial_implementation`?

If it were up to me I'd just delete these entries. I'm very confident I could find invocations of APIs that work today that didn't work in ancient times that we don't have notes for, and shouldn't add notes for IMHO.

This tension could also be resolved with a better mechanism for applying notes to a range of releases, and making sure that ancient notes are hidden on MDN.

sideshowbarker:
This one needs rebasing…

foolip:
I wrote http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=9482 to test the `cloneNode()` behavior a bit, and it's looking like this was only ever a Firefox problem.

This could be represented as notes for Firefox, but it just seems like a net negative to represent this fact at all in BCD. Unless we have a way of hiding the information everywhere web developers might see it, either a note or a separate `partial_implementation` range is something a web developer might spend a few seconds looking at, for no good at all, since Firefox 29 was released in 2014.

I'm now much more enthusiastic about deleting :)

sideshowbarker (approved):

ddbeck:
@foolip @sideshowbarker This got merged too quickly, unfortunately. If we're going to remove these features completely, then content changes are needed on both pages (which mention these historic oddities).

foolip:
@ddbeck I've sent https://github.com/mdn/content/pull/6934 to also remove the notes from content.

PR #11153, opened 2021-06-18
Prepare release v3.3.9

ddbeck:


RELEASE_NOTES.md
```diff
@@ -1,5 +1,27 @@
 # Release notes
 
+## [v3.3.9](https://github.com/mdn/browser-compat-data/releases/tag/v3.3.9)
+
+RELEASE DATE TK
+
+### Notable changes
```
foolip, line 7: In https://github.com/mdn/browser-compat-data/pull/11174, iOS 4.3 was removed and a few existing entries changed to 4.2, which has the same browser.

ddbeck: Thanks, @foolip. Done. 👍

PR #11162, opened 2021-06-20
Update Safari versions for fullscreen properties

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the fullscreen-related members of the `Element` and `Document` APIs, based upon manual testing.  Additionally, this corrects any data which indicates a version of Safari iOS lower than 12, as the Full Screen API wasn't introduced until then (per [CanIUse](https://caniuse.com/fullscreen), confirmed in BrowserStack).

Test Code Used: `'webkitRequestFullScreen' in document.body;`


foolip:
There are many more entries for the Fullscreen API, can you check all entries that contain "fullscreen" (case insensitive) and see if any of those also need updating?

foolip:
Is also see the data is incomplete here. As you can see in https://blog.foolip.org/2014/07/21/history-of-the-fullscreen-api/ there are multiple capitalizations of this API, introduced at different times.

foolip:
> document.forms[0]

I'm curious why you've used this test code here and in https://github.com/mdn/browser-compat-data/pull/11164? It will be null or undefined unless there's a `<form>` element in the page. Using `document.body` or `document.documentElement` (the `<html>` element) should work just as well. Or just a `<div>` element for testing these APIs.

api/Document.json
```diff
@@ -5425,10 +5425,12 @@
               "version_added": "32"
             },
             "safari": {
-              "version_added": null
+              "version_added": "5.1",
+              "alternative_name": "webkitfullscreenchange"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "5.1",
```
foolip, line 5432: 5.1 + 5.1 aren't matching versions of Safari and iOS. Is this based on mirroring or testing?

Also, I think the fullscreen API was supported much later on iOS and may still not be fully supported, so some testing of this on real devices would be great.

vinyldarkscratch:
There's no particular reason for `document.forms` vs. other attributes, really -- I just remember that being the way I had to test element API support in IE, so when dealing with older browsers, that's what sticks in my head!  I've been running the test code in either Google.com or on the collector's page.

foolip:
I see. It seems fine to use `document.forms[0]` in this case, but when the code depends on there being some markup present as well (a `<form>` element) including that markup as well would be great. But then it would be simpler to test `document.body` :)

api/Document.json
vinyldarkscratch: All the iOS/iPadOS data has been fixed!  5.1 was based upon mirroring, but looking around, there were many other entries with incorrect data, so I fixed them up as well.

api/Document.json
```diff
@@ -4659,7 +4659,9 @@
               "prefix": "webkit"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": "12",
```
foolip, line 4662: I could only find iPadOS 11 and 12.2 to test http://mdn-bcd-collector.appspot.com/tests/api/Document/webkitExitFullscreen, but this is consistent with those results.

https://www.reddit.com/r/apple/comments/b5ketr/psa_ios_122_has_fullscreen_api_custom_html5_video/ talks about it first being in 12, so this seems correct 👍

api/Document.json
```diff
@@ -4659,7 +4659,9 @@
...
+              "prefix": "webkit",
+              "notes": "Only available on iPad, not on iPhone."
```
foolip, line 4664: For other similar notes about platform support we've added `partial_implementation` false, can you do that here too?

api/Element.json
```diff
@@ -2907,10 +2907,10 @@
               "version_added": "43"
             },
             "safari": {
-              "version_added": null
+              "version_added": false
```
foolip, line 2910: There's a "webkitfullscreenchange" event, it should go back as far as `element.webkitRequestFullScreen()` or `element.webkitRequestFullscreen()`, whichever is older.

api/Element.json
```diff
@@ -2987,10 +2987,10 @@
...
```
foolip, line 2990: "webkitfullscreenerror" is supported.

api/Element.json
```diff
@@ -6487,11 +6487,11 @@
               }
             ],
             "safari": {
-              "version_added": "6",
+              "version_added": "5.1",
```
foolip, line 6490: Is this 6&rarr;5.1 change based on testing something in both 5.1 and 6? Early support is tricky, there's both `webkitRequestFullscreen` and `webkitRequestFullScreen` which the data doesn't capture, and it's easy to get things wrong here I think.

api/Element.json
```diff
@@ -6487,11 +6487,11 @@
...
               "prefix": "webkit"
             },
             "safari_ios": {
-              "version_added": "6",
+              "version_added": "12",
               "prefix": "webkit",
               "notes": "Only available on iPad, not on iPhone. Shows an overlay button which can not be disabled."
```
foolip, line 6496: Per https://www.reddit.com/r/apple/comments/b5ketr/psa_ios_122_has_fullscreen_api_custom_html5_video/ it seems like this note only applies since 12.2, but that's fine, no need to capture that fact.

foolip (changes requested):

api/Element.json
vinyldarkscratch: Yep, this came from testing for `webkitRequestFullscreen` (no capital S) specifically!

api/Element.json
foolip: Ping @vinyldarkscratch, this is blocking this PR.

foolip (approved):

PR #11164, opened 2021-06-20
Add Chrome/Safari versions for api.Element.requestPointerLock

vinyldarkscratch:
This PR adds real values for Chrome and Safari for the `requestPointerLock` member of the `Element` API.  By running `document.forms[0].webkitRequestPointerLock;` in various versions of Chrome and Safari, I have determined that Safari has never supported pointer locks before Safari 10.1, with or without a prefix.  Additionally, Chrome has removed the prefixed version of `requestPointerLock` in the following version after adding the non-prefixed version, so it is not present in Edge.

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -8786,10 +8786,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": "10.1"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": false
             },
             "samsunginternet_android": [
               {
@@ -8864,10 +8864,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": "10.1"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": false
             },
             "samsunginternet_android": [
               {
--- a/api/Element.json
+++ b/api/Element.json
@@ -6627,7 +6627,8 @@
                 "version_added": "37"
               },
               {
-                "version_added": true,
+                "version_added": "22",
+                "version_removed": "38",
                 "prefix": "webkit"
               }
             ],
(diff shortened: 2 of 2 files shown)


foolip:
I've sent https://github.com/foolip/mdn-bcd-collector/pull/1249 to add custom tests for all of the prefixed API here. @vinyldarkscratch would you mind going over all entries with "pointerlock" in their name and make them consistently represent support for the prefixed API?

foolip:
I've tested https://mdn-bcd-collector.appspot.com/tests/api/Element/webkitRequestPointerLock and https://mdn-bcd-collector.appspot.com/tests/api/Element/requestPointerLock in Safari 10 and 10.1 to confirm 10.1 is when the unprefixed API was introduced, and Safari never had the prefixed variant.

api/Document.json
```diff
@@ -8786,10 +8786,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": "10.1"
```
foolip, line 8789: This matches the existing `onpointerlockchange` data 👍

api/Document.json
```diff
@@ -8864,10 +8864,10 @@
...
```
foolip, line 8867: This matches the existing `onpointerlockerror` data 👍

api/Element.json
```diff
@@ -6627,7 +6627,8 @@
                 "version_added": "37"
               },
               {
-                "version_added": true,
+                "version_added": "22",
```
foolip, line 6630: This matches https://caniuse.com/pointerlock, except caniuse.com doesn't capture the fact that Chrome 37 had both the prefixed and unprefixed API. (Confirmed on BrowserStack just now.)

foolip (approved):

PR #11167, opened 2021-06-20
Fix DOMPoint»w|x|y|z spec URLs

sideshowbarker:
There aren’t actually any #dom-dompoint-w|x|y|z fragment IDs in the https://drafts.fxtf.org/geometry/ spec. Instead the spec has shared text that the `DOMPointReadOnly` and `DOMPoint` IDLs both link to, at, e.g., https://drafts.fxtf.org/geometry/#dom-dompointreadonly-x, which says:

> The x attribute, on getting, must return the x coordinate value. For the DOMPoint interface, setting the x attribute must set the x coordinate to the new value.

That is, the first sentence of that shared text applies to both the `DOMPointReadOnly` and `DOMPoint` interfaces, while the second sentence applies only to the `DOMPoint` interface.

cc @teoli2003

### DIFF
--- a/api/DOMPoint.json
+++ b/api/DOMPoint.json
@@ -150,7 +150,7 @@
       "w": {
         "__compat": {
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/DOMPoint/w",
-          "spec_url": "https://drafts.fxtf.org/geometry/#dom-dompoint-w",
+          "spec_url": "https://drafts.fxtf.org/geometry/#dom-dompointreadonly-w",
           "support": {
             "chrome": {
               "version_added": "61"
@@ -247,7 +247,7 @@
       "x": {
         "__compat": {
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/DOMPoint/x",
-          "spec_url": "https://drafts.fxtf.org/geometry/#dom-dompoint-x",
+          "spec_url": "https://drafts.fxtf.org/geometry/#dom-dompointreadonly-x",
           "support": {
             "chrome": {
               "version_added": "61"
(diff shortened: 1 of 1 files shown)


teoli2003:
Thanks. Good catch.

teoli2003:
@sideshowbarker Shouldn't your amazing fragment tester being added to the CI or to `npm test` so that we catch these before it lands? It would be great!

Earthman100:
Ha no as it is a personal Dom that is set for personal use as it does not communicate ith the eb for exchange only to view and authenticate by its set rules as it is similar to verizon or At&t, They are not the same network nor would they have other networks making posts above the owner of the document and or server.


Sincerely, Sean P. Kropaczewski

sideshowbarker:
> @sideshowbarker Shouldn't your amazing fragment tester being added to the CI or to `npm test` so that we catch these before it lands? It would be great!

Yeah I guess I should open a PR to add it.

The source is currently at https://github.com/w3c/mdn-spec-links/blob/master/.check-spec-urls.js and I’m anyway running it regularly as part of building all the https://github.com/w3c/mdn-spec-links data. It takes about 35 seconds to run, so I guess that wouldn’t slow down BCD too much.

I should note though the the data it relies on to do its checks is a 19MB list at https://github.com/w3c/mdn-spec-links/blob/master/SPECURLS.json that has every single fragment ID for every single spec that we have any spec URLs for in BCD.

The script which builds that `SPECURLS.json` file  is at https://github.com/w3c/mdn-spec-links/blob/master/.make-SPECURLS.js

It takes about 20 to 25 minutes for that script to (re)build the `SPECURLS.json` file — so I don’t run it every day. Instead I only re-build the file whenever the `.check-spec-urls.js` script reports any errors.

But I think GitHub Actions has some way to run cron jobs, so I guess I could also set up something that re-builds the `SPECURLS.json` file once a day or so, and checks it back in. That certainly should be often enough.

teoli2003:
Yes, I think GitHub Actions can run cron jobs and open issues. See  https://github.com/w3c/browser-specs/issues/317 that does this I guess.

cc/ @Elchi3 & @ddbeck

Elchi3:
Yes! Sounds like something to outline in a new issue and then implement if we can come up with a good plan for automation.

Elchi3 (approved):

PR #11173, opened 2021-06-21
Remove Safari iOS 3.1

vinyldarkscratch:
This PR removes Safari iOS 3.1 from the browser data, and changes everything set to Safari iOS 3.1 to mirrored data from Safari Desktop.  Safari iOS 3.1's WebKit version is the exact same as iOS 3.0's, as confirmed by querying WhatIsMyBrowser.

Using the Safari 3 emulator in BrowserStack, I've spot-checked a few entries to ensure this methodology is correct.  I have also found that all of the data for Safari iOS 3.1 came from the initial wiki tables, and most entries are accompanied by Safari Desktop set to 3.1, which further leads me to believe that Safari iOS 3.1 was mistakenly added.

~~Blocked by #11188.~~

### DIFF
--- a/css/at-rules/media.json
+++ b/css/at-rules/media.json
@@ -917,7 +917,7 @@
                 "version_added": "1.3"
               },
               "safari_ios": {
-                "version_added": "1"
+                "version_added": "3.1"
               },
               "samsunginternet_android": {
                 "version_added": "1.0"
--- a/css/at-rules/media.json
+++ b/css/at-rules/media.json
@@ -917,7 +917,7 @@
                 "version_added": "1.3"
               },
               "safari_ios": {
-                "version_added": "3.1"
+                "version_added": "1"
               },
               "samsunginternet_android": {
                 "version_added": "1.0"
--- a/css/properties/empty-cells.json
+++ b/css/properties/empty-cells.json
@@ -34,7 +34,7 @@
               "version_added": "1.2"
             },
             "safari_ios": {
-              "version_added": "3.1"
+              "version_added": "1"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
--- a/css/properties/outline.json
+++ b/css/properties/outline.json
@@ -43,7 +43,7 @@
               "version_added": "1.2"
             },
             "safari_ios": {
-              "version_added": "3.1"
+              "version_added": "1"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
--- a/css/properties/unicode-bidi.json
+++ b/css/properties/unicode-bidi.json
@@ -34,7 +34,7 @@
               "version_added": "1.3"
             },
             "safari_ios": {
-              "version_added": "3.1"
+              "version_added": "1"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
--- a/html/elements/input/email.json
+++ b/html/elements/input/email.json
@@ -36,7 +36,7 @@
                 "version_added": true
               },
               "safari_ios": {
-                "version_added": "2",
+                "version_added": "3",
                 "notes": [
                   "Doesn't do validation, but instead offers a custom 'email' keyboard, which is designed to make entering email addresses easier.",
                   "The custom 'email' keyboard does not provide a comma key, so users cannot enter multiple email addresses.",
--- a/html/elements/input/email.json
+++ b/html/elements/input/email.json
@@ -36,7 +36,7 @@
                 "version_added": true
               },
               "safari_ios": {
-                "version_added": "3.1",
+                "version_added": "2",
                 "notes": [
                   "Doesn't do validation, but instead offers a custom 'email' keyboard, which is designed to make entering email addresses easier.",
                   "The custom 'email' keyboard does not provide a comma key, so users cannot enter multiple email addresses.",
--- a/api/ApplicationCache.json
+++ b/api/ApplicationCache.json
@@ -36,7 +36,7 @@
             "version_added": "4"
           },
           "safari_ios": {
-            "version_added": "3.1"
+            "version_added": "3"
           },
           "samsunginternet_android": {
             "version_added": "1.0"
@@ -139,7 +139,7 @@
               "version_added": "4"
             },
             "safari_ios": {
-              "version_added": "3.1"
+              "version_added": "3"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
@@ -191,7 +191,7 @@
               "version_added": "4"
             },
             "safari_ios": {
-              "version_added": "3.1"
+              "version_added": "3"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
@@ -243,7 +243,7 @@
               "version_added": "4"
             },
             "safari_ios": {
-              "version_added": "3.1"
+              "version_added": "3"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
@@ -295,7 +295,7 @@
               "version_added": "4"
             },
             "safari_ios": {
-              "version_added": "3.1"
+              "version_added": "3"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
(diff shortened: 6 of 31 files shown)


foolip:
> confirmed by querying WhatIsMyBrowser

How can I do this? Did you visit https://www.whatismybrowser.com/ with iOS 3.0 and 3.1, or is there a way to query their database?

> mirrored data from Safari Desktop

Did you use the mirroring script? Since the [mapping](https://gist.github.com/foolip/6872cf5350dccf6224fce984fb73fba1) before Safari / iOS 9 is not as trivial, I'm skeptical of any existing code for this :)

foolip:
I've looked for UA strings and came up with this:

https://developers.whatismybrowser.com/useragents/parse/841safari-ios-iphone-webkit has:

> Mozilla/5.0 (iPhone; U; CPU iPhone OS 3_0 like Mac OS X; en-us) AppleWebKit/528.18 (KHTML, like Gecko) Version/4.0 Mobile/7A341 Safari/528.16

https://user-agents.net/s/o05iLj has:

> Mozilla/5.0 (iPhone; U; CPU iPhone OS 3_0 like Mac OS X; en-us) AppleWebKit/528.18 (KHTML, like Gecko) Version/4.0 Mobile/7A341 Safari/528.16; Ipad

https://thadafinser.github.io/UserAgentParserComparison/v5/user-agent-detail/fb/5e/fb5ef5de-9363-45a6-b364-585feccd2cab.html has:

> Mozilla/5.0 (iPhone; U; CPU iPhone OS 3_1 like Mac OS X; HW iPhone1,1; de_de) AppleWebKit/525.18.1 (KHTML, like Gecko) (AdMob-iSDK-20090617)


https://user-agents.net/s/rxAjIRSBrt has:

> Mozilla/5.0 (iPhone; U; CPU iPhone OS 3_1 like Mac OS X; ru-ru) AppleWebKit/528.18 (KHTML, like Gecko) Version/4.0 Mobile/7C144 Safari/528.16

There's some weird tokens in some of those strings, but 528.18 is in all of them.

And it seems reasonable to say that Safari 4 ≈ iOS 3. But only as an approximation, the WebKit versions don't match.

foolip:
Let's also consider the UA string of iOS 3.2 from https://stackoverflow.com/a/5255476:

> Mozilla/5.0 (iPhone; U; CPU OS 3_2 like Mac OS X; en-us) AppleWebKit/531.21.10 (KHTML, like Gecko) Version/4.0.4 Mobile/7B334b Safari/531.21.10

That 4.0.4 probably means Safari 4, and 531.21.10 also appears in https://en.wikipedia.org/wiki/Safari_version_history#Safari_4 so it all makes sense. It does however make me question whether we should consider iOS 3.2 significant.

foolip:
If we trust our `safari.json` and `safari_ios.json` the sequence of releases around this time was:
- 2009-06-08: Safari 4 using WebKit 530.17
- 2009-06-17: iOS 3 using WebKit 528.18
- 2010-04-03: iOS 3.2 using WebKit 531.21

If we throw [Wikipedia](https://en.wikipedia.org/wiki/Safari_version_history#Safari_4) into the mix we could add:
- 2009-11-11: Safari 4.0.4 using WebKit 531.21.10
- 2010-03-11: Safari 4.0.5 using WebKit 531.22.7

I don't know how the branching was done, but if we just consider the WebKit version numbers and the UA strings (above) it seems like Safari 4 maps to either iOS 3 or 3.2. It seems unlikely to me that there's much difference between the two since the same differences on desktop were just treated as patch versions.

foolip:
To review this, it's also necessary to understand (believe) the Safari 3.1 &rarr; iOS 2 mapping. [Wikipedia](https://en.wikipedia.org/wiki/Safari_version_history#Safari_3) supports the 525.13 WebKit version we have in `safari.json` and the UA string for iOS 2 seems to be:

https://www.w3.org/2007/03/mth/9e573d0661a52f38282af0a625fd32fc.html has:

> Mozilla/5.0 (iPhone; U; CPU iPhone OS 2_0 like Mac OS X; en-us) AppleWebKit/525.18.1 (KHTML, like Gecko) Version/3.1.1 Mobile/5A347 Safari/525.20

https://developers.whatismybrowser.com/useragents/parse/829764safari-ios-ipod-webkit has:

> Mozilla/5.0 (iPod; U; CPU iPhone OS 2_0 like Mac OS X; en-us) AppleWebKit/525.18.1 (KHTML, like Gecko) Version/3.1.1 Mobile/5A347 Safari/525.20

Note the 3.1.1 there, which is the Safari version. So that all supports the mapping Safari 3.1 &rarr; iOS 2.

css/at-rules/media.json
```diff
@@ -917,7 +917,7 @@
                 "version_added": "1.3"
               },
               "safari_ios": {
-                "version_added": "3.1"
+                "version_added": "1"
```
foolip, line 920: That 1.3 &rarr; 1 mapping here is correct, but are you confident that the Safari data is correct and not the iOS data?

css/properties/empty-cells.json
```diff
@@ -34,7 +34,7 @@
               "version_added": "1.2"
             },
             "safari_ios": {
-              "version_added": "3.1"
+              "version_added": "1"
```
foolip, line 37: Is the Safari 1.2 info reliable?

css/properties/outline.json
```diff
@@ -43,7 +43,7 @@
               "version_added": "1.2"
             },
             "safari_ios": {
-              "version_added": "3.1"
+              "version_added": "1"
```
foolip, line 46: Is the Safari 1.2 info reliable?

css/properties/unicode-bidi.json
```diff
@@ -34,7 +34,7 @@
               "version_added": "1.3"
             },
             "safari_ios": {
-              "version_added": "3.1"
+              "version_added": "1"
```
foolip, line 37: Is the Safari 1.3 info reliable?

html/elements/input/email.json
```diff
@@ -36,7 +36,7 @@
                 "version_added": true
               },
               "safari_ios": {
-                "version_added": "3.1",
+                "version_added": "2",
```
foolip, line 39: The Safari data here is just true. This feature is also likely OS-dependent and needs manual testing to see what the UI looks like. Is iOS 2 based on testing, or a general 3.1 &rarr; 2 replacement?

foolip (commented):
I trust the cases that are now 4+3 or 3.1+2 and have commented on all cases that aren't either pair. With those checked I think this change is good to go.

css/at-rules/media.json
vinyldarkscratch: I've opened #11188 to fix the inaccurate data here.

css/properties/empty-cells.json
vinyldarkscratch: Just confirmed the data is accurate!

css/properties/outline.json
vinyldarkscratch: Confirmed!  `outline` has an effect in Safari 1.2.

css/properties/unicode-bidi.json
vinyldarkscratch: Yep, support in Safari 1.3 confirmed!

html/elements/input/email.json
vinyldarkscratch: This one is a general replacement, as I have no way of testing old iOS versions (I've only got BrowserStack right now).  After running a quick test in iOS 3.0 and 4.0, however, I do notice that the special email keyboard doesn't show up in iOS 3.0, so it probably _was_ introduced in 3.1...

Should we just say iOS 3 for this?

foolip:
@vinyldarkscratch what's this blocked on now? Do I just need to review again?

vinyldarkscratch:
It just needs a re-review now, yep!

foolip (approved):
I've reviewed top-to-bottom again without finding any irregularities to poke at.

foolip:
This will also allow the mirror migration script to replace a lot more data.

PR #11174, opened 2021-06-21
Remove Safari iOS 4.3

vinyldarkscratch:
This PR removes Safari iOS 4.3 from the browser data, and adjusts everything that is set as iOS 4.3 to 4.2 instead. Safari iOS 4.3's WebKit version is the exact same as iOS 4.2's, as confirmed by querying WhatIsMyBrowser.

### DIFF
--- a/browsers/safari_ios.json
+++ b/browsers/safari_ios.json
@@ -45,12 +45,6 @@
           "engine_version": "533.17",
           "release_date": "2010-11-22"
         },
-        "4.3": {
-          "status": "retired",
-          "engine": "WebKit",
-          "engine_version": "533.17",
-          "release_date": "2011-03-09"
-        },
         "5": {
           "status": "retired",
           "engine": "WebKit",
--- a/css/selectors/placeholder.json
+++ b/css/selectors/placeholder.json
@@ -93,7 +93,7 @@
               },
               {
                 "prefix": "-webkit-input-",
-                "version_added": "4.3"
+                "version_added": "4.2"
               }
             ],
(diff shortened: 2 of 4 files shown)


foolip:
This one looks small and simple enough. I found two sources for what the UA string on iOS 4.3 was:

https://stackoverflow.com/a/5738659 says:
> Mozilla/5.0 (iPhone; U; CPU iPhone OS 4_3_2 like Mac OS X; en-us) AppleWebKit/533.17.9 (KHTML, like Gecko) Version/5.0.2 Mobile/8H7 Safari/6533.18.5

https://gist.github.com/moomerman/1240682 says among other things:
> Mozilla/5.0 (iPhone; U; CPU iPhone OS 4_2_1 like Mac OS X; en-us) AppleWebKit/533.17.9 (KHTML, like Gecko) Version/5.0.2 Mobile/8C148 Safari/6533.18.
> Mozilla/5.0 (iPhone; U; CPU iPhone OS 4_3_1 like Mac OS X; en-us) AppleWebKit/533.17.9 (KHTML, like Gecko) Version/5.0.2 Mobile/8G4 Safari/6533.18.5

So 533.17.9 seems very likely correct, for 4.3 and we can fold it into 4.2. It also seems like the Safari version was 5.0.2, which makes sense since the WebKit version of Safari 5 we have is 533.16.

foolip (approved):
The Safari versions here are all 5, and that makes sense considering the WebKit versions and that the UA string for iOS had 5.0.2 here.

PR #11175, opened 2021-06-21
Remove Safari iOS 5.1

vinyldarkscratch:
This PR removes Safari iOS 5.1 from the browser data, and changes everything set to Safari iOS 5.1 to mirrored data from Safari Desktop. Safari iOS 5.1's WebKit version is the exact same as iOS 5.0's, as confirmed by querying WhatIsMyBrowser.

Using the Safari iOS 4 emulator in BrowserStack, I've spot-checked a few entries to ensure this methodology is correct. I have also found that all of the data for Safari iOS 351 came from the initial wiki tables, and most entries are accompanied by Safari Desktop set to 5.1, which further leads me to believe that Safari iOS 5.1 was mistakenly added.

An additional note is that worker support was indeed pushed back to a later release in Safari iOS than Safari Desktop, probably due to OS implementation or the likes.  I've set its version to 5.0.


foolip:
I've found some additional sources of the UA strings:

https://stackoverflow.com/a/7825888 says:

> Mozilla/5.0 (iPhone; CPU iPhone OS 5_0 like Mac OS X) AppleWebKit/534.46 (KHTML, like Gecko) Version/5.1 Mobile/9A334 Safari/7534.48.3

https://pypi.org/project/user-agents-next/ says, in an example:

> Mozilla/5.0 (iPhone; CPU iPhone OS 5_1 like Mac OS X) AppleWebKit/534.46 (KHTML, like Gecko) Version/5.1 Mobile/9B179 Safari/7534.48.3

So it looks like the WebKit version is 534.46 for both and the corresponding Safari version is 5.1 for both.

html/elements/input/range.json
```diff
@@ -36,7 +36,7 @@
                 "version_added": "3.1"
               },
               "safari_ios": {
-                "version_added": "5.1"
+                "version_added": "2"
```
foolip, line 39: Is Safari 3.1 above correct?

javascript/builtins/DataView.json
```diff
@@ -139,7 +139,7 @@
                   "version_added": "5.1"
                 },
                 "safari_ios": {
-                  "version_added": "5.1"
+                  "version_added": "4.2"
```
foolip, line 142: Are you sure this shouldn't be 5.1+5?

foolip (commented):
I've taken all the matching 5+5.1 as trustworthy, and have commented on any Safari + iOS pairs other than that. With those investigated I'd be confident that this isn't changing data in the wrong direction.

api/KeyboardEvent.json
```diff
@@ -1441,7 +1441,7 @@
               "version_added": "5"
             },
             "safari_ios": {
-              "version_added": "5.1"
+              "version_added": "4.2"
```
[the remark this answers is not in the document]
vinyldarkscratch: The WebKit versions for Safari 5 and iOS 4.2 match almost perfectly, which is why I'm thinking iOS 4.2 is accurate.  I have no way of testing this as of the time of this writing, so I'm purely going off WebKit versions.  (This goes for all of the other 5.1 -> 4.2 changes in this PR.)

foolip:
@vinyldarkscratch are you planning to go through the comments to finish this?

vinyldarkscratch:
I am, yeah -- I found out that BrowserStack has an iPad on iOS 4.3 emulator, so I can collect the results for iOS 4.2!  I'll be tackling that soon.

api/KeyboardEvent.json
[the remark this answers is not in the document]
vinyldarkscratch: Seems that the `keyCode` is just "0" in iOS 4.2.  I'll set this to iOS 5!

html/elements/input/range.json
vinyldarkscratch: Just researched this some more.  The Safari 3.1 is correct, however it appears that iOS support wasn't introduced until Safari iOS 5.  Fixing!

api/KeyboardEvent.json
```diff
@@ -1441,7 +1441,7 @@
               "version_added": "5"
             },
             "safari_ios": {
-              "version_added": "5.1"
+              "version_added": "5"
```
foolip, line 1444: Continuing from https://github.com/mdn/browser-compat-data/pull/11175#discussion_r655585705

It sounds like `keyCode` was in iOS 4.2 but always 0. If so I would still set this to 4.2, maybe it only worked with a keyboard attached, or is broken in the emulator. Without the ability to test, I'd prefer to assume the matching release:

```suggestion
              "version_added": "4.2"
```

api/MessagePort.json
```diff
@@ -37,7 +37,7 @@
             "version_added": "5"
           },
           "safari_ios": {
-            "version_added": "5.1"
+            "version_added": "5"
```
foolip, line 40: You updated everything in MessageChannel.json. Should everything in this file also be 4.2?

```suggestion
            "version_added": "4.2"
```

api/WebSocket.json
```diff
@@ -1050,7 +1050,7 @@
               "version_added": "5"
             },
             "safari_ios": {
-              "version_added": "5.1"
+              "version_added": "5"
```
foolip, line 1053: Are you sure this shouldn't be 4.2? (I can't find a previous thread, maybe I missed this in the first round.)

api/Window.json
```diff
@@ -6503,11 +6503,11 @@
               "version_added": "11.5"
             },
             "safari": {
-              "version_added": "6",
+              "version_added": "5",
               "notes": "Before version 10, Safari would fire a <code>popstate</code> event on page load."
             },
             "safari_ios": {
-              "version_added": "5.1",
+              "version_added": "5",
```
foolip, line 6510: Shouldn't this be 4.2?

api/WindowEventHandlers.json
```diff
@@ -528,10 +528,10 @@
               "version_added": "11.5"
             },
             "safari": {
-              "version_added": "6"
+              "version_added": "5"
             },
             "safari_ios": {
-              "version_added": "5.1"
+              "version_added": "5"
```
foolip, line 534: Shouldn't this be 4.2?

html/elements/input/range.json
```diff
@@ -36,7 +36,7 @@
                 "version_added": "3.1"
               },
               "safari_ios": {
-                "version_added": "5.1"
+                "version_added": "5"
```
foolip, line 39: Discussed in https://github.com/mdn/browser-compat-data/pull/11175#discussion_r671003559

foolip (commented):
I've gone through everything again and found a few more irregular things that I'm not sure about.

api/WebSocket.json
vinyldarkscratch: This one's for worker support, which was introduced in Safari iOS 5.  The rest of the API is set to 4.2, which I've just double-checked is accurate!

foolip: Ah, thanks!

foolip (approved):
Reviewed end-to-end yet again, all looks nice and orderly now! Thanks Vinyl!

PR #11156, opened 2021-06-19
Remove Safari 6.1

queengooborg:
This PR removes Safari 6.1 from the data and converts all of its entries to Safari 7.  Details are described in https://github.com/mdn/browser-compat-data/issues/9423, but to summarize, it seems that Safari 6.1 and 6.2 are simply backports of Safari 7.0 and 7.1 for older macOS versions, similar to [Safari 4.1 and 5.0](https://github.com/mdn/browser-compat-data/issues/4679).

PR #11184, opened 2021-06-21
Change Safari 6.1 to 7 where the iOS version is already 7

foolip:
This subset of https://github.com/mdn/browser-compat-data/pull/11156
does not need much scrutiny, as this results in the expected matching
of Safari and iOS versions. There may still be errors in this data, but
then it's wrong for both Safaris.

Part of https://github.com/mdn/browser-compat-data/issues/9423.

### DIFF
--- a/api/AudioBufferSourceNode.json
+++ b/api/AudioBufferSourceNode.json
@@ -428,7 +428,7 @@
               "version_added": "14"
             },
             "safari": {
-              "version_added": "6.1"
+              "version_added": "7"
             },
             "safari_ios": {
               "version_added": "7"
--- a/api/AudioScheduledSourceNode.json
+++ b/api/AudioScheduledSourceNode.json
@@ -265,7 +265,7 @@
               "version_added": "14"
             },
             "safari": {
-              "version_added": "6.1"
+              "version_added": "7"
             },
             "safari_ios": {
               "version_added": "7"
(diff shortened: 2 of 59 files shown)


api/Window.json
```diff
@@ -1041,11 +1041,11 @@
             },
             "safari": [
               {
-                "version_added": "6.1"
+                "version_added": "7"
               },
               {
                 "version_added": "6",
-                "version_removed": "6.1",
+                "version_removed": "7",
```
foolip, line 1048: This case is a little bit irregular since there's no corresponding prefixed entry for safari_ios. There probably should be.

api/BarProp.json
```diff
@@ -28,7 +28,7 @@
             "version_added": "16"
           },
           "safari": {
-            "version_added": "6.1"
+            "version_added": "7"
```
foolip, line 31: Lint is failing due to this change, so I'll back it out.

api/Window.json
foolip: http://mdn-bcd-collector.appspot.com/tests/api/Window/webkitCancelAnimationFrame reveals there's more wrong here, the API is still supported!

I'll back out the change from this PR.

foolip:
@vinyldarkscratch this is a substantial subset of https://github.com/mdn/browser-compat-data/pull/11156. I've gone over it twice myself to only leave the stuff that's very regular looking. I think landing this (and https://github.com/mdn/browser-compat-data/pull/11185) and then rebasing https://github.com/mdn/browser-compat-data/pull/11156 would make review easier.

vinyldarkscratch (approved):
Thanks for splitting these changes out, looking good to me!  👍

PR #11185, opened 2021-06-21
Update Safari versions for BarProp

foolip:
This matches window.toolbar and other window.*bar properties updated in
https://github.com/mdn/browser-compat-data/pull/6825 based on
mdn-bcd-collector results and manual testing.

Additionally, http://mdn-bcd-collector.appspot.com/tests/api/BarProp/visible
was tested in Safari 4 and iOS 3 (the oldest available on BrowserStack)
to confirm support there. Support going back to 3 / 1 is assumed based
on the other data.

### DIFF
--- a/api/BarProp.json
+++ b/api/BarProp.json
@@ -28,10 +28,10 @@
             "version_added": "16"
           },
           "safari": {
-            "version_added": "6.1"
+            "version_added": "3"
           },
           "safari_ios": {
-            "version_added": "7"
+            "version_added": "1"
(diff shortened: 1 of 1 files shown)


foolip:
@vinyldarkscratch this is a change I had to break out of https://github.com/mdn/browser-compat-data/pull/11184 to pass the lint.

vinyldarkscratch (approved):
Great catch, thanks for tracking this one down!

PR #4369, opened 2019-06-24
Set real values for Chromium + Safari CSS @media properties

vinyldarkscratch:
This PR is to assist in the efforts of #3710, or more specifically, #4294.  I've browsed through WebKit and Blink to identify the exact versions where features were added/removed (if supported at all).  Based on the following data, this sets versions within Chrome, Chrome Android, Safari, Safari iOS, Opera, Opera Android, Samsung Internet, and WebView:

<details>
	<summary>css.at-rules.media.color, css.at-rules.media.device-aspect-ratio, css.at-rules.media.device-height, css.at-rules.media.device-width, css.at-rules.media.grid, css.at-rules.media.height, css.at-rules.media.width</summary>
	<ol>
		<li>CSS3 Media Queries implementation in WebKit 1f63b91e8004c50e4ab184a522913cadf5866ddd</li>
		<li>Commit defines WebCore/css/MediaQueryEvaluator.cpp</li>
		<li>Parsing script found in file in commit</li>
		<li>File initially found in 521.13, contains parseAspectRatio function</li>
		<li><strong>Safari 3, Chrome 1</strong></li>
	</ol>
</details>

<details>
	<summary>css.at-rules.media.aspect-ratio, css.at-rules.media.orientation</summary>
	<ol>
		<li>Support added in WebKit 9476eb7cbc7ca31b649e1a54ea3ecf69f00f1a76</li>
		<li><a href="https://bugs.webkit.org/show_bug.cgi?id=25197">https://bugs.webkit.org/show_bug.cgi?id=25197</a></li>
		<li>Introduced in 532.0</li>
		<li><strong>Safari 4.1, Chrome 3</strong></li>
		<li>Note: Opera was set to "53" for some odd reason.  This was changed to "true" to leave room for pre-15 Opera implementation.</li>
	</ol>
</details>

<details>
	<summary>css.at-rules.media.color-gamut</summary>
	<ol>
		<li>Support added in WebKit 1452c11ffcad445fd8a4b631cd4bdd6d167cec99</li>
		<li><a href="https://bugs.webkit.org/show_bug.cgi?id=155994">https://bugs.webkit.org/show_bug.cgi?id=155994</a></li>
		<li>Introduced in 602.1.30</li>
[...]

PR #11188, opened 2021-06-22
Add Safari versions for media CSS @rule

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `media` CSS @rule.  Since there is nothing in the [source code for Safari 2](https://trac.webkit.org/browser/webkit/tags/old/Safari-421/WebCore/khtml/css) for media queries, it doesn't make sense for this to be set to Safari 1.3.  As I have found in #4369, support for pretty much every part of media queries hadn't been introduced until at least Safari 3.

### DIFF
--- a/css/at-rules/media.json
+++ b/css/at-rules/media.json
@@ -35,7 +35,7 @@
               "version_added": "10.1"
             },
             "safari": {
-              "version_added": "1.3"
+              "version_added": "3"
             },
             "safari_ios": {
               "version_added": "1"
@@ -914,10 +914,10 @@
                 "version_added": "10.1"
               },
               "safari": {
-                "version_added": "1.3"
+                "version_added": "3"
               },
               "safari_ios": {
-                "version_added": "3.1"
+                "version_added": "3"
               },
               "samsunginternet_android": {
                 "version_added": "1.0"
--- a/css/at-rules/media.json
+++ b/css/at-rules/media.json
@@ -917,7 +917,7 @@
                 "version_added": "3"
               },
               "safari_ios": {
-                "version_added": "3"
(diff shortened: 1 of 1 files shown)


css/at-rules/media.json
```diff
@@ -914,10 +914,10 @@
                 "version_added": "10.1"
               },
               "safari": {
-                "version_added": "1.3"
+                "version_added": "3"
               },
               "safari_ios": {
-                "version_added": "3.1"
+                "version_added": "3"
```
foolip, line 920: Should this be 1 like above?

foolip:
You linked to https://trac.webkit.org/browser/webkit/tags/old/Safari-421/WebCore/khtml/css, did you mean https://trac.webkit.org/browser/webkit/tags/old/Safari-412/WebCore/khtml/css? (421 vs. 412)

Judging by https://caniuse.com/css-mediaqueries there was initial partial support. I think relying on source code for this is not going to be too reliable. Maybe a simple test can be written instead?

vinyldarkscratch:
I did mean to link the earlier tag, but I think that helps show the lack of support!

It appears that CanIUse doesn't track data as far back as Safari 2 as it is (they have Safari 3.1 listed as the first version).  I ran a quick test with a simple media query (`@media screen and (max-width: 700px) {}`) in both Safari 2 and 3, and have confirmed there's no support in Safari 2.

foolip (approved):

foolip (approved):
This also matches https://developer.mozilla.org/en-US/docs/Web/API/CSSMediaRule#browser_compatibility, good!

Issue #10956, opened 2021-06-11
Mirroring script mirrors Safari 10.1 to iOS 11, should be 10.3


foolip:
Steps to reproduce:
> npm run mirror -- -m always safari_ios api/CSSNamespaceRule.json

This will replace the existing "10.3" (correct) with "11". This file and `-m always` is used just to illustrate the problem, this is mostly a problem when adding and mirroring new data.

It's likely there's a lot of incorrectly mirrored data as a result of this bug.

cc @apple-web-evangelist @jensimmons @gsnedders @vinyldarkscratch

foolip:
The WebKit version is "603.2.1" for both Safari 10.1 and iOS 10.3, so it's not because of a WebKit version mismatch...

PR #11191, opened 2021-06-22
Sort browser releases in mirroring script

queengooborg:
This PR fixes the bug reported in #10956.  Since JavaScript orders the keys of a dictionary not in the way we would like it to, we need to sort the browser keys using our trusty version comparison dependency.

### DIFF
--- a/scripts/mirror.js
+++ b/scripts/mirror.js
@@ -47,29 +47,41 @@ const { argv } = require('yargs').command(
 
 /**
  * @param {string} dest_browser
- * @param {ReleaseStatement} source_browser_release
+ * @param {ReleaseStatement} source_release
  * @returns {ReleaseStatement|boolean}
  */
-const getMatchingBrowserVersion = (dest_browser, source_browser_release) => {
+const getMatchingBrowserVersion = (dest_browser, source_release) => {
   const browserData = browsers[dest_browser];
   const releaseKeys = Object.keys(browserData.releases);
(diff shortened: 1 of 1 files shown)


foolip (approved):

foolip:
@ddbeck or @Elchi3 for review and merging?

foolip:
I've tried running the script with local changes to see if mirroring now matches https://gist.github.com/foolip/6872cf5350dccf6224fce984fb73fba1. It still doesn't for these cases:
- 3 &rarr; 2, should be 1
- 4 &rarr; 3.2, but what is even correct?
- 5.1 &rarr; 6, should be 5 or 5.1
- 7 &rarr; 8, should be 7
- 14.1 &rarr; "Invalid argument not valid semver ('611.1.21.161.3' received)" 😆 
- 15 &rarrl; "Invalid argument expected string"

vinyldarkscratch:
The issues with Safari should be handled with the aid of the new additions in #11299, where we'll change the mirroring script to use hard coded mappings between the two.  Due to the variations in engine versions between iOS and Desktop, we can't really use our generic mapping algorithm, unfortunately.

Regarding the other crashes, #11333 should handle the Safari 14.1 issue, and I'll see if I can come up with a fix for the issue with Safari 15!

queengooborg:
I'm merging this PR since I have an R+ from a fellow BCD owner!

PR #11195, opened 2021-06-22
Add label usage docs

ddbeck:
#### Summary

This PR adds some documentation for label usage on mdn/browser-compat-data and captures some changes to how I intend to use the needs-release-note label.

This fell out of a discussion I had with @Elchi3 yesterday. Mostly, I wanted a vehicle to change how I used the needs-release-note label (which is to remove the label, once the note is done). But there's also a new "needs content change" label which isn't exactly obvious, if you're not already in the know. I figure this might also be a good place to capture other workflow stuff, eventually.

#### Related issues

This is somewhat related to but doesn't re&ZeroWidthSpace;solve https://github.com/mdn/browser-compat-data/issues/4178.

### DIFF
--- a/docs/workflow.md
+++ b/docs/workflow.md
@@ -1,0 +1,57 @@
+# Issue and pull requests workflow
+
+Or, _How changes are considered, accepted, merged, and published_
+
+## Label usage
+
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
Thanks Daniel! 👍

Issue #6175, opened 2020-05-18
Reduce the number of special case feature groups


ddbeck:
Some consumers have to handle special groups of features in the structure of the data. We should reduce or eliminate these special cases.

In general, we have a few broad namespaces in BCD, in which features with `__compat` data reside. The `api` namespace is typical of this: `api.<featurename>` is a parent of `__compat` data. `css` is more finely divided, with `css.properties`, `css.selectors`, etc.

The trouble is within a namespace, such as [`css.properties.gap`](https://github.com/mdn/browser-compat-data/blob/master/css/properties/gap.json). `css.properties.gap` is _not_ a parent to `__compat` data. Instead, it's a surprise namespace for `flex_context`, `grid_context`, etc. We should avoid these surprises in the structure of the data.

The first step to resolving this would be to make a list of all paths through the structure which are not parents to `__compat`.

jculver050:
Close

ddbeck:
This is dupe of https://github.com/mdn/browser-compat-data/issues/13804 (well, actually, a subset of). Closing in favor of the other issue, since it's a bit more comprehensive.

PR #11214, opened 2021-06-23
Fix place-* css properties

teoli2003:
These 3 properties had no bcd directly on them, but only on their subfeatures. I added it (taking the earliest one). That way mdn_url and spec_url are at the right level.

### DIFF
--- a/css/properties/place-self.json
+++ b/css/properties/place-self.json
@@ -1,4 +1,4 @@
-  {
+{
   "css": {
     "properties": {
       "place-self": {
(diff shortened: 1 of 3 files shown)


css/properties/place-self.json
```diff
@@ -1,12 +1,57 @@
-{
+  {
```
foolip, line 1: What's this change and why isn't it failing the lint? 🧐

teoli2003: Weird. Fixed in next commit.

foolip (approved):

ddbeck:
Related to https://github.com/mdn/browser-compat-data/issues/6175

PR #11223, opened 2021-06-24
Add Internet Explorer 11 as a range

vinyldarkscratch:
This PR adds `≤11` as a valid range for Internet Explorer.  With IE's market share now [less than 1%](https://gs.statcounter.com/browser-market-share), along with the age of its releases, I don't think most developers will care about how far back something has been supported in IE (or in most cases, IE support altogether).  Additionally, this will aid in speeding up the removal of null and true values.

This PR is inspired by a comment from @foolip in #10381.

### DIFF
--- a/schemas/compat-data-schema.md
+++ b/schemas/compat-data-schema.md
@@ -275,6 +275,7 @@ For certain browsers, ranged versions are allowed as it is sometimes impractical
   - "≤79" (supported in some version Chromium-based Edge and possibly in EdgeHTML-based Edge)
 - Internet Explorer
   - "≤6" (earliest IE version supported in BrowserStack)
+  - "≤11" (supported in some version of IE)
 - Opera
   - "≤12.1" (supported in some version of Presto-based Opera)
   - "≤15" (supported in some version of Chromium-based Opera and possibly in Presto-based Opera)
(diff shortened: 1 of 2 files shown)


foolip (approved):
I think we should do this 👍

ddbeck:
This is going to need documentation changes: https://github.com/mdn/browser-compat-data/blob/main/schemas/compat-data-schema.md#ranged-versions

schemas/compat-data-schema.md
```diff
@@ -275,6 +275,7 @@ For certain browsers, ranged versions are allowed as it is sometimes impractical
   - "≤79" (supported in some version Chromium-based Edge and possibly in EdgeHTML-based Edge)
```
foolip, line 275: I notice this documentation is misleading. It's not in "some" version of Chromium-based Edge, to use this range one has to test Edge 79 specifically.

Same issue with Opera 15 range.

We should probably remove those ranges :)

foolip (approved):
Anyway, the added docs look good.

lucalves (approved):

foolip:
Ping @ddbeck for review.

Elchi3 (approved):

ddbeck (approved):
Let's do it! ✔️

PR #11234, opened 2021-06-25
Fix feature structure of HTMLCanvasElement API (and correct version numbers)

vinyldarkscratch:
This PR fixes up the names of the features within the `HTMLCanvasElement` API.  Many of the feature names did not conform to the [documented guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#parameters-and-parameter-object-features), so this is an attempt to get them to conform properly.  This also adds the `2d` context to our data, which was originally missing.

As a part of the research of adding in the context and organizing the options, I've found some errors within our data.  This PR fixes the data as well.  (I would've separated it into another PR, but I thought it would be better to tackle it in one go.)


api/HTMLCanvasElement.json
```diff
@@ -147,6 +147,55 @@
             "deprecated": false
           }
         },
+        "alpha_attribute": {
```
foolip, line 150: This should actually be "options_alpha_parameter" since it represents the "alpha" dictionary member of the second argument to `getContext`.

Unfortunately there are no spec links for these since it's not a Web IDL dictionary but is rather treated as a JavaScript object.

foolip (changes requested):

api/HTMLCanvasElement.json
```diff
@@ -195,9 +195,58 @@
             }
           }
         },
-        "desynchronized": {
+        "options_alpha_parameter": {
           "__compat": {
-            "description": "<code>desynchronized</code> attribute",
+            "description": "<code>options.alpha</code> parameter",
+            "spec_url": "https://www.khronos.org/registry/webgl/specs/latest/1.0/#WEBGLCONTEXTATTRIBUTES",
```
foolip, line 201: I would say https://www.khronos.org/registry/webgl/specs/latest/1.0/#5.2.1 is a better URL for this, or links to the individual bits of that section if IDs exist.

Note also that this and other attributes may be supported by 2d canvas and webgl2, in which case those should also be linked.

foolip: `alpha` is also supported by the "2d" and "bitmaprenderer" contexts:
https://html.spec.whatwg.org/multipage/canvas.html#2d-context-creation-algorithm
https://html.spec.whatwg.org/multipage/canvas.html#imagebitmaprenderingcontext-creation-algorithm

Looking at this I thought we should group these parameters under the context features. In other words, wthat e should have these entries:
- api.HTMLCanvasElement.getContext.2d_context.options_alpha_parameter
- api.HTMLCanvasElement.getContext.2d_context.options_desynchronized_parameter
- api.HTMLCanvasElement.getContext.2d_context.options_colorSpace_parameter
- api.HTMLCanvasElement.getContext.2d_context.options_willReadFrequently_parameter
- api.HTMLCanvasElement.getContext.bitmaprenderer_context.options_alpha_parameter
- api.HTMLCanvasElement.getContext.webgl_context.options_alpha_parameter
- api.HTMLCanvasElement.getContext.webgl_context.options_antialias_parameter
- api.HTMLCanvasElement.getContext.webgl_context.options_depth_parameter
- api.HTMLCanvasElement.getContext.webgl_context.options_desynchronized_parameter
- api.HTMLCanvasElement.getContext.webgl_context.options_failIfMajorPerformanceCaveat_parameter
- api.HTMLCanvasElement.getContext.webgl_context.options_powerPreference_parameter
- api.HTMLCanvasElement.getContext.webgl_context.options_premultipliedAlpha_parameter
- api.HTMLCanvasElement.getContext.webgl_context.options_preserveDrawingBuffer_parameter
- api.HTMLCanvasElement.getContext.webgl_context.options_stencil_parameter
- api.HTMLCanvasElement.getContext.webgl2_context._REPEAT_THE_ABOVE_

However, due mainly due to the duplication between "webgl" and "webgl2" entries, this seems unwieldy. The parameters supported in more than one context (treating WebGL as one) are:
- `alpha` in 2d, bitmaprenderer and webgl(2)
- `desynchronized` in 2d and webgl(2)

So we could probably do merged entries representing both and use notes or `partial_implementation` if there's any complication. The downside is that which parameters that work for a given context won't be clear, and people might expect "colorSpace" to work with WebGL.

I'd like to hear from @ddbeck and @Elchi3 how they'd deal with this situation.

ddbeck: > So we could probably do merged entries representing both and use notes or partial_implementation if there's any complication. The downside is that which parameters that work for a given context won't be clear, and people might expect "colorSpace" to work with WebGL.

As much as I hate all the length and repetition of the duplication, it has the benefit of being direct. Deduplicating seems to maximize complexity for readers and BCD contributors alike. And if we truly hate the longer option, then I imagine it'll be easier to collapse things that it would be to go the other direction.

That's my take. I don't feel strongly about this stuff, though.

foolip: OK, I could definitely live with the deduplicated approach. If I squint and think about this like mixins merging together different things, I'm horrified by my suggestion to merge the entries :)

@vinyldarkscratch do you want to try rearranging the existing entries like that? No need to add what's missing I think, just sorting out what we have is more than enough for this PR :)

vinyldarkscratch: Finished sorting out the options into their respective contexts!  I ended up having to fix some of the data as well, so I've re-titled this PR to increase its scope.  I added the missing `2d` context, but I haven't added any of the individual options -- that'll be a later thing!

foolip (commented):
The structure looks great now! I have a lot of questions about the versions. This might turn into a giant rabbit hole, so if some of it can be turned into true or ranges to avoid digging too deep, I'd be OK with that. We'll come back to it :)

Or, we can excavate this rabbit hole and build a tidy little museum down there.

api/HTMLCanvasElement.json
```diff
@@ -185,171 +185,62 @@
...
+              "spec_url": "https://www.khronos.org/registry/webgl/specs/latest/1.0/#WEBGLCONTEXTATTRIBUTES",
+              "support": {
+                "chrome": {
+                  "version_added": "56"
                 },
-                {
-                  "version_added": "75",
-                  "notes": "ChromeOS only"
+                "chrome_android": {
+                  "version_added": false
```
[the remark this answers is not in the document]
vinyldarkscratch: This had come from the original data for this option, but I had a sneaking suspicion it might have been incorrect.  I haven't verified this data yet, however.

api/HTMLCanvasElement.json
```diff
@@ -515,54 +626,372 @@
...
               "deprecated": false
             }
+          },
+          "options_alpha_parameter": {
+            "__compat": {
+              "description": "<code>options.alpha</code> parameter",
+              "spec_url": "https://www.khronos.org/registry/webgl/specs/latest/1.0/#WEBGLCONTEXTATTRIBUTES",
+              "support": {
+                "chrome": {
+                  "version_added": "32"
```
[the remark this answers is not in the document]
vinyldarkscratch: This is from the original data; I haven't done any new testing in this PR.

api/HTMLCanvasElement.json
```diff
@@ -461,6 +357,215 @@
...
+                    "version_added": "81",
+                    "notes": "ChromeOS and Windows"
+                  },
+                  {
+                    "version_added": "75",
+                    "notes": "ChromeOS only"
+                  },
+                  {
+                    "version_added": "71",
+                    "alternative_name": "lowLatency"
```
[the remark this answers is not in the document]
vinyldarkscratch: Seems that https://www.chromestatus.com/feature/6360971442388992 indicates this was implmemented right in Chrome 75, including on Chrome Android, so this is probably wrong.  I'm not sure how to test for support, though, so I can't confirm the OS differences.

api/HTMLCanvasElement.json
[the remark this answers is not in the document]
vinyldarkscratch: Ran a test using the following code and found there is indeed Android support.  Updating!

```html
<div id="test">
	<canvas id="canvas"></canvas>
</div>

<script>
	var canvas = document.getElementById('canvas');
	var ctx = canvas.getContext('2d', {alpha: true});

	ctx.fillStyle = "#ff0000aa";
	ctx.fillRect(10, 10, 100, 100);
</script>
```

foolip:
@vinyldarkscratch I notice in another review that `OffscreenCanvas` also has a similar structure. Can you reorganize that too so that the style is kept consistent between the two?

api/HTMLCanvasElement.json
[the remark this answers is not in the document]
vinyldarkscratch: Follow-up: I've now tested this and confirmed the `alpha` parameter is supported since Chrome 32!

foolip:
@vinyldarkscratch there are still unresolved comments here, over to you :)

api/HTMLCanvasElement.json
```diff
@@ -461,6 +357,215 @@
...
+                  "version_added": "12"
+                },
+                "firefox": {
+                  "version_added": "41"
+                },
+                "firefox_android": {
+                  "version_added": "41"
+                },
+                "ie": {
+                  "version_added": true
```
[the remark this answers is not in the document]
vinyldarkscratch: I'm unsure, this came from the original data.  I'm actually not sure how to test this particular option.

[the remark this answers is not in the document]
foolip: OK, let's leave it for now, it's true so we'll get back to it.

api/HTMLCanvasElement.json
```diff
@@ -32,7 +32,7 @@
             "notes": "Opera Mini 5.0 and later has partial support."
           },
           "safari": {
-            "version_added": "3"
+            "version_added": "2"
```
foolip, line 35: This is consistent with `CanvasRenderingContext2D` 👍

api/HTMLCanvasElement.json
```diff
@@ -461,6 +357,212 @@
...
+          },
+          "options_desynchronized_parameter": {
+            "__compat": {
+              "description": "<code>options.desynchronized</code> parameter",
+              "spec_url": "https://www.khronos.org/registry/webgl/specs/latest/1.0/#WEBGLCONTEXTATTRIBUTES",
+              "support": {
+                "chrome": [
+                  {
+                    "version_added": "81",
+                    "notes": "ChromeOS and Windows"
```
foolip, line 418: Looks like https://bugs.chromium.org/p/chromium/issues/detail?id=945835 is the bug about enabling this.

foolip (approved):

ddbeck:
A summary of the feature name changes that happened in this PR:

## Removed

- `api.HTMLCanvasElement.getContext.desynchronized`
- `api.HTMLCanvasElement.getContext.failIfMajorPerformanceCaveat`
- `api.HTMLCanvasElement.getContext.powerPreference`
- `api.HTMLCanvasElement.getContext.2d_alpha_context`
- `api.HTMLCanvasElement.toBlob.Image_quality`

## Added

- `api.HTMLCanvasElement.getContext.bitmaprenderer_context.options_alpha_parameter`
- `api.HTMLCanvasElement.getContext.webgl_context.options_alpha_parameter`
- `api.HTMLCanvasElement.getContext.webgl_context.options_desynchronized_parameter`
- `api.HTMLCanvasElement.getContext.webgl_context.options_failIfMajorPerformanceCaveat_parameter`
- `api.HTMLCanvasElement.getContext.webgl_context.options_powerPreference_parameter`
- `api.HTMLCanvasElement.getContext.webgl2_context.options_alpha_parameter`
- `api.HTMLCanvasElement.getContext.webgl2_context.options_desynchronized_parameter`
- `api.HTMLCanvasElement.getContext.webgl2_context.options_failIfMajorPerformanceCaveat_parameter`
- `api.HTMLCanvasElement.getContext.webgl2_context.options_powerPreference_parameter`
- `api.HTMLCanvasElement.getContext.2d_context`
- `api.HTMLCanvasElement.getContext.2d_context.options_alpha_parameter`
- `api.HTMLCanvasElement.getContext.2d_context.options_desynchronized_parameter`
- `api.HTMLCanvasElement.toBlob.quality_parameter`

PR #11235, opened 2021-06-25
Add HTTP Sec-CH-UA-* hints

### LINKED ISSUE #11190: Sec-CH-UA User Client Hints - Time to document?
*User Agent Client Hints*  went into [Chrome 89](https://www.chromestatus.com/feature/5995832180473856). 
- The [spec](https://wicg.github.io/ua-client-hints/) is a W3C Draft Community Group Report - which presumably counts as "on a standard track".
- Mozilla's [position](https://mozilla.github.io/standards-positions/) is "not harmful"
- There is no signal from Safari according to [that Chrome 89 info](https://www.chromestatus.com/feature/5995832180473856)

The MDN convention on [When to document new technologies](https://developer.mozilla.org/en-US/docs/MDN/Guidelines/Conventions_definitions#when_to_document_new_technologies) doesn't give me clear direction on whether we document this on MDN and in BCD. I feel it probably is time, but perhaps I'm overly influenced by chromium's effective market share.

Is it time? 

FYI @Elchi3 @ddbeck

hamishwillee:
Fixes #11190

Adds header entries for all the User Agent client hints and makes corresponding updates the `Accept-CH` header.
- Spec: https://wicg.github.io/ua-client-hints/
- Added to Chrome in 89 as per https://www.chromestatus.com/feature/5995832180473856
- Demo/testing gitchme: https://user-agent-client-hints.glitch.me/

Appropriate versions added to all chromiums. There is no corresponding version for Samsung Internet yet.

This does not include the header for [Sec-CH-UA-Bitness](https://wicg.github.io/ua-client-hints/#sec-ch-ua-bitness) as that is not tested by the demo and does not appear to be implemented on quick code scan.

### DIFF
--- a/http/headers/accept-ch.json
+++ b/http/headers/accept-ch.json
@@ -93,6 +93,342 @@
             "status": {
               "experimental": true,
               "standard_track": false,
+              "deprecated": true
+            }
+          }
+        },
+        "Sec-CH-UA": {
+          "__compat": {
+            "description": "<code>Sec-CH-UA</code>",
+            "support": {
+              "chrome": {
+                "version_added": "89"
+              },
+              "chrome_android": {
+                "version_added": "89"
+              },
+              "edge": {
+                "version_added": "89"
+              },
+              "firefox": {
+                "version_added": false
+              },
+              "firefox_android": {
+                "version_added": false
+              },
+              "ie": {
+                "version_added": false
+              },
+              "opera": {
+                "version_added": "75"
+              },
+              "opera_android": {
+                "version_added": "63"
+              },
+              "safari": {
+                "version_added": false
+              },
+              "safari_ios": {
+                "version_added": false
+              },
+              "samsunginternet_android": {
+                "version_added": false
+              },
+              "webview_android": {
+                "version_added": "89"
+              }
+            },
+            "status": {
+              "experimental": true,
+              "standard_track": true,
+              "deprecated": false
+            }
+          }
+        },
+        "Sec-CH-UA-Arch": {
+          "__compat": {
+            "description": "<code>Sec-CH-UA-Arch</code>",
+            "support": {
+              "chrome": {
+                "version_added": "89"
+              },
+              "chrome_android": {
+                "version_added": "89"
+              },
+              "edge": {
+                "version_added": "89"
+              },
+              "firefox": {
+                "version_added": false
+              },
+              "firefox_android": {
+                "version_added": false
+              },
+              "ie": {
+                "version_added": false
+              },
+              "opera": {
+                "version_added": "75"
+              },
+              "opera_android": {
+                "version_added": "63"
+              },
+              "safari": {
+                "version_added": false
+              },
+              "safari_ios": {
+                "version_added": false
+              },
+              "samsunginternet_android": {
+                "version_added": false
+              },
+              "webview_android": {
+                "version_added": "89"
+              }
+            },
+            "status": {
+              "experimental": true,
+              "standard_track": true,
+              "deprecated": false
+            }
+          }
+        },
+        "Sec-CH-UA-Full-Version": {
+          "__compat": {
+            "description": "<code>Sec-CH-UA-Full-Version</code>",
+            "support": {
+              "chrome": {
+                "version_added": "89"
+              },
+              "chrome_android": {
+                "version_added": "89"
+              },
+              "edge": {
+                "version_added": "89"
+              },
+              "firefox": {
+                "version_added": false
+              },
+              "firefox_android": {
+                "version_added": false
+              },
+              "ie": {
+                "version_added": false
+              },
+              "opera": {
+                "version_added": "75"
+              },
+              "opera_android": {
+                "version_added": "63"
+              },
+              "safari": {
+                "version_added": false
+              },
+              "safari_ios": {
+                "version_added": false
+              },
+              "samsunginternet_android": {
+                "version_added": false
+              },
+              "webview_android": {
+                "version_added": "89"
+              }
+            },
+            "status": {
+              "experimental": true,
+              "standard_track": true,
+              "deprecated": false
+            }
+          }
+        },
+        "Sec-CH-UA-Mobile": {
+          "__compat": {
+            "description": "<code>Sec-CH-UA-Mobile</code>",
+            "support": {
+              "chrome": {
+                "version_added": "89"
+              },
+              "chrome_android": {
+                "version_added": "89"
+              },
+              "edge": {
+                "version_added": "89"
+              },
+              "firefox": {
+                "version_added": false
+              },
+              "firefox_android": {
+                "version_added": false
+              },
+              "ie": {
+                "version_added": false
+              },
+              "opera": {
+                "version_added": "75"
+              },
+              "opera_android": {
+                "version_added": "63"
+              },
+              "safari": {
+                "version_added": false
+              },
+              "safari_ios": {
+                "version_added": false
+              },
+              "samsunginternet_android": {
+                "version_added": false
+              },
+              "webview_android": {
+                "version_added": "89"
+              }
+            },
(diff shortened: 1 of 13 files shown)


http/headers/accept-ch.json
```diff
@@ -93,6 +93,342 @@
             "status": {
               "experimental": true,
               "standard_track": false,
+              "deprecated": true
+            }
+          }
+        },
+        "Sec-CH-UA": {
```
ddbeck, line 100: @hamishwillee: Help me understand something here. Servers include `Accept-CH` headers in responses; clients are supposed to react to those headers by setting corresponding headers in their next request. A server sends `Accept-CH: Sec-CH-UA` and a client is supposed set `Sec-CH-UA: somevalue`.

My question is this: what is this feature representing, exactly? That a client sends `Sec-CH-UA` headers? That a client sends `Sec-CH-UA` in response to `Accept-CH: Sec-CH-UA`? Or something else to do with clients processing `Accept-CH: Sec-CH-UA` responses?

I think the answer to that question will suggest something about whether this feature should be a child of `Accept-CH` or ought to be its own entry as a request header alone.

hamishwillee: @ddbeck Sorry for delay in response! 

> Help me understand something here. Servers include `Accept-CH` headers in responses; clients are supposed to react to those headers by setting corresponding headers in their next request. A server sends `Accept-CH: Sec-CH-UA` and a client is supposed set `Sec-CH-UA: somevalue`.

Yes, with very minor caveats:
- A server specifies  `Accept-CH` with a header in any/every response to tell clients what hints it is capable of using, if provided. 
- Clients may opt-in to responding with these hints (i.e. it is optional, and defined by the client). So more "could" than "is supposed to".
- The client can respond if and when it wants, not just the "next request". Could be the next request, or it could be every subsequent request where it makes sense to do so. 

> My question is this: what is this feature representing, exactly? That a client sends `Sec-CH-UA` headers? That a client sends `Sec-CH-UA` in response to `Accept-CH: Sec-CH-UA`? Or something else to do with clients processing `Accept-CH: Sec-CH-UA` responses?

Good question. For Accept-CH I think it just means client "understands" this directive and may send the header indicated in the directive as a response.

> I think the answer to that question will suggest something about whether this feature should be a child of `Accept-CH` or ought to be its own entry as a request header alone.

Currently it is both. 
-  I did not invent this approach - it just extends what was done before.
- It is useful to have both because the docs for Accept-CH then show the versions in which you can add the various directives. 
- Just FYI, Accept-CH is in a different spec from the actual headers it will get in responses (i.e. Accept-CH is in the client hints infrastructure spec, but there are numerous different client hints that live in their own specs.

Hope that clarifies. Happy to be lead by you!

hamishwillee: Can I please get a review on this?

ddbeck: > > My question is this: what is this feature representing, exactly? That a client sends `Sec-CH-UA` headers? That a client sends `Sec-CH-UA` in response to `Accept-CH: Sec-CH-UA`? Or something else to do with clients processing `Accept-CH: Sec-CH-UA` responses?
> 
> Good question. For Accept-CH I think it just means client "understands" this directive and may send the header indicated in the directive as a response.

Sorry, I have to poke at this a little more (because I don't get it—not as a critique). On the `Accept-CH` side, what does "understand" mean? How would I test this to know if it's supported or not, since there's nuance about some-or-all subsequent requests? (I saw the link to the demo, but the view source link doesn't work and I can't figure out what's supposed to happen with just the browser dev tools.)

Perhaps more critically, when would the compat data for `http.headers.Accept-CH.Sec-CH-UA` be different from the compat data for `http.headers.Sec-CH-UA`?

> > I think the answer to that question will suggest something about whether this feature should be a child of `Accept-CH` or ought to be its own entry as a request header alone.
>
> …
>
> * I did not invent this approach - it just extends what was done before.

Can you point to an example of this and maybe provide a little commentary on it? I feel a bit dumb here, but I can't figure out what this structure is similar to.

> * Just FYI, Accept-CH is in a different spec from the actual headers it will get in responses (i.e. Accept-CH is in the client hints infrastructure spec, but there are numerous different client hints that live in their own specs.

That's fine. That's why we can put a spec URL on any feature. 😆

ddbeck (commented):
Sorry, @hamishwillee. More questions for you. It's a bit frustrating, because there's a lot of fuzziness about what compat data means for HTTP.

hamishwillee:
@ddbeck   I'm sure you not understanding is me, not you. If the below doesn't help, we should call in jpmedley :-)

> Perhaps more critically, when would the compat data for http.headers.Accept-CH.Sec-CH-UA be different from the compat data for http.headers.Sec-CH-UA?

Never - versions will never be different / will always be the same.

A server sends the  `Accept-CH` in a response to tell the client which client hint headers it is capable of using (if it gets them in a subsequent request). The client can then choose to send these hints in subsequent requests, where appropriate. 

So a browser "understanding" `Sec-CH-UA` in the `Accept-CH` response header implies that it also understands that it can send `Sec-CH-UA` _when appropriate_ in future requests. 

It doesn't change things with respect to your question, but there is a small rub to this that some headers may be sent even if not requested in an `ACCEPT-CH` (as I understand it). These are called the low entropy hints. They are basically considered safe as they don't have much info that can be used for fingerprinting. These are listed in https://wicg.github.io/client-hints-infrastructure/#low-entropy-table

Did that help?

>  On the Accept-CH side, what does "understand" mean?

I _think_ answered above. It means that the browser might include the specified hint in future requests, with the expectation that the server is capable of formatting the response based on that hint.

> How do I test it

With respect to "how to test it", I used https://user-agent-client-hints.glitch.me/ 
Below is side by side test in Chrome and Firefox. 
image

No idea how this works, but the fact it shows the sorts of things I expect in the different browsers is indicative.

ddbeck (changes requested):
That sort of helps. At least, it helps enough.

Since we don't have standalone guidelines for HTTP, let's do something a bit weird to stake out what we're doing more clearly so that, should we ever actually clean up the mess that is HTTP data, we'll at least have a clearer idea of what we've done here:

* For each of the `http.header.Sec-CH-UA*` features, add a description field which clearly identifies it as a request header (e.g., `<code>Sec-CH-UA-Arch</code> request header`, or similar—use your judgment).
* For each of the children of `http.headers.Accept-CH`, add some text to the description that explains we're talking about header list item tokens. That is, show we're representing the response header values with these subfeatures. I'm not settled on what this text should be, but I trust you to come up with something.

What do you think of that?

http/headers/accept-ch.json
```diff
@@ -93,6 +93,342 @@
             "status": {
               "experimental": true,
               "standard_track": false,
+              "deprecated": true
```
hamishwillee, line 96: FYI the "old" spec versions are deprecated, and either replaced with Sec- versions in new spec or just dropped in a few cases.

hamishwillee:
@ddbeck Thanks. Firstly, I have changed as you suggested:
- Header descriptions like.: `<code>Sec-CH-UA-Arch</code> request header`
- Accept-CH tokens like: "`<code>Sec-CH-UA-Arch</code> token`

Then:
- Accept-CH header description to: `<code>Accept-CH</code> client hint accept (response) header`
- Added additional tokens for all the other client hint headers, and made their descriptions also consistent. 

I was considering providing more information for the headers themselves - e.g. `<code>Sec-CH-UA-Arch</code> user-agent client request header`. This is useful information. My assumption though, based on the patterns used throughout BCD is that the descriptions must be succinct and minimalist.

FYI only, there are a bunch more client hint headers coming - e.g. replacements for Width etc which are now deprecated. Waiting for them to appear in a Chrome build before adding them.

ddbeck (approved):
Looks great. Thank you, @hamishwillee!

PR #11244, opened 2021-06-25
missing shadowRoot property and AriaMixin spec URLs

rachelandrew:
I've been documenting ElementInternals in https://github.com/mdn/content/pull/6335

This adds a missing property `shadowRoot`, and also all of the spec URLs for `AriaMixin` in ElementInternals.


Elchi3 (approved):
Thanks! 👍

PR #11259, opened 2021-06-26
Add spec_url & mdn_url to DataTransferIterm.getAsFileSystemHandle

teoli2003:
Both the spec_url & mdn_url were missing for this method.

### DIFF
--- a/api/DataTransferItem.json
+++ b/api/DataTransferItem.json
@@ -99,8 +99,8 @@
       },
       "getAsFileSystemHandle": {
         "__compat": {
-          "spec_url": "https://wicg.github.io/file-system-access/#dom-datatransferitem-getasfilesystemhandle",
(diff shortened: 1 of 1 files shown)


api/DataTransferItem.json
```diff
@@ -99,6 +99,8 @@
       },
       "getAsFileSystemHandle": {
         "__compat": {
+          "spec_url": "https://wicg.github.io/file-system-access/#dom-datatransferitem-getasfilesystemhandle",
```
foolip, line 102: Looks like there's no lint for this, but can you put mdn_url before spec_url?

foolip (changes requested):

api/DataTransferItem.json
teoli2003: Sure! Fix in next commit.

foolip (approved):

PR #11266, opened 2021-06-26
Remove Safari 3.2 from BCD

vinyldarkscratch:
This PR removes Safari 3.2 from BCD.  After comparing the results collected from the mdn-bcd-collector project for both 3.1 and 3.2, I had only found one difference between them: the User Agent string.  On top of that, I've spot checked a number of the entries reported as Safari 3.2 that the collector doesn't check, and found pretty much everything to be supported in Safari 3.1, so there is no need to track Safari 3.2 separately.

### DIFF
--- a/css/properties/-webkit-mask-box-image.json
+++ b/css/properties/-webkit-mask-box-image.json
@@ -33,7 +33,7 @@
               "version_added": "3.1"
             },
             "safari_ios": {
-              "version_added": "3.2"
+              "version_added": "2"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
--- a/css/properties/-webkit-mask-box-image.json
+++ b/css/properties/-webkit-mask-box-image.json
@@ -30,7 +30,7 @@
               "version_added": "14"
             },
             "safari": {
-              "version_added": "3.2"
+              "version_added": "3.1"
             },
             "safari_ios": {
               "version_added": "3.2"
--- a/api/EventTarget.json
+++ b/api/EventTarget.json
@@ -625,7 +625,7 @@
               "version_added": "10.1"
             },
             "safari": {
-              "version_added": "3.2"
+              "version_added": "3.1"
             },
             "safari_ios": {
               "version_added": "3"
--- a/api/HTMLMediaElement.json
+++ b/api/HTMLMediaElement.json
@@ -304,7 +304,7 @@
               "version_added": "≤12.1"
             },
             "safari": {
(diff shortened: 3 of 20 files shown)


foolip:
Pasting what @vinyldarkscratch said on chat since it covers what was spot checked:

> I'm getting a PR written up right now, and after removing 3.2 from the browser data and running the linter, it seems that there's only a few entries that have differences: a few APIs (which we most certainly test), some CSS properties and selectors (I've spot checked a few of them real quick, and confirmed that 3.1 is correct), the SVG `<font>` element, and the `void` operator. All in all, there's only a few potentially correct differences, but it's not enough to consider it a new browser version.

css/properties/-webkit-mask-box-image.json
```diff
@@ -30,7 +30,7 @@
               "version_added": "14"
             },
             "safari": {
-              "version_added": "3.2"
+              "version_added": "3.1"
             },
             "safari_ios": {
               "version_added": "3.2"
```
foolip, line 36: This was probably wrong, just 3.2 in both Safaris. iOS 2 is more likely. Since it will look like someone tried to convey "iOS support came after macOS" with these changes, is there any way to check if iOS 3.2 is correct?

vinyldarkscratch: I actually think we should probably just get rid of this entry altogether.  Taking a look at `mask-border`, it specifies `-webkit-mask-box-image` as an alternative name.  (The data in `mask-border` also says Safari 3.1 / iOS 3, which leads me to believe iOS 3.2 is incorrect.)

vinyldarkscratch:
Gotta love it when you are deleting your local branch and accidentally select the remote... -_-

css/properties/-webkit-mask-box-image.json
foolip: This was just the first entry with iOS 3.2, there are a bunch and it would be good to look at all of them to see if the iOS data is likely to be wrong. (If left unchanged, it's more likely to look like a deliberate differences when the versions don't align.)

vinyldarkscratch: I've gone through and updated all of the entries where both iOS and macOS are set to 3.2, and adjusted them!  I've spot checked a few of the entries to confirm they're supposed to be in Safari 3.1, and none of them are entries that should vary in between iOS and macOS editions.

foolip (approved):

PR #11272, opened 2021-06-28
Add Safari versions for api.Element.click_event.on_disabled_form_elements

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `click_event.on_disabled_form_elements` member of the `Element` API, based upon manual testing.

Test Code Used:
```html
<textarea id="input" disabled></textarea>

<script>
  document.getElementById('input').addEventListener('click', function() {
    console.log('Hi!');
  });
</script>
````

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -1788,10 +1788,10 @@
                 "notes": "Only works for <code>&lt;textarea&gt;</code> elements and some <code>&lt;input&gt;</code> element types."
               },
               "safari": {
-                "version_added": null
+                "version_added": false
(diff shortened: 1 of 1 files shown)


foolip:
Can you share the full document used for testing when some markup is also involved? I assume there was a disabled attribute in the input element here, but the PR description doesn't say.

vinyldarkscratch:
Sorry about that, fixed!

foolip (approved):

PR #11285, opened 2021-06-28
Add missing mdn_url

teoli2003:
While working on a script to check the coherence between mdn_url here and the content of the browser-compat in MDN's front-runner, I found a few existing MDN pages without the mdn_url clause here.

I'm fixing these.

All these URLs do exist.

### DIFF
--- a/api/HTMLHtmlElement.json
+++ b/api/HTMLHtmlElement.json
@@ -50,6 +50,7 @@
       },
       "version": {
         "__compat": {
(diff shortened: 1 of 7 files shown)


foolip (approved):

sideshowbarker:
Ideally this PR would have also added `spec_url` values along with the `mdn_url` values. See #11443.

Going forward, this is actually something we could lint for. In fact, https://github.com/mdn/browser-compat-data/pull/11430 adds the linting for it. cc @vinyldarkscratch

PR #11286, opened 2021-06-28
Add `accepts_flags` to the browsers schema as an optional hint

ddbeck:
#### Summary

This PR proposes an addition to the schema, `accepts_flags` for both browsers as a whole and individual releases. This is me trying to having it both ways: providing an easy option (per-browser) for easy cases and a harder option (per-release) for harder cases.

I defined this part of the schema as an optional hint, rather than to trying to use it to drive the linter immediately or do anything more restrictive. I'm proposing it in this shallow state now, so that we can roll it into the next breaking release (which we require for schema changes).

#### Related issues

https://github.com/mdn/browser-compat-data/issues/10616

### DIFF
--- a/browsers/samsunginternet_android.json
+++ b/browsers/samsunginternet_android.json
@@ -2,6 +2,7 @@
   "browsers": {
     "samsunginternet_android": {
       "name": "Samsung Internet",
+      "accepts_flags": false,
       "releases": {
         "1.0": {
           "release_date": "2013-04-27",
--- a/browsers/webview_android.json
+++ b/browsers/webview_android.json
@@ -2,6 +2,7 @@
   "browsers": {
     "webview_android": {
       "name": "WebView Android",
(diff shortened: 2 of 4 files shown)


ddbeck:
(Flagged as not ready, so we don't merge it before the next breaking release, likely to take place in July.)

Elchi3 (approved):
I like this approach 👍

vinyldarkscratch (approved):
I love this approach as well, so much cleaner than just putting the data into a linter, allowing for our consumers to easily utilize the data too!

ddbeck:
Since this is approved and I'm planning for the next release (Thursday) to be v4.0.0, I'm merging this now.

queengooborg:
I'm opting to remove the release-level property in #15990 because of a lack of implementation.  We've only been using the browser-level version in practice and in linting.

PR #11292, opened 2021-06-28
Add Safari versions for GlobalEventHandlers API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `GlobalEventHandlers` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.8).

Tests Used:
https://mdn-bcd-collector.appspot.com/tests/api/Document/onloadend
https://mdn-bcd-collector.appspot.com/tests/api/Document/onshow

### DIFF
--- a/api/GlobalEventHandlers.json
+++ b/api/GlobalEventHandlers.json
@@ -2202,10 +2202,10 @@
               "version_added": false
             },
             "safari": {
-              "version_added": null
+              "version_added": false
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": false
             },
             "samsunginternet_android": {
               "version_added": false
@@ -4321,10 +4321,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": false
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": false
             },
             "samsunginternet_android": {
               "version_added": false


foolip:
https://mdn-bcd-collector.appspot.com/tests/api/GlobalEventHandlers is 404 and there are many things that could be tested to determine the right versions here, to the extend that a merged mixin view of things can be right. Which tests were used?

vinyldarkscratch:
Fixed the links now!

foolip:
Both "loadend" and "show" are in https://github.com/WebKit/WebKit/blob/main/Source/WebCore/dom/EventNames.h suggesting some level of support. Can you check loadevent_event and show_event entries and see if they reveal anything?

foolip:
I've confirmed `onloadend` is an odd mishap in Gecko and filed https://bugzilla.mozilla.org/show_bug.cgi?id=1718651.

Looking into `onshow` now.

foolip:
OK, `onshow` is also a Gecko thing, actually disabled:
https://searchfox.org/mozilla-central/rev/ed5c7b4262d4e30656886a45321cbbf4200de349/dom/webidl/EventHandler.webidl#85-86

I'm pretty sure these changes are correct for WebKit now. I'll send other PRs to fix additional things here.

foolip (approved):

PR #11293, opened 2021-06-29
Add Safari versions for api.KeyboardEvent.getModifierState.parameters

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `getModifierState.parameters` member of the `KeyboardEvent` API, based upon commit history and date.

Commit: https://trac.webkit.org/changeset/206725/webkit / https://trac.webkit.org/changeset/206828/webkit / https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/dom/UIEventWithKeyState.cpp#L62

### DIFF
--- a/api/KeyboardEvent.json
+++ b/api/KeyboardEvent.json
@@ -563,10 +563,10 @@
                 "version_added": "35"
               },
               "safari": {
-                "version_added": null
+                "version_added": "10.1"
               },
               "safari_ios": {
-                "version_added": null
+                "version_added": "10.3"
               },
               "samsunginternet_android": {
                 "version_added": "5.0"
@@ -659,10 +659,10 @@
                 "version_added": "35"
               },
               "safari": {
-                "version_added": null
+                "version_added": false
               },
               "safari_ios": {
-                "version_added": null
+                "version_added": false
               },
               "samsunginternet_android": {
                 "version_added": "5.0"
@@ -707,10 +707,10 @@
                 "version_added": true
               },
               "safari": {
-                "version_added": null
+                "version_added": false
               },
               "safari_ios": {
-                "version_added": null
+                "version_added": false
               },
               "samsunginternet_android": {
                 "version_added": true
@@ -755,10 +755,10 @@
                 "version_added": true
               },
               "safari": {
-                "version_added": null
+                "version_added": false
               },
               "safari_ios": {
-                "version_added": null
+                "version_added": false
               },
               "samsunginternet_android": {
                 "version_added": true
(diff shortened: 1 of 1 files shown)


foolip:
The features here are misnamed IMHO, I've filed https://github.com/mdn/browser-compat-data/issues/11295. That doesn't block this PR though.

foolip:
> based upon commit history and date

I don't trust the date alone to always lead to the right answer since the way that Safari does release branches is a bit opaque. I would suggest always also looking at the WebKit trunk version, and https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Configurations/Version.xcconfig?rev=206828 says 603.1.9

That does map to Safari 10.1.

api/KeyboardEvent.json
```diff
@@ -901,10 +901,10 @@
                 "version_added": "35"
               },
               "safari": {
-                "version_added": null
+                "version_added": false
```
foolip, line 904: This "OS" modifier isn't in the FIXME, but it's also not one of the supported strings in the source.

api/KeyboardEvent.json
```diff
@@ -951,10 +951,10 @@
...
```
foolip, line 954: Confirmed by "FIXME: The specification also has Fn, FnLock, Hyper, NumLock, Super, ScrollLock, Symbol, SymbolLock." comment.

api/KeyboardEvent.json
```diff
@@ -1047,10 +1047,10 @@
                 "version_added": true
               },
               "safari": {
-                "version_added": null
+                "version_added": false
```
foolip, line 1050: Confirmed by "FIXME: The specification also has Fn, FnLock, Hyper, NumLock, Super, ScrollLock, Symbol, SymbolLock." comment.

api/KeyboardEvent.json
```diff
@@ -1095,10 +1095,10 @@
                 "version_added": "35"
               },
               "safari": {
-                "version_added": null
+                "version_added": false
```
foolip, line 1098: Confirmed by "FIXME: The specification also has Fn, FnLock, Hyper, NumLock, Super, ScrollLock, Symbol, SymbolLock." comment.

api/KeyboardEvent.json
```diff
@@ -1143,10 +1143,10 @@
                 "version_added": true
               },
               "safari": {
-                "version_added": null
+                "version_added": false
```
foolip, line 1146: Confirmed by "FIXME: The specification also has Fn, FnLock, Hyper, NumLock, Super, ScrollLock, Symbol, SymbolLock." comment.

foolip (approved):
Yay!

PR #11297, opened 2021-06-29
Add Safari versions for api.Element.error_event

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `error_event` member of the `Element` API, based upon manual testing.

Test Code Used: 
```html
<div id="test">
	<div class="controls">
	  <button id="img-error" type="button">Generate image error</button>
	  <img id="bad-img" />
	</div>
</div>

<script>
	var badImg = document.getElementById('bad-img');
	badImg.addEventListener('error', function(event) {
		alert('Error!');
	});

	var imgError = document.getElementById('img-error');
	imgError.addEventListener('click', function() {
	    badImg.setAttribute('src', 'i-dont-exist');
	});
</script>
```

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -2718,10 +2718,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": "3"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "1"
             },
             "samsunginternet_android": {
               "version_added": true
--- a/api/Element.json
+++ b/api/Element.json
@@ -2718,7 +2718,7 @@
               "version_added": null
             },
             "safari": {
-              "version_added": "3"
+              "version_added": "1"
             },
             "safari_ios": {
               "version_added": "1"


foolip (commented):
This seems like an overly generic entry to me, I interpret it as the possibility that an error event can be fired at any kind of element. I'm not sure what to do with that information.

Anyway, let's make the data match the earliest time an error event can be fired. I think `` and `<script>` elements are the oldest cases on the platform.

In https://developer.mozilla.org/en-US/docs/Web/API/HTMLImageElement#browser_compatibility the Safari data for `onerror` is 1. That might not be right, do you have any way of testing?

vinyldarkscratch:
I took another look at this and realized that trusting `textContent` was a bad idea.  I modified the test code to use an `alert` instead, and found that support was around since Safari 1.  Fixed!

foolip:
Excellent, that's what I thought!

foolip (approved):

PR #11298, opened 2021-06-29
Fix redirected mdn_url for <image> functions

teoli2003:
In issue mdn/content#2852  (fixed in mdn/content#6113 ), CSS functions have been moved around.

This PR fixes the `mdn_url` to avoid the need for a redirection.

Note that paint() is in the progress of being moved in mdn/content#6445, this slightly anticipates its landing by already using the new url

### DIFF
--- a/css/types/image.json
+++ b/css/types/image.json
@@ -328,7 +328,7 @@
           "conic-gradient": {
             "__compat": {
               "description": "<code>conic-gradient()</code>",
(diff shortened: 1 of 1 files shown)


vinyldarkscratch (approved):
Looks good to me, thanks!  I quickly verified one of the links, and am confident the other ones are all good too.

PR #11302, opened 2021-06-29
@supports selector() supported by Safari 14.1

gynekolog:
#### Summary
An updated version of the merged and reverted pull request https://github.com/mdn/browser-compat-data/pull/6591

#### Summary from the previous pull request:
The implementation patch for this feature was landed in WebKit.
https://bugs.webkit.org/show_bug.cgi?id=199237

### DIFF
--- a/css/at-rules/supports.json
+++ b/css/at-rules/supports.json
@@ -109,12 +109,10 @@
                 "notes": "See <a href='https://crbug.com/979041'>bug 979041</a>."
               },
               "safari": {
-                "version_added": false,
-                "notes": "See <a href='https://webkit.org/b/199237'>bug 199237</a>"
+                "version_added": "14.1"
               },
               "safari_ios": {
-                "version_added": false,
-                "notes": "See <a href='https://webkit.org/b/199237'>bug 199237</a>"
+                "version_added": "14.1"
               },
               "samsunginternet_android": {
                 "version_added": "13.0"


foolip:
@jdatapple can you confirm/review this?

css/at-rules/supports.json
```diff
@@ -109,12 +109,10 @@
...
               },
               "safari": {
-                "version_added": false,
-                "notes": "See <a href='https://webkit.org/b/199237'>bug 199237</a>"
+                "version_added": "14.1"
               },
               "safari_ios": {
-                "version_added": false,
-                "notes": "See <a href='https://webkit.org/b/199237'>bug 199237</a>"
+                "version_added": "14.1"
```
jdatapple, line 115: This should be 14.5. 

Safari 14.1 looks correct on macOS, but the iOS version is wrong. 

This change would have been previewed available in Safari Technology Preview 114 and shipped with Safari 14.1 on macOS and Safari on iOS 14.5.

jdatapple (changes requested):

css/at-rules/supports.json
foolip: In a situation like (when I'm very sure of the suggested change) this I usually apply the change myself rather than waiting for the contributor to do it.

foolip: I'll go ahead and make the change and merge.

foolip (approved):

foolip:
Thanks @gynekolog!

css/at-rules/supports.json
gynekolog: @foolip Sorry for the incomplete pull request and thank you for the fix.

foolip: No problem, it was a simple fix!

Issue #8947, opened 2021-02-03
html.elements.menuitem - Should mark all properties behind flag


evilpie:
PR #8019 marked the main `<menuitem>` feature as behind a flag. Looking at the MDN page https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menuitem, all other attributes `checked` are not marked as behind a flag.

ddbeck:
This is related to #6509. We never really specified whether flags ought to cascade to descendant features. I think we need to settle #6509 before committing to resolving this _within_ BCD, or whether this should be a BCD consumer problem (e.g., that the table renderer needs to sort ths out).

foolip:
I just came across this issue as well and it goes beyond the question of cascading flags. Parts of this feature are still marked as unconditionally supported in Firefox but are in fact behind a flag ("dom.menuitem.enabled") now:
- https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/contextMenu
- api.GlobalEventHandlers.onshow (no MDN page)
- `label` and `type` in https://developer.mozilla.org/en-US/docs/Web/API/HTMLMenuElement#browser_compatibility
- https://developer.mozilla.org/en-US/docs/Web/API/HTMLMenuItemElement#browser_compatibility
- Probably https://developer.mozilla.org/docs/Web/API/Element/show_event and other things one would find by searching for "contextmenu"

foolip:
This is all complicated by the fact that `<menu>` remains in HTML but is really only for semantics now:
https://html.spec.whatwg.org/multipage/grouping-content.html#the-menu-element

Care has to be taken with `<menu>` and `HTMLMenuElement`.

PR #11303, opened 2021-06-29
Update support for <menu>, <menuitem> and related APIs/events

foolip:
Most of this was only ever implemented in Gecko, and is now behind the
"dom.menuitem.enabled" flag.

A few slices of it remain in HTML and should be supported, namely &lt;menu>
itself, the HTMLMenuElement interface, and the compact attribute. The
latter however is obsolete, so marked as deprecated.

Existing data and notes were preserved as far as possible, even where
this leaves things in an inconsistent and likely incorrect state. This
is because all of this will probably eventually be removed and it would
be a lot of work to determine when exactly each facet of this was
supported in Firefox on different platforms. Assume by default that it
was from 8-85 without a flag, and behind a flag from 85.

Fixes https://github.com/mdn/browser-compat-data/issues/8947.

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -8048,12 +8048,38 @@
             "edge": {
               "version_added": false
             },
-            "firefox": {
-              "version_added": true
-            },
-            "firefox_android": {
-              "version_added": true
-            },
+            "firefox": [
+              {
+                "version_added": "85",
+                "flags": [
+                  {
+                    "type": "preference",
+                    "name": "dom.menuitem.enabled",
+                    "value_to_set": "true"
+                  }
+                ]
+              },
+              {
+                "version_added": "8",
+                "version_removed": "85"
+              }
+            ],
+            "firefox_android": [
+              {
+                "version_added": "85",
+                "flags": [
+                  {
+                    "type": "preference",
+                    "name": "dom.menuitem.enabled",
+                    "value_to_set": "true"
+                  }
+                ]
+              },
+              {
+                "version_added": "8",
+                "version_removed": "85"
+              }
+            ],
             "ie": {
               "version_added": false
             },
@@ -8078,7 +8104,7 @@
           },
           "status": {
             "experimental": false,
-            "standard_track": true,
+            "standard_track": false,
             "deprecated": true
           }
         }
--- a/html/elements/menu.json
+++ b/html/elements/menu.json
@@ -7,42 +7,40 @@
           "spec_url": "https://html.spec.whatwg.org/multipage/grouping-content.html#the-menu-element",
           "support": {
             "chrome": {
-              "version_added": false
+              "version_added": "1"
             },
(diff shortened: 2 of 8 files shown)


foolip:
This was big and complicated and I probably missed something.

api/Element.json
```diff
@@ -8048,12 +8048,38 @@
...
+                "flags": [
+                  {
+                    "type": "preference",
+                    "name": "dom.menuitem.enabled",
+                    "value_to_set": "true"
+                  }
+                ]
+              },
+              {
+                "version_added": "8",
```
ddbeck, line 8063: Where did version 8 come from, anyway?

html/elements/menu.json
```diff
@@ -93,9 +91,9 @@
               }
             },
             "status": {
-              "experimental": true,
-              "standard_track": true,
-              "deprecated": false
+              "experimental": false,
+              "standard_track": false,
+              "deprecated": true
```
ddbeck, line 96: This feature is all-false, deprecated, and non-standard. Should this be deleted instead?

html/elements/menu.json
```diff
@@ -288,8 +366,8 @@
                 }
               },
               "status": {
-                "experimental": true,
-                "standard_track": true,
+                "experimental": false,
+                "standard_track": false,
```
ddbeck, line 370: Another candidate for deletion?

ddbeck (commented):
Just a few questions for you, @foolip. Thank you!

api/Element.json
foolip: I initially trusted the original data and have confirmed support in 8-84 (not in 7 or 85) for `HTMLMenuItemElement` and api.HTMLElement.contextMenu.

It's very hard to confirm the same for events, and actually it looks like the `onshow` properties were added in Firefox 9, not 8. I suspect that the event was supported in Firefox 8 and the event handler property was forgotten but quickly added.

My suggestion is to leave the data consistent but incorrect, if we change how events are represented it truly won't be worth the effort to figure this out, even if we somehow leave a trace leading back to this PR.

html/elements/menu.json
foolip: Yes, the removal is up for review at https://github.com/mdn/browser-compat-data/pull/6955. One of the PRs will have to be rebased depending on which lands first. (I'd prefer to not do the removal here.)

html/elements/menu.json
foolip: Also part of https://github.com/mdn/browser-compat-data/pull/6955.

html/elements/menu.json
ddbeck: Great. Happy to let the removals happen elsewhere.

ddbeck (approved):
Thank you, @foolip. 🎉

Issue #11301, opened 2021-06-29
Add specific bcd for CSS functions of <transform-function> and <filter-function>


teoli2003:
In browser-compat-data, most CSS functions are considered feature and have their own entry. We have two cases:
1. Either they are placed as a sub-feature of the type they "returns". E.g:
- `linear-gradient()` in `css.types.image.gradient.linear-gradient`
- `paint()` in `css.types.image.paint`
- `hsl()` in `css.types.color.hsl`
2. Or they are placed as a sub-feature of the property they can be used with. E.g:
- `fit-content()` in `css.properties.grid-template-rows.fit-content` (and others)
- `repeat() `in` css.properties.grid-template-rows.repeat`
- `minmax()` in `css.properties.grid-template-columns.minmax`
and many more.

There are a few exceptions. 

The functions of `<transform-function>`, like `translate()` or `skewX()` are not defined as subfeatures but are covered either by the top level `css.types.transform-function` or by `css.types.transform-function.3d` (if they are 2d, respectively 3d functions).

The functiions of `<filter-function>`, like `sepia()`, `blur()` or  `brightness()` are also not defined as subfeatures and covered by the top level `css.properties.filter` and `css.properties.backdrop-filter` (as `backdrop-filter` has been recently added, the `filter-function` type was needed, and is a recent addition to the spec, not yet reflected on MDN content.

We should "orthogonalize" these two set of functions and my plan is to:
1. List all the `<transform-function>` as subfeatures of `css.types.transform-function` (we have the bcd data, there are `mdn_url` and `spec_url `to add to each of these subfeatures); remove `css.types.transform-function.3d` that is useless and its data distributed to each relevant transform functions.
2. Create `<filter-function>` with the different functions as subfeature (and their bcd indicating when they have been added).
3. Make sure the MDN web docs pages point correctly to these features (trivial for `<transform-function>`s, a bit more work for `<filter-function>`s, as we need to make them work for both `filter` and `backdrop-filter`).

teoli2003:
Both PR got merged. This is done!

PR #11304, opened 2021-06-29
Populate css.types.transform-function with functions

teoli2003:
This is the first part of #11301.

It deals with the CSS type `<transform-function>` and the CSS functions it defines.

What Is done in this PR:

1. I added entries for the different transform functions, including the `mdn_url `and `bcd_url` clauses. (I had to update 1 mdn url in mdn/content#6453 ). The `support` data is taken from the original data (`css.types.transform-function` for 2d functions, `css.types.transform-function.3d` for 3d functions).
2. I adapted the notes so only pertinent ones appear inside each CSS function.
3. I removed `css.types.transform-function.3d` that is of no more use now that each CSS function has its own entry. (Does this need an entry in the release notes?)
4. I updated the `spec_url` entry of `css.types.transform-function` as [CSS Transforms Module Level 1](https://drafts.csswg.org/css-transforms/#transform-functions) is not superseded by [CSS Transform Level 2](https://drafts.csswg.org/css-transforms-2/#transform-functions). The latter only describes the new features. This was not necessary for the subfeatures as they are described either in the first one, either in the latter (at least for the moment).

Once this lands, I'll update the mdn pages to link to the right keys added here.

### DIFF
--- a/css/types/transform-function.json
+++ b/css/types/transform-function.json
@@ -5,7 +5,10 @@
         "__compat": {
           "description": "<code>&lt;transform-function&gt;</code>",
           "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/transform-function",
-          "spec_url": "https://drafts.csswg.org/css-transforms/#transform-functions",
+          "spec_url": [
+            "https://drafts.csswg.org/css-transforms/#transform-functions",
+            "https://drafts.csswg.org/css-transforms-2/#transform-functions"
+          ],
           "support": {
             "chrome": {
               "version_added": "1"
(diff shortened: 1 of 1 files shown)


ddbeck:
I'm not sure I got enough sleep last night to understand the underlying issue. I'll come back to that. 😴

> Once this lands, I'll update the mdn pages to link to the right keys added here.

We recently started using the [needs content update](https://github.com/mdn/browser-compat-data/blob/main/docs/workflow.md#needs-content-update-) label to flag things that affect content like this. The summary is that we should hold off merging changes in BCD until there's a PR open on mdn/content and linked here (to avoid getting into a state where changes get into BCD and never get fixed in content—I'm not worried anyone will do this intentionally, but things slip through the cracks).

teoli2003:
Thanks for the info about the flag! I will create the content PR and link it here.

Note that I can't add labels to PR, even those that I created, Github's permission required for applying label is `Triage` and not the `Read` permissions, default for users.

teoli2003:
Ping @ddbeck

ddbeck (commented):
I'm sorry this took so long to review, @teoli2003. It was a bit big and I kept putting it off. 😬

(I retitled the PR to capture what I think this does a little better. Feel free to revise, if I've misunderstood.)

I'm happy with expanding the features listed here. Seems really sensible. I do have one question, however: where did the versions come from for the functions themselves?

For example, some are supported from Chrome 1 and others from Chrome 12 and I don't think there's an explicit explanation for this and I'm guessing you didn't test every one of these manually to find out.

teoli2003:
No worry for the delay.

I took it from the `transform` CSS properties: 2D functions were added when the prefixed versions was added (Chrome 1) and 3D ones in Chrome 12.

Screenshot 2021-08-03 at 14 39 26

ddbeck (approved):
OK, makes sense. Thank you! 🎉

PR #11306, opened 2021-06-29
Mark api.GlobalEventHandlers.onloadend as non-standard and link Gecko bug

foolip:


foolip:
@ddbeck would you accept a PR that just removes this entry instead? It's still in Gecko, but it's totally pointless for anyone to know about this property AFAICT, unless you really enjoy `document.onloadend = function() { ...}` + `document.dispatchEvent(new Event('loadend'))` as a way to communicate with yourself in Firefox only :)

ddbeck:
Are [the docs for this](https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onloadend) a complete fabrication or something? Does it not actually do anything for images?

foolip:
Huh, it's a thing. I put the example code into https://software.hixie.ch/utilities/js/live-dom-viewer/?saved=9457 and Firefox does fire "loadstart" and "loadend" events for images, which Chrome and Safari don't.

These are non-standard events and this `onloadend` entry is the only thing testifying to their existence in BCD. If we want to represent support for this in Firefox, I think `loadstart_event` and `loadend_event` under api.HTMLImageElement might be better.

I found https://bugzilla.mozilla.org/show_bug.cgi?id=1574487 which is the better bug to link to here.

sideshowbarker:
Rightly I guess this should be marked non-standard and deprecated MDN as well — but I guess we don’t need to block this BCD PR on that happening first.

PR #11308, opened 2021-06-29
Update Chromium versions for init_delegatesFocus_parameter entry

foolip:
This goes back to when Shadow DOM v1 was first enabled in Chromium:
https://storage.googleapis.com/chromium-find-releases-static/972.html#97235496c3dee67efb24f9a7893232df493dcd1e

Thus the versions match the parent feature, attachShadow.


vinyldarkscratch (approved):
Thanks for tracking this down, this is looking good to me!

PR #11316, opened 2021-06-30
Add Safari versions for NotificationEvent API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `NotificationEvent` API, based upon results from WPT.fyi.

Test Used: https://wpt.fyi/results/notifications/idlharness.https.any.serviceworker.html?label=experimental&label=master&aligned

### DIFF
--- a/api/NotificationEvent.json
+++ b/api/NotificationEvent.json
@@ -31,10 +31,10 @@
             "version_added": "37"
           },
           "safari": {
-            "version_added": null
+            "version_added": false
           },
           "safari_ios": {
-            "version_added": null
+            "version_added": false
           },
           "samsunginternet_android": {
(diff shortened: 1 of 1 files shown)


foolip:
Note that https://wpt.fyi/results/notifications/idlharness.https.any.serviceworker.html?label=experimental&label=master&aligned (idlharness.js) is very surface-level tests and no better than https://mdn-bcd-collector.appspot.com/tests/api/NotificationEvent, in fact a bit worse for our purposes since it tests more than existence and could fail in cases BCD should say something is supported.

foolip (approved):
Confirmed, no `NotificationEvent` in WebKit source.

PR #11317, opened 2021-06-30
Remove long_to_double from MouseEvent API

vinyldarkscratch:
This PR removes the various `long_to_double` entries from the MouseEvent API.  Overall, the difference between a `long` and a `double` is more floating point accuracy, but I can't imagine a scenario where a web developer will require such high levels of precision, not to mention this is near impossible to test, and only really observable in the WebIDL.

### DIFF
--- a/api/MouseEvent.json
+++ b/api/MouseEvent.json
@@ -101,54 +101,6 @@
             "deprecated": false
           }
         },
-        "long_to_double": {
-          "__compat": {
-            "description": "Redefined <code>mouseEventInit</code> fields from <code>long</code> to <code>double</code>",
-            "support": {
-              "chrome": {
-                "version_added": "56"
-              },
-              "chrome_android": {
-                "version_added": "56"
-              },
-              "edge": {
-                "version_added": "≤79"
-              },
-              "firefox": {
-                "version_added": null
-              },
-              "firefox_android": {
-                "version_added": null
-              },
-              "ie": {
-                "version_added": false
-              },
-              "opera": {
-                "version_added": null
-              },
-              "opera_android": {
-                "version_added": null
-              },
-              "safari": {
-                "version_added": null
-              },
-              "safari_ios": {
-                "version_added": null
-              },
-              "samsunginternet_android": {
-                "version_added": "6.0"
-              },
-              "webview_android": {
-                "version_added": "56"
-              }
-            },
-            "status": {
-              "experimental": false,
-              "standard_track": true,
-              "deprecated": false
-            }
-          }
-        },
         "region_support": {
           "__compat": {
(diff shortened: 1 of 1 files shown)


foolip:
These entries are very poorly described, but are about this change I think:
https://chromestatus.com/feature/5117956774166528

(The versions don't match, so I'm not sure...)

https://chromestatus.com/feature/6169687914184704 is a similar change.

vinyldarkscratch:
I'm now realizing that I was seriously confusing `long` with `float`, and was thinking that `long` was a decimal format...oops.  As such, I think we should keep it, but maybe give it a better name?  (And maybe de-duplicate the entry, so we only have one `long_to_double`?)

foolip:
Yeah, I think a single subfeature per interface where some members changed from always integer to sometimes floating point would be good. "fractional coordinates" is actually a pretty good description I think.

I see that https://chromestatus.com/feature/6169687914184704 says:
> only browser-fired events that will change are PointerEvents. MouseEvents will continue to have integral coordinates for backward compatibility

If that's still true, then these entries on `MouseEvent` seems kind of pointless.

foolip:
This all originates from the wiki table:
https://web.archive.org/web/20180421035120/https://developer.mozilla.org/en-US/docs/Web/API/MouseEvent#Browser_compatibility

I'm pretty sure that the single "MouseEvent.screenX, MouseEvent.screenY, MouseEvent.clientX, and MouseEvent.Y are double instead of long." feature there was based on https://chromestatus.com/feature/6169687914184704 and was then turned into multiple features when it was ported to BCD in https://github.com/mdn/browser-compat-data/pull/1883.

I've just tested in Chrome 91 to confirm that `MouseEvent` coordinates are still integers, while `PointerEvent` has fractional coordinates. In fact, even though `MouseEvent` does have some `double` members in Chromium's IDL, it doesn't seem possible to actually observe that, because those same getters floor the returned value to integers:
https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/events/mouse_event.h;l=138-146;drc=c629498025fa2b88f251047c22b06ca840c05f5e

A further nail in the coffin is that the spec uses `long` for these members:
https://w3c.github.io/uievents/#idl-mouseevent

foolip (approved):

foolip:
@ddbeck let me know if you need any more info for the release notes here, or if it seems suspect when you try to explain it.

foolip:
Note also that the real change/feature is already represented here:
https://developer.mozilla.org/en-US/docs/Web/API/PointerEvent/pointerType#browser_compatibility

foolip:
And I've sent https://github.com/mdn/browser-compat-data/pull/11327 to link bugs for that entry.

PR #11318, opened 2021-06-30
Add Safari versions for RTCTrackEventInit API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `RTCTrackEventInit` API, based upon commit history and date.

Commit: https://trac.webkit.org/browser/webkit/tags/Safari-604.2.4/Source/WebCore/Modules/mediastream/RTCTrackEvent.idl

### DIFF
--- a/api/RTCTrackEventInit.json
+++ b/api/RTCTrackEventInit.json
@@ -30,10 +30,10 @@
             "version_added": "43"
           },
           "safari": {
-            "version_added": null
+            "version_added": "11"
           },
           "safari_ios": {
-            "version_added": null
+            "version_added": "11"
           },
           "samsunginternet_android": {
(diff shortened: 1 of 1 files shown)


foolip:
@vinyldarkscratch if you've tracked down a WebKit commit where something was added, linking to it would be very helpful for verifying. With just https://trac.webkit.org/browser/webkit/tags/Safari-604.2.4/Source/WebCore/Modules/mediastream/RTCTrackEvent.idl it's not clear if it might also have been in some previous release.

foolip (approved):
This now matches RTCTrackEvent.json apart from `mdn_url` and `spec_url` which makes me confident it's correct without looking closer.

PR #11321, opened 2021-06-30
Add Safari versions for MediaStreamEvent API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `MediaStreamEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.8).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/MediaStreamEvent

### DIFF
--- a/api/MediaStreamEvent.json
+++ b/api/MediaStreamEvent.json
@@ -29,10 +29,12 @@
             "version_added": true
           },
           "safari": {
-            "version_added": false
+            "version_added": "11",
+            "version_removed": "12"
           },
           "safari_ios": {
-            "version_added": false
+            "version_added": "11",
+            "version_removed": "12"
           },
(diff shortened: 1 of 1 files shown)


foolip:
Searching for "MediaStreamEvent" in WebKit source shows it changelogs a fair bit, revealing it was once there. It was removed in https://trac.webkit.org/changeset/235484/webkit

Using https://mdn-bcd-collector.appspot.com/tests/api/MediaStreamEvent I can confirm this was in 11.1, but I'm not sure exactly when it was added and removed. The time it was added can't be determined from source code because of the runtime flags involved here.

Can you add `version_added` and `version_removed`?

foolip (approved):

PR #11323, opened 2021-06-30
Remove RequestDestination enum from BCD

vinyldarkscratch:
This PR removes the `RequestDestination` enum from BCD.  We don't have any data for the individual values to begin with, plus we don't want to be keeping enum data in BCD anyways.


foolip (approved):
LGTM but this should be merged together with the content PR.

vinyldarkscratch:
Content PR merged!  Merging this one myself since it has been approved.

PR #11324, opened 2021-06-30
Add Safari versions and bugs for preventScroll_option entries

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `focus.preventScroll_option` member of the `SVGElement` API, based upon information in a tracking bug.

Tracking Bug: https://webkit.org/b/178583

### DIFF
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -1059,10 +1059,12 @@
                 "version_added": "47"
               },
               "safari": {
-                "version_added": false
+                "version_added": false,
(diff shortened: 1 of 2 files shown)


foolip (approved):
I went ahead and linked the same bug for the other `preventScroll` entry.

foolip:
I've sent https://github.com/mdn/browser-compat-data/pull/11328 to touch up these entries some more.

PR #11328, opened 2021-06-30
Rename and complete the entries for focus({preventScroll}) parameters

foolip:
The naming now matches the guideline:
https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#parameters-and-parameter-object-features

Chromium support for this parameter for SVGElement came when the focus()
methods of HTMLElement and SVGElement were consolidated into
HTMLOrForeignElement. Before that only the HTMLElement variant had the
FocusOptions dictionary as an argument:
https://chromium.googlesource.com/chromium/src/+/5f506297541b9c0a041701401eeb133a16619bf9
https://storage.googleapis.com/chromium-find-releases-static/5f5.html#5f506297541b9c0a041701401eeb133a16619bf9

The other data is mirrored and IE is just assumed to be false.

### DIFF
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -1030,9 +1030,9 @@
             "deprecated": false
           }
         },
-        "preventScroll_option": {
+        "options_preventScroll_parameter": {
           "__compat": {
-            "description": "<code>preventScroll</code> Boolean option",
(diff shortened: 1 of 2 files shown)


vinyldarkscratch (approved):
Thanks @foolip, this is looking good to me!

PR #11332, opened 2021-06-30
Part 1 of 2: copy feature policy to permissions policy.

jpmedley:
**PLEASE READ CAREFULLY. A unique problem requires a unique approach, which is described below.**

The rename of 'Feature Policy' to 'Permissions Policy' requires that BCD be moved as well. Regardless of whether BCD is changed before MDN content or vice-versa, there will be a stage where BCD keys in MDN documents will point to data that isn't there. To get around this, I have duplicated the data. In other words, when this is deployed there will be data under the keys `Feature-Policy` and `Permissions-Policy`. This is by design. When the new MDN pages are deployed, I'll come back remove the data recorded under `Feature-Policy`. 

https://github.com/mdn/content/issues/1831

### DIFF
--- a/http/headers/permissions-policy.json
+++ b/http/headers/permissions-policy.json
@@ -1,20 +1,41 @@
 {
   "http": {
     "headers": {
-      "Feature-Policy": {
+      "Permissions-Policy": {
         "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/HTTP/Headers/Feature-Policy",
+          "mdn_url": "https://developer.mozilla.org/docs/Web/HTTP/Headers/Permissions-Policy",
           "spec_url": "https://w3c.github.io/webappsec-permissions-policy/#permissions-policy-http-header-field",
           "support": {
-            "chrome": {
-              "version_added": "60"
-            },
-            "chrome_android": {
-              "version_added": "60"
-            },
-            "edge": {
-              "version_added": "79"
-            },
+            "chrome": [
+              {
+                "version_added": "88"
+              },
+              {
+                "alternative_name": "Feature-Policy",
+                "version_added": "60",
+                "version_removed": "87"
+              }
+            ],
+            "chrome_android": [
+              {
+                "version_added": "88"
+              },
+              {
+                "alternative_name": "Feature-Policy",
+                "version_added": "60",
+                "version_removed": "87"
+              }
+            ],
+            "edge": [
+              {
+                "version_added": "88"
+              },
+              {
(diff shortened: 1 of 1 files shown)


Lukepearshotmail:
Feature-Policy to permissions
Polich

jpmedley:
What lead me to this was the large number of pages that would contain missing BCD and the unknown amount of time needed to correct it. I'm working as quickly as I can to keep that time short, but I expect the review process to be tedious and somewhat drawn out. To help, I've already split the affected pages into 4 separate PRs which I recommend merging in close proximity. The PR to remove the duplicate content would be removed as soon as the updated files are deployed.

I would suggest that any changes to this data go exclusively to the replacement key: `permissions-policy`.

If you have a better idea, I'm all ears.

jpmedley:
[I've opened four PRs](https://github.com/mdn/content/pulls/jpmedley) to replace 'feature policy' with 'permissions policy'. I believe this is all of the content that needs to be updated.

http/headers/permissions-policy.json
```diff
@@ -0,0 +1,1682 @@
...
+          "spec_url": "https://w3c.github.io/webappsec-permissions-policy/#permissions-policy-http-header-field",
+          "support": {
+            "chrome": [
+              {
+                "version_added": "88"
+              },
+              {
+                "alternative_name": "Feature-Policy",
+                "version_added": "60",
+                "version_removed": "87"
```
foolip, line 16: Was support for the `Feature-Policy` header really removed, one release before its replacement even?

hamishwillee:
Strictly speaking, a spec rename should be accomplished by renaming the feature-policy json to permission policy, updating the content, and adding an alternative name. However I don't think BCD can properly represent the relationship using alternative names because these can't be fully qualified - as a result it might become quite messy to indicate the things that do and do not have an alternative name as the spec evolves. 

Proposal: We treat `Feature-Policy` and `Permission-Policy` as separate technologies at the BCD level. Practically speaking this means that this PR would be modified to:
- have start dates based on when `Permission-Policy` actually appeared
- also modify `Feature-Policy` with end-dates for support on any browsers that stopped using this name and do not support it as an alternative name.

If we treat them as separate implementations then the versioning can be presented very clearly. The only issue we'd have is whether we "try" to present alternative names at all. I'd say "not".

@ddbeck FYI

ddbeck:
@jpmedley I'm going to close this PR, since the corresponding content PRs have also been closed. Feel free to open a new one with the additions of permissions policy, but let's treat feature policy and permission policy as two closely related but distinct features. Thank you!

PR #9800, opened 2021-04-08
Update FormData compat data for Safari

foolip:
Most of the API was added in WebKit trunk version 605.1.6:
https://trac.webkit.org/changeset/221839/webkit
https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Configurations/Version.xcconfig?rev=221839

This maps to Safari 11.1 / iOS 11.3.

Worker exposure happened later in WebKit trunk version 609.1.11:
https://trac.webkit.org/changeset/252349/webkit
https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Configurations/Version.xcconfig?rev=252349

That maps to Safari 13.1 / iOS 13.4.

https://mdn-bcd-collector.appspot.com/tests/api/FormData was also tested
in Safari 11 + 11.1 and 13 + 13.1 on Sauce Labs to confirm.

Follow-up to https://github.com/mdn/browser-compat-data/issues/9425 + https://github.com/mdn/browser-compat-data/pull/9722.

PR #11333, opened 2021-06-30
Simplify engine version for Safari 14.1

queengooborg:
This PR simplifies the engine version for Safari 14.1 in our data, stripping out the micro and nano version numbers.  Not only does this level of information not really matter to us, but `compareVersions` doesn't like them, and the micro/nano versions have already changed in 14.1.1.

### DIFF
--- a/browsers/safari.json
+++ b/browsers/safari.json
@@ -173,7 +173,7 @@
           "release_notes": "https://developer.apple.com/documentation/safari-release-notes/safari-14_1-release-notes",
           "status": "current",
           "engine": "WebKit",
-          "engine_version": "611.1.21.161.3"
+          "engine_version": "611.1.21"
         },
         "15": {
           "release_notes": "https://developer.apple.com/documentation/safari-release-notes/safari-15-beta-release-notes",
--- a/browsers/safari_ios.json
+++ b/browsers/safari_ios.json
@@ -157,7 +157,7 @@
           "release_notes": "https://developer.apple.com/documentation/safari-release-notes/safari-14_1-release-notes",
           "status": "current",
           "engine": "WebKit",
-          "engine_version": "611.1.21.161.6"
+          "engine_version": "611.1.21"
         },
         "15": {
           "release_notes": "https://developer.apple.com/documentation/safari-release-notes/safari-15-beta-release-notes",


foolip:
> the micro/nano versions have already changed in 14.1.1

This is true, but not really the criteria we should use. The bundle version I see is 16611.2.7.1.4, so that would be WebKit 611.2.7.1.4. In other words, the minor and patch versions have also changed, but we're not going to remove those parts.

foolip (approved):
This makes sense to me, but I'd like to hear from @jdatapple too.

foolip:
Ping @jdatapple, @jensimmons or @gsnedders for review

jdatapple (commented):
Without an understanding of how MDN makes use of these engine version numbers, it's not clear how to provide useful feedback on these changes.

foolip:
These engine versions never show up on MDN or any rendering of compat data that I know, rather it's used by BCD contributors.

When I'm unable to pin down the exact version by testing on BrowserStack or Sauce, I use the WebKit versions numbers to make an educated guess of which Safari or iOS version a feature was shipped in if I know the WebKit revision that introduced it and I'm confident there's no compile or runtime flags involved. https://github.com/mdn/browser-compat-data/pull/9625 is an example of this, and [this search](https://github.com/mdn/browser-compat-data/pulls?q=is%3Apr+xcconfig) has many more.

The useful precision for this purpose is only as many parts of the WebKit version that change on trunk, which has been 3 for a long time now.

gsnedders:
> The useful precision for this purpose is only as many parts of the WebKit version that change on trunk, which has been 3 for a long time now.

Note that this isn't comparable to what we do for Blink, where we only ever have the major version despite the x.y.z changing on trunk.

At the moment we're overly specific for WebKit, given the value we give isn't necessarily constant for all bug fix releases.

foolip:
That's true, but there does seem to be a practical use for the 3rd part of the version. In https://github.com/mdn/browser-compat-data/pull/9800 I ended up comparing 609.1.11 to 609.1.20 to conclude that https://trac.webkit.org/changeset/252349/webkit was in Safari 131. If we had recorded only 609 or 609.1 in BCD, I'd be less confident.

Now, given more information about how release branching for Safari works, fewer or even *no* WebKit version components could work. What I'm really trying to guess in these cases is the branch points, the point before which all commits are going to be included in the release unless reverted. (I know it's not infallible, but sometimes the only option I can see to pick a version number.)

queengooborg:
Where are we at with this PR?  Is this something that needs more discussion, or can we move forward with this?

ashnicdill:
Does this involve a reply I sent to someone involving this information? I
didn't do anything with safari. I have other emails from GitHub about
tracking bugs on Chrome that I didn't do. My husband is very smart, knows
how to build apps, and I need to figure out how to make it stop. We have
been separated for months. He has never had permission to use my phone and
I do believe has access to 100% of it. I can't even change some settings. I
bought this new Motorola because he did the same thing to my last phone.
Can someone help me? Please?

foolip:
I'm going to go ahead and merge this. While the WebKit version number isn't a perfect mapping to WebKit trunk, it's useful enough in practice. We definitely don't need 5 version components. Maybe we could remove even more, but at least 3 is the most we could have use for in BCD.

PR #11335, opened 2021-06-30
Bump NodeJS requirement for contributors to v14

### LINKED ISSUE #11311: Drop NodeJS v10 (and maybe even more?)
Since NodeJS v10 is no longer in LTS, I'd say that it's time for us to drop support for it.  However, since BCD consumers are no longer bound to any NodeJS versions thanks to our bundled releases, I'm thinking that we could raise the minimum version even higher, dropping v12 as well.  This NodeJS requirement only affects the BCD contributors, and already we're not running tests for anything lower than NodeJS v14 in GitHub Actions.  Additionally, by raising the NodeJS requirement higher, we'll be able to upgrade our linter/test dependencies easier.

The big questions to ask are:
- What NodeJS versions are contributors using now?
- Are they willing to upgrade to the latest Active LTS or current NodeJS versions?

What do you think. @ddbeck?

queengooborg:
This PR fixes #11311 by bumping the minimum NodeJS requirement for contributors from v10 to v14.  We're already not performing tests on NodeJS versions earlier than v14 (active LTS), and this requirement will not affect BCD consumers, only contributors.

### DIFF
--- a/package.json
+++ b/package.json
@@ -5,7 +5,7 @@
   "main": "index.js",
   "types": "index.d.ts",
   "engines": {
-    "node": ">=10.0.0"
+    "node": ">=14.0.0"
   },
   "repository": {
     "type": "git",


lucalves (approved):

foolip (approved):
https://github.com/mdn/browser-compat-data/blob/main/.github/workflows/node.js.yml is testing Node.js 14 so this makes sense.

The "engines" field isn't included in the released package:

https://github.com/mdn/browser-compat-data/blob/30c8c2521a690061ce8a25be48fe111558f4e6a1/scripts/release-build.js#L39-L64

PR #11338, opened 2021-07-01
Add Safari versions for api.EventSource.EventSource.cors_support

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `EventSource.cors_support` member of the `EventSource` API, based upon commit history and date.

Commit: https://trac.webkit.org/changeset/138083/webkit

### DIFF
--- a/api/EventSource.json
+++ b/api/EventSource.json
@@ -126,10 +126,10 @@
                 "version_added": "12"
               },
               "safari": {
-                "version_added": null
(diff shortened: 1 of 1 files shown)


foolip (approved):
https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Configurations/Version.xcconfig?rev=138083 says 537.22 👍

PR #11340, opened 2021-07-01
Link Chrome bug for RTCPeerConnection.addStream() and related APIs

foolip:


api/RTCPeerConnection.json
```diff
@@ -237,13 +237,14 @@
       "addStream": {
         "__compat": {
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/RTCPeerConnection/addStream",
-          "spec_url": "https://w3c.github.io/webrtc-pc/#legacy-interface-extensions",
```
foolip, line 240: `addStream` isn't actually defined in this section, there's just a note about how to polyfill it.

ddbeck (approved):
Thank you! 🎉

PR #11354, opened 2021-07-02
Fix multi display mode css

teoli2003:
These properties had no bcd directly on them, but only on their subfeatures. I added it (taking the earliest one). That way mdn_url and spec_url are at the right level.

This is similar to what we did in #11214.

(No mdn content will be broken by this change, it will be improved and we will be able to source spec_url through {{Specifications}})

### DIFF
--- a/css/properties/align-content.json
+++ b/css/properties/align-content.json
@@ -2,60 +2,129 @@
   "css": {
     "properties": {
       "align-content": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/align-content",
+          "spec_url": [
+            "https://drafts.csswg.org/css-align/#align-justify-content",
+            "https://drafts.csswg.org/css-flexbox/#align-content-property"
+          ],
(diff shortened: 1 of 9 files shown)


teoli2003:
@ddeck, I think we can remove the needs content update label (I can't myself)

ddbeck:
Related to https://github.com/mdn/browser-compat-data/issues/6175.

ddbeck:
Also, since this seems to be addressing https://github.com/mdn/browser-compat-data/issues/6175, I'd like to see that issue updated to explain what you're planning to do and what's out of scope for the work you've decided to take on, in that issue.

ddbeck:
@teoli2003 Not trying to add any pressure here, but I was wondering: is this something you plan to return to?

ddbeck:
This has gotten rather stale and conflicted, so I'm going to close this PR. See https://github.com/mdn/browser-compat-data/issues/13804 for guidance on how we might move forward on this sort of thing.

PR #11360, opened 2021-07-02
Add .github/workflows/ping-other-repos.yml

sideshowbarker:
For every push to the repo, the `.github/workflows/ping-other-repos.yml` file added in this change creates a GitHub `repository_dispatch` event to send a ping to the https://github.com/w3c/mdn-spec-links/ repo (where it triggers the build of the contents of that repo).

Along with https://github.com/w3c/mdn-spec-links/, other repos can be added to this later — any repos that want to have some action triggered each time a push to the BCD repo occurs.

https://docs.github.com/en/rest/reference/repos#create-a-repository-dispatch-event has the documentation for the API that this relies on.

---

@ddbeck, @Elchi3 — for this to actually work, a new `secrets.SIDESHOWBARKER` Actions repository secret needs to be added to the repo using the form at https://github.com/mdn/browser-compat-data/settings/secrets/actions/new. I can give you the value of the secret in a Matrix direct message.


.github/workflows/ping-other-repos.yml
```diff
@@ -0,0 +1,25 @@
+name: Remote Dispatch Action Initiator
```
ddbeck, line 1: I like how the file name described this and thought it would be nice if it would be visible in the GitHub UI

```suggestion
name: Ping other repos
```

.github/workflows/ping-other-repos.yml
```diff
@@ -0,0 +1,25 @@
...
+
+on:
+  push:
+  repository_dispatch:
```
ddbeck, line 5: If I understand this correctly, this workflow will run when another repository dispatches _back_ to mdn/browser-compat-data. My guess here is that this was so you could test things, but I'm not sure we need to do this in the final workflow. At very least, it would never actually happen, unless we also provide a PAT to every repository that we dispatch _to_.

.github/workflows/ping-other-repos.yml
```diff
@@ -0,0 +1,25 @@
...
+    runs-on: ubuntu-latest
+    steps:
+      - name: PING - Dispatch initiating repository event
+        if: github.event.action != 'pong'
+        run: |
+          curl -X POST https://api.github.com/repos/w3c/mdn-spec-links/dispatches \
+          -H 'Accept: application/vnd.github.everest-preview+json' \
+          -u ${{ secrets.SIDESHOWBARKER }} \
+          --data '{"event_type": "ping", "client_payload": { "repository": "'"$GITHUB_REPOSITORY"'" }}'
+      - name: ACK - Acknowledge pong from remote repository
```
ddbeck, line 18: As mentioned up top, I think we can drop this step?

.github/workflows/ping-other-repos.yml
```diff
@@ -0,0 +1,25 @@
+name: Remote Dispatch Action Initiator
+
+on:
+  push:
```
ddbeck, line 4: Do you want to be pinged on every push to every branch on mdn/browser-compat-data? Or just to `main`?

.github/workflows/ping-other-repos.yml
```diff
@@ -0,0 +1,25 @@
...
+  repository_dispatch:
+
+jobs:
+  ping-pong:
+    runs-on: ubuntu-latest
+    steps:
+      - name: PING - Dispatch initiating repository event
+        if: github.event.action != 'pong'
```
ddbeck, line 12: Can we guard against the secret being unset as well? Otherwise, every fork will attempt to complete this step and fail, because it doesn't have the right permissions.

.github/workflows/ping-other-repos.yml
```diff
@@ -0,0 +1,25 @@
+name: Remote Dispatch Action Initiator
+
+on:
+  push:
+  repository_dispatch:
+
+jobs:
+  ping-pong:
+    runs-on: ubuntu-latest
```
ddbeck, line 9: As an alternative (or in addition) to guarding against unset secrets, we could add an `if` here to prevent running the workflow on forks:

(I think this is right—I haven't tested it.)

```suggestion
    runs-on: ubuntu-latest
    if: ${{ github.respository == 'mdn/browser-compat-data' }}
```

.github/workflows/ping-other-repos.yml
```diff
@@ -0,0 +1,25 @@
...
+    steps:
+      - name: PING - Dispatch initiating repository event
+        if: github.event.action != 'pong'
+        run: |
```
ddbeck, line 13: I'm not exactly sure if we should—or if you've already considered this—but maybe we ought to use an off-the-shelf step for this? Something like https://github.com/marketplace/actions/repository-dispatch would be a little more readable, at least.

ddbeck (commented):
Interesting! I'm amenable to this, though there are some details I wonder about (see line comments for more). I had thought about building some tools that reacted to changes in BCD this way, but never really got beyond the "maybe repository dispatch will do this?" phase.

.github/workflows/ping-other-repos.yml
sideshowbarker: > If I understand this correctly, this workflow will run when another repository dispatches _back_ to mdn/browser-compat-data. My guess here is that this was so you could test things, but I'm not sure we need to do this in the final workflow.

Yeah, agreed, it’s not necessary — now removed

.github/workflows/ping-other-repos.yml
sideshowbarker: > Do you want to be pinged on every push to every branch on mdn/browser-compat-data? Or just to `main`?

Good point — there’s no need for this to run on pushes to branches, so I changed to run only for pushes to the main branch

.github/workflows/ping-other-repos.yml
sideshowbarker: > Can we guard against the secret being unset as well? Otherwise, every fork will attempt to complete this step and fail, because it doesn't have the right permissions.

Good point. Added a conditional for it.

And note that per https://github.com/actions/runner/issues/520:
> Unfortunately, GitHub Actions currently does not support checking the presence of a secret in a conditional

So, as suggested in https://github.com/actions/runner/issues/520#issuecomment-860043020, I did it by copying the secret to an environment variable first, then checking if that environment variable is set.

.github/workflows/ping-other-repos.yml
sideshowbarker: > As mentioned up top, I think we can drop this step?

Yep, dropped the ACK

.github/workflows/ping-other-repos.yml
sideshowbarker: > I'm not exactly sure if we should—or if you've already considered this—but maybe we ought to use an off-the-shelf step for this? Something like [github.com/marketplace/actions/repository-dispatch](https://github.com/marketplace/actions/repository-dispatch) would be a little more readable, at least.

I’m looking into that right now

.github/workflows/ping-other-repos.yml
sideshowbarker: > Can we guard against the secret being unset as well? Otherwise, every fork will attempt to complete this step and fail, because it doesn't have the right permissions.

On second thought, looking back at the source, it already has this:
```yaml
if: ${{ github.respository == 'mdn/browser-compat-data' }}
```
So, would that not be sufficient for ensuring it doesn’t run for forks?

.github/workflows/ping-other-repos.yml
sideshowbarker: > > I'm not exactly sure if we should—or if you've already considered this—but maybe we ought to use an off-the-shelf step for this? Something like [github.com/marketplace/actions/repository-dispatch](https://github.com/marketplace/actions/repository-dispatch) would be a little more readable, at least.
> 
> I’m looking into that right now

OK — thanks much for suggesting; it’s definitely way more readable, so I’ve switched the source to using that intead

.github/workflows/ping-other-repos.yml
ddbeck: In that case:

```suggestion
```

.github/workflows/ping-other-repos.yml
```diff
@@ -0,0 +1,19 @@
...
+  push:
+    branches: [main]
+  repository_dispatch:
+
+jobs:
+  ping-pong:
+    runs-on: ubuntu-latest
+    env:
+      SIDESHOWBARKER_TOKEN: ${{ secrets.SIDESHOWBARKER }}
+    if: ${{ github.respository == 'mdn/browser-compat-data' && env.SIDESHOWBARKER_TOKEN != '' }}
```
ddbeck, line 13: Oh, I see. It's probably unnecessary to do both guards, right? I mean, maybe there's a case to guard against unset secrets at the level of each step, if we wanted to add more dispatches. But let's cross that bridge when we get to it.

```suggestion
    if: ${{ github.respository == 'mdn/browser-compat-data' }}
```

.github/workflows/ping-other-repos.yml
```diff
@@ -0,0 +1,19 @@
...
+    steps:
+      - name: Repository Dispatch
+        uses: peter-evans/repository-dispatch@v1
+        with:
+          token: ${{ secrets.SIDESHOWBARKER }}
+          event-type: ping
```
ddbeck, line 19: Maybe we ought to still have a payload? Something like…

```suggestion
          event-type: ping
          client-payload: '{"repository": "${{ github.repository }}", "ref": "${{ github.ref }}", "sha": "${{ github.sha }}"}'
```

.github/workflows/ping-other-repos.yml
```diff
@@ -0,0 +1,19 @@
+name: Ping other repos
+
+on:
+  push:
+    branches: [main]
+  repository_dispatch:
+
+jobs:
+  ping-pong:
```
ddbeck, line 9: I guess we're no longer `pong`-ing now, too

```suggestion
  ping:
```

ddbeck (changes requested):
This is looking really tidy. A couple little suggestions now.

.github/workflows/ping-other-repos.yml
sideshowbarker: Yup, removed `repository_dispatch`

ddbeck (approved):
OK, let's give this a try (pending set up of secrets).

PR #11368, opened 2021-07-03
Add CSS <filter-function> and its 10 functions

teoli2003:
This is the second and last bcd PR for #11301 (it can be landed independantly).

Initially there wasn't a `<filter-function>` CSS types, the 10 functions were defined directly on the CSS `filter`property.

With the addition of `backdrop-filter`, supporting the same filters, this changed and the 10 CSS functions are now defined on a type. MDN has created these pages long ago, but always linked to the bcd of the CSS `filter` property (that don't individualize the functions).

This fixes this by adding bcd for `<filter-function>`(implicitely added to the platform when `filter` has been added) and for each of the 10 functions, as subfeatures (added at the same time, but more were added in the future).

The `mdn_url` (already existing) and `spec_url` have been added too to point to the right pages/fragments.

This can land independantly of any other PR, nothing will break (even temporarily)

### DIFF
--- a/css/types/filter-function.json
+++ b/css/types/filter-function.json
@@ -1,0 +1,555 @@
+{
+  "css": {
+    "types": {
+      "filter-function": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/filter-function",
+          "spec_url": "https://drafts.fxtf.org/filter-effects/#typedef-filter-function",
+          "support": {
+            "chrome": {
+              "version_added": "18"
+            },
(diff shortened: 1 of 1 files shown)


teoli2003:
No worry for the delay.

I took it from the `transform` CSS properties: 2D functions were added when the prefixed versions was added (Chrome 1) and 3D ones in Chrome 12.

Screenshot 2021-08-03 at 14 39 26

ddbeck (approved):
Looks good. Thank you! 🎉

Issue #11208, opened 2021-06-23
CSS types – basic-shape's ellipse() missing


teoli2003:
#### What information was incorrect, unhelpful, or incomplete?
`ellipse()` is missing (we have the others like `circle()` or `polygon()` and the live sample on MDN Web docs shows that some browsers support it at least.

#### Did you test this? If so, how?
(Live sample)

#### Can you link to any release notes, bugs, pull requests, or MDN pages related to this?
https://developer.mozilla.org/en-US/docs/Web/CSS/basic-shape/ellipse()

Blakelist7:
Hey, @teoli2003 if I am not wrong, you wish to add an image like <a href="https://developer.mozilla.org/en-US/docs/Web/CSS/basic-shape/circle()#examples">circle</a> to <a href="https://developer.mozilla.org/en-US/docs/Web/CSS/basic-shape/ellipse()#basic_ellipse_example">Ellipse</a>. Also if this is what this issue says, I don't think we have an image for <a href="https://developer.mozilla.org/en-US/docs/Web/CSS/basic-shape/polygon()#basic_polygon_example">Polygon</a> as well.

teoli2003:
HI @Blakelist7, no that's about adding compatibility data for ellipse in https://github.com/mdn/browser-compat-data/blob/main/css/types/basic-shape.json.

It shouldn't be too difficult to dig this info out, but I hadn't time for it that day. For example, https://bugzilla.mozilla.org/show_bug.cgi?id=1326406 clearly hint at Fx 54, the same as `circle()`. But we still need chromium info, then test with old browser to validate.

Blakelist7:
ohk, now I get it, we have to add a table that shows which browser's support `ellipse()`
right?

Could you assign this to me?
Thanks :)

teoli2003:
Hi @blakelist7. I can't assign it to it (I don't have the rights to do it). Feel free to submit the PR for it, though.

What has to be done is to _edit the **json** file_ I linked too, not to touch the content page: browser compat tables are filled from data in these json files.

PR #11369, opened 2021-07-03
Add data for CSS ellipse() method

Blakelist7:
Add data for`ellipse()`.
Fixes: #11208

### DIFF
--- a/css/types/basic-shape.json
+++ b/css/types/basic-shape.json
@@ -345,7 +345,7 @@
               "version_added": "37"
             },
             "edge": {
-              "version_added": "80"
+              "version_added": "79"
             },
             "firefox": {
               "version_added": "62"
--- a/css/types/basic-shape.json
+++ b/css/types/basic-shape.json
@@ -331,6 +331,56 @@
             }
           }
         }
+      },
+      "ellipse": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/basic-shape/ellipse()",
+          "spec_url": "https://drafts.csswg.org/css-shapes/#funcdef-polygon",
+          "description": "<code>ellipse()</code>",
+          "support": {
+            "chrome": {
+              "version_added": "37"
+            },
+            "chrome_android": {
+              "version_added": "37"
+            },
+            "edge": {
+              "version_added": "80"
+            },
+            "firefox": {
+              "version_added": "62"
+            },
+            "firefox_android": {
+              "version_added": "54"
+            },
+            "ie": {
+              "version_added": false
+            },
+            "opera": {
+              "version_added": "24"
+            },
+            "opera_android": {
+              "version_added": "24"
+            },
+            "safari": {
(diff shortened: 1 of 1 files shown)


Blakelist7:
@teoli2003 please review this PR.

teoli2003:
Hi! @Blakelist7 Thanks a lot! Can you tell us how you determined the values?

Blakelist7:
Sure! I checked if the example was working on a specific version using https://caniuse.com/

teoli2003:
Thanks! You found Edge 80, this is strange as all the other entries on this file have Edge 79. Can you double-check?

Blakelist7:
Actually, in the mentioned site edge version's starts from 80 so I was not able to check for 79.

teoli2003:
@foolip and @vinyldarkscratch, any hint about how to test for Edge 79? (I would be surprised it doesn't work there)

vinyldarkscratch:
Can you share the exact CanIUse entry you obtained your data from?  You can get a permalink by clicking the "#" icon on the entry!

I'm also guessing you went to BrowserStack to test the support, which yes, doesn't have Edge 79.  However, I highly doubt that a feature like this would have been excluded in Edge 79 considering its established support in Chrome.  Have you also tested support in Edge 18 or older?

Blakelist7:
This is my first time using CanIUse, so firstly I searched for `basic-shape` and from there I used BrowserSlack to test.
Yes, I tested for Edge 18 and it is not compatible.

foolip:
> @foolip and @vinyldarkscratch, any hint about how to test for Edge 79? (I would be surprised it doesn't work there)

If testing on BrowserStack I just test Edge 80 and assume 79 will be the same unless the feature was introduced in Chromium 80.

Or one can use Sauce Labs which also has Edge 79.

Blakelist7:
> > @foolip and @vinyldarkscratch, any hint about how to test for Edge 79? (I would be surprised it doesn't work there)
> 
> If testing on BrowserStack I just test Edge 80 and assume 79 will be the same unless the feature was introduced in Chromium 80.
> 
> Or one can use Sauce Labs which also has Edge 79.

Thank you for the resource and I just checked using Sauce Labs, yes it works on Edge 79. :)

teoli2003:
Cool. Thanks. LGTM (I can't formally approve, though)

vinyldarkscratch (approved):
Thanks for the PR!  I just confirmed the version numbers real quick, and this is looking good to me!

teoli2003:
@Blakelist7 Congrats for your first bcd commit!

Blakelist7:
@teoli2003 
Also as I can check for `circle()`, `polygon()` and `inset()` their examples are not supported in FF 54, so should I create an issue here in BCD or content?

vinyldarkscratch:
Since it involves browser compatibility, let's open the issue here?

Blakelist7:
> Since it involves browser compatibility, let's open the issue here?

Cool

PR #11370, opened 2021-07-03
Add Safari versions for api.Element.keyboard_events

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the keyboard events members of the `Element` API, based upon manual testing.

Test Code Used:
```js
document.body.addEventListener('keydown', function(e) {
	alert('keydown');
});
document.body.addEventListener('keypress', function(e) {
	alert('keypress');
});
document.body.addEventListener('keyup', function(e) {
	alert('keyup');
});
```

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -4694,10 +4694,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": "1.2"
             },
             "safari_ios": {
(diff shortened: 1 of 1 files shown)


foolip (approved):
Seems plausible that keyup/keydown could have been supported before keypress.

PR #11403, opened 2021-07-06
Add spec_url for window.closed

teoli2003:
`window.console` had no bcd. It was added at the same time as `Console`, so I copied the bcd. For the `spec_url` entry, as `Console` is a namespace and not an interface, I linked to the webidl entry that explain that a property is automatically on the global scope for namespaces. (This is similar to what we do for iterator methods like `foreach()`, or for `toString`: we link to the webidl spec.

`window.closed` was just missing the `spec_url` clause.

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -1373,59 +1373,6 @@
           }
         }
       },
-      "console": {
-        "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/Window/console",
-          "spec_url": "https://console.spec.whatwg.org/#console-namespace",
-          "support": {
-            "chrome": {
-              "version_added": "1"
-            },
-            "chrome_android": {
-              "version_added": "18"
-            },
-            "edge": {
-              "version_added": "12"
-            },
-            "firefox": {
-              "version_added": "4"
-            },
-            "firefox_android": {
-              "version_added": "4"
-            },
-            "ie": {
-              "version_added": "8",
-              "notes": "In Internet Explorer 8 and 9, the <code>console</code> object is <code>undefined</code> when the developer tools are not open. This behavior was fixed in Internet Explorer 10."
-            },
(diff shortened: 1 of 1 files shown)


api/Window.json
```diff
@@ -1371,6 +1373,59 @@
           }
         }
       },
+      "console": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/Window/console",
+          "spec_url": "https://heycam.github.io/webidl/#dfn-namespace-object",
```
foolip, line 1379: This isn't the right spec URL IMHO, it should be https://console.spec.whatwg.org/#console-namespace

Web IDL's definition of namespaces explains why that ends up as a property on the global object, but that's true of interfaces too.

And speaking of the global object, the console object is available in workers too, so should this really be on Window?

teoli2003: I'll update the spec_url.

Wrt `Window.console`, `Window` is the global scope for non-worker case. As it is available in workers, it means we probably should also have a WorkerGlobalScope.console entry.

But… I never encoutered namespaces before, so I've never thought about it. Maybe we should open a follow-up issue for console in workers? (I am also fine to remove it from this PR)

teoli2003: Backed out.

api/Window.json
```diff
@@ -1271,6 +1271,8 @@
       },
       "closed": {
         "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/Window/closed",
+          "spec_url": "https://html.spec.whatwg.org/multipage/window-object.html#dom-window-closed",
```
foolip, line 1275:
```suggestion
          "spec_url": "https://html.spec.whatwg.org/multipage/window-object.html#dom-window-closed-dev",
```

foolip (approved):
Looks good with -dev appended.

PR #11406, opened 2021-07-06
Updates Edge manifest support

diekus:
#### Summary
Adds manifest supported versions for MS Edge

There are no tests to perform, the only changes are from null to the string version number


Elchi3 (approved):
Thanks and congrats on landing your first change here 👍 🎉

PR #11413, opened 2021-07-07
Demix Navigator* mixins into Navigator and WorkerNavigator

teoli2003:
This is a companion PR to mdn/content#6637

- I moved `Navigation*.*` info to the corresponding `Navigation.*` entries.
- I added it to `WorkerNavigation.*`, they are those of `WorkerNavigation`. I set some values to `true` because `WorkerNavigation` was set to `true` and it is probably a feature as old as it. They will be fixed at the same time (when we test for `WorkerNavigation`, we will test for the subfeatures too).
- I removed the `Navigation*.json` files.


vinyldarkscratch (approved):
I have been looking forward to this for quite some time, thank you for taking the initiative @teoli2003!  LGTM!

vinyldarkscratch:
A little extra note on this that I forgot to mention: for most of the mixins, we have been keeping them defined within separate files, but moving them to a `_mixins` subdirectory (which is hidden from consumers as the mixin members are flattened into their inheriting interfaces; i.e. `NavigatorPlugins__Navigator.json` defines `api.Navigator.plugins`).  However, since these mixins are highly specific for the navigator API, I feel that it makes more sense to ignore there are mixins to begin with and just flatten the data altogether.

PR #11414, opened 2021-07-07
Demix paymentDetailsBase and remove it

teoli2003:
This is the companion PR of mdn/content#6640

I deleted `PaymentDetailsInit` that is a dictionary and was never documented on MDN.
I moved the content of `PaymentDetailsBase` into the only (dictionary) that implemented this mixin: `PaymentDetailsUpdate` (it will go away too at some point, but there is a lot of links to it on MDN).
I deleted `PaymentDetailsBase`.
I removed 404 `mdn_url` of `PaymentDetailsBase` that will never be documented.


vinyldarkscratch (approved):
Looking good to me, thanks!  The content PR is already merged, so this is all ready to go!

PR #11416, opened 2021-07-07
Update Firefox compat of URL.pathname, URL.search

### LINKED ISSUE #10817: URL.search / URL.pathname - compat note about Firefox<53 seems wrong
# Relevant docs

https://developer.mozilla.org/en-US/docs/Web/API/URL
https://developer.mozilla.org/en-US/docs/Web/API/URL/search
https://developer.mozilla.org/en-US/docs/Web/API/URL/pathname

The compat tables say that Firefox<53 returns buggy values and it's marked 🔴 like "no support" which is wrong.
It should be marked with perhaps 🟠 "partial support"; and the note should explain the issue is for non-HTTP(s) URLs only

image

> pathname and search returned the wrong values so that for a URL of http://z.com/x?a=true&b=false, pathname would return "/x?a=true&b=false" and search would return "", rather than "/x" and "?a=true&b=false" respectively.

I tested on BrowserStack with Firefox 53, 48, 32 (win10), 26 (win 8.1), and I don't repro:

image

```
new URL('http://z.com/x?a=true&b=false').search
> "?a=true&b=false"
new URL('http://z.com/x?a=true&b=false').pathname
> "/x"
```

Source of this info is this PR: https://github.com/mdn/browser-compat-data/pull/2163/files#diff-911b66f26bac12a132458b40ad7f763dfe6ee02096a63ad583be768f0e709adbR643

jakub-g:
# Summary

See #10817. Currently the compat table renders as below:
image

IMO the 🔴 for Firefox<53 is too much; the API *is* there; it has a bug, _but_ not for the common case (`http`/`https`).
(The release notes for Firefox 53 are wrong too BTW).
So I think 🟢 with a note is a better fit here.

Example of an existing article using this convention: [elementFromPoint](https://developer.mozilla.org/en-US/docs/Web/API/Document/elementFromPoint) article, Chrome support is `1*` and bug is put as a comment; it's not `66` due to the bug.
image

But if you want I can bring the old schedule back and mark it with 🟠 `"partial_implementation": true`?

(The wording of the bug was also wrong; the example shows `http://` but the bug does not happen for `http://`; I updated the example too)

# Test results and supporting details

Relevant firefox bug: [bugzilla](https://bugzilla.mozilla.org/show_bug.cgi?id=1310483)

Tests in Firefox 48:
image

# Related issues

Fixes #10817

### DIFF
--- a/api/URL.json
+++ b/api/URL.json
@@ -627,26 +627,14 @@
             "edge": {
               "version_added": "13"
             },
-            "firefox": [
-              {
-                "version_added": "53"
-              },
-              {
-                "version_added": "22",
-                "version_removed": "53",
-                "notes": "<code>pathname</code> and <code>search</code> returned the wrong values so that for a URL of <code>http://z.com/x?a=true&b=false</code>, <code>pathname</code> would return \"/x?a=true&b=false\" and <code>search</code> would return \"\", rather than \"/x\" and \"?a=true&b=false\" respectively."
-              }
-            ],
-            "firefox_android": [
-              {
-                "version_added": "53"
-              },
-              {
-                "version_added": "22",
-                "version_removed": "53",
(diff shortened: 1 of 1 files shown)


vinyldarkscratch (approved):
This is much cleaner and far more consistent with our data.  Thanks for the PR, LGTM!

PR #11422, opened 2021-07-08
Add Firefox versions for api.CanvasRenderingContext2D.imageSmoothingEnabled

vinyldarkscratch:
This PR adds real values for Firefox and Firefox Android for the `imageSmoothingEnabled` member of the `CanvasRenderingContext2D` API, based upon manual testing.

Test Code Used:
```html
<canvas id="canvas"></canvas>

<script>
	var canvas = document.createElement('canvas');
	var ctx = canvas.getContext('2d');
	alert(ctx.mozImageSmoothingEnabled);
</script>
```

### DIFF
--- a/api/CanvasRenderingContext2D.json
+++ b/api/CanvasRenderingContext2D.json
@@ -2125,15 +2125,21 @@
                 "prefix": "webkit"
               }
             ],
-            "edge": {
-              "version_added": "15"
-            },
+            "edge": [
+              {
+                "version_added": "15"
+              },
+              {
+                "version_added": "12",
+                "prefix": "ms"
+              }
+            ],
             "firefox": [
               {
                 "version_added": "51"
               },
               {
-                "version_added": true,
+                "version_added": "3.6",
                 "version_removed": "51",
                 "prefix": "moz"
               }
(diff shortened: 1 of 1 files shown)


api/CanvasRenderingContext2D.json
```diff
@@ -2125,15 +2125,21 @@
                 "prefix": "webkit"
               }
             ],
-            "edge": {
-              "version_added": "15"
-            },
+            "edge": [
```
foolip, line 2128: Can you update the title description to cover Edge?

api/CanvasRenderingContext2D.json
```diff
@@ -2125,15 +2125,21 @@
...
+              {
+                "version_added": "15"
+              },
+              {
+                "version_added": "12",
+                "prefix": "ms"
```
foolip, line 2134: Can you link the test used for this? Also I assume it was removed in 79 or earlier?

foolip (changes requested):

api/CanvasRenderingContext2D.json
vinyldarkscratch: Didn't mean to include Edge changes here, oops!

foolip (approved):

PR #11425, opened 2021-07-08
Add missing NavigatorUAData API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing `NavigatorUAData` API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.10).

Spec: https://wicg.github.io/ua-client-hints/#navigatoruadata

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/NavigatorUAData


teoli2003:
Shouldn't bcd for the mixin be added at the same time?
Screenshot 2021-07-08 at 10 14 00
That is `api.Navigator.userAgentData` and `api.WorkerNavigator.userAgentData`

vinyldarkscratch:
Good catch, thanks @teoli2003!

api/Navigator.json
```diff
@@ -3332,6 +3332,54 @@
...
+              "version_added": false
+            },
+            "safari_ios": {
+              "version_added": false
+            },
+            "samsunginternet_android": {
+              "version_added": false
+            },
+            "webview_android": {
+              "version_added": false
```
foolip, line 3373: No WebView?

foolip (changes requested):

api/Navigator.json
vinyldarkscratch: Apparently the WebView version we have in our data is an older version (I haven't updated my testing Android in a while).  Lemme collect results for the latest WebView, one moment...

vinyldarkscratch: There seems to be a bug that's preventing people from updating their WebView (or Chrome) to the latest release on Android devices, and it's affecting my device as well.  I will not be able to test in the latest WebView until that issue is resolved.  Should I just mark it as `null` for now?

foolip: Via [source.chromium.org](https://source.chromium.org/chromium/chromium/src/+/main:android_webview/tools/system_webview_shell/test/data/webexposed/not-webview-exposed.txt;l=153;drc=19e16336f47125b3e4b5c34cac0e93a2f4c921cf) I found my way to https://crbug.com/921655, turns out this isn't enabled in WebView. Let's link https://crbug.com/921655 though.

foolip (approved):

PR #11429, opened 2021-07-08
Add missing XRWebGLBinding API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing `XRWebGLBinding` API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.10).

Spec: https://immersive-web.github.io/layers/#XRWebGLBindingtype

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/XRWebGLBinding

### DIFF
--- a/api/XRWebGLBinding.json
+++ b/api/XRWebGLBinding.json
@@ -1,0 +1,101 @@
+{
+  "api": {
+    "XRWebGLBinding": {
+      "__compat": {
+        "spec_url": "https://immersive-web.github.io/layers/#XRWebGLBindingtype",
+        "support": {
+          "chrome": {
+            "version_added": "89"
+          },
+          "chrome_android": {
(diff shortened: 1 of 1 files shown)


Elchi3 (commented):
What about the other API members the spec mentions?

vinyldarkscratch:
None of those members are supported in any browser, so the [script I'm using](https://github.com/foolip/mdn-bcd-collector/blob/main/add-new-bcd.js) doesn't add them to prevent a bunch of all-false data from getting added.  Happy to add them in, though!

Elchi3 (approved):
okay, cool, let's not add all-false features.

PR #11430, opened 2021-07-08
Add test-spec-links.js linter

queengooborg:
This PR adds a new linter to BCD, which checks the spec URLs and confirms whether they are valid based upon a [W3C database](https://github.com/w3c/mdn-spec-links) as well as the statuses (non-standard and deprecated).  This linter also checks for the presence of an MDN URL and alerts us when one is required.

Since there are a number of files that are in need of a spec URL being added (amongst other issues), this linter will **FAIL** for the time being.

Thanks to the amazing @sideshowbarker for writing the code that went into this linter!  This was a joint effort to take his [checker](https://github.com/w3c/mdn-spec-links/blob/main/.spec-urls-check.js) and turn it into a linter for everyone in BCD.  This PR wouldn't be here without his wonderful work!


sideshowbarker (approved):

sideshowbarker:
I would vote for making the “requires an mdn_url” case a warning case that doesn’t case a non-zero exit status. And the same for the “marked as non-standard, but has a spec_url” and “marked as deprecated, but has a spec_url” cases.

I believe the “requires a spec_url” case is th eonly case we should consider making a non-zero-exit failure case.

And further, maybe the other messages should not even be emitted by default but instead maybe only optionally — by choosing some `lint-full` option or something.

I say all that because the “requires a spec_url” cases are really the only ones that I think of as errors that are absolutely in need of fixing. They are gaps in the data that cause it to be missing essential information.

Therefore, I think this lint should allow those ”requires a spec_url” cases to be brought to everybody’s attention without getting overrun/obscured by the volume of the other messages.

vinyldarkscratch:
I had similar thoughts of reworking the linter to allow for warnings.  I'm planning to discuss the linters and their structure with Daniel when he gets back from his vacation!

queengooborg:
I'm going to go ahead and close this PR because I now realize that we'll probably end up just converting the existing spec check, and the statuses are handled by #15889.

PR #11438, opened 2021-07-08
Remove deprecated XDomainRequest API

vinyldarkscratch:
This PR removes the long-deprecated, non-standard `XDomainRequest` API from BCD.  This API was an IE-only interface to implement CORS, but was removed in IE 10 in favor of using the standard XMLHttpRequest.


foolip (approved):
Approving but not merging so that this can be merged together with the content PR.

sideshowbarker:
I’ve merged the content change, so I’ll go ahead and merge this too

PR #11441, opened 2021-07-08
Add missing Counter API

queengooborg:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing `Counter` API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.10).

Spec: https://www.w3.org/TR/DOM-Level-2-Style/css.html#CSS-Counter

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Counter

### DIFF
--- a/api/Counter.json
+++ b/api/Counter.json
@@ -1,0 +1,203 @@
+{
+  "api": {
+    "Counter": {
+      "__compat": {
+        "spec_url": "https://www.w3.org/TR/DOM-Level-2-Style/css.html#CSS-Counter",
+        "support": {
+          "chrome": {
+            "version_added": "1",
+            "version_removed": "40"
+          },
+          "chrome_android": {
+            "version_added": false
+          },
+          "edge": {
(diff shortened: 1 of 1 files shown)


vinyldarkscratch:
Considering that this is an API from an older version of the spec, I think that we should maybe just not add it.  I'll leave this PR open for some discussion, however.

foolip:
There are failing checks, can you look into that?

Also weren't any of the members supported in Chrome? I would expect the initial support to match Safari...

vinyldarkscratch:
Marked this PR as a draft for the time being while we decide whether or not this should be added to BCD.  We'll come back to this another time.

queengooborg:
Considering that this is no longer in an active spec, I don't think that this is worth adding into BCD.  Closing this PR.

PR #11446, opened 2021-07-09
Adding cause option to Error constructor

HolgerJeromin:
#### Summary
Adding cause option to Error constructor.

https://v8.dev/features/error-cause

### DIFF
--- a/javascript/builtins/Error.json
+++ b/javascript/builtins/Error.json
@@ -110,10 +110,10 @@
               "spec_url": "https://tc39.es/proposal-error-cause/#sec-errorobjects-install-error-cause",
               "support": {
                 "chrome": {
-                  "version_added": 93
+                  "version_added": "93"
                 },
                 "chrome_android": {
-                  "version_added": 93
+                  "version_added": "93"
                 },
                 "edge": {
                   "version_added": false
@@ -162,10 +162,10 @@
             "spec_url": "https://tc39.es/proposal-error-cause/#sec-errorobjects-install-error-cause",
             "support": {
               "chrome": {
-                "version_added": 93
+                "version_added": "93"
               },
               "chrome_android": {
-                "version_added": 93
+                "version_added": "93"
               },
               "edge": {
                 "version_added": false
--- a/javascript/builtins/Error.json
+++ b/javascript/builtins/Error.json
@@ -146,11 +146,11 @@
                   "version_added": false
                 },
                 "webview_android": {
-                  "version_added": false
+                  "version_added": "93"
                 }
               },
               "status": {
-                "experimental": true,
+                "experimental": false,
                 "standard_track": true,
                 "deprecated": false
               }
@@ -198,11 +198,11 @@
                 "version_added": false
               },
               "webview_android": {
-                "version_added": false
+                "version_added": "93"
               }
             },
             "status": {
-              "experimental": true,
+              "experimental": false,
               "standard_track": true,
               "deprecated": false
             }
--- a/javascript/builtins/Error.json
+++ b/javascript/builtins/Error.json
@@ -119,10 +119,10 @@
                   "version_added": false
                 },
                 "firefox": {
-                  "version_added": false
+                  "version_added": "91"
                 },
                 "firefox_android": {
-                  "version_added": false
+                  "version_added": "91"
                 },
                 "ie": {
                   "version_added": false
@@ -171,10 +171,10 @@
                 "version_added": false
               },
               "firefox": {
-                "version_added": false
+                "version_added": "91"
               },
               "firefox_android": {
-                "version_added": false
+                "version_added": "91"
               },
               "ie": {
                 "version_added": false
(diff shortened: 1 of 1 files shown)


HolgerJeromin (commented):
~~The linked blog entry says this is supported in firefox too. But I was not able to find this feature in firefox console.~~
Will be fixed by https://github.com/v8/v8.dev/pull/566

teoli2003:
Also, there is mdn/content#6714 that plans to add all the relevant docs about it (and bcd for Firefox).

HolgerJeromin:
Thanks @teoli2003 for the pointer. Updated my v8.dev PR.

hamishwillee:
@codehag This appears to have been implemented in FF91: https://bugzilla.mozilla.org/show_bug.cgi?id=1679653. Can you confirm that it has actually shipped (and isn't restricted to Nightly). [I scanned the code and shipping email group and saw nothing to confirm either way - still feels wrong that release build isn't indicated on the bug itself. ]

@Elchi3 ///

1. Following our discussion, my understanding is that the data for Error is  incorrectly formatted (not the proposed changes, but the original). Specifically, the structure is currently like:
```
Error
  Error constructor
    options_cause_parameter
  cause
  column number
  filename
  ...
```
This means that we're duplicating cause, and the other options are not indicated as values you can pass to the constructor. 

I believe the right way to to this is:
```
Error
  Error constructor
    options_parameter
       options_cause_parameter
        options_column_number_parameter      
        options_filename_parameter    
        ...
  ...
```
Do we change it?

2. The options parameter and `cause` are in the spec, but the other parameters like the filename are firefox-only. Is there any particular way that should be used to indicate non-specification subfeatures?

hamishwillee:
@ddbeck @Elchi3 Can I please get review/advice on this? This is correct based on the current pattern (and data). 

However you might want me to change the structure as discussed in https://github.com/mdn/browser-compat-data/pull/11446#issuecomment-887093398

ddbeck:
@hamishwillee to your questions:

> is it OK if
> 1. We add the additional level of `options_parameter`?

No, the structure of the data for `options.cause` is correct. See the guideline for [Parameters and parameter object features](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#parameters-and-parameter-object-features).

> 2. We remove `cause` outside of the constructor?

No, because I don't believe it's duplicate data. There's a property of `Error` named `cause` (i.e., `'cause' in Error.prototype → true`). There's a `cause` parameter in the `Error()` constructor. These are two distinct features.

ddbeck (changes requested):
Thanks for this PR! Two suggestions:

- According to [this Chrome status entry](https://chromestatus.com/features/5727099325251584), this appears to have shipped in WebView Android too.
- Since this is supported in both Safari and Chrome, there are now enough implementations to switch the `experimental` status to `false`, for both features.

hamishwillee:
@HolgerJeromin Can you please modify as suggested above https://github.com/mdn/browser-compat-data/pull/11446#pullrequestreview-721204669 -

In addition, if you could mark the cause option as released in version 91 for Firefox and Firefox android that would be awesome!

I can't edit your PR or I'd do it myself.

HolgerJeromin:
I will do on monday or tuesday after my holiday. If you want it faster feel free to copy my PR.

hamishwillee:
Thanks @HolgerJeromin - that would be fine. Much appreciated.

HolgerJeromin:
I added webview and toggled the experimental switch.

HolgerJeromin:
@teoli2003 wrote: 

> Also, there is mdn/content#6714 that plans to add all the relevant docs about it (**and bcd for Firefox**).

@hamishwillee linked this PR in https://github.com/mdn/content/issues/6714#issuecomment-887096001 and kind of claimed this very PR is the BCD for firefox (or at least does not link another one).

Now I am a bit confused. Perhaps the easiest thing would be for me to add firefox compat data in this PR. But I do not want to add even more confusion :-)

hamishwillee:
Thanks @HolgerJeromin 
1. This addresses both the comments in https://github.com/mdn/browser-compat-data/pull/11446#pullrequestreview-721204669
2. I was hoping you could also add "91" to the firefox and firefox android versions in the same places `options_cause_parameter` and `cause`

@ddbeck If you get here before @HolgerJeromin does, can you please just merge this and I'll do this as a separate PR. If he gets to it first then I won't have to.

HolgerJeromin:
I interpreted 

> Also, there is mdn/content#6714 that plans to add all the relevant docs about it (and bcd for Firefox).

as: "Please do not touch firefox because we want to do it by ourselves."
But as I now understood I was happy to add it.

hamishwillee:
Thanks very much @HolgerJeromin - exactly what I hoped for.

@ddbeck Can you please review this?

ddbeck (approved):
Looks good. Thank you! 🎉

PR #11454, opened 2021-07-09
Remove deprecated DOMActivate_event

vinyldarkscratch:
This PR removes the deprecated `DOMActivate` event from BCD due to lack of support.  I have not been able to get the example on the MDN page working in any browser I've tested (Chrome 1, 5, 12, 25, 91; Firefox 89; Safari 14.1), not to mention it has been deprecated.

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -55,56 +55,6 @@
           "deprecated": false
         }
       },
-      "DOMActivate_event": {
-        "__compat": {
-          "description": "<code>DOMActivate</code> event",
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/Element/DOMActivate_event",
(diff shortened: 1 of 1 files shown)


foolip:
Is an MDN change also needed? Have you searched source code for this event?

foolip:
This is actually supported, but https://developer.mozilla.org/en-US/docs/Web/API/Element/DOMActivate_event#examples is very weird, `<handler type="application/ecmascript" ev:event="DOMActivate"> change(evt); </handler>` is something I've never seen before and which isn't supported.

I've confirmed support in Chrome with this simple example:
```html
<!DOCTYPE html>
<button>click me</button>
<script>
document.querySelector('button').addEventListener('DOMActivate', console.log);
</script>
```

foolip:
Closing this in favor of https://github.com/mdn/browser-compat-data/pull/11487.

PR #11455, opened 2021-07-09
Add Firefox versions for TextTrackList API

vinyldarkscratch:
This PR adds real values for Firefox and Firefox Android for the `TextTrackList` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.10).

### DIFF
--- a/api/TextTrackList.json
+++ b/api/TextTrackList.json
@@ -15,10 +15,10 @@
             "version_added": "12"
           },
           "firefox": {
-            "version_added": true
+            "version_added": "31"
           },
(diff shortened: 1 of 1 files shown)


vinyldarkscratch:
Note: this PR also adds version numbers for the API's events.  Their data was derived from the event handlers (AKA `on*`).  Since the event handlers were supported since the API was first introduced, it doesn't make sense if the events themselves were not supported until later.

foolip (approved):

PR #11457, opened 2021-07-09
Add Firefox versions for Pointer Lock API

vinyldarkscratch:
This PR adds real values for Firefox and Firefox Android for the `pointerlock_event` member of the `Document` API.  The data for these events was mirrored from their event handler counterparts.


foolip (commented):
Were there moz-prefixed events before Firefox 50? The data for `exitPointerLock` would suggest that, but it's incomplete.

foolip:
Ping @vinyldarkscratch

foolip:
@vinyldarkscratch I've changed the title to what I think the scope of this should be: any entry in any file with https://w3c.github.io/pointerlock/ as the spec URL. Over to you again :)

foolip:
@vinyldarkscratch can you also check api.MouseEvent.movementX/Y, are the versions right or should they match what you've done here?

foolip:
> @vinyldarkscratch can you also check api.MouseEvent.movementX/Y, are the versions right or should they match what you've done here?

I've confirmed Firefox 41 for the unprefixed properties, so this did come before Pointer Lock API bits.

foolip (approved):

PR #11458, opened 2021-07-09
Add Safari versions for copy/cut/paste_event

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `copy_event`, `cut-_event`, and `paste_event` members of the `Document`, `Element`, and `Window` APIs.  The data for Safari Desktop was mirrored from Chrome, whereas the Safari iOS version was explicitly set to 3.0 based upon an [article](https://www.engadget.com/2009-03-17-iphone-finally-gets-copy-and-paste.html) stating that clipboard support was introduced in iOS 3.0.  For the `clipboardData` member, the data was collected by running the live test on the [MDN article](https://developer.mozilla.org/en-US/docs/Web/API/Element/copy_event) and checking the result.

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -1766,10 +1766,10 @@
               "version_added": "14"
             },
             "safari": {
-              "version_added": true
+              "version_added": "≤4"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "3"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
@@ -3050,10 +3050,10 @@
               "version_added": "14"
             },
             "safari": {
-              "version_added": true
+              "version_added": "≤4"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "3"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
@@ -8166,10 +8166,10 @@
               "version_added": "14"
             },
             "safari": {
-              "version_added": true
+              "version_added": "≤4"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "3"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
--- a/api/Document.json
+++ b/api/Document.json
@@ -1812,10 +1812,10 @@
                 "version_added": "43"
               },
               "safari": {
-                "version_added": true
+                "version_added": "10.1"
               },
               "safari_ios": {
-                "version_added": null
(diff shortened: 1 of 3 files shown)


foolip:
> The data for Safari Desktop was mirrored from Chrome

For anything with OS integration, which is the case here, it does not seem like a safe assumption that early support in Chrome implies early support in Safari. Did you also do some testing for these events?

foolip:
Hmm, the `clipboardData` subfeatures of the *_event features here are very unusual, in fact unique in BCD. These are the only other subfeatures of *_event entries, and they're not about properties of the fired event:
api.Element.click_event.on_disabled_form_elements
api.Window.beforeunload_event.custom_text_support
api.Window.beforeunload_event.event_returnvalue_activation
api.Window.beforeunload_event.preventdefault_activation
api.Window.beforeunload_event.return_string_activation

I think we should delete these entries and fix up the api.ClipboardEvent entry to capture the existing data.

Do you want to do that as part of this PR, or separately?

Lukepearshotmail:
Yes pleaseOn 12 Jul 2021 21:28, Philip Jägenstedt ***@***.***> wrote:
Hmm, the clipboardData subfeatures of the *_event features here are very unusual, in fact unique in BCD. These are the only other subfeatures of *_event entries, and they're not about properties of the fired event:
api.Element.click_event.on_disabled_form_elements
api.Window.beforeunload_event.custom_text_support
api.Window.beforeunload_event.event_returnvalue_activation
api.Window.beforeunload_event.preventdefault_activation
api.Window.beforeunload_event.return_string_activation
I think we should delete these entries and fix up the api.ClipboardEvent entry to capture the existing data.
Do you want to do that as part of this PR, or separately?

vinyldarkscratch:
I did run a test in BrowserStack on Safari 4, apologies for not mentioning it!  Since I'm back home, I also ran tests in Safari 2.0 and 3.0, so I can now specify exact version numbers!

vinyldarkscratch:
As far as the `clipboardData` subfeatures, I'd like to take care of that in a separate PR and get this one merged as it is.

foolip (approved):
Thanks Vinyl!

foolip:
I had a branch sitting with the removal, so I rebased it and sent https://github.com/mdn/browser-compat-data/pull/11524.

PR #11459, opened 2021-07-09
Remove redundant api.HTMLFrameSetElement.event_handlers feature

vinyldarkscratch:
This PR removes api.HTMLFrameSetElement.event_handlers from our data.  Since the API doesn't define its own event handlers as the MDN article states (and I question whether this API actually has an `onstorage` event handler), there's no reason to have this feature around.

### DIFF
--- a/api/HTMLFrameSetElement.json
+++ b/api/HTMLFrameSetElement.json
@@ -95,54 +95,6 @@
           }
         }
       },
-      "event_handlers": {
-        "__compat": {
-          "description": "<code>onXYZ</code> event handlers",
-          "support": {
-            "chrome": {
-              "version_added": true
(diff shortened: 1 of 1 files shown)


foolip:
Per https://html.spec.whatwg.org/multipage/obsolete.html#htmlframesetelement this interface does include the WindowEventHandlers mixin, so this isn't all fiction. I'm not sure we should add all of those entries though...

@ddbeck for thoughts.

ddbeck (approved):
We probably _should_ have the mixin-originating features too. If you want to make that happen, then great, but if you'd rather open an issue saying that it ought to be done _eventually_ and merge this as-is, I'm fine with that too.

PR #11462, opened 2021-07-10
Add Firefox versions for api.KeyboardEvent.KeyboardEvent.code_and_key_in_init

vinyldarkscratch:
This PR adds real values for Firefox and Firefox Android for the `KeyboardEvent.code_and_key_in_init` member of the `KeyboardEvent` API.  The `key` attribute had been around since the constructor's implementation (https://bugzil.la/930893), and the `code` attribute when it was added to the interface (https://bugzil.la/865649).

### DIFF
--- a/api/KeyboardEvent.json
+++ b/api/KeyboardEvent.json
@@ -111,10 +111,10 @@
                 "version_added": "≤79"
               },
               "firefox": {
-                "version_added": null
+                "version_added": "38"
               },
               "firefox_android": {
-                "version_added": null
+                "version_added": "38"
               },
               "ie": {
                 "version_added": false


foolip:
> The `key` attribute had been around since the constructor's implementation (https://bugzil.la/930893), and the `code` attribute when it was added to the interface (https://bugzil.la/865649).

The data for the constructor says 31 though. Did you use 38 since that's when *both* became available?

I think this entry should be split into two, if both of these are useful to know. I would argue the constructor is mostly useless to begin with and we could just as well remove this entry. But neither splitting nor removing need happen in this PR.

vinyldarkscratch:
I think it makes more sense to separate this entry, combining the two individual parameters isn't very useful in my opinion.  Due to my unreliable internet at the time, however, I didn't want to dive down that path, so I decided to just fill the data for this as it is.

foolip (approved):
This seems correct, and I'll send a PR to remove the entry :)

foolip:
I have sent https://github.com/mdn/browser-compat-data/pull/11530.

PR #11488, opened 2021-07-12
Use rfc-editor.org as URL targets for IETF RFCs

dontcallmedom:
Starting from 8650, they're much more readable on rfc-editor.org than on datatracker.ietf.org; browser-specs is aligning to using this as the default target https://github.com/w3c/browser-specs/issues/342 starting with release 2.2.0


Elchi3 (approved):
Nice! 🎉  Thank you! 👍

PR #11492, opened 2021-07-12
Fixed erroneous edge support data for some mask related CSS properties

lukewarlow:
#### Summary

Fixed erroneous edge support data for some mask related CSS properties

Just changed Edge to reflect that version 79 onward matches Chromes support data.

### DIFF
--- a/css/properties/mask.json
+++ b/css/properties/mask.json
@@ -30,9 +30,17 @@
                 "notes": "The prefixed property can be used with SVG and HTML with a slightly different syntax, which allows setting the non-standard <a href='https://developer.mozilla.org/docs/Web/CSS/-webkit-mask-attachment'><code>-webkit-mask-attachment</code></a> property."
               }
             ],
-            "edge": {
-              "version_added": "12"
-            },
+            "edge": [
+              {
+                "prefix": "-webkit-",
+                "version_added": "79",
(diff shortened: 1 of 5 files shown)


css/properties/mask.json
```diff
@@ -30,9 +30,17 @@
                 "notes": "The prefixed property can be used with SVG and HTML with a slightly different syntax, which allows setting the non-standard <a href='https://developer.mozilla.org/docs/Web/CSS/-webkit-mask-attachment'><code>-webkit-mask-attachment</code></a> property."
               }
             ],
-            "edge": {
-              "version_added": "12"
-            },
+            "edge": [
+              {
+                "prefix": "-webkit-",
```
foolip, line 35: There's also a `partial_implementation` entry for Chrome without a prefix, and I can confirm with https://mdn-bcd-collector.appspot.com/tests/css/properties/mask that *something* is supported. Is the "While the property is recognized, values applied to it don't have any effect." note there accurate, and if so can you mirror it to Edge as well?

lukewarlow: Mirrored that across to Edge.

foolip: I've confirmed with https://software.hixie.ch/utilities/js/live-dom-viewer/?saved=9481 that it still doesn't work. https://bugs.chromium.org/p/chromium/issues/detail?id=432153 is the bug for this. I'll add that to the notes.

foolip: Actually, I've asked in https://bugs.chromium.org/p/chromium/issues/detail?id=432153 if it's the right bug, I'll link it if the answer is yes.

foolip (approved):

PR #11498, opened 2021-07-13
Removed experimental flag from VisualViewport data

lukewarlow:
#### Summary

Changed experimental to false as this API has at least 2 implementations.


foolip (approved):
Thanks @lukewarlow! As it happens I sent https://github.com/mdn/browser-compat-data/pull/11506 today, which should make it easy to fix lots of these issues in one go.

PR #1738, opened 2018-04-05
adding compat data for Selection

katzrkool:

PR #11500, opened 2021-07-13
Rename/move entries for Selection method parameters

foolip:
Rename two of them to be more consistent with other parameter entries in BCD and importantly to have underscores in the name.
 
 The subfeatures for modify() are moved to collapse(). The modify() method doesn't have a node or offset parameter, and this data originates from https://github.com/mdn/browser-compat-data/pull/1738 where there appears to have been a mishap while splitting a "params" subfeature for collapse().

### DIFF
--- a/api/Selection.json
+++ b/api/Selection.json
@@ -244,6 +244,102 @@
             "standard_track": true,
             "deprecated": false
           }
+        },
+        "node_parameter_nullable": {
+          "__compat": {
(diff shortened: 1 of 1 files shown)


foolip:
Since I sent https://github.com/mdn/browser-compat-data/pull/11506 today, I'll back out the experimental status changes. Those could all be done by the migration script.

sideshowbarker (approved):

PR #11502, opened 2021-07-13
Make entries for parameter object features more consistent

foolip:
This updates lots of entries that have "option" somewhere in the
description, but isn't exhaustive.

This largely aligns with this guideline:
https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#parameters-and-parameter-object-features

Not everything is covered by the guideline, but a consistent style was
used for parameters being optional and having default values.

Spec links were added where possible.

### DIFF
--- a/api/CSSStyleSheet.json
+++ b/api/CSSStyleSheet.json
@@ -294,9 +294,9 @@
             "deprecated": false
           }
         },
-        "optional_index": {
+        "index_parameter_optional": {
           "__compat": {
-            "description": "<code>index</code> is optional",
+            "description": "<code>index</code> parameter is optional",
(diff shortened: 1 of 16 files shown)


foolip:
I did the sorting as a separate commit to make review easier. I would suggest reviewing the first commit and then jus trusting that the "npm run fix" commit is what it claims to be.

Elchi3 (approved):
Looks good to me! 👍 

I'm not sure if there is anyone who consumed these inconsistent keys, so not sure how release note-worthy all of these changes are in details. @ddbeck maybe a summary that says this made parameter features more consistent is sufficient enough.

PR #11505, opened 2021-07-13
adds spec data to MIDI interfaces

rachelandrew:
Adds spec URLs to the MIDI interfaces as I'm creating the documentation pages.

I also removed a duplicate statechange event from MIDIAccess, as this seemed to be a duplicate of onstatechange.

### DIFF
--- a/api/MIDIAccess.json
+++ b/api/MIDIAccess.json
@@ -195,6 +195,55 @@
           }
         }
       },
+      "statechange_event": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/MIDIAccess/statechange_event",
+          "description": "<code>statechange</code> event",
+          "support": {
+            "chrome": {
+              "version_added": "43"
+            },
+            "chrome_android": {
+              "version_added": "43"
+            },
+            "edge": {
+              "version_added": "≤79"
+            },
+            "firefox": {
+              "version_added": false
+            },
+            "firefox_android": {
+              "version_added": false
+            },
+            "ie": {
+              "version_added": false
+            },
+            "opera": {
+              "version_added": null
+            },
+            "opera_android": {
+              "version_added": null
+            },
+            "safari": {
+              "version_added": false
+            },
+            "safari_ios": {
+              "version_added": false
+            },
+            "samsunginternet_android": {
+              "version_added": "4.0"
+            },
+            "webview_android": {
+              "version_added": "43"
+            }
+          },
+          "status": {
+            "experimental": true,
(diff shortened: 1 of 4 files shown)


foolip:
> Adds spec URLs to the MIDI interfaces as I'm creating the documentation pages.

Excellent! I can review this.

> I also removed a duplicate statechange event from MIDIAccess, as this seemed to be a duplicate of onstatechange.

Unfortunately support of events is a bit of a mess in BCD, with https://github.com/mdn/browser-compat-data/issues/7545 describing the current state of things with no clear conclusion for how to improve it. We have lots of APIs with both `onfoo` and `foo_event` entries, and while I've never asked for `foo_event` entries to be added, I also don't think we should remove them until that issue has been resolved somehow.

api/MIDIAccess.json
```diff
@@ -191,58 +195,10 @@
           }
         }
       },
-      "statechange_event": {
```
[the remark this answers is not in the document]
rachelandrew: I've just re-added that.

api/MIDIInput.json
```diff
@@ -51,6 +51,7 @@
       "midimessage_event": {
         "__compat": {
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/MIDIInput/midimessage_event",
+          "spec_url": "https://webaudio.github.io/web-midi-api/#dom-midimessageevent",
```
foolip, line 54: This URL is for the `MIDIMessageEvent` interface, the same as https://webaudio.github.io/web-midi-api/#midimessageevent-interface which is already linked in MIDIMessageEvent.json.

Unfortunately, there is nothing linkable for the concept of the "midimessage" event firing, it's just an algorithm in https://webaudio.github.io/web-midi-api/#dom-midiinput-onmidimessage.

I realize this is a pedantic distinction, and not one that's actually useful to make in documentation, but again see https://github.com/mdn/browser-compat-data/issues/7545 about the current mess. I'm thinking at least it's good to be consistent about what the "midimessage_event" vs. "onmidimessage" entry means so that if such entries are later merged, differences between them are clues that there's something worth noting for web developers. (Search for "addEventListener" in Document.json for attempts to capture issues like this in the current structure.)

foolip (commented):
Oops, I found another thing.

api/MIDIInput.json
rachelandrew: I've removed that link for now, pending a fix to the whole situation.

foolip (approved):
Thanks @rachelandrew for your patience!

PR #11506, opened 2021-07-13
Migration 007: Set experimental to false

foolip:
This is to bring the data in line with the recently added guideline:
https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#choosing-an-experimental-status

Lots of features which are now widely supported are still marked as
experimental. This migration script conservatively sets experimental to
false for features which are supported in Chrome+Firefox+Safari.

It could later be tweaked to do the same for 2-engine support.

### DIFF
--- a/scripts/migrations/007-experimental-false.js
+++ b/scripts/migrations/007-experimental-false.js
@@ -1,0 +1,106 @@
+/* Any copyright is dedicated to the Public Domain.
+ * http://creativecommons.org/publicdomain/zero/1.0/ */
+
+'use strict';
+const fs = require('fs');
+const path = require('path');
+
+const { walk } = require('../../utils');
+
+/**
+ * @param {object} bcd Parsed BCD object to be updated in place.
+ */
+const fixExperimental = (bcd) => {
+  for (const { path, compat } of walk(undefined, bcd)) {
+    if (!compat?.status?.experimental) {
+      continue;
+    }
+
+    // This entry is marked as experimental. Check which browsers support it.
+
+    const supportedIn = new Set();
+
+    for (const [browser, support] of Object.entries(compat.support)) {
+      // Consider only the first part of an array statement.
+      const statement = Array.isArray(support) ? support[0] : support;
+      // Ignore anything behind flag, prefix or alternative name
+      if (statement.flags || statement.prefix || statement.alternative_name) {
+        continue;
+      }
+      if (statement.version_added && !statement.version_removed) {
+        supportedIn.add(browser);
+      }
+    }
+
+    const widelySupported = ['chrome', 'firefox', 'safari'].every((browser) =>
+      supportedIn.has(browser),
+    );
+    if (widelySupported) {
+      compat.status.experimental = false;
+    }
+  }
+};
+
+/**
+ * @param {string} filename Filename of BCD to be updated in place.
+ */
+const fixExperimentalFile = (filename) => {
+  const actual = fs.readFileSync(filename, 'utf-8').trim();
+  const bcd = JSON.parse(actual);
+  fixExperimental(bcd);
+  const expected = JSON.stringify(bcd, null, 2);
+
+  if (actual !== expected) {
+    fs.writeFileSync(filename, expected + '\n', 'utf-8');
+  }
+};
+
+if (require.main === module) {
+  /**
+   * @param {string[]} files
+   */
+  function load(...files) {
+    for (let file of files) {
+      if (file.indexOf(__dirname) !== 0) {
+        file = path.resolve(__dirname, '..', '..', file);
+      }
+
+      if (!fs.existsSync(file)) {
+        continue; // Ignore non-existent files
+      }
+
+      if (fs.statSync(file).isFile()) {
+        if (path.extname(file) === '.json') {
+          fixExperimentalFile(file);
+        }
+
+        continue;
+      }
+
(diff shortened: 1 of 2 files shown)


sideshowbarker:
[X] code reviewed
[X] tested (affects 553 features in 116 files)

So I’d be happy to merge this, but I don’t know who else should also sign off on it — and can’t remember what the next step is after this is actually merged…

foolip:
I also don't know if anyone is responsible for running the migration script after it's merged. I was planning to submit PRs with what it produces, and the edit it to cover 2-engine cases too.

foolip:
I've sent https://github.com/mdn/browser-compat-data/pull/11521 with the changes this script makes.

vinyldarkscratch:
@sideshowbarker The repository owners are the ones who must review and merge PRs involving linters, schema, infrastructure, and scripts like this one.  Our rights as peers doesn't allow us to merge script PRs like this one.

@foolip I don't think we have any particular guideline on who performs the migration PRs, but in the past for the migrations I've written, I've handled the whole process!

foolip:
Ping @ddbeck for review.

ddbeck (changes requested):
This looks good. I have one suggestion.

We do have a doc covering this procedure here: https://github.com/mdn/browser-compat-data/blob/main/docs/migrations.md. The only part we haven't really satisfied yet is a standalone test, however. I think you could extract the for loop to a separate function, write a couple examples (e.g., one showing a feature that needs to be marked false, and another that doesn't), and demonstrate it that way. That would be a nice confidence builder and satisfy our documented expectations.

queengooborg:
@foolip Ping!

foolip:
@queengooborg there we go, test added!

queengooborg:
Quick request: may we actually call this migration 007 instead? We've opened PRs for migrations titled [005](https://github.com/mdn/browser-compat-data/milestone/26) and [006](https://github.com/mdn/browser-compat-data/milestone/27), so this would help us maintain continuity!

(In doing so, you also get to make lots of James Bond jokes.  :P)

foolip:
Sure, let this script be known as James Bond 007: Experimental Migration.

foolip:
@queengooborg is this blocked on anything, or ready to go?

scripts/migrations/007-experimental-false.test.js
```diff
@@ -0,0 +1,69 @@
...
+              standard_track: true,
+              deprecated: false,
+            },
+          },
+        },
+      },
+    };
+    assert.equal(bcd.api.fetch.__compat.status.experimental, true);
+    fixExperimental(bcd);
+    assert.equal(bcd.api.fetch.__compat.status.experimental, true);
```
queengooborg, line 67: Shouldn't experimental be set to `false`, since the feature is supported in two different browser engines?

scripts/migrations/007-experimental-false.js
```diff
@@ -0,0 +1,106 @@
...
+      // Ignore anything behind flag, prefix or alternative name
+      if (statement.flags || statement.prefix || statement.alternative_name) {
+        continue;
+      }
+      if (statement.version_added && !statement.version_removed) {
+        supportedIn.add(browser);
+      }
+    }
+
+    const widelySupported = ['chrome', 'firefox', 'safari'].every(browser =>
```
queengooborg, line 35: What about mobile browsers?  If a mobile-only feature is supported in both Chrome Android and Safari iOS, shouldn't it also be non-experimental?

queengooborg (changes requested):

scripts/migrations/007-experimental-false.test.js
foolip: I have a follow-up to count engines instead, but I'd like to update and use this script in multiple phases, to start with the really simple cases. More scrutiny in review will be needed for the "two different engines" changes I think.

scripts/migrations/007-experimental-false.js
foolip: My follow-up does this.

foolip:
@queengooborg I went to create a follow-up PR, but then thought it's simple enough to just do it right away.

scripts/migrations/007-experimental-false.js
```diff
@@ -0,0 +1,123 @@
...
+    let engineCount = 0;
+    for (const engine of ['Blink', 'Gecko', 'WebKit']) {
+      if (engineSupport.has(engine)) {
+        engineCount++;
+      }
+    }
+
+    if (engineCount > 1) {
+      compat.status.experimental = false;
+    }
```
queengooborg, line 57: Could we perhaps transform this migration to go both ways, setting `true` and `false`?  

```suggestion
    compat.status.experimental = engineCount < 2;
```

foolip: That wouldn't match our guidelines:
https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#choosing-an-experimental-status

If we want to infer experimental status entirely by script, we also need to take age into account.

But if we think that is possible/worthwhile, then we should delete the experimental status from source and make it a build step, so not this piece of code.

queengooborg: That makes sense -- in that case, this is LGTM and we can follow up in the future to add more and potentially automate the experimental status flag!

queengooborg (approved):

PR #11512, opened 2021-07-14
FF version update

Rumyra:
#### Summary

Updating Firefox version

tests pass

### DIFF
--- a/browsers/firefox_android.json
+++ b/browsers/firefox_android.json
@@ -511,27 +511,29 @@
         "89": {
           "release_date": "2021-06-01",
           "release_notes": "https://developer.mozilla.org/docs/Mozilla/Firefox/Releases/89",
-          "status": "current",
+          "status": "retired",
           "engine": "Gecko",
           "engine_version": "89"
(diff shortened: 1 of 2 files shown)


browsers/firefox_android.json
```diff
@@ -511,27 +511,29 @@
...
-          "status": "nightly",
+          "status": "beta",
           "engine": "Gecko",
           "engine_version": "91"
         },
         "92": {
           "release_date": "2021-09-07",
           "status": "planned",
+          "release_notes": "https://developer.mozilla.org/docs/Mozilla/Firefox/Releases/92",
+          "status": "nightly",
```
foolip, line 536: Note that this added a second "status" key. Surprisingly that didn't fail any tests, but anyway it was the only difference to https://github.com/mdn/browser-compat-data/pull/11511.

foolip:
https://github.com/mdn/browser-compat-data/pull/11511 was sent just before this, and I've merged that instead. I checked for differences and commented on the only one.

PR #11516, opened 2021-07-14
Updates to font-size-adjust

rachelandrew:
Various updates to `font-size-adjust`

I am working on https://github.com/mdn/content/issues/6720

This PR adds a section for the two-value syntax now supported in Firefox.

While testing I realized the single value did not work in Chrome. It was behind a flag, however has been moved to test and therefore does not seem to be accessible using Experimental Web Platform Features now. https://bugs.chromium.org/p/chromium/issues/detail?id=451346#c33 I have tested this and the removal seems to apply from Chrome 91.

I also updated the spec to point to the level 5 spec, as the level 4 spec does not include the two-value syntax.

### DIFF
--- a/css/properties/font-size-adjust.json
+++ b/css/properties/font-size-adjust.json
@@ -4,10 +4,11 @@
       "font-size-adjust": {
         "__compat": {
           "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/font-size-adjust",
-          "spec_url": "https://drafts.csswg.org/css-fonts/#font-size-adjust-prop",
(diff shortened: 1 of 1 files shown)


vinyldarkscratch (changes requested):
Thanks for tracking this down, this is looking good to me!  Can we also add the `version_removed` to Opera 77, Opera Android 64, and WebView 91?

vinyldarkscratch (approved):
Ah my bad, I didn't notice the fact this was supported behind a flag in Chrome, so WebView should've remained `false`!  Thanks for the quick fix, this is looking good to me!

PR #11520, opened 2021-07-14
Drop Sanitizer.p.sanitizeToString

sideshowbarker:
https://github.com/WICG/sanitizer-api/commit/f1b12c1 (https://github.com/WICG/sanitizer-api/pull/99) completely removed `Sanitizer.prototype.sanitizeToString` from the HTML Sanitizer API spec.

It never shipped in any browsers other than Firefox, and in Firefox it only ever shipped behind a flag.

So there’s no value to developers in keeping it around forever in BCD and in MDN. It was never a cross-browser feature of the platform.

### DIFF
--- a/api/Sanitizer.json
+++ b/api/Sanitizer.json
@@ -188,69 +188,6 @@
             "deprecated": false
           }
         }
-      },
-      "sanitizeToString": {
-        "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/Sanitizer/sanitizeToString",
-          "spec_url": "https://wicg.github.io/sanitizer-api/#dom-sanitizer-sanitizetostring",
-          "support": {
-            "chrome": {
-              "version_added": false
-            },
-            "chrome_android": {
-              "version_added": false
-            },
-            "edge": {
-              "version_added": false
-            },
-            "firefox": {
-              "version_added": "83",
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "dom.security.sanitizer.enabled",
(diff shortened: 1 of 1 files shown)


foolip:
I agree that this should be removed, but I notice that https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-features doesn't have a provision for it. I think we should remove things behind flags immediately when they're removed from implementation and not wait two years.

foolip:
@ddbeck to bless ignoring the guideline :)

foolip (approved):

ddbeck (approved):
#blessed

ddbeck:
Yeah, I'd support an amendment to the guideline to allow this sort of thing to go away faster. 👍

ddbeck:
Oh, and I got ahead of things a little bit here. @sideshowbarker, would you mind doing something about this page (and any that reference it)? https://developer.mozilla.org/en-US/docs/Web/API/Sanitizer/sanitizeToString

sideshowbarker:
> @sideshowbarker, would you mind doing something about this page (and any that reference it)? [developer.mozilla.org/en-US/docs/Web/API/Sanitizer/sanitizeToString](https://developer.mozilla.org/en-US/docs/Web/API/Sanitizer/sanitizeToString)

Yes, I plan to — though it’s going to require more than just touching that page and the references to it. There are multiple articles with code examples which use that method — so all of those examples need to be updated to use the new method from the spec which was added to replace `sanitizeToString()`

PR #11521, opened 2021-07-14
Set experimental to false for widely supported features

foolip:
This is entirely produced using the migration script:
https://github.com/mdn/browser-compat-data/pull/11506

### DIFF
--- a/api/AbortController.json
+++ b/api/AbortController.json
@@ -60,7 +60,7 @@
           }
         },
         "status": {
-          "experimental": true,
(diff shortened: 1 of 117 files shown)


ddbeck:
(Marked as not ready, awaiting https://github.com/mdn/browser-compat-data/pull/11506.)

queengooborg:
I'm going to close this PR as a rebase is needed and since the changes are purely scripted it'd be better to re-run the script when the script is ready to be merged.

PR #11524, opened 2021-07-14
Remove clipboardData subfeatures in favor of ClipboardEvent

foolip:
These subfeatures mean the same thing as the
api.ClipboardEvent.clipboardData entry, namely whether copy/cut/paste
events fired will have the clipboardData property.

Fill out more of the ClipboardEvent data on the assumption that the now
removed data was correct.

Follow-up to https://github.com/mdn/browser-compat-data/pull/11458.


vinyldarkscratch (approved):
Looking good to me, thank you!

PR #11528, opened 2021-07-14
Add spec links for some <iframe> attributes

foolip:


foolip:
@sideshowbarker I suspect there aren't the URLs you think we should use?

sideshowbarker (approved):

sideshowbarker:
> @sideshowbarker I suspect there aren't the URLs you think we should use?

Actually, these are perfect. For content attributes (as opposed to DOM/IDL attributes), I think the choice for the target anchor to use is typically unambiguous — so the ones here are consistent with the existing ones we already have in BCD for content attributes.

All that said, however, since these features don’t have their own MDN articles, adding them to BCD won’t have any observable effect in MDN, right?

But of course there may be other tooling that’ll want to do something with them, so let’s go ahead and merge them.

foolip:
I thought perhaps you would have preferred to link to https://html.spec.whatwg.org/multipage/iframe-embed-object.html#the-iframe-element:attr-iframe-allow which is domintro-like.

The reason I added these is because of the `allowfullscreen` attribute, so that there's no mystery about why it doesn't link to the Fullscreen spec.

sideshowbarker:
> I thought perhaps you would have preferred to link to [html.spec.whatwg.org/multipage/iframe-embed-object.html#the-iframe-element:attr-iframe-allow](https://html.spec.whatwg.org/multipage/iframe-embed-object.html#the-iframe-element:attr-iframe-allow) which is domintro-like.

Ah those

While those are an analog to the domintro stuff, I think the rationale that can be articulated for instead using the anchors — `dfn`s — those link to is, In the case of content attributes, the `dfn` parts themselves actually are written specifically for web developers, and do in fact actually state normative requirements for authors: document-conformance requirements (aka authoring requirements).

That’s in contrast to the case of the `dfn`s for DOM/IDL attributes, where those parts are written specifically for implementors, and state normative requirements for implementations.

foolip:
Ah, I see. That makes sense actually, the mere definition of a content attribute can never be the implementation requirements, that's all in how it is used, typically in some algorithm elsewhere.

PR #11530, opened 2021-07-14
Remove api.KeyboardEvent.KeyboardEvent.code_and_key_in_init entry

foolip:
This is similar to https://github.com/mdn/browser-compat-data/pull/11499.

The constructor isn't especially useful to begin with, keeping track of
when individual object parameters were supported isn't practically
useful. What matters in practice is when the properties on the event
were supported, which is already captured.


ddbeck (approved):
Seems reasonable. Thank you! 🎉

PR #11531, opened 2021-07-14
Updated Edge support for `hyphens` property

lukewarlow:
#### Summary

Updates Edge support data for the `hyphens` CSS property

Added chrome related support data to Edge. And marked the EdgeHTML specific support as removed in v79.


foolip (approved):
Thanks for continuing to fix a bunch of these clearly wrong claims that new Edge continues to support -ms-prefixed stuff.

PR #11552, opened 2021-07-16
Update Edge support from String.prototype.trimStart/trimEnd

foolip:
Confirmed by testing in Edge 18 and 79 on Sauce using these tests:
https://mdn-bcd-collector.appspot.com/tests/javascript/builtins/String


foolip:
This is the first error I found after adding JavaScript builtins to mdn-bcd-collector.

Elchi3 (approved):

PR #11558, opened 2021-07-16
Add missing features for CloseEvent API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing features of the CloseEvent API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.10).

Spec: https://html.spec.whatwg.org/multipage/web-sockets.html#the-closeevent-interface

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/CloseEvent

Note: Chrome/Safari data was derived from commit history instead -- see https://source.chromium.org/chromium/chromium/src/+/63fc605d4f4d52aafa6f568b096c44ff9dd55021 and https://source.chromium.org/chromium/chromium/src/+/cb8badea0d665f6bb1107f88626e38edf820767d.

### DIFF
--- a/api/CloseEvent.json
+++ b/api/CloseEvent.json
@@ -125,10 +125,10 @@
               "version_added": "10"
             },
             "opera": {
-              "version_added": "15"
+              "version_added": "≤15"
             },
             "opera_android": {
-              "version_added": "14"
+              "version_added": "≤14"
             },
             "safari": {
               "version_added": "6"
@@ -224,10 +224,10 @@
               "version_added": "10"
             },
             "opera": {
-              "version_added": "15"
+              "version_added": "≤15"
             },
             "opera_android": {
-              "version_added": "14"
+              "version_added": "≤14"
             },
             "safari": {
               "version_added": "6"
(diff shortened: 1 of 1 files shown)


api/CloseEvent.json
```diff
@@ -102,6 +102,54 @@
...
+              "version_added": "8"
+            },
+            "firefox_android": {
+              "version_added": "8"
+            },
+            "ie": {
+              "version_added": "10"
+            },
+            "opera": {
+              "version_added": "15"
```
foolip, line 128: It seems incredibly suspect that the event interface itself is marked as supported from Opera 12.1 but none of the members are. That seems implausible. Can you do some testing in Presto, maybe add an instance to custom tests?

vinyldarkscratch: I've set these to ranged values for the time being since Opera isn't one of the eight main browsers BCD is concerned about.  I'll come back to it another time!

foolip (approved):

PR #11561, opened 2021-07-17

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing features of the DOMMatrix API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.10).

Spec: https://drafts.fxtf.org/geometry/#dommatrix

### DIFF
--- a/api/DOMMatrix.json
+++ b/api/DOMMatrix.json
@@ -98,6 +98,150 @@
           }
         }
       },
+      "fromFloat32Array": {
+        "__compat": {
+          "spec_url": "https://drafts.fxtf.org/geometry/#dom-dommatrixreadonly-fromfloat32array",
+          "support": {
+            "chrome": {
+              "version_added": "61"
+            },
+            "chrome_android": {
+              "version_added": "61"
+            },
+            "edge": {
+              "version_added": "79"
+            },
+            "firefox": {
+              "version_added": "69"
+            },
+            "firefox_android": {
+              "version_added": "79"
+            },
+            "ie": {
+              "version_added": false
+            },
+            "opera": {
+              "version_added": "48"
+            },
+            "opera_android": {
+              "version_added": "45"
+            },
+            "safari": {
+              "version_added": "11"
+            },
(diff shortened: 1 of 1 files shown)


saschanaz:
Could you also cover the attributes? `DOMPoint` has the data for inherited atttributes (#10975) but `DOMMatrix` does not.

vinyldarkscratch:
Sorry, I'm not entirely certain what you're asking for, @saschanaz?  Can you clarify?

saschanaz:
It has quite a lot writable attributes (different from ones from its ancestor interface) but not in BCD. https://developer.mozilla.org/en-US/docs/Web/API/DOMMatrix#properties

vinyldarkscratch:
I see, so you're looking for the attributes inherited from `DOMMatrixReadOnly`.  Currently, I'm focusing on all that's added by the mdn-bcd-collector project, and these inherited attributes aren't covered by it.  On other words, I would like to get these features added for now, and follow up with the rest in another PR!

saschanaz:
>https://github.com/foolip/mdn-bcd-collector/blob/149912853621f7bec4d3c5afccb85c72e75779af/build.js#L259

I wonder it's because it filters out inherited members here 🤔🤔 I'll file an issue there.

foolip:
https://github.com/foolip/mdn-bcd-collector/pull/1320 is a new release including the new tests. After we get through the current burndown list (base on an older version of the collector) we'll run the script again, and these entries should then show up.

@saschanaz feel free to file an issue about these missing properties if you want to make sure it's not forever forgotten. This PR is adding static methods which are already on `DOMMatrixReadOnly`, and is an improvement in itself, so I'll merge it.

foolip (approved):

foolip:
I've sent https://github.com/mdn/browser-compat-data/pull/11627 to fix up some descriptions and URLs.

PR #11562, opened 2021-07-17
Add missing api.DOMRect.fromRect feature

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing DOMRect API feature, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.10).

Spec: https://drafts.fxtf.org/geometry/#dom-domrect-fromrect

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/DOMRect/fromRect

### DIFF
--- a/api/DOMRect.json
+++ b/api/DOMRect.json
@@ -176,6 +176,54 @@
           }
         }
       },
+      "fromRect": {
+        "__compat": {
+          "spec_url": "https://drafts.fxtf.org/geometry/#dom-domrect-fromrect",
+          "support": {
+            "chrome": {
+              "version_added": "61"
+            },
+            "chrome_android": {
+              "version_added": "61"
+            },
+            "edge": {
+              "version_added": "79"
+            },
+            "firefox": {
+              "version_added": "69"
+            },
+            "firefox_android": {
+              "version_added": "79"
+            },
+            "ie": {
+              "version_added": false
+            },
+            "opera": {
+              "version_added": "48"
+            },
+            "opera_android": {
+              "version_added": "45"
+            },
+            "safari": {
+              "version_added": "10.1"
+            },
+            "safari_ios": {
+              "version_added": "10.3"
+            },
+            "samsunginternet_android": {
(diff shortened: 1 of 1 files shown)


saschanaz:
Same here: https://github.com/mdn/browser-compat-data/pull/11561#issuecomment-881893951

foolip (approved):
Matches https://developer.mozilla.org/en-US/docs/Web/API/DOMRectReadOnly/fromRect apart from Firefox for Android which is wrong (false) in the existing data.

foolip:
@saschanaz are you asking for some change in this PR?

saschanaz:
Yeah, I want `api.DOMRect.x` etc. Currently MDN just refers to the ancestor's properties, but `Object.getOwnPropertyDescriptor(DOMRect.prototype, "x") !== Object.getOwnPropertyDescriptor(DOMRectReadOnly.prototype, "x")`.

#10975 properly added `api.DOMPoint.x` in addition to existing `api.DOMPointReadOnly.x`, I want to see the same here (or in a followup).

vinyldarkscratch:
As I've stated in https://github.com/mdn/browser-compat-data/pull/11561#issuecomment-882038499, I'd like to look into the inherited attributes in a follow-up PR, and just focus on this feature for the time being!

foolip:
@saschanaz since it's almost always possible to find something adjacent that's missing or wrong, I usually try to review and land small PRs to avoid many review roundtrips. An exception is if a PR makes BCD seemingly contradictory even if the lint passes, like when the data for two features I know were shipped together don't match. In other cases I'll either file an issue or merge and send the followup myself.

In this case I guess the follow-up would be to fix up all the attributes on `DOMRect` and `DOMRectReadOnly`. (Note that api.DOMRectReadOnly.fromRect is already in BCD.)

PR #11574, opened 2021-07-19
Adds information about content:none

rachelandrew:
Working on https://github.com/mdn/content/issues/6716

Firefox 91 implements `content: none` on elements. I do not believe any other UA has shipped this.

Intent to Ship: https://groups.google.com/a/mozilla.org/g/dev-platform/c/oWc1K4Uur2Y

### DIFF
--- a/css/properties/content.json
+++ b/css/properties/content.json
@@ -147,7 +147,7 @@
         },
         "none": {
           "__compat": {
-            "description": "<code>content: none</code> for elements.",
+            "description": "<code>content: none</code> for elements",
             "support": {
               "chrome": {
                 "version_added": false
--- a/css/properties/content.json
+++ b/css/properties/content.json
@@ -145,6 +145,66 @@
             }
           }
         },
+        "none": {
+          "__compat": {
+            "description": "<code>content: none</code> for elements.",
+            "support": {
+              "chrome": {
+                "version_added": false
+              },
+              "chrome_android": {
+                "version_added": false
+              },
+              "edge": {
+                "version_added": false
+              },
+              "firefox": {
+                "version_added": "91",
+                "flags": [
+                  {
+                    "type": "preference",
+                    "name": "layout.css.element-content-none.enabled"
+                  }
+                ]
+              },
+              "firefox_android": {
+                "version_added": "91",
+                "flags": [
+                  {
+                    "type": "preference",
+                    "name": "layout.css.element-content-none.enabled"
+                  }
+                ]
+              },
+              "ie": {
+                "version_added": false
+              },
+              "opera": {
+                "version_added": false
+              },
+              "opera_android": {
+                "version_added": false
+              },
+              "safari": {
+                "version_added": false
+              },
+              "safari_ios": {
+                "version_added": false
+              },
+              "samsunginternet_android": {
+                "version_added": false
+              },
+              "webview_android": {
+                "version_added": false
+              }
+            },
+            "status": {
+              "experimental": true,
+              "standard_track": true,
+              "deprecated": false
+            }
+          }
+        },
         "url": {
           "__compat": {
             "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/url()",
--- a/css/properties/content.json
+++ b/css/properties/content.json
@@ -145,7 +145,7 @@
             }
           }
         },
-        "none": {
+        "none_applies_to_elements": {
           "__compat": {
             "description": "<code>content: none</code> for elements",
             "support": {


css/properties/content.json
```diff
@@ -145,6 +145,66 @@
             }
           }
         },
+        "none": {
+          "__compat": {
+            "description": "<code>content: none</code> for elements.",
```
ddbeck, line 150: style nit

```suggestion
            "description": "<code>content: none</code> for elements",
```

ddbeck (changes requested):
One tiny suggestion in a line comment, but I have a question which might impact the structure here: is `none` itself widely supported? I was unfamiliar with it until today and that has me wondering whether there ought to be a `none` value feature (with this new feature underneath it).

rachelandrew:
> One tiny suggestion in a line comment, but I have a question which might impact the structure here: is `none` itself widely supported? I was unfamiliar with it until today and that has me wondering whether there ought to be a `none` value feature (with this new feature underneath it).

I'm not sure I understand the question, I've made this a separate feature because it's not widely supported. It's a new value for `content`. I'm not sure what change you are suggesting.

ddbeck:
> > One tiny suggestion in a line comment, but I have a question which might impact the structure here: is `none` itself widely supported? I was unfamiliar with it until today and that has me wondering whether there ought to be a `none` value feature (with this new feature underneath it).
> 
> I'm not sure I understand the question, I've made this a separate feature because it's not widely supported. It's a new value for `content`. I'm not sure what change you are suggesting.

I'm not suggesting anything yet because I don't understand the feature. I read "`content: none` for elements" to mean that your new feature represents the `none` value for `content` for regular elements as opposed to pseudo-elements. Is that a correct reading? Or to put this another way, is `content: none` supported in _any_ context prior to Firefox 91?

rachelandrew:
> I'm not suggesting anything yet because I don't understand the feature. I read "`content: none` for elements" to mean that your new feature represents the `none` value for `content` for regular elements as opposed to pseudo-elements. Is that a correct reading? Or to put this another way, is `content: none` supported in _any_ context prior to Firefox 91?

content: none is supported by pseudo-elements for generated content, this is the support on elements which is being implemented in Firefox. It was the bit about a "none value feature" I didn't understand.

ddbeck:
Thanks for the additional info, @rachelandrew.

Can you rename the key for this feature to something like `none_applies_to_elements`?

Here's my thinking: as written in the PR, this feature's key is `"none"` but the description (and the substance, confirmed by your most recent comment) is actually for a particular application of the `none` value, not whether the value is recognized or does anything in general. In the abstract, I was thinking the hierarchy of features was like this:

* `css`
  * `properties`
    * `content`
      *  `none` ← doesn't exist
         * `applies_to_elements` ← this PR

We could, in fact, have this structure—I'd welcome that, actually—but to create the `applies_to_elements` feature would require setting values for the `none` feature. I'm guessing you don't actually want to do that work, so we need to make it clear that we've flattened out the structure.

rachelandrew:
I feel like I'm being really dim here, but I don't understand the level of "none" doesn't exist. Or the distinction of whether the value is recognized, it's definitely recognized, you can confirm that in DevTools. 

So let's try again. CSS has a `content` property. It is typically used as a property for a pseudo-element created with `::before` or `::after` (and some other pseudos) in order to insert something via CSS. It also has a value of `none`, which you would use to prevent anything from being inserted. `content: none` is generally a thing.

It is not, however, supported on other elements, things other than generated content. For example, a `<p>`. However, in the spec it can be used on other elements, in which case it prevents the children of that element from displaying. Firefox is implementing that case. 

So I don't understand what is special about this, that requires this amount of complication. I'll happily "do that work", it doesn't make any difference to me what I spend my Mozilla hours on, so if this is what is wanted I can do it. However, at the moment I don't understand what work there is to be done, or what you mean by "setting values for the none feature". Sorry for the confusion!

ddbeck:
I'm sorry I've been unclear so far. Let me try again with an example and if this doesn't work, maybe we could have a synchronous conversation instead?

Suppose I've got some CSS like this:

```
p::before {
  content: '\00B6';
}

p#pseudo::before {
  /* Firefox ??: behavior for pseudo-elements */
  content: none;
}

p#element {
  /* Firefox 91: new behavior with regular elements */
  content: none;
}
```

It's my guess that the support for the rule `p#special::before` is rather long standing, while the `p#element` usage is new in Firefox 91. Conventionally, I'd expect `css.properties.content.none` to cover the broadest behavior of `none` (pseudo + regular elements), not a newer portion of the overall behavior of `none`.

As the PR is written, if we ever want to represent that broader behavior, we'll have to rename the feature entry you've created, which I want to avoid.

Or to put this in slightly more formal terms: there exists support for `content: none` for any use whatsoever and there exists support for using `content: none` for regular elements, which is a subset of any use whatsoever. The PR covers only the subset, but it does so using the dotted path (`css.properties.content.none`) that we would ordinarily use to refer to the superset.

rachelandrew:
Maybe this is the confusion, `::before` is the pseudo-element. So your first two rulesets are both targetting pseudo-elements, and both have the same support. There is no special firefox behavior for a pseudo-element applied to an id.

https://developer.mozilla.org/en-US/docs/Web/CSS/::before

So we are only talking about two things, not three. The behavior of `content: none` where the selector is a pseudo-element and the behavior where the selector is a regular element in the DOM.

ddbeck:
> So we are only talking about two things, not three. The behavior of `content: none` where the selector is a pseudo-element and the behavior where the selector is a regular element in the DOM.

Yes, that's exactly right ([here's a code pen to demonstrate that I know how `::before` works](https://codepen.io/ddbeck/pen/ZEKxNYv)). This PR only covers one of these things (the latter). What I'm asking for is that the dotted path to this data (i.e., `css.properties.content.something`) clearly indicates which of the two things it covers.

rachelandrew:
I'm just trying to find out why what I'm proposing is incorrect. As far as I can tell from your suggestion, you think there is some status where "none" doesn't exist. I'm trying to explain that it does and has always existed (or as far as we care about anyway), or to find out in what state you think it doesn't exist. Maybe I'm wrong, maybe there is such a situation, I just don't know of it.

Going back to your initial question, "I have a question which might impact the structure here: is none itself widely supported?"

Yes, it is widely supported. On pseudo-elements. Not on elements.

I'm not arguing with you, I just don't understand, as to my mind what you are suggesting doesn't map to my understanding of the support for this feature so I don't know how to make the change.

When I added this I was thinking about how we indicate support for other partial implementations, for example, the `gap` property. In that case, we list it under support in a `grid_context` and so on. Can you explain why this partial implementation is different from those partial implementations? Maybe that's the issue, I see this as being equivalent to that.

ddbeck:
> I'm just trying to find out why what I'm proposing is incorrect.

Generally, this is about consistency: I think that if I'm faced with a property that has some keyword values, then I should use the literal keys for those keyword values, to represent the longest-standing standard behavior.

> As far as I can tell from your suggestion, you think there is some status where "none" doesn't exist. I'm trying to explain that it does and has always existed (or as far as we care about anyway), or to find out in what state you think it doesn't exist. Maybe I'm wrong, maybe there is such a situation, I just don't know of it.

My thinking isn't what matters here. BCD doesn't have _data_ that shows, either way, whether `none` was supported ever since ~time immemorial~ `content` was supported. If BCD _did_ have that data, I would expect it to be in the feature `css.properties.content.none`, which collides with this PR.

> Going back to your initial question, "I have a question which might impact the structure here: is none itself widely supported?"
> 
> Yes, it is widely supported. On pseudo-elements. Not on elements.
> 
> I'm not arguing with you, I just don't understand, as to my mind what you are suggesting doesn't map to my understanding of the support for this feature so I don't know how to make the change.

Suppose you already had version information for the pseudo-elements behavior, in addition the element behavior you're adding already. Where would you put that data, yourself? What feature would you add, what would you name it, and what would its parents/siblings/children be? Would any of that conflict with the data you've already constructed for this PR? Knowing how you would accommodate more data might help me understand where you're coming from here. 

My answer to those questions is basically the outline in https://github.com/mdn/browser-compat-data/pull/11574#issuecomment-889138032. I don't mind that we don't actually have the data for `none`-for-pseudo-elements but I do not want to foreclose on adding it in the future. If `none`-for-regular-elements has a distinct dotted path, then there's still room for the other data, someday.

> When I added this I was thinking about how we indicate support for other partial implementations, for example, the `gap` property. In that case, we list it under support in a `grid_context` and so on. Can you explain why this partial implementation is different from those partial implementations? Maybe that's the issue, I see this as being equivalent to that.

I'm not sure it's helpful to think of these as partial implementations. In part, we're not using [the `partial_implementation` attribute](https://github.com/mdn/browser-compat-data/blob/main/schemas/compat-data-schema.md#partial_implementation) and because there's no incompatibility between any generally available browser today.

For a number of reasons, `grid_context` etc. aren't a good model either. Mainly, one-off namespaces that we don't document make for a confusing API to BCD (e.g., `css.properties.gap` doesn't have any direct compat data unlike most other properties 😞). That said, `gap` with children `grid_context` and `flex_context` do achieve the aim of explicitly separating the grid and flex behaviors in the dotted path name; this PR doesn't satisfy that yet, for pseudo-element and regular element behaviors of `content: none`.

css/properties/content.json
```diff
@@ -145,6 +145,66 @@
             }
           }
         },
+        "none": {
```
rachelandrew, line 148:
```suggestion
        "none_applies_to_elements": {
```

rachelandrew:
I've renamed the feature as per your original suggestion, as I still don't really understand why you are making a special case out of this one. The `none` value on pseudos dates back to the original pseudo spec, so support is pretty likely to be the same as support for the property itself. We don't typically break out every value of a property, we just mention new ones which is what I thought I was doing.

ddbeck:
Thanks, Rachael. I will merge this.

> We don't typically break out every value of a property, we just mention new ones which is what I thought I was doing.

It's true that we don't typically break out every value, but I want to avoid shadowing the names of values that do exist with other data. I'm running out of ways to write this out, so if you want to discuss this any further, please email me (or send a message on Slack/Matrix/etc.) and we can schedule a time to talk about it synchronously.

PR #11576, opened 2021-07-19
Add missing api.HTMLSlotElement.assign feature

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing  member of the HTMLSlotElement API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.10).

Spec: https://html.spec.whatwg.org/multipage/scripting.html#dom-slot-assign

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/HTMLSlotElement/assign

### DIFF
--- a/api/HTMLSlotElement.json
+++ b/api/HTMLSlotElement.json
@@ -90,7 +90,7 @@
             }
           },
           "status": {
-            "experimental": false,
+            "experimental": true,
             "standard_track": true,
             "deprecated": false
           }
--- a/api/HTMLSlotElement.json
+++ b/api/HTMLSlotElement.json
@@ -48,6 +48,54 @@
           "deprecated": false
         }
       },
+      "assign": {
+        "__compat": {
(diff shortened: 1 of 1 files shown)


api/HTMLSlotElement.json
```diff
@@ -48,6 +48,54 @@
...
+            },
+            "samsunginternet_android": {
+              "version_added": "14.0"
+            },
+            "webview_android": {
+              "version_added": "86"
+            }
+          },
+          "status": {
+            "experimental": false,
```
foolip, line 93: Should be true since this is single-engine and less than two years old.

foolip (changes requested):

foolip (approved):

PR #11579, opened 2021-07-19
Update Edge support for Fullscreen API

foolip:
Since Edge 12 is not available on BrowserStack or Sauce, Web API
Confluence was used for this:
http://web-confluence.appspot.com/#!/catalog?releases=%5B%22Edge_12.10240_Windows_10.0%22,%22Edge_13.10586_Windows_10.0%22,%22Edge_14.14393_Windows_10.0%22,%22Edge_15.15063_Windows_10.0%22,%22Edge_16.16299_Windows_10.0%22,%22Edge_17.17134_Windows_10.0%22,%22Edge_18.17763_Windows_10.0%22%5D&q=%22fullscreen%22

In short, the unprefixed API was in Edge 12-13, and then removed due to
web compat issues. The webkit-prefixed API remained.

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -4528,9 +4528,19 @@
                 "prefix": "webkit"
               }
             ],
-            "edge": {
-              "version_added": "12"
-            },
+            "edge": [
+              {
+                "version_added": "79"
+              },
+              {
+                "version_added": "12",
+                "prefix": "webkit"
+              },
+              {
+                "version_added": "12",
+                "version_removed": "14"
+              }
+            ],
             "firefox": [
               {
                 "version_added": "64"
@@ -5212,7 +5222,7 @@
                 "version_added": "79"
               },
               {
-                "version_added": "≤79",
+                "version_added": "12",
                 "alternative_name": "webkitIsFullScreen"
               }
             ],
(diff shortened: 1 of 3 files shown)


api/Document.json
```diff
@@ -5212,7 +5222,7 @@
                 "version_added": "79"
               },
               {
-                "version_added": "≤79",
+                "version_added": "12",
```
foolip, line 5225: Note that `document.fullscreen` was added to the spec as Firefox was trying to ship the unprefixed API and ran into some compat issues. It didn't exist at the time of Edge 12, so it's omission here is correct.

api/Element.json
```diff
@@ -6380,12 +6420,11 @@
                 "version_added": "79"
               },
               {
-                "version_added": "79",
+                "version_added": "12",
```
foolip, line 6423: This entry was just a bit messed up, these edits make it consistent with the other changes in this PR.

vinyldarkscratch (approved):
It's odd for Edge to have removed the non-prefixed version and kept the prefixed one.  Thanks for researching this!

foolip:
> It's odd for Edge to have removed the non-prefixed version and kept the prefixed one. Thanks for researching this!

What happened was that Edge ran into site compat issues. Because the fullscreen API was prefixed-only for such a long time, sites accidentally found ways of depending on this and breaking if the unprefixed API was supported, among other things by having typos in code paths that were only hit when the unprefixed API was supported. This forced the Edge team to disable it. There was no problem supported in the prefixed API, however.

We had to work through a lot of stuff like this to get the unprefixed API shipped in Chrome too.

PR #11582, opened 2021-07-19
Add missing features for HTMLSourceElement API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing features of the HTMLSourceElement API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.10).

Spec: https://html.spec.whatwg.org/multipage/embedded-content-other.html#dom-dim-height / https://html.spec.whatwg.org/multipage/embedded-content-other.html#dom-dim-width

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/HTMLSourceElement

### DIFF
--- a/api/HTMLSourceElement.json
+++ b/api/HTMLSourceElement.json
@@ -48,6 +48,54 @@
           "deprecated": false
         }
       },
+      "height": {
+        "__compat": {
+          "spec_url": "https://html.spec.whatwg.org/multipage/embedded-content-other.html#dom-dim-height",
+          "support": {
+            "chrome": {
+              "version_added": "90"
+            },
+            "chrome_android": {
+              "version_added": "90"
+            },
+            "edge": {
+              "version_added": "90"
+            },
+            "firefox": {
+              "version_added": false
+            },
+            "firefox_android": {
+              "version_added": false
+            },
+            "ie": {
+              "version_added": false
+            },
+            "opera": {
+              "version_added": "76"
+            },
+            "opera_android": {
+              "version_added": "64"
+            },
+            "safari": {
+              "version_added": false
+            },
+            "safari_ios": {
+              "version_added": false
+            },
+            "samsunginternet_android": {
+              "version_added": "15.0"
+            },
+            "webview_android": {
+              "version_added": "90"
+            }
+          },
+          "status": {
+            "experimental": false,
+            "standard_track": true,
+            "deprecated": false
+          }
+        }
+      },
       "keySystem": {
         "__compat": {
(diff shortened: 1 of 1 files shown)


api/HTMLSourceElement.json
```diff
@@ -48,6 +48,54 @@
           "deprecated": false
         }
       },
+      "height": {
+        "__compat": {
+          "spec_url": "https://html.spec.whatwg.org/multipage/embedded-content-other.html#dom-dim-height",
```
foolip, line 53: @sideshowbarker do you think we should use this URL or https://html.spec.whatwg.org/multipage/embedded-content-other.html#attr-dim-height here?

api/HTMLSourceElement.json
```diff
@@ -48,6 +48,54 @@
...
+            },
+            "samsunginternet_android": {
+              "version_added": "15.0"
+            },
+            "webview_android": {
+              "version_added": "90"
+            }
+          },
+          "status": {
+            "experimental": false,
```
foolip, line 93:
```suggestion
            "experimental": true,
```

api/HTMLSourceElement.json
```diff
@@ -346,6 +394,54 @@
...
```
foolip, line 440:
```suggestion
            "experimental": true,
```

api/HTMLSourceElement.json
sideshowbarker: > @sideshowbarker do you think we should use this URL or [html.spec.whatwg.org/multipage/embedded-content-other.html#attr-dim-height](https://html.spec.whatwg.org/multipage/embedded-content-other.html#attr-dim-height) here?

I think that because this is for the API, we shouldn’t use the one for content/markup attribute. So that seems to leave us with https://html.spec.whatwg.org/multipage/embedded-content-other.html#dom-dim-height (the existing spec URL in this PR).

In general we have quite a number of these cases where DOM/IDL attributes, the anchor/target text just says, *“This should reflect the content attribute.”* But arguably the fact that the spec doesn’t state any requirements beyond just that is a significant piece of information. In other words, it’s useful for developers to know that this is a case where there are no requirements beyond just reflecting the content attribute.

foolip: Thanks Mike, that makes sense. I actually missed that the https://html.spec.whatwg.org/multipage/embedded-content-other.html#attr-dim-height doesn't say "IDL attribute" and assumed they were talking about the same thing.

foolip (approved):

PR #11589, opened 2021-07-19
Add missing features for Path2D API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing features of the Path2D API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.10).

Spec: https://html.spec.whatwg.org/multipage/canvas.html#canvaspath

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Path2D

### DIFF
--- a/api/Path2D.json
+++ b/api/Path2D.json
@@ -147,6 +147,438 @@
             "deprecated": false
           }
         }
+      },
+      "arc": {
+        "__compat": {
+          "spec_url": "https://html.spec.whatwg.org/multipage/canvas.html#dom-context-2d-arc-dev",
+          "support": {
+            "chrome": {
+              "version_added": "36"
+            },
+            "chrome_android": {
+              "version_added": "36"
+            },
+            "edge": {
+              "version_added": "14"
+            },
+            "firefox": {
+              "version_added": "31"
+            },
+            "firefox_android": {
+              "version_added": "31"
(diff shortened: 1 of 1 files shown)


foolip (approved):

api/Path2D.json
```diff
@@ -147,6 +147,438 @@
...
+              "version_added": false
+            },
+            "opera": {
+              "version_added": "23"
+            },
+            "opera_android": {
+              "version_added": "24"
+            },
+            "safari": {
+              "version_added": "8"
```
foolip, line 180: It's weird that none of the members of the interface are marked as supported in Safari 7 but the interface itself is. Can you check if that's really correct?

vinyldarkscratch: It appears that I was the one who set this to Safari 7.  This was probably an issue due to the confusion regarding Safari 7.1 (it was in #7950, which was last year, ergo we had no idea Safari 7.1 was an 8.0 backport).  I'll update this in a quick follow-up PR!

foolip: Ah, I see, thanks!

foolip (approved):

PR #11591, opened 2021-07-19
Add missing api.Performance.measureUserAgentSpecificMemory feature

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing `measureUserAgentSpecificMemory` member of the Performance API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.10).

Spec: https://wicg.github.io/performance-measure-memory/#ref-for-dom-performance-measureuseragentspecificmemory⑤

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Performance/measureUserAgentSpecificMemory

### DIFF
--- a/api/Performance.json
+++ b/api/Performance.json
@@ -699,6 +699,54 @@
           }
         }
       },
+      "measureUserAgentSpecificMemory": {
+        "__compat": {
+          "spec_url": "https://wicg.github.io/performance-measure-memory/#ref-for-dom-performance-measureuseragentspecificmemory⑤",
+          "support": {
+            "chrome": {
+              "version_added": "89"
+            },
+            "chrome_android": {
+              "version_added": "89"
+            },
+            "edge": {
+              "version_added": "89"
+            },
+            "firefox": {
+              "version_added": false
+            },
+            "firefox_android": {
+              "version_added": false
+            },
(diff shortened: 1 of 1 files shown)


api/Performance.json
```diff
@@ -699,6 +699,54 @@
...
+            },
+            "samsunginternet_android": {
+              "version_added": false
+            },
+            "webview_android": {
+              "version_added": "89"
+            }
+          },
+          "status": {
+            "experimental": false,
```
foolip, line 744: Should be true.

api/Performance.json
```diff
@@ -699,6 +699,54 @@
           }
         }
       },
+      "measureUserAgentSpecificMemory": {
+        "__compat": {
+          "spec_url": "https://wicg.github.io/performance-measure-memory/#ref-for-dom-performance-measureuseragentspecificmemory⑤",
```
foolip, line 704: Yep, this is right even though the URL is weird.

foolip (changes requested):

foolip (approved):

PR #11594, opened 2021-07-19
Add missing api.PopStateEvent.PopStateEvent feature

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing `constructor` member of the PopStateEvent API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.10).

Spec: https://html.spec.whatwg.org/multipage/browsing-the-web.html#the-popstateevent-interface

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/PopStateEvent/PopStateEvent

### DIFF
--- a/api/PopStateEvent.json
+++ b/api/PopStateEvent.json
@@ -48,6 +48,55 @@
           "deprecated": false
         }
       },
+      "PopStateEvent": {
+        "__compat": {
+          "description": "<code>PopStateEvent()</code> constructor",
+          "spec_url": "https://html.spec.whatwg.org/multipage/browsing-the-web.html#the-popstateevent-interface",
+          "support": {
+            "chrome": {
+              "version_added": "16"
+            },
+            "chrome_android": {
+              "version_added": "18"
+            },
+            "edge": {
+              "version_added": "14"
+            },
+            "firefox": {
+              "version_added": "11"
+            },
+            "firefox_android": {
+              "version_added": "14"
+            },
+            "ie": {
+              "version_added": false
+            },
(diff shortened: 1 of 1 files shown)


api/PopStateEvent.json
```diff
@@ -48,6 +48,55 @@
           "deprecated": false
         }
       },
+      "PopStateEvent": {
+        "__compat": {
+          "description": "<code>PopStateEvent()</code> constructor",
+          "spec_url": "https://html.spec.whatwg.org/multipage/browsing-the-web.html#the-popstateevent-interface",
```
foolip, line 54: @sideshowbarker for constructors with nothing in particular to link to, do you think we should do this, or just omit the spec URL?

sideshowbarker: > @sideshowbarker for constructors with nothing in particular to link to, do you think we should do this

I have a slight preference for doing that — though I agree it’s not a very friendly place to drop developers into.

> or just omit the spec URL?

If we omit the spec URL completely, then we’re not giving developers any help at all if they want to confirm what the spec says are the arguments for the constructor.

The IDL at least minimally shows what the arguments for the constructor are, so it seems like that makes the spec link marginally useful.

foolip: OK, so let's link to the most useful thing we can find then, if the spec defines the constructor at all.

foolip (approved):

PR #11597, opened 2021-07-19
Add missing api.SVGAElement.href feature

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing `href` member of the SVGAElement API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.10).

Spec: https://html.spec.whatwg.org/multipage/links.html#dom-hyperlink-href

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/SVGAElement/href

### DIFF
--- a/api/SVGAElement.json
+++ b/api/SVGAElement.json
@@ -95,6 +95,54 @@
           }
         }
       },
+      "href": {
+        "__compat": {
+          "spec_url": "https://html.spec.whatwg.org/multipage/links.html#dom-hyperlink-href",
+          "support": {
+            "chrome": {
+              "version_added": "1"
+            },
+            "chrome_android": {
+              "version_added": "18"
+            },
+            "edge": {
+              "version_added": "12"
+            },
+            "firefox": {
+              "version_added": "4"
+            },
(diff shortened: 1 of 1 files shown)


api/SVGAElement.json
```diff
@@ -95,6 +95,54 @@
           }
         }
       },
+      "href": {
+        "__compat": {
+          "spec_url": "https://html.spec.whatwg.org/multipage/links.html#dom-hyperlink-href",
```
foolip, line 100:
```suggestion
          "spec_url": "https://svgwg.org/svg2-draft/single-page.html#types-__svg__SVGURIReference__href",
```

The `href` property defined in HTML is a `DOMString`, while this is an `SVGAnimatedString`. See also https://github.com/w3c/svgwg/issues/312 for some weirdness with how this property is defined. There's no weirdness in implementation though, so we can add this in BCD no problem.

foolip (approved):

PR #11599, opened 2021-07-19
Add missing features for SVGElement API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing features of the SVGElement API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.1.10).

### DIFF
--- a/api/SVGElement.json
+++ b/api/SVGElement.json
@@ -347,6 +347,54 @@
           }
         }
       },
+      "nonce": {
+        "__compat": {
+          "spec_url": "https://html.spec.whatwg.org/multipage/urls-and-fetching.html#dom-noncedelement-nonce",
+          "support": {
+            "chrome": {
+              "version_added": "61"
+            },
+            "chrome_android": {
+              "version_added": "61"
+            },
+            "edge": {
+              "version_added": "79"
+            },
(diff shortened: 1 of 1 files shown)


api/SVGElement.json
```diff
@@ -347,6 +347,54 @@
           }
         }
       },
+      "nonce": {
+        "__compat": {
+          "spec_url": "https://html.spec.whatwg.org/multipage/urls-and-fetching.html#dom-noncedelement-nonce",
```
foolip, line 352: @sideshowbarker can you tell why the domintro block just above this doesn't have an ID?

sideshowbarker: > @sideshowbarker can you tell why the domintro block just above this doesn't have an ID?

I don’t know — but I’ll take a look at the spec source and what wattsi is doing with it, and figure out if I can manage to get an ID generated for it.

foolip: Thanks! I'll go ahead and merge it like this in the meantime.

foolip (approved):

PR #11610, opened 2021-07-20
Add missing features for SVGTransformList API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing features of the SVGTransformList API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.1).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/SVGTransformList

### DIFF
--- a/api/SVGTransformList.json
+++ b/api/SVGTransformList.json
@@ -48,6 +48,342 @@
           "deprecated": false
         }
       },
+      "appendItem": {
+        "__compat": {
+          "spec_url": "https://svgwg.org/svg2-draft/types.html#__svg__SVGNameList__appendItem",
+          "support": {
+            "chrome": {
+              "version_added": "5"
+            },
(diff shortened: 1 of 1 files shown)


api/SVGTransformList.json
```diff
@@ -48,6 +48,342 @@
           "deprecated": false
         }
       },
+      "appendItem": {
+        "__compat": {
+          "spec_url": "https://svgwg.org/svg2-draft/types.html#__svg__SVGNameList__appendItem",
```
foolip, line 53: This looks wrong but is right, `SVGNameList` is a sort of template for lots of similar interfaces.

foolip (approved):

PR #5584, opened 2020-01-27
Rename SpeechRecognitionError to SpeechRecognitionErrorEvent

sideshowbarker:
https://github.com/WICG/speech-api/pull/23 renamed `SpeechRecognitionError` to `SpeechRecognitionErrorEvent`. (No engines other than Blink/Chromium yet implement it as `SpeechRecognitionErrorEvent`, and no others ever implemented `SpeechRecognitionError`.) https://github.com/chromium/chromium/commit/5dbd1fc 

---

cc @jpmedley for confirmation of the Chrome, etc., version numbers — and for additional feedback about whether the patch here is the right way to handle this kind of case.

PR #11613, opened 2021-07-20
Correct data for SpeechRecognitionErrorEvent and add its constructor

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing `SpeechRecognitionErrorEvent` member of the SpeechRecognitionErrorEvent API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.1).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/SpeechRecognitionErrorEvent/SpeechRecognitionErrorEvent

### DIFF
--- a/api/SpeechRecognitionErrorEvent.json
+++ b/api/SpeechRecognitionErrorEvent.json
@@ -62,10 +62,12 @@
           "spec_url": "https://wicg.github.io/speech-api/#dom-speechrecognitionerrorevent-speechrecognitionerrorevent",
           "support": {
             "chrome": {
-              "version_added": false
+              "alternative_name": "webkitSpeechRecognitionError",
+              "version_added": "33"
             },
             "chrome_android": {
-              "version_added": false
+              "alternative_name": "webkitSpeechRecognitionError",
+              "version_added": "33"
             },
             "edge": {
               "version_added": false
@@ -92,10 +94,12 @@
               "version_added": "14.5"
             },
             "samsunginternet_android": {
-              "version_added": false
+              "alternative_name": "webkitSpeechRecognitionError",
+              "version_added": "2.0"
             },
             "webview_android": {
-              "version_added": false
+              "alternative_name": "webkitSpeechRecognitionError",
+              "version_added": "33"
             }
           },
           "status": {
--- a/api/SpeechRecognitionErrorEvent.json
+++ b/api/SpeechRecognitionErrorEvent.json
@@ -8,12 +8,12 @@
           "chrome": {
             "alternative_name": "webkitSpeechRecognitionError",
             "version_added": "33",
-            "notes": "Since Chrome 77, you'll need to serve your code through a web server for recognition to work."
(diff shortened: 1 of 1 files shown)


api/SpeechRecognitionErrorEvent.json
```diff
@@ -80,6 +80,55 @@
           "deprecated": false
         }
       },
+      "SpeechRecognitionErrorEvent": {
+        "__compat": {
+          "description": "<code>SpeechRecognitionErrorEvent()</code> constructor",
+          "spec_url": "https://wicg.github.io/speech-api/#dom-speechrecognitionerrorevent-speechrecognitionerrorevent",
+          "support": {
+            "chrome": {
+              "version_added": false
```
foolip, line 89: The oddly named `webkitSpeechRecognitionError` constructor is supported, can you take the results from http://mdn-bcd-collector.appspot.com/tests/api/webkitSpeechRecognitionError/webkitSpeechRecognitionError and put in here?

Note that the data for the interface itself is wrong, it suggests this event interface is no longer prefixed in Chrome, but it is. Can you fix that at the same time?

foolip (changes requested):

api/SpeechRecognitionErrorEvent.json
```diff
@@ -5,28 +5,16 @@
...
-            {
-              "alternative_name": "webkitSpeechRecognitionError",
-              "version_added": "33",
-              "version_removed": "77"
-            }
-          ],
+          "chrome": {
+            "alternative_name": "webkitSpeechRecognitionError",
+            "version_added": "33",
+            "notes": "Since Chrome 77, you'll need to serve your code through a web server for recognition to work."
```
foolip, line 11: This note is all over the speech recognition APIs but I'd like to trim to to just the main entry point for the API.

There actually isn't anything worth noting about Chrome 77. This data was added in https://github.com/mdn/browser-compat-data/pull/5584 with reference to https://github.com/chromium/chromium/commit/5dbd1fc, but all that did was to change the name one sees when doing things like stringifying an instance. `webkitSpeechRecognitionError` is still the only way to get the interface object without an instance.

So let's get rid of the Chrome 77 stuff and the notes.

@sideshowbarker FYI

foolip (changes requested):

api/SpeechRecognitionErrorEvent.json
```diff
@@ -51,35 +39,76 @@
...
+              "alternative_name": "webkitSpeechRecognitionError",
+              "version_added": "2.0"
+            },
+            "webview_android": {
+              "alternative_name": "webkitSpeechRecognitionError",
+              "version_added": "≤37"
+            }
+          },
+          "status": {
+            "experimental": true,
```
foolip, line 106: This matches the parent feature, so alright then. It can be fixed together separately.

foolip (approved):

PR #11620, opened 2021-07-20
Add missing DOMRectList API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing `DOMRectList` API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.1).

### DIFF
--- a/api/DOMRectList.json
+++ b/api/DOMRectList.json
@@ -1,0 +1,148 @@
+{
+  "api": {
+    "DOMRectList": {
+      "__compat": {
+        "spec_url": "https://drafts.fxtf.org/geometry/#DOMRectList",
+        "support": {
+          "chrome": {
+            "version_added": "2"
+          },
+          "chrome_android": {
+            "version_added": "18"
+          },
+          "edge": {
+            "version_added": "12"
+          },
+          "firefox": {
+            "version_added": "3"
+          },
+          "firefox_android": {
+            "version_added": "4"
+          },
+          "ie": {
+            "version_added": "9"
+          },
+          "opera": {
+            "version_added": "≤12.1"
+          },
+          "opera_android": {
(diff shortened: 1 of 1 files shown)


api/DOMRectList.json
```diff
@@ -0,0 +1,148 @@
+{
+  "api": {
+    "DOMRectList": {
+      "__compat": {
+        "spec_url": "https://drafts.fxtf.org/geometry/#DOMRectList",
+        "support": {
+          "chrome": {
+            "version_added": "2"
```
foolip, line 8: https://mdn-bcd-collector.appspot.com/tests/api/DOMRectList papers over this in order to test the attributes on an instance, but the original name for `DOMRectList` was `ClientRectList`. Can you compare to the api.DOMRect entry and add similar ranges? I would expect the version ranges are the same, but it's best to test to confirm.

foolip (changes requested):

api/DOMRectList.json
```diff
@@ -0,0 +1,226 @@
...
+            },
+            {
+              "alternative_name": "ClientRectList",
+              "version_added": "4",
+              "version_removed": "27"
+            }
+          ],
+          "ie": {
+            "alternative_name": "ClientRectList",
+            "version_added": "9"
```
foolip, line 59: https://developer.mozilla.org/en-US/docs/Web/API/Element/getClientRects#browser_compatibility says IE 5 for `element.getClientRects()`. Looks like in IE 8 it returned a `TextRectangleList`. Let's not bother recording that in BCD :)

foolip (approved):

PR #11631, opened 2021-07-20
Add CSSGroupingRule properties

rachelandrew:
CSSGroupingRule was missing properties, this PR adds them and their specdata https://drafts.csswg.org/cssom/#cssgroupingrule

### DIFF
--- a/api/CSSGroupingRule.json
+++ b/api/CSSGroupingRule.json
@@ -47,6 +47,153 @@
           "standard_track": true,
           "deprecated": false
         }
+      },
+      "cssRules": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/CSSGroupingRule/cssRules",
+          "spec_url": "https://drafts.csswg.org/cssom/#dom-cssgroupingrule-cssrules",
+          "support": {
+            "chrome": {
+              "version_added": "45"
+            },
+            "chrome_android": {
+              "version_added": "45"
+            },
+            "edge": {
+              "version_added": "12"
+            },
+            "firefox": {
+              "version_added": "20"
+            },
+            "firefox_android": {
+              "version_added": "20"
+            },
+            "ie": {
+              "version_added": false
+            },
+            "opera": {
+              "version_added": "32"
+            },
+            "opera_android": {
+              "version_added": "32"
+            },
+            "safari": {
+              "version_added": false
+            },
+            "safari_ios": {
+              "version_added": false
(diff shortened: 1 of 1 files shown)


api/CSSGroupingRule.json
```diff
@@ -47,6 +47,153 @@
...
           "deprecated": false
         }
+      },
+      "cssRules": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/CSSGroupingRule/cssRules",
+          "spec_url": "https://drafts.csswg.org/cssom/#dom-cssgroupingrule-cssrules",
+          "support": {
+            "chrome": {
+              "version_added": "45"
```
foolip, line 57: This could be correct, but due to https://github.com/mdn/browser-compat-data/issues/7844 I'm suspicious of any Chrome 45 versions. What testing was done to get this version? There are generated tests for this in mdn-bcd-collector, but they were broken. I've sent https://github.com/foolip/mdn-bcd-collector/pull/1323 to fix it and then it should be possible to use these tests:
https://staging-dot-mdn-bcd-collector.appspot.com/tests/api/CSSGroupingRule

rachelandrew: I copied the data from the main interface (that I haven't changed), after searching to see if there was any indication that this had been partly implemented. So if it's wrong, the main interface is likely wrong too.

foolip: Looks like this came from https://bugs.chromium.org/p/chromium/issues/detail?id=496381, and the date *does* line up with Chrome 45. I've confirmed that the `CSSGroupingRule` is there in Chrome 45, but not in 44.

That just shuffled the prototype chain around a little bit and is a case of https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#apis-moved-on-the-prototype-chain, but sorting that out need not block this PR. I'll review.

api/CSSGroupingRule.json
```diff
@@ -47,6 +47,153 @@
...
+              "version_added": false
+            },
+            "opera": {
+              "version_added": "32"
+            },
+            "opera_android": {
+              "version_added": "32"
+            },
+            "safari": {
+              "version_added": false
```
foolip, line 81: Safari does have `CSSGroupingRule`, but the parent feature is wrong too. I've filed https://github.com/mdn/browser-compat-data/issues/11644.

foolip (approved):

PR #11632, opened 2021-07-20
Add Firefox versions for api.Document.exitPointerEvent

vinyldarkscratch:
This PR adds real values for Firefox and Firefox Android for the `exitPointerEvent` member of the `Document` API, based upon manual testing.

Test Code Used: `document.mozExitPointerEvent`

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -7791,12 +7791,26 @@
             "edge": {
               "version_added": "13"
             },
-            "firefox": {
-              "version_added": "50"
-            },
-            "firefox_android": {
-              "version_added": "50"
-            },
+            "firefox": [
+              {
+                "version_added": "50"
+              },
+              {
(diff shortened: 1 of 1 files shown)


foolip:
@vinyldarkscratch do you want to go ahead and tackle all of the entries that have https://w3c.github.io/pointerlock/ as `spec_url`? That would make it easy to tell if everything seems to add up.

foolip:
@vinyldarkscratch just so there's no misunderstanding, there's more stuff (in other files) with https://w3c.github.io/pointerlock/ as the spec URL.

foolip:
But maybe fold this into https://github.com/mdn/browser-compat-data/pull/11457 to have a single PR with everything moz-prefixed for this API?

vinyldarkscratch:
Merged the two PRs together now!

PR #11634, opened 2021-07-20
Remove repeated notes about Speech Recognition requirements

foolip:
These notes don't need to be repeated everywhere. They are left on the
entry point to the API, api.SpeechRecognition.


vinyldarkscratch (approved):
I like notes like this not to be repeated on every member -- this is looking good to me!

PR #11640, opened 2021-07-20
Add missing api.FontFaceSet.forEach feature

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing `forEach` member of the FontFaceSet API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.1).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/FontFaceSet/forEach

### DIFF
--- a/api/FontFaceSet.json
+++ b/api/FontFaceSet.json
@@ -367,7 +367,7 @@
               "version_added": "10"
             },
             "safari_ios": {
-              "version_added": "10.3"
+              "version_added": "10"
             },
             "samsunginternet_android": {
               "version_added": false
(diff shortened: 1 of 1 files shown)


api/FontFaceSet.json
```diff
@@ -336,6 +336,53 @@
...
+              "version_added": false
+            },
+            "opera_android": {
+              "version_added": false
+            },
+            "safari": {
+              "version_added": "10"
+            },
+            "safari_ios": {
+              "version_added": "10.3"
```
foolip, line 370: This is the only "10.3" in the file, is it really correct? Shouldn't it be "10"?

foolip (approved):

PR #11660, opened 2021-07-21
Update RTCDataChannel (spec_url, flags)

teoli2003:
So here is what I did:
- most of the entries are no more experimental (two implementations for years). I removed them
- `id` is supported by Firefox (see https://bugzilla.mozilla.org/show_bug.cgi?id=1531904 that _update_ it in Firefox 68). I updated it to `true` as there are alot of `true` entries for Fx on this interface. I removed the experimental flag there too.
- `onerror` is also supported by Firefox and therefore not experimental (was put in the webidl when the interface was added to Firefox)
- `reliable` is no more on the standard track
- `priority` was missing `spec_url` and its `mdn_url` was 404.

### DIFF
--- a/api/RTCDataChannel.json
+++ b/api/RTCDataChannel.json
@@ -57,7 +57,7 @@
           }
         },
         "status": {
-          "experimental": true,
+          "experimental": false,
           "standard_track": true,
           "deprecated": false
         }
@@ -105,7 +105,7 @@
             }
           },
           "status": {
-            "experimental": true,
+            "experimental": false,
             "standard_track": true,
             "deprecated": false
           }
@@ -154,7 +154,7 @@
             }
           },
           "status": {
-            "experimental": true,
+            "experimental": false,
             "standard_track": true,
             "deprecated": false
           }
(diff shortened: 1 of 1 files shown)


api/RTCDataChannel.json
```diff
@@ -430,10 +430,10 @@
               "version_added": "79"
             },
             "firefox": {
-              "version_added": false
+              "version_added": true
```
ddbeck, line 433: Out of curiosity, I tested this and it works as far back as version 17 (which clashes with the root feature's 18 for DataChannel). Same goes for `onerror`. Not sure whether you want to fight with that oddity in this PR or not.

ddbeck (commented):
Looks fine, with one avenue for building on this, if you care to do so.

api/RTCDataChannel.json
teoli2003: If possible, as there a lot of `true` already for Firefox,  I would like not to fight it now. There is likely a lot of oddities for Firefox. (And I'm only cleaning a bit the content to kill a macro)

PR #11672, opened 2021-07-22
Updates to CSSStyleSheet

rachelandrew:
Adds the spec and MDN URLs to the missing CSSStyleSheet constructor and replace, replaceSync pages.

Also, removed experimental and added to standard as these are in a spec, are in all the Chromium browsers, and are being implemented in Firefox.

While looking up the status of these I discovered that there is a behind a flag implementation in Firefox, so I have updated the Firefox support and linked to the bugs.

### DIFF
--- a/api/CSSStyleSheet.json
+++ b/api/CSSStyleSheet.json
@@ -51,6 +51,8 @@
       "CSSStyleSheet": {
         "__compat": {
           "description": "<code>CSSStyleSheet()</code> constructor",
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/CSSStyleSheet/CSSStyleSheet",
+          "spec_url": "https://drafts.csswg.org/cssom/#dom-cssstylesheet-cssstylesheet",
           "support": {
             "chrome": {
               "version_added": "73"
@@ -62,7 +64,15 @@
               "version_added": "79"
             },
             "firefox": {
-              "version_added": false
+              "version_added": "73",
+              "notes": "See <a href='https://bugzil.la/1520690'>bug 1520690</a>",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "layout.css.constructable-stylesheets.enabled",
+                  "value_to_set": "true"
+                }
+              ]
             },
             "firefox_android": {
               "version_added": false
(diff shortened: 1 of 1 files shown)


foolip:
> Also, removed experimental and added to standard as these are in a spec, are in all the Chromium browsers, and are being implemented in Firefox.

I think that's a correct interpretation of https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#choosing-an-experimental-status since the first "supported by two or more engines" isn't conditional on being enabled by default. @ddbeck is that the intention, that a single-engine implementation that gets another implementation behind a flag immediately becomes non-experimental?

Happily this PR isn't blocked on an answer, because the "feature has been supported by default and without major changes by some browser for two or more years" case also applies here.

api/CSSStyleSheet.json
```diff
@@ -62,7 +64,15 @@
               "version_added": "79"
             },
             "firefox": {
-              "version_added": false
+              "version_added": "73",
```
foolip, line 67: I can't find Firefox 73 in https://bugzil.la/1520690, but Chrome 73 is mentioned. Is this a mixup?

rachelandrew: that's the meta bug, it is in the experimental features list https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Experimental_features#constructable_stylesheets and there are a whole raft of individual issues starting with https://bugzilla.mozilla.org/show_bug.cgi?id=1604296

foolip: I see, thanks! It ultimately doesn't matter too much if the version is wrong for features behind a flag, just thought this might be a mistake.

foolip (approved):

ddbeck:
> I think that's a correct interpretation of https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#choosing-an-experimental-status since the first "supported by two or more engines" isn't conditional on being enabled by default. @ddbeck is that the intention, that a single-engine implementation that gets another implementation behind a flag immediately becomes non-experimental?

@foolip Right now, the exact text requires the two years _and_ that any flagged implementations be compatible. I'd be OK with a proposal making it an _or_, however.

foolip:
That seems OK actually, it seems reasonable to go back to experimental if a second but incompatible implementation comes along. I doubt this scenario is actually going to happen much if at all.

PR #11673, opened 2021-07-22
Adds spec and MDN URL to stringifier.

rachelandrew:
Adding a spec and MDN URL to the CSSStyleValue stringifier, as I'm adding the missing page.

### DIFF
--- a/api/CSSStyleValue.json
+++ b/api/CSSStyleValue.json
@@ -148,6 +148,8 @@
       },
       "toString": {
         "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/CSSStyleValue/toString",
+          "spec_url": "https://drafts.css-houdini.org/css-typed-om/#CSSStyleValue-stringification-behavior",
           "support": {
             "chrome": {
               "version_added": "66"


api/CSSStyleValue.json
```diff
@@ -148,6 +148,8 @@
       },
       "toString": {
         "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/CSSStyleValue/toString",
+          "spec_url": "https://drafts.css-houdini.org/css-typed-om/#CSSStyleValue-stringification-behavior",
```
foolip, line 152: @sideshowbarker what do you think about what's best to link for `toString` methods? https://drafts.css-houdini.org/css-typed-om/#stylevalue-serialization is another candidate here, but I'm not sure if there's a consistent style to follow.

sideshowbarker: > @sideshowbarker what do you think about what's best to link for `toString` methods? [drafts.css-houdini.org/css-typed-om/#stylevalue-serialization](https://drafts.css-houdini.org/css-typed-om/#stylevalue-serialization) is another candidate here, but I'm not sure if there's a consistent style to follow.

I believe https://drafts.css-houdini.org/css-typed-om/#stylevalue-serialization would be better because it provides more information for developers — an actual prose explanation. https://drafts.css-houdini.org/css-typed-om/#CSSStyleValue-stringification-behavior is worse because — although it is technically the most precise anchor to reference — I think to most developers it’s not going to mean much, and maybe just leave them scratching their heads.

The general principle we’ve been trying to consistently follow is, We should link to whatever spec sections provide information written specifically for web developers — rather than spec sections that are intended for implementors and mostly only meaningful to implementors. That’s in keeping with the goal of the spec citation being in part so that developers reading it can confirm that what we’ve got documented in MDN is consistent with what he relevant specs say.

rachelandrew:
```suggestion
          "spec_url": "https://drafts.css-houdini.org/css-typed-om/#stylevalue-serialization",
```

sideshowbarker (approved):

PR #11674, opened 2021-07-22
Add note about Chromium returning Element.

jpmedley:
This supports a content PR: https://github.com/mdn/content/pull/7136

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -2049,13 +2049,16 @@
           "spec_url": "https://dom.spec.whatwg.org/#ref-for-dom-document-createelement①",
           "support": {
             "chrome": {
-              "version_added": "1"
+              "version_added": "1",
+              "notes": "Chromium still returns an Element."
             },
             "chrome_android": {
-              "version_added": "18"
+              "version_added": "18",
+              "notes": "Chromium still returns an Element."
             },
             "edge": {
-              "version_added": "12"
+              "version_added": "12",
+              "notes": "Chromium still returns an Element."
             },
             "firefox": {
               "version_added": "1",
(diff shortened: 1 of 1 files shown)


foolip:
See https://github.com/mdn/content/pull/7136#issuecomment-885681355 for why I think this is incorrect.

api/Document.json
```diff
@@ -2049,13 +2049,16 @@
           "spec_url": "https://dom.spec.whatwg.org/#ref-for-dom-document-createelement①",
           "support": {
             "chrome": {
-              "version_added": "1"
+              "version_added": "1",
+              "notes": "Chromium still returns an Element."
```
ddbeck, line 2053: Setting aside the correctness discussion on the content PR, this note isn't descriptive enough to stand alone (on caniuse and other non-MDN contexts). "Still" with respect to when or what? If Chrome etc. are doing something non-standard or otherwise surprising, the notes should spell that out.

Additionally, avoid referencing browser engines in notes and use the actual browser's name instead (i.e., Chrome not Chromium, Edge not Chromium, and so on).

ddbeck (changes requested):

jpmedley:
@foolip is correct. The IDL is actually irrelevant to what's returned. 

Side note. I"m surprised there aren't more instances like this. The IDL as a shorthand stands independent from the actual implementation. In Chrome, it's used to autogenerate .cpp and .h files.

foolip:
> Side note. I"m surprised there aren't more instances like this. The IDL as a shorthand stands independent from the actual implementation. In Chrome, it's used to autogenerate .cpp and .h files.

Although the default behavior of the bindings generator has been very different in the past when it comes to type checking, currently it's only where Chromium uses custom bindings or some non-standard extended attributes that one needs to doubt the accuracy.

This isn't a case of anything unusual happening with types, it's just that `HTMLElement` inherits from `Element`, so a return type of `Element` doesn't mean that the returned object *isn't* an `HTMLElement`, just that it *is* an `Element`.

ddbeck:
Looking at other things related to this one (particularly https://github.com/mdn/content/pull/7971 and the text highlighting the behavior there), I think the right move here is to close this PR and let the docs explain the typical case.

PR #11677, opened 2021-07-22

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing features of the DOMMatrix API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.2).

### DIFF
--- a/api/DOMMatrix.json
+++ b/api/DOMMatrix.json
@@ -100,7 +100,6 @@
       },
       "a": {
         "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/DOMMatrixReadOnly/a",
           "spec_url": "https://drafts.fxtf.org/geometry/#dom-dommatrixreadonly-a",
           "support": {
             "chrome": {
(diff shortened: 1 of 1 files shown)


api/DOMMatrix.json
```diff
@@ -98,10 +98,10 @@
           }
         }
       },
-      "fromFloat32Array": {
+      "a": {
         "__compat": {
-          "spec_url": "https://drafts.fxtf.org/geometry/#dom-dommatrix-fromfloat32array",
-          "description": "<code>fromFloat32Array()</code> static method",
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/DOMMatrixReadOnly/a",
```
foolip, line 103: Can you make sure `mdn_url` and `spec_url` don't contain "readonly"?

foolip (changes requested):

api/DOMMatrix.json
vinyldarkscratch: The spec URL must contain the `readonly` segment, but I've removed the MDN URLs since they're 404s anyways.

foolip: I see, there's just one definition, and it actually says what to do for both 👍

foolip (approved):

PR #11678, opened 2021-07-22
Add spec URLs for a-f/mXX attributes of DOMMatrixReadOnly

vinyldarkscratch:
This PR adds the spec URLs for the `a` through `f` and `mXX` attributes of the DOMMatrixReadOnly API.


foolip (approved):
I checked a few, they worked!

Issue #8745, opened 2021-01-13
Sort out data for autofocus


foolip:
This is very similar to https://github.com/mdn/browser-compat-data/issues/7519, but need not be solved together with that issue.

Per [spec](https://html.spec.whatwg.org/#htmlorsvgelement), there should be `autofocus` members on `HTMLElement` and `SVGElement`. BCD instead has it on `HTMLButtonElement`, `HTMLInputElement` and `HTMLSelectElement`. Probably it was on those interfaces in implementations originally as well, but has migrated to `HTMLElement` at least in Chromium.

This will need careful investigation of when `autofocus` was supported on which types of HTML elements, and potentially notes.

https://github.com/mdn/browser-compat-data/pull/8112 was a PR to add api.HTMLElement.autofocus, but it should be done together with the removal of the others.

foolip:
In https://github.com/mdn/browser-compat-data/pull/8112/files#r548334335 @vinyldarkscratch found that the move happened in [Chrome 79](https://storage.googleapis.com/chromium-find-releases-static/0fb.html#0fbf9bdd80bfef802dcc57ab0c295a95a3fa39b3).

foolip:
I've just discovered that the attributes is additionally on `HTMLTextAreaElement` in Firefox and Safari, and one `SVGGraphicsElement` in Firefox.

jculver050:
Close

foolip:
https://github.com/mdn/browser-compat-data/pull/11680 sorted out the mess. I'm not sure the "Supported for <code>HTMLButtonElement</code>, <code>HTMLInputElement</code>, <code>HTMLSelectElement</code>, and <code>HTMLTextAreaElement</code>" notes are 100% accurate going all the way back to the first versions claimed to have partial support, but it's more than good enough.

PR #11680, opened 2021-07-22
Add missing autofocus feature

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing `autofocus` member of the HTMLElement and SVGElement APIs, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.1).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/HTMLElement/autofocus

### DIFF
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -490,42 +490,108 @@
         "__compat": {
           "spec_url": "https://html.spec.whatwg.org/multipage/interaction.html#dom-fe-autofocus",
           "support": {
-            "chrome": {
-              "version_added": "79"
-            },
-            "chrome_android": {
-              "version_added": "79"
-            },
-            "edge": {
-              "version_added": "79"
-            },
+            "chrome": [
+              {
+                "version_added": "79"
+              },
+              {
+                "version_added": "1",
+                "version_removed": "79",
+                "partial_implementation": true,
+                "notes": "Before Chrome 79, this attribute was implemented on <code>HTMLButtonElement</code>, <code>HTMLInputElement</code>, <code>HTMLSelectElement</code>, and <code>HTMLTextAreaElement</code>."
+              }
+            ],
+            "chrome_android": [
+              {
+                "version_added": "79"
+              },
+              {
+                "version_added": "18",
+                "version_removed": "79",
+                "partial_implementation": true,
+                "notes": "Before Chrome 79, this attribute was implemented on <code>HTMLButtonElement</code>, <code>HTMLInputElement</code>, <code>HTMLSelectElement</code>, and <code>HTMLTextAreaElement</code>."
+              }
+            ],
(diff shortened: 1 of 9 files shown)


foolip:
Oh, this runs into https://github.com/mdn/browser-compat-data/issues/8745. Can you try resolving all of that mess?

foolip:
For the avoidance of doubt, I'd also like to see other "autofocus" entries removed to resolve #8745.

api/HTMLElement.json
```diff
@@ -490,42 +490,108 @@
...
-            },
+            "chrome": [
+              {
+                "version_added": "79"
+              },
+              {
+                "version_added": "1",
+                "version_removed": "79",
+                "partial_implementation": true,
+                "notes": "Before Chrome 79, this attribute was implemented on <code>HTMLButtonElement</code>, <code>HTMLInputElement</code>, <code>HTMLSelectElement</code>, and <code>HTMLTextAreaElement</code>."
```
foolip, line 501: When the note is on a statement with a removal version, I don't think it's necessary to mention the browser and version. Also we should say "property" instead of "attribute", but I suggest a simpler wording that avoid the issue:

```suggestion
                "notes": "Supported for <code>HTMLButtonElement</code>, <code>HTMLInputElement</code>, <code>HTMLSelectElement</code>, and <code>HTMLTextAreaElement</code>."
```

foolip:
There are also "autofocus" entries in html/elements/ which should be replaced by one in html/global_attributes.json with https://html.spec.whatwg.org/multipage/interaction.html#the-autofocus-attribute as the spec URL.

That would fix the broken BCD in https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autofocus#browser_compatibility :)

foolip (approved):
I'm pretty sure that if one digs in here one could find that support for `autofocus` was added for different HTML elements at different times, but that does not seems worthwhile to me.

foolip:
> I'm pretty sure that if one digs in here one could find that support for `autofocus` was added for different HTML elements at different times, but that does not seems worthwhile to me.

I couldn't resist and confirmed these were in Firefox 4 and Chrome 15 at least:
http://mdn-bcd-collector.appspot.com/tests/api/HTMLButtonElement/autofocus
http://mdn-bcd-collector.appspot.com/tests/api/HTMLInputElement/autofocus
http://mdn-bcd-collector.appspot.com/tests/api/HTMLSelectElement/autofocus
http://mdn-bcd-collector.appspot.com/tests/api/HTMLTextAreaElement/autofocus

If the data is slightly wrong, it's at most slightly wrong even further back than that. Safari not tested since Chrome was WebKit back then.

PR #11681, opened 2021-07-22
Add missing api.Navigator.storage feature

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing `storage` member of the Navigator API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.1).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Navigator/storage

### DIFF
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -3134,7 +3134,7 @@
       },
       "storage": {
         "__compat": {
-          "mdn_url": "https://developer.mozilla.org/en-US/docs/Web/API/Navigator/storage",
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/Navigator/storage",
           "spec_url": "https://storage.spec.whatwg.org/#dom-navigatorstorage-storage",
           "support": {
             "chrome": {
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -3132,6 +3132,54 @@
           }
         }
       },
+      "storage": {
+        "__compat": {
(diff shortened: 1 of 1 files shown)


api/Navigator.json
```diff
@@ -3132,6 +3132,54 @@
           }
         }
       },
+      "storage": {
+        "__compat": {
```
foolip, line 3136: Can you link https://developer.mozilla.org/en-US/docs/Web/API/Navigator/storage?

foolip (changes requested):

api/Navigator.json
```diff
@@ -3132,6 +3132,55 @@
...
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/Navigator/storage",
+          "spec_url": "https://storage.spec.whatwg.org/#dom-navigatorstorage-storage",
+          "support": {
+            "chrome": {
+              "version_added": "55"
```
foolip, line 3141: This doesn't match https://developer.mozilla.org/en-US/docs/Web/API/StorageManager#browser_compatibility, can you investigate?

vinyldarkscratch: It appears the collector changes the version from 48 to 55 for the StorageManager.  I'll send a PR to correct the data for StorageManager!

foolip: Thanks!

foolip (approved):

PR #11682, opened 2021-07-22
Add missing api.SVGElement.autofocus feature

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing `autofocus` member of the SVGElement API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.1).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/SVGElement/autofocus

### DIFF
--- a/api/SVGElement.json
+++ b/api/SVGElement.json
@@ -48,6 +48,54 @@
           "deprecated": false
         }
       },
+      "autofocus": {
+        "__compat": {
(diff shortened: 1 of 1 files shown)


foolip:
Can you fold this into https://github.com/mdn/browser-compat-data/pull/11680 to deal with everything "autofocus" at once?

PR #11685, opened 2021-07-22
Add missing api.BluetoothDevice.ongattserverdisconnected feature

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing `ongattserverdisconnected` member of the BluetoothDevice API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.1).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/BluetoothDevice/ongattserverdisconnected


foolip:
@vinyldarkscratch this should have multiple statements with per-platform support, right? Differences between macOS and Windows are why add-new-bcd couldn't add this originally.

vinyldarkscratch:
Yes, the Bluetooth APIs have differences in support between the OSes, however the data for the API as a whole doesn't have multiple support statements.  I'd like to come back to that and make the appropriate corrections later.

foolip:
If we're not cascading the platform support notes, shouldn't we at least use the earliest versions? Otherwise it will look like this API wasn't supported at all before.

vinyldarkscratch:
Fair point -- I've opened up https://github.com/mdn/browser-compat-data/pull/11701 to take care of the corrections for the API, and will update this PR to match the earliest version.  This will cause the linter to fail until #11701 is merged.

PR #11686, opened 2021-07-22
Add missing api.BluetoothRemoteGATTCharacteristic.oncharacteristicvaluechanged feature

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing `oncharacteristicvaluechanged` member of the BluetoothRemoteGATTCharacteristic API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.1).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/BluetoothRemoteGATTCharacteristic/oncharacteristicvaluechanged

### DIFF
--- a/api/BluetoothRemoteGATTCharacteristic.json
+++ b/api/BluetoothRemoteGATTCharacteristic.json
@@ -308,6 +308,54 @@
           }
         }
       },
+      "oncharacteristicvaluechanged": {
+        "__compat": {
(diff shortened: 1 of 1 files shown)


foolip:
Same question as in https://github.com/mdn/browser-compat-data/pull/11685#issuecomment-885459162.

PR #11687, opened 2021-07-22
Add missing features for WorkerNavigator API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing features of the WorkerNavigator API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.1).

### DIFF
--- a/api/WorkerNavigator.json
+++ b/api/WorkerNavigator.json
@@ -902,10 +902,10 @@
           "spec_url": "https://wicg.github.io/webusb/#dom-workernavigator-usb",
           "support": {
             "chrome": {
-              "version_added": "69"
+              "version_added": "70"
             },
             "chrome_android": {
-              "version_added": "69"
+              "version_added": "70"
             },
             "edge": {
               "version_added": "79"
@@ -920,10 +920,10 @@
               "version_added": false
             },
             "opera": {
-              "version_added": "56"
+              "version_added": "57"
             },
             "opera_android": {
-              "version_added": "48"
+              "version_added": "49"
             },
             "safari": {
               "version_added": false
@@ -935,7 +935,7 @@
               "version_added": "10.0"
             },
             "webview_android": {
-              "version_added": "69"
+              "version_added": "70"
             }
           },
           "status": {
--- a/api/WorkerNavigator.json
+++ b/api/WorkerNavigator.json
@@ -244,6 +244,54 @@
           }
         }
       },
+      "deviceMemory": {
+        "__compat": {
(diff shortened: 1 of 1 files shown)


api/WorkerNavigator.json
```diff
@@ -244,6 +244,54 @@
           }
         }
       },
+      "deviceMemory": {
+        "__compat": {
+          "spec_url": "https://w3c.github.io/device-memory/#sec-device-memory-js-api",
+          "support": {
+            "chrome": {
+              "version_added": "65"
```
foolip, line 252: This doesn't match api.Navigator.deviceMemory, but I've confirmed that shipped in 63 and worker support came in 65.

api/WorkerNavigator.json
```diff
@@ -401,6 +449,102 @@
...
+            "deprecated": false
+          }
+        }
+      },
+      "mediaCapabilities": {
+        "__compat": {
+          "spec_url": "https://w3c.github.io/media-capabilities/#dom-navigator-mediacapabilities",
+          "support": {
+            "chrome": {
+              "version_added": "76"
```
foolip, line 505: Doesn't match api.Navigator.mediaCapabilities, but I've confirmed the difference is real, worker support came later, tracked by https://www.chromestatus.com/feature/5730672977117184.

api/WorkerNavigator.json
```diff
@@ -753,6 +897,54 @@
           }
         }
       },
+      "usb": {
+        "__compat": {
+          "spec_url": "https://wicg.github.io/webusb/#dom-workernavigator-usb",
+          "support": {
+            "chrome": {
+              "version_added": "69"
```
foolip, line 905: https://storage.googleapis.com/chromium-find-releases-static/627.html#6279f700aae1f5edbc51e8dacfabab7efce95d2f says 70 and https://bugs.chromium.org/p/chromium/issues/detail?id=837406#c15 also suggests it missed the 69 branch.

It looks like what happened is that *just* `navigator.usb` accidentally shipped in Chrome 69 in https://bugs.chromium.org/p/chromium/issues/detail?id=837406#c15

Let's mark this as supported for "70" anyway and pretend that little mishap didn't happen:

 ```suggestion
              "version_added": "70"
```

Can you mirror that to other browsers?

foolip (changes requested):

foolip (approved):

PR #11694, opened 2021-07-23
Remove ClientRect alternative name and notes from DOMRectReadOnly

foolip:
This is already represented on DOMRect for all browsers.


vinyldarkscratch (approved):

foolip:
I've sent https://github.com/mdn/browser-compat-data/pull/11960 which effectively reverts this.

PR #11701, opened 2021-07-23
Update Chromium versions for BluetoothDevice API

vinyldarkscratch:
This PR corrects the version numbers for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `BluetoothDevice` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.2).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/BluetoothDevice

### DIFF
--- a/api/BluetoothDevice.json
+++ b/api/BluetoothDevice.json
@@ -5,15 +5,51 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/BluetoothDevice",
         "spec_url": "https://webbluetoothcg.github.io/web-bluetooth/#bluetoothdevice-interface",
         "support": {
-          "chrome": {
-            "version_added": "74"
-          },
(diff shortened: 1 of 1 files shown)


vinyldarkscratch:
Note: the data was copied from the `Bluetooth` API, and confirmed by testing in Chrome 55 and 56 on macOS, and Chrome 69 and 70 on Windows.

foolip (changes requested):
Can you copy the style you used in https://github.com/mdn/browser-compat-data/pull/10197 instead?

vinyldarkscratch:
Well, since we're here, we might as well correct all of the Bluetooth API data!  It seems that the majority of the data we have in BCD has this same issue.  I'll open a new, re-scoped PR to tackle it all at once.

PR #11704, opened 2021-07-23
Add IE versions for api.Element.fullscreen_event

vinyldarkscratch:
This PR adds real values for Internet Explorer for the `fullscreen_event` member of the `Element` API.  The data was copied from their event handler counterparts.

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -2868,7 +2868,7 @@
               }
             ],
             "ie": {
-              "version_added": null
+              "version_added": false
             },
             "opera": [
               {
@@ -2988,7 +2988,7 @@
               }
             ],
             "ie": {
-              "version_added": null
+              "version_added": false
             },
             "opera": [
               {


api/Element.json
```diff
@@ -2868,7 +2868,7 @@
               }
             ],
             "ie": {
-              "version_added": null
+              "version_added": false
```
foolip, line 2871: `Document` has a "MSFullscreenChange" altname for IE11. Can you check if that event was fired on just `Document` or on `Element` and bubbling? If it was fired on elements then this should change.

api/Element.json
```diff
@@ -2988,7 +2988,7 @@
...
```
foolip, line 2991: Same but "MSFullscreenError"

api/Element.json
vinyldarkscratch: As far as I can tell, there appears to be no Element-level support for the fullscreen API in IE.  The Document API defines an `onmsfullscreenchange` handler, but the Element API does not.

foolip: OK, that's enough of a clue, no testing needed!

foolip (approved):

PR #11717, opened 2021-07-24
Remove api.MediaDevices.stereo_audio_capture

vinyldarkscratch:
This PR removes the "stereo audio capture" feature of the MediaDevices API.  As far as I can tell, this is a Firefox-only change, and I haven't been able to find any code in other browsers that suggests the same functionality.  Additionally, there are hardly any consumer-level microphones that contain more than one audio channel.

### DIFF
--- a/api/MediaDevices.json
+++ b/api/MediaDevices.json
@@ -499,54 +499,6 @@
             "deprecated": false
           }
         }
-      },
-      "stereo_audio_capture": {
-        "__compat": {
-          "description": "Stereo audio capture",
-          "support": {
-            "chrome": {
-              "version_added": null
-            },
-            "chrome_android": {
-              "version_added": null
-            },
-            "edge": {
-              "version_added": null
-            },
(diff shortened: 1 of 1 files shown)


foolip:
I would expect that other browsers also support stereo microphones but haven't done any testing.

This data really just amount to saying that older Firefox doesn't support it, and it does not seem worth preserving even as a note IMHO.

Still I'd prefer if someone at Mozilla reviewed this and makes the call.

saschanaz:
It seems this is from https://bugzilla.mozilla.org/show_bug.cgi?id=971528.

ddbeck:
I'm fine with this going away. I'm not quite sure how we'd go about testing this on historic versions. If this really matters to somebody, then we'll need someone with the requisite hardware, interest, and know-how to test this on different browsers for us. To that future audio expert reading this, I welcome your issue or PR. 😄

ddbeck (approved):

PR #11718, opened 2021-07-24
Update Firefox versions for MediaStream API

queengooborg:
This PR updates and corrects the real values for Firefox and Firefox Android for the `MediaStream` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.1).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/MediaStream

### DIFF
--- a/api/MediaStream.json
+++ b/api/MediaStream.json
@@ -27,10 +27,10 @@
             "version_added": "12"
           },
           "firefox": {
-            "version_added": "15"
+            "version_added": "36"
           },
           "firefox_android": {
-            "version_added": "15"
+            "version_added": "36"
           },
           "ie": {
             "version_added": false
@@ -103,7 +103,7 @@
               "version_added": "44"
             },
             "firefox_android": {
-              "version_added": "42"
+              "version_added": "44"
             },
             "ie": {
               "version_added": false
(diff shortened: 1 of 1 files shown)


api/MediaStream.json
```diff
@@ -27,10 +27,10 @@
             "version_added": "12"
           },
           "firefox": {
-            "version_added": "15"
```
foolip, line 30: Have you checked where this data came from, and if it's consistent with APIs that take or return MediaStreams?

Maybe this interface has been prefixed, or the interface object hasn't always been exposed?

foolip: Per https://developer.mozilla.org/en-US/docs/Web/API/Navigator/getUserMedia#browser_compatibility it has to be available since 17 at least.

Also it should probably match MediaStreamTrack.

foolip:
Oh, I see there's a problem with the tests here. They depend on `navigator.mediaDevices.getUserMedia()` but that's much more recent that `MediaStream` itself and `navigator.getUserMedia()` with prefixed variants.

queengooborg:
Closing this PR for now and will open a new one when the tests are updated!

PR #11722, opened 2021-07-24
Remove irrelevant api.Document.popupNode feature

vinyldarkscratch:
This PR removes the irrelevant `popupNode` member of the `Document` API. The documentation for XUL documents has been removed from MDN, creating a broken link, and many other XUL features have been removed back in Firefox 64.

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -9377,55 +9377,6 @@
           }
         }
       },
-      "popupNode": {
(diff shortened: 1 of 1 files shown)


foolip (approved):
Approving but not merging so that https://developer.mozilla.org/docs/Web/API/Document/popupNode can be removed first.

PR #11737, opened 2021-07-25
Add IE/Edge versions for Window API

vinyldarkscratch:
This PR adds real values for Internet Explorer and Edge for the `Window` API.  The data was copied from their event handler counterparts in `api.Window`, `api.WindowEventHandlers`, and `api.GlobalEventHandlers`.

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -309,7 +309,7 @@
             ],
             "edge": [
               {
-                "version_added": "≤79"
+                "version_added": "18"
               },
               {
                 "version_added": "≤79",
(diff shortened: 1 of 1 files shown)


api/Window.json
```diff
@@ -1632,7 +1632,7 @@
               "version_added": true
             },
             "edge": {
-              "version_added": "≤18"
+              "version_added": "12"
```
foolip, line 1635: Can you double check this? I’d be surprised if devicemotion and deviceorientation events were really supported. Would that be for laptops with motion sensors then?

vinyldarkscratch: I have no way to directly test this since I have no mobile Windows device, but this matches with `ondevicemotion`, and the collector agrees with its results.

foolip: OK, let's go with that then :)

foolip (approved):

PR #11745, opened 2021-07-25
Add Chrome/Safari versions for HTMLMediaElement API

queengooborg:
This PR adds real values for Chrome and Safari for the `HTMLMediaElement` API.  The data was copied from their event handler counterparts in `GlobalEventHandlers`.

### DIFF
--- a/api/HTMLMediaElement.json
+++ b/api/HTMLMediaElement.json
@@ -1371,7 +1371,7 @@
               "version_added": "1.0"
             },
             "webview_android": {
-              "version_added": "≤37"
+              "version_added": "1"
             }
           },
           "status": {
@@ -1663,7 +1663,7 @@
               "version_added": "1.0"
             },
             "webview_android": {
-              "version_added": "≤37"
+              "version_added": "1"
             }
           },
           "status": {
--- a/api/HTMLMediaElement.json
+++ b/api/HTMLMediaElement.json
@@ -1362,10 +1362,10 @@
               "version_added": "12"
             },
             "safari": {
-              "version_added": "≤4"
+              "version_added": "1.3"
             },
             "safari_ios": {
-              "version_added": "≤3"
+              "version_added": "1"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
@@ -1630,10 +1630,10 @@
           "spec_url": "https://html.spec.whatwg.org/multipage/media.html#event-media-loadstart",
           "support": {
             "chrome": {
-              "version_added": "32"
+              "version_added": "1"
             },
             "chrome_android": {
-              "version_added": "32"
+              "version_added": "18"
             },
             "edge": {
               "version_added": "12"
@@ -1647,35 +1647,23 @@
             "ie": {
               "version_added": "9"
             },
-            "opera": [
-              {
-                "version_added": "19"
-              },
-              {
-                "version_added": "≤12.1",
-                "version_removed": "15"
-              }
-            ],
-            "opera_android": [
-              {
-                "version_added": "19"
-              },
-              {
-                "version_added": "≤12.1",
-                "version_removed": "14"
-              }
-            ],
+            "opera": {
+              "version_added": "≤12.1"
+            },
+            "opera_android": {
+              "version_added": "≤12.1"
+            },
             "safari": {
-              "version_added": "9"
+              "version_added": "4"
             },
             "safari_ios": {
-              "version_added": "9"
+              "version_added": "3"
             },
             "samsunginternet_android": {
-              "version_added": "2.0"
+              "version_added": "1.0"
             },
             "webview_android": {
-              "version_added": "4.4.3"
+              "version_added": "≤37"
             }
           },
           "status": {
@@ -3105,10 +3093,10 @@
           "spec_url": "https://html.spec.whatwg.org/multipage/media.html#event-media-progress",
           "support": {
             "chrome": {
-              "version_added": "32"
+              "version_added": "1"
             },
             "chrome_android": {
-              "version_added": "32"
+              "version_added": "18"
             },
             "edge": {
               "version_added": "12"
@@ -3122,35 +3110,23 @@
             "ie": {
               "version_added": "9"
             },
-            "opera": [
-              {
-                "version_added": "19"
-              },
-              {
-                "version_added": "≤12.1",
-                "version_removed": "15"
-              }
-            ],
-            "opera_android": [
-              {
-                "version_added": "19"
-              },
-              {
-                "version_added": "≤12.1",
-                "version_removed": "14"
-              }
-            ],
+            "opera": {
+              "version_added": "≤12.1"
+            },
+            "opera_android": {
+              "version_added": "≤12.1"
+            },
             "safari": {
-              "version_added": "9"
+              "version_added": "1.3"
             },
             "safari_ios": {
-              "version_added": "9"
+              "version_added": "1"
             },
             "samsunginternet_android": {
-              "version_added": "2.0"
+              "version_added": "1.0"
             },
             "webview_android": {
-              "version_added": "4.4.3"
+              "version_added": "≤37"
             }
           },
           "status": {
--- a/api/HTMLMediaElement.json
+++ b/api/HTMLMediaElement.json
@@ -3126,7 +3126,7 @@
               "version_added": "1.0"
             },
             "webview_android": {
-              "version_added": "≤37"
+              "version_added": "1"
             }
           },
           "status": {
--- a/api/HTMLMediaElement.json
+++ b/api/HTMLMediaElement.json
@@ -55,10 +55,10 @@
           "spec_url": "https://html.spec.whatwg.org/multipage/media.html#event-media-abort",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "1"
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": "18"
             },
             "edge": {
               "version_added": "12"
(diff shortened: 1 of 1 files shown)


foolip:
Copying data from `GlobalEventHandlers` isn't reliable if there are multiple events by that name, which is the case for "error" and probably others here.

Most of this is probably supported since the initial support for media elements, can you test the events manually to check?

queengooborg:
These are actually already set to the first browser editions when `HTMLMediaElement` was present!

foolip:
Only some of the data matches api.HTMLMediaElement itself. I'll leave a few comments as an example.

api/HTMLMediaElement.json
```diff
@@ -1338,10 +1338,10 @@
           "spec_url": "https://html.spec.whatwg.org/multipage/media.html#event-media-error",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "10"
```
foolip, line 1341: I doubt this is correct. The `MediaError` data says Chrome 1, and even though having `mediaElement.error` (which is a `MediaError`) without the "error" event is possible, I'm pretty sure they went together from the very start.

api/HTMLMediaElement.json
```diff
@@ -1356,22 +1356,22 @@
...
-              "version_added": true
+              "version_added": "11.6"
             },
             "opera_android": {
-              "version_added": null
+              "version_added": "12"
             },
             "safari": {
-              "version_added": null
+              "version_added": "6"
```
foolip, line 1365: This too doesn't seem plausible.

api/HTMLMediaElement.json
```diff
@@ -1630,10 +1630,10 @@
           "spec_url": "https://html.spec.whatwg.org/multipage/media.html#event-media-loadstart",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "32"
```
foolip, line 1633: I'm sure the "loadstart" was supported earlier than this just from memory.

api/HTMLMediaElement.json
```diff
@@ -1647,23 +1647,35 @@
             "ie": {
               "version_added": "9"
             },
-            "opera": {
-              "version_added": true
-            },
-            "opera_android": {
-              "version_added": null
-            },
+            "opera": [
```
foolip, line 1650: This should be simplified if the right version for Chrome is found. No gap in support here.

api/HTMLMediaElement.json
```diff
@@ -3093,10 +3105,10 @@
           "spec_url": "https://html.spec.whatwg.org/multipage/media.html#event-media-progress",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "32"
```
foolip, line 3108: This also seems too late.

foolip (changes requested):

api/HTMLMediaElement.json
foolip: https://trac.webkit.org/changeset/36893/webkit is the patch that introduced the "loadstart" event. That was before the release of Chrome 1 by a few months, so let's assume:

```suggestion
              "version_added": "1"
```

api/HTMLMediaElement.json
```diff
@@ -1630,10 +1630,10 @@
           "spec_url": "https://html.spec.whatwg.org/multipage/media.html#event-media-loadstart",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "32"
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": "32"
```
foolip, line 1636:
```suggestion
              "version_added": "1"
```

api/HTMLMediaElement.json
```diff
@@ -1647,23 +1647,35 @@
...
+                "version_added": "19"
+              },
+              {
+                "version_added": "≤12.1",
+                "version_removed": "14"
+              }
+            ],
             "safari": {
-              "version_added": null
+              "version_added": "9"
```
foolip, line 1669:
```suggestion
              "version_added": "4"
```

Based on WebKit trunk version 528.4:
https://trac.webkit.org/changeset/36893/webkit
https://trac.webkit.org/browser/webkit/trunk/WebCore/Configurations/Version.xcconfig?rev=36893

api/HTMLMediaElement.json
```diff
@@ -1647,23 +1647,35 @@
...
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "9"
```
foolip, line 1672:
```suggestion
              "version_added": "3"
```

foolip (commented):
"loadstart" event tracked down

api/HTMLMediaElement.json
```diff
@@ -1356,22 +1356,22 @@
...
-              "version_added": true
+              "version_added": "11.6"
             },
             "opera_android": {
-              "version_added": null
+              "version_added": "12"
             },
             "safari": {
-              "version_added": null
+              "version_added": "≤4"
```
foolip, line 1365: The "error" event was in the initial media element support in WebKit:
https://trac.webkit.org/browser/webkit/trunk/WebCore/html/HTMLMediaElement.cpp?rev=27277#L337

So:

```suggestion
              "version_added": "1.3"
```

api/HTMLMediaElement.json
```diff
@@ -1356,22 +1356,22 @@
...
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "≤3"
```
foolip, line 1368:
```suggestion
              "version_added": "1"
```

foolip (commented):
"error" event researched...

api/HTMLMediaElement.json
foolip: Yep, it was in the initial WebKit support:
https://trac.webkit.org/browser/webkit/trunk/WebCore/html/HTMLMediaElement.cpp?rev=27277#L452

```suggestion
              "version_added": "1"
```

api/HTMLMediaElement.json
```diff
@@ -3093,10 +3105,10 @@
           "spec_url": "https://html.spec.whatwg.org/multipage/media.html#event-media-progress",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "32"
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": "32"
```
foolip, line 3111:
```suggestion
              "version_added": "18"
```

api/HTMLMediaElement.json
```diff
@@ -3110,23 +3122,35 @@
             "ie": {
               "version_added": "9"
             },
-            "opera": {
-              "version_added": true
-            },
-            "opera_android": {
-              "version_added": null
-            },
+            "opera": [
```
foolip, line 3125: This should be simplified to just `"version_added": "≤12.1"`

api/HTMLMediaElement.json
```diff
@@ -3110,23 +3122,35 @@
...
+                "version_added": "19"
+              },
+              {
+                "version_added": "≤12.1",
+                "version_removed": "14"
+              }
+            ],
             "safari": {
-              "version_added": null
+              "version_added": "9"
```
foolip, line 3144:
```suggestion
              "version_added": "1.3"
```

api/HTMLMediaElement.json
```diff
@@ -3110,23 +3122,35 @@
...
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "9"
```
foolip, line 3147:
```suggestion
              "version_added": "1"
```

api/HTMLMediaElement.json
```diff
@@ -3110,23 +3122,35 @@
...
             },
             "samsunginternet_android": {
-              "version_added": true
+              "version_added": "2.0"
```
foolip, line 3150:
```suggestion
              "version_added": "1.0"
```

api/HTMLMediaElement.json
```diff
@@ -3110,23 +3122,35 @@
...
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "4.4.3"
```
foolip, line 3153:
```suggestion
              "version_added": "1"
```

foolip (commented):
And finally, "progress" event pinned down

api/HTMLMediaElement.json
```diff
@@ -1647,23 +1647,35 @@
...
-              "version_added": null
+              "version_added": "9"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "9"
             },
             "samsunginternet_android": {
-              "version_added": true
+              "version_added": "2.0"
```
foolip, line 1675:
```suggestion
              "version_added": "1"
```

api/HTMLMediaElement.json
```diff
@@ -1647,23 +1647,35 @@
...
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "4.4.3"
```
foolip, line 1678:
```suggestion
              "version_added": "1"
```

api/HTMLMediaElement.json
```diff
@@ -1356,22 +1356,22 @@
...
-              "version_added": null
+              "version_added": "≤3"
             },
             "samsunginternet_android": {
-              "version_added": true
+              "version_added": "1.0"
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "≤37"
```
foolip, line 1374:
```suggestion
              "version_added": "1"
```

api/HTMLMediaElement.json
```diff
@@ -3111,22 +3111,22 @@
...
-              "version_added": null
+              "version_added": "1"
             },
             "samsunginternet_android": {
-              "version_added": true
+              "version_added": "1.0"
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "≤37"
```
foolip, line 3129:
```suggestion
              "version_added": "1"
```

foolip (approved):

PR #11748, opened 2021-07-25
Add Safari versions for api.PaymentResponse.onpayerdetailchange

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `onpayerdetailchange` member of the `PaymentResponse` API.  The data for the event handler was obtained from the mdn-bcd-collector project, while the event itself was copied from the event handler.

### DIFF
--- a/api/PaymentResponse.json
+++ b/api/PaymentResponse.json
@@ -238,7 +238,7 @@
               "version_added": "12.1"
             },
             "safari_ios": {
-              "version_added": "11.3"
+              "version_added": "12.2"
             },
             "samsunginternet_android": {
               "version_added": true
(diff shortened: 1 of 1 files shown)


api/PaymentResponse.json
```diff
@@ -238,7 +238,7 @@
               "version_added": "12.1"
             },
             "safari_ios": {
-              "version_added": "11.3"
```
foolip, line 241: Can you check where this data came from? Maybe it's correct? :)

vinyldarkscratch: This came from https://github.com/mdn/browser-compat-data/pull/6679/files#diff-ebf770809c04c0bcb7fed9d6007fea3b3dd550547b7b6c2831e79faf8479a494L352-R237, but in the same PR, I set Safari Desktop to "12.1", so I am confident this was originally a copy-paste error.

foolip (approved):

PR #11749, opened 2021-07-25
Add Safari versions for RTCDataChannel API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `RTCDataChannel` API.  The data was copied from their event handler counterparts.

### DIFF
--- a/api/RTCDataChannel.json
+++ b/api/RTCDataChannel.json
@@ -346,10 +346,10 @@
               "version_added": "43"
             },
             "safari": {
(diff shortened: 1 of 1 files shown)


foolip:
> > The data was copied from their event handler counterparts.
> 
> Is this a copy-paste error, or what event goes together with `RTCDataChannel`?

Oh, I see, this is updating event entries :)

foolip (approved):

PR #11750, opened 2021-07-25
Add Safari versions for TextTrackCue API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `TextTrackCue` API.  The data was copied from their event handler counterparts.

### DIFF
--- a/api/TextTrackCue.json
+++ b/api/TextTrackCue.json
@@ -126,10 +126,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": "6"
             },
             "safari_ios": {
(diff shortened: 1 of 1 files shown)


foolip:
Waiting for https://github.com/mdn/browser-compat-data/pull/11751 so that the versions make sense together. 6 or 7?

foolip:
> Waiting for #11751 so that the versions make sense together. 6 or 7?

6/7 for Safari/iOS is consistent with https://github.com/mdn/browser-compat-data/pull/11751.

foolip (approved):

PR #11751, opened 2021-07-25
Add Safari versions for TextTrackList API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `TextTrackList` API.  The data was copied from their event handler counterparts.

### DIFF
--- a/api/TextTrackList.json
+++ b/api/TextTrackList.json
@@ -79,10 +79,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": "6"
             },
             "safari_ios": {
-              "version_added": null
+              "version_added": "7"
             },
             "samsunginternet_android": {
               "version_added": true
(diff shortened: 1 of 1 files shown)


api/TextTrackList.json
```diff
@@ -79,10 +79,10 @@
               "version_added": null
             },
             "safari": {
-              "version_added": null
+              "version_added": "6"
```
foolip, line 82: 6 here but 7 below. Is that correct?

vinyldarkscratch: Oddly enough, it is; `addTrack` was added before the other two events.

If you're talking about Safari iOS being 7, that's also correct; support wasn't implemented in iOS until iOS 7.0.

foolip: Thanks for confirming!

foolip (approved):

PR #11752, opened 2021-07-25
Add all browsers versions for WebSocket API

vinyldarkscratch:
This PR adds real values for all browsers for the `WebSocket` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.2).

### DIFF
--- a/api/WebSocket.json
+++ b/api/WebSocket.json
@@ -89,10 +89,10 @@
               "version_added": "10"
             },
             "opera": {
-              "version_added": "≤15"
+              "version_added": "12.1"
             },
             "opera_android": {
-              "version_added": "≤14"
+              "version_added": "12.1"
             },
             "safari": {
               "version_added": "5"
@@ -1009,10 +1009,10 @@
               "version_added": "10"
             },
             "opera": {
-              "version_added": "≤15"
+              "version_added": "12.1"
             },
             "opera_android": {
-              "version_added": "≤14"
+              "version_added": "12.1"
             },
             "safari": {
               "version_added": "5"
--- a/api/WebSocket.json
+++ b/api/WebSocket.json
@@ -27,8 +27,7 @@
           ],
           "firefox_android": [
             {
-              "version_added": "14",
-              "notes": "See <a href='https://bugzil.la/695635'>bug 695635</a>."
+              "version_added": "14"
             },
             {
               "version_added": "7",
--- a/api/WebSocket.json
+++ b/api/WebSocket.json
@@ -23,11 +23,6 @@
               "version_removed": "11",
               "prefix": "Moz",
               "notes": "Message size limited to 16 MB (see <a href='https://bugzil.la/711205'>bug 711205</a>)."
-            },
-            {
-              "version_added": "4",
-              "version_removed": "6",
-              "notes": "Message size limited to 16 MB (see <a href='https://bugzil.la/711205'>bug 711205</a>)."
             }
           ],
           "firefox_android": [
@@ -39,11 +34,6 @@
               "version_added": "7",
               "version_removed": "14",
               "prefix": "Moz",
-              "notes": "Message size limited to 16 MB (see <a href='https://bugzil.la/711205'>bug 711205</a>)."
-            },
-            {
-              "version_added": "4",
-              "version_removed": "6",
               "notes": "Message size limited to 16 MB (see <a href='https://bugzil.la/711205'>bug 711205</a>)."
             }
           ],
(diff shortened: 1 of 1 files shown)


api/WebSocket.json
```diff
@@ -23,11 +23,6 @@
               "version_removed": "11",
               "prefix": "Moz",
               "notes": "Message size limited to 16 MB (see <a href='https://bugzil.la/711205'>bug 711205</a>)."
-            },
-            {
-              "version_added": "4",
```
foolip, line 28: Why are you deleting this bit? Is it wrong? Maybe it was prefixed and needs fixing?

vinyldarkscratch: I tried using the WebSocket API in these earlier versions, but `WebSocket` was undefined in these earlier versions, either prefixed or not.  As such, I've removed this statement.

foolip: I see, maybe it was behind a flag but not correctly recorded as such.

api/WebSocket.json
```diff
@@ -82,47 +71,40 @@
...
+            },
             "firefox_android": {
               "version_added": "7"
             },
             "ie": {
               "version_added": "10"
             },
             "opera": {
-              "version_added": true
+              "version_added": "≤15"
```
foolip, line 92: The data for constructors should usually match the parent feature, especially for things where the constructor is the main entry point to the API.

The parent feature says 12.1 and https://dev.opera.com/blog/opera-12-10-is-out/ mentions "Web Sockets" so this should also be 12.1:

```suggestion
              "version_added": "12.1"
```

api/WebSocket.json
```diff
@@ -82,47 +71,40 @@
...
             },
             "opera_android": {
-              "version_added": null
+              "version_added": "≤14"
```
foolip, line 95:
```suggestion
              "version_added": "12.1"
```

api/WebSocket.json
```diff
@@ -1041,10 +1009,10 @@
               "version_added": "10"
             },
             "opera": {
-              "version_added": null
+              "version_added": "≤15"
```
foolip, line 1012: http://mdn-bcd-collector.appspot.com/tests/api/WebSocket?exposure=Worker confirms support in Opera 12.16.

```suggestion
              "version_added": "12.1"
```

api/WebSocket.json
```diff
@@ -1041,10 +1009,10 @@
...
             },
             "opera_android": {
-              "version_added": null
+              "version_added": "≤14"
```
foolip, line 1015:
```suggestion
              "version_added": "12.1"
```

api/WebSocket.json
```diff
@@ -23,28 +23,17 @@
...
-            {
-              "version_added": "4",
-              "version_removed": "6",
-              "notes": "Message size limited to 16 MB (see <a href='https://bugzil.la/711205'>bug 711205</a>)."
             }
           ],
           "firefox_android": [
             {
-              "version_added": "14",
-              "notes": "See <a href='https://bugzil.la/695635'>bug 695635</a>."
```
foolip, line 36: I made this edit. https://bugzil.la/695635 is the unprefixed bug and could have been linked in the prefixed entry, but is really just unnessary to link anywhere, and was left over in https://github.com/mdn/browser-compat-data/pull/6608.

api/WebSocket.json
```diff
@@ -138,40 +120,40 @@
...
             "edge": {
               "version_added": "12"
             },
             "firefox": {
-              "version_added": true
+              "version_added": "11"
             },
             "firefox_android": {
-              "version_added": true
+              "version_added": "14"
```
foolip, line 135:
```suggestion
              "version_added": "11"
```

Assuming there was no testing of older Firefox for Android, I think this should be 11 matching desktop. The other data is the same for desktop and Android, the only difference being in when the unprefixing happened. I kind of doubt that's correct, but that's a separate question.

api/WebSocket.json
```diff
@@ -82,47 +71,40 @@
...
-                "version_added": "7"
-              },
-              {
-                "version_added": "4",
-                "version_removed": "7",
-                "notes": "Parameter <code>protocols</code> not supported."
-              }
-            ],
+            "firefox": {
+              "version_added": "7"
```
foolip, line 83: Can you duplicate the "Moz" prefix data here? We need to do that for constructors IMO since that's where the prefix is used.

api/WebSocket.json
```diff
@@ -23,28 +23,17 @@
...
-              "version_added": "4",
-              "version_removed": "6",
-              "notes": "Message size limited to 16 MB (see <a href='https://bugzil.la/711205'>bug 711205</a>)."
             }
           ],
           "firefox_android": [
             {
-              "version_added": "14",
-              "notes": "See <a href='https://bugzil.la/695635'>bug 695635</a>."
+              "version_added": "14"
```
foolip, line 30: Can you check where this version 14 (as opposed to 11 for desktop) cam from? I might have been different for desktop and Android, but it seems just as likely to be a mistake. Since I'm asking you to copy the data to the constructor, it would be good to check it first.

foolip (changes requested):

api/WebSocket.json
vinyldarkscratch: That's quite possible, yeah -- if that's the case though, then it's flag data so old we don't need to keep it around!  :P

api/WebSocket.json
vinyldarkscratch: There actually is no Firefox Android 11 through 13 release, which is why this is set to 14.

foolip: Oh :)

foolip (approved):

foolip:
With this merged, we're now at 96.09% real+ranged values for APIs 🥳

PR #11753, opened 2021-07-25
Remove irrelevant api.Window.beforeunload_event.custom_text_support feature

vinyldarkscratch:
This PR removes the irrelevant `beforeunload_event.custom_text_support` member of the `Window` API as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-features). The lack of current support has been confirmed by the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.2), even if the current BCD suggests support.

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -686,63 +686,6 @@
             "deprecated": false
           }
         },
-        "custom_text_support": {
-          "__compat": {
-            "description": "Custom text support",
-            "support": {
-              "chrome": {
-                "version_added": true,
-                "version_removed": "51"
-              },
-              "chrome_android": {
-                "version_added": true,
-                "version_removed": "51"
-              },
-              "edge": {
-                "version_added": false
-              },
-              "firefox": {
-                "version_added": true,
-                "version_removed": "44"
-              },
-              "firefox_android": {
-                "version_added": true,
-                "version_removed": "44"
-              },
-              "ie": {
-                "version_added": true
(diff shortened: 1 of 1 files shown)


ddbeck:
@vinyldarkscratch can you link to the content PR for this?

vinyldarkscratch:
There is actually no mention of this subfeature anywhere within the MDN page, so there is no need for a content PR!

ddbeck:
I don't think that's right, or I've badly misunderstood the page and this feature. From [the reference page](https://developer.mozilla.org/en-US/docs/Web/API/Window/beforeunload_event):

> Some browsers used to display the returned string in the confirmation dialog, enabling the event handler to display a custom message to the user. However, this is deprecated and no longer supported in most browsers.

vinyldarkscratch:
Wow, that's super embarrassing -- I legitimately have glossed over that paragraph both times I checked the page (and I thought I did a search for "custom" before, but I guess not).  Eheh...whoops!  A content PR has now been created!

ddbeck (approved):
Thanks, @vinyldarkscratch!

And I hope it wasn't too embarrassing. Honestly, it's a good thing that "reading every word" (or even _writing_ every word) is not a job requirement for being a tech writer, or I'd have a very different career. 😅

PR #11755, opened 2021-07-25
Add Chrome/Safari versions for api.HTMLInputElement.search_event

vinyldarkscratch:
This PR adds real values for Chrome and Safari for the `search_event` member of the `HTMLInputElement` API.  The data was copied from its event handler counterparts.


foolip:
Can you test this? I wonder if there are multiple "search" events and if it's actually fired going back to Safari 1. I find it suspicious that Edge wouldn't have supported this if it's that old.

foolip (approved):

PR #11757, opened 2021-07-25
Add all browsers versions for XMLHttpRequest[EventTarget] API

vinyldarkscratch:
This PR adds real values for all browsers for the `XMLHttpRequest` and `XMLHttpRequestEventTarget` APIs.  The data was collected from the mdn-bcd-collector project (v3.2.8), and the event data was copied from their event handler counterparts (from `XMLHttpRequestEventTarget`).

### DIFF
--- a/api/XMLHttpRequest.json
+++ b/api/XMLHttpRequest.json
@@ -108,40 +108,40 @@
           "spec_url": "https://xhr.spec.whatwg.org/#the-abort()-method",
           "support": {
             "chrome": {
-              "version_added": "18"
+              "version_added": "1"
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": "18"
             },
             "edge": {
               "version_added": "12"
             },
             "firefox": {
-              "version_added": true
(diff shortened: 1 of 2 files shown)


foolip:
> The data was copied from their even handler counterparts (from `GlobalEventHandlers`).

This isn't a reliable approach because the event handler properties on `GlobalEventHandlers` are for different events with the same names. In particular "error" is many different events.

I'll review and comment where I'm not sure about the results.

foolip:
I've sent https://github.com/foolip/mdn-bcd-collector/pull/1376 to add more tests to help fill out `XMLHttpRequestEventTarget`. With that filled out I think we could copy the data from those event handler properties and have better confidence in the data.

vinyldarkscratch:
I've updated this PR to include results from `XMLHttpRequestEventTarget`, and the event data is now copied from there.  This one's ready to go!

foolip (approved):

foolip:
Thanks @vinyldarkscratch, this was a big one!

PR #11761, opened 2021-07-25
Update Safari versions for FileReaderSync API

vinyldarkscratch:
This PR updates and corrects the real values for Safari (Desktop and iOS/iPadOS) for the `FileReaderSync` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.2).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/FileReaderSync

### DIFF
--- a/api/FileReaderSync.json
+++ b/api/FileReaderSync.json
@@ -30,12 +30,10 @@
             "version_added": "≤12.1"
           },
           "safari": {
-            "version_added": "6",
-            "version_removed": "7"
+            "version_added": "6"
           },
           "safari_ios": {
-            "version_added": "6",
-            "version_removed": "7"
+            "version_added": "6"
           },
           "samsunginternet_android": {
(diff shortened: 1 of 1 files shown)


foolip (approved):

foolip:
Can you find where this data was added and if any evidence was given? It sure seems wrong, but maybe there's something interesting to note about this...

vinyldarkscratch:
The data originally came from https://github.com/mdn/browser-compat-data/pull/7377, which was a PR using collector results.  I genuinely don't know why the collector was reporting a lack of support before, as there have been no custom tests for this interface...  Maybe a bug we fixed in the updater script?

foolip:
I see. I trust current collector results over past collector results, so won't dig into this more.

foolip (approved):

PR #11762, opened 2021-07-25
Add/correct data for worker support for various APIs

vinyldarkscratch:
This PR sets real values for worker support across various PRs.  The results were collected using tests from the mdn-bcd-collector project.

### DIFF
--- a/api/CustomEvent.json
+++ b/api/CustomEvent.json
@@ -225,13 +225,13 @@
           "description": "Available in workers",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "43"
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": "43"
             },
             "edge": {
-              "version_added": "12"
+              "version_added": "17"
             },
             "firefox": {
               "version_added": "48"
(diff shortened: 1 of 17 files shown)


api/CustomEvent.json
```diff
@@ -225,13 +225,13 @@
           "description": "Available in workers",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "43"
```
foolip, line 228: Confirmed by the date of https://bugs.chromium.org/p/chromium/issues/detail?id=370011#c7 + https://www.chromium.org/developers/calendar

foolip:
@vinyldarkscratch can you describe the methodology here? Have you manually tested and looked at the "Window exposure" variants of tests, or have you used a script and collected results? Depending on what you've done, I might not need to review very carefully.

vinyldarkscratch:
I went through each of the APIs' tests in the collector and looked at the "Worker exposure" support entry.  After narrowing down the version number, I then put it in BCD!

foolip:
Thanks @vinyldarkscratch! That sounds pretty time-consuming, but it should give the correct results!

api/WindowOrWorkerGlobalScope.json
```diff
@@ -858,7 +858,7 @@
               "version_added": "8"
             },
             "samsunginternet_android": {
-              "version_added": "2.0"
+              "version_added": "1.5"
```
foolip, line 861: This seems like a random change, but indeed Chrome 24 would mirror to Samsung 1.5.

foolip (approved):

PR #11765, opened 2021-07-25
Add Safari iOS versions for api.Element.contextmenu_event

### LINKED ISSUE #6376: contextmenu event: Safari on iOS does not support it.
The contextmenu event isn't supported on iOS 13 and later and needs a lot of quirky hacks with a custom setTimeout() callback in order to behave the same as on other Browsers.

A lot of other issues in libraries claim that this is the result of intermixing touch and pointer events, though this assumption in general is wrong.

In this very simple test case, it won't work on iOS Safari 13+, and will instead open the magnifier overlay that magnifies the text (even if none is there) under it.

Even if you add the `-webkit-user-select: none` (aka `user-select: none` in WebKit nightly) CSS fix to the element, the `contextmenu` event won't be fired.

```html
<!DOCTYPE html>
<html>
	<head>
<style>
#test {
	display: block;
	width: 200px;
	height: 200px;
	background: red;
	/* This Fix ain't working either */
	/* -webkit-user-select: none; */
}
#test.contextmenu {
	background: green;
}
</style>
	</head>
	<body>
		<div id="test"></div>
	<script>
		let element = document.querySelector('#test');
		if (element !== null) {
			element.addEventListener('contextmenu', () => {

				element.className = 'contextmenu';

				setTimeout(() => {
					element.className = '';
				}, 2000);

			});
		}
	</script>
	</body>
</html>
```

vinyldarkscratch:
This PR adds real values for Safari iOS/iPadOS for the `contextmenu_event` member of the `Element` API.  This PR fixes #6376, which is where the data comes from.

Tracking Bug: https://webkit.org/b/213953

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -2245,7 +2245,7 @@
               "version_added": "3"
             },
             "safari_ios": {
-              "version_added": "1"
+              "version_added": false
             },
             "samsunginternet_android": {
               "version_added": "1.0"
--- a/api/Element.json
+++ b/api/Element.json
@@ -2245,7 +2245,8 @@
               "version_added": "3"
             },
             "safari_ios": {
-              "version_added": false
+              "version_added": false,
+              "notes": "See <a href='https://webkit.org/b/213953'>bug 213953</a>."
             },
             "samsunginternet_android": {
               "version_added": "1.0"


foolip (changes requested):
Can you link to the bug as well?

vinyldarkscratch:
Bug link added!

foolip (approved):

PR #11770, opened 2021-07-26
Firefox 91 gamepad permission policy

hamishwillee:
FF91 adds `gamepad` feature policy (permission policy) directive. This is spec non-compliant in that the default policy is * (all) rather than `self` (is a temporary bug fix):
- FF bug: https://bugzilla.mozilla.org/show_bug.cgi?id=1704005
- FF bug where permission is modified to EAll: https://bugzilla.mozilla.org/show_bug.cgi?id=1718221
- Docs work for this being tracked in: https://github.com/mdn/content/issues/6723

This was added to Chromium based things in v86 behind a preference: https://bugs.chromium.org/p/chromium/issues/detail?id=1011006. This is supposed to be removed soon. There is no evidence this is in Safari - see Moz bug above.

### DIFF
--- a/http/headers/feature-policy.json
+++ b/http/headers/feature-policy.json
@@ -656,14 +656,7 @@
                 ]
               },
               "opera_android": {
-                "version_added": "61",
-                "flags": [
-                  {
-                    "type": "preference",
-                    "name": "#restrict-gamepad-access",
-                    "value_to_set": "Enabled"
-                  }
-                ]
+                "version_added": false
               },
               "safari": {
                 "version_added": false
--- a/http/headers/feature-policy.json
+++ b/http/headers/feature-policy.json
@@ -602,7 +602,7 @@
         "gamepad": {
           "__compat": {
             "mdn_url": "https://developer.mozilla.org/docs/Web/HTTP/Headers/Feature-Policy/gamepad",
-            "spec_url": "https://w3c.github.io/permissions/#gamepad",
+            "spec_url": "https://www.w3.org/TR/gamepad/#dfn-gamepad",
             "support": {
               "chrome": {
                 "version_added": "86",
--- a/http/headers/feature-policy.json
+++ b/http/headers/feature-policy.json
@@ -636,11 +636,11 @@
               },
               "firefox": {
                 "version_added": "91",
-                "notes": "The default allowlist is <code>all</code> instead of <code>self</code> (as required by the specification)."
+                "notes": "The default allowlist is <code>*</code> instead of <code>self</code> (as required by the specification)."
               },
               "firefox_android": {
                 "version_added": "91",
-                "notes": "The default allowlist is <code>all</code> instead of <code>self</code> (as required by the specification)."
+                "notes": "The default allowlist is <code>*</code> instead of <code>self</code> (as required by the specification)."
               },
               "ie": {
                 "version_added": false
--- a/http/headers/feature-policy.json
+++ b/http/headers/feature-policy.json
@@ -599,6 +599,92 @@
             }
           }
         },
+        "gamepad": {
+          "__compat": {
+            "mdn_url": "https://developer.mozilla.org/docs/Web/HTTP/Headers/Feature-Policy/gamepad",
+            "spec_url": "https://w3c.github.io/permissions/#gamepad",
+            "support": {
+              "chrome": {
+                "version_added": "86",
+                "flags": [
+                  {
+                    "type": "preference",
+                    "name": "#restrict-gamepad-access",
+                    "value_to_set": "Enabled"
+                  }
+                ]
+              },
+              "chrome_android": {
+                "version_added": "86",
+                "flags": [
+                  {
+                    "type": "preference",
+                    "name": "#restrict-gamepad-access",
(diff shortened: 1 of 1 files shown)


api/Permissions.json
```diff
@@ -385,6 +385,54 @@
           }
         }
       },
+      "gamepad_permission": {
+        "__compat": {
+          "description": "<code>gamepad</code> permission",
+          "support": {
+            "chrome": {
+              "version_added": "86"
```
hamishwillee, line 393: Note below, that the corresponding directive for this permission is behind a flag. It is not clear to me if this is behind a flag as well, but it doesn't look like it on code inspection - i.e. you can check for this permission (not tested)

api/Permissions.json
```diff
@@ -385,6 +385,54 @@
...
+            "safari": {
+              "version_added": false
+            },
+            "safari_ios": {
+              "version_added": false
+            },
+            "samsunginternet_android": {
+              "version_added": "14.0"
+            },
+            "webview_android": {
```
hamishwillee, line 425: Note, permissions API not supported on webview android, so this must be false.

http/headers/feature-policy.json
```diff
@@ -599,6 +599,92 @@
...
+          "__compat": {
+            "mdn_url": "https://developer.mozilla.org/docs/Web/HTTP/Headers/Feature-Policy/gamepad",
+            "spec_url": "https://w3c.github.io/permissions/#gamepad",
+            "support": {
+              "chrome": {
+                "version_added": "86",
+                "flags": [
+                  {
+                    "type": "preference",
+                    "name": "#restrict-gamepad-access",
```
hamishwillee, line 612: Note, this is what is listed in various docs as the permission, but I can't find it on my normal chrome. Might be some kind of Canary-only thing?

http/headers/feature-policy.json
```diff
@@ -599,6 +599,92 @@
...
+                  }
+                ]
+              },
+              "safari": {
+                "version_added": false
+              },
+              "safari_ios": {
+                "version_added": false
+              },
+              "samsunginternet_android": {
```
hamishwillee, line 674: Note, set this and the next one as false as they have no permission API

api/Permissions.json
hamishwillee: Gamepad permission not present for this API - as noted in https://bugzilla.mozilla.org/show_bug.cgi?id=1704005#c30

http/headers/feature-policy.json
```diff
@@ -599,6 +599,92 @@
             }
           }
         },
+        "gamepad": {
+          "__compat": {
+            "mdn_url": "https://developer.mozilla.org/docs/Web/HTTP/Headers/Feature-Policy/gamepad",
+            "spec_url": "https://w3c.github.io/permissions/#gamepad",
```
hamishwillee, line 605: @marcoscaceres Where is the HTTP Feature-Policy/Permission Policy _directive_ for  gamepad defined? This `spec_url` https://w3c.github.io/permissions/#gamepad is the permission API doc, which does not define `gamepad`. 

The reason I used this is that the other Feature-Policy directive links do link to this spec - e.g. Feature-Policy accelerometer.

hamishwillee:
@ddbeck Could you please review or help find a reviewer for this?

http/headers/feature-policy.json
marcoscaceres: I'm going to define it in the Gamepad spec so it's linkable. Just need to fix a couple of things first.

marcoscaceres: FYI, this will hopeful make things more sane https://github.com/w3c/permissions/issues/263 ... or at least I hope.

hamishwillee: Thanks @marcoscaceres ! So to be clear - can I use this current URL for now?

marcoscaceres: Sorry for the delay! I was republishing the Gamepad spec to have the permissions link:

```suggestion
            "spec_url": "https://www.w3.org/TR/gamepad/#dfn-gamepad",
```

http/headers/feature-policy.json
ddbeck: I was able to confirm this in 86 (and not in 85) on BrowserStack. (Though for future reference, Canary-only features are `false`, at least until the `"preview"` lands.)

http/headers/feature-policy.json
```diff
@@ -599,6 +599,92 @@
...
+                "flags": [
+                  {
+                    "type": "preference",
+                    "name": "#restrict-gamepad-access",
+                    "value_to_set": "Enabled"
+                  }
+                ]
+              },
+              "opera_android": {
+                "version_added": "61",
```
ddbeck, line 659: Opera for Android doesn't support flags. This should be `false`.

ddbeck (changes requested):
A couple of suggestions for you. Thanks, @hamishwillee.

hamishwillee:
Thanks very much. Your changes/suggestions accepted.

ddbeck (approved):
Thank you!

PR #11771, opened 2021-07-26
Gamepad API - secure_context_required

hamishwillee:
FF91 requires `gamepad` has secure context: https://bugzilla.mozilla.org/show_bug.cgi?id=1704005. Further,  Chrome also puts gamepad in secure context from v86, though I read this as being behind a flag: https://bugs.chromium.org/p/chromium/issues/detail?id=1099544

This updates theMDN listed [GamePad APIs](https://developer.mozilla.org/en-US/docs/Web/API/Gamepad_API): 
`Gamepad`, `GamepadButton`, `GamepadEvent`, `GamepadHapticActuator`, `GamepadPose`. 

I did a check on `Navigator.getGamepads()` and that also requires a secure context. Since not everything in Navigator does, do I need to do add secure_context_required as a _subfeature_ of the method??
I can't tell if  `Window.ongamepadconnected` event and disconnected would require secure context - I get null on both secure and insecure contexts.

Related docs work can be tracked in https://github.com/mdn/content/issues/6723. This is associated with the change in https://github.com/mdn/browser-compat-data/pull/11770

### DIFF
--- a/api/Gamepad.json
+++ b/api/Gamepad.json
@@ -817,6 +817,89 @@
           }
         }
       },
+      "secure_context_required": {
+        "__compat": {
+          "description": "Secure context required",
+          "support": {
+            "chrome": {
+              "version_added": "86",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "#restrict-gamepad-access",
+                  "value_to_set": "Enabled"
+                }
+              ]
+            },
+            "chrome_android": {
+              "version_added": "86",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "#restrict-gamepad-access",
+                  "value_to_set": "Enabled"
+                }
+              ]
+            },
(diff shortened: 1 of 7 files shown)


vinyldarkscratch:
If `Navigator.getGamepads()` also requires secure contexts, then yes, we should add `secure_context_required` as a subfeature to the method!  (So we'll have a `api.Navigator.getGamepads.secure_context_required` feature.)

So far, this is LGTM!

hamishwillee:
Thanks @vinyldarkscratch . I've added api.Navigator.getGamepads.secure_context_required along with the same thing for the connected/disconnected event handlers. Mirroring the other cases and setting as false if the parent is false. 

I think this is now ready for re-review.

vinyldarkscratch (approved):
Thanks for the quick response, this looks all good to go now!

PR #11779, opened 2021-07-26
alignment values for flexbox in Chrome 93 and a question

rachelandrew:
The alignment values of `start`, `end`, `left`, and `right` as they apply in flex layout, are in Chrome 93.

I believe that our values for `align-self` were incorrect previously for Chromium (stating support for the `-self` property and false for the `-items` property).

https://www.chromestatus.com/feature/5777880099323904

Before this is merged ...

I'm a bit dubious about the support data for Firefox, data says Firefox 45 for start/end and left/right. However, based on https://bugzilla.mozilla.org/show_bug.cgi?id=1472843 some of these didn't land until 63. Doing a test `justify-content: right` works in 61 and up. So much later than 45.

So I'm proposing to add to this a change to the values as listed in the above bug to correct the Firefox data.

Also, to add the line for `self-start/self-end` which is also shipping in Chromium 93, and is referenced in the above Firefox bug, so I think 63 or thereabouts, I'll test it.

### DIFF
--- a/css/properties/align-content.json
+++ b/css/properties/align-content.json
@@ -209,16 +209,13 @@
               "description": "<code>left</code> and <code>right</code>",
               "support": {
                 "chrome": {
-                  "version_added": false,
-                  "notes": "This value is recognized, but has no effect."
+                  "version_added": "93"
                 },
                 "chrome_android": {
-                  "version_added": false,
-                  "notes": "This value is recognized, but has no effect."
+                  "version_added": "93"
                 },
                 "edge": {
-                  "version_added": false,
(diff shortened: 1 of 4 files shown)


vinyldarkscratch (approved):
This is looking good to me, thank you!

Regarding the Firefox data, I agree that we should change the versions.  If you want to add them into this PR, I'll hold off on merging for the time being, or if you want to create a separate PR for Firefox changes, let me know and I'll merge this PR!

vinyldarkscratch (approved):
I'll assume you'll take care of the Firefox data in a follow-up PR, so I'll merge this one now!

PR #11782, opened 2021-07-26
Update support for logical assignment operators in opera

lukewarlow:
#### Summary

Update support for logical assignment operators in Opera

Just copies blink support across to Opera and Opera Android


vinyldarkscratch (approved):
This makes sense to me, thanks!  👍

PR #11808, opened 2021-07-29
Fix support of CompositionEvent constructor by IE

KentoMoriwaki:
#### Summary

IE does not support CompositionEvent constructor.

It should be the same as other Event class (e.g. [KeyboardEvent](https://developer.mozilla.org/en-US/docs/Web/API/KeyboardEvent/KeyboardEvent#browser_compatibility))

### DIFF
--- a/api/CompositionEvent.json
+++ b/api/CompositionEvent.json
@@ -70,7 +70,7 @@
               "version_added": "53"
             },
             "ie": {
-              "version_added": true
+              "version_added": false
             },
             "opera": {
               "version_added": false


ddbeck (approved):
Hi @KentoMoriwaki and thanks for this PR. I see it's your first to the project—welcome!

I confirmed this by testing this in IE11 on BrowserStack against https://mdn-bcd-collector.appspot.com/tests/api/CompositionEvent/CompositionEvent. So I think this is good to go.

Thank you again!

PR #11838, opened 2021-08-01
Correct versions for WEBGL_compressed_texture_pvrtc API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `WEBGL_compressed_texture_pvrtc` API, based upon commit history and date.

Commit: https://source.chromium.org/chromium/chromium/src/+/4cb6a26d8b06a60da0e7460eedb5034d5d25ae30

Additionally, this PR corrects the data for Firefox in the process.  Since this is an API designed solely for specific mobile devices, desktop support wouldn't be a thing.

### DIFF
--- a/api/WEBGL_compressed_texture_pvrtc.json
+++ b/api/WEBGL_compressed_texture_pvrtc.json
@@ -14,7 +14,10 @@
           "edge": {
             "version_added": false
           },
-          "firefox": [
+          "firefox": {
+            "version_added": false
+          },
+          "firefox_android": [
             {
               "version_added": "18"
             },
@@ -24,9 +27,6 @@
               "version_removed": "58"
             }
           ],
-          "firefox_android": {
-            "version_added": null
-          },
           "ie": {
             "version_added": false
           },
--- a/api/WEBGL_compressed_texture_pvrtc.json
+++ b/api/WEBGL_compressed_texture_pvrtc.json
@@ -6,10 +6,10 @@
         "spec_url": "https://www.khronos.org/registry/webgl/extensions/WEBGL_compressed_texture_pvrtc/",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "28"
           },
           "chrome_android": {
-            "version_added": true
+            "version_added": "28"
           },
           "edge": {
             "version_added": "79"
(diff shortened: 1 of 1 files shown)


api/WEBGL_compressed_texture_pvrtc.json
```diff
@@ -6,10 +6,10 @@
         "spec_url": "https://www.khronos.org/registry/webgl/extensions/WEBGL_compressed_texture_pvrtc/",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "28"
```
foolip, line 9: The commit you linked says this might not actually be supported on desktop. It depends on the hardware I guess, and perhaps the OS.

Not sure how to represent this. Is there any pattern we can copy?

vinyldarkscratch: I wasn't certain about the desktop support either, especially since the existing data claims that BCD supports it on desktops and it was set to `true` originally, which came from #3704.  However, the documentation talks about support almost completely on mobile devices, so I'm thinking that setting desktop browsers to `false` is the right way to go about this?

foolip: Yeah, from https://en.wikipedia.org/wiki/PVRTC it's clear that this is only for devices with PowerVR GPUs, and that seems to be a mobile/tablet thing only. Let's set desktop to false.

foolip:
```suggestion
            "version_added": false
```

api/WEBGL_compressed_texture_pvrtc.json
```diff
@@ -31,10 +31,10 @@
             "version_added": false
           },
           "opera": {
-            "version_added": true
+            "version_added": "15"
```
foolip, line 34:
```suggestion
            "version_added": false
```

api/WEBGL_compressed_texture_pvrtc.json
```diff
@@ -6,10 +6,10 @@
...
           "chrome": {
-            "version_added": true
+            "version_added": "28"
           },
           "chrome_android": {
-            "version_added": true
+            "version_added": "28"
           },
           "edge": {
             "version_added": "79"
```
foolip, line 15:
```suggestion
            "version_added": false
```

foolip:
@vinyldarkscratch can you also update Firefox in this PR? We should mark it as not supported on desktop, but that requires updating the Firefox Android data to not just erase the existing data.

foolip (approved):

PR #11839, opened 2021-08-01
Add all browsers versions for WorkerLocation API

vinyldarkscratch:
This PR adds real values for all browsers for the `WorkerLocation` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.2).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/WorkerLocation

### DIFF
--- a/api/WorkerLocation.json
+++ b/api/WorkerLocation.json
@@ -39,7 +39,7 @@
             "version_added": "1.0"
           },
           "webview_android": {
-            "version_added": "37"
+            "version_added": "≤37"
           }
         },
         "status": {
@@ -87,7 +87,7 @@
               "version_added": "1.0"
             },
             "webview_android": {
-              "version_added": "37"
+              "version_added": "≤37"
             }
           },
           "status": {
@@ -136,7 +136,7 @@
               "version_added": "1.0"
             },
             "webview_android": {
-              "version_added": "37"
+              "version_added": "≤37"
             }
           },
           "status": {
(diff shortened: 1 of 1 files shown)


api/WorkerLocation.json
```diff
@@ -24,22 +24,22 @@
...
-            "version_added": true
+            "version_added": "5"
           },
           "samsunginternet_android": {
-            "version_added": true
+            "version_added": "1.0"
           },
           "webview_android": {
-            "version_added": true
+            "version_added": "37"
```
foolip, line 42: Shouldn't this be a range or do you have WebView testing for lots of versions set up now?

api/WorkerLocation.json
```diff
@@ -513,22 +513,22 @@
               "version_added": null
             },
             "opera": {
-              "version_added": true
+              "version_added": "≤15"
```
foolip, line 516: Why a range here?

vinyldarkscratch: Admittedly, the collector doesn't actually test `toString`.  The Safari and other Chromium data was mirrored from Chrome Desktop.  I hadn't tested this function in Opera Presto, and since Opera doesn't contribute to our current real values goal, I didn't dig into this further.

foolip: I see, that's fine. There aren't many "≤14" or "≤15" left and I've been keen to remove it as an allowed range, but fine to keep using it to not get bogged down with testing Presto in our current effort.

foolip (approved):

PR #11848, opened 2021-08-01
Update all browsers versions for MouseEvent API

vinyldarkscratch:
This PR updates and corrects the real values for all browsers for the `MouseEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.2).

### DIFF
--- a/api/MouseEvent.json
+++ b/api/MouseEvent.json
@@ -237,7 +237,7 @@
               "version_added": "11.3"
             },
             "samsunginternet_android": {
-              "version_added": "5.0"
+              "version_added": "4.0"
             },
             "webview_android": {
               "version_added": "43"
@@ -611,7 +611,7 @@
                 "version_added": "37"
               },
               {
-                "version_added": true,
+                "version_added": "25",
                 "version_removed": "37",
                 "prefix": "webkit"
               }
@@ -642,12 +642,26 @@
             "ie": {
               "version_added": false
             },
-            "opera": {
-              "version_added": "24"
-            },
-            "opera_android": {
-              "version_added": "24"
-            },
+            "opera": [
+              {
+                "version_added": "24"
+              },
+              {
+                "version_added": "15",
+                "version_removed": "24",
+                "prefix": "webkit"
+              }
+            ],
+            "opera_android": [
+              {
+                "version_added": "24"
+              },
+              {
+                "version_added": "14",
+                "version_removed": "24",
+                "prefix": "webkit"
+              }
+            ],
             "safari": [
               {
                 "version_added": "9"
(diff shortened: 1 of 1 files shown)


api/MouseEvent.json
```diff
@@ -59,10 +59,10 @@
           "description": "<code>MouseEvent()</code> constructor",
           "support": {
             "chrome": {
-              "version_added": "47"
+              "version_added": "26"
```
foolip, line 62: Confirmed with https://staging-dot-mdn-bcd-collector.appspot.com/tests/api/MouseEvent/MouseEvent.

api/MouseEvent.json
```diff
@@ -1024,26 +1024,36 @@
...
-                {
-                  "type": "preference",
-                  "name": "canvas.hitregions.enabled",
-                  "value_to_set": "true"
-                }
-              ]
-            },
+            "firefox": [
+              {
+                "version_added": "32"
```
foolip, line 1029: This is api.MouseEvent.region and appears to be a tricky case. The pref "canvas.hitregions.enabled" actually isn't in [`MouseEvent.webidl`](https://github.com/mozilla/gecko-dev/blob/af6b24b75f82e702476056d6822cb4a2bcf1f985/dom/webidl/MouseEvent.webidl) any longer, but it *is* in [`CanvasRenderingContext2D.webidl`](https://github.com/mozilla/gecko-dev/blob/af6b24b75f82e702476056d6822cb4a2bcf1f985/dom/webidl/CanvasRenderingContext2D.webidl#L333-L338) and seems to be off by default.

I strongly suspect that `region` is always the empty string without the pref enabled, and that we should mark this as `partial_support` or just pretend it's not there and keep it as false.

Can you break out these changes to a separate PR so it doesn't block the rest of this PR?

api/MouseEvent.json
```diff
@@ -1303,13 +1313,13 @@
               "version_added": "3.1"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "≤3"
```
foolip, line 1316: If we have confidence in 3.1 as the Safari version I think we could set iOS to 2, it can't be wrong by much:

```suggestion
              "version_added": "2"
```

foolip (changes requested):

api/MouseEvent.json
```diff
@@ -237,10 +237,10 @@
               "version_added": "11.3"
             },
             "samsunginternet_android": {
-              "version_added": false
+              "version_added": "5.0"
```
foolip, line 240: Is this based on the mirroring script? If so can you file an issue, because based on manual mirroring it should be:

```suggestion
              "version_added": "4.0"
```

foolip (changes requested):

PR #11850, opened 2021-08-01
Add all browsers versions for PageTransitionEvent API

vinyldarkscratch:
This PR adds real values for all browsers for the `PageTransitionEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.2).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/PageTransitionEvent

### DIFF
--- a/api/PageTransitionEvent.json
+++ b/api/PageTransitionEvent.json
@@ -6,40 +6,40 @@
         "spec_url": "https://html.spec.whatwg.org/multipage/browsing-the-web.html#the-pagetransitionevent-interface",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "4"
           },
           "chrome_android": {
-            "version_added": true
+            "version_added": "18"
           },
           "edge": {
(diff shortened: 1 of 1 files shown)


foolip:
We don't have `onpageshow` and `onpagehide` entries to compare the data for, so to have some confidence the interface was exposed from the get-go I looked up https://trac.webkit.org/changeset/47824/webkit, where indeed `PageTransitionEvent` is in `DOMWindow.idl`.

The date also lines up with the Chrome/Safari versions here.

foolip (approved):

foolip:
We don't have `onpageshow` and `onpagehide` entries to compare the data for, so to have some confidence the interface was exposed from the get-go I looked up https://trac.webkit.org/changeset/47824/webkit, where indeed `PageTransitionEvent` is in `DOMWindow.idl`.

The date also lines up with the Chrome/Safari versions here.

PR #11851, opened 2021-08-01
Add Safari versions for PopStateEvent API

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `PopStateEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.2).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/PopStateEvent

### DIFF
--- a/api/PopStateEvent.json
+++ b/api/PopStateEvent.json
@@ -30,10 +30,10 @@
             "version_added": true
           },
           "safari": {
-            "version_added": true
+            "version_added": "6"
           },
           "safari_ios": {
-            "version_added": true
+            "version_added": "6"
           },
           "samsunginternet_android": {
             "version_added": "1.0"
@@ -126,10 +126,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "10"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "10"
             },
             "samsunginternet_android": {
               "version_added": "1.0"


api/PopStateEvent.json
```diff
@@ -30,10 +30,10 @@
             "version_added": true
           },
           "safari": {
-            "version_added": true
+            "version_added": "6"
```
foolip, line 33: This doesn't match `onpopstate` and "popstate_event" data, which is 5/4.2 for Safari/iOS. Can you investigate if this is a case of the `PopStateEvent` interface not being exposed, or perhaps the "popstate" event originally used a plain `Event`?

The updated https://staging-dot-mdn-bcd-collector.appspot.com/tests/api/PopStateEvent should help answer this, but it looks like the harness is broken on Safari 5 now, so that's something to look into as well.

foolip: I think this is the same problem as in https://github.com/mdn/browser-compat-data/pull/11881#issuecomment-896595350, https://trac.webkit.org/changeset/94493/webkit also added `window.PopStateEvent`.

vinyldarkscratch: The issues with Safari 5 are actually caused by one of the extensions BrowserStack has loaded, of which appears to conflict with the JSON3 polyfill.  Running Safari 5 on my local VM works no problem.

Using the latest collector (v3.2.9), I tested this event once again in Safari 5.1 and 6, and have found that the `PopStateEvent` was implemented in Safari 6.  As a result, I'm certain that the `onpopstate` handler used a plain event instance before.

foolip: @vinyldarkscratch can you file an issue about the problem with the JSON3 polyfill? It sounds like we're loading it too early, and loading it when we're going to send the results would be better?

vinyldarkscratch: Regarding the event by the way, it was a regular event until Safari 6.

foolip: Great, thanks for confirming! There's already an issue about Safari 5 now so let's goooooo!

foolip (approved):

PR #11854, opened 2021-08-01
Update IE versions for External API

vinyldarkscratch:
This PR updates and corrects the real values for Internet Explorer for the `External` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.2).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/External

### DIFF
--- a/api/External.json
+++ b/api/External.json
@@ -20,7 +20,7 @@
             "version_added": true
           },
           "ie": {
-            "version_added": false
+            "version_added": "≤6"
           },
           "opera": {
             "version_added": "15"
@@ -68,7 +68,7 @@
               "version_added": true
             },
             "ie": {
-              "version_added": false
+              "version_added": "7"
             },
             "opera": {
               "version_added": "15"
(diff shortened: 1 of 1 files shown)


api/External.json
```diff
@@ -20,7 +20,7 @@
             "version_added": true
           },
           "ie": {
-            "version_added": false
+            "version_added": "≤6"
```
foolip, line 23: What properties did the object have in IE6, given that the methods below are set to 7?

vinyldarkscratch: I have no clue honestly, probably some proprietary IE-only implementation.  Though `window.external` _is_ defined in IE 6 (and as BCD states, all the way back to IE 4).

foolip: Fair enough. It's not inconsistent, just curious since some of the data is missing (and shouldn't be added).

foolip (approved):

PR #11856, opened 2021-08-01
Update all browsers versions for Plugin API

vinyldarkscratch:
This PR updates and corrects the real values for all browsers for the `Plugin` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.9).

### DIFF
--- a/api/Plugin.json
+++ b/api/Plugin.json
@@ -24,10 +24,10 @@
             "version_added": "4"
           },
           "opera": {
-            "version_added": "15"
+            "version_added": "≤12.1"
           },
           "opera_android": {
-            "version_added": "14"
+            "version_added": "≤12.1"
           },
           "safari": {
             "version_added": "1"
@@ -53,10 +53,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Plugin/description",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "1"
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": "18"
             },
             "edge": {
               "version_added": "12"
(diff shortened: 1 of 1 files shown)


api/Plugin.json
```diff
@@ -9,7 +9,7 @@
             "version_added": "1"
           },
           "chrome_android": {
-            "version_added": "18"
+            "version_added": false
```
foolip, line 12: `Plugin` if defined in Chrome for Android 91. I think there's a problem with http://mdn-bcd-collector.appspot.com/tests/api/Plugin, where an empty `navigator.plugins` array results in `instance` being undefined. But the interface object `Plugin` is exposed, and is in that sense supported, it's just that there aren't any instances because there are no plugins on mobile.

And indeed you've noticed this before:
https://github.com/foolip/mdn-bcd-collector/blob/90d055d06a50b9106f1d417aeec6c785efcfec2c/custom-tests.yaml#L1302

foolip (changes requested):

api/Plugin.json
foolip: I've sent https://github.com/foolip/mdn-bcd-collector/pull/1377 to make the test return null instead of false in this case, to avoid making incorrect updates.

vinyldarkscratch: Well, if mobile browsers don't have support for plugins, then perhaps we should set it to `false`, even if the APIs themselves are supported?

foolip: I could see the argument two ways:
- Set `navigator.plugins`, `PluginArray` and `Plugins` to false because there can be no actual plugins
- Set them to true because the APIs are there are correctly reflect that there can be no plugins

The previous state of the data was the second option, and that's what I lean towards. In this PR, can you keep things that way, and send another PR that changes all of the data for mobile if you think that would be better? Or perhaps add notes to all of these APIs?

vinyldarkscratch: I've reverted setting Chrome Android to `false` now, so this is ready to go!

foolip (approved):

PR #11861, opened 2021-08-01
Add all browsers versions for Document API

queengooborg:
This PR adds real values for all browsers for the `Document` API.  The data was copied from their matching event handlers (such as in `GlobalEventHandlers`), as well as manual testing.

Test Code Used:

animation events: https://github.com/mdn/browser-compat-data/pull/11861#discussion_r717507599

copy/cut/paste:
```html
<div id="test">
	<input value="The quick brown fox jumps over the lazy dog." />
</div>

<script>
	document.addEventListener('copy', function() {
		alert('Foo One!');
	});
	document.oncopy = function() {
		alert('Foo Two!');
	};
	document.addEventListener('cut', function() {
		alert('Bar One!');
	});
	document.oncut = function() {
		alert('Bar Two!');
	};
	document.addEventListener('paste', function() {
		alert('Baz One!');
	});
	document.onpaste = function() {
		alert('Baz Two!');
	};
</script>
```

readystatechange: http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=9673

selectionchange:
```html
<div id="test">
	<p>The quick brown fox jumps over the lazy dog.</p>
</div>

<script>
	document.addEventListener('selectionchange', function() {
		alert('Foo!');
	});
</script>
```

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -623,10 +623,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": true
+              "version_added": "51"
             },
             "firefox_android": {
-              "version_added": true
+              "version_added": "51"
             },
             "ie": {
               "version_added": "10"
@@ -1751,10 +1751,10 @@
               "version_added": "≤18"
             },
             "firefox": {
-              "version_added": true
+              "version_added": "9"
             },
             "firefox_android": {
-              "version_added": true
+              "version_added": "9"
             },
             "ie": {
               "version_added": false
@@ -2988,10 +2988,10 @@
               "version_added": "≤18"
             },
             "firefox": {
-              "version_added": true
+              "version_added": "9"
             },
             "firefox_android": {
-              "version_added": true
+              "version_added": "9"
             },
             "ie": {
               "version_added": false
@@ -8188,10 +8188,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": true
+              "version_added": "9"
             },
             "firefox_android": {
-              "version_added": true
+              "version_added": "9"
             },
             "ie": {
               "version_added": "11"
@@ -9756,40 +9756,40 @@
           "spec_url": "https://html.spec.whatwg.org/multipage/indices.html#event-readystatechange",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "9"
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": "18"
             },
             "edge": {
               "version_added": "12"
             },
             "firefox": {
-              "version_added": true
+              "version_added": "9"
             },
             "firefox_android": {
-              "version_added": true
+              "version_added": "9"
             },
             "ie": {
-              "version_added": true
+              "version_added": "4"
             },
             "opera": {
-              "version_added": true
+              "version_added": "≤12.1"
             },
             "opera_android": {
-              "version_added": true
+              "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": true
+              "version_added": "5.1"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "5"
             },
             "samsunginternet_android": {
-              "version_added": true
+              "version_added": "1.0"
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "≤37"
             }
           },
           "status": {
@@ -10466,10 +10466,10 @@
           "description": "<code>selectionchange</code> event",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "11"
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": "18"
             },
             "edge": {
               "version_added": "12"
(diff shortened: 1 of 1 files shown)


api/Document.json
```diff
@@ -623,10 +623,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": true
+              "version_added": "51"
```
foolip, line 626: This does match results for http://mdn-bcd-collector.appspot.com/tests/api/Document/onanimationend and https://mdn-bcd-collector.appspot.com/tests/api/HTMLElement/onanimationend but still seems way too late.

`AnimationEvent` is marked as shipped in Firefox 6 and the `animation` CSS property unprefixed in 16. Then why no "animationend" events until that much later? Can you do some manual testing of when the event is fired?

foolip:
I only commented on one thing, but I have low confidence in most things around events. Can you do some testing for each event where it's easy enough, and comment on ones that are too hard to test?

We can't make everything perfect here, but by copying dubious data (`GlobalEventHandlers` is a mixin and represents many things at the same time) I'm worried that we're making the data seem more consistent and thus likely accurate than it really is.

api/Document.json
queengooborg: I've given these a manual test in older Firefox editions.  It turns out that there was a big difference between when the events were supported vs. the event handlers.  I've updated the data using the following tests (which is a modified version of the test on https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/animationend_event for compatibility with older browser versions):

```html
<head>
		<!-- <meta content="width=device-width"> -->
		<style id="persistent-style">
			/* Persistent styling */
			body {
				background-color: black;
				color: white;
				margin: 0;
				padding: 0;
			}
			canvas {
				/*background-color: white;*/
			}
		</style>
		<style id="test-style">
			.container {
			  height: 3rem;
			}

			.event-log {
			  width: 25rem;
			  height: 2rem;
			  border: 1px solid black;
			  margin: .2rem;
			  padding: .2rem;
			}

			.animation.active {
			  -moz-animation-duration: 2s;
			  -moz-animation-name: slidein;
			  -moz-animation-iteration-count: 2;
			}

			@-moz-keyframes slidein {
			  from {
			    margin-left: 100%;
			    width: 300%;
			  }

			  to {
			    margin-left: 0%;
			    width: 100%;
			  }
			}
		</style>
	</head>
	<body>
		<div id="test">
			<div class="animation-example">
		    <div class="container">
		        <p class="animation">You chose a cold night to visit our planet.</p>
		    </div>
		    <button class="activate" type="button">Activate animation</button>
		    <div class="event-log"></div>
			</div>
		</div>

		<script>
			var animation = document.querySelector('p.animation');
			var animationEventLog = document.querySelector('.animation-example>.event-log');
			var applyAnimation = document.querySelector('.animation-example>button.activate');
			var iterationCount = 0;

			animation.addEventListener('animationstart', function() {
			  animationEventLog.textContent = 'animation started' ;
			});

			animation.addEventListener('animationiteration', function() {
			  iterationCount++;
			  animationEventLog.textContent = 'animation iterations: ' + iterationCount;
			});

			animation.addEventListener('animationend', function() {
			  animationEventLog.textContent = 'animation ended';
			  animation.classList.remove('active');
			  applyAnimation.textContent = "Activate animation";
			});

			animation.addEventListener('animationcancel', function() {
			  animationEventLog.textContent = 'animation canceled';
			});

			applyAnimation.addEventListener('click', function() {
			  animation.classList.toggle('active');
			  animationEventLog.textContent = '';
			  iterationCount = 0;
			  var active = animation.classList.contains('active');
			  if (active) {
			    applyAnimation.textContent = "Cancel animation";
			  } else {
			    applyAnimation.textContent = "Activate animation";
			  }
			});
		</script>
	</body>
```

foolip:
@queengooborg I'm not sure how to review the recent changes here. How were the versions for the animation and copy/cut/paste events determined?

queengooborg:
I put the test code I used in the PR's description!

api/Document.json
```diff
@@ -9756,40 +9756,40 @@
           "spec_url": "https://html.spec.whatwg.org/multipage/indices.html#event-readystatechange",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "9"
```
foolip, line 9759: I would bet real money that this is also a case of the event being supported earlier than the `onreadystatechange` property which this is mirroring data from. It doesn't matter much so if we had a range we could just drop that in, but I expect this is wrong.

Fortunately this is very easy to test, can you try http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=9673 in older browsers?

api/Document.json
```diff
@@ -10466,10 +10466,10 @@
           "description": "<code>selectionchange</code> event",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "11"
```
foolip, line 10469: Can you test if the event is fired earlier than this?

foolip: It's refreshing to find data that isn't wrong :)

api/Document.json
```diff
@@ -9756,40 +9756,40 @@
...
             "chrome_android": {
-              "version_added": true
+              "version_added": "18"
             },
             "edge": {
               "version_added": "12"
             },
             "firefox": {
-              "version_added": true
+              "version_added": "9"
```
foolip, line 9768: I tried to confirm 9 with http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=9673 but it looks like it was first in Firefox 4:

```suggestion
              "version_added": "4"
```

api/Document.json
```diff
@@ -9756,40 +9756,40 @@
...
             },
             "firefox_android": {
-              "version_added": true
+              "version_added": "9"
```
foolip, line 9771:
```suggestion
              "version_added": "4"
```

foolip (approved):

PR #11863, opened 2021-08-01
Add all browsers versions for FileReader API

queengooborg:
This PR adds real values for all browsers for the `FileReader` API.  The data for the events was copied from their event handler counterparts.  Data for the other features was mirrored from Chrome to Chrome Android and Samsung Internet.  Finally, this corrects the Safari data.

### DIFF
--- a/api/FileReader.json
+++ b/api/FileReader.json
@@ -108,7 +108,7 @@
               "version_added": "7"
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": "18"
             },
             "edge": {
               "version_added": "12"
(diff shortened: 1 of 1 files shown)


ddbeck:
> Finally, any mentions of Safari 6.1 was corrected to Safari 7, since we are removing the 6.1 backport data.

We don't seem to have consensus on that yet (see https://github.com/mdn/browser-compat-data/issues/9423#issuecomment-889162769), only for Safari 6.2. Maybe roll that part back? Or block this PR on a resolution of the data guideline?

vinyldarkscratch:
I've just checked Safari 6 in the collector, and it turns out pretty much every feature was added in Safari 6, not 7 (6.1), so we had bad data to begin with.  I've made the appropriate corrections now!

ddbeck (approved):
Thank you! 🎉

PR #11866, opened 2021-08-01
Add Safari versions for api.NodeList.length

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `length` member of the `NodeList` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.2).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/NodeList/length

### DIFF
--- a/api/NodeList.json
+++ b/api/NodeList.json
@@ -271,10 +271,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "≤4"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "≤3"
             },
             "samsunginternet_android": {
               "version_added": true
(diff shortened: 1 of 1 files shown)


lucalves (approved):

api/NodeList.json
```diff
@@ -271,10 +271,10 @@
               "version_added": true
             },
             "safari": {
-              "version_added": true
+              "version_added": "≤4"
```
foolip, line 274: Can you set this to 1 to match the parent feature? This interface makes no sense without the `length` property.

foolip (changes requested):

foolip (approved):

PR #11882, opened 2021-08-02
Correct Chromium versions for HashChangeEvent API

queengooborg:
This PR corrects real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `HashChangeEvent` API, based upon manual testing.

Test Code Used: 
```js
window.addEventListener('hashchange', function(event) {
  console.log(event);
});

if (location.hash === '#foobar') {
	location.hash = '#foo';
} else {
	location.hash = '#foobar';
}

```


foolip:
Doesn't this duplicate part of https://github.com/mdn/browser-compat-data/pull/11883?

vinyldarkscratch:
It doesn't; this PR is exclusively correcting the existing data, while the other PR is filling in new data.  It appears my PR-creating script is slightly bugged and doesn't properly change the title and description for corrections, I'll have to fix that...

foolip (changes requested):
This will leave BCD inconsistent, as the "hashchange_event" and "onhashchange" still say Chrome 5.

foolip:
> Test Code Used: https://developer.mozilla.org/en-US/docs/Web/API/HashChangeEvent#foobar

Do you mean https://developer.mozilla.org/en-US/docs/Web/API/HashChangeEvent#examples? Can you paste in the exact code tested, for the benefit of future double checking?

vinyldarkscratch:
I've now updated the description to include the proper test code!

foolip:
Looks like support for the event itself and the `HashChangeEvent` interface came about a year apart in WebKit:
https://trac.webkit.org/changeset/46892/webkit
https://trac.webkit.org/changeset/67898/webkit

To avoid having to research this again if we restructure event support data in some way, it would be good to add an initial `partial_support` range for the "hashchange_event" entry with a note that it was fired as an `Event` and not `HashChangeEvent`. @vinyldarkscratch if you're able to test for these two things separately, can you add the correct data?

vinyldarkscratch:
I won't be able to test this until my laptop is back from repairs, unfortunately...  I'll mark this as a draft for now and come back to it once I am able to!

queengooborg:
It appears that in Chrome 7, a regular `Event` was fired, but then in Chrome 8 it was a `HashChangeEvent`.  Data updates incoming!

api/Window.json
```diff
@@ -2678,7 +2678,8 @@
           "spec_url": "https://html.spec.whatwg.org/multipage/indices.html#event-hashchange",
           "support": {
             "chrome": {
-              "version_added": "5"
+              "version_added": "7",
+              "notes": "Before Chrome 8, a regular <code>Event</code> instance was thrown, rather than a <code>HashChangeEvent</code> instance."
```
foolip, line 2682:
```suggestion
              "notes": "In Chrome 8, a regular <code>Event</code> object is fired, rather than a <code>HashChangeEvent</code> object."
```

But how about using two ranges for this, to hide away the irrelevant note for readers on MDN?

foolip (changes requested):

api/Window.json
```diff
@@ -2677,10 +2677,16 @@
...
-              "notes": "Before Chrome 8, a regular <code>Event</code> instance was thrown, rather than a <code>HashChangeEvent</code> instance."
-            },
+            "chrome": [
+              {
+                "version_added": "8"
+              },
+              {
+                "version_added": "7",
+                "version_removed": "8",
+                "notes": "In Chrome 7, a regular <code>Event</code> object is fired, rather than a <code>HashChangeEvent</code> object."
```
foolip, line 2687:
```suggestion
                "notes": "A regular <code>Event</code> object is fired, rather than a <code>HashChangeEvent</code> object."
```

foolip (approved):

foolip (approved):

foolip:
Just as I was about to merge this when I noticed a potential problem in the test code:
```js
if (location.hash === '#foobar') {
	location.hash = '#foo';
} else {
	location.hash = '#foobar';
}

window.addEventListener('hashchange', function(event) {
  console.log(event);
});
```

The event handler here is added after `location.hash` is updated. If the event was initially fired *synchronously* when `location.hash` changes, then this test code wouldn't have caught that. @queengooborg can you test Chrome 5-7 again with this code instead?

```js
window.addEventListener('hashchange', function(event) {
  console.log(event);
});

if (location.hash === '#foobar') {
	location.hash = '#foo';
} else {
	location.hash = '#foobar';
}
```

And also trying to cause the hash to change by clicking links in the page, that will be a different code path than modifying `location.hash` from script.

queengooborg:
Good catch!  I ran the test code in Chrome 1-5 and found that the event was supported in earlier Chrome versions, all the way back to Chrome 4.  I've updated the data accordingly!

foolip:
Thanks @queengooborg! Sorting out such details isn't worthwhile in general, but if changing some existing data I want to be sure it's not regressing correct claims.

foolip (approved):

PR #11883, opened 2021-08-02
Add Chrome/Safari versions for HashChangeEvent API

vinyldarkscratch:
This PR adds real values for Chrome and Safari for the `HashChangeEvent` API, based upon manual testing.

Test Code Used:
```js
if (location.hash === '#foobar') {
	location.hash = '#foo';
} else {
	location.hash = '#foobar';
}

window.addEventListener('hashchange', function(event) {
	alert(event.oldURL);
	alert(event.newURL);
});
```

### DIFF
--- a/api/HashChangeEvent.json
+++ b/api/HashChangeEvent.json
@@ -103,10 +103,10 @@
           "spec_url": "https://html.spec.whatwg.org/multipage/browsing-the-web.html#dom-hashchangeevent-newurl-dev",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "8"
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": "18"
             },
             "edge": {
               "version_added": "12"
@@ -121,22 +121,22 @@
               "version_added": false
             },
             "opera": {
-              "version_added": true
+              "version_added": "≤15"
             },
             "opera_android": {
-              "version_added": true
+              "version_added": "≤14"
             },
             "safari": {
-              "version_added": true
+              "version_added": "5.1"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "6"
             },
             "samsunginternet_android": {
-              "version_added": true
+              "version_added": "1.0"
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "≤37"
             }
           },
           "status": {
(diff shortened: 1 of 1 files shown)


api/HashChangeEvent.json
```diff
@@ -121,22 +121,22 @@
               "version_added": false
             },
             "opera": {
-              "version_added": true
+              "version_added": "≤15"
```
foolip, line 124: Can you test Presto while you have the test handy?

foolip (changes requested):

api/HashChangeEvent.json
```diff
@@ -121,22 +121,22 @@
               "version_added": false
             },
             "opera": {
-              "version_added": true
+              "version_added": "15"
```
foolip, line 124: Support for the event interface but none of the properties seemed suspect so I tested Opera 12.16 a bit. `document.createEvent('HashChangeEvent').newURL` is the empty string meaning this was supported in Presto.

foolip (changes requested):

api/HashChangeEvent.json
foolip: Oh, and actually https://staging-dot-mdn-bcd-collector.appspot.com/tests/api/HashChangeEvent can test this now with your recent custom test additions.

api/HashChangeEvent.json
```diff
@@ -121,22 +121,22 @@
...
-              "version_added": true
+              "version_added": "≤12.1"
             },
             "opera_android": {
-              "version_added": true
+              "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": true
+              "version_added": "5.1"
```
foolip, line 130: This also doesn't quite match `onhashchange` data. Can you make bring them into alignment?

api/HashChangeEvent.json
```diff
@@ -103,10 +103,10 @@
           "spec_url": "https://html.spec.whatwg.org/multipage/browsing-the-web.html#dom-hashchangeevent-newurl-dev",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "8"
```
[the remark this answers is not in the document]
vinyldarkscratch: It was just a plain `Event` beforehand; same with Safari.

foolip (approved):

PR #11896, opened 2021-08-03
Improve ranged version descriptions

ddbeck:
#### Summary

This revises the wording describing each of the ranged versions permitted in the schema. The original wording was somewhat misleading, using "some version" when it meant a specific version, with possibly more besides.

This is not meant to change the schema, just how we talk about it.

#### Related issues

Based on feedback from @foolip in https://github.com/mdn/browser-compat-data/pull/11223/files#r658244508.


Elchi3 (approved):
Thanks, this is much better! 👍

foolip (commented):
Look good, thanks!

PR #11909, opened 2021-08-04
Add data guideline for Safari backported releases

vinyldarkscratch:
This PR is an attempt to add a guideline for the Safari backport releases (4.1, 6.1, 6.2, and 7.1) based upon the discussions in #9423.  This removes our initial guideline where we state the version number should be for the backported release, adding a section just below it to document these special Safari releases and what to do to deal with them.

Note: I'm writing this section under the assumption that we will remove ALL backported releases from Safari, though we haven't actually come to a conclusion quite yet.  I'm happy to change this PR as needed!

### DIFF
--- a/docs/data-guidelines.md
+++ b/docs/data-guidelines.md
@@ -166,15 +166,28 @@ For example, although the UI Events specification defines a [`DOM_KEY_LOCATION_S
 
 This guideline was proposed in [#7936](https://github.com/mdn/browser-compat-data/issues/7585), based in part on previous discussion in [#7585](https://github.com/mdn/browser-compat-data/issues/7585).
 
-## Release lines and backported features
+## Release lines
 
 Use version numbers to reflect which _release line_ (major or minor but not patch-level releases) first supported a feature, rather than absolute version numbers.
 
-Typically, BCD does not record absolute version numbers, such as Chrome 76.0.3809.46; instead BCD records significant releases (such as Chrome 76). Use the earliest applicable release line for recording support for a given feature, even when that support change was backported to a previous release of the browser.
-
-For example, if the current release of browser X is version 10.2, but a new feature was backported to previous versions including a new 9.7.1 release, then the supported version is 9.7 (not 10.2 or 9.7.1).
+Typically, BCD does not record absolute version numbers, such as Chrome 76.0.3809.46; instead BCD records significant releases (such as Chrome 76). Use the earliest applicable release line for recording support for a given feature. For example, if a feature was not added in Chrome 76.0.3700.43, but added in Chrome 76.0.3809.46, then the supported version is 76.
 
 This decision was made in [#3953, under the expectation that most users are likely to run the latest minor version of their browser](https://github.com/mdn/browser-compat-data/pull/3953#issuecomment-485847399), but not necessarily the latest version overall.
+
+## Safari backported releases
+
+Safari has released a number of versions as backports of newer releases. These releases have roughly the same features, with exceptions to OS-dependent interfaces such as notifications and media playback.
+
+These releases include the following:
+
+- Safari 4.1 <- 5.0
+- Safari 6.1 <- 7.0
+- Safari 6.2 <- 8.0
+- Safari 7.1 <- 8.0
+
+If a feature is found to be supported in one of these backport releases, set the verion number to the following major version. For example, if a new feature was added in Safari 7.0 and it is present in Safari 6.1, then the supported version is 7.0 (not 6 or 6.1).
+
+This decision was made in [#4679](https://github.com/mdn/browser-compat-data/issues/4679) and [#9423](https://github.com/mdn/browser-compat-data/issues/9423).
 
 ## Safari for iOS versioning
 


foolip (approved):

docs/data-guidelines.md
```diff
@@ -166,16 +166,29 @@ For example, although the UI Events specification defines a [`DOM_KEY_LOCATION_S
 
 This guideline was proposed in [#7936](https://github.com/mdn/browser-compat-data/issues/7585), based in part on previous discussion in [#7585](https://github.com/mdn/browser-compat-data/issues/7585).
 
-## Release lines and backported features
+## Release lines
```
ddbeck, line 169: I don't think we should change the existing guideline. It applies to all browsers, not just Safari. I think this has come up for other browsers (perhaps when `SharedArrayBuffer` was removed?), though it's definitely less prominent.

docs/data-guidelines.md
```diff
@@ -166,16 +166,29 @@ For example, although the UI Events specification defines a [`DOM_KEY_LOCATION_S
...
+## Safari backported releases
+
+Safari has released a number of versions as backports of newer releases. These releases have roughly the same features, with exceptions to OS-dependent interfaces such as notifications and media playback.
+
+These releases include the following:
+
+- Safari 4.1 <- 5.0
+- Safari 6.1 <- 7.0
+- Safari 6.2 <- 8.0
+- Safari 7.1 <- 8.0
```
ddbeck, line 186: I think we should use words here instead of an arrow. I got really confused by this list paired with the example after. I interpreted this list to mean something like, "if something is found to be supported in Safari 8, set the value to Safari 7.1", which I think is the opposite of what's intended?

docs/data-guidelines.md
```diff
@@ -166,16 +166,29 @@ For example, although the UI Events specification defines a [`DOM_KEY_LOCATION_S
...
 
-Typically, BCD does not record absolute version numbers, such as Chrome 76.0.3809.46; instead BCD records significant releases (such as Chrome 76). Use the earliest applicable release line for recording support for a given feature, even when that support change was backported to a previous release of the browser.
-
-For example, if the current release of browser X is version 10.2, but a new feature was backported to previous versions including a new 9.7.1 release, then the supported version is 9.7 (not 10.2 or 9.7.1).
+Typically, BCD does not record absolute version numbers, such as Chrome 76.0.3809.46; instead BCD records significant releases (such as Chrome 76). Use the earliest applicable release line for recording support for a given feature. For example, if a feature was not added in Chrome 76.0.3700.43, but added in Chrome 76.0.3809.46, then the supported version is 76.
 
 This decision was made in [#3953, under the expectation that most users are likely to run the latest minor version of their browser](https://github.com/mdn/browser-compat-data/pull/3953#issuecomment-485847399), but not necessarily the latest version overall.
 
+## Safari backported releases
```
ddbeck, line 177: For this section overall, I'd like to see the structure inverted, to lead with the high-level imperative (i.e., if a release is a known backport, then use the next major version) then fill in with details about backports, versions, etc.

ddbeck (changes requested):
A couple ideas for you, though subject to change depending on the outcome of the issue discussion.

ddbeck:
@vinyldarkscratch hey Vinyl, on the most recent BCD call we talked about this. It sounds like this is ready to re-visit (with the overall decision reflected by the text, my quibbles notwithstanding).

vinyldarkscratch:
> @vinyldarkscratch do you have access to both 1.3 and 2 in order to test for differences between the two?

I do have a Safari 1.3 VM, but unfortunately I don't have access to it at the moment while my laptop is undergoing repairs.  When I get my laptop back, though, I'll be sure to check them out!  (Sadly checking differences will be harder since the collector isn't supported below Safari 3.)

foolip:
I see, hope your laptops comes back strong and healthy from the repairs :)

This isn't really blocking this issue, we can write a guideline and then see after the fact whether Safari 1.3 should be removed or not based on that guideline.

vinyldarkscratch:
Marking this as a draft for the time being!  I'll be revising these docs based upon your review, @ddbeck, and will mark this as ready once I do!

docs/data-guidelines.md
```diff
@@ -166,16 +166,28 @@ For example, although the UI Events specification defines a [`DOM_KEY_LOCATION_S
...
+Typically, BCD does not record absolute version numbers, such as Chrome 76.0.3809.46; instead BCD records significant releases (such as Chrome 76). Use the earliest applicable release line for recording support for a given feature. For example, if a feature was not added in Chrome 76.0.3700.43, but added in Chrome 76.0.3809.46, then the supported version is 76.
 
 This decision was made in [#3953, under the expectation that most users are likely to run the latest minor version of their browser](https://github.com/mdn/browser-compat-data/pull/3953#issuecomment-485847399), but not necessarily the latest version overall.
 
+## Backported releases
+
+Some browsers have created backport releases to include all or most of the features in a new major version, targeted for older operating systems. If a feature is found to be supported in one of these backport releases, set the verion number to the following major version. For example, if a new feature was added in Safari 7.0 and it is present in Safari 6.1, then the supported version is 7.0 (not 6 or 6.1).
```
ddbeck, line 179: I've suggested a pretty substantial rewrite of this paragraph. What I'm I'm trying to achieve here:

* Downplaying the reason for backports (e.g., Node.js uses backporting a lot, but for different reasons)
* Suggesting what to do for backports that we don't have a rule for (e.g., we overlook the complexity of Node.js backports because they don't seem to backport new features far enough for it to matter)

```suggestion
Some browsers have backport releases, where a feature is added or removed in two or more versions at once. If not otherwise covered by this guideline, use the earliest applicable version (as described in the [Release lines](#release-lines) guideline). In some cases, however, you must set the the version number to the following major version. For example, if a new feature was added in Safari 7.0 and in Safari 6.1, then the supported version is 7.0 (not 6 or 6.1).
```

docs/data-guidelines.md
```diff
@@ -166,16 +166,28 @@ For example, although the UI Events specification defines a [`DOM_KEY_LOCATION_S
...
+
+Known backport releases include the following:
+
+- Safari
+  - 4.1 is a backport of 5.0
+  - 6.1 is a backport of 7.0
+  - 6.2 is a backport of 8.0
+  - 7.1 is a backport of 8.0
```
ddbeck, line 187: I'm still confused by this. I think it's because it says what something _is_ and not what to _do_. What do you think of a table like this, maybe?

| If the browser is… | and the version is… | then set the value to… |
|---------|--------------------|------------------------|
| Safari  | 4.1                | 5.0                    |
| Safari  | 6.1                | 7.0                    |
| Safari  | 6.2                | 8.0                    |
| Safari  | 7.1                | 8.0                    |

ddbeck (changes requested):
I think we're getting close! Some suggestions in line comments.

junel17 (commented):
docs/data-guidelines.md

PR #11924, opened 2021-08-05
Adds avoid-page and avoid-column support for break-inside for Firefox

SebastianZ:
#### Summary
Firefox 92 added support for the keywords `avoid-page` and `avoid-column` of the `break-inside` property. See [bug 1722945](https://bugzilla.mozilla.org/show_bug.cgi?id=1722945).

While I was on that I realized that the descriptions were wrong. They included the keywords `column` and `page` which are only available for `break-before` and `break-after` (see https://www.w3.org/TR/css-break-3/#break-between and https://www.w3.org/TR/css-break-3/#break-within). So I removed them.

https://jsfiddle.net/SebastianZ/qukjxtyh/

Sebastian

### DIFF
--- a/css/properties/break-inside.json
+++ b/css/properties/break-inside.json
@@ -67,9 +67,9 @@
               "deprecated": false
             }
           },
-          "column": {
+          "avoid-column": {
             "__compat": {
-              "description": "<code>column</code> and <code>avoid-column</code>",
+              "description": "<code>avoid-column</code>",
               "support": {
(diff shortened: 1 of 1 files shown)


hamishwillee:
As an aside, see https://github.com/mdn/content/issues/7717. There appears to be no top level spec.

ddbeck (approved):
Looks good. Thank you! 🎉

PR #11935, opened 2021-08-07
docs(HTMLElement.inert): add missing spec

yinonov:
#### Summary

Add missing spec link for HTMLElement.inert

### DIFF
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -1185,6 +1185,7 @@
       "inert": {
         "__compat": {
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLElement/inert",
+          "spec_url": "https://html.spec.whatwg.org/multipage/interaction.html#inert-subtrees",
           "support": {
             "chrome": {
               "version_added": "60",


ddbeck:
Thanks for this PR, @yinonov. Also I see it's your first to the project—welcome!

It looks like this attribute isn't part of the HTML spec yet; see https://github.com/whatwg/html/pull/4288. I'm not sure whether we ought to link to inertness spec section, especially since it explicitly says:

> This section does not define or create any content attribute named "inert".

@Elchi3 @sideshowbarker do you have a take on what (if anything) ought to be used as the spec URL for this? I'm still uncertain on what constitutes a good and correct spec URL (apart from "passes the linter"). I could use some guidelines here (and in general).

sideshowbarker:
> @Elchi3 @sideshowbarker do you have a take on what (if anything) ought to be used as the spec URL for this?

For cases where something isn’t actually in a published spec, we have not been adding an spec URL.

As noted above, the proposed specification text for `inert` only exists in a patch in a PR branch of the HTML spec repo. We’ve intentionally chosen not to link to that.

sideshowbarker:
…and to be clear: the section the `https://html.spec.whatwg.org/multipage/interaction.html#inert-subtrees` URL in the patch in this PR links to is not a definition of the `inert` attribute. So we definitely don’t want to use that URL

ddbeck:
@sideshowbarker 

> We’ve intentionally chosen not to link to that.

OK, great. In a case like this, what's the right move for the specifications section on the MDN page? Right now it's got the missing spec box.

sideshowbarker:
> OK, great. In a case like this, what's the right move for the specifications section on the MDN page? Right now it's got the missing spec box.

I think what @teoli2003 has been doing in those cases is just adding a sentence like *“This feature is no part of any current specification.”°, and then not including the `{{Specification}}` macro.

yinonov:
hey sorry for missing out the small (big 😁) letters in my reference. but t thanks for the detailed correspondence here 👍

ddbeck:
@yinonov Thank you for taking the time to contribute here. Since there's no change to make to this file (yet), I'll close this PR. If you'd like to apply sideshowbarker's change to the content, then you can open a PR against https://github.com/mdn/content/blob/main/files/en-us/web/api/htmlelement/inert/index.html#L46-L48. Thank you again!

PR #11941, opened 2021-08-09
RTCIceCandidate - init dictionary into options

hamishwillee:
The [RTCIceCandidate constructor](https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidate/RTCIceCandidate) takes an (optional) options dictionary that is of type [RTCIceCandidateInit](https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidateInit) in the spec. My understanding is that we're removing dictionaries and enums across MDN and BCD, so I have moved this with structure as shown.

The spec refers to this parameter as` candidateInitDict`. I have just called this `candidateInfo_parameter` because that makes it easier to map to docs where you don't want to be referring to `candidate.candidate` property.
```
RTCIceCandidate
    RTCIceCandidate constructor
       candidateInfo_parameter
            candidateInfo_candidate_parameter
            candidateInfo_spdMid_parameter
            etc.
    candidate
    spdMid
    etc.
 ```           

Note, the dictionary looks like it appears in versions BEFORE the constructor appears as shown below. But I this is because the constructor early versions had a different prefix. 
    image

Further, still got to do the following - but waiting on confirmation above is sane.
 1. The [RTCPeerConnection.addIceCandidate()](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/addIceCandidate) also takes this as an options - just follow same process for that?
 1. The `RTCIceCandidateInit` is also returned by [RTCIceCandidate. toJSON()](https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidate/toJSON) - same process for that? 
 1. When all that done, I just delete the `RTCIceCandidateInit` in BCD?

### CODE REVIEW

File: api/RTCIceCandidate.json [hunk 1]
```diff
@@ -124,6 +124,253 @@
...
+                },
+                "chrome_android": {
+                  "version_added": true
+                },
+                "edge": {
+                  "version_added": "≤18"
+                },
+                "firefox": {
+                  "version_added": "22",
+                  "notes": "Prior to Firefox 68, the <code>candidate</code> property was required; Firefox 68 and later make it optional per a mid-2017 specification update."
```
ddbeck, line 192: This note's a bit complicated for what it's saying. How about this (with the caveat that the name of the parameter might be different)?

```suggestion
                  "notes": "Before Firefox 68, <code>candidateInfo.candidate</code> is required."
```

api/RTCIceCandidate.json
```diff
@@ -124,6 +124,253 @@
...
+                },
+                "firefox_android": {
+                  "version_added": true,
+                  "notes": "Prior to Firefox 68, the <code>candidate</code> property was required; Firefox 68 and later make it optional per a mid-2017 specification update."
```
ddbeck, line 196:
```suggestion
                  "notes": "Before Firefox 68, <code>candidateInfo.candidate</code> is required."
```

ddbeck (changes requested):
One big question here: where did the version data come from? Given the non-real data, I assume it's not been tested. How confident are you in where that data came from?

Regarding your questions:

> 1. The [RTCPeerConnection.addIceCandidate()](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/addIceCandidate) also takes this as an options - just follow same process for that?

Yep, this seems fine as a general approach. But if the guidelines are leading you to do something that feels foolish, please speak up—I wrote them to make _some_ sense. 😄

> 2. The `RTCIceCandidateInit` is also returned by [RTCIceCandidate. toJSON()](https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidate/toJSON) - same process for that?

I'm not sure `toJSON()` needs subfeatures. If there's some rich compatibility story there (e.g., it returns something different than the same version's constructor accepts as an argument), then maybe there's a story to tell, but otherwise I think the compat question is: it spits out a string, y/n?

> 3. When all that done, I just delete the `RTCIceCandidateInit` in BCD?

And on MDN, yes. 👍 🎉

api/RTCIceCandidate.json
hamishwillee: Sure!

api/RTCIceCandidate.json
```diff
@@ -124,6 +124,253 @@
             "standard_track": true,
             "deprecated": false
           }
+        },
+        "candidateInfo_parameter": {
```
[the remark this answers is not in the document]
hamishwillee: Sorry for the confusion. The constructor takes a dictionary of options as a parameter. The question "what should the parameter to the constructor be called".

TLDR: Can I use  `candidateInfo` as the name for this parameter to match MDN docs, ignoring the different name used in the spec (`candidateInitDict`.) and for the same options object passed to `addIceCandidate()` (`candidate`).

----

The long version

Some of the options (sorry for repetition):
- [In MDN](https://developer.mozilla.org/en-US/docs/Web/API/RTCIceCandidate/RTCIceCandidate#parameters) this parameter is called `candidateInfo`. That is what I have chosen for the name of the parameter in BCD.
- [The spec](https://w3c.github.io/webrtc-pc/#dom-rtcicecandidate-constructor) just refers to it by `candidateInitDict`. I don't want to call something Dict while we're busy removing the dict terminology. But some might think this useful as a clear mapping for people reading the spec.
- `candidate` is the name of the parameter passed to the [RTCPeerConnection.addIceCandidate()](https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/addIceCandidate#parameters) used in both the spec and MDN. This is the same dictionary as passed to the constructor - though it is described differently. I don't like this name because it ends up being hard to describe whether you mean the parameter `candidate`, the value `candidate.candidate`, the string `candidate` etc. I mention it because it is the same options object, but I don't want to use it here. And I would prefer not to use in in MDN or `addIceCandidate()`. So kind of a "question" - more for when doing BCD for `addIceCandidate()`

hamishwillee:
Thanks @ddbeck - helpful as always

> One big question here: where did the version data come from? Given the non-real data, I assume it's not been tested. How confident are you in where that data came from?

The truth? You can't handle the truth! Yada yada yada.

It comes from the existing data, so I am fairly confident it no worse than the existing data. Specifically `RTCIceCandidateInit` BCD data copied in verbatim, as shown below.  `RTCIceCandidateInit` is the name of this option as per the spec. 

image

Make sense. 


>> The RTCPeerConnection.addIceCandidate() also takes this as an options - just follow same process for that?

> Yep, this seems fine as a general approach. But if the guidelines are leading you to do something that feels foolish, please speak up—I wrote them to make some sense. 😄

No, they really consistent (which I like). My only question here is what should I call the parameter? As discussed in the "long version part here: https://github.com/mdn/browser-compat-data/pull/11941/files#r692577572 - the spec and mdn call this parameter `candidate` but since the first option is also `candidate` the spec are hard to read ('what candidate do I mean here')

>> The RTCIceCandidateInit is also returned by RTCIceCandidate. toJSON() - same process for that?

> I'm not sure toJSON() needs subfeatures. If there's some rich compatibility story there (e.g., it returns something different than the same version's constructor accepts as an argument), then maybe there's a story to tell, but otherwise I think the compat question is: it spits out a string, y/n?

No. It spits out an object with the same format that you can stringify using the normal javascript method. The format of the object is the same as the dictionary. So presumably the object you get will depend on the version you use. 

My thinking was that this just duplicates the others.

api/RTCIceCandidate.json
```diff
@@ -124,6 +124,253 @@
...
+          "__compat": {
+            "description": "<code>candidateInfo</code> parameter",
+            "spec_url": "https://w3c.github.io/webrtc-pc/#dom-rtcicecandidateinit",
```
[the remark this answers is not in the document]
sideshowbarker: > @Elchi3 @sideshowbarker Do you have any recommendations for parameter spec URLs? There aren't a lot of existing examples. This one points to the spec for the type of the parameter (the dictionary), but I've also seen some of these point to the specification saying that this is a parameter of the method. It makes sense to me that the members would get the dict specification URL, but I'm not sure what to choose at this boundary.

I think it depends a lot on how the spec is structured, and what amount of detail in provides for developers (rather than just for implementors), and where it provides that developer-focused information. In other words, I think it’s a category where the answer is a not-so-satisfying *“it depends”* — in practice, it needs looking at case-by-case.

In this case, the https://w3c.github.io/webrtc-pc/#dom-rtcicecandidateinit URL clearly seems like right target to be citing.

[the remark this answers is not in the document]
ddbeck: Thank you, @sideshowbarker. As "it depends" go, that's not a bad one. 👍

api/RTCIceCandidate.json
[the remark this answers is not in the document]
ddbeck: Ahh, OK. I think I get it now: this constructor and `addIceCandidate()` both accept the same dict as a parameter, but the spec (and the two pages) use a different name for the two parameters. Bummer. I hate all the options, which, as I see them, are:

1. Follow the spec in each case (i.e., use `candidateInitDict` and `candidate`). Truest to the spec, but harder to reuse the text in both places and mentions dicts (though I think that's not too harmful in and of itself).
2. Follow the spec in one case and mirror the other (i.e., `candidate` and `candidate` — doesn't make sense to use the `init` terminology since only one is `init`-ing). Easy on authors, true-ish to the spec, but confusing to readers in the `candidate.candidate` case.
3. Follow the spec nowhere by using with `candidateInfo` in both places. Easy on authors and readers, but departs from the spec. For both option 2 and 3, we might someday end up reverting this, if we ever start validating content against the IDL.

Reluctantly, I accept option 3. Maybe we could aslo open an issue against the spec to ask the editors to rename the parameter in both places as a courtesy to beleaguered technical writers everywhere?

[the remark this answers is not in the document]
ddbeck: Oh, I guess the actual action in this PR: can you change the notes in the parent feature to match the `candidateInfo` name? That will at least eliminate one of the names.

(I guess you could rename `candidate` to `candidateInfo` for `addIceCandidate()` too, but only if you're going to do the content work as well. We don't need to drag everything into the scope of this PR, unless you're excited to do so.)

ddbeck (changes requested):

ddbeck:
Oh I missed a couple things:

> What versions should I use?

Honestly, you could set this stuff to `true` or `false` instead of making an inference about the versions. I don't like adding new non-real data, but if it's dodgy or complicated, I'd rather just admit to that (and clean it up as part of our other data quality efforts).

> No. [`toJSON()`] spits out an object

Of _course_ it does, that's not confusing or anything. 🙄 In any case, I agree: let's not add subfeatures to `toJSON()`. I don't think there's anything interesting to say there.

hamishwillee:
@ddbeck As per your suggestion in  https://github.com/mdn/browser-compat-data/pull/11941#discussion_r693003685 I changed the name in the notes from "options" to "candidateInfo". 

Unfortunately this can no longer merge as now we can't use `true` for chrome and others in javascript APIs. Note further that the options data we are trying to record was deleted in https://github.com/mdn/browser-compat-data/issues/6810 (deleted).

Perhaps we should just close this and pretend there is no compatibility issue? Looking at the table it is _likely_ that most of them were added near the beginning of the existence of the constructor back in FF22 and Chrome 11 timeframes. Further the most recent "known addition" was `usernameFragment ` in FF67. So if you hold that changes back in time before FF67 are probably not worth trying to establish we might just let this go.



Advice please.

hamishwillee:
PS I decided that if we do follow up with `RTCPeerConnection.addIceCandidate()` we would just live with the name `candidate` for the option.

api/RTCIceCandidate.json
```diff
@@ -78,23 +78,23 @@
             "firefox": [
               {
                 "version_added": "44",
-                "notes": "Before Firefox 68, the constructor's <code>options</code> parameter was required."
+                "notes": "Before Firefox 68, the constructor's <code>candidateInfo</code> parameter was required."
```
ddbeck, line 81: Change the notes like this, I think.

```suggestion
                "notes": "Before Firefox 68, the constructor's <code>candidateInfo</code> parameter is required and the <code>candidateInfo</code> object must have a <code>candidate</code> property."
```

api/RTCIceCandidate.json
```diff
@@ -124,6 +124,253 @@
             "standard_track": true,
             "deprecated": false
           }
+        },
```
ddbeck, line 127: Drop all of the changes starting here, on the presumption of compatibility.

ddbeck (changes requested):
> Perhaps we should just close this and pretend there is no compatibility issue - after all that's what the merged PR did :-) ?

Very nearly! I've made a couple suggestions to change the notes and drop the stuff we don't need any more.

api/RTCIceCandidate.json
```diff
@@ -78,23 +78,23 @@
...
               {
                 "version_added": "44",
-                "notes": "Before Firefox 68, the constructor's <code>options</code> parameter was required."
+                "notes": "Before Firefox 68, the constructor's <code>candidateInfo</code> parameter was required."
               },
               {
                 "prefix": "moz",
                 "version_added": "22",
-                "notes": "Before Firefox 68, the constructor's <code>options</code> parameter was required."
+                "notes": "Before Firefox 68, the constructor's <code>candidateInfo</code> parameter was required."
```
hamishwillee, line 86:
```suggestion
                "notes": "Before Firefox 68, the constructor's <code>candidateInfo</code> parameter is required and the <code>candidateInfo</code> object must have a <code>candidate</code> property."
```

api/RTCIceCandidate.json
```diff
@@ -78,23 +78,23 @@
...
               }
             ],
             "firefox_android": [
               {
                 "version_added": "44",
-                "notes": "Before Firefox 68, the constructor's <code>options</code> parameter was required."
+                "notes": "Before Firefox 68, the constructor's <code>candidateInfo</code> parameter was required."
```
hamishwillee, line 92:
```suggestion
                "notes": "Before Firefox 68, the constructor's <code>candidateInfo</code> parameter is required and the <code>candidateInfo</code> object must have a <code>candidate</code> property."
```

api/RTCIceCandidate.json
```diff
@@ -78,23 +78,23 @@
...
               },
               {
                 "prefix": "moz",
                 "version_added": "22",
-                "notes": "Before Firefox 68, the constructor's <code>options</code> parameter was required."
+                "notes": "Before Firefox 68, the constructor's <code>candidateInfo</code> parameter was required."
```
hamishwillee, line 97:
```suggestion
                "notes": "Before Firefox 68, the constructor's <code>candidateInfo</code> parameteris required and the <code>candidateInfo</code> object must have a <code>candidate</code> property."
```

hamishwillee:
Thanks @ddbeck , in that case, we might be done!

api/RTCIceCandidate.json
```diff
@@ -78,23 +78,23 @@
...
               {
                 "version_added": "44",
-                "notes": "Before Firefox 68, the constructor's <code>options</code> parameter was required."
+                "notes": "Before Firefox 68, the constructor's <code>candidateInfo</code> parameter is required and the <code>candidateInfo</code> object must have a <code>candidate</code> property."
               },
               {
                 "prefix": "moz",
                 "version_added": "22",
-                "notes": "Before Firefox 68, the constructor's <code>options</code> parameter was required."
+                "notes": "Before Firefox 68, the constructor's <code>candidateInfo</code> parameteris required and the <code>candidateInfo</code> object must have a <code>candidate</code> property."
```
ddbeck, line 97:
```suggestion
                "notes": "Before Firefox 68, the constructor's <code>candidateInfo</code> parameter is required and the <code>candidateInfo</code> object must have a <code>candidate</code> property."
```

ddbeck (changes requested):
One last tiny fix to make. Thank you!

PR #11679, opened 2021-07-22
Add missing features for DOMRect API

vinyldarkscratch:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing features of the DOMRect API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.1).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/DOMRect

PR #11960, opened 2021-08-10
Add back ClientRect alias for DOMRectReadOnly

foolip:
Since the properties top, right, bottom and left are on DOMRectReadOnly,
it's necessary to treat ClientRect as an alias of both DOMRect and
DOMRectReadOnly. See discussion starting here:
https://github.com/mdn/browser-compat-data/pull/11679#issuecomment-885951832

This reverts https://github.com/mdn/browser-compat-data/pull/11694 but
goes much further in representing the ClientRect support than before.

### DIFF
--- a/api/DOMRectReadOnly.json
+++ b/api/DOMRectReadOnly.json
@@ -5,42 +5,130 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/DOMRectReadOnly",
         "spec_url": "https://drafts.fxtf.org/geometry/#DOMRect",
         "support": {
-          "chrome": {
-            "version_added": "61"
-          },
-          "chrome_android": {
-            "version_added": "61"
-          },
-          "edge": {
-            "version_added": "79"
-          },
-          "firefox": {
-            "version_added": "31"
-          },
-          "firefox_android": {
-            "version_added": "31"
-          },
+          "chrome": [
+            {
+              "version_added": "61"
+            },
+            {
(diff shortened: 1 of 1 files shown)


api/DOMRectReadOnly.json
```diff
@@ -5,42 +5,130 @@
...
+              "version_added": "12",
+              "version_removed": "79"
+            }
+          ],
+          "firefox": [
+            {
+              "version_added": "31"
+            },
+            {
+              "alternative_name": "DOMRect",
```
foolip, line 43: I had to add this to not have a hole in the support for Firefox. It looks like Firefox first just renamed `ClientRect` to `DOMRect` and then split it into `DOMRect` and `DOMRectReadOnly` with all the extra bits. It's a bit weird to say that `DOMRect` was an alias of `DOMRectReadOnly`, but only because of their current relationship. The `DOMRect` in Firefox 27-30 was just like `ClientRect` in 3-26 AFAICT.

I'm not sure if some note would help explain this?

lucalves (approved):

api/DOMRectReadOnly.json
ddbeck: Yeah, I'm not sure there's a note to be written that doesn't exaggerate what the looks-weird-now-but-made-sense-then quality of the data. Let's leave this as-is. 👍

ddbeck (approved):
Thanks, @foolip! I'm happy with this approach. 🎉

PR #11966, opened 2021-08-10
Remove notes about Opera Mini for HTMLCanvasElement

foolip:
BCD doesn't have Opera Mini as a browser, and these are the only
references to "Opera Mini" in any notes.


ddbeck (approved):
Nice clean up. Thank you! 🎉

PR #11967, opened 2021-08-10
Add api.WindowWorkerGlobalScope.structuredClone

lucacasonato:
#### Summary

Adds data entry for `self.structuredClone()`. See spec change:
https://github.com/whatwg/html/commit/1b5099fcf55cc6332fb9af5d36ac4db47749438b

No browser support, except for in Deno.
https://wpt.fyi/results/html/webappapis/structured-clone?label=master&label=experimental&product=chrome&product=edge&product=firefox&product=safari&product=deno&aligned

### DIFF
--- a/api/WindowOrWorkerGlobalScope.json
+++ b/api/WindowOrWorkerGlobalScope.json
@@ -1357,6 +1357,66 @@
             }
           }
         }
+      },
+      "structuredClone": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/WindowOrWorkerGlobalScope/structuredClone",
+          "spec_url": "https://html.spec.whatwg.org/multipage/structured-data.html#dom-structuredclone",
+          "support": {
+            "chrome": {
+              "version_added": false
+            },
+            "chrome_android": {
+              "version_added": false
+            },
+            "deno": {
+              "version_added": "1.13",
+              "partial_implementation": true,
+              "notes": [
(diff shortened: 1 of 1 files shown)


api/WindowOrWorkerGlobalScope.json
```diff
@@ -1357,6 +1357,66 @@
...
+            },
+            "samsunginternet_android": {
+              "version_added": false
+            },
+            "webview_android": {
+              "version_added": false
+            }
+          },
+          "status": {
+            "experimental": false,
```
ddbeck, line 1415: Generally, we need two distinct implementations to set experimental to `true`. Check out [the data guideline](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#choosing-an-experimental-status) for _all_ the nuance.

```suggestion
            "experimental": true,
```

ddbeck (changes requested):
One suggestion in a line comment. Otherwise, this is looking good. Thank you, @lucacasonato.

api/WindowOrWorkerGlobalScope.json
lucacasonato: Whoops. Fixed it now. This is what I get for copy pasting :sweat_smile:

ddbeck: Happens to all of us!

ddbeck (approved):
Thank you! 🎉

PR #11990, opened 2021-08-12
Fix traverse script after yargs upgrade

foolip:
https://github.com/mdn/browser-compat-data/pull/11916 broke the traverse
script, probably due to https://github.com/yargs/yargs/pull/1992.

The problem now is that for `npm run chrome all 90`, `values`
will be ["null", "true", "90"], due to the combining of default and
given argument, where before `values` would have been ["90"].

We were using arrays for both folder and value, which doesn't really
make sense since only the last positional argument can be variadic:
https://github.com/yargs/yargs/blob/master/docs/advanced.md#variadic-positional-arguments

Since we use comma-separated values, just use strings instead to avoid
the problem.

### DIFF
--- a/scripts/traverse.js
+++ b/scripts/traverse.js
@@ -12,13 +12,13 @@ const { argv } = require('yargs').command(
       })
       .positional('folder', {
         describe: 'The folder(s) to test (set to "all" for all folders)',
-        type: 'array',
+        type: 'string',
         default: 'all',
       })
       .positional('value', {
         describe: 'The value(s) to test against',
-        type: 'array',
-        default: ['null', 'true'],
+        type: 'string',
+        default: 'null,true',
       })
       .option('depth', {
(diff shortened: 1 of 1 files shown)


bcoe:
@foolip could I bother you to try `yargs@17.1.1-candidate.0`, which contains a fix from @jly36963.

If it solves the problem we can release ASAP.

lucalves (approved):

ddbeck:
@bcoe Thanks for the fix. I tested the patch release, which resolved the problem for me, and merged https://github.com/mdn/browser-compat-data/pull/12013 with the upgrade.

@foolip the regression is undone, but do you think we ought to make this change anyway?

foolip:
@ddbeck I think what we're doing is a bit weird, but I don't think it's worth making this change when nothing is broken.

PR #11993, opened 2021-08-13
Add permission and feature policy speaker-selection

hamishwillee:
FF92 adds support for the `speaker-selection` permission and associated Permission-Policy (Feature-Policy on FF) - see https://bugzilla.mozilla.org/show_bug.cgi?id=1577199

I tried this on my phone with the first (enumerateDevices) test here: https://wpt.live/audio-output/. Test is not all that helpful. But anyway, if you look at https://bugzilla.mozilla.org/show_bug.cgi?id=1577199#c21 you will see that the evidence is that this feature is not supported on Safari or Chrome.

Tracking other docs work for this in https://github.com/mdn/content/issues/7746#issuecomment-898177902

### DIFF
--- a/api/Permissions.json
+++ b/api/Permissions.json
@@ -1185,7 +1185,7 @@
             }
           },
           "status": {
-            "experimental": false,
+            "experimental": true,
             "standard_track": true,
             "deprecated": false
           }
--- a/api/Permissions.json
+++ b/api/Permissions.json
@@ -1140,6 +1140,56 @@
             "deprecated": true
           }
         }
+      },
+      "speaker-selection_permission": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/Permissions/speaker-selection",
+          "spec_url": "https://w3c.github.io/permissions/#dom-permissionname-speaker-selection",
+          "description": "<code>speaker-selection</code> permission",
+          "support": {
+            "chrome": {
+              "version_added": false
+            },
+            "chrome_android": {
+              "version_added": false
+            },
+            "edge": {
+              "version_added": false
+            },
+            "firefox": {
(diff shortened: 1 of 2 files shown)


api/Permissions.json
```diff
@@ -1140,6 +1140,56 @@
...
+            },
+            "samsunginternet_android": {
+              "version_added": false
+            },
+            "webview_android": {
+              "version_added": false
+            }
+          },
+          "status": {
+            "experimental": false,
```
ddbeck, line 1188: Experimental is usually `true` for single implementations (see [Choosing an experimental status](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#choosing-an-experimental-status))

```suggestion
            "experimental": true,
```

http/headers/feature-policy.json
```diff
@@ -1319,6 +1319,55 @@
...
+              },
+              "safari_ios": {
+                "version_added": false
+              },
+              "samsunginternet_android": {
+                "version_added": "14.0"
+              },
+              "webview_android": {
+                "version_added": "84"
+              }
```
ddbeck, line 1362: These values look suspect, based on the other data here and your PR description. Is this a copy and paste glitch, maybe?

ddbeck (changes requested):
A couple of things to fix up or clarify here, @hamishwillee. Thank you!

api/Permissions.json
hamishwillee: Fair enough!

http/headers/feature-policy.json
hamishwillee: You're right - sloppy. Fixed.

ddbeck (approved):
Thank you! 🎉

Issue #11994, opened 2021-08-13
api.InputEvent.InputEvent - Safari 14.1.2 does not support initData in InputEvent constructor correctly 


ZhangYiJiang:
Browser: Safari 14.1.2  
OS: macOS Catalina 10.15.7

#### What information was incorrect, unhelpful, or incomplete?

`InputEvent()` constructor's compatibility chart suggests Safari is fully compliant, but it does not correctly support `initData`. This makes it difficult to use `InputEvent` constructor since the properties defined in `initData` are immutable (readonly) on the actual event object, which means the constructor cannot be used. 

#### What did you expect to see?

A note warning about Safari not supporting `initData` correctly

#### Did you test this? If so, how?

Webkit bug tracker: https://bugs.webkit.org/show_bug.cgi?id=170416

In Safari's console - 

```
> e = new InputEvent('beforeinput', { inputType: 'insertText' }) 
< InputEvent {isTrusted: false, inputType: "", data: null, dataTransfer: null, getTargetRanges: function, …}
> e.inputType
< ""
```

Compared to in Firefox 

``` 
> e = new InputEvent('beforeinput', { inputType: 'insertText' })
< beforeinput { target: null, isTrusted: false, isComposing: false, inputType: "insertText", detail: 0, layerX: 0, layerY: 0, which: 0, rangeOffset: 0, SCROLL_PAGE_UP: -32768, … }
> e.inputType
< "insertText"
``` 

<details>
<summary>MDN page report details</summary>

* Query: `api.InputEvent.InputEvent`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/InputEvent/InputEvent
* Report started: 2021-08-13T16:21:29.960Z

</details>

gsnedders:
#7287 is relevant here; WebKit has:

```
dictionary InputEventInit : UIEventInit {
    DOMString? data = null;
};
```

As such the rest of the dictionary is unsupported.

PR #11995, opened 2021-08-13
Add partial implementation notes for Safari `InputEvent()` constructor

ZhangYiJiang:
#### Summary

Add partial implementation notes for Safari `InputEvent()` constructor 

In Safari's console - 

```
> e = new InputEvent('beforeinput', { inputType: 'insertText' }) 
< InputEvent {isTrusted: false, inputType: "", data: null, dataTransfer: null, getTargetRanges: function, …}
> e.inputType
< ""
```

Compared to in Firefox 

``` 
> e = new InputEvent('beforeinput', { inputType: 'insertText' })
< beforeinput { target: null, isTrusted: false, isComposing: false, inputType: "insertText", detail: 0, layerX: 0, layerY: 0, which: 0, rangeOffset: 0, SCROLL_PAGE_UP: -32768, … }
> e.inputType
< "insertText"
``` 

https://bugs.webkit.org/show_bug.cgi?id=170416

#### Related issues

Fixes https://github.com/mdn/browser-compat-data/issues/11994

### DIFF
--- a/api/InputEvent.json
+++ b/api/InputEvent.json
@@ -84,12 +84,12 @@
             "safari": {
               "version_added": "10.1",
               "partial_implementation": true,
-              "notes": [
-                "Does not support inputEventInit parameter, see <a href='https://bugs.webkit.org/show_bug.cgi?id=170416'>bug 170416</a>"
-              ]
+              "notes": "The <code>inputEventInit</code> parameter is not supported. See <a href='https://webkit.org/b/170416'>bug 170416</a>."
             },
             "safari_ios": {
-              "version_added": "10.3"
+              "version_added": "10.3",
+              "partial_implementation": true,
+              "notes": "The <code>inputEventInit</code> parameter is not supported. See <a href='https://webkit.org/b/170416'>bug 170416</a>."
             },
             "samsunginternet_android": {
               "version_added": "8.0"
--- a/api/InputEvent.json
+++ b/api/InputEvent.json
@@ -82,7 +82,11 @@
               "version_added": "44"
             },
             "safari": {
-              "version_added": "10.1"
+              "version_added": "10.1",
+              "partial_implementation": true,
+              "notes": [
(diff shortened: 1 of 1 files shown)


ddbeck (changes requested):
Hi @ZhangYiJiang and thank your for this PR. I see that it's your first to the project—welcome!

I've got a few suggestions for you.

- Most important: this appears to apply to safari_ios as well. Can you copy your changes to that browser too?
- In a line comment, I've suggested some formatting changes.
- Optionally: if you prefer to help make this data richer, you could create an `inputEventInit_parameter` subfeature and report this in that subfeature. [See our parameter features guideline for more information](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#parameters-and-parameter-object-features). This is optional, but if you'd like to do it, I'm happy to review the expansion to the data.

CI is failing with a couple of suggestions too (to use the preferred URL format and to use a string rather than array of strings for the note).

Thanks again for your help!

api/InputEvent.json
```diff
@@ -82,7 +82,11 @@
               "version_added": "44"
             },
             "safari": {
-              "version_added": "10.1"
+              "version_added": "10.1",
+              "partial_implementation": true,
+              "notes": [
+                "Does not support inputEventInit parameter, see <a href='https://bugs.webkit.org/show_bug.cgi?id=170416'>bug 170416</a>"
```
ddbeck, line 88:
```suggestion
                "The <code>inputEventInit</code> parameter is not supported. See <a href='https://bugs.webkit.org/show_bug.cgi?id=170416'>bug 170416</a>."
```

ZhangYiJiang:
I believe I've fixed the issues. I've opted not to create a subfeature because the `inputEventInit` is fairly integral to the feature itself - without it the constructor is almost entirely useless since the properties defined in `inputEventInit` are readonly and cannot be set on the object after initialization

ddbeck (approved):
Makes sense. Thank you, @ZhangYiJiang! 🎉

PR #12016, opened 2021-08-16
Bug 1614919 browser.identity.launchWebAuthFlow() exposes redirect_url

rebloor:
#### Summary
Added notes to:
* getRedirectURL regarding the use of a loopback address
* launchWebAuthFlow  noting the need for redirect_uri to be set to the URL returned by identity.getRedirectURL()

Ran npm test, no issues found

#### Related issues
[Bug 1614919](https://bugzilla.mozilla.org/show_bug.cgi?id=1614919)

### DIFF
--- a/webextensions/api/identity.json
+++ b/webextensions/api/identity.json
@@ -12,9 +12,15 @@
               "edge": {
                 "version_added": "79"
               },
-              "firefox": {
-                "version_added": "53"
-              },
+              "firefox": [
+                {
+                  "version_added": "53"
+                },
+                {
+                  "version_added": "86",
+                  "notes": "The loopback address http://127.0.0.1/mozoauth2/[subdomain of URL returned by identity.getRedirectURL()] can be used as the redirect URL to enable the use of Google OAuth flows."
+                }
+              ],
               "firefox_android": {
                 "version_added": false
               },
@@ -45,6 +51,10 @@
                   "version_added": "53",
                   "version_removed": "63",
                   "notes": "The <code>redirect_uri</code> URL parameter is required."
+                },
+                {
+                  "version_added": "75",
+                  "notes": "The <code>redirect_uri</code> URL parameter must be set to the URL returned by <code>identity.getRedirectURL()</code>."
                 }
               ],
               "firefox_android": {
--- a/webextensions/api/identity.json
+++ b/webextensions/api/identity.json
@@ -18,7 +18,7 @@
                 },
                 {
                   "version_added": "86",
-                  "notes": "The loopback address http://127.0.0.1/mozoauth2/[subdomain of URL returned by identity.getRedirectURL()] can be used as the redirect URL to enable the use of Google OAuth flows."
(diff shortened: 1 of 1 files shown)


webextensions/api/identity.json
```diff
@@ -12,9 +12,15 @@
...
-              "firefox": {
-                "version_added": "53"
-              },
+              "firefox": [
+                {
+                  "version_added": "53"
+                },
+                {
+                  "version_added": "86",
+                  "notes": "The loopback address http://127.0.0.1/mozoauth2/[subdomain of URL returned by identity.getRedirectURL()] can be used as the redirect URL to enable the use of Google OAuth flows."
```
Rob--W, line 21: This level of detail is too much. Let's refer to the inline documentation instead, like this:

```html
The <code>redirect_url</code> parameter now supports a loopback address, see <a href="https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/identity#getting_the_redirect_url">Getting the redirect URL</a>.
```

Rob--W (changes requested):

webextensions/api/identity.json
```diff
@@ -12,9 +12,15 @@
...
-              "firefox": {
-                "version_added": "53"
-              },
+              "firefox": [
+                {
+                  "version_added": "53"
+                },
+                {
+                  "version_added": "86",
+                  "notes": "The <code>redirect_url</code> parameter now supports a loopback address, see <a href='https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/API/identity#getting_the_redirect_url'>Getting the redirect URL</a> for details."
```
Rob--W, line 21: The `redirect_url` property belongs to the `launchWebAuthFlow` method, not here.

webextensions/api/identity.json
```diff
@@ -45,6 +51,10 @@
                   "version_added": "53",
                   "version_removed": "63",
```
Rob--W, line 52: This is rendered incorrectly, as "Removed in 63 and later": https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/identity/launchWebAuthFlow#browser_compatibility

Could you change this to `"version_added": 63",` and note that the parameter is optional? It has the same meaning and would be rendered correctly.

webextensions/api/identity.json
```diff
@@ -45,6 +51,10 @@
...
                   "notes": "The <code>redirect_uri</code> URL parameter is required."
+                },
+                {
+                  "version_added": "75",
+                  "notes": "The <code>redirect_uri</code> URL parameter must be set to the URL returned by <code>identity.getRedirectURL()</code>."
```
Rob--W, line 57: With the lot of comments about `redirect_url`, how about adding a new row? `redirect_url`.

Note that Chrome, Opera and Edge does not support the property; their documentation is at https://developer.chrome.com/docs/extensions/reference/identity/#type-WebAuthFlowDetails

Rob--W (changes requested):

rebloor:
@Rob--W Thanks for the feedback, would you like to confirm the changes are expected?

Rob--W (approved):
Looks great, thanks! In this form, it is very apparent that the `redirect_url` parameter has gone through several revisions.

PR #12019, opened 2021-08-16
Update all browsers versions for KeyboardEvent API

vinyldarkscratch:
This PR updates and corrects the real values for all browsers for the `KeyboardEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.2).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/KeyboardEvent

### DIFF
--- a/api/KeyboardEvent.json
+++ b/api/KeyboardEvent.json
@@ -55,10 +55,10 @@
           "description": "<code>KeyboardEvent()</code> constructor",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "26"
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": "26"
             },
             "edge": {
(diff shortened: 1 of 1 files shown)


api/KeyboardEvent.json
```diff
@@ -300,10 +300,10 @@
           "spec_url": "https://w3c.github.io/uievents/#dom-keyboardevent-getmodifierstate",
           "support": {
             "chrome": {
-              "version_added": "31"
+              "version_added": "30"
```
foolip, line 303: Curiously it looks like the Opera data is already correct if mirroring Chromium 30. I'm guessing that was originally due to a mistake, but now made correct.

foolip (approved):

PR #12025, opened 2021-08-16
Add all browsers versions for TransitionEvent API

vinyldarkscratch:
This PR adds real values for all browsers for the `TransitionEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.2).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/TransitionEvent

### DIFF
--- a/api/TransitionEvent.json
+++ b/api/TransitionEvent.json
@@ -38,17 +38,29 @@
             "version_added": "10"
           },
           "opera": {
-            "version_added": true
+            "version_added": "≤12.1"
           },
           "opera_android": {
-            "version_added": true
+            "version_added": "≤12.1"
           },
-          "safari": {
-            "version_added": true
-          },
-          "safari_ios": {
-            "version_added": true
-          },
+          "safari": [
+            {
+              "version_added": "7"
+            },
+            {
+              "prefix": "WebKit",
+              "version_added": "≤4"
+            }
+          ],
+          "safari_ios": [
+            {
+              "version_added": "7"
+            },
+            {
+              "prefix": "WebKit",
+              "version_added": "≤3"
+            }
+          ],
           "samsunginternet_android": [
             {
               "version_added": "2.0"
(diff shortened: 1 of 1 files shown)


api/TransitionEvent.json
```diff
@@ -180,12 +192,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/TransitionEvent/initTransitionEvent",
           "support": {
             "chrome": {
-              "version_added": false,
-              "notes": "Removal version unknown."
+              "version_added": false
```
foolip, line 195: Actually there was something to those notes, https://developer.apple.com/documentation/webkitjs/webkittransitionevent/1806991-initwebkittransitionevent used to exist, but was removed from WebKit in https://trac.webkit.org/changeset/103746/webkit.

I think we also need to update the custom tests for this to try creating a `WebKitTransitionEvent` to cover everything.

foolip (changes requested):

api/TransitionEvent.json
vinyldarkscratch: I think having a note saying "Removal version unknown" is unhelpful, and if we're to keep this data, we should make it machine-readable with `version_removed: true`.  Though based upon that commit and the fact that the collector doesn't see any support, we can guesstimate that removal happened around or before Chrome 27 (when the interface was unprefixed, since that's what our collector is testing) and Safari 6 (based upon the date).  Since it was removed from browsers so long ago, I think we should remove this feature from BCD.

foolip: I agree the note should be removed somehow. Unfortunately the entry can't be removed until January since it was removed in Edge 79. Can you make a guess of version added and removed for now? It doesn't matter if it's not perfect, it'll be removed soon enough.

vinyldarkscratch: Alright, I've tracked this down and have found the versions when they were removed!

foolip (approved):

PR #12026, opened 2021-08-16
Add all browsers versions for UIEvent API

vinyldarkscratch:
This PR adds real values for all browsers for the `UIEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.2).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/UIEvent

### DIFF
--- a/api/UIEvent.json
+++ b/api/UIEvent.json
@@ -6,10 +6,10 @@
         "spec_url": "https://w3c.github.io/uievents/#idl-uievent",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "1"
           },
           "chrome_android": {
-            "version_added": true
+            "version_added": "18"
           },
           "edge": {
             "version_added": "12"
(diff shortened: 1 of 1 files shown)


api/UIEvent.json
```diff
@@ -255,13 +255,15 @@
...
+              "version_removed": "45"
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": "18",
+              "version_removed": "45"
             },
             "edge": {
-              "version_added": "12"
+              "version_added": false
```
foolip, line 266: Many of the changes from a specific version to false in this PR have me wondering... since `UIEvent` is an abstract base class and browsers don't fire any events that are direct `UIEvent` instances, maybe `layerX` was just on `MouseEvent` or something and should be considered supported? In other words, it might be a case of https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#apis-moved-on-the-prototype-chain

Can you do some more checking of this?

vinyldarkscratch: I've reverted the changes from true to false for the time being to save some reviewing time.  I'll revisit these changes in a later PR!

PR #12027, opened 2021-08-16
Add all browsers versions for WheelEvent API

vinyldarkscratch:
This PR adds real values for all browsers for the `WheelEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.2).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/WheelEvent

### DIFF
--- a/api/WheelEvent.json
+++ b/api/WheelEvent.json
@@ -6,10 +6,10 @@
         "spec_url": "https://w3c.github.io/uievents/#interface-wheelevent",
         "support": {
           "chrome": {
-            "version_added": "31"
+            "version_added": "1"
           },
           "chrome_android": {
-            "version_added": true
+            "version_added": "18"
           },
           "edge": {
             "version_added": "12"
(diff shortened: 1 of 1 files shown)


foolip:
There was a lot of Safari 6.1 in the original data so I wanted to confirm that's definitely wrong by testing https://mdn-bcd-collector.appspot.com/tests/api/WheelEvent in Safari 5 on BrowserStack, but it seems broken. Is the setup no longer working for old Safari?

Also it looks like there was originally a `initWebKitWheelEvent` which was removed in Chromium in https://chromium.googlesource.com/chromium/src/+/73fa6fe327bfe6230b7dc89704b850a8d28b59ae and maybe in WebKit separately, so there was probably a `WebKitWheelEvent` too. Can you add custom tests for that and retest?

foolip:
Actually there was no `WebKitWheelEvent` I've found:
https://github.com/foolip/mdn-bcd-collector/blob/0ad73398f39ed35df2cbc7e6824cfddcd86bc8e7/custom-idl/uievents.idl#L95-L98

foolip:
I've filed https://github.com/foolip/mdn-bcd-collector/issues/1416 about the remaining problem.

foolip (approved):

PR #12028, opened 2021-08-16
Add Firefox versions for VTTCue API

vinyldarkscratch:
This PR adds real values for Firefox and Firefox Android for the `VTTCue` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.2).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/VTTCue

### DIFF
--- a/api/VTTCue.json
+++ b/api/VTTCue.json
@@ -15,10 +15,10 @@
             "version_added": "12"
           },
           "firefox": {
-            "version_added": "26"
+            "version_added": "31"
           },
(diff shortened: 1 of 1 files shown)


lucalves (approved):

foolip (approved):
This now matches `TextTrackCue` which makes much more sense. The previous state would have failed a consistency check for inheritance if we had one.

Issue #11288, opened 2021-06-28
`api.Notification` - Notes for Chrome are incorrectly structured and dubious


ddbeck:
#### What information was incorrect, unhelpful, or incomplete?

The notes field for the Chrome entry for `api.Notifications` contains a very long note that seems to be more like four notes in one:

https://github.com/mdn/browser-compat-data/blob/31e399e619b4180a42174d794c825d43af821238/api/Notification.json#L11

(It's unclear what all the notes mean, exactly (e.g., it references `navigator.webkitNotifications`, which I'm not certain is a real thing), so I didn't want to simply split them.)

#### What did you expect to see?

At very least, an array of notes.

#### Did you test this? If so, how?

No. The veracity of the notes needs to investigating.

#### Can you link to any release notes, bugs, pull requests, or MDN pages related to this?

Probably related to some of the problems with `api.Notification.Notification` raised in https://github.com/mdn/browser-compat-data/issues/7923.

foolip:
These notes were rewritten in https://github.com/mdn/browser-compat-data/pull/12032 + https://github.com/mdn/browser-compat-data/pull/15822.

PR #12032, opened 2021-08-17
Correct data for Notification API

queengooborg:
This PR updates all of the data for the Notification API, adding real values and correcting the existing data.  Changes are as described:

- The data talks about support with an older version of the specification in early Chrome versions.  From what I have determined, the `webkitNotifications` API has a completely different method of interaction from the Notification API we know today.  The data represents this in the form of notes and a support statement with a `prefix`; this is wrong, as the name differs with an extra "s" at the end, and the interaction differs enough that I would consider it to be a completely different API.  With all of this in consideration, I've decided to remove this data, as we would per the irrelevant feature removal guidelines if this were recorded as a separate interface.
- The note regarding lack of `permission` support in earlier Chrome versions has been removed.  We're already tracking this feature in BCD, so there's no reason to record this in a note.
- Firefox Android's prefix was set to "webkit", which was incorrect.  This PR fixes that and sets it to "moz" appropriately.
- The prefix data was copied to the `icon` feature accidentally.  This removes the prefix statements, as this feature was never supported behind a prefix of its own.
- Finally, this updates all the data using results from the mdn-bcd-collector project (v3.2.9), adding real values where missing and correcting the existing data as needed.

### DIFF
--- a/api/Notification.json
+++ b/api/Notification.json
@@ -5,18 +5,13 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/Notification",
         "spec_url": "https://notifications.spec.whatwg.org/#api",
         "support": {
-          "chrome": [
-            {
-              "version_added": "22",
-              "notes": "Before Chrome 22, the support for notification followed an old prefixed version of the specification and used the <code>navigator.webkitNotifications</code> object to instantiate a new notification. Before Chrome 32, <code>Notification.permission</code> was not supported. Before Chrome 42, service worker additions were not supported. Starting in Chrome 49, notifications do not work in incognito mode."
-            },
-            {
-              "prefix": "webkit",
-              "version_added": "5"
-            }
-          ],
+          "chrome": {
+            "version_added": "20",
+            "notes": "Before Chrome 20, the support for notification followed an old prefixed version of the specification and used the <code>navigator.webkitNotifications</code> object to instantiate a new notification. Before Chrome 42, service worker additions were not supported. Starting in Chrome 49, notifications do not work in incognito mode."
+          },
           "chrome_android": {
-            "version_added": true
+            "version_added": "25",
+            "notes": "Before Chrome 20, the support for notification followed an old prefixed version of the specification and used the <code>navigator.webkitNotifications</code> object to instantiate a new notification. Before Chrome 42, service worker additions were not supported. Starting in Chrome 49, notifications do not work in incognito mode."
           },
           "edge": {
             "version_added": "14"
@@ -35,7 +30,7 @@
               "version_added": "22"
             },
             {
-              "prefix": "webkit",
+              "prefix": "moz",
               "version_added": "4"
             }
           ],
@@ -43,10 +38,10 @@
             "version_added": false
           },
           "opera": {
-            "version_added": "25"
+            "version_added": "23"
           },
           "opera_android": {
-            "version_added": true
+            "version_added": "24"
           },
           "safari": {
             "version_added": "7"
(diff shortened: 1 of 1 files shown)


ddbeck:
Probably related to https://github.com/mdn/browser-compat-data/issues/11288 and https://github.com/mdn/browser-compat-data/issues/7923.

And a question for you, Vinyl, on this point:

> the name differs with an extra "s" at the end, and the interaction differs enough that I would consider it to be a completely different API. With all of this in consideration, I've decided to remove this data, as we would per the irrelevant feature removal guidelines if this were recorded as a separate interface.

I don't understand the sequence of events here, or whether the irrelevance guideline applies. What happened to the alternative name? Does it no longer exist? Or are you ignoring it because it's different API?

vinyldarkscratch:
Apologies, let me attempt to clarify what's going on.

Before Chrome 20, notifications were supported using a separate API, `NotificationCenter`, which was accessible via the `window.webkitNotifications` (not `navigator`) variable.  This interface involves creating notifications using its own `createNotification()` method, and it doesn't expose the same features (`body`, `timestamp`, etc.).  The way BCD is representing this is with a prefix statement, but this is inaccurate, as it's an entirely different API, not a prefix or an alt. name of the existing one.  If we continued to call it a prefix/alt. name, I feel this would be misleading for web developers.

If we were to have a file for this `NotificationCenter` API, we would remove it anyways according to the irrelevant features guideline.  It has been deprecated over two years ago and was only available in Chrome.  Thus, I came to the conclusion that simply removing this information was the best course of action.

> Probably related to #11288 and #7923.

It is related to #11288, yes -- I wouldn't go as far to say this "fixes" #11288, but it is a step closer!

ddbeck:
Thank you, Vinyl. That clears up a lot. Based on what you've described here, it completely makes sense to me to treat the `webkitNotifications` API as a separate thing.

One last point of clarification, for this: 

>  It has been deprecated over two years ago and was only available in Chrome.

When you say "deprecated" do you mean removed two years ago? Or does the API still exist but emits warnings or something?

vinyldarkscratch:
I meant to say "removed" rather than "deprecated", whoops!

api/Notification.json
```diff
@@ -5,18 +5,13 @@
...
-              "notes": "Before Chrome 22, the support for notification followed an old prefixed version of the specification and used the <code>navigator.webkitNotifications</code> object to instantiate a new notification. Before Chrome 32, <code>Notification.permission</code> was not supported. Before Chrome 42, service worker additions were not supported. Starting in Chrome 49, notifications do not work in incognito mode."
-            },
-            {
-              "prefix": "webkit",
-              "version_added": "5"
-            }
-          ],
+          "chrome": {
+            "version_added": "20",
+            "notes": "Before Chrome 42, service worker additions were not supported. Starting in Chrome 49, notifications do not work in incognito mode."
```
ddbeck, line 10: OK, these are clearly two distinct notes now. Can we split these notes into an array?

* From Chrome 49…
* Before Chrome 42…

ddbeck (changes requested):
OK, cool. I've run out of time to fully review this, but I did catch one thing

ddbeck (commented):
One last question, Vinyl: is the moz prefix on the constructor real? As in, is it genuinely `mozNotification.mozNotification()`?

queengooborg:
I did verify the prefix is correct and in earlier versions, developers will need to use `new mozNotification()`!

ddbeck (approved):
Thanks! 🎉

PR #12033, opened 2021-08-17
Add Firefox versions for PerformanceEntry API

queengooborg:
This PR adds real values for Firefox and Firefox Android for the `PerformanceEntry` API, based upon browser bug history.

Bugs:
https://bugzil.la/822480
https://bugzil.la/1047848

### DIFF
--- a/api/PerformanceEntry.json
+++ b/api/PerformanceEntry.json
@@ -27,10 +27,10 @@
             "version_added": "12"
           },
           "firefox": {
-            "version_added": "38"
+            "version_added": "35"
           },
           "firefox_android": {
-            "version_added": "38"
+            "version_added": "35"
           },
           "ie": {
             "version_added": "10"
@@ -103,10 +103,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "38"
+              "version_added": "35"
             },
             "firefox_android": {
-              "version_added": "38"
+              "version_added": "35"
             },
             "ie": {
               "version_added": "10"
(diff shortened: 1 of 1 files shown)


foolip (changes requested):
The original "25" for Firefox Android came from https://github.com/mdn/browser-compat-data/pull/1298 with no discussion about the version that I can see.

https://developer.mozilla.org/en-US/docs/Web/API/Performance#browser_compatibility says `performance.getEntries()` and a bunch of other things were shipped in Firefox 35, but that implies support for `PerformanceEntry` since that's what's returned. Can you double check which is correct and change things to be consistent?

vinyldarkscratch:
I think the original "25" was either a mistake (copied from Chrome Android), or bad data from the wiki tables (which could have also been caused by someone copying the value from Chrome Android).

Regarding the values in this PR however, it seems this is a bit of a problem with our test.  I'm submitting a PR to the collector now to resolve this and use `getEntries()` instead!

vinyldarkscratch:
The new collector updates will fix this, yes!  Since my laptop's out of commission for a while though, it'll be some time before I can get updated results.

vinyldarkscratch:
Hmmm...the recent collector results (v3.1.12) also indicate support in Firefox 38.  Looks like I'll need to revise the custom test again and try to generate an entry before getting any of them, probably.

queengooborg:
https://bugzilla.mozilla.org/show_bug.cgi?id=822480 marks this interface as implemented in Firefox 31.  Based upon the collector results for `getEntries()`, I'm assuming that it was behind a flag to begin with, so I'll set this to "35" to match.

foolip (approved):
Matching the methods that return`PerformanceEntry` makes sense!

PR #12039, opened 2021-08-17
Adds `color-contrast()` from Color Level 5

argyleink:
#### Summary
I noticed other color functions from color level 5 are available to query, but `color-contrast()` was missing. 

Spec https://www.w3.org/TR/css-color-5/#colorcontrast
Webkit commit https://trac.webkit.org/changeset/273683/webkit/

### DIFF
--- a/css/types/color.json
+++ b/css/types/color.json
@@ -258,7 +258,7 @@
                 ]
               },
               "safari_ios": {
-                "version_added": "15"
+                "version_added": false
               },
               "samsunginternet_android": {
                 "version_added": false
--- a/css/types/color.json
+++ b/css/types/color.json
@@ -248,7 +248,14 @@
                 "version_added": false
               },
               "safari": {
-                "version_added": "15"
+                "version_added": "15",
+                "flags": [
+                  {
+                    "type": "preference",
+                    "name": "CSS color-contrast()",
+                    "value_to_set": "true"
+                  }
+                ]
               },
               "safari_ios": {
                 "version_added": "15"
--- a/css/types/color.json
+++ b/css/types/color.json
@@ -281,6 +281,56 @@
             }
           }
         },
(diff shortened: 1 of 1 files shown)


css/types/color.json
```diff
@@ -281,6 +281,56 @@
...
+                "version_added": false
+              },
+              "opera": {
+                "version_added": false
+              },
+              "opera_android": {
+                "version_added": false
+              },
+              "safari": {
+                "version_added": "15"
```
argyleink, line 315: needs enabled with the flag `CSSColorContrastEnabled` which is in the dropdown menu.. todo: need to go find the way to reference this.

argyleink:
this PR also helps resolve this warning message [on MDN](https://developer.mozilla.org/en-US/docs/Web/CSS/color_value/color-contrast())

Screen Shot 2021-08-17 at 9 34 29 AM

css/types/color.json
ddbeck: Thanks for getting started on this, @argyleink!

This can be done with the `flags` key. See https://github.com/mdn/browser-compat-data/blob/main/css/properties/scroll-behavior.json#L33 for an example and [the schema docs](https://github.com/mdn/browser-compat-data/blob/main/schemas/compat-data-schema.md#flags) for details. Also note, we usually use the string from the actual UI, rather than name of the preference in source.

I'm not sure that Safari for iOS supports flags, so depending on the menu options in iOS, it might be that it's still `false`.

ddbeck (changes requested):
Also, the CI is failing because of the sort order. `npm run fix` ought to resolve this (and `npm test` ought to confirm).

css/types/color.json
argyleink: thanks for the pointers! items have been attended to 👍🏻

PR #12055, opened 2021-08-18
Add Safari preview support for BroadcastChannel behind flag

lukewarlow:
#### Summary

Safari Tech Preview now supports Broadcast Channel behind an experimental feature flag.

image

### DIFF
--- a/api/BroadcastChannel.json
+++ b/api/BroadcastChannel.json
@@ -43,8 +43,7 @@
             "flags": [
               {
                 "name": "BroadcastChannel API",
-                "type": "preference",
-                "value_to_set": "enabled"
+                "type": "preference"
               }
             ]
           },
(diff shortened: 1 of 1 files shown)


api/BroadcastChannel.json
```diff
@@ -39,7 +39,14 @@
             "version_added": "41"
           },
           "safari": {
-            "version_added": false
+            "version_added": "preview",
+            "flags": [
+              {
+                "name": "BroadcastChannel API",
+                "type": "preference",
+                "value_to_set": "enabled"
```
ddbeck, line 47: Conventionally, we've omitted `value_to_set` for Safari prefs since there's no enabled string in the UI and it seems a bit noisy to include the check mark. This change would need to be applied throughout.

```suggestion
                "type": "preference"
```

ddbeck (changes requested):
Thanks for this PR, @lukewarlow. A suggestion for you in a line comment.

ddbeck (approved):
Thank you! 🎉

PR #12084, opened 2021-08-21
Add Chromium versions for DeviceMotionEvent API

queengooborg:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `DeviceMotionEvent` API, based upon commit history and date.

Commit: https://source.chromium.org/chromium/chromium/src/+/8136adc9f9f8e570d8a39a1fb4c19dd6cb67b548

### DIFF
--- a/api/DeviceMotionEvent.json
+++ b/api/DeviceMotionEvent.json
@@ -104,10 +104,10 @@
           "spec_url": "https://w3c.github.io/deviceorientation/#ref-for-dom-devicemotionevent-acceleration③",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "11"
             },
             "chrome_android": {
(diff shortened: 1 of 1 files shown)


vinyldarkscratch:
The commit I linked had a date prior to the Chrome version that's indicated in BCD (Chrome 11), so I set it to Chrome 11 accordingly.

foolip:
> The commit I linked had a date prior to the Chrome version that's indicated in BCD (Chrome 11), so I set it to Chrome 11 accordingly.

Given `Conditional=DEVICE_ORIENTATION` in the IDL here, it's hard to use source code to determine when `DeviceMotionEvent` itself was enabled, so let's just trust the existing data and use that as you say.

PR #12091, opened 2021-08-21
Add Firefox versions for api.FileSystemDirectoryReader.readEntries

queengooborg:
This PR adds real values for Firefox and Firefox Android for the `readEntries` member of the `FileSystemDirectoryReader` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.9).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/FileSystemDirectoryReader/readEntries

### DIFF
--- a/api/FileSystemDirectoryReader.json
+++ b/api/FileSystemDirectoryReader.json
@@ -69,10 +69,10 @@
               "version_added": false
             },
             "firefox": {
-              "version_added": "51"
+              "version_added": "50"
             },
             "firefox_android": {
-              "version_added": "51"
+              "version_added": "50"
             },
             "ie": {
               "version_added": false
--- a/api/FileSystemDirectoryReader.json
+++ b/api/FileSystemDirectoryReader.json
@@ -69,10 +69,10 @@
               "version_added": false
             },
             "firefox": {
-              "version_added": true
+              "version_added": "51"
             },
             "firefox_android": {
-              "version_added": true
+              "version_added": "51"
             },
             "ie": {
               "version_added": false


vinyldarkscratch:
It looks like this interface was originally called `DirectoryReader`, but then renamed to `FileSystemDirectoryReader`.  What's odd, though, is that the [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1295570) and [changeset](https://hg.mozilla.org/mozilla-central/rev/4a589e795dcf0942de5cef5361ec41c8b9d73a5a) both indicate that the change was made in Firefox 51, yet our tests are suggesting otherwise?

Also, this API looks to have been added in https://bugzilla.mozilla.org/show_bug.cgi?id=1265767, which suggests Firefox 50.  I ran some tests in Firefox 47, 49, and 50 to see if `DirectoryReader` would be defined, but I turned up nothing.

TL;DR: this PR looks to be correct.

foolip:
> It looks like this interface was originally called `DirectoryReader`, but then renamed to `FileSystemDirectoryReader`. What's odd, though, is that the [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1295570) and [changeset](https://hg.mozilla.org/mozilla-central/rev/4a589e795dcf0942de5cef5361ec41c8b9d73a5a) both indicate that the change was made in Firefox 51, yet our tests are suggesting otherwise?

At the time, Gecko's Web IDL had `[NoInterfaceObject]`, so neither the previous `DirectoryReader` nor the new `FileSystemDirectoryReader` would have been exposed on the global object, which is all the generated tests can look for. We'd need to test an instance here.

I think there are two options:
- Mark it as supported since Firefox 50 and ignore the fact that the interface object wasn't exposed
- Partial support in Firefox 50 because the interface object wasn't exposed

I favor the first approach, because the interface object doesn't matter here in practice, it's not a constructor and isn't used in any way when actually using this API, AFAICT.

queengooborg:
I like the first approach as well, let's go for that!  I've made the appropriate changes to this PR.

foolip (approved):

PR #12092, opened 2021-08-21
Add Safari versions for api.HTMLCollection.length

vinyldarkscratch:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `length` member of the `HTMLCollection` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.9).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/HTMLCollection/length

### DIFF
--- a/api/HTMLCollection.json
+++ b/api/HTMLCollection.json
@@ -127,10 +127,10 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": true
+              "version_added": "≤4"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "≤3"
             },
             "samsunginternet_android": {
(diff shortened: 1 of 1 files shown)


api/HTMLCollection.json
```diff
@@ -127,10 +127,10 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": true
+              "version_added": "≤4"
```
foolip, line 130: Let's just set it to 1. The interface makes no sense without the length property, and the rest of the data is 1.

foolip (changes requested):

foolip (approved):

PR #12093, opened 2021-08-21
Add Chromium versions for HTMLMediaElement API

vinyldarkscratch:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `HTMLMediaElement` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.9).

### DIFF
--- a/api/HTMLMediaElement.json
+++ b/api/HTMLMediaElement.json
@@ -2209,7 +2209,7 @@
               "version_added": "43"
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": "43"
             },
             "edge": {
               "version_added": "12"
@@ -2236,10 +2236,10 @@
               "version_added": "3"
             },
             "samsunginternet_android": {
-              "version_added": true
+              "version_added": "4.0"
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "43"
             }
           },
           "status": {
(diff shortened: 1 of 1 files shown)


api/HTMLMediaElement.json
```diff
@@ -2209,7 +2209,7 @@
               "version_added": "43"
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": "43"
```
foolip, line 2212: Most of the data being mirrored here is incorrect, see https://github.com/mdn/browser-compat-data/issues/7843.

Can you prepare a separate PR updating only the "43" values and mirroring from the new correct data?

foolip (changes requested):

api/HTMLMediaElement.json
vinyldarkscratch: Okay, this is odd -- the collector isn't changing the value for this feature, and yet running the test in Chrome 42 shows this is clearly supported beforehand.  Yikes...

I've been noticing some odd issues with updates to Chrome Desktop's data recently.  :/

foolip: Can you file a collector issue about that? It's probably due to one of the TODOs in the updater code...

PR #12107, opened 2021-08-22
Add all browsers versions for MediaStreamTrackEvent API

queengooborg:
This PR adds real values for all browsers for the `MediaStreamTrackEvent` API, based upon commit history and date ([for Safari](https://github.com/mdn/browser-compat-data/pull/12107#issuecomment-907246933)) as well as mirroring from `api.MediaStream.addtrack_event`.


queengooborg:
It looks like there are a lot of really interesting things regarding these APIs, and it has been a rabbit hole with #12108 as well.  Hmm...

queengooborg:
Alright, so I've mirrored the data from the relevant `addtrack` event on the `MediaStream` API.  This seems most reasonable as the event's data matches the `addTrack` method's (and I can't see how an `addtrack` even would fire if you can't...well, add a track 😛 )!

foolip (approved):

PR #12108, opened 2021-08-22
Add all browsers versions for TrackEvent API

queengooborg:
This PR adds real values for all browsers for the `TrackEvent` API.  The data was copied from the earliest event that would fire this instance, specifically `api.TextTrackList.addtrack_event`.  Rresults from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v4.0.0) (specifically, https://mdn-bcd-collector.appspot.com/tests/api/TrackEvent) seem to report that this interface is supported earlier, but I believe this to be a case of accidental early exposure.


vinyldarkscratch:
This is certainly an odd discrepancy!  Sometimes these sorts of discrepancies can be worked out easily with additional updates from the collector, which I suspect may be all we'd need for this one (but unfortunately I won't be able to collect results for browser versions far back enough until my laptop's back from repairs).  Other times, it requires some deeper searching to figure out the whole story, like in the case of #12152 and #12134.  Typically when a discrepancy is found, Philip will ask me to research this further and see what the real story is behind the API and/or its feature(s).

Once I get my laptop back, I'll see what I can find out about this event!  I'll mark it as a draft PR for the time being.

queengooborg:
Hmm, getting to the bottom of this one is proving to be challenging.  Using results from the collector, support is still claimed to be introduced in these versions (with the exception of Firefox which is reporting an earlier version).  Perhaps this is a case of the event being exposed on accident?  I'll need to dig deeper into this...

queengooborg:
Okay, so I've taken another look at this, and I just can't figure out how this interface would be naturally obtained in such early versions of these browsers.  I have to believe that this is just a case of WebView (and apparently IE) exposing this interface earlier than it's used in the rest of the code.  I've gone ahead and mirrored the data from the `addtrack_event` event of the `TextTrackList` (the only one of three APIs that isn't behind a flag) to this interface, but I've left Firefox as-is because a `TrackEvent` instance _can_ be constructed in Firefox 27.

I feel that digging further into this isn't worthwhile, as we're talking about browser versions released 10+ years ago, which a very small percentage of individuals will care about.  (Additionally, this PR is one of the final three needed to get us to 100% real+ranged values!)

foolip (approved):

PR #12111, opened 2021-08-22
Add Firefox versions for RTCRtpReceiver API; correct related API features

queengooborg:
This PR adds real values for Firefox and Firefox Android for the `RTCRtpReceiver` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.9).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/RTCRtpReceiver

### DIFF
--- a/api/RTCPeerConnection.json
+++ b/api/RTCPeerConnection.json
@@ -1515,10 +1515,10 @@
               "version_added": "79"
             },
             "firefox": {
-              "version_added": "22"
+              "version_added": "34"
             },
             "firefox_android": {
-              "version_added": "44"
+              "version_added": "34"
             },
             "ie": {
               "version_added": false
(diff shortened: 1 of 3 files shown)


vinyldarkscratch:
Ah, we don't have a custom test for this API...  Will reopen with new collector results when they're ready!

queengooborg:
After some further digging around, it looks like this _was_ implemented in Firefox 34 (https://bugzilla.mozilla.org/show_bug.cgi?id=1032835), so this data is correct, and the other data is not.  I'll update this PR to update the other APIs soon!

foolip:
@queengooborg can you check if the sender APIs need some update too?

foolip (approved):
I feel like there's still a lot wrong with this data, but this is an improvement and makes things more consistent. RTCRtpSender.json already looks good.

foolip:
I've filed https://github.com/mdn/browser-compat-data/issues/12312 about a problem with the Firefox for Android data I noticed while reviewing this.

PR #12125, opened 2021-08-22
Add insertFromComposition and deleteByComposition inputTypes

ZhangYiJiang:
#### Summary

Add Input Events Level 2 `insertFromComposition` and `deleteByComposition` inputType events 

For each browser, on macOS pinyin IME input <kbd>a 1</kbd> (inserting 啊), then selecting the inserted text and input <kbd>b 1</kbd> (deleting 啊 and replacing it with 不). According to https://w3c.github.io/input-events/#event-order-during-ime-composition we should see 

- deleteByComposition beforeInput and input events for the second test since 啊 is being deleted 
- insertFromComposition beforeInput and input events for both cases 

Firefox: https://input-inspector.vercel.app/profiles/8n74DRvc3g5dhl6pMYQo - neither event fires  
Chrome: https://input-inspector.vercel.app/profiles/AXWWGKNTootDNMo6pXcY - neither event fires  
Safari: https://input-inspector.vercel.app/profiles/Hocw2xtbGkIbqgFwXzQP - insertFromComposition fires but deleteByComposition does not. I don't know when this was added, so this is just left as `true` 

Firefox - the original bug for supporting inputEvent notes both these level 2 events will not be implemented yet https://bugzilla.mozilla.org/show_bug.cgi?id=1447239

### DIFF
--- a/api/InputEvent.json
+++ b/api/InputEvent.json
@@ -322,6 +322,102 @@
             "deprecated": false
           }
         },
+        "deleteByComposition": {
+          "__compat": {
+            "description": "<code>deleteByComposition</code> input type",
+            "support": {
+              "chrome": {
+                "version_added": false
+              },
+              "chrome_android": {
+                "version_added": false
+              },
+              "edge": {
+                "version_added": false
+              },
+              "firefox": {
+                "version_added": false
+              },
+              "firefox_android": {
+                "version_added": false
+              },
+              "ie": {
+                "version_added": false
+              },
+              "opera": {
+                "version_added": false
+              },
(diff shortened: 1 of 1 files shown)


vinyldarkscratch:
Thank you for your PR, and welcome to BCD!  Currently, we are in the process of deciding whether or not to allow all-false entries in BCD -- see #10619 for more details.  I will be marking this as "not ready" accordingly.

queengooborg:
I'm going to close this now since it's been over two months and there has been no response and I'm seeing no signals that any browser supports these.  Thank you anyways!

PR #12130, opened 2021-08-23
Adding note about disabledFeatures static property

rachelandrew:
Working on https://github.com/mdn/content/issues/7745

I wasn't altogether sure where to note this. So tell me if you think it should be elsewhere and I'll re-do it.

The spec is here: https://html.spec.whatwg.org/multipage/custom-elements.html#dom-customelementregistry-define-dev (step 6 mentions it). So that's why I've added it to define.

I found the Chromium information here https://bugs.chromium.org/p/chromium/issues/detail?id=972951&q=disabledFeatures&can=1 and tested that it is in 77. I also checked Safari where it does not appear to be implemented.

### DIFF
--- a/api/CustomElementRegistry.json
+++ b/api/CustomElementRegistry.json
@@ -105,6 +105,10 @@
           "support": {
             "chrome": [
               {
+                "version_added": "77",
+                "notes": "Support for disabledFeatures static property."
+              },
+              {
                 "version_added": "67",
                 "notes": "Support for 'Customized built-in elements' as well."
               },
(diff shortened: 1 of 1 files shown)


queengooborg:
I think that we should add this as a subfeature, rather than a set of notes, especially since this change applies to so many browsers.  Could you add it as a subfeature instead, perhaps titled `supports_disabledFeatures_static_method`?

queengooborg:
I've opened a new PR to supersede this one.  Thanks for researching this!

PR #12138, opened 2021-08-23
Update Safari versions for DOMTokenList API

queengooborg:
This PR updates and corrects the real values for Safari (Desktop and iOS/iPadOS) for the `DOMTokenList` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.9).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/DOMTokenList

### DIFF
--- a/api/DOMTokenList.json
+++ b/api/DOMTokenList.json
@@ -33,7 +33,7 @@
             "version_added": "6"
           },
           "safari_ios": {
-            "version_added": "5"
+            "version_added": "6"
           },
           "samsunginternet_android": {
             "version_added": "1.0"
@@ -81,7 +81,7 @@
               "version_added": "6"
             },
             "safari_ios": {
-              "version_added": "5"
+              "version_added": "6"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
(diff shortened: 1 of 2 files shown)


lucalves (approved):

queengooborg:
Based upon additional testing, it does seem like the interface was accidentally exposed in earlier Safari versions.  I've updated the data to match the new tests and the `classList` feature!

api/DOMTokenList.json
```diff
@@ -30,7 +30,7 @@
             "version_added": "11.5"
           },
           "safari": {
-            "version_added": "5.1"
+            "version_added": "6"
           },
           "safari_ios": {
             "version_added": "5"
```
foolip, line 36: Does this also need to be updated to 6? https://developer.mozilla.org/en-US/docs/Web/API/Element/classList#browser_compatibility has a range starting at 5, but that seems implausible, that might just be wrong too.

queengooborg: It seems to be just mirroring the original desktop data.  Fixing now!

foolip (approved):

PR #12144, opened 2021-08-23
Update Safari versions for HTMLInputElement API

vinyldarkscratch:
This PR updates and corrects the real values for Safari (Desktop and iOS/iPadOS) for the `HTMLInputElement` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.9).

### DIFF
--- a/api/HTMLInputElement.json
+++ b/api/HTMLInputElement.json
@@ -693,7 +693,7 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "5.1"
+              "version_added": "5"
             },
             "safari_ios": {
               "version_added": "5"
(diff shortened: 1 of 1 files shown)


api/HTMLInputElement.json
```diff
@@ -693,7 +693,7 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "5.1"
+              "version_added": "5"
             },
             "safari_ios": {
               "version_added": "5"
```
foolip, line 699: Does this also need to change? Per https://github.com/mdn/browser-compat-data/pull/11299 the pairing should be 5/4.2, not 5/5.

foolip (approved):

PR #12148, opened 2021-08-23
Update Safari versions for TextTrackCue API

vinyldarkscratch:
This PR updates and corrects the real values for Safari (Desktop and iOS/iPadOS) for the `TextTrackCue` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.9).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/TextTrackCue

### DIFF
--- a/api/TextTrackCue.json
+++ b/api/TextTrackCue.json
@@ -78,10 +78,10 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "6.1"
+              "version_added": "6"
             },
             "safari_ios": {
-              "version_added": "8"
+              "version_added": "7"
             },
             "samsunginternet_android": {
               "version_added": "1.5"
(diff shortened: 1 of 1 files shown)


api/TextTrackCue.json
```diff
@@ -78,10 +78,10 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "6.1"
+              "version_added": "6"
             },
             "safari_ios": {
-              "version_added": "8"
+              "version_added": "7"
```
foolip, line 84: Is this based on mirroring? Wasn't a bunch of text track stuff enabled in iOS 8?

vinyldarkscratch: This was based on testing, actually.  The text track APIs were enabled in iOS 7 rather than iOS 8.

foolip: Alright then!

foolip (approved):

PR #12154, opened 2021-08-23
Update Node.js details

jasnell:
#### Summary
Updates various Node.js Web API compatibilities

### DIFF
--- a/api/Blob.json
+++ b/api/Blob.json
@@ -28,7 +28,11 @@
           },
           "nodejs": {
             "version_added": "15.7.0",
-            "notes": "Experimental implementation. Must be imported using require('buffer').Blob or import { Blob } from 'buffer'."
+            "alternative_name": "buffer.Blob",
+            "notes": [
+              "Experimental implementation.",
+              "Must be imported using require('buffer').Blob or import { Blob } from 'buffer'."
+            ]
           },
           "opera": {
             "version_added": "11"
--- a/api/Blob.json
+++ b/api/Blob.json
@@ -27,7 +27,7 @@
             "version_added": "10"
           },
           "nodejs": {
-            "version_added": "`5.7.0",
+            "version_added": "15.7.0",
             "notes": "Experimental implementation. Must be imported using require('buffer').Blob or import { Blob } from 'buffer'."
           },
           "opera": {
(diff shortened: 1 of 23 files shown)


api/BroadcastChannel.json
```diff
@@ -32,6 +32,9 @@
           "ie": {
             "version_added": false
           },
+          "nodejs": {
+            "version_added": "15.4.0",
```
lucacasonato, line 36: I suggest you add a note that this needs to be imported from "worker_threads" and that it is experimental. That would align with how api.MessageChannel was documented prior to being available as a global.

lucacasonato (commented):
Ooh nice! Didn't know Node has WHATWG streams now. Such a great win for compat. I do think it is worth adding `"notes"` that the interfaces need to imported from "streams/web" + that they are marked as experimental in Node docs.

api/Blob.json
```diff
@@ -26,6 +26,10 @@
           "ie": {
             "version_added": "10"
           },
+          "nodejs": {
+            "version_added": "15.7.0",
+            "notes": "Experimental implementation. Must be imported using require('buffer').Blob or import { Blob } from 'buffer'."
```
ddbeck, line 31: It's not true everywhere, but often structure this information with `"alternative_name"`.

Additionally, notes can be an array. I'd suggest two separate notes here, for consistency with similar cases:

- `"Experimental implementation."`
- `"Must be imported from the <code>buffer</code> module."`

```suggestion
            "version_added": "15.7.0",
            "alternative_name": "buffer.Blob",
            "notes": "Experimental implementation. Must be imported using require('buffer').Blob or import { Blob } from 'buffer'."
```

ddbeck (changes requested):
Thanks for this PR, @jasnell. Also, I see it's your first to the project—welcome!

I've got a few consistency suggestions in line comments, but no major concerns heree. Let me know if there are any questions. Thanks again!

ddbeck:
Hi @jasnell, sorry for the delay on the original review. Let me know if you plan to come back to this. If not, we can take over getting these changes merged. Thank you!

jasnell:
Yes, sorry, I do want to get this going! I missed the questions that were there.

ddbeck (approved):
Looks good. Thank you! 🎉

PR #12155, opened 2021-08-23
Bug-1630415-Make-nativeMessaging-an-optional-permission

rebloor:
Adds nativeMessaging to the list of supported optional permission as per [Bug 1630415 ](https://bugzilla.mozilla.org/show_bug.cgi?id=1630415).

npm test run with no errors


rpl (approved):
lgtm

PR #12159, opened 2021-08-23
WHATWG streams are partially not experimental

### LINKED ISSUE #11915: api.ReadableStream - Stability level
All `ReadableStream` APIs are shown as _experimental_, but at the same time the [Stream spec](https://streams.spec.whatwg.org) is a living standard.

I expect the `ReadableStream` APIs to be a stable API or, if there is a reason why these are still experimental even if the spec is standard, a sort of _stability note_ would be helpful to understand the actual situation.

I just opened the spec to see if it was still in draft or a living standard.

<details>

* Query: `api.ReadableStream`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/ReadableStream
* Report started: 2021-08-05T10:04:40.799Z

</details>

lucacasonato:
#### Summary

These interfaces are exposed in at least two browsers (usually Chromium + WebKit + Deno). Interfaces which are not yet implemented in both of WebKit and Firefox, or which are mistakenly not exposed (`ReadableStreamDefaultReader`, `ReadableStreamDefaultController`) are still marked as experimental.

Fixes #11915

### DIFF
--- a/api/ReadableStream.json
+++ b/api/ReadableStream.json
@@ -46,7 +46,7 @@
           }
         },
         "status": {
-          "experimental": true,
+          "experimental": false,
           "standard_track": true,
           "deprecated": false
         }
@@ -98,7 +98,7 @@
             }
           },
           "status": {
-            "experimental": true,
+            "experimental": false,
             "standard_track": true,
             "deprecated": false
           }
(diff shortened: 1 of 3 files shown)


ddbeck (approved):
Nice clean up. Thank you! 🎉

PR #5904, opened 2020-03-28
Implement ESLint (basic rules)

queengooborg:
This PR implements ESLint with basic rules to enforce code styling within our linters and scripts to get the ball rolling again towards #5252.  Currently, this PR will fail as there are already errors within our scripts (these will be fixed in other PRs).

PR #12161, opened 2021-08-23
ESM all the things

lucacasonato:
This PR changes all internal code (scripts/, test/, util/) to ESM. The
built package for NPM now contains an ESM and CJS entrypoint.

ESM is supported in all LTS Node versions now.


ddbeck:
This is interesting, @lucacasonato. Thanks for opening it. I've got a couple of questions for you.

I do quite like the import/export syntax over CommonJS, though I'll have to admit to some ignorance here on the advantages of ESM generally. Was there anything in particular that motivated this PR?

And on this point in particular:

> The built package for NPM now contains an ESM and CJS entrypoint.

This is interesting and reminds me of a problem reported when we started pre-building the package in the first place, https://github.com/mdn/browser-compat-data/pull/10033. One thing I've noticed in this PR is that the built package contains the data twice: once as `data.json` and once again as the default export of `index.js`. Do you know if there's a way we might avoid doubling the size of package?

ddbeck:
Oh, a side note: I haven't tested it yet, but upgrading `ora` to 6.0.0 (https://github.com/mdn/browser-compat-data/pull/12164) would probably be unblocked by merging this PR.

lucacasonato:
> I do quite like the import/export syntax over CommonJS, though I'll have to admit to some ignorance here on the advantages of ESM generally. Was there anything in particular that motivated this PR?

Yeah. The main motivation was making sure the module would be directly consumable by browsers or Deno. (`import bcd from "https://unpkg.com/@mdn/browser-compat-data/index.js"`). As long as it uses `require`, that doesn't work.

Also tooling is generally moving in the direction of ESM, so would be good to keep the project up to date in that sense. Many developers just starting out with JS will feel a lot more familiar with ESM compared to CJS.

> One thing I've noticed in this PR is that the built package contains the data twice: once as data.json and once again as the default export of index.js. Do you know if there's a way we might avoid doubling the size of package?

Indeed - this is an unfortunate side-effect caused by the lack of support for importing JSON files directly in ESM right now. This can be resolved in a couple of months once [import assertions](https://github.com/tc39/proposal-import-assertions) have stabilized across the ecosystem. The `type: json` assertion is already supported in Chrome, but not yet in Node, Deno or TSC.

Once import assertions have gained stability across the ecosystem, the ESM entrypoint can be changed to the following:

```js
import bcd from "./data.json" asserts { type: "json" };
export default bcd;
```

ddbeck:
Thanks for the additional info, @lucacasonato. That's quite helpful.

@Elchi3 do you have an opinion on whether we ought to take on that size penalty? The package used to be quite a bit larger anyway, so maybe it's small price to pay for expanding the environments that BCD works in.

ddbeck:
@foolip and @vinyldarkscratch: I think you've been most in the weeds for the internal tooling. Any quick impressions on switching to ES modules? I think this might impact a couple of other PRs y'all have open.

vinyldarkscratch:
I can't really say that I have much experiences in the benefits and downsides to using ES modules vs. CommonJS `require()`.  I've personally stuck with CommonJS for most projects since I'm more familiar with it, and from what I've read online it seems to have kept pretty consistent across all NodeJS versions.

@lucacasonato is there a particular problem you're trying to solve that's led you to want to switch to ESM for everything?

lucacasonato:
>@lucacasonato is there a particular problem you're trying to solve that's led you to want to switch to ESM for everything?

Yeah, mentioned them 3 comments up (https://github.com/mdn/browser-compat-data/pull/12161#issuecomment-904552209). Additionally @ddbeck brought up a good point about `ora` moving to ESM only, so not being able to update it anymore: https://github.com/mdn/browser-compat-data/pull/12164

> In regards to adding ESM bindings to the releases, I am definitely down for that. Can we split that out into its own PR for easier and quicker review?

Will do!

Elchi3:
> @Elchi3 do you have an opinion on whether we ought to take on that size penalty? The package used to be quite a bit larger anyway, so maybe it's small price to pay for expanding the environments that BCD works in.

I think it's fine but we could check in with the folks that were particularly interested in a smaller package size. I think it was @bershanskiy who investigated this in an earlier PR. Anton, let us know if you have thoughts :)

bershanskiy:
> I think it's fine but we could check in with the folks that were particularly interested in a smaller package size. I think it was @bershanskiy who investigated this in an earlier PR. Anton, let us know if you have thoughts :)

Thanks for notifying me of this change. I do like ESM better over CJS, but doubling of produced output size is unfortunate.

Please note that the source and the output can evolve independently. That is, the source and tests (`scripts/`, `test/`, `util/`) can fully migrate to ESM without creating any changes in the produced output. [This freedom to alter](https://github.com/mdn/browser-compat-data/pull/7398#discussion_r613956075) source code, test infrastructure, supported Node versions, and everyting else was one of the expressed goals of bundling.

> >One thing I've noticed in this PR is that the built package contains the data twice: once as data.json and once again as the default export of index.js. Do you know if there's a way we might avoid doubling the size of package?
>
> Indeed - this is an unfortunate side-effect caused by the lack of support for importing JSON files directly in ESM right now. This can be resolved in a couple of months once import assertions have stabilized across the ecosystem. The type: json assertion is already supported in Chrome, but not yet in Node, Deno or TSC.

Import assertions are an important aspect of ESM. Since import assertions should roll out across the ecosystem in a couple months, may be we should continue old style of releases for now and then migrate to ESM once they mature?

vinyldarkscratch:
> > @lucacasonato is there a particular problem you're trying to solve that's led you to want to switch to ESM for everything?
> 
> Yeah, mentioned them 3 comments up ([#12161 (comment)](https://github.com/mdn/browser-compat-data/pull/12161#issuecomment-904552209)). Additionally @ddbeck brought up a good point about `ora` moving to ESM only, so not being able to update it anymore: #12164

Ah!  Apologies, I had initially reviewed this PR on my iPhone, so I hadn't seen any of the earlier comments.  😅   I'm all for providing consumer-facing ESM bindings, so that we may maximize compatibility with ease!  For our internal code, if our dependencies are switching to becoming ESM-only, then I'll give a +1 for migrating to ESM ourselves, especially since it has no impact on our consumers.

> @foolip and @vinyldarkscratch: I think you've been most in the weeds for the internal tooling. Any quick impressions on switching to ES modules? I think this might impact a couple of other PRs y'all have open.

This might cause some merge conflicts with PRs we have open, but since it's all in regards to the imports up top, resolving the conflicts seems straightforward, so I'm not worried about this!

scripts/fix-browser-order.js
```diff
@@ -53,14 +55,15 @@ const fixBrowserOrder = filename => {
...
+const self = fileURLToPath(import.meta.url);
+if (process.argv[1] === self) {
   /**
    * @param {string[]} files
    */
   function load(...files) {
     for (let file of files) {
-      if (file.indexOf(__dirname) !== 0) {
-        file = path.resolve(__dirname, '..', file);
+      if (file.indexOf(dirname) !== 0) {
```
saschanaz, line 65: Time to migrate to `startsWith()`?

index.js
```diff
@@ -32,7 +34,7 @@ function load() {
   }
 
   for (dir of arguments) {
```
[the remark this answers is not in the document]
queengooborg: BCD doesn't actually have ESLint implemented!  I've got a PR waiting for review (see #5904), but no progress.  (I understand why, though; the BCD owners have had a lot of transitions and the priority is the data itself.)

ddbeck:
Coming back to this briefly: having read the discussion, I want to highlight @bershanskiy:

> Please note that the source and the output can evolve independently.

> Import assertions are an important aspect of ESM. Since import assertions should roll out across the ecosystem in a couple months, may be we should continue old style of releases for now and then migrate to ESM once they mature?

These are both good points. To unblock this PR, here's what I'm thinking:

- Go ahead with ESM-ing everything on the development side of the package boundary. I don't see any objections to that component, at least. It's a big chunk of the work and would unblock that dependabot PR. And, if I understand correctly, that might allow for BCD-from-source to work under Deno.
- Look at ESM-ing the stuff on the other side of the package boundary separately. Perhaps there's something we can do with pre-releases and/or npm release tags to offer the oversized ESM version of the package (e.g., offering up `v4.1.0.esm-dev` alongside the standard releases) separately from the regular package, until import assertions become more widely available?

@lucacasonato what do you think?

queengooborg:
Could I request we wait on ESM-ing BCD's internals until I've converted mdn-bcd-collector?  Since the collector project interacts with BCD on a write-basis, we need to import the development version instead, so we need to follow along with the ESM conversion, and it'll take a little longer than normal due to our testing setup.

(Note: we are a special case!  Normal BCD consumers will not have to worry about this, since they are using pre-built bundles!)

ddbeck:
@queengooborg 

> Could I request we wait on ESM-ing BCD's internals until I've converted mdn-bcd-collector?

Apart from giving you some time to do it, do you need anything else on the BCD side of things to make this easier?

queengooborg:
> Apart from giving you some time to do it, do you need anything else on the BCD side of things to make this easier?

Actually funny enough, I had finished the conversion earlier today, and it's simply pending review now!  😁

foolip:
mdn-bcd-collector is converted to ESM now, thanks to @queengooborg!

saschanaz:
Can the test changes be landed first? That should cause no breaking change.

queengooborg:
Hey @lucacasonato, just a head's up: we've been hard at work merging numerous infra PRs for BCD which may cause additional merge conflicts!

index.js
```diff
@@ -32,7 +34,7 @@ function load() {
...
-    dir = path.resolve(__dirname, dir);
+    dir = path.resolve(dirname, dir);
```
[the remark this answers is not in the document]
caugner: Indeed, given that we use `new URL('.', import.meta.url)` above, the proposed change makes sense.

schemas/browsers-schema.md
```diff
@@ -80,7 +80,7 @@ The release objects consist of the following properties:
 This structure is exported for consumers of `@mdn/browser-compat-data`:
 
 ```js
-> const compat = require('@mdn/browser-compat-data');
-> compat.browsers.firefox.releases['1.5'].status;
+> import bcd from '@mdn/browser-compat-data';
+> bcd.browsers.firefox.releases['1.5'].status;
```
caugner, line 84: This looks like Node.js REPL, which doesn't seem to support import statements yet:

```console
% node
Welcome to Node.js v16.15.0.
Type ".help" for more information.
> import bcd from '@mdn/browser-compat-data';
import bcd from '@mdn/browser-compat-data';
^^^^^^

Uncaught:
SyntaxError: Cannot use import statement inside the Node.js REPL, alternatively use dynamic import
```

Maybe it would make sense to describe both CommonJS and Module ways of accessing it?

index.js
```diff
@@ -1,6 +1,8 @@
-'use strict';
-const fs = require('fs');
-const path = require('path');
+import fs from 'node:fs';
+import path from 'node:path';
+import { fileURLToPath } from 'node:url';
```
caugner, line 3: +1 for using [`node:`-prefixed imports](https://2ality.com/2021/12/node-protocol-imports.html)

scripts/fix-browser-order.js
caugner: Let's limit this PR to ESM changes.

scripts/compare-features.test.js
```diff
@@ -3,9 +3,10 @@
  * http://creativecommons.org/publicdomain/zero/1.0/ */
 
 'use strict';
```
caugner, line 5: Remove `'use strict';` from all ES modules, as they're implicitly strict (unless I'm mistaken).

```suggestion
```

caugner (changes requested):
LGTM, just minor nits.

package.json
```diff
@@ -4,6 +4,7 @@
   "description": "Browser compatibility data provided by MDN Web Docs",
   "main": "index.js",
   "types": "index.d.ts",
+  "type": "module",
```
lucacasonato, line 7: This may be a problem without landing #15775 first.

lucacasonato (commented):
This looks good to me now.

index.js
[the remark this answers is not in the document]
lucacasonato: Can't do this, because the Node LTS release that is being used by BCD does not support URLs as an arg to fs functions yet.

package.json
queengooborg: This will be alright since we generate a stripped-down package.json using only select properties for the releases!

queengooborg (changes requested):
LGTM overall, thank you!  Just a small nit regarding the readme:

index.js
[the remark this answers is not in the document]
caugner: @lucacasonato What's the difference to `new URL('.', import.meta.url)` in line 5?

README.md
```diff
@@ -25,7 +25,7 @@ npm install @mdn/browser-compat-data
 ## Usage
 
 ```js
-const bcd = require('@mdn/browser-compat-data');
+import bcd from '@mdn/browser-compat-data';
```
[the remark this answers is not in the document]
caugner: Just to clarify, this change would have to wait until the first ESM-compatible version of BCD is released, right?

PS: I guess that will be a major version bump?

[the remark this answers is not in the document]
queengooborg: That's correct, see #15775 for the ESM-ready public builds!

queengooborg:
Oh, before we merge this PR by the way, I'd like to land #15944 first.  In my testing, I found that Ajv's "strict mode" warnings get repeated for every run of the schema validation in ESM, so errors get buried underneath the warnings.  Why this only occurs on ESM, I'm not sure.

foolip:
@queengooborg there are lots of conflicts again. Would this be ready to merge after another rebase? I haven't reviewed it at all, it's so big :)

PR #12169, opened 2021-08-24
Package data for NPM as ESM and CJS

lucacasonato:
Previously only a CJS entrypoint was provided. This provides a ESM
entrypoint that can be directly used from Deno or a browser.


scripts/release-build.js
```diff
@@ -38,7 +43,11 @@ async function copyFiles() {
 
 function createManifest() {
   const full = require('../package.json');
-  const minimal = { main: 'index.js' };
+  const minimal = {
+    main: 'index.cjs',
+    exports: { import: './index.js', require: './index.cjs' },
```
vinyldarkscratch, line 48: Just a note for @ddbeck -- with the inclusion of an `exports` in our `package.json`, we'll need to bump the major semver.  An [article](https://redfin.engineering/node-modules-at-war-why-commonjs-and-es-modules-cant-get-along-9617135eeca1#1152) I have been reading states the following:

> Beware: adding an exports map is always a “semver major” breaking change. By default, your users can reach into your package and require() any script they want, even files that you intended to be internal. The exports map ensures that users can only require/import the entry points that you deliberately expose.

While we don't really expose any other files besides `data.json` and `index.js`, it is a potentially breaking change.  But despite that, it feels fitting to bump the major version regardless -- ESM support is a big deal after all!

index.d.ts
```diff
@@ -1,9 +1,6 @@
 /* Any copyright is dedicated to the Public Domain.
  * http://creativecommons.org/publicdomain/zero/1.0/ */
-/// <reference path="./types.d.ts"/>
+
 import { CompatData } from './types';
 
-// This is necessary to have intellisense in projects which
-// import data from this package.
-declare const compatData: CompatData;
-export = compatData;
```
vinyldarkscratch, line 9: Does replacing this with `export default CompatData` fix the Intellisense issues?  (I don't really use IntelliJ so I am unsure!)

scripts/release-build.js
```diff
@@ -15,18 +15,23 @@ function createDataBundle() {
...
 
-async function writeIndex() {
-  const dest = path.resolve(directory, 'index.js');
+async function writeIndexCJS() {
+  const dest = path.resolve(directory, 'index.cjs');
   const content = `module.exports = require("./data.json");\n`;
   await fs.writeFile(dest, content);
 }
 
+async function writeIndexJS(data) {
```
vinyldarkscratch, line 29: For clarity, can we call this function `writeIndexMJS` instead?

vinyldarkscratch (commented):
Thank you for splitting this out, I'm super excited for this!  😄   I have just a small suggestion and a question, but from a quick skim of the code this is looking good to me overall!

scripts/release-build.js
```diff
@@ -15,18 +15,23 @@ function createDataBundle() {
...
   const dest = path.resolve(directory, 'data.json');
-  const data = createDataBundle();
   await fs.writeFile(dest, data);
 }
 
-async function writeIndex() {
-  const dest = path.resolve(directory, 'index.js');
+async function writeIndexCJS() {
+  const dest = path.resolve(directory, 'index.cjs');
   const content = `module.exports = require("./data.json");\n`;
```
saschanaz, line 25: This function was to keep compatibility with ESM where JSON couldn't be directly imported (#10033). Seems like it's not needed here anymore.

lucacasonato: Makes sense. I changed package.json to the following:
```json
{
    "main": "data.json",
    "exports": { "import": "./index.js", "require": "./data.json" },
}
```

queengooborg:
Okay, so I had a feeling this might be a breaking change, and it looks like my suspicions were correct.  BCD's exported packages don't expose the version number, so I require the included `package.json` and use that instead.  By adding the `exports` key to the `package.json`, I wouldn't be able to require any file I wanted from the package, so my program would break.  We should be sure to bump the major semver when this is launched!

lucacasonato:
Import assertions are now widely available in server side runtimes. Chrome supports them, and so do both Deno (since 1.17) and Node (since v17). As such we could now keep the JSON in a seperate file for publishing again, and then re-export it from the ESM module using import assertions.

@queengooborg We could do the same with the version from package.json.

lucacasonato:
Actually - scrap that. Node 17 does not support JSON modules without the `--experimental-json-modules` flag... :/

I guess the current thing here is the best thing we could do for now.

lucacasonato:
Node unflagged import assertions in v17.5.0 (released 2022-02-10). Probably too soon to rely on them?

queengooborg:
Yeah, I'd say it's too soon, especially since our developer requirement is NodeJS v14.  Thanks for double-checking!

lucacasonato:
Makes sense. Once the requirement is NodeJS v18, this is the patch to apply to clean up:

```patchFrom c1922a92bd2597120b81d2008028098c965b00f0 Mon Sep 17 00:00:00 2001
From: Luca Casonato <hello@lcas.dev>
Date: Thu, 7 Apr 2022 23:29:06 +0200
Subject: [PATCH] Use import assertions to import data in ESM

---
 scripts/release-build.js | 6 +++---
 1 file changed, 3 insertions(+), 3 deletions(-)

diff --git a/scripts/release-build.js b/scripts/release-build.js
index eaba25e7e..fd2a20a9c 100644
--- a/scripts/release-build.js
+++ b/scripts/release-build.js
@@ -20,9 +20,9 @@ async function writeData(data) {
 }
 
-async function writeIndexESM(data) {
+async function writeIndexESM() {
   const dest = path.resolve(directory, 'index.js');
-  const content = `export default ${data};\n`;
+  const content = `import data from "./data.json" assert { type: "json" };\nexport default data;\n`;
   await fs.writeFile(dest, content);
 }
 
@@ -91,7 +91,7 @@ async function main() {
 
   await writeManifest();
   await writeData(data);
-  await writeIndexESM(data);
+  await writeIndexESM();
   await copyFiles();
 
   console.log('Data bundle is ready');
-- 
2.35.1

```

queengooborg:
Just a quick request, would you mind also updating the usage section in the readme to include an example via import, like so?

```
// Import BCD into your project
import bcd from '@mdn/browser-compat-data';
// ...or...
const bcd = require('@mdn/browser-compat-data');

// Grab the desired support statement
const support = bcd.css.properties.background; // returns a compat data object (see schema)
```

queengooborg (approved):
I went ahead and updated the readme to include an ESM import example, but all in all this is looking great to me!  Thank you so much for this, I'm super excited to see this in the next release!

PR #12171, opened 2021-08-25
Fix crypto.randomUUID() spec URL

sideshowbarker:
The `crypto.randomUUID()` method isn’t in the Web Crypto spec but instead in a separate spec.

The spec URL chosen here is intentionally the anchor for the closest heading rather than the specific anchor for the definition of method.

The rationale for that choice is that this is a case where landing developers at a section heading gives them a lot more context — because the section has a note explaining the method in developer-friendly language, and if we instead landed developers at the formal normative definition, the only way they’d see the more developer-friendly explanation would be if they knew to scroll back up vertically to find it.

cc @lucacasonato

### DIFF
--- a/api/Crypto.json
+++ b/api/Crypto.json
@@ -122,7 +122,7 @@
       "randomUUID": {
         "__compat": {
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Crypto/randomUUID",
-          "spec_url": "https://w3c.github.io/webcrypto/#Crypto-method-randomUUID",
+          "spec_url": "https://wicg.github.io/uuid/#extensions-to-the-crypto-interface",
           "description": "<code>randomUUID()</code>",
           "support": {
             "chrome": {


Shago2739:
new desing to fix problem whit criptocurrency now

Shago2739:
@Shago2739  hello world nextsteps on the house

Elchi3 (approved):
Thanks Mike!

(We really need to write up guidelines for `spec_url `submissions!)

sideshowbarker:
> (We really need to write up guidelines for `spec_url `submissions!)

Agreed — I’ll try to carve out some time this week to do that

lucacasonato:
Thanks! My overzealous copy paste at work again 😅

sideshowbarker:
> Thanks! My overzealous copy paste at work again 😅

No worries — thanks for having gotten the feature into BCD. There’s almost always some kind of kind up that eventually needs to be done some time after something first gets landed.

PR #12181, opened 2021-08-25
HTMLSlotElement is not experimental

rachelandrew:
Part of https://github.com/mdn/content/issues/7756

The `assign()` method is in Firefox 92. I also changed the API to not be experimental as there seems to be wide implementation.


ddbeck (approved):
Looks good. Thank you! 🎉

PR #12182, opened 2021-08-25
Reorganize scripts within scripts/ folder

queengooborg:
This PR moves the fix and release scripts into their own `fix/` and `release/` subdirectories within `scripts/` respectively.


caugner (approved):
LGTM, exactly what I was hoping to see.

PR #12185, opened 2021-08-26
Remove KeyboardEvent.which and MouseEvent.which

foolip:
UIEvent.which is already in the data, so update that to reflect the
partial support from when `which` was on KeyboardEvent and MouseEvent.

The existing data from MouseEvent is not perfectly preserved, in
particular the early Opera and Safari versions have not been verified,
so instead ranges are used in the merged data.

### DIFF
--- a/api/KeyboardEvent.json
+++ b/api/KeyboardEvent.json
@@ -1706,57 +1706,6 @@
             "deprecated": false
           }
         }
-      },
-      "which": {
-        "__compat": {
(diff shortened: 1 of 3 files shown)


queengooborg (approved):
Got the content PR all done for this, and it has now been merged -- the BCD updates here are LGTM!  👍

foolip:
Thanks @queengooborg!

PR #12194, opened 2021-08-27
FF supports navigator.share behind pref

hamishwillee:
`Navigator.share()` was marked false for FireFox desktop. However testing shows that it does work for desktop Windows (at least). I think it was added in FF71 in this [bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1312422) behind the preference `dom.webshare.enabled`.
Note that FF android is not blocked by this preference.

I also did some testing and verified that `Navigator.share()` and `Navigator.canShare()` work on Windows Opera and Edge, and also Opera Android. I have *guessed* the versions for desktop match Chrome (but without the partial implementation warning due, I think, to MacOS issues). 

Upshot, those changes might perhaps revert to `true` if you think this is not a safe assumption.

FYI these methods are protected by web-share permission. I have not added subfeatures for that, because I don't know if it is a compatibility issue (at all).

### DIFF
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -3355,8 +3355,7 @@
             },
             "firefox_android": {
               "version_added": "79",
-              "partial_implementation": true,
-              "notes": "Firefox for Android does not support sharing text or files."
+              "partial_implementation": true
             },
             "ie": {
               "version_added": false


hamishwillee:
@ddbeck Further to this, the [webshare spec](https://w3c.github.io/web-share) states that `navigator.share()` should support sharing a base set of data types that might be shared, which includes: title, text, url, an array files

I just found this bug https://bugzilla.mozilla.org/show_bug.cgi?id=1635700 which indicates that FF does not support file sharing. 

Should we note this by marking the support in firefox as partial OR by adding a subfeature for `files`, `url`, whatever. Thoughts? 

@saschanaz Do you know when/if this feature is expected for Firefox?

api/Navigator.json
```diff
@@ -554,10 +554,10 @@
               "version_added": false
             },
             "opera": {
-              "version_added": false
+              "version_added": "75"
             },
             "opera_android": {
-              "version_added": false
+              "version_added": true
```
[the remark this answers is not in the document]
queengooborg: I think it's safe to say "54" for this one -- Opera Android 54 correlates with Chrome Android 76 (there was no Opera Android release matching Chrome 75).

ddbeck:
> but without the partial implementation warning due, I think, to MacOS issues

Sorry, I didn't quite follow this. You didn't mark the Mac partial implementation because you didn't test it? Or had reason to believe that this is a Chrome-specific limitation? The rest of this seems fine to me, I just want to figure out when/where we ought to be copying notes.

api/Navigator.json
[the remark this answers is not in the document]
hamishwillee: Thanks!

hamishwillee:
> but without the partial implementation warning due, I think, to MacOS issues

Coming back to this after a couple of weeks - yes, that note is gibberish sorry. 

What I am saying is that Chrome marks `canShare()` support as partial - should I copy this "partial" to the other chromiums?

The only reason I am questioning this is because I don't know the reason this was marked as partial support in the first place.  If this could be because it is not supported on macOS then I could omit this caveat on android platforms. 

I guess I should add "partial" and the note for Opera though, as that runs on Mac.

Does the question make sense now?

ddbeck:
> Does the question make sense now?

Yes, this makes sense now. Reading the linked bug, it appears to be Mac specific, so I'd copy it to Edge and Opera. 👍

hamishwillee:
Thanks @ddbeck ,
I've added the note and partial_implementation to opera and Edge. Should be good now.

saschanaz:
> I just found this bug https://bugzilla.mozilla.org/show_bug.cgi?id=1635700 which indicates that FF does not support file sharing.
> 
> Should we note this by marking the support in firefox as partial OR by adding a subfeature for `files`, `url`, whatever. Thoughts?
> 
> @saschanaz Do you know when/if this feature is expected for Firefox?

It's not expected anytime soon, and `files` support is [added later to the spec separately](https://github.com/w3c/web-share/pull/133), so 👍 to add a separate `files` data.

saschanaz:
And also probably a good chance to use `preview` since it's enabled only on Nightly/early beta.

hamishwillee:
@ddbeck  @saschanaz OK, I added a files subfeature with versions based on 
-  Safari/Safari iOS in v15 - https://blog.bitsrc.io/sharing-files-from-ios-15-safari-to-apps-using-web-share-c0e98f6a4971 [Note, that files may have been in Safari v14 behind a flag, but I couldn't find the information and not sure it is necessary?]
- Chrome  https://www.chromestatus.com/feature/4777349178458112 - v76 android, 89 Chromium. Rolled it out everwhere.

Question. Could/should we add subfeatures for text as well? The reason I ask is that Firefox for android does not support text sharing (at the moment captured with a note).

> And also probably a good chance to use preview since it's enabled only on Nightly/early beta.

@ddbeck How do you do both a preference and a preview version? This is what it looks like now
```
            "firefox": {
              "version_added": "71",
              "flags": [
                {
                  "type": "preference",
                  "name": "#dom.webshare.enabled",
                  "value_to_set": "enabled"
                }
              ]
            },
```
This was added in nightly in FF92 - see https://bugzilla.mozilla.org/show_bug.cgi?id=1641280
I tried a few variants like this, but it can't cope with "preview". Ideas?

```json
            "firefox": [
              {
                "version_added": "preview"
              },
              {
                "version_added": "71",
                "version_removed": "91",
                "flags": [
                  {
                    "type": "preference",
                    "name": "#dom.webshare.enabled",
                    "value_to_set": "enabled"
                  }
                ]
              }
            ],
```

api/Navigator.json
```diff
@@ -3259,10 +3271,61 @@
             }
           },
           "status": {
-            "experimental": true,
+            "experimental": false,
             "standard_track": true,
             "deprecated": false
           }
+        },
+        "files": {
```
ddbeck, line 3279: Can we give this a richer name and description? For the name, perhaps named `data_parameter_files` and something like `"Files as a <code>data</code> paramter"`? Riffing on [the parameter naming conventions](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#parameters-and-parameter-object-features), mainly.

ddbeck (changes requested):
Thanks, Hamish. Some general commentary here and a specific suggestion in a line comment.

> Could/should we add subfeatures for text as well? The reason I ask is that Firefox for android does not support text sharing (at the moment captured with a note).

Yes, I endorse this. 👍

> This was added in nightly in FF92 - see https://bugzilla.mozilla.org/show_bug.cgi?id=1641280 . I think it was only added for Windows though - dunno if that matters (i.e. it doesn't if preview is assumed to be Windows for Firefox).
I tried a few variants like this, but it can't cope with "preview". Ideas?

There's a lot to unpack here:

- [Operating system limitations imply `partial_implementation`](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#operating-system-limitations-imply-partial_implementation). `"preview"` shouldn't change this.
- You said 'it can't cope with "preview"'. Sorry if I'm being dense, but can you elaborate on this? What's "it" and what is not coping, in this situation?
- Based on the example, I'd interpret that to mean:
  - In Nightly-only, it's enabled by default.
  - In 71-91, it's available behind a pref.
  - In 92+ (stable), it's no longer supported in anyway.
  
  Is that the story you want to tell? If it isn't, what is?

api/Navigator.json
saschanaz: That sounds like `.share(files)` but this is more like `.share({ title, text, files })`;

ddbeck: ohhh, I misread the docs. I saw "Possible values" and thought it was one argument.

The correction here would be to closely follow the data guideline: `data_files_parameter` with description  `"<code>data.filescode> parameter"`.

hamishwillee:
Thanks, all helpful. I'll make the required changes tomorrow.

@ddbeck 
1. By "can't cope with preview", I mean I tried the following and got an error related to using the term "preview". I tried a number of other variants, and they gave me an error warning too. So my question was "what does it have to look like". I will try this again tomorrow and post the actual error for this to provide more data. 
```
            "firefox": [
              {
                "version_added": "preview"
              },
              {
                "version_added": "71",
                "version_removed": "91",
                "flags": [
                  {
                    "type": "preference",
                    "name": "#dom.webshare.enabled",
                    "value_to_set": "enabled"
                  }
                ]
              }
            ],
```
2. With respect to the story ...
3. 
The story that I "know" is that prior to FF92 it is available behind pref ( 71-91), in FF92 it is available in Nightly by default and behind the pref for release. 

The restriction to Windows confuses me. My original understanding was that the preference change in Nightly was only for Windows, but that doesn't make much sense. 

@saschanaz is this a partial implementation? - ie this only works on Windows, or is it that we only enabled it for Windows, or is it that it behaves differently on Windows? Can you confirm?

saschanaz:
> @saschanaz is this a partial implementation? - ie this only works on Windows, or is it that we only enabled it for Windows, or is it that it behaves differently on Windows? Can you confirm?

It's enabled on Windows and Android, and should generally work same way on both, except a nit: Windows does not support tracking share action properly, so it resolves after opening the share popup, not after completing the share action.

hamishwillee:
> except a nit: Windows does not support tracking share action properly, so it resolves after opening the share popup, not after completing the share action. (Same for Chrome)

@saschanaz Ah, that wasn't clear to me, and makes this wrong: https://developer.mozilla.org/en-US/docs/Web/API/Navigator/share
image

My understanding was that is that this should be re-written as:

> **On Windows** the method resolves a Promise with `undefined` as soon as the **share popup has launched.** **On other platforms it resolves the promise with `undefined` once data is successfully passed to the share target**.

Is that correct? Note the assertions:
- in both cases it resolves with undefined.
- "data is passed to the share target" (i.e. bluetooth) NOT actually sent. So completion means "prompt has closed and data sent to the OS API that does the sending"

hamishwillee:
@ddbeck I have added `data_parameter_files` and `data_parameter_text` to `share` and `canShare`. These mirror each other. The Chrome mac bug refers to files and it isn't clear if it also applies to text - but it might so I have left that in. 

With respect to adding preview, I still don't get how to do that. The story I am trying to tell is that it was added in version 71 behind a preference. From version 92 it is released (no pref required) in Nightly. 

I tried the following, but get `Error: Invalid argument not valid semver ('preview' received)`

```
            "firefox": [
              {
                "version_added": "preview"
              },
              {
                "version_added": "71",
                "flags": [
                  {
                    "type": "preference",
                    "name": "#dom.webshare.enabled",
                    "value_to_set": "enabled"
                  }
                ]
              }
            ],
```

saschanaz:
>resolved when "data is passed to the share target" (i.e. bluetooth) NOT when actually sent/recieved.

I don't think this can be a strong assertion as each OS may behave different. But correct at least on Android.

hamishwillee:
@ddbeck When you get a chance, this needs re-review. As discussed, this is not an immediate priority, just keeping it visible.

ddbeck (changes requested):
Thanks, Hamish. Some suggestions in line comments.

Regarding `Error: Invalid argument not valid semver ('preview' received)`, that's a bug. I'll open an issue about it.

hamishwillee:
@ddbeck Thanks for the review. I've "accepted all".

Issue #11383, opened 2021-07-04
css.properties.padding-block/margin-block/margin-inline - Supported in Edge


saschanaz:
#### What information was incorrect, unhelpful, or incomplete?

They are supported in Edge

#10751 only fixed padding-inline and the remainings are still to be fixed.

#### What did you expect to see?

Proper version numbers

#### Did you test this? If so, how?

The MDN examples work

<details>
<summary>MDN page report details</summary>

* Query: `css.properties.padding-block`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/CSS/padding-block
* Report started: 2021-07-04T15:34:27.990Z

</details>

saschanaz:
It could also be worth to check opera for Android support.

foolip:
I can confirm with these test on Edge 91:
https://mdn-bcd-collector.appspot.com/tests/css/properties/padding-block
https://mdn-bcd-collector.appspot.com/tests/css/properties/margin-block
https://mdn-bcd-collector.appspot.com/tests/css/properties/margin-inline

I haven't tested Opera for Android, but think we can safely mirror all Chromium-based browsers here.

foolip:
@vinyldarkscratch can you fix this? Probably mdn-bcd-collector can just fill in the right versions.

PR #12202, opened 2021-08-27
Mirror Chrome to Edge for CSS features shipped in Chrome 87

foolip:
Fixes https://github.com/mdn/browser-compat-data/issues/11383.


vinyldarkscratch (approved):
Thanks for tackling these, looks good to me!  👍

PR #12249, opened 2021-09-03
Firefox 93 supports datetime-local

### LINKED ISSUE #12248: html.elements.input.datetime-local - Firefox 93 support
Firefox 93 supports datetime-local.

I've tested this with the FF nightly build on https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local

See also https://bugzilla.mozilla.org/show_bug.cgi?id=1283388

lafeber:
#### Summary
Firefox 93 supports datetime-local, updates version to reflect this.

Downloaded nightly build, checked if it worked on https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input/datetime-local

See https://bugzilla.mozilla.org/show_bug.cgi?id=1283388

#### Related issues
Fixes #12248

### DIFF
--- a/html/elements/input/datetime-local.json
+++ b/html/elements/input/datetime-local.json
@@ -18,8 +18,7 @@
                 "version_added": "12"
               },
               "firefox": {
-                "version_added": false,
-                "notes": "See <a href='https://bugzil.la/888320'>bug 888320</a> and <a href='https://wiki.mozilla.org/TPE_DOM/Date_time_input_types'>TPE DOM/Date time input types</a>."
+                "version_added": "93"
               },
               "firefox_android": {
                 "version_added": true


html/elements/input/datetime-local.json
```diff
@@ -18,8 +18,7 @@
                 "version_added": "12"
               },
               "firefox": {
-                "version_added": false,
-                "notes": "See <a href='https://bugzil.la/888320'>bug 888320</a> and <a href='https://wiki.mozilla.org/TPE_DOM/Date_time_input_types'>TPE DOM/Date time input types</a>."
+                "version_added": "93"
               },
               "firefox_android": {
                 "version_added": true
```
ddbeck, line 24: This appears to have shipped on Android as well

```suggestion
                "version_added": "93"
```

ddbeck (changes requested):
Thanks for this PR, @lafeber. I see it's your first to the project—welcome! I have one suggestion to expand on your change, in a line comment. Thanks again!

html/elements/input/datetime-local.json
lukewarlow: This exists in Firefox Android already. Potentially better to make a separate issue and find the actual version it was introduced? Its existed since at least 91 though i suspect potentially much earlier.

lukewarlow: Just installed and tested Firefox 55 for Android and it supports this input. Can't test much older than that because it crashes on launch on my phone. But it's possible we can find the version it was added in release notes somewhere?

lukewarlow: I've also just noticed it is already set to true so potentially best to just leave it at that because getting exact versions doesn't materially help anyone as support has existed for so long.

lafeber:
@ddbeck Thank you for the suggestion, it's implemented.

html/elements/input/datetime-local.json
```diff
@@ -18,11 +18,10 @@
                 "version_added": "12"
               },
               "firefox": {
-                "version_added": false,
-                "notes": "See <a href='https://bugzil.la/888320'>bug 888320</a> and <a href='https://wiki.mozilla.org/TPE_DOM/Date_time_input_types'>TPE DOM/Date time input types</a>."
+                "version_added": "93"
               },
               "firefox_android": {
-                "version_added": true
+                "version_added": "93"
```
lukewarlow, line 24: Just so it doesn't get lost with the other discussion. This has existed in Firefox Android forever (for all modern intents and purposes). `true` was a better representation of the situation that support for this browser isn't something to worry about.

lafeber: You're absolutely right, I've changed it back. FF Android is currently on version 90, so changing this to 93 would not make sense.

html/elements/input/datetime-local.json
ddbeck: Good catch, @lukewarlow! Sorry for the bad suggestion, @lafeber.

ddbeck (approved):

PR #12258, opened 2021-09-04
Added Samsung Internet 16.0 Beta

lukewarlow:
#### Summary

Added Samsung Internet 16.0 as a beta release and marked 15.0 as stable.

Also copied across blink 92 support data to Samsung Internet 16.0

image

https://play.google.com/store/apps/details?id=com.sec.android.app.sbrowser.beta

### DIFF
--- a/api/Crypto.json
+++ b/api/Crypto.json
@@ -162,7 +162,7 @@
               "version_added": false
             },
             "samsunginternet_android": {
-              "version_added": false
+              "version_added": "16.0"
             },
             "webview_android": {
               "version_added": "92"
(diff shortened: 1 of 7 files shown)


lukewarlow:
The release date for Samsung Internet 15.0 and potentially 14.2 (and others) seems to be the release data of the beta versions. I've followed suit and done the same for 16.0 but this is not the way release dates are done for other browsers. Should I make an issue to look into this?

vinyldarkscratch:
That's a good catch, @lukewarlow!  We should be using the dates when they no longer were in beta.  Let's omit 16.0's release date and open either an issue or a PR to correct the dates for 14.2/15.0?

vinyldarkscratch (approved):

PR #12268, opened 2021-09-05
Update data for API globals

queengooborg:
This PR updates the data for the API globals, specifically the worker support data.  Some of these features needed to be changed for consistency as well.  All data comes from the mdn-bcd-collector project v3.1.12.

### DIFF
--- a/api/_globals/atob.json
+++ b/api/_globals/atob.json
@@ -92,7 +92,7 @@
               "version_added": "17"
             },
             "opera_android": {
-              "version_added": "17"
+              "version_added": "18"
             },
             "safari": {
               "version_added": "10"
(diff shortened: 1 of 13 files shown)


foolip:
Can you describe how you mapped results for this? The collector hasn't been updated to deal with the globals reorganization yet, so I presume it was partly manual.

queengooborg:
Right, I probably should've specified that!  To map the mdn-bcd-collector results to these global properties, I ran the `add-new-bcd` script to get the corresponding results for both `Window` and `WorkerGlobalScope`.  Using that, I then mapped the results on the `Window` API to the main feature, and `WorkerGlobalScope` to `worker_support`.

foolip:
Thanks @queengooborg, that's what I thought.

foolip (approved):

PR #12269, opened 2021-09-05
Add Firefox versions for api.Document.documentURI.readonly

queengooborg:
This PR adds real values for Firefox and Firefox Android for the `documentURI.readonly` member of the `Document` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.12).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Document/documentURI.readonly

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -3395,10 +3395,10 @@
                 "version_added": "≤79"
               },
               "firefox": {
-                "version_added": true
+                "version_added": "10"
               },
               "firefox_android": {
-                "version_added": true
+                "version_added": "10"
               },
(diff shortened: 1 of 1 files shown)


foolip (approved):

foolip:
I've sent https://github.com/mdn/browser-compat-data/pull/12276 to remove this feature instead. In general, for lots of behavioral subfeatures like this, I would suggest checking to see if it really represents some non-standard behavior that is irrelevant today. There seems to be a lot like this.

PR #12271, opened 2021-09-05
Correct Samsung Internet support for `appearance`

lukewarlow:
#### Summary

Corrected the Samsung Internet support data for the `appearance` property.

Just copied across blink support data.

### DIFF
--- a/css/properties/appearance.json
+++ b/css/properties/appearance.json
@@ -99,11 +99,16 @@
               "partial_implementation": true,
               "prefix": "-webkit-"
             },
-            "samsunginternet_android": {
-              "version_added": "1.0",
-              "partial_implementation": true,
-              "prefix": "-webkit-"
-            },
+            "samsunginternet_android": [
+              {
+                "version_added": "14.0"
+              },
+              {
+                "version_added": "1.0",
+                "partial_implementation": true,
+                "prefix": "-webkit-"
+              }
+            ],
             "webview_android": [
               {
                 "version_added": "84"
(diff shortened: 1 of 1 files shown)


css/properties/appearance.json
```diff
@@ -347,10 +352,15 @@
                 "version_added": "1",
                 "partial_implementation": true
               },
-              "samsunginternet_android": {
-                "version_added": "1.0",
-                "partial_implementation": true
-              },
+              "samsunginternet_android": [
+                {
+                  "version_added": "13.2"
```
queengooborg, line 357:
```suggestion
                  "version_added": "13.0"
```

css/properties/appearance.json
```diff
@@ -506,10 +516,15 @@
...
```
queengooborg, line 521:
```suggestion
                  "version_added": "13.0"
```

queengooborg (changes requested):
Good catch, thank you!  I'm assuming the version numbers were based upon mirroring, could we change the version numbers from "13.2" to "13.0" (which matches Chrome 83)?  If you've done testing or have found documentation that indicates it was actually added in 13.2, feel free to disregard my change request!

css/properties/appearance.json
lukewarlow: Good catch I just assumed 13.0 was on an Earlier chromium version. Corrected that now

queengooborg (approved):
Thank you for the quick fix, this is looking good to me!  Aside from 1.5, a Chrome engine bump has happened in every major Samsung Internet release, though some features were introduced in the minor releases.

PR #12272, opened 2021-09-05
Correct Firefox Android support for individual transform properties

lukewarlow:
#### Summary

Adds Firefox Android support to `translate`, `scale` and `rotate` individual CSS properties.

Copies Firefox support data to Firefox Android.

Verified to exist as of Firefox Android 92. So assuming that it's been supported since 79.

### DIFF
--- a/css/properties/rotate.json
+++ b/css/properties/rotate.json
@@ -31,16 +31,22 @@
                 ]
               }
             ],
-            "firefox_android": {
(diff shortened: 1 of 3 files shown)


queengooborg (approved):
Good catch, thanks for fixing this!  This looks good to me considering the Firefox Desktop data.

PR #12275, opened 2021-09-06
Add Safari versions for api.Element.animate/getAnimations

queengooborg:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `animate` and `getAnimations` members of the `Element` API.  It turns out the "CSS Animations via Web Animations" flag had no effect upon these properties.  The data was obtained by going through SauceLabs, checking each individual version, turning on the flag (if it was available), and seeing what the first version that supported the property was.

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -695,9 +695,20 @@
                 ]
               }
             ],
-            "safari_ios": {
-              "version_added": "13.4"
-            },
+            "safari_ios": [
+              {
+                "version_added": "13.4"
+              },
+              {
+                "version_added": "11.3",
+                "flags": [
+                  {
+                    "type": "preference",
+                    "name": "Web Animations"
+                  }
+                ]
+              }
+            ],
             "samsunginternet_android": {
               "version_added": "3.0"
             },
--- a/api/Element.json
+++ b/api/Element.json
@@ -3463,28 +3463,42 @@
                 ]
               }
             ],
-            "safari": {
-              "version_added": "11.1",
-              "partial_implementation": true,
-              "notes": "Does not support the <code>subtree</code> option.",
-              "flags": [
-                {
-                  "type": "preference",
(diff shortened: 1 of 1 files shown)


api/Element.json
```diff
@@ -3457,32 +3453,24 @@
...
-                },
-                {
-                  "type": "preference",
-                  "name": "CSS Animations via Web Animations"
                 }
               ]
             },
             "safari_ios": {
-              "version_added": true,
+              "version_added": "11.3",
```
foolip, line 3467: This data suggests that `element.getAnimations()` isn't enabled by default in Safari, but it is. Can you add another statement for that too? I've checked all "Web Animations" and this seems to be the only such mistake.

foolip (changes requested):

api/Element.json
```diff
@@ -3452,28 +3463,42 @@
...
-                  "type": "preference",
-                  "name": "Web Animations"
-                }
-              ]
-            },
+            "safari": [
+              {
+                "version_added": "13.1",
+                "partial_implementation": true,
+                "notes": "Does not support the <code>subtree</code> option."
```
foolip, line 3470: Confirmed: https://github.com/WebKit/WebKit/blob/27bbb8b26e7c86cef190b68a12127e492187e2ca/Source/WebCore/animation/DocumentOrShadowRoot%2BWebAnimations.idl#L28

foolip (approved):

api/Element.json
foolip: Oops, that was the wrong bit of IDL. It is actually supported:
https://github.com/WebKit/WebKit/blob/27bbb8b26e7c86cef190b68a12127e492187e2ca/Source/WebCore/animation/Animatable.idl#L29

I'll send a PR to fix.

PR #12276, opened 2021-09-06
Remove irrelevant documentURI.readonly feature

foolip:
The original behavior, being able to set document.documentURI, is a
superset of the current behavior, and there isn't anything worth noting
to users of the API today. The mutable behavior, if seen as its own
feature, has been removed for long enough that it's irrelevant.

https://mdn-bcd-collector.appspot.com/tests/api/Document/documentURI/readonly
was used to determine that documentURI was made readonly in Edge 17,
the only part of the data that could make this change <2 years.

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -3380,54 +3380,6 @@
             "standard_track": true,
             "deprecated": false
           }
-        },
-        "readonly": {
-          "__compat": {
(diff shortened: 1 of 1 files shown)


queengooborg (approved):
I'm all for removing this!  👍

PR #12277, opened 2021-09-06
Update flag data for Payment APIs; delete PaymentDetailsUpdate dictionary

queengooborg:
This PR's goal is to delete the `PaymentDetailsUpdate` dictionary from BCD.  First, this copies the Firefox flag, note and version data from `PaymentDetailsUpdate` to `PaymentRequest`, `PaymentMethodChangeEvent`, `PaymentRequestUpdateEvent` and `PaymentResponse`, which fixes consistency issues.  Afterwards, the entries within the dictionary are copied as subfeatures to both `PaymentRequest.show` and `PaymentRequestUpdateEvent.updateWith`, the two methods that use this dictionary.  Finally, the dictionary is deleted.


api/PaymentRequest.json
```diff
@@ -15,12 +15,36 @@
             "version_added": "15"
           },
           "firefox": {
-            "version_added": false,
-            "notes": "Available only in nightly builds. Requires <code>dom.payments.request.enabled</code> to be set to <code>true</code> and the comma-delineated list in <code>dom.payments.request.supportedRegions</code> to contain one or more of the supported 2-character ISO locales, currently US and CA."
+            "version_added": "55",
+            "notes": "Available only in nightly builds.",
```
foolip, line 19: This not appears to be wrong. The default value of dom.payments.request.enabled appears to be false in both Firefox Nightly (94) and Firefox 92. (If it were actually true, then we should use "preview" as the version number, not a note like this.)

I see this note is everywhere for payments stuff, can you either clean it all up or file an issue about it?

foolip (changes requested):

api/PaymentRequest.json
queengooborg: I've converted this note into the appropriate flag data for all of the instances in both this API and `PaymentRequestUpdateEvent`.  I'll get `PaymentMethodChangeEvent` and `PaymentResponse` updated shortly!

Catclaws12: > I've converted this note into the appropriate flag data for all of the instances in both this API and `PaymentRequestUpdateEvent`.  I'll get `PaymentMethodChangeEvent` and `PaymentResponse` updated shortly!

- - - **_[]()_**

api/PaymentMethodChangeEvent.json
```diff
@@ -15,12 +15,36 @@
             "version_added": "79"
           },
           "firefox": {
-            "version_added": false,
-            "notes": "Available only in nightly builds. Requires <code>dom.payments.request.enabled</code> to be set to <code>true</code> and the comma-delineated list in <code>dom.payments.request.supportedRegions</code> to contain one or more of the supported 2-character ISO locales, currently US and CA."
+            "version_added": "55",
+            "notes": "Available only in nightly builds.",
```
foolip, line 19: The note is still here :)

queengooborg: Ahh, I misread your comment before, my bad!

api/PaymentRequest.json
```diff
@@ -866,6 +1068,378 @@
             "standard_track": true,
             "deprecated": false
           }
+        },
+        "detailsPromise_displayItems_parameter": {
```
foolip, line 1072: This structure is the same as we'd have for `request.show({ displayItems: ...})` but the parameter here is a promise, not an object (IDL dictionary). So... I'm not sure if this needs an update to the guideline or  something.

Before going down that path, is it really necessary to represent this parameter at all? The data looks almost the same as the parent feature, if it's only a small difference between Chrome 60 and 61 I don't think that warrants a subfeature, and not even a note on the main feature.

foolip:
The `PaymentDetailsUpdate` dictionary is also used (as `optional Promise<PaymentDetailsUpdate> detailsPromise`) for `PaymentRequest`'s `show()` method. If it is worth capturing this data at all, it should be duplicated there. I'm not totally sure that it is however and would like to hear from @ddbeck or @Elchi3 how they'd deal with this situation.

foolip:
I've filed https://github.com/mdn/content/issues/9362 about a content issue here. For these methods that can take either _X_ or a promise resolving to _X_ I think it makes sense in BCD to just treat it as _X_, but it is novel and maybe the data guideline needs to be updated for this.

ddbeck:
@queengooborg I see that a bunch of the suggestions have resolved, but not others. Anything I can do to help this along? Or is it just a work in progress?

queengooborg:
Sorry, this one slipped my mind!  I just applied the suggested changes and I see the two comments regarding the nightly-availability builds.

api/PaymentResponse.json
```diff
@@ -15,12 +15,37 @@
...
+                "value_to_set": "A comma-delineated list of one or more 2-character ISO country codes indicating the countries in which to support payments (for example, <code>US,CA</code>."
+              }
+            ]
           },
           "firefox_android": {
-            "version_added": false,
-            "notes": "Available only in nightly builds. Requires <code>dom.payments.request.enabled</code> to be set to <code>true</code> and the comma-delineated list in <code>dom.payments.request.supportedRegions</code> to contain one or more of the supported 2-character ISO locales, currently US and CA."
+            "version_added": "55",
+            "version_removed": "79",
+            "notes": "Available only in nightly builds.",
```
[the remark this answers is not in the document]
queengooborg: So in other words, the data should be just like the following?

```json
"firefox_android": {
  "version_added": "55",
  "version_removed": "79",
  "flags": ...
}
```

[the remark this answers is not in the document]
ddbeck: That's right!

api/PaymentResponse.json
```diff
@@ -15,12 +15,37 @@
             "version_added": "15"
           },
           "firefox": {
-            "version_added": false,
-            "notes": "Available only in nightly builds. Requires <code>dom.payments.request.enabled</code> to be set to <code>true</code> and the comma-delineated list in <code>dom.payments.request.supportedRegions</code> to contain one or more of the supported 2-character ISO locales, currently US and CA."
+            "version_added": "55",
+            "notes": "Available only in nightly builds.",
```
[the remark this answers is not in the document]
ddbeck: To elaborate on this point: this note probably needs to be removed.

If it's true that this is nightly-only, then we'd represent that as `"version_added": "preview"` but that doesn't seem consistent with the rest of this PR.

api/PaymentResponse.json
```diff
@@ -66,12 +89,37 @@
               "version_added": "15"
             },
             "firefox": {
-              "version_added": false,
-              "notes": "Available only in nightly builds."
+              "version_added": "55",
+              "notes": "Available only in nightly builds.",
```
ddbeck, line 93: It's been a while since I've looked at this, but I _think_ we talked about dropping these notes on the suspicion that they're not accurate (otherwise, we could use `"preview"` values, if they're true).

queengooborg: Whoops!  Somehow I didn't catch these notes in the removal, fixed!

ddbeck (approved):
Looks good. Thank you! 🎉

Catclaws12:
- [ ] [@**__**](https://discord.com/channels/@me/905526921985675315/905821257243717662)

PR #12278, opened 2021-09-06
Remove incorrect note for element.getAnimations() in Safari

foolip:
This was implemented in https://trac.webkit.org/changeset/255149/webkit
which using WebKit version mapping would be Safari 14, but it must have
been backported.

https://software.hixie.ch/utilities/js/live-dom-viewer/?saved=9607 was
used to confirm support for the subtree parameter in Safari 13.1 on
Sauce Labs.

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -3465,9 +3465,7 @@
             ],
             "safari": [
               {
-                "version_added": "13.1",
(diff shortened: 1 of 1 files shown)


queengooborg (approved):
Good catch, thanks!

PR #12280, opened 2021-09-06
Add Safari versions for TextMetrics API

queengooborg:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `TextMetrics` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.2.12).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/TextMetrics

### DIFF
--- a/api/TextMetrics.json
+++ b/api/TextMetrics.json
@@ -33,7 +33,7 @@
             "version_added": "3.1"
           },
           "safari_ios": {
-            "version_added": true
+            "version_added": "3"
           },
           "samsunginternet_android": {
             "version_added": "1.0"
@@ -102,10 +102,10 @@
               "version_added": "55"
             },
             "safari": {
-              "version_added": true
+              "version_added": "11.1"
             },
             "safari_ios": {
-              "version_added": true
+              "version_added": "11.3"
             },
             "samsunginternet_android": {
               "version_added": "12.0"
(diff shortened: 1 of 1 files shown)


api/TextMetrics.json
```diff
@@ -33,7 +33,7 @@
             "version_added": "3.1"
           },
           "safari_ios": {
-            "version_added": true
+            "version_added": "3"
```
foolip, line 36: This doesn't match https://developer.mozilla.org/en-US/docs/Web/API/CanvasRenderingContext2D/measureText#browser_compatibility and I'm pretty sure the two should be the same version. Can you check double check the `measureText()` data and see if they really do disagree?

If `TextMetrics` was exposed before `measureText()`, then most likely it was just a missing flag in the IDL and accidental. Since that `TextMetrics` interface couldn't be used for anything, I think we should then still use the version from `measureText()`.

foolip (changes requested):

api/TextMetrics.json
queengooborg: This was from manual mirroring, and the `width` property is also set to iOS 3.2, so I trust the version number for `measureText()` more.  I'm unable to test anything earlier than Safari 4 at the moment either, so I don't have a way to verify whether this was actually exposed in Safari 3.1 or not.  (Both instances where "3.1" shows up came from the initial wiki migration.)

foolip (approved):

PR #12281, opened 2021-09-06
Adds font-synthesis: small-caps

rachelandrew:
Part of https://github.com/mdn/content/issues/8611

Firefox 93 adds support for `font-synthesis: small-caps`.

### DIFF
--- a/css/properties/font-synthesis.json
+++ b/css/properties/font-synthesis.json
@@ -93,7 +93,7 @@
             "status": {
               "experimental": true,
               "standard_track": true,
-              "deprecated": true
+              "deprecated": false
             }
           }
         }
--- a/css/properties/font-synthesis.json
+++ b/css/properties/font-synthesis.json
@@ -48,6 +48,54 @@
             "standard_track": true,
             "deprecated": false
           }
+        },
+        "small-caps": {
+          "__compat": {
+            "description": "<code>small-caps</code>",
+            "support": {
+              "chrome": {
+                "version_added": false
+              },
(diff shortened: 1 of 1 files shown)


css/properties/font-synthesis.json
```diff
@@ -48,6 +48,54 @@
...
+              },
+              "samsunginternet_android": {
+                "version_added": false
+              },
+              "webview_android": {
+                "version_added": false
+              }
+            },
+            "status": {
+              "experimental": false,
```
ddbeck, line 94: If there's only one implementation, then this should be experimental

```suggestion
              "experimental": true,
```

css/properties/font-synthesis.json
```diff
@@ -48,6 +48,54 @@
...
+              "standard_track": true,
+              "deprecated": true
```
ddbeck, line 96: Is this supposed to be marked as deprecated? The spec doesn't indicate anything about this being discouraged.

ddbeck (changes requested):
A couple of questions for you. Thank you, @rachelandrew.

css/properties/font-synthesis.json
rachelandrew:
```suggestion
              "deprecated": false
```

rachelandrew: No - I put it on the wrong key, which is how today is going, sorry!

ddbeck (approved):
Thank you! 🎉

PR #12289, opened 2021-09-07
onsecuritypolicyviolation global event handler and friends

hamishwillee:
The HTML spec added a global event handler `onsecuritypolicyviolation` in https://github.com/whatwg/html/pull/2651.
- Firefox adding in FF93 : https://bugzilla.mozilla.org/show_bug.cgi?id=1727302
- Webkit adding/added in https://bugs.webkit.org/show_bug.cgi?id=229381 
- Chromium not yet added https://bugs.chromium.org/p/chromium/issues/detail?id=1242893

I need help to work out whether this is in Safari yet. I also updated the `SecurityPolicyViolationEvent` to indicate it is not experimental - based on broad rollout. 

Note that the event fired by element is not in BCD - I added that in a separate PR: #12333

### DIFF
--- a/api/GlobalEventHandlers.json
+++ b/api/GlobalEventHandlers.json
@@ -3931,6 +3931,55 @@
           }
         }
       },
+      "onsecuritypolicyviolation": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/GlobalEventHandlers/onsecuritypolicyviolation",
+          "spec_url": "https://html.spec.whatwg.org/multipage/webappapis.html#handler-onsecuritypolicyviolation",
+          "support": {
+            "chrome": {
+              "version_added": false
+            },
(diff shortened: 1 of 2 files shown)


queengooborg (approved):
Looks good to me, thank you!

I looked into the WebKit commit more, and based upon the version number in the corresponding config file (https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Configurations/Version.xcconfig?rev=281569), the major version is set to "613" so I think this one will be landing in Safari 15.  Would you please set Safari to `"preview"` accordingly?

hamishwillee:
@queengooborg Thanks! I've updated Safari to "preview". This should be good to merge.

queengooborg (approved):

hamishwillee:
Thanks @queengooborg !! Also for the check on Safari!

PR #12291, opened 2021-09-07
Add Chromium versions for api.FetchEvent.request

queengooborg:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `request` member of the `FetchEvent` API, based upon commit history and date.

Commit: https://source.chromium.org/chromium/chromium/src/+/451a831fc3cb3c0b845298abbdde172219444d9f

### DIFF
--- a/api/FetchEvent.json
+++ b/api/FetchEvent.json
@@ -356,10 +356,10 @@
           "spec_url": "https://w3c.github.io/ServiceWorker/#fetch-event-request",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "40"
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": "40"
             },
             "edge": {
               "version_added": "17"
@@ -374,10 +374,10 @@
               "version_added": false
             },
             "opera": {
-              "version_added": true
+              "version_added": "27"
             },
             "opera_android": {
-              "version_added": true
+              "version_added": "27"
             },
             "safari": {
               "version_added": "11.1"
@@ -386,10 +386,10 @@
               "version_added": "11.3"
             },
             "samsunginternet_android": {
-              "version_added": true
+              "version_added": "4.0"
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "40"
             }
           },
           "status": {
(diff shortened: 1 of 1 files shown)


queengooborg:
The date of this particular commit is months before Chrome 40's feature freeze, which is when this event interface was implemented.  As such, I also set this feature to Chrome 40.

foolip:
I thought this would be easy to verify, but I guess the reason you didn't use the collector here is that it's broken. I've sent https://github.com/foolip/mdn-bcd-collector/pull/1479 + https://github.com/foolip/mdn-bcd-collector/pull/1481 but that won't be enough to test this I think.

https://developer.mozilla.org/en-US/docs/Web/API/Request#browser_compatibility says that `Request` was in Chrome 42, and the same is true of [`fetch()`](https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API#browser_compatibility) itself, so I guess the `FetchEvent` data is just wrong. Can you bump it to 42? That could still be wrong, but at least less wrong than now.

api/FetchEvent.json
```diff
@@ -356,10 +356,10 @@
           "spec_url": "https://w3c.github.io/ServiceWorker/#fetch-event-request",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "40"
```
foolip, line 359: This change, but everywhere and mirrored:

```suggestion
              "version_added": "42"
```

foolip (changes requested):

foolip:
Trying to confirm this again, I found an interesting note that might explain this. https://caniuse.com/fetch says "Only available in Chrome and Opera within ServiceWorkers" for Chrome 40-41, and if that's true for the whole fetch API things would make more sense. That would mean that `FetchEvent` actually was supported from Chrome 40, and we should mark `fetch()`, `Request` and probably a bunch of other things as supported from Chrome 40.

@queengooborg do you want to tackle that in this PR, or perhaps revert back to 40 and file an issue about the mess?

queengooborg:
I've reverted the changes from Chrome 42 -> 40 and filed an issue.  We should be able to fix this in the next collector version!

foolip (approved):

PR #12292, opened 2021-09-07
Delete RTCIceCandidateInit dictionary from BCD

queengooborg:
This PR removes the `RTCIceCandidateInit` dictionary from BCD.


api/RTCIceCandidate.json
```diff
@@ -124,6 +124,204 @@
             "standard_track": true,
             "deprecated": false
           }
+        },
+        "candidateInfo_candidate_parameter": {
```
foolip, line 128: These parameters all correspond to a property (IDL attributes) of `RTCIceCandidate`. I'm pretty sure that if we track down the versions, they'll be the same as the properties of the same name.

This is very common for constructor init dicts, and in these cases I don't think there's much use in having them in BCD.

queengooborg: The data already looks like it matches pretty well -- I'm down for removing these constructor subfeatures!

foolip (approved):

foolip:
@queengooborg I'll leave it to you to merge in case you want to rewrite the commit message.

PR #3287, opened 2019-01-11
[All] Improvements to RTCPeerConnection.json

jpmedley:
The non-Chrome changes came from Confluence. The Confluence data on this interface turned out to be really bad. You will wonder if some of these changes are correct. Here's the proof.

The webkit prefix was removed in Chrome 56. It appears that the prefix was only ever on the interface and not it’s individual members. I might have found this odd, but I’ve seen it before.
[constructor](https://chromium.googlesource.com/chromium/src/+/4426ed8c87c03f2c5d38aa75648154fc22cc564f)
Missing from IDL, not supported: [defaultIceServers, peerIdentity, sctp,onidentityresult, onidpassertionerror, onidpvalidationerror, onpeeridentity, getIdentityAssertion, setIdentityProvider, onicecandidateerror, getDefaultIceServers](https://cs.chromium.org/chromium/src/third_party/blink/renderer/modules/peerconnection/rtc_peer_connection.idl)
Promise-based [createAnswer, createOffer](https://storage.googleapis.com/chromium-find-releases-static/703.html#70302a3d99ed149d59acf592edf9d32d4fd0dbda) and [createAnswer added](https://chromium.googlesource.com/chromium/src/+/8f0a7f9dbbc977f76149b9e420b56bcaf5c81dc1), [createOffer added](https://chromium.googlesource.com/chromium/src/+/10646e4b58290fdc9942f3c0758de2c65649162d)
Promise-based [addIceCandidate, setLocalDescription, setRemoteDescription](https://chromium.googlesource.com/chromium/src/+/e16c96f5a5b517a64fdbc5be75f05e359e04b461)
[addTrack(), getSenders(), onTrack, removeTrack()](https://storage.googleapis.com/chromium-find-releases-static/fb2.html#fb265faa85701894a930a1024fba9c03c8f6a859)
[iceConnectionState, signalingState](https://chromium.googlesource.com/chromium/src/+/6c690b360cffb5f300f05382356d2fec3703f9dd)
[iceGatheringState](https://chromium.googlesource.com/chromium/src/+/b052a77a71916dfbd69029263ef437f98408df5f)
[localDescription, remoteDescription, setLocalDescription(), setRemoteDescription()](https://chromium.googlesource.com/chromium/src/+/d2df6ba6560e99f17cd900fdaa24f0ef4fafe203)
[...]

PR #1070, opened 2018-02-13
Add compat data for RTCPeerConnection

bunnybooboo:
Via #987 
https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection

PR #12300, opened 2021-09-08
Correct data for the RTCPeerConnection API

### LINKED ISSUE #11158: Meaningless "Promise-based version." notes for RTCPeerConnection
For example:
https://github.com/mdn/browser-compat-data/blob/2d29a84f220d74b6d957ec4767ffb55b88ee0b5f/api/RTCPeerConnection.json#L105-L114

This is a note for the `new RTCPeerConnection()` constructor, and makes no sense. This seems to be have copied and spread around at some point, and only partially clean up.

queengooborg:
This PR corrects the data throughout the RTCPeerConnection API, which was a **big** mess, mostly due to copy-paste error.  The fixes include the following:

- ~~Add data regarding the prefixed constructor~~ This one was done in #12196 already, though the non-prefixed version number was a little off based on my testing, so this is a quick correction.
- ~~Updates the notes regarding the default `rtcpMuxPolicy` value in Chrome when constructing a new instance.  The notes didn't mention what the new default value is, which is slightly unhelpful.  (There's only two values available, but the note update removes the need to scroll up on the page and find that out.)~~  Cherry-picked in #12301.
- Converts the "Promise-based version" notes into subfeatures.  It is more standard to use a subfeature rather than the notes.  Additionally, they were copied and pasted all across the Opera data, so it caused some major data discrepancies.  (Particularly, claiming support for features in Opera that weren't supported, wrong version numbers...)  Fixes #11158.
- Corrects the data regarding promise-based versions for Chrome.  After further digging in, it turns out that most of these features had the wrong version number.  New data comes from the commit history ([1](https://source.chromium.org/chromium/chromium/src/+/e16c96f5a5b517a64fdbc5be75f05e359e04b461), [2](https://source.chromium.org/chromium/chromium/src/+/70302a3d99ed149d59acf592edf9d32d4fd0dbda), [3](https://source.chromium.org/chromium/chromium/src/+/68f8e256dc7c7369127bd345100bcf776fa8acc0)).
- Corrects the Samsung Internet and Opera/Opera Android data. In the case of Samsung Internet, it appears to have been set to 6.0 most everywhere after a mirroring from the old, incorrect wiki data (#1606), and Opera's data initially came from the wiki tables (#1070).  When the Chrome data was corrected, however (see #3287), only the Chrome, Chrome Android, and WebView data was adjusted, leaving these other browsers untouched.
- Corrects the version number for `createDataChannel` based upon results from the mdn-bcd-collector project (v3.3.0).
- Finally, mirrors the Chrome data to both Chrome Android and WebView.  Both had been left as `true` for many of the features, however since we have ranged values, we're able to replace the `true` with `≤37` for WebView, and mirror Chrome straight to Chrome Android since we know Chrome Android has supported this API back then.

There's a lot of changes made in this PR, so I would love to get it reviewed and merged ASAP!


api/RTCPeerConnection.json
```diff
@@ -7,31 +7,25 @@
         "support": {
           "chrome": [
             {
-              "version_added": "55"
+              "version_added": "56"
```
foolip, line 10: This matches https://caniuse.com/rtcpeerconnection 👍

api/RTCPeerConnection.json
```diff
@@ -7,31 +7,25 @@
...
             },
             {
-              "prefix": "webkit",
-              "version_added": "15"
+              "version_added": "25",
+              "prefix": "webkit"
             }
           ],
+          "edge": {
+            "version_added": "15"
```
foolip, line 27: Edge should also have the prefixed variant, can you revert these edits?

api/RTCPeerConnection.json
```diff
@@ -110,33 +104,27 @@
           "support": {
             "chrome": [
               {
-                "version_added": "55",
-                "notes": "Before Chrome 63 the default value for the <code>RTCConfiguration.rtcpMuxPolicy</code> parameter was <code>&quot;negotiate&quot;</code>"
+                "version_added": "56",
+                "notes": "In Chrome 63, the default value for the <code>configuration.rtcpMuxPolicy</code> parameter changed from <code>&quot;negotiate&quot;</code> to <code>&quot;required&quot;</code>."
```
foolip, line 108: Can you update the notes in a separate PR? I'd like to ask @ddbeck to review them instead of me.

api/RTCPeerConnection.json
```diff
@@ -187,17 +175,23 @@
...
               }
             ],
-            "webview_android": {
-              "version_added": "≤37",
-              "notes": "Before WebView 63 the default value for the <code>RTCConfiguration.rtcpMuxPolicy</code> parameter was <code>&quot;negotiate&quot;</code>"
-            }
+            "webview_android": [
+              {
+                "version_added": "56",
+                "notes": "In Chrome 63, the default value for the <code>configuration.rtcpMuxPolicy</code> parameter changed from <code>&quot;negotiate&quot;</code> to <code>&quot;required&quot;</code>."
```
foolip, line 188: Oops, WebView&rarr;Chrome here. As requested, making the note changes separately would be great.

api/RTCPeerConnection.json
```diff
@@ -211,24 +205,12 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/RTCPeerConnection/addIceCandidate",
           "spec_url": "https://w3c.github.io/webrtc-pc/#dom-peerconnection-addicecandidate",
           "support": {
-            "chrome": [
-              {
-                "version_added": "51",
-                "notes": "Promise-based version."
```
foolip, line 217: These notes aren't a mistake, `addIceCandidate()` did change from callback-based to promise-based. I'd suggest a `partial_implementation` range for the callback variant.

api/RTCPeerConnection.json
```diff
@@ -715,24 +655,12 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/RTCPeerConnection/createAnswer",
           "spec_url": "https://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection-createanswer",
           "support": {
-            "chrome": [
-              {
-                "version_added": "51",
-                "notes": "Promise-based version."
```
foolip, line 721: This is also real, `createAnswer()` now returns a promise.

api/RTCPeerConnection.json
```diff
@@ -929,24 +801,12 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/RTCPeerConnection/createOffer",
           "spec_url": "https://w3c.github.io/webrtc-pc/#dom-rtcpeerconnection-createoffer",
           "support": {
-            "chrome": [
-              {
-                "version_added": "51",
-                "notes": "Promise-based version."
```
foolip, line 935: This is also real.

api/RTCPeerConnection.json
```diff
@@ -1600,32 +1348,12 @@
...
           "spec_url": "https://w3c.github.io/webrtc-pc/#widl-RTCPeerConnection-getStats-Promise-RTCStatsReport--MediaStreamTrack-selector",
           "support": {
-            "chrome": [
-              {
-                "version_added": "58",
-                "notes": "Promise resolves with <code>RTCStatsReport</code>."
-              },
-              {
-                "version_added": "54",
-                "notes": "Promise-based version."
```
foolip, line 1610: This is also real, shouldn't just be removed.

api/RTCPeerConnection.json
```diff
@@ -1600,32 +1348,12 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/RTCPeerConnection/getStats",
           "spec_url": "https://w3c.github.io/webrtc-pc/#widl-RTCPeerConnection-getStats-Promise-RTCStatsReport--MediaStreamTrack-selector",
           "support": {
-            "chrome": [
-              {
-                "version_added": "58",
-                "notes": "Promise resolves with <code>RTCStatsReport</code>."
```
foolip, line 1606: What did this note mean? Are you sure it doesn't matter?

api/RTCPeerConnection.json
```diff
@@ -1736,13 +1444,16 @@
             "description": "Uses callbacks instead of promises",
```
foolip, line 1444: Can you delete this subfeature and use the same `partial_implementation` ranges you'll be used for other methods that went from callbacks-only to callbacks-and-promises?

api/RTCPeerConnection.json
```diff
@@ -3720,24 +3155,12 @@
           "spec_url": "https://w3c.github.io/webrtc-pc/#dom-peerconnection-setlocaldescription",
           "description": "<code>setLocalDescription()</code>",
           "support": {
-            "chrome": [
-              {
-                "version_added": "51",
-                "notes": "Promise-based version."
```
foolip, line 3726: This is real.

api/RTCPeerConnection.json
```diff
@@ -3859,24 +3254,12 @@
           "spec_url": "https://w3c.github.io/webrtc-pc/#dom-peerconnection-setremotedescription",
           "description": "<code>setRemoteDescription()</code>",
           "support": {
-            "chrome": [
-              {
-                "version_added": "51",
-                "notes": "Promise-based version."
```
foolip, line 3865: This is real.

foolip (changes requested):

api/RTCPeerConnection.json
queengooborg: After digging through the commit history, it looks like this comes from #3287, and is about https://chromestatus.com/feature/5665052275245056.  It seems to be a change in the object that's returned by `getStats()`, but honestly, I'm not seeing anything different between Chrome 57 and the latest Chrome (or Chrome 58 and 60 for that matter).

Screen Shot 2021-09-08 at 02 39 40

I didn't think this note was that important, particularly for such an old Chrome version, but now it looks like it's incorrect as well.

api/RTCPeerConnection.json
queengooborg: The spec had actually made a change from callbacks to promise-based syntax (as suggested in https://groups.google.com/a/chromium.org/g/blink-dev/c/V7-02zJiVzw), so maybe it would be better to create a subfeature for these, `returns_promise`?  (There's no data guideline for this, just an open discussion in #4468, but a quick search in BCD shows this is the more common way to represent the transition.  This file is also the only one with that `with_callback` subfeature so we can reverse it and turn it into `returns_promise`.)

queengooborg: I've now noticed your PR to make the name of the subfeatures consistent in #11630!

foolip: Sure, using `returns_promise` subfeatures sounds OK. It does make it more prominent, and I suspect supported this from the beginning, but I think it'll be OK.

api/RTCPeerConnection.json
foolip: I think what happened here is that the promise-based variant has always returned something different from the callback-based variant. So a `returns_promise` subfeature should be enough here.

api/RTCPeerConnection.json
queengooborg: All done, and ready for re-review!

foolip:
I'd like to hold off with this until @ddbeck gets a chance to review https://github.com/mdn/browser-compat-data/pull/11630, since it assumes the proposed guideline.

api/RTCPeerConnection.json
```diff
@@ -243,58 +237,78 @@
...
           },
           "status": {
             "experimental": false,
             "standard_track": true,
             "deprecated": false
           }
+        },
+        "returns_promise": {
+          "__compat": {
+            "description": "Returns a promise",
```
foolip, line 267:
```suggestion
            "description": "Returns a <code>Promise</code>",
```

foolip (changes requested):
Descriptions for all `returns_promise` entries need updating.

api/RTCPeerConnection.json
```diff
@@ -959,58 +927,78 @@
...
+                "version_added": false
+              },
+              "opera": {
+                "version_added": "43"
+              },
+              "opera_android": {
+                "version_added": "43"
+              },
+              "safari": {
+                "version_added": false
```
foolip, line 984: Can you double check this?

api/RTCPeerConnection.json
```diff
@@ -1639,89 +1523,69 @@
...
         },
-        "selector_parameter": {
+        "returns_promise": {
           "__compat": {
-            "description": "<code>selector</code> parameter",
+            "description": "Returns a <code>Promise</code>",
             "support": {
               "chrome": {
-                "version_added": "67"
+                "version_added": "51"
```
foolip, line 1555: Was both 54 and 58 in the original data wrong?

api/RTCPeerConnection.json
```diff
@@ -4035,6 +3588,54 @@
...
+                "version_added": false
+              },
+              "opera": {
+                "version_added": "43"
+              },
+              "opera_android": {
+                "version_added": "43"
+              },
+              "safari": {
+                "version_added": false
```
foolip, line 3621: Can you double check this?

api/RTCPeerConnection.json
```diff
@@ -243,58 +237,78 @@
...
```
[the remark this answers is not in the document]
queengooborg: Confirmed all of them, good catch, thanks!

api/RTCPeerConnection.json
queengooborg: Looks like there were some problems with this data, fixed!

foolip (approved):
I've reviewed everything again and can't spot any more issues. Thanks for sorting out this huge mess, this is much, much better now! 🥳

foolip:
@ddbeck I think the new subfeatures for promises are release note worthy.

ddbeck:
@foolip Good idea. In [the forthcoming release notes](https://github.com/mdn/browser-compat-data/pull/12543), I've included this with the text for https://github.com/mdn/browser-compat-data/pull/11630. If you want to review today before the release, that would be helpful (but not required).

foolip:
@ddbeck I've reviewed the release notes after the fact, looks good, thank you!

PR #12301, opened 2021-09-08
Update notes regarding default value of an RTCPeerConnection option

queengooborg:
Cherry-picked from #12300.  This PR updates the notes regarding the default value of `rtcpMuxPolicy` for Chrome when creating a new instance and indicates what the new default value is.  Currently, the notes don't mention what the new default value is, which is slightly unhelpful.  (There's only two values available, but this note update removes the need to scroll up on the page and find that out.)

### DIFF
--- a/api/RTCPeerConnection.json
+++ b/api/RTCPeerConnection.json
@@ -111,7 +111,7 @@
             "chrome": [
               {
                 "version_added": "55",
-                "notes": "In Chrome 63, the default value for the <code>configuration.rtcpMuxPolicy</code> parameter changed from <code>&quot;negotiate&quot;</code> to <code>&quot;required&quot;</code>."
+                "notes": "Before Chrome 63, the default value for the <code>configuration.rtcpMuxPolicy</code> parameter is <code>&quot;negotiate&quot;</code> instead of <code>&quot;required&quot;</code>."
               },
               {
                 "prefix": "webkit",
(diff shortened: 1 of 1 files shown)


foolip:
@ddbeck I believe you have fairly strong opinions on the wording of notes like this?

api/RTCPeerConnection.json
```diff
@@ -111,7 +111,7 @@
             "chrome": [
               {
                 "version_added": "55",
-                "notes": "Before Chrome 63 the default value for the <code>RTCConfiguration.rtcpMuxPolicy</code> parameter was <code>&quot;negotiate&quot;</code>"
+                "notes": "In Chrome 63, the default value for the <code>configuration.rtcpMuxPolicy</code> parameter changed from <code>&quot;negotiate&quot;</code> to <code>&quot;required&quot;</code>."
```
ddbeck, line 114: Yeah, I like the idea of making the note standalone better. A couple of adjustments to suggest anyway:

- Avoiding "In version X" when we mean "In several versions of a range starting or ending with X" since "In X" might be interpreted as applying to X and only X.
- Present tense, a style convention I like (kinda related to above, but also older releases aren't usualy changed retroactively so it's a little weird to talk about the behavior in the past tense, though this is less weird the older a release is)

```suggestion
                "notes": "Before Chrome 63, the default value for the <code>configuration.rtcpMuxPolicy</code> parameter is <code>&quot;negotiate&quot;</code> instead of <code>&quot;required&quot;</code>."
```

jpmedley: +1

jpmedley (approved):

PR #12317, opened 2021-09-09
Mark AbortSignal.abort() as shipped in Chromium 93

foolip:
https://www.chromestatus.com/feature/5642501387976704

Also confirmed manually in Chrome 93.


lucalves (approved):

ddbeck (approved):
Thanks, @foolip!

PR #12323, opened 2021-09-09
Set WebAuthn APIs as no longer experimental

lukewarlow:
#### Summary

Unmarks webauthn APIs with cross-browser support as no longer experimental

Just based on existing support data


lukewarlow:
The support data for Webauthn overall is of fairly poor quality, I'm going to be doing follow up MRs that improve the browser specific support data once I've worked out the versions better. Firefox Android 92 recently added support which caused me to find these issues.

PR #12325, opened 2021-09-09
Set WebAuthn APIs as supported in Firefox Android 92

lukewarlow:
#### Summary

This updates Firefox Android support for WebAuthn related APIs.

I've set them all as fully supported in Firefox 92. And left the preview entries as partial implementations.

I will file off an issue to investigate the extent of the implementation pre-92. I know that it was partially implemented as it passes basic feature detection but fails weirdly on https://webauthn.io

Tested on https://webauthn.io

And https://www.mozilla.org/en-US/firefox/android/92.0/releasenotes/#:~:text=Added%20support%20for%20Web%20Authentication%20API

### DIFF
--- a/api/AuthenticatorAssertionResponse.json
+++ b/api/AuthenticatorAssertionResponse.json
@@ -18,10 +18,17 @@
             "version_added": "60",
             "notes": "Only supports USB U2F tokens."
           },
-          "firefox_android": {
-            "version_added": "60",
-            "notes": "Only supports USB U2F tokens."
-          },
+          "firefox_android": [
+            {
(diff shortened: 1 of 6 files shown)


lukewarlow:
https://developer.mozilla.org/en-US/docs/Web/API/Web_Authentication_API - Follow up issue about the notes

ddbeck (approved):
It turns out those false to 92 entries will get resolved more cleanly with https://github.com/mdn/browser-compat-data/pull/12771. I'll go ahead and merge this one. Thank you, @lukewarlow.

PR #12332, opened 2021-09-10
Selection API: change experimental to false

hamishwillee:
[Selection](https://developer.mozilla.org/en-US/docs/Web/API/Selection) appears to be well supported. This removes the markup that it is experimental.

As discussed in #12126


ddbeck (approved):
Looks good. Thank you! 🎉

PR #12333, opened 2021-09-10
Add Element securitypolicyviolation_event

hamishwillee:
This adds `securitypolicyviolation_event` to the `Element` API as a follow on from #12289.

This appears to have been a long term omission. 
- I have added it to Element because that is what the [spec says here](https://html.spec.whatwg.org/#event-securitypolicyviolation)
- I have matched the versions of the associated event object [SecurityPolicyViolationEvent](https://developer.mozilla.org/en-US/docs/Web/API/SecurityPolicyViolationEvent) which I would have expected to have been added at the same time.
- I did not add the spec URL because it causes an assertion error: "spec_url": "https://html.spec.whatwg.org/#event-securitypolicyviolation",

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -7648,6 +7648,55 @@
           }
         }
       },
+      "securitypolicyviolation_event": {
+        "__compat": {
+          "description": "<code>securitypolicyviolation</code> event",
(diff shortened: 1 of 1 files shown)


hamishwillee:
@ddbeck Would be great if you could review this please.

PR #12336, opened 2021-09-10
FF93 onslotchange to globalevents and shadowroot

hamishwillee:
FF93 added the event handler `onslotchange`  to  ShadowRoot and also the GlobalEventHandler. This PR adds the version to FF for ShadowRoot and creates new entry for the GlobalEvent Handler. 

Note, I have set this as experimental, as only one implementation

- FF bug https://bugzilla.mozilla.org/show_bug.cgi?id=1501983. I have also tested this and verified in FF93 (and not present in FF92)
- Tested Chrome and it is not present - intent to ship is recent and here: https://groups.google.com/a/mozilla.org/g/dev-platform/c/tHIn-Ce1DSE
- Safari is in preview apparently https://bugs.webkit.org/show_bug.cgi?id=191310 though I have not tested this.
- Other docs work can be tracked in https://github.com/mdn/content/issues/8624

### DIFF
--- a/api/ShadowRoot.json
+++ b/api/ShadowRoot.json
@@ -307,10 +307,10 @@
               "version_added": false
             },
             "firefox": {
-              "version_added": false
+              "version_added": "93"
             },
             "firefox_android": {
-              "version_added": false
+              "version_added": "93"
             },
             "ie": {
               "version_added": false
--- a/api/GlobalEventHandlers.json
+++ b/api/GlobalEventHandlers.json
@@ -4325,6 +4325,55 @@
           }
         }
       },
+      "onslotchange": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/GlobalEventHandlers/onslotchange",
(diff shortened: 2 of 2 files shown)


hamishwillee:
@ddbeck Would be great if this could be reviewed.

api/ShadowRoot.json
```diff
@@ -307,10 +307,10 @@
               "version_added": false
             },
             "firefox": {
-              "version_added": false
+              "version_added": "93"
             },
             "firefox_android": {
-              "version_added": false
+              "version_added": "93"
```
ddbeck, line 313: With Safari, experimental can be `false`

ddbeck (changes requested):
One little fix needed. Thank you, @hamishwillee!

api/ShadowRoot.json
hamishwillee: Thanks. I fixed this, and a couple of other cases in the docs - including the top level API.

hamishwillee:
Done. There were a few cases that were incorrectly marked experimental in ShadowRoot (unrelated to this change). I fixed those - note that there are still experimental cases in this API though - methods that have only one implementer.

PR #12343, opened 2021-09-11
Misc cleanup in scripts/

saschanaz:
#### Summary

Random things in scripts/

### DIFF
--- a/scripts/migrations/002-remove-webview-flags.test.js
+++ b/scripts/migrations/002-remove-webview-flags.test.js
@@ -3,10 +3,7 @@
  * http://creativecommons.org/publicdomain/zero/1.0/ */
 
 'use strict';
-const fs = require('fs');
-const path = require('path');
 const chalk = require('chalk');
-const { platform } = require('os');
 
 const { removeWebViewFlags } = require('./002-remove-webview-flags.js');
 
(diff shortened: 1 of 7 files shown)


scripts/migrations/002-remove-webview-flags.test.js
```diff
@@ -3,10 +3,7 @@
  * http://creativecommons.org/publicdomain/zero/1.0/ */
```
saschanaz, line 3: Are the migrations/ things still needed? @queengooborg

queengooborg: We aren't really using the migration scripts for anything, but I'll defer to @ddbeck on whether we should remove them.

scripts/release-stats.js
```diff
@@ -30,25 +27,21 @@ const getJSON = url =>
...
 };
 
-const confirm = str => !['n', 'no'].includes(str.toLowerCase());
-
 const prompt = async questions => {
   const results = {};
   for (const q of questions) {
-    const options = q.type === confirm ? '(Y/n) ' : '';
-    results[q.name] = await question(`${q.message} ${options}`).then(q.type);
+    results[q.name] = await question(q.message).then(q.type);
```
ddbeck, line 44: Can I have the trailing space back? Looks a little nicer 😄

```suggestion
    results[q.name] = await question(`${q.message} `).then(q.type);
```

ddbeck (changes requested):
A couple little things

ddbeck (approved):
Thank you! 🎉

Issue #12099, opened 2021-08-21
CSS color-mix & color-contrast - Available on iOS 15 Safari as experimental


formspiel:
Both CSS properties are available as experimental features in Safari iOS (beta 6). See screenshot
![image](https://user-images.githubusercontent.com/1277141/130336003-a044e43e-763e-4006-93a4-bfeb13b7e86f.png)

ddbeck:
Fixed by #12361. Thanks for the report!

PR #12361, opened 2021-09-12
Add safari version for css color-mix()

aeharding:
#### Summary

`color-mix()` is behind a flag in Safari 15.

 * https://trac.webkit.org/changeset/273244/webkit/
 * https://webkit.org/blog/11577/release-notes-for-safari-technology-preview-122/#post-11577:~:text=CSS%20Color,-Added

I also installed Safari 15 (beta) on my device, and verified the flag is there and works as expected.

image

#### Related issues

N/A

### DIFF
--- a/css/types/color.json
+++ b/css/types/color.json
@@ -257,7 +257,13 @@
                 ]
               },
               "safari_ios": {
-                "version_added": false
+                "version_added": "15",
+                "flags": [
+                  {
+                    "type": "preference",
+                    "name": "CSS color-contrast()"
+                  }
+                ]
               },
               "samsunginternet_android": {
                 "version_added": false
@@ -327,7 +333,13 @@
                 ]
               },
               "safari_ios": {
-                "version_added": false
+                "version_added": "15",
+                "flags": [
+                  {
+                    "type": "preference",
+                    "name": "CSS color-mix()"
+                  }
+                ]
               },
               "samsunginternet_android": {
                 "version_added": false
--- a/css/types/color.json
+++ b/css/types/color.json
@@ -319,7 +319,14 @@
                 "version_added": false
               },
               "safari": {
-                "version_added": false
+                "version_added": "15",
+                "flags": [
+                  {
+                    "type": "preference",
+                    "name": "CSS color-mix()",
+                    "value_to_set": "true"
+                  }
+                ]
               },
               "safari_ios": {
                 "version_added": false
--- a/css/types/color.json
+++ b/css/types/color.json
@@ -252,8 +252,7 @@
                 "flags": [
                   {
                     "type": "preference",
-                    "name": "CSS color-contrast()",
-                    "value_to_set": "true"
+                    "name": "CSS color-contrast()"
                   }
                 ]
               },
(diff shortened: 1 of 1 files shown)


css/types/color.json
```diff
@@ -319,7 +319,14 @@
                 "version_added": false
               },
               "safari": {
-                "version_added": false
+                "version_added": "15",
+                "flags": [
+                  {
+                    "type": "preference",
+                    "name": "CSS color-mix()",
+                    "value_to_set": "true"
```
ddbeck, line 327: IIRC, `value_to_set` is optional. Since there's no actual value to choose in the Safari UI, how about we omit it?

```suggestion
                    "name": "CSS color-mix()"
```

ddbeck (changes requested):
Hi @aeharding and thank you for this PR. I see that it's your first to the project—welcome!

I have one suggestion in a line comment. Otherwise, this looks good.

css/types/color.json
aeharding: Hi @ddbeck, I was just following the pattern I saw here:

https://github.com/mdn/browser-compat-data/blob/db3cb3ec3984f24f277daa25a28eb15b2c45f936/css/types/color.json#L256

Do you want me to remove it from `color-contrast()` as well?

aeharding: Fixed :)

lukewarlow:
Would be good to add this information to iOS Safari too as that also has these features behind a flag

ddbeck:
@lukewarlow what does the UI for browser preferences look like on iOS? Do you have a screenshot? I've never seen anything like that and was under the impression that prefs were unsupported on iOS

lukewarlow:
> @lukewarlow what does the UI for browser preferences look like on iOS? Do you have a screenshot? I've never seen anything like that and was under the impression that prefs were unsupported on iOS

I didn't realise it was a thing until very recently too. It's hidden away in Settings -> Safari -> Advanced -> Experimental Features.

Here's a screenshot from an iPad.

image

lukewarlow:
I'm happy to do a follow up PR that adds all the flag data for Safari iOS rather than doing just these two in this PR.

aeharding:
@lukewarlow nice catch, I've added that to this PR (might as well get it done now). Can you please review? I assume the naming of the preferences are the same, and it's only available starting in Safari for iOS v15.

aeharding:
This should also close #12099

lukewarlow:
Yeah the flags are the same as desktop generally (and specifically in this case). Lgtm fwiw

ddbeck:
@lukewarlow 

> I'm happy to do a follow up PR that adds all the flag data for Safari iOS rather than doing just these two in this PR.

Yeah, I think doing this in a separate PR would be the way to go. 👍

ddbeck (approved):
This looks read to go now. Thank you! 🎉

lukewarlow:
> @lukewarlow
> 
> > I'm happy to do a follow up PR that adds all the flag data for Safari iOS rather than doing just these two in this PR.
> 
> Yeah, I think doing this in a separate PR would be the way to go. 👍

https://github.com/mdn/browser-compat-data/pull/12500 - I've started work on this here. Will do another one adding Safari 14 flags in future.

PR #12368, opened 2021-09-13
Drop the “type” attribute from ImageDecoder

sideshowbarker:
https://w3c.github.io/webcodecs/#imagedecoder-interface doesn’t show any `type` attribute for the `ImageDecoder` interface/object.

---

Maybe this was supposed to be for `isTypeSupported`?

### DIFF
--- a/api/ImageDecoder.json
+++ b/api/ImageDecoder.json
@@ -441,8 +441,6 @@
       },
       "type": {
         "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/ImageDecoder/type",
-          "spec_url": "https://w3c.github.io/webcodecs/#dom-imagedecoder-type",
           "support": {
             "chrome": {
               "version_added": "94"
@@ -482,8 +480,8 @@
             }
           },
           "status": {
-            "experimental": true,
-            "standard_track": true,
+            "experimental": false,
+            "standard_track": false,
             "deprecated": false
           }
         }


rachelandrew:
The boilerplate was generated from the Chromium idl, which does seem to include `type` https://chromium.googlesource.com/chromium/src/+/refs/heads/main/third_party/blink/renderer/modules/webcodecs/image_decoder.idl

I had a poke around the issues in the repo to see if I could find anything about it (sometimes I find things pending spec edits) but haven't so far. I'll try to find out.

sideshowbarker:
> I had a poke around the issues in the repo to see if I could find anything about it (sometimes I find things pending spec edits) but haven't so far. I'll try to find out. 

@rachelandrew Any update on that?

ddbeck:
As a quick check on this, I tried `"type" in ImageDecoder.prototype` in Chrome and it exists. Maybe the right move here is to remove the spec URL and set it as non-standard track instead of removing it?

rachelandrew:
Sorry for the delay, I found the original discussion about this eventually. So when I spoke to one of the editors about this I was told that `ImageDecoder.type` is there by mistake. It's redundant, and they intend to deprecate it from Chrome.

sideshowbarker:
> As a quick check on this, I tried `"type" in ImageDecoder.prototype` in Chrome and it exists. Maybe the right move here is to remove the spec URL and set it as non-standard track instead of removing it?

Make sense — I’ve pushed an update which does that

ddbeck (approved):
Thank you! 🎉

ddbeck:
(Closed and reopened to force CI to become unstuck.)

PR #12373, opened 2021-09-13
Adding options parameter to ResizeObserver

rachelandrew:
As discussed in https://github.com/mdn/content/issues/8622

### DIFF
--- a/api/ResizeObserver.json
+++ b/api/ResizeObserver.json
@@ -197,7 +197,7 @@
         },
         "options_box_parameter": {
           "__compat": {
-            "description": "<code>options</code> parameter",
+            "description": "<code>options.box</code> parameter",
             "support": {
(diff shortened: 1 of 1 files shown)


api/ResizeObserver.json
```diff
@@ -194,6 +194,63 @@
             "standard_track": true,
             "deprecated": false
           }
+        },
+        "options_box_parameter": {
+          "__compat": {
+            "description": "<code>options</code> parameter",
```
ddbeck, line 200:
```suggestion
            "description": "<code>options.box</code> parameter",
```

ddbeck (changes requested):
Looks great, @rachelandrew! Just one fix and we're good to go.

ddbeck (approved):
Thank you!

PR #12376, opened 2021-09-13
Remove irrelevant Transferable API

queengooborg:
This PR removes the irrelevant `Transferable` API from BCD.  This used to be an abstract interface, but has long since become a WebIDL keyword.


Elchi3 (approved):
Thanks for the quick fix!

PR #12382, opened 2021-09-13
Add Chromium versions for api.MouseEvent.region

queengooborg:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `region` member of the `MouseEvent` API, based upon commit history and date.

Commit: https://storage.googleapis.com/chromium-find-releases-static/f18.html#f18571e9097d0533f6fc18d62b3ab006e1d6c6d4

### DIFF
--- a/api/MouseEvent.json
+++ b/api/MouseEvent.json
@@ -1030,7 +1030,7 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/MouseEvent/region",
           "support": {
             "chrome": {
-              "version_added": true,
+              "version_added": "51",
               "flags": [
                 {
                   "type": "preference",
@@ -1040,10 +1040,17 @@
               ]
             },
             "chrome_android": {
-              "version_added": false
+              "version_added": "51",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "Experimental Web Platform Features",
+                  "value_to_set": "true"
+                }
+              ]
             },
             "edge": {
-              "version_added": "≤79",
+              "version_added": "79",
               "flags": [
                 {
                   "type": "preference",
@@ -1076,10 +1083,24 @@
               "version_added": false
             },
             "opera": {
-              "version_added": false
+              "version_added": "38",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "Experimental Web Platform Features",
+                  "value_to_set": "true"
+                }
+              ]
             },
             "opera_android": {
-              "version_added": false
+              "version_added": "41",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "Experimental Web Platform Features",
+                  "value_to_set": "true"
+                }
+              ]
             },
             "safari": {
               "version_added": false


jpmedley:
@queengooborg when I see a feature that's been behind a Chrome flag for three years, I'm suspicious that it has a future. My research so far makes me think that this is supposed to be removed from Chrome flag and all, but it somehow slipped through the cracks. 

Do you know anything more about this?

foolip:
@jpmedley this is still in Chromium source:
https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/events/mouse_event.idl;l=68-70;drc=035ad13ca11618ee85f9b32e45a36ddae6104424

But it's still not enabled by default:
https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/platform/runtime_enabled_features.json5;l=352-353;drc=0d6681f391c7f9f33d7bd6c5e4438454385817ea

Unfortunately we don't have a guideline for removing features like this.

foolip (approved):

queengooborg:
I can't say I know too much detail about this API, but I do remember hit regions remaining an experimental feature for the time being, even if it has been so for a long time.  I'm down for marking this as simply unsupported if that is preferred!

jpmedley:
@foolip I wish you hadn't merged this. I know it was in the source, and I know it hasn't been enabled by default. Those weren't the issue. I'm trying to determine if this feature has a chance of moving to stable. I dug into the intents, bugs, and other activity around this and it shows every sign of being an abandoned feature that wasn't cleaned up.

jpmedley:
I just got this from one of the developers on this. I asked him to comment here: "The feature is implemented in Chrome, but we never shipped it and probably never will."

queengooborg:
At the very least, the data merged in this PR isn't inaccurate, since it is a feature implemented behind a flag.  It's probably better that we've merged this as-is, as this issue is bigger than just this one feature and I believe needs some further discussion.

I'll open up an issue to continue this discussion further and ping you, @jpmedley, with some more details!

foolip:
@jpmedley Sorry, I didn't take your comment to be asking for changes to this PR, just asking out of curiosity. If you want to signal strongly to other reviewers to not merge, I'd suggest the "request changes" review state, or adding the "not ready" label.

AFAIK we don't have a way to represent "experiment that will likely never ship" in BCD, or do you have an idea for it?

jpmedley:
I'm saying it should be removed. The only reason to use this is for a developer to try it. The developer said they're likely to remove it, but he's pulling his punches. The team deleted the Chrome Status entry related to it. 

User should never use this either. In fact, users should never flip a Chrome flag. Ever. They're for trying out features.

foolip:
I would love for this feature to be removed, but since there are still two experimental implementations behind flags, it seems like a novel case that may require tweaking the [irrelevant features guideline](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-features). @jpmedley do you want to send a PR and make the case for it?

@junov note that this was removed from the spec way back in https://github.com/whatwg/html/pull/1942.

PR #12383, opened 2021-09-13
Remove irrelevant api.HTMLSourceElement.keySystem feature

queengooborg:
This PR removes the irrelevant `keySystem` member of the `HTMLSourceElement` API as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-features). The lack of current support has been confirmed by the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.3.0), even if the current BCD suggests support.

### DIFF
--- a/api/HTMLSourceElement.json
+++ b/api/HTMLSourceElement.json
@@ -96,65 +96,6 @@
           }
         }
       },
-      "keySystem": {
-        "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLSourceElement/keySystem",
-          "support": {
-            "chrome": {
-              "version_added": true
-            },
-            "chrome_android": {
-              "version_added": true
-            },
-            "edge": {
-              "version_added": "≤18"
-            },
-            "firefox": {
-              "version_added": true,
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "media.eme.enabled",
-                  "value_to_set": "true"
-                }
-              ]
-            },
-            "firefox_android": {
-              "version_added": true,
-              "flags": [
-                {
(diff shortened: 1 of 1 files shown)


queengooborg:
Additional comments: it seems that the `media.eme.enabled` flag has been enabled by default, but this feature is not around in modern Firefox versions.  None of the other browsers support the feature.  I don't see anything in the spec about this feature, and the MDN page is already deleted.

foolip:
This was probably a precursor to https://developer.mozilla.org/docs/Web/API/MediaKeySystemAccess/keySystem

api/HTMLSourceElement.json
```diff
@@ -96,65 +96,6 @@
           }
         }
       },
-      "keySystem": {
-        "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLSourceElement/keySystem",
-          "support": {
-            "chrome": {
-              "version_added": true
```
foolip, line 104: I've confirmed that Chromium doesn't has `keySystem` on `HTMLSourceElement`.

api/HTMLSourceElement.json
```diff
@@ -96,65 +96,6 @@
...
-            },
-            "chrome_android": {
-              "version_added": true
-            },
-            "edge": {
-              "version_added": "≤18"
```
foolip, line 110: This seems to also have been wrong, https://developer.mozilla.org/docs/Web/API/MediaKeySystemAccess/keySystem return false in Edge 18.

foolip (approved):

PR #12393, opened 2021-09-14
Remove wrong mdn_url from HTMLBaseFontElement

teoli2003:
These mdn urls don't exist.

They will never exist as this feature is in the progress of being completely removed; only waiting for IE11 to be retired.

(They lead to MDN flaws, hiding real flaws)

### DIFF
--- a/api/HTMLBaseFontElement.json
+++ b/api/HTMLBaseFontElement.json
@@ -53,7 +53,6 @@
       },
       "color": {
         "__compat": {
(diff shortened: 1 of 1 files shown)


foolip (approved):
This will also remove broken links from https://developer.mozilla.org/en-US/docs/Web/API/HTMLBaseFontElement#browser_compatibility 👍

teoli2003:
> This will also remove broken links from https://developer.mozilla.org/en-US/docs/Web/API/HTMLBaseFontElement#browser_compatibility 👍

Yes, this is why I detected them in the first place.

PR #12397, opened 2021-09-15
Remove MediaStreamConstraints dictionary from BCD

queengooborg:
This PR removes the `MediaStreamConstraints` dictionary from BCD, folding its two parameters into the only function that uses it (`api.MediaDevices.getUserMedia`).  This is also preliminary work for the removal of the `MediaTrackConstraints` dictionary as well.


foolip:
Is there an MDN PR for this?

I don't think that capturing whether `audio` and `video` parameters are supported is the thing that matters to web developers, rather it's the next level, constraints like `deviceId` (all tracks), `echoCancellation` (audio tracks), or `frameRate` (video tracks). That's further complicated by `exact` and `ideal` in many contexts.

Ultimately I don't think https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#parameters-and-parameter-object-features will work here, because something like `deviceId` isn't *one* parameter, but a parameter that can appear in many places.

api/MediaDevices.json
```diff
@@ -402,6 +402,102 @@
             "deprecated": false
           }
         },
+        "constraints_audio_parameter": {
```
saschanaz, line 405: I'd like to use this data for automation, is there a written rule to name this? What happens when the spec changes the argument name? Will the name here also change?

foolip: The rules are in https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#parameters-and-parameter-object-features. For the most part the names match the spec and I've been changing some to match, but I can't imagine they all match.

I would guess that in the vast majority of cases there's just one dictionary argument and the argument name can be ignored / assumed to match. That only leaves cases with multiple dictionary arguments needing to match on the name.

ashnicdill:

api/MediaDevices.json
queengooborg: I've removed these subfeatures since their data matches their parent, and as indicated, there's no need for the individual `audio` and `video` features anyways!

foolip (approved):

PR #12399, opened 2021-09-15
Add Chromium versions for RTCPeerConnectionIceEvent API

queengooborg:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `RTCPeerConnectionIceEvent` API, based upon commit history and date.

Commit: https://source.chromium.org/chromium/chromium/src/+/17f2e5e5850823844c8d6127b1b8ec068a7cc736

### DIFF
--- a/api/RTCPeerConnectionIceEvent.json
+++ b/api/RTCPeerConnectionIceEvent.json
@@ -8,7 +8,7 @@
           "chrome": [
             {
               "alternative_name": "RTCIceCandidateEvent",
-              "version_added": true,
+              "version_added": "24",
               "version_removed": "56"
             },
             {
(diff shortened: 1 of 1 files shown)


foolip:
The `candidate` member is in the linked commit, but the data in this file isn't updated. Can you update that to match?

foolip (approved):
Consistency check: https://source.chromium.org/chromium/chromium/src/+/17f2e5e5850823844c8d6127b1b8ec068a7cc736 also adds `addIceCandidate` and `onicecandidate` and those are also marked as Chrome 24 in https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection#browser_compatibility 👍

PR #12403, opened 2021-09-15
Remove Window.routeEvent

teoli2003:
The last browser to support this was Firefox 25. I'm about to delete the related MDN page.


teoli2003:
This will need a "label:release_notes" or similar (but I can't set it)

Elchi3 (approved):

PR #12414, opened 2021-09-16
Add Chromium versions for api.Window.showModalDialog

queengooborg:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `showModalDialog` member of the `Window` API by mirroring the data.

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -8843,7 +8843,7 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Window/showModalDialog",
           "support": {
             "chrome": {
-              "version_added": true,
+              "version_added": "6",
               "version_removed": "43"
             },
             "chrome_android": {


foolip:
> This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `showModalDialog` member of the `Window` API by mirroring the data.

Only Chrome has changed and that can't be mirrored from any other browser, so how was the version (Chrome 6) determined?

queengooborg:
It did come from mirroring the data from Safari.  Since I still don't have my laptop I couldn't verify the version number myself, so admittedly this isn't the most accurate data.  My thought process was that since this feature is so old and removed a long time ago from Chrome, we could merge this as it is simply to get our real values up (and it shouldn't matter to web developers this far back), and then fix the version number once I get my laptop back.

foolip:
OK, a guesstimate is good enough here.

foolip (approved):

PR #12418, opened 2021-09-16
Add Chromium versions for api.TimeRanges.length

queengooborg:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `length` member of the `TimeRanges` API, based upon commit history and date.

Commit: https://source.chromium.org/chromium/chromium/src/+/f85a45d545671a8a59a97c18c5f313fce3b0c2f7


foolip (approved):
This matches the parent feature `TimeRanges`, which makes sense.

PR #12446, opened 2021-09-17
HTTP WWW-Authenticate - add scheme and algorithm info

### LINKED ISSUE #12370: HTTP authentication header algorithms as subfeatures?
FF93 starts to support SHA256 in HTTP Authorization (https://bugzilla.mozilla.org/show_bug.cgi?id=472823). Previously only md5 could be used, which is deprecated/obsolete. Docs work tracked in https://github.com/mdn/content/issues/8682

This impacts the supported algorithms in [HTTP Authentication](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication) - specifically the algorithms that might be specified in [Authentication](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization) and [WWW-Authenticate](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate)

Anyway, the questions here are: 
- should we add the possible authentication mechanisms AND algorithms for the authentication headers as subfeatures?
- Can we please add specs for for the mechanisms (whether or not we do list them out as subfeatures). For example:
  - https://datatracker.ietf.org/doc/html/rfc7616 - HTTP Digest Authorization
  - https://datatracker.ietf.org/doc/html/rfc7617 - Basi

hamishwillee:
[WWW-Authenticate](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/WWW-Authenticate) and [Authorization](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization) together are the HTTP headers used in the general [HTTP authentication framework](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication). Note, compatibility for these should mirror each other.

Thing is, this is a framework, not a specific implementation. So saying that Firefox or whatever supports this in version X tells you very little. The useful compatibility information is 
- which authentication methods are supported - basic, digest, kerberos etc.
- within some authentication methods, which algorithms are supported.

What I have done here is a draft that attempts to start adding that information. I really don't want to go all the way with this, but I'd like to at least cover the methods supported by Firefox and Chrome, down to algorithm level. Right now this just starts the job for Firefox, since there is not much point me chasing this too far if BCD team disagree with this approach. 

What I know. 
- Firefox supports:  basic, digest (on all platforms), and ntlm and negotiate (kerberos) depending on whether OS support is available ([from here](https://bugzilla.mozilla.org/show_bug.cgi?id=472823#c49))
   - Firefox supports digest authentication with MD5 and SHA256 (from FF93). Not SHA 512.
   - Basic doesn't use any algorithms. 
   - I whether NTLM/negotiate use algorithms that require separate support statements.
- Chrome supports the same set, but it isn't clear what algorithms: https://www.chromium.org/developers/design-documents/http-authentication , and no OS restrictions other than Chrome OS are noted. 
  - Chrome does NOT support SHA256 as of Sept 29 https://bugs.chromium.org/p/chromium/issues/detail?id=1160478 - nothing supports SHA512. 
- In addition to those schemes above [there are more listed here](https://developer.mozilla.org/en-US/docs/Web/HTTP/Authentication#authentication_schemes) - including HOBA, Mutual, Bearer, AWS4-HMAC-SHA256. I propose not to add info about support for those until yet, but people could add that if they knew something was supported.
- Safari itself has no clear statement about what it supports but there is enough info to be sure it does for Basic authentication: https://developer.apple.com/forums/thread/96293 - albeit poorly. No clear info on digest support. If someone had a mac you could test maybe using a method from here https://httpbin.org/#/Auth

Questions
- Is this approach acceptable? Should I keep going?
- The specs are giving me errors. How can I fix this?
- [Authorization](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization) should mirror this. OK to just copy this when we're done? - EDITED. Yes.

Fixes #12370. Actually, it's a draft to help test whether the approach is reasonable.

### DIFF
--- a/http/headers/www-authenticate.json
+++ b/http/headers/www-authenticate.json
@@ -48,6 +48,346 @@
             "standard_track": true,
             "deprecated": false
           }
+        },
+        "Basic": {
+          "__compat": {
+            "description": "<code>Basic</code> authentication",
+            "spec_url": "https://httpwg.org/specs/rfc7617#top",
+            "support": {
+              "chrome": {
+                "version_added": "1"
+              },
+              "chrome_android": {
+                "version_added": true
+              },
+              "edge": {
+                "version_added": "12"
+              },
+              "firefox": {
+                "version_added": "1"
+              },
+              "firefox_android": {
+                "version_added": true
+              },
+              "ie": {
+                "version_added": "1"
+              },
+              "opera": {
+                "version_added": true
+              },
+              "opera_android": {
+                "version_added": true
+              },
+              "safari": {
+                "version_added": null
+              },
+              "safari_ios": {
+                "version_added": null
+              },
+              "samsunginternet_android": {
+                "version_added": true
+              },
+              "webview_android": {
+                "version_added": "37"
+              }
+            },
+            "status": {
+              "experimental": false,
+              "standard_track": true,
+              "deprecated": false
+            }
+          }
+        },
+        "Digest": {
+          "__compat": {
+            "description": "<code>Digest</code> authentication",
+            "spec_url": "https://httpwg.org/specs/rfc7616#top",
+            "support": {
+              "chrome": {
+                "version_added": null
+              },
+              "chrome_android": {
+                "version_added": null
+              },
+              "edge": {
+                "version_added": null
+              },
+              "firefox": {
+                "version_added": true
+              },
+              "firefox_android": {
+                "version_added": true
+              },
+              "ie": {
+                "version_added": null
+              },
+              "opera": {
+                "version_added": null
+              },
+              "opera_android": {
+                "version_added": null
+              },
+              "safari": {
+                "version_added": null
+              },
+              "safari_ios": {
+                "version_added": null
+              },
+              "samsunginternet_android": {
+                "version_added": null
+              },
+              "webview_android": {
+                "version_added": null
+              }
+            },
+            "status": {
+              "experimental": false,
+              "standard_track": true,
+              "deprecated": false
+            }
+          },
+          "SHA-256": {
+            "__compat": {
+              "description": "SHA2-256 digest authentication",
+              "support": {
+                "chrome": {
+                  "version_added": null
+                },
+                "chrome_android": {
+                  "version_added": null
+                },
+                "edge": {
+                  "version_added": null
+                },
+                "firefox": {
+                  "version_added": "93"
+                },
+                "firefox_android": {
+                  "version_added": "93"
+                },
+                "ie": {
+                  "version_added": null
+                },
+                "opera": {
+                  "version_added": null
+                },
+                "opera_android": {
+                  "version_added": null
+                },
+                "safari": {
+                  "version_added": null
+                },
+                "safari_ios": {
+                  "version_added": null
+                },
+                "samsunginternet_android": {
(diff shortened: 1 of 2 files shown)


hamishwillee:
@ddbeck If you could push this to someone that would be great.

ashnicdill:

ddbeck:
> * Is this approach acceptable? Should I keep going?

Yeah, this makes sense. A couple things to keep in mind: `partial_implementation` will probably apply to OS-specific limitations (e.g., some auth supported on Windows but not macOS or what have you). Notes will be needed.

One additional question at the end here, however.

> * The specs are giving me errors. How can I fix this?

The immediate cause of the failures is that URLs are missing fragments, for deep linking to the relevant portion of the spec.

That said, we typically use the https://datatrack.ietf.org/ URLs. And I didn't check whether the linked specs would validate. Also, the URL will need to be in [w3c/browser-specs](https://github.com/w3c/browser-specs) (there's a separate test for that at the end, after you've cleared the schema validation).

> * [Authorization](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization) should mirror this. OK to just copy this when we're done?

I'm not decided on this yet, but why both? On initial reading, I was thinking, should this be on `Authorization` only, as that's the request header? I'm not sure what browsers are obligated to do with `WWW-Authenticate` responses. It might be easier to overlook these features on `WWW-Authenticate` and centralize what we know about browser behavior on `Authorization`. (That said, you're way ahead of me on how this stuff works, so if you say it's relevant to both, I'll defer to that.)

hamishwillee:
@ddbeck OK, I have created https://github.com/w3c/browser-specs/pull/420 to add specs for the https://www.rfc-editor.org/ cases.

Also created https://github.com/w3c/browser-specs/pull/421 for the Microsoft spec for [NTLM authentication](https://docs.microsoft.com/en-us/windows/win32/secauthn/microsoft-ntlm). Let's see how that goes.

http/headers/www-authenticate.json
```diff
@@ -48,6 +48,346 @@
...
+              "experimental": false,
+              "standard_track": true,
+              "deprecated": false
+            }
+          }
+        },
+        "NTLM": {
+          "__compat": {
+            "description": "<code>NTLM</code> authentication",
+            "spec_url": "https://docs.microsoft.com/en-us/openspecs/windows_protocols/ms-ntht/#published-version",
```
ddbeck, line 346: OK, it looks like we need to omit this here. See https://github.com/mdn/browser-compat-data/pull/13383#issuecomment-971825712.

We can still link to this spec from the MDN page if you like (e.g., from the specification section or see also section), but it's probably best to highlight that it's not a standard, exactly.

```suggestion
```

ddbeck (changes requested):

hamishwillee:
@ddbeck Happy days. NTLM spec removed and I have mirrored the WWW-authenticate to the Authorization header (changing the main spec url only).

Should be good to merge. 
After that, I'll update https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Authorization to use the spec_url.

ddbeck (approved):
Thank you! 🎉

PR #12450, opened 2021-09-17
Add Firefox versions for api.IDBTransaction.objectStoreNames

queengooborg:
This PR adds real values for Firefox and Firefox Android for the `objectStoreNames` member of the `IDBTransaction` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.3.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/IDBTransaction/objectStoreNames

### DIFF
--- a/api/IDBTransaction.json
+++ b/api/IDBTransaction.json
@@ -648,10 +648,10 @@
               "version_added": "79"
             },
             "firefox": {
-              "version_added": true
+              "version_added": "10"
             },
             "firefox_android": {
-              "version_added": true
+              "version_added": "22"
             },
             "ie": {
               "version_added": false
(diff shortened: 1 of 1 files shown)


api/IDBTransaction.json
```diff
@@ -663,10 +663,10 @@
               "version_added": "35"
             },
             "safari": {
-              "version_added": true
+              "version_added": "10.1"
```
foolip, line 666: Could this be a case of a false negative due to missing interface object? What does the test do in Safari 9/10?

queengooborg: I looked at the commit history and it agrees with the collector's results: https://github.com/WebKit/WebKit/commit/e2bfdd0c9f5996ade274100781279db0ed00da5f

foolip: Thanks for checking! Indeed `objectStoreNames` was added later than the interface itself.

foolip (approved):

PR #12451, opened 2021-09-17
Add all browsers versions for InputDeviceInfo API

queengooborg:
This PR adds real values for all browsers for the `InputDeviceInfo` API.  This is the interface returned by `MediaDevices.enumerateDevices()`, it doesn't make sense for this interface to have any compatibility differences.  The data was mirrored from `enumerateDevices()` accordingly.


foolip (approved):
https://mdn-bcd-collector.appspot.com/tests/api/MediaDevices/enumerateDevices looks solid so I trust the data. (Didn't check if it was filled by the collector, just hoping.)

PR #12457, opened 2021-09-17
Add Chromium versions for MutationRecord API

queengooborg:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `MutationRecord` API, based upon commit history and date.

Commits: https://source.chromium.org/chromium/chromium/src/+/312007166b154657025bd14c72574888e2779076
https://source.chromium.org/chromium/chromium/src/+/4e5368c3ba5b6cdf4a878fe39888f090d77dc585

### DIFF
--- a/api/MutationRecord.json
+++ b/api/MutationRecord.json
@@ -6,10 +6,10 @@
         "spec_url": "https://dom.spec.whatwg.org/#interface-mutationrecord",
         "support": {
           "chrome": {
-            "version_added": true
+            "version_added": "28"
           },
           "chrome_android": {
-            "version_added": true
+            "version_added": "28"
           },
           "edge": {
             "version_added": "12"
@@ -24,10 +24,10 @@
             "version_added": "11"
           },
           "opera": {
-            "version_added": true
+            "version_added": "15"
           },
           "opera_android": {
-            "version_added": true
+            "version_added": "14"
           },
           "safari": {
             "version_added": "7"
@@ -36,10 +36,10 @@
             "version_added": "7"
           },
           "samsunginternet_android": {
-            "version_added": true
+            "version_added": "1.5"
           },
           "webview_android": {
-            "version_added": true
+            "version_added": "≤37"
           }
         },
         "status": {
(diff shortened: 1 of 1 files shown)


foolip:
This doesn't match https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver/takeRecords#browser_compatibility, what does `takeRecords()` in Chrome 18-27 if this is correct?

queengooborg:
The collector is saying that the `MutationObserver` API wasn't around until Chrome 26.  The commit history I'm looking at seems to agree too: https://source.chromium.org/chromium/chromium/src/+/18551dbaecaa6417390f443739e671cca55ab984

The commit history also indicates that the `takeRecords` member always returned a `MutationRecord` object (https://source.chromium.org/chromium/chromium/src/+/4e5368c3ba5b6cdf4a878fe39888f090d77dc585), but oddly enough the collector also indicates that the `MutationRecord` interface was implemented in Chrome 28.  Considering this, I'll change the version to "26", and submit another PR to correct `MutationObserver`!

foolip (approved):
Thanks for checking out the discrepancy!

I wasn't totally sure about the `takeRecords()` data, but I could confirm Chrome 20 using the updated tests in https://github.com/foolip/mdn-bcd-collector/pull/1511.

foolip:
Oh, wait, actually `takeRecords()` isn't the oldest way to get a `MutationRecord`, which explains why the interface wasn't added in https://source.chromium.org/chromium/chromium/src/+/4e5368c3ba5b6cdf4a878fe39888f090d77dc585.

Before this, `MutationRecord` instances were delivered via the callback. So maybe a custom test is needed here, or source inspection to see if all of the members of `MutationRecord` were present from the beginning.

queengooborg:
I'm unsure what callback are you referring to, could you clarify?

foolip:
That's the first argument to the [`MutationObserver` constructor](https://developer.mozilla.org/en-US/docs/Web/API/MutationObserver/MutationObserver).

foolip (approved):

PR #12459, opened 2021-09-17
Add Firefox versions for api.OfflineAudioContext.length

queengooborg:
This PR adds real values for Firefox and Firefox Android for the `length` member of the `OfflineAudioContext` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.3.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/OfflineAudioContext/length

### DIFF
--- a/api/OfflineAudioContext.json
+++ b/api/OfflineAudioContext.json
@@ -323,10 +323,10 @@
               "version_added": "14"
             },
             "firefox": {
-              "version_added": true
+              "version_added": "49"
             },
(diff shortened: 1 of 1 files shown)


foolip (approved):
Unlike `length` properties of Array-like interfaces, it's not so strange that this was added after the interface itself, since it's a nice-to-have, returning the `length` parameter used in the constructor.

PR #12461, opened 2021-09-17
Update Firefox versions for RTCDataChannel API

queengooborg:
This PR updates and corrects the real values for Firefox and Firefox Android for the `RTCDataChannel` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.3.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/RTCDataChannel

### DIFF
--- a/api/RTCDataChannel.json
+++ b/api/RTCDataChannel.json
@@ -14,26 +14,12 @@
           "edge": {
             "version_added": "79"
           },
-          "firefox": [
-            {
-              "version_added": "24"
-            },
-            {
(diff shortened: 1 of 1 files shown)


queengooborg:
I removed the alternative name data because a) this is an interface that isn't interacted with directly, so alt. names don't particularly matter, and b) it suggests the interface was accessible before `RTCPeerConnection.createDataChannel()`, the only method used to obtain an instance, which doesn't seem right.

PR #12462, opened 2021-09-17
Move the performance attribute to api.performance (in _globals)

foolip:
This was missed in https://github.com/mdn/browser-compat-data/pull/11518.

This is defined using WindowOrWorkerGlobalScope in the spec:
https://w3c.github.io/hr-time/#the-performance-attribute

None of the data was validated or changed, just moved.

### DIFF
--- a/api/_globals/performance.json
+++ b/api/_globals/performance.json
@@ -1,0 +1,107 @@
+{
+  "api": {
+    "performance": {
+      "__compat": {
+        "mdn_url": "https://developer.mozilla.org/docs/Web/API/performance",
+        "spec_url": "https://w3c.github.io/hr-time/#the-performance-attribute",
+        "support": {
+          "chrome": {
+            "version_added": "6"
+          },
+          "chrome_android": {
+            "version_added": true
+          },
+          "deno": {
+            "version_added": "1.0"
+          },
+          "edge": {
+            "version_added": "12"
+          },
+          "firefox": {
+            "version_added": "7"
+          },
+          "firefox_android": {
+            "version_added": "7"
+          },
+          "ie": {
+            "version_added": "9"
+          },
+          "opera": {
+            "version_added": "15"
+          },
+          "opera_android": {
+            "version_added": "14"
+          },
+          "safari": {
+            "version_added": "8"
+          },
+          "safari_ios": {
+            "version_added": "9"
+          },
+          "samsunginternet_android": {
+            "version_added": true
+          },
+          "webview_android": {
+            "version_added": true
+          }
+        },
+        "status": {
+          "experimental": false,
+          "standard_track": true,
+          "deprecated": false
+        }
+      },
+      "worker_support": {
+        "__compat": {
+          "description": "Available in workers",
+          "support": {
+            "chrome": {
+              "version_added": true
+            },
+            "chrome_android": {
+              "version_added": true
+            },
+            "deno": {
+              "version_added": "1.0"
+            },
+            "edge": {
+              "version_added": "≤79"
+            },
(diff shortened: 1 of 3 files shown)


foolip:
@Elchi3 could I tempt you into preparing the content-side change for this, since you already did a whole lot of these and know all the bits to update?

Elchi3:
I'm happy to help with the content work on this one but it runs into the problem that I ran into with `crypto`: The namespace is already occupied. (see https://github.com/mdn/browser-compat-data/pull/11518#issuecomment-906336064)

Here, `performance` also conflicts with the interface: https://developer.mozilla.org/en-US/docs/Web/API/Performance

api/_globals/performance.json
```diff
@@ -0,0 +1,107 @@
+{
+  "api": {
+    "performance": {
+      "__compat": {
+        "mdn_url": "https://developer.mozilla.org/docs/Web/API/performance",
```
Elchi3, line 5: This MDN URL is already taken. Can't move this global there.

foolip:
Oh, I didn't realize this would happen, that MDN URLs are case-insensitive... any ideas for how to resolve this?

Elchi3:
> Oh, I didn't realize this would happen, that MDN URLs are case-insensitive... any ideas for how to resolve this?

No yet. One could have an MDN URL like `performance_global`. Other ideas?

foolip:
Could we disambiguate "Performance (interface)" and "performance (property)" Wikipedia-style? One of them should probably be very minimal and just link to the other.

Elchi3:
I'm not a big fan of disambiguation pages. I think the current Performance (interface) page is great and having a disambiguation in its place wouldn't be ideal, imo. Or where do you propose that such disambiguation would happen?

foolip:
I agree the page with all the `Performance` members is the best place to land. How about https://developer.mozilla.org/en-US/docs/Web/API/Performance stays the same but https://developer.mozilla.org/en-US/docs/Web/API/performance_property is added?

Perhaps more radical would be to just have one page and not use the BCD for `api.performance` at all, since it should match `api.Performance`.

Elchi3:
> I agree the page with all the `Performance` members is the best place to land. How about https://developer.mozilla.org/en-US/docs/Web/API/Performance stays the same but https://developer.mozilla.org/en-US/docs/Web/API/performance_property is added?

I think this is a good idea.

api/_globals/performance.json
Elchi3:
```suggestion
        "mdn_url": "https://developer.mozilla.org/docs/Web/API/performance_property",
```

Elchi3 (changes requested):

ashnicdill:
I never requested this either. I only use my Motorola phone. I think he
somehow copied my sim card and that's how he is controlling everything.
I've changed it before, got a new phone. This is a major invasion of
privacy and this needs to stop. My husband is very smart, knows how to
build apps, and I need to figure out how to make it stop. We have been
separated for months. He has never had permission to use my phone and I do
believe has access to 100% of it. I can't even change some settings. I
bought this new Motorola because he did the same thing to my last phone.
Can someone help me? Please? Also, can I please get some type of proof
where this is coming from because it isn't me!!

foolip:
@Elchi3 do you think this change would be trivial on the MDN side, or is something novel needed for the first some_property page?

Elchi3:
I've exercised the change in https://github.com/mdn/content/pull/9279. I think it could work as proposed.

Elchi3 (approved):
Content work was merged. Lets do this

PR #12463, opened 2021-09-17
css: bg-attachment: fixed not supported on iOS

Epskampie:
`background-attachment: fixed` doesn't work since iOS 13, as you can verify using [this simple test](http://tools.simonsimon.nl/parallax.html) I wrote, or see in these links:

https://bugs.webkit.org/show_bug.cgi?id=196327
https://developer.apple.com/forums/thread/99883

### DIFF
--- a/css/properties/background-attachment.json
+++ b/css/properties/background-attachment.json
@@ -90,9 +90,8 @@
               ],
               "safari_ios": {
                 "version_added": "5",
-                "version_removed": "13",
                 "partial_implementation": true,
-                "notes": "<code>local</code> is recognized but has no effect."
+                "notes": "<code>fixed<code> is recognized but has no effect. See <a href='https://webkit.org/b/219324'>bug 219324</a>."
               },
               "samsunginternet_android": {
                 "version_added": "1.0"
--- a/css/properties/background-attachment.json
+++ b/css/properties/background-attachment.json
@@ -90,6 +90,7 @@
               ],
               "safari_ios": {
                 "version_added": "5",
+                "version_removed": "13",
                 "partial_implementation": true,
                 "notes": "<code>local</code> is recognized but has no effect."
               },


css/properties/background-attachment.json
```diff
@@ -90,6 +90,7 @@
               ],
               "safari_ios": {
                 "version_added": "5",
+                "version_removed": "13",
                 "partial_implementation": true,
                 "notes": "<code>local</code> is recognized but has no effect."
```
ddbeck, line 95: Safari still _says_ that `fixed` is supported. This is significant for the cascade in ways that being wholly unsupported wouldn't be. So I think we need to stick with this being supported, but fix the note.

```suggestion
                "partial_implementation": true,
                "notes": "<code>fixed<code> is recognized but has no effect. See <a href='https://webkit.org/b/219324'>bug 219324</a>."
```

css/properties/background-attachment.json
```diff
@@ -90,6 +90,7 @@
               ],
```
ddbeck, line 90: It's out of range for the diff suggestion, but we could also make a correction for Safari desktop. I suspect the right thing is to have a single object (like Chrome etc.) with `{ "version_added": "3.1" }`. The local notes and misbehavior do not seem to apply to desktop Safari.

ddbeck (changes requested):
Hi @Epskampie and thanks for this PR. I see it's your first to the project—welcome!

I think we're correctly getting the gist of the problem here with the current data, but the notes were improperly copied from `local`. I think we can fix this with the bug link that you mentioned. See my line comments. Thank you again!

css/properties/background-attachment.json
Epskampie: I'll open another PR for that after verifying.

Epskampie:
Sorry @ddbeck  I didn't get notifications for this, so I only saw your remarks now. I've applied them, it seems correct.

ddbeck (approved):
Looks good. Thank you! 🎉

PR #12464, opened 2021-09-17
Node.js has global atob/btoa in version 16

freaktechnik:
#### Summary
Notes availability of global `atob` and `btoa` in node.js version 16.0.0.

https://nodejs.org/api/globals.html#globals_atob_data

Note that the documentation marks it as "legacy" which means it shouldn't be used in new code, but is unlikely to be removed. Not sure if that's worth noting in the compat data.

### DIFF
--- a/api/_globals/atob.json
+++ b/api/_globals/atob.json
@@ -38,6 +38,9 @@
           "ie": {
             "version_added": "10"
           },
+          "nodejs": {
+            "version_added": "16.0.0"
+          },
           "opera": {
             "version_added": "10.5"
           },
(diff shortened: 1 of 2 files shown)


ddbeck (approved):
Thank you, @freaktechnik. Looks good. 👍

> Note that the documentation marks it as "legacy" which means it shouldn't be used in new code, but is unlikely to be removed. Not sure if that's worth noting in the compat data.

Thanks for mentioning this too. I dug into it a little bit. Reading https://github.com/nodejs/node/pull/37529, it appears that was introduced for the purpose of web compatibility. Based on the discussion there, it seems like the legacy note is for the benefit of developers who don't need web compat. That feels a bit out of scope for us, so let's not fuss with mentioning it.

PR #12467, opened 2021-09-17
Add a guideline for when to use `"preview"`

ddbeck:
#### Summary

Adds a guideline to help contributors decide when to use `"preview"` values

#### Related issues

Fixes https://github.com/mdn/browser-compat-data/issues/12344


Elchi3 (approved):
Thanks Daniel! This is what I had in mind with "preview" and I think it is also what we've discussed with other participants that shaped this feature.

PR #12490, opened 2021-09-20
Remove ScrollToOptions dictionary; fix scroll options data

queengooborg:
This PR performs two actions related to the options for `scroll`, `scrollTo`, and `scrollBy`.

First, the `ScrollToOptions` dictionary is deleted, and its properties are converted into subfeatures of the dependent methods.

Second, this updates and corrects the data for these different options.  While trying to fill out the real values for both Firefox and Safari, I noticed a few discrepancies with the Chrome data as well.  This PR fixes them using the following commit/bug data:
- options.left, options.top
  - Chrome: 41 - https://source.chromium.org/chromium/chromium/src/+/8b3ec94a0d488e26d2854f6465480106487090d9
  - Firefox: 36 - https://bugzil.la/1045754
  - Safari: 10.1 (603.1.5) - https://trac.webkit.org/changeset/205505/webkit
- options.behavior
  - Chrome: 35 - https://source.chromium.org/chromium/chromium/src/+/d5703dc81f333467a9fd2b929ec56d1f8c2410ca
  - Firefox: 27 - https://bugzil.la/918345
  - Safari: 14 (610.1.1) - https://trac.webkit.org/browser/webkit/trunk/Source/WebKit/ChangeLog?rev=254790

Additionally, this removes the IE and Edge partial implementation block on `api.Window.scrollBy` describing that only `scrollBy(x, y)` is supported.  This is referring to the lack of support for the `options` parameter, which is already taken care of by the subfeature.


foolip:
I've written http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=9636 to test for support with feature detection. Can you double check data that was based on commit dates using this? It's hard to tell if there are any runtime flags involved here.

ashnicdill:

api/Element.json
```diff
@@ -6883,11 +6883,107 @@
                 "version_added": "45"
               },
               "safari": {
+                "version_added": "14"
```
foolip, line 6886: Tested 13.1 and 14.1 with http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=9636 and it's added in 14.1, so this is probably right.

api/Element.json
```diff
@@ -6883,11 +6883,107 @@
...
                 "version_added": false
               },
+              "opera": {
+                "version_added": "48"
+              },
+              "opera_android": {
+                "version_added": "45"
+              },
+              "safari": {
+                "version_added": "10.1"
```
foolip, line 6934: Confirmed with http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=9636 by testing both 10 and 10.1.

api/Window.json
```diff
@@ -7715,45 +7715,141 @@
...
         },
-        "options_parameter": {
+        "options_behavior_parameter": {
           "__compat": {
-            "description": "<code>options</code> parameter",
+            "description": "<code>options.behavior</code> parameter",
             "support": {
               "chrome": {
-                "version_added": "45"
+                "version_added": "35"
```
foolip, line 7723: This is based on https://source.chromium.org/chromium/chromium/src/+/d5703dc81f333467a9fd2b929ec56d1f8c2410ca but "These behavior changes are protected by a runtime flag" in the commit message.

Based on http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=9636 and some manual tweaking of the test to see the error, I think it was all enabled in Chrome 61:

```suggestion
                "version_added": "61"
```

Can you mirror the same change throughout?

api/Window.json
```diff
@@ -7715,45 +7715,141 @@
...
               "chrome_android": {
-                "version_added": "45"
+                "version_added": "35"
               },
               "edge": {
                 "version_added": "79"
               },
               "firefox": {
-                "version_added": true
+                "version_added": "27"
```
foolip, line 7732: Based on http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=9636 all of these APIs were introduced in Firefox 36:

```suggestion
                "version_added": "36"
```

foolip (changes requested):
Trying to verify revealed some errors. After applying fixes, can you check if all of the entries have the exact same data? I would expect only two variants of the data due to the different versions for `behavior` and `top`/`left` in Safari, but everything else to stay the same across the board.

foolip (approved):

PR #12501, opened 2021-09-20
Mark cross-engine MediaSession APIs as no longer experimental

lukewarlow:
#### Summary

Marks MediaSession APIs (all the ones with at least 2 engines with support) as no longer experimental

Just based on existing support data.

### DIFF
--- a/api/MediaSession.json
+++ b/api/MediaSession.json
@@ -57,7 +57,7 @@
           }
         },
         "status": {
-          "experimental": true,
+          "experimental": false,
           "standard_track": true,
           "deprecated": false
         }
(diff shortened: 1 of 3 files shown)


ashnicdill:
Again, does this mean whatever this is has stopped? My husband is very
smart, knows how to build apps, and I need to figure out how to make it
stop. We have been separated for months. He has never had permission to use
my phone and I do believe has access to 100% of it. I can't even change
some settings. I bought this new Motorola because he did the same thing to
my last phone. Can someone help me? Please?

foolip (approved):
Thanks @lukewarlow!

PR #12505, opened 2021-09-21
Add Intl.getSupportedValuesOf()

hamishwillee:
This adds `Intl.getSupportedValuesOf()` which was added to FF93 in https://bugzilla.mozilla.org/show_bug.cgi?id=1670033.
Docs work tracking in https://github.com/mdn/content/issues/8618

I have tested and it is not in Chrome, which matches the tracking bug: https://bugs.chromium.org/p/v8/issues/detail?id=10743. I can't find evidence of it in Safari either (but have not tested). 

NOTE, the spec this is in is at stage 3: https://github.com/tc39/proposal-intl-enumeration
The URL for the "proposal spec" is not allowed (https://tc39.es/proposal-intl-enumeration/#sec-intl.supportedvaluesof). What I have done is put in the spec URL where the record will eventually go .

Marked experimental as FF is currently only implementation.

### DIFF
--- a/javascript/builtins/intl/Intl.json
+++ b/javascript/builtins/intl/Intl.json
@@ -110,6 +110,61 @@
             }
           }
         },
+        "supportedValuesOf": {
+          "__compat": {
+            "mdn_url": "https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Intl/supportedValuesOf",
+            "spec_url": "https://tc39.es/ecma402/#sec-intl.supportedvaluesof",
+            "support": {
+              "chrome": {
+                "version_added": false
+              },
+              "chrome_android": {
+                "version_added": false
+              },
+              "deno": {
+                "version_added": false
+              },
+              "edge": {
+                "version_added": false
+              },
+              "firefox": {
+                "version_added": "93"
+              },
+              "firefox_android": {
+                "version_added": "93"
+              },
+              "ie": {
+                "version_added": false
+              },
+              "nodejs": {
+                "version_added": false
+              },
+              "opera": {
+                "version_added": false
+              },
+              "opera_android": {
+                "version_added": false
+              },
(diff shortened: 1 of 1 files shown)


ashnicdill:
So does tracking bug mean someone is tracking what I am doing? My husband
is very smart, knows how to build apps, and I need to figure out how to
make it stop. We have been separated for months. He has never had
permission to use my phone and I do believe has access to 100% of it. I
can't even change some settings. I bought this new Motorola because he did
the same thing to my last phone. Can someone help me? Please?

Thank you,
Ashley Dill

javascript/builtins/intl/Intl.json
```diff
@@ -110,6 +110,61 @@
             }
           }
         },
+        "supportedValuesOf": {
+          "__compat": {
+            "mdn_url": "https://developer.mozilla.org/docs/Web/JavaScript/Reference/Global_Objects/Intl/supportedValuesOf",
+            "spec_url": "https://tc39.es/ecma402/#sec-intl.supportedvaluesof",
```
ddbeck, line 116: This is a little fussy, but… would you set this to the proposal URL after all?

The spec is in w3c/browser-specs now, so when dependabot picks up the release later today, I can merge it. Then I can rebase _this_ PR with the updated `package.json`, which will unbreak the tests and everything will be very tidy.

ddbeck (changes requested):
One nitpicky suggestion for you, @hamishwillee. Thank you!

javascript/builtins/intl/Intl.json
hamishwillee:
```suggestion
            "spec_url": "https://tc39.es/proposal-intl-enumeration/#sec-intl.supportedvaluesof",
```
@ddbeck Absolutely! Done - handed over to you now.

PR #12508, opened 2021-09-21
JS class static inititialization blocks

hamishwillee:
FF93 just added support for class static initialization blocks: https://bugzilla.mozilla.org/show_bug.cgi?id=1725689 (actually they were present behind a preference in FF92, but I didn't know about that, and I'm not sure it's useful).

This is present in FF93. It is present in v8 engine [from v9.4](https://v8.dev/blog/v8-release-94) went into Chrome 94 and Deno 1.14. I also tested this in nodejs using the latest NVM I can get v16.9.1 on Ubuntu and that says we're still at v8 engine `9.3.345.16-node.12`.

Test code is just a block of javascript that either dumps output to console or gives me a syntax error:
```js
class C {
  // This block will run when the class itself is evaluated
  static { 
    console.log("C's static block");
  }
}
```

Explainer on this feature here: https://v8.dev/features/class-static-initializer-blocks (MDN docs under construction)

### DIFF
--- a/javascript/classes.json
+++ b/javascript/classes.json
@@ -160,62 +160,6 @@
           "deprecated": false
         }
       },
-      "static_initialization_blocks": {
-        "__compat": {
-          "description": "Class static initialization blocks",
-          "mdn_url": "https://developer.mozilla.org/docs/Web/JavaScript/Reference/Classes/Class_static_initialization_blocks",
-          "spec_url": "https://tc39.es/ecma262/multipage/ecmascript-language-functions-and-classes.html#prod-ClassStaticBlock",
-          "support": {
-            "chrome": {
-              "version_added": "94"
-            },
-            "chrome_android": {
-              "version_added": "94"
-            },
-            "deno": {
-              "version_added": "1.14"
-            },
-            "edge": {
-              "version_added": "94"
-            },
-            "firefox": {
-              "version_added": "93"
-            },
-            "firefox_android": {
-              "version_added": "93"
-            },
-            "ie": {
-              "version_added": false
-            },
-            "nodejs": {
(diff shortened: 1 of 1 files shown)


javascript/classes.json
```diff
@@ -160,6 +160,62 @@
           "deprecated": false
         }
       },
+      "class_static_initialization_blocks": {
```
hamishwillee, line 163: Note the inventive name :-)

hamishwillee:
@ddbeck Would be great to get this reviewed.

javascript/classes.json
ddbeck: Let's omit "class" since this feature is nested under classes anyway. And then it'll sort nicely with the other `static` stuff. (This suggestion will fail the tests, unfortunately—you'll have to `npm run fix` I think)

```suggestion
      "static_initialization_blocks": {
```

javascript/classes.json
```diff
@@ -160,6 +160,62 @@
...
+            },
+            "samsunginternet_android": {
+              "version_added": false
+            },
+            "webview_android": {
+              "version_added": "94"
+            }
+          },
+          "status": {
+            "experimental": true,
```
ddbeck, line 213:
```suggestion
            "experimental": false,
```

ddbeck (changes requested):
Two small suggestions for you, @hamishwillee. Thank you!

hamishwillee:
Thanks @ddbeck - I must cut our meetings short more often from now on :-). Fixed and ready for re-review

ddbeck (approved):
Looks good. Thank you! 🎉

PR #12512, opened 2021-09-21
Add webextensions.onCommand.tab

bershanskiy:
#### Summary

Add compatibility data for WebExtension Commands API callback arguments. Specifically, this adds the second `tab` argument. `onCommand.name.__compat` is taken from `onCommand.__compat`.

[Chromium docs](https://developer.chrome.com/docs/extensions/reference/commands/#event-onCommand) showing "Chrome 86+" next to `tabs`.

### DIFF
--- a/webextensions/api/commands.json
+++ b/webextensions/api/commands.json
@@ -25,6 +25,54 @@
                 "version_added": "14"
               }
             }
+          },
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Thank you!

PR #12518, opened 2021-09-21
Remove irrelevant SVGZoomAndPan API

queengooborg:
This PR removes the irrelevant `SVGZoomAndPan` API as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-features). The lack of current support has been confirmed by the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.3.0), even if the current BCD suggests support.


foolip:
@queengooborg maybe you can make "zoomAndPan" attribute removal into this as well? Those are present, but the properties by the same name aren't.

ddbeck:
> @ddbeck do you think we can remove this half-present and almost-old-enough-for-removal feature?

We're what, a month off and the API is useless? Go ahead. 👍

foolip:
Alright, breaking the rules with abandon!

PR #11700, opened 2021-07-23
Update Firefox and Safari support for `display-capture` feature-policy

lukewarlow:
#### Summary

This PR updates Firefox and Safari support for the `display-capture` feature policy. Also unmarks as experimental as it has at least 2 stable implementations.

#### Test results and supporting details

For Safari it's an educated guess based on the fact Safari implemented getDisplayMedia in Safari 13, and [this commit ](https://github.com/WebKit/WebKit/commit/c06555f8c754ac0ffa5e3a3f8929150f2b81931f) which implements this feature was done in May 2019 which was before Safari 13 release.

I have also verified that it is at the very least implemented in Safari 14.1.

For the Firefox version, I tested with https://plastic-brief-carnation.glitch.me/ on browser stack and version 74 is when it worked as expected. I cannot find any mention of this in any Firefox release notes (I checked from version 67 onwards).

Tested on Firefox 92 and Safari 15 with https://demo.lukewarlow.uk/feature-policy/display-capture/ and can see that it doesn't support the header.

Firefox Android doesn't support get display media, and even when that's enabled (`media.getdisplaymedia.enabled`) it still doesn't actually work. So this feature policy can't be implemented in any meaningful way.

#### Related issues

Fixes #11554

Closes https://github.com/mdn/browser-compat-data/issues/12708

PR #12525, opened 2021-09-21
Set Chrome 94 as current

lukewarlow:
#### Summary

Set Chrome, Chrome Android and Webview Android v94 as current.

https://chromereleases.googleblog.com/2021/09/stable-channel-update-for-desktop_21.html

https://chromereleases.googleblog.com/2021/09/chrome-for-android-update_21.html

https://www.chromestatus.com/features/schedule - for the Chrome 95, and 96 release dates.


lukewarlow:
Related to this MR it would be good if https://github.com/mdn/browser-compat-data/pull/11700 could get in. As it's a potentially impactful change in Chrome 94

queengooborg (approved):

PR #12536, opened 2021-09-22
Update api.Animation.pending as shipped in Chromium 76

foolip:
Chrome 76 is from mdn-bcd-collector results:
http://mdn-bcd-collector.appspot.com/tests/api/Animation/pending

The rest is mirrored, with the assumption that this wasn't in Edge 18.

Also fix the incorrect spec URL.

### DIFF
--- a/api/Animation.json
+++ b/api/Animation.json
@@ -661,16 +661,16 @@
       "pending": {
         "__compat": {
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Animation/pending",
-          "spec_url": "https://drafts.csswg.org/web-animations/#dom-animation-currenttime",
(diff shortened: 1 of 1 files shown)


ddbeck (changes requested):
This looks good, except for one thing: can you change the experimental status while we're here? That should be `false`.

foolip:
> This looks good, except for one thing: can you change the experimental status while we're here? That should be `false`.

I did, but flipped it back because the whole file is wrong and I'd rather it was all changed at once. I can send a separate PR for it though.

foolip:
I've sent https://github.com/mdn/browser-compat-data/pull/12539 for that.

ddbeck (approved):
Great. Thanks for the other PR, @foolip. Let's go ahead with this one now.

Issue #12545, opened 2021-09-22
Feature Policy - Firefox data is all flagged


lukewarlow:
#### What information was incorrect, unhelpful, or incomplete?

Firefox support for "Feature-Policy" header all states that it's only supported behind a flag.

#### What did you expect to see?

Some if not all have been partially supported since Firefox ~~[73](https://bugzilla.mozilla.org/show_bug.cgi?id=1600883)~~ [74](https://bugzilla.mozilla.org/show_bug.cgi?id=1617219). Only working inside the allow attribute.

#### Did you test this? If so, how?

I've validated that for the "display-capture" feature policy it definitely works as of at least version 74.

#### Can you link to any release notes, bugs, pull requests, or MDN pages related to this?

https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Feature-Policy

lukewarlow:
Follow up issue after noticing more than just display capture was wrong when doing https://github.com/mdn/browser-compat-data/pull/11700

lukewarlow:
This data was originally added as a full support in 74. But was later reverted in https://github.com/mdn/browser-compat-data/pull/9280. It should have been changed to only partial implementations with a note (like has been done for Safari).

lukewarlow:
Behind the `dom.security.featurePolicy.experimental.enabled` flag, the following policies are supported. Not sure on exact version they were added yet.

`autoplay`
`document-domain`
`encrypted-media`
`midi`
`payment`
`xr-spatial-tracking`

queengooborg:
It appears that the data has been fixed at least since the rename to `Permissions-Policy`, so I'm going to close this issue.  If I'm missing anything, please let me know.

PR #12548, opened 2021-09-22
Update Firefox support for feature-policy

lukewarlow:
#### Summary

This updates Firefox support data for the feature policy functionality.

Tested on latest Firefox using https://demo.lukewarlow.uk/feature-policy/

https://bugzilla.mozilla.org/show_bug.cgi?id=1617219 - Shows it was enabled in 74

https://bugzilla.mozilla.org/show_bug.cgi?id=1483631 - Is what I've used to work out the initial values that were supported. Though I suspect others beyond the media and geolocation values were also supported.

`speaker-selection` and `gamepad` was tested on latest version.

#### Related issues

https://github.com/mdn/browser-compat-data/issues/12545 - The experimental features mentioned in a comment on this issue, I haven't added because I'm not sure how best to track their supported versions, or how valuable the data actually is.

### DIFF
--- a/http/headers/feature-policy.json
+++ b/http/headers/feature-policy.json
@@ -1446,64 +1446,30 @@
               "edge": {
                 "version_added": false
               },
-              "firefox": [
-                {
-                  "version_added": "92",
-                  "partial_implementation": true,
-                  "notes": "Only supported through the <code>allow</code> attribute on <code>&lt;iframe&gt;</code> elements.",
-                  "flags": [
-                    {
-                      "type": "preference",
-                      "name": "media.setsinkid.enabled",
-                      "value_to_set": "true"
-                    }
-                  ]
-                },
-                {
-                  "version_added": "92",
-                  "flags": [
-                    {
-                      "type": "preference",
-                      "name": "dom.security.featurePolicy.header.enabled",
-                      "value_to_set": "true"
-                    },
-                    {
-                      "type": "preference",
-                      "name": "media.setsinkid.enabled",
-                      "value_to_set": "true"
-                    }
-                  ]
-                }
-              ],
-              "firefox_android": [
-                {
-                  "version_added": "92",
-                  "partial_implementation": true,
-                  "notes": "Only supported through the <code>allow</code> attribute on <code>&lt;iframe&gt;</code> elements.",
-                  "flags": [
-                    {
-                      "type": "preference",
-                      "name": "media.setsinkid.enabled",
-                      "value_to_set": "true"
-                    }
-                  ]
-                },
-                {
-                  "version_added": "92",
-                  "flags": [
-                    {
-                      "type": "preference",
-                      "name": "dom.security.featurePolicy.header.enabled",
-                      "value_to_set": "true"
-                    },
-                    {
-                      "type": "preference",
-                      "name": "media.setsinkid.enabled",
-                      "value_to_set": "true"
-                    }
-                  ]
-                }
-              ],
+              "firefox": {
+                "version_added": "92",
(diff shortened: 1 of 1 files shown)


lukewarlow:
https://github.com/mdn/browser-compat-data/pull/11700 - handles the `display-capture` feature policy

lukewarlow:
Behind the `dom.security.featurePolicy.experimental.enabled` flag, the following policies are supported. Not sure on exact version they were added yet.

autoplay
document-domain
encrypted-media
midi
payment
xr-spatial-tracking

http/headers/feature-policy.json
```diff
@@ -1333,12 +1432,64 @@
...
+                  "partial_implementation": true,
+                  "notes": "Only supported through the <code>allow</code> attribute on <code>&lt;iframe&gt;</code> elements.",
+                  "flags": [
+                    {
+                      "type": "preference",
+                      "name": "media.setsinkid.enabled",
+                      "value_to_set": "true"
+                    }
+                  ]
+                },
```
lukewarlow, line 1447: I'm not sure if this partial implementation entry should exist until the `media.setsinkid.enabled` flag is enabled by default in Firefox? While this feature policy is recognised by Firefox without this flag, it seems pointless to mark as supported without it having any noticeable effect.

http/headers/feature-policy.json
```diff
@@ -1683,28 +1834,41 @@
...
-                    "name": "dom.security.featurePolicy.header.enabled",
-                    "value_to_set": "true"
-                  }
-                ]
-              },
+              "firefox": [
+                {
+                  "version_added": "81",
+                  "partial_implementation": true,
+                  "notes": "Only supported through the <code>allow</code> attribute on <code>&lt;iframe&gt;</code> elements. Firefox recognizes the <code>web-share</code> permissions policy, but this has no effect in versions of Firefox that do not support the <a href='https://developer.mozilla.org/docs/Web/API/Navigator/share'><code>share()</code></a> method."
```
lukewarlow, line 1841: I don't know if it's actually valuable to show that it supports this feature policy when no desktop Firefox versions support web share? (according to the data, and not macOS from my testing).

lukewarlow: https://github.com/mdn/browser-compat-data/pull/12194 - Might be relevant here

http/headers/feature-policy.json
ddbeck: I think the way you've structured this is a little unusual. Partial + flags isn't terribly common, but it comes up from time to time. What I'm confused by is how it relates to the support statement immediately following. Since neither statement has a `version_removed`, it seems to say that it's partially supported with only one flag set and fully supported with both?

I think the right move here might be one support statement with the note and partial from above and the array of flags from below. Does that make sense?

ddbeck (changes requested):

http/headers/feature-policy.json
lukewarlow: The problem it's trying to solve is that the feature policy is supported in header form with two flags enabled. But only on the iframe (see notes) without one of the extra flags set. So I'm not sure the one entry would make much sense. I'm thinking maybe just keep the top entry and remove the bottom one, as the second flag isn't overly useful?

ddbeck: > It would seem that Firefox don't intend to enable feature policy header (it's been superseded by permissions policy) hence my thinking that the flag might not be too useful.

Yes, that's a strong point in favor of not mentioning the other case. I'd drop the second statement, in that case.

ddbeck (approved):
Thank you! 🎉

PR #12552, opened 2021-09-23
Remove Window.minimize/maximize

teoli2003:
They were XUL-only.

We removed them in mdn/content: mdn/content#9175 (the other was not documented)


ddbeck (approved):
Looks good. Thank you! 🎉

PR #1212, opened 2018-02-27
adding compat data for MediaStream

mec:
https://developer.mozilla.org/en-US/docs/Web/API/MediaStream

PR #12556, opened 2021-09-23
Fix support for MediaStream.id and MediaStreamTrack.id

foolip:
Current support in Chrome 94 can be easily confirmed:
https://mdn-bcd-collector.appspot.com/tests/api/MediaStream/id
https://mdn-bcd-collector.appspot.com/tests/api/MediaStreamTrack/id

These were both introduced in WebKit trunk version 537.26:
https://trac.webkit.org/changeset/139352/webkit
https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Configurations/Version.xcconfig?rev=139352

That maps to Chrome 26, matching the existing `version_added`.

The original data here came from wiki migration:
https://github.com/mdn/browser-compat-data/pull/1212

The data for api.MediaStreamTrack also has to be updated to pass the
lint. Support for MediaStreamTrack in Chrome 26 can be confirmed with
this test:
http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=9656

The interface object wasn't exposed in Chrome 26-28, which is the most
likely explanation for why support was set to 29 in this PR:
https://github.com/mdn/browser-compat-data/pull/5844

Don't both with `partial_implementation` and notes for this, since
there's no constructor and it doesn't really matter if the interface
object is exposed or not.

### DIFF
--- a/api/MediaStream.json
+++ b/api/MediaStream.json
@@ -586,16 +586,13 @@
           "spec_url": "https://w3c.github.io/mediacapture-main/#dom-mediastream-id",
           "support": {
             "chrome": {
-              "version_added": "26",
-              "version_removed": "54"
+              "version_added": "26"
             },
             "chrome_android": {
-              "version_added": "26",
-              "version_removed": "54"
+              "version_added": "26"
             },
             "edge": {
-              "version_added": "12",
-              "version_removed": "79"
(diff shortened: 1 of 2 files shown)


queengooborg (approved):
Thanks for digging into this, looking good to me!

PR #12561, opened 2021-09-24
Adds spec_url for Document.clear()

teoli2003:
HTML5 actually defines Document.clear()… as doing nothing.


queengooborg (approved):
Since it's supposed to do nothing, this is probably a special-case candidate for removal.  Something to consider in the future though.  Thanks for the PR!

PR #12565, opened 2021-09-24
Remove irrelevant MouseWheelEvent API

queengooborg:
This PR removes the irrelevant `MouseWheelEvent` API as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-features). The lack of current support has been confirmed by the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.3.0), even if the current BCD suggests support.  I've also manually confirmed a lack of support by checking Chrome's and Safari's sources, and checked Edge for compatibility.


foolip:
Was this in Edge 18? If so it's still too early to remove per guideline.

queengooborg:
It was not in Edge 18, I did check that!

foolip:
Excellent!

foolip (approved):
Not merging, please land the MDN change first.

hamishwillee:
What about the associated event https://developer.mozilla.org/en-US/docs/Web/API/GlobalEventHandlers/onmousewheel ?

Even though we redirect, could you clean up some of the internal links? e.g. 
- https://developer.mozilla.org/en-US/docs/Web/API/WheelEvent

might be simple enough just to turn them from links to code-only.

queengooborg:
I've opened a follow-up PR to the content repo!  I'll go ahead and merge this since the main removal is complete.

ddbeck:
> I've opened a follow-up PR to the content repo! I'll go ahead and merge this since the main removal is complete.

@queengooborg could you link to the corresponding content PR? It's not clear to me that the needs-content-update label was satisfied before merging.

PR #12577, opened 2021-09-25
Add Chromium/Edge versions for PushSubscriptionOptions API

queengooborg:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) and Edge for the `PushSubscriptionOptions` API, while also correcting the versions for `PushSubscription.options`, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.3.0).

### DIFF
--- a/api/PushSubscription.json
+++ b/api/PushSubscription.json
@@ -206,13 +206,13 @@
           "spec_url": "https://w3c.github.io/push-api/#dom-pushsubscription-options",
           "support": {
             "chrome": {
-              "version_added": "42"
+              "version_added": "54"
             },
             "chrome_android": {
-              "version_added": "42"
+              "version_added": "54"
             },
             "edge": {
-              "version_added": "16"
+              "version_added": "17"
             },
             "firefox": {
               "version_added": "44",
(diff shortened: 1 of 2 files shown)


foolip (changes requested):
This data should likely match https://developer.mozilla.org/en-US/docs/Web/API/PushSubscription/options#browser_compatibility, it's not possible for `myPushSubscription.options` to be supported without `PushSubscriptionOptions`.

queengooborg:
It appears that the collector is also indicating that `api.PushSubscription.options` was also added in Chrome 54, rather than Chrome 42 as currently indicated.  The original data comes from the wiki migration so I trust the collector results more.  The [commit history](https://storage.googleapis.com/chromium-find-releases-static/888.html#8882a829851f9bfe6e1857c9d8b0a7c679d89c00) also backs this up.

I'll also fix `api.PushSubscription.options` in this PR!

foolip:
@queengooborg can you also check if other related APIs have the same error and fix them up at the same time?

queengooborg:
The rest of the Push-related APIs are looking good to me!  The collector _is_ changing them up, but it's mostly changes like Chrome 42 -> 43, all-false, etc. which I don't trust without some custom tests.

foolip (approved):

Issue #12574, opened 2021-09-25
api.Crypto.randomUUID - Change usage version for Node.js


beeloon:
#### What information was incorrect, unhelpful, or incomplete?
Version in which crypto.randomUUID, can be used with Node.js.

#### What did you expect to see?
Changing version from v.16.7.0 to  v14.17.0. 

#### Did you test this? If so, how?
This feature works fine on v14.17.0. 
Here's official doc [link](https://nodejs.org/docs/latest-v14.x/api/crypto.html#crypto_crypto_randomuuid_options)

<details>
<summary>MDN page report details</summary>

* Query: `api.Crypto.randomUUID`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/Crypto/randomUUID
* Report started: 2021-09-25T07:48:57.844Z

</details>

queengooborg:
Closing since #12578 covers this!

PR #12578, opened 2021-09-25
Update Node.js version for Crypto.randomUUID API(mdn#12574)

beeloon:
#### Summary

Fix for Node.js  version for crypto.randomUUID.

Function randomUUID was tested on v.14.7.0, here's results
Screenshot from 2021-09-25 11-21-15

#### Related issues

Issue #12574.

### DIFF
--- a/api/Crypto.json
+++ b/api/Crypto.json
@@ -147,7 +147,8 @@
               "version_added": false
             },
             "nodejs": {
-              "version_added": "14.17.0"
+              "version_added": "16.7.0",
+              "notes": "From version 14.17.0, a similar method is available as <a href='https://nodejs.org/docs/latest-v14.x/api/crypto.html#crypto_crypto'><code>crypto.randomUUID()</code></a>."
             },
             "opera": {
               "version_added": false
--- a/api/Crypto.json
+++ b/api/Crypto.json
@@ -147,7 +147,7 @@
               "version_added": false
             },
             "nodejs": {
-              "version_added": "16.7.0"
+              "version_added": "14.17.0"
             },
             "opera": {
               "version_added": false


lukewarlow:
Need to add v14.17.0 to the nodejs.json file. And update the Crypto.json file to include the new version on the base feature aswell.

lukewarlow:
However, it's not actually available on the global like it is in node 16.7+. And also isn't available on a webcrypto export from the crypto module, like the current note on the base feature says for node 15. So I'm not sure if it's actually correct to add this to MDN.

As this is a node specific implementation of a thing rather than being a web API?

beeloon:
@lukewarlow 
thanks for mention this :rocket: 
at first i thought about availability of using the feature on concrete version
but did not understand the specification as well as you
so you might be right, and we don't need to add this info :bow:

lukewarlow:
Definitely worth waiting to see what one of the maintainers has to say I might be completely wrong :)

ddbeck (dismissed):
Thanks for this PR, @beeloon. I see it's your first to the project—welcome!

(And thank you for the initial review, @lukewarlow. It really sped up sorting this one out.)

After looking through the docs, I think we ought to treat Node's `crypto.randomUUID()` method as a separate API from `webcrypto.Crypto.randomUUID()`, though they happen to share a method name (you might've noticed that Node's version supports an options object that the standard Web Crypto API doesn't). This is a really fine distinction, so your proposed change was reasonable, but a little difficult to reconcile with the features above.

All that said, we **can** note the availability of Node's API, but I don't think we should change the version number here. I've left a line comment to suggest a path forward.

(There's also a problem with the way this whole interface is represented without an `alternative_name`. I'll open a separate PR to address that.)

Again, thank you!

ddbeck:
I'd still like to see the suggestion applied, so I'm going to apply it and merge this anyway. Thank you, @beeloon!

beeloon:
@ddbeck Thanks :bow:

Issue #12340, opened 2021-09-10
Fetch APIs - Inconsistent with CanIUse


queengooborg:
As brought up in https://github.com/mdn/browser-compat-data/pull/12291#issuecomment-916774685, it seems that our data is inconsistent with CanIUse.  https://caniuse.com/fetch  We should update our data to maintain consistency.

foolip:
In particular, caniuse.com has a note that says "Only available in Chrome and Opera within ServiceWorkers" for Chrome 40-41. This is probably correct, and we should figure out which APIs are supported in service worker scope in Chrome 40-41 and use `partial_implementation` there.

The collector probably can't answer these questions now, because these API will tend to now use `[Exposed=(Window,Worker)]`, which means we'll generate tests for dedicated workers, but we need to detect support in service workers specifically, which we don't generate tests for.

PR #12579, opened 2021-09-25
Update all browsers versions for Request + Response APIs

queengooborg:
This PR updates real values for all browsers for the `Request` and `Response` APIs, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v3.3.0).

Tests Used:
https://mdn-bcd-collector.appspot.com/tests/api/Request
https://mdn-bcd-collector.appspot.com/tests/api/Response

### DIFF
--- a/api/Request.json
+++ b/api/Request.json
@@ -13,7 +13,7 @@
               "version_added": "40",
               "version_removed": "42",
               "partial_implementation": true,
-              "notes": "Only available via Service Workers."
+              "notes": "Only available via service workers."
             }
           ],
           "chrome_android": [
@@ -24,7 +24,7 @@
               "version_added": "40",
               "version_removed": "42",
               "partial_implementation": true,
-              "notes": "Only available via Service Workers."
+              "notes": "Only available via service workers."
             }
           ],
           "deno": {
@@ -50,7 +50,7 @@
               "version_added": "27",
               "version_removed": "29",
               "partial_implementation": true,
-              "notes": "Only available via Service Workers."
+              "notes": "Only available via service workers."
             }
           ],
           "opera_android": [
@@ -61,7 +61,7 @@
               "version_added": "27",
               "version_removed": "29",
               "partial_implementation": true,
-              "notes": "Only available via Service Workers."
+              "notes": "Only available via service workers."
             }
           ],
           "safari": {
@@ -81,7 +81,7 @@
               "version_added": "40",
               "version_removed": "42",
               "partial_implementation": true,
-              "notes": "Only available via Service Workers."
+              "notes": "Only available via service workers."
             }
           ]
         },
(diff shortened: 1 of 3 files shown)


api/Request.json
```diff
@@ -6,16 +6,16 @@
         "spec_url": "https://fetch.spec.whatwg.org/#request-class",
         "support": {
           "chrome": {
-            "version_added": "42"
+            "version_added": "40"
```
foolip, line 9: This doesn't match the data for `fetch()` for reasons covered in https://github.com/mdn/browser-compat-data/issues/12340.

It would be great to keep things consistent, with 40 everywhere or 42 everywhere.

foolip (changes requested):
If you want to split out the changes that just fix true values, that might speed up progress on that.

api/Request.json
queengooborg: Ah right, I forgot about that issue -- I think that having "40" everywhere makes more sense and tells a more accurate story, maybe with a note (or possibly a partial_implementation range) stating it's only available in service workers?

api/Request.json
```diff
@@ -5,17 +5,33 @@
...
-          },
+          "chrome": [
+            {
+              "version_added": "42"
+            },
+            {
+              "version_added": "40",
+              "version_removed": "42",
+              "partial_implementation": true,
+              "notes": "Only available via Service Workers."
```
foolip, line 16:
```suggestion
              "notes": "Only available in service workers"
```

This is to match the style of "worker_support" descriptions and the capitalization of "service workers" in the prose of https://developer.mozilla.org/en-US/docs/Web/API/Service_Worker_API

Can you make the same change everywhere?

foolip (changes requested):

api/Response.json
```diff
@@ -264,10 +304,10 @@
           "spec_url": "https://fetch.spec.whatwg.org/#ref-for-dom-response-error①",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": "43"
```
foolip, line 307: This is suspect, a case of https://github.com/mdn/browser-compat-data/issues/7843?

api/_mixins/LinkStyle__ProcessingInstruction.json
```diff
@@ -13,7 +13,7 @@
               "version_added": "18"
             },
             "edge": {
-              "version_added": "12"
+              "version_added": "79"
```
foolip, line 16: The changes to this file should be reverted.

foolip (changes requested):

api/Response.json
queengooborg: It does look strange, but we have custom tests for these APIs, so this is accurate.

foolip: Aha, sorry I didn't check the tests.

foolip (approved):

PR #12591, opened 2021-09-26
Update support data for the datalist element

lukewarlow:
#### Summary

Added Opera Android support, also change Firefox Android support entry to better reflect that it's a partial implementation for v79+.

Tested latest Opera Android and the dropdown does show. Assuming it was added to the release based on Chrome 33, based on Chrome Android support data.

Also confirmed Firefox Android still doesn't render a dropdown (or other UI) for the options.

### DIFF
--- a/html/elements/datalist.json
+++ b/html/elements/datalist.json
@@ -18,10 +18,17 @@
             "firefox": {
               "version_added": "4"
             },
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
Thanks! 👍

PR #12600, opened 2021-09-27
Change Chrome's SameSite cookie version

adamtaylor13:
#### Summary

The Chromium Project says that Chrome versions 51 to 66 (inclusive) will outright reject a `SameSite=None` cookie.

Reference: https://www.chromium.org/updates/same-site/incompatible-clients

### DIFF
--- a/http/headers/set-cookie.json
+++ b/http/headers/set-cookie.json
@@ -335,13 +335,13 @@
                 ],
                 "chrome_android": [
                   {
-                      "version_added": "67"
+                    "version_added": "67"
                   },
                   {
-                      "version_added": "51",
-                      "version_removed": "67",
-                      "partial_implementation": true,
-                      "notes": "Rejects cookies with <code>SameSite=None</code>. See <a href='https://www.chromium.org/updates/same-site/incompatible-clients'>SameSite=None: Known Incompatible Clients</a>."
+                    "version_added": "51",
+                    "version_removed": "67",
+                    "partial_implementation": true,
+                    "notes": "Rejects cookies with <code>SameSite=None</code>. See <a href='https://www.chromium.org/updates/same-site/incompatible-clients'>SameSite=None: Known Incompatible Clients</a>."
                   }
                 ],
                 "edge": {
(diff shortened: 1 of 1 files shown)


lukewarlow:
https://www.chromium.org/updates/same-site/incompatible-clients#:~:text=This%20also%20affects%20older%20versions%20of%20Chromium-derived%20browsers%2C%20as%20well%20as%20Android%20WebView. States that it affects all chromium derivates. So Chrome Android, Edge, Opera, Opera Android, Samsung Internet and WebView Android all need updating too.

lukewarlow:
I would also somewhat copy how Safari is treated. 

Keep the old version mark it as partial and add a note with a short description and link to that reference.

And then add a new one that's no longer partial from 67 onwards (same for other Chromium browsers)

adamtaylor13:
> States that it affects all chromium derivates.

Ahh, I didn't fully understand the implications of that statement, but that makes sense. I'll add this in!

adamtaylor13:
So sorry for the delay! Yep, I'm still keen on seeing this through. So I committed the suggestion by @queengooborg above—should I do the same general process (adding the "notes" property) for all the Chromium derivates listed by @lukewarlow ?

ddbeck (changes requested):
I think @lukewarlow had the right idea, to follow what was done for Safari.

I think it would look something like this for Chrome:

```json
[
  {
    "version_added": "67"
  },
  {
    "version_added": "51",
    "version_removed": "67",
    "notes": "Rejects cookies with <code>SameSite=None</code>. See <a href='https://www.chromium.org/updates/same-site/incompatible-clients'>SameSite=None: Known Incompatible Clients</a>."
  }
]
```

I would also like to see the data extended to the Chromium-derivatives. This would map 67 as Luke has already provided. 51 would map as follows:

- Chrome for Android: 51
- Opera: 51
- Opera Android: 41
- Samsung Internet: 5.4
- WebView: 51

(Edge wouldn't be affected.)

Thank you!

adamtaylor13:
@ddbeck Okay, I believe I understood the directions, but please double-check me in this [most-recent commit](https://github.com/mdn/browser-compat-data/pull/12600/commits/0d4ed21af238bf4603ff42fe46c2bfe2700e209e). I took the list you gave, and replaced the "version_added" property with the version you specified in the list of Chromium-derivatives.

Does that look right to you?

adamtaylor13:
Whoops, missed a trailing comma there. Perhaps view _this most recent commit_ instead: [commit](https://github.com/mdn/browser-compat-data/pull/12600/commits/e7675a246f28616f5f66b62848fd98b8b894dc58)

ddbeck (approved):
Well done, @adamtaylor13! This was a bit complicated and I didn't explain it particularly well—I'm glad you figured it out. Looks good. Thank you! 🎉

PR #12602, opened 2021-09-27
css: "start", "end" values in flex context

ZerdoX-x:
#### Summary
Removing wrong & adding support data: "start" and "end" values in flex context (align-self, align-items, align-content, justify-content)

Updated browsers:
Safari, safari mobile, opera mobile, samsung internet – Incorrect support data removed
Opera 79 – Support data added

Tested by me on Linux in WebKitGTK 2.32.4 and my friends on OSX and IOS (different safari versions)
Tested in Opera 79.0.4143.50
Tested in Opera Mobile 65.1.3381.61266
Tested in Samsung Internet 15

`align-self: end;` demo: https://codepen.io/zerdox-x/pen/rNwQXRe
`align-items: end;` demo: https://codepen.io/zerdox-x/pen/MWoxXdK
`align-content: end;` demo: https://codepen.io/zerdox-x/pen/powYZoj
`justify-content: end;` demo: https://codepen.io/zerdox-x/pen/JjJeVyR


lukewarlow:
The support data for some of the other Chromium based browsers is incorrect too. Specifically Opera, Opera Android and Samsung Internet.

ZerdoX-x:
Thank you. So this should be ok now. Feel free to report info if you'll find more issues with this.

ZerdoX-x:
Let me know if something holding back from reviewing & merging

lukewarlow:
Opera 79 correlates to Chrome 93 which is what the existing support data says so I'm assuming it's correct.

Not sure if the notes should be kept for the version_added false entries (Samsung and Opera Android) but one of the maintainers will let you know if that's an issue when they review.

ZerdoX-x:
So can someone review this too please?

css/properties/align-content.json
```diff
@@ -387,12 +387,10 @@
                   "version_added": false
                 },
                 "opera": {
-                  "version_added": false,
-                  "notes": "This value is recognized, but has no effect."
```
ddbeck, line 391: Unfortunately, these notes aren't structured properly since this affects the cascade, it should be listed as `partial_implementation`. Something like this breaks in versions ???-79:

```
  align-content: space-evenly;
  align-content: end;
```

Fixing that is quite the effort, however, since we don't know _when_ the the recognition began. Instead, let's work around this by putting what we do into a note, how about this:

> Before version X, this value is recognized, but has no effect.

where X is the `version_added` value.

Then we'll be able to reconstruct the right structure later, after some testing. (Or, if you're interested in doing that testing, I can suggest an alternative structure).

This would apply to all of the notes like this in the PR.

ddbeck (changes requested):
Hi @ZerdoX-x and thanks for this PR. I see it's your first to the project—welcome! I do think we need to do something about the old notes. I've left a line comment to try to explain. Let me know if there are any questions. Thanks again!

ZerdoX-x:
@ddbeck here you are

ddbeck (approved):
Thank you! 🎉

PR #12615, opened 2021-09-28
Fix the order of RTCPeerConnectionIceEvent entries

foolip:
Currently it's red on MDN because the removed ranges is first:
https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnectionIceEvent


queengooborg (approved):
Good catch, thanks!

PR #12635, opened 2021-09-29
URLPattern is stabilized in Chrome/Edge 95

lucacasonato:
URLPattern is unflagged in Chrome 95, which is currently in beta. See
https://www.chromestatus.com/feature/5731920199548928

### DIFF
--- a/api/URLPattern.json
+++ b/api/URLPattern.json
@@ -4,26 +4,38 @@
       "__compat": {
         "spec_url": "https://wicg.github.io/urlpattern/#urlpattern",
         "support": {
-          "chrome": {
-            "version_added": "93",
-            "flags": [
-              {
-                "type": "preference",
-                "name": "#enable-experimental-web-platform-features",
-                "value_to_set": "Enabled"
-              }
-            ]
-          },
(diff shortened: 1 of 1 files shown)


lukewarlow:
Should the support data for the other entries in that file actually say 1.15 instead of 1.14? Or at least have 1.14 be flagged. (Same goes for Chrome/Edge 93 support)

lucacasonato:
I don't know what the precedent here is. Is there a style guide? Does the flag data need to be present on all entries, or just the top level one?

cc @ddbeck

lukewarlow:
Currently it's showing as supported on MDN which doesn't correctly represent the status of the API. So I would assume flag data on all entries, or just on the top and false / only the release version on the others.

lucacasonato:
@lukewarlow I have removed flag info entirely now as it is irrelevant to most users.

queengooborg:
Unfortunately per BCD data guidelines, we cannot remove data regarding support behind a flag until it has been enabled by default (or the flag has been removed) over two years ago.  Flag data _is_ supposed to copy down to children, can we do that instead?

PR #12661, opened 2021-09-30
Add Chrome ≤15 and Firefox ≤4 as allowed ranges

### LINKED ISSUE #12626: Add ranges for Chrome and Firefox
During my weekly discussion with @foolip, we were talking about adding ranges for Chrome for the earliest versions available in BrowserStack and SauceLabs.  I think it would make sense to do the same for Firefox as well.

This issue is to propose that we add Chrome ≤15, Chrome ≤26, and Firefox ≤4 as allowed ranges.

queengooborg:
This PR fixes #12626 by adding ranges for the earliest Chrome and Firefox versions supported by BrowserStack.  This should help us get to 100% real values quicker, and these browser versions are so old that most consumers will not care about them.

### DIFF
--- a/test/linter/test-versions.js
+++ b/test/linter/test-versions.js
@@ -16,7 +16,9 @@ const validBrowserVersions = {};
 
 /** @type {Object<string, string[]>} */
 const VERSION_RANGE_BROWSERS = {
+  chrome: ['≤15'],
   edge: ['≤18', '≤79'],
+  firefox: ['≤4'],
   ie: ['≤6', '≤11'],
   opera: ['≤12.1', '≤15'],
   opera_android: ['≤12.1', '≤14'],


ddbeck (changes requested):
This changes need to be documented in https://github.com/mdn/browser-compat-data/blob/main/schemas/compat-data-schema.md#ranged-versions. Thanks!

schemas/compat-data-schema.md
```diff
@@ -267,6 +267,10 @@ The following table indicates initial versions for browsers in BCD. These are th
 
 For certain browsers, ranged versions are allowed as it is sometimes impractical to find out in which early version of a browser a feature shipped. Ranged versions should be used sparingly and only when it is impossible to find out the version number a feature initially shipped in. The following ranged version values are allowed:
 
+- Chrome
+  - "≤15" (the earliest Chrome version testable in BrowserStack and possibly earlier)
+- Firefox
+  - "≤4" (the earliest Firefox version testable in BrowserStack and possibly earlier)
```
foolip, line 273: BrowserStack actually has Firefox 3 on Windows XP and Windows 7:

```suggestion
  - "≤3" (the earliest Firefox version testable in BrowserStack and possibly earlier)
```

foolip (changes requested):

PR #12671, opened 2021-10-01
self.structuredClone() added to FF94

hamishwillee:
- Support added for `structuredClone()` in FF94: https://bugzilla.mozilla.org/show_bug.cgi?id=1722576
- Docs tracked in https://github.com/mdn/content/issues/9370

This is not present on chromium yet - verified using
```
const original = { name: "MDN" };
const clone = structuredClone(original);
```
(returns error about function not being defined).

### DIFF
--- a/api/_globals/structuredClone.json
+++ b/api/_globals/structuredClone.json
@@ -30,10 +30,10 @@
             "version_added": false
           },
           "firefox": {
-            "version_added": false
+            "version_added": "94"
           },
           "firefox_android": {
-            "version_added": false
+            "version_added": "94"
           },
           "ie": {
             "version_added": false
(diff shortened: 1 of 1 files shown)


api/_globals/structuredClone.json
```diff
@@ -61,7 +61,7 @@
           }
         },
         "status": {
-          "experimental": true,
+          "experimental": false,
```
hamishwillee, line 64: NOTE, changed this to false now is on 2 platforms - OK? 

BUT , I did NOT change the worker support setting experimental setting though, because deno does not support this in workers according to existing setup, untested). Feels wrong.

ddbeck: Worker support in Deno is `true`. Let's set 'em both to `experimental: false`

ddbeck (changes requested):

api/_globals/structuredClone.json
hamishwillee: Dunno how I missed that. Thanks. Fixed.

Issue #12663, opened 2021-09-30
getVersionAdded function in test-consistency.js - doesn't handle `preview` versions in array syntax. 


lukewarlow:
The linter doesn't like a preview version in array syntax.

```json
 [
    {
      "version_added": "preview"
    },
    {
      "version_added": "10.1",
      "version_removed": "11"
    }
  ],
```

PR #12673, opened 2021-10-01
Fix preview consistency check

ddbeck:
#### Summary

@hamishwillee got an unexpected consistency failure with `"preview"` values. This PR adds some details to remove it.

See https://github.com/mdn/browser-compat-data/pull/12194#issuecomment-921359678 for a test case.

You can also check out the first commit on this branch (5525c22) to see the tests fail, then check out the head to make them turn green.

### DIFF
--- a/package.json
+++ b/package.json
@@ -40,7 +40,7 @@
   },
   "scripts": {
     "confluence": "node ./node_modules/mdn-confluence/main/generate.es6.js --output-dir=. --bcd-module=./index.js",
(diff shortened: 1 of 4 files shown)


lukewarlow:
Might be a fix for https://github.com/mdn/browser-compat-data/issues/12663 ?

queengooborg (approved):
Thanks for tackling this @ddbeck, this is looking great to me!

PR #12697, opened 2021-10-04
FF94: enterkeyhint support added

hamishwillee:
FF94 ships enterkeyhint - as per https://bugzilla.mozilla.org/show_bug.cgi?id=1648332 , which was formerly behind the preference `dom.forms.enterkeyhint`.

Other docs work for this can be tracked in https://github.com/mdn/content/issues/9367

### DIFF
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -1075,25 +1075,24 @@
             "edge": {
               "version_added": "79"
             },
-            "firefox": {
-              "version_added": "79",
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "dom.forms.enterkeyhint",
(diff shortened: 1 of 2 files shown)


api/HTMLElement.json
```diff
@@ -1075,25 +1075,24 @@
...
+                  {
+                    "type": "preference",
+                    "name": "dom.forms.enterkeyhint",
+                    "value_to_set": "true"
+                  }
+                ]
+              },
+            ],
             "firefox_android": {
-              "version_added": "79",
```
hamishwillee, line 1089: Note, I removed this because FF android from v79 does not have method to set that preference - as discussed in #9679

Issue #12685, opened 2021-10-02
`standard_track` is redundant when there is `spec_url`


saschanaz:
Having a spec URL should automatically say it's on a standard track. But since not all data have `spec_url`, maybe at least make sure `standard_track` is consistent with the value of `spec_url` and ultimately remove the field?

foolip:
See https://github.com/mdn/browser-compat-data/issues/11216 for creating a guideline around `spec_url` generally.

gsnedders:
How far away are we from having `spec_url` everywhere?

saschanaz:
Modifying `required` field in `compat_statement` in `schemas/compat-data.schema.json` suggests there are *tons* of data that lacks `spec_url`.

gsnedders:
Filed #12709 to add what webref knows about, but there's still a _lot_ of omitted `spec_url`, especially with things like `worker_support` and event handlers and optional parameters/dictionary members.

martinthomson:
Why would a specification imply that something is a standard (or on track to be one)?  Any crank on the Internet can write a specification.  Many do.  Some of them end up on standards tracks of various sorts.

Mozilla has [a more advanced set of definitions](https://wiki.mozilla.org/ExposureGuidelines#Standardization_requirements_for_shipping_features) that might be a better substitute than testing for the presence of a specification.  Of course, that might mean some judgment is involved, which means documenting some guidelines.

FWIW, WICG documents - or any CG document - should not be marked as standards track.  Yes, they might eventually become standards (many have done so), but many others never go anywhere.

(Chris Wilson suggested that maybe you flag incubations; I would say that `standards_track` being false with a specification is exactly what an incubation should look like.)

sideshowbarker:
As I’ve said at https://github.com/mdn/browser-compat-data/issues/6103#issuecomment-625024050 and elsewhere, we are going to be much better off if we just remove `standard_track` altogether. What’s “standard” is subjective, and among the various stakeholders involved in wider discussions about what should be considered a standard and what shouldn’t be, we’ve never — in years of having discussions about it — arrived at any agreement.

More details and examples about why trying to identify what’s “standard” and what’s not is problematic:

* https://github.com/mdn/browser-compat-data/issues/6103#issuecomment-625003114
* https://github.com/mdn/browser-compat-data/issues/6103#issuecomment-624996231

[...]

PR #12709, opened 2021-10-04
Add many more spec_url entries based on webref data

gsnedders:
This is a simple addition of many more spec_url entries for api/* based on webref.

### DIFF
--- a/api/Attr.json
+++ b/api/Attr.json
@@ -141,7 +141,8 @@
             "experimental": false,
             "standard_track": true,
             "deprecated": false
-          }
+          },
+          "spec_url": "https://dom.spec.whatwg.org/#dom-attr-name"
         }
       },
       "namespaceURI": {
@@ -237,7 +238,8 @@
             "experimental": false,
             "standard_track": true,
             "deprecated": false
-          }
+          },
+          "spec_url": "https://dom.spec.whatwg.org/#dom-attr-ownerelement"
         }
       },
       "prefix": {
@@ -333,7 +335,8 @@
             "experimental": false,
             "standard_track": true,
             "deprecated": false
-          }
+          },
+          "spec_url": "https://dom.spec.whatwg.org/#dom-attr-specified"
         }
       },
       "value": {
@@ -380,7 +383,8 @@
             "experimental": false,
             "standard_track": true,
             "deprecated": false
-          }
+          },
+          "spec_url": "https://dom.spec.whatwg.org/#dom-attr-value"
         }
       }
     }
--- a/api/AudioListener.json
+++ b/api/AudioListener.json
@@ -458,7 +458,8 @@
             "experimental": false,
             "standard_track": false,
             "deprecated": true
-          }
+          },
+          "spec_url": "https://webaudio.github.io/web-audio-api/#dom-audiolistener-setorientation"
         }
       },
       "setPosition": {
--- a/api/AudioParam.json
+++ b/api/AudioParam.json
@@ -93,7 +93,8 @@
             "experimental": false,
             "standard_track": true,
             "deprecated": false
-          }
+          },
+          "spec_url": "https://webaudio.github.io/web-audio-api/#dom-audioparam-automationrate"
         }
       },
       "cancelAndHoldAtTime": {
--- a/api/AudioProcessingEvent.json
+++ b/api/AudioProcessingEvent.json
@@ -141,7 +141,8 @@
             "experimental": false,
             "standard_track": true,
             "deprecated": true
-          }
+          },
+          "spec_url": "https://webaudio.github.io/web-audio-api/#dom-audioprocessingevent-inputbuffer"
         }
       },
       "outputBuffer": {
@@ -189,7 +190,8 @@
             "experimental": false,
             "standard_track": true,
             "deprecated": true
-          }
+          },
+          "spec_url": "https://webaudio.github.io/web-audio-api/#dom-audioprocessingevent-outputbuffer"
         }
       },
       "playbackTime": {
@@ -237,7 +239,8 @@
             "experimental": false,
             "standard_track": true,
             "deprecated": true
-          }
+          },
+          "spec_url": "https://webaudio.github.io/web-audio-api/#dom-audioprocessingevent-playbacktime"
         }
       }
     }
--- a/api/AudioWorkletGlobalScope.json
+++ b/api/AudioWorkletGlobalScope.json
@@ -93,7 +93,8 @@
             "experimental": false,
             "standard_track": true,
             "deprecated": false
-          }
+          },
+          "spec_url": "https://webaudio.github.io/web-audio-api/#dom-audioworkletglobalscope-currentframe"
         }
       },
       "currentTime": {
@@ -141,7 +142,8 @@
             "experimental": false,
             "standard_track": true,
             "deprecated": false
-          }
+          },
+          "spec_url": "https://webaudio.github.io/web-audio-api/#dom-audioworkletglobalscope-currenttime"
         }
       },
       "registerProcessor": {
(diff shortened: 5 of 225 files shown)


gsnedders:
I _think_ my suggestion to review this is to look at cases where it adds multiple `spec_url` and decide whether it's doing something sensible, as in the general (simple) case I'm not worried by the data being added.

gsnedders:
Specifically, that is this subset:

```diff
diff --git a/api/HTMLCollection.json b/api/HTMLCollection.json
index eec99ac4d..e016caded 100644
--- a/api/HTMLCollection.json
+++ b/api/HTMLCollection.json
@@ -191,7 +191,11 @@
             "experimental": false,
             "standard_track": true,
             "deprecated": false
-          }
+          },
+          "spec_url": [
+            "https://dom.spec.whatwg.org/#dom-htmlcollection-nameditem",
+            "https://dom.spec.whatwg.org/#dom-htmlcollection-nameditem-key"
+          ]
         }
       }
     }
diff --git a/api/HTMLEmbedElement.json b/api/HTMLEmbedElement.json
index 64376e9a7..7c3459e44 100644
--- a/api/HTMLEmbedElement.json
+++ b/api/HTMLEmbedElement.json
@@ -286,7 +286,11 @@
             "experimental": false,
             "standard_track": true,
             "deprecated": false
-          }
+          },
+          "spec_url": [
+            "https://html.spec.whatwg.org/multipage/iframe-embed-object.html#dom-embed-src",
+            "https://w3c.github.io/webappsec-trusted-types/dist/spec/#dom-htmlembedelement-src"
+          ]
         }
       },
       "type": {
diff --git a/api/HTMLObjectElement.json b/api/HTMLObjectElement.json
index d61a873a7..0d700f090 100644
--- a/api/HTMLObjectElement.json
+++ b/api/HTMLObjectElement.json
@@ -334,7 +334,11 @@
             "experimental": false,
             "standard_track": true,
             "deprecated": true
-          }
+          },
+          "spec_url": [
+            "https://html.spec.whatwg.org/multipage/obsolete.html#dom-object-codebase",
+            "https://w3c.github.io/webappsec-trusted-types/dist/spec/#dom-htmlobjectelement-codebase"
+          ]
         }
       },
       "codeType": {
diff --git a/api/HTMLScriptElement.json b/api/HTMLScriptElement.json
index 5f6ee5158..e909722b4 100644
--- a/api/HTMLScriptElement.json
+++ b/api/HTMLScriptElement.json
@@ -518,7 +518,11 @@
             "experimental": false,
             "standard_track": true,
             "deprecated": false
-          }
+          },
+          "spec_url": [
+            "https://html.spec.whatwg.org/multipage/scripting.html#dom-script-src",
+            "https://w3c.github.io/webappsec-trusted-types/dist/spec/#dom-htmlscriptelement-src"
+          ]
         }
       },
       "text": {
@@ -565,7 +569,11 @@
             "experimental": false,
             "standard_track": true,
             "deprecated": false
-          }
+          },
+          "spec_url": [
+            "https://html.spec.whatwg.org/multipage/scripting.html#dom-script-text",
+            "https://w3c.github.io/webappsec-trusted-types/dist/spec/#dom-htmlscriptelement-text"
+          ]
         }
       },
       "type": {
diff --git a/api/HTMLTableColElement.json b/api/HTMLTableColElement.json
index 5645baee0..26fe2cb01 100644
--- a/api/HTMLTableColElement.json
+++ b/api/HTMLTableColElement.json
@@ -237,7 +237,11 @@
             "experimental": false,
             "standard_track": true,
             "deprecated": false
-          }
+          },
+          "spec_url": [
+            "https://html.spec.whatwg.org/multipage/tables.html#dom-col-span",
+            "https://html.spec.whatwg.org/multipage/tables.html#dom-colgroup-span"
+          ]
         }
       },
       "vAlign": {
diff --git a/api/ServiceWorker.json b/api/ServiceWorker.json
index 4e650a8db..8b033ed4c 100644
--- a/api/ServiceWorker.json
+++ b/api/ServiceWorker.json
@@ -192,7 +192,11 @@
             "experimental": false,
             "standard_track": true,
             "deprecated": false
-          }
+          },
+          "spec_url": [
+            "https://w3c.github.io/ServiceWorker/#dom-serviceworker-postmessage",
+            "https://w3c.github.io/ServiceWorker/#dom-serviceworker-postmessage-message-options"
+          ]
         }
       },
       "scriptURL": {
```

Earthman100:
Docker has Web-Versions on CustomEvents Saved in Blobs to Users UUIDS as well as UUIDs on Docker Generators to CustomEvents for People trying to break the Law



Sincerely, Sean P. Kropaczewski

sideshowbarker:
After looking through the patch a bit, I can there are a number of cases where it’s adding spec URLs for features that do in fact also have MDN URLs in BCD. However, in every case of those which I’ve checked so far, those MDN URLs are 404s; that is, we don’t yet actually have an article in MDN for that feature.

(This is another reason why it’s a pretty bad idea for us to be adding MDN URLs to BCD for articles that don’t actually exist yet.)

> in the general (simple) case I'm not worried by the data being added

Glancing through a sample of them, I agree they seem to not look incorrect — but that does not make me confident that in any particular case, it’s actually the best anchor to be using. For all the spec URLs we landed when we did the initial adding of all of them, we checked them one-by-one manually to make sure they were the best URLs to be citing. And in some non-insignificant number of cases, we ended deciding to use different URLs that were a better choice to be directing developer readers to.

sideshowbarker:
> @sideshowbarker This is, as linked above, motivated by #12685: given `standard_track` is currently required (if I'm reading the schema correctly), if we want to move to inferring `standard_track` from `spec_url` (or making assertions linking them) then we need to ensure we have `spec_url` wherever possible.

ah OK yeah — that makes sense

gsnedders:
> Glancing through a sample of them, I agree they seem to not look incorrect — but that does not make me confident that in any particular case, it’s actually the best anchor to be using. For all the spec URLs we landed when we did the initial adding of all of them, we checked them one-by-one manually to make sure they were the best URLs to be citing. And in some non-insignificant number of cases, we ended deciding to use different URLs that were a better choice to be directing developer readers to.

I guess my other take—wrt this—is that a bad spec link, pointing to something overly specific, is probably better than no spec link.

api/Attr.json
```diff
@@ -141,7 +141,8 @@
             "experimental": false,
             "standard_track": true,
             "deprecated": false
-          }
+          },
+          "spec_url": "https://dom.spec.whatwg.org/#dom-attr-name"
```
ddbeck, line 145: I'm not in a position (yet) to review the individual entries here, but I do want to highlight a structural problem: `spec_url` ought to be a sibling of `support` and `status` (usually coming before both, right after the `__compat` line).

ddbeck (changes requested):

api/Attr.json
gsnedders: It is a sibling of support and status? (It's at the end because, uh, it got added programmatically and as is often the case with JS dictionaries insertion order being used. I guess I could change the script to add it at the beginning.)

ddbeck: Sorry, I misread the nesting. 🤦 You're right, it's at the right place, but it would be nice to see it at the top (I realize that the actual representation is unordered, but it's nice to have a specific order in the source).

queengooborg:
Hey @gsnedders, this PR is awaiting both a rebase and requested changes.  Are you able and willing to come back to this PR?

queengooborg:
Thanks for your work on this PR, and sharing the script you used to create it!  Since there has been no response, I opened #16051 to supersede this.

gsnedders:
Oops! I clearly got distracted then forgot about this, sorry!

PR #12718, opened 2021-10-05
Remove PerformanceFrameTiming

saschanaz:
Fixes test failure in #12716

https://w3c.github.io/web-performance/meetings/2021/2021-09-16/index.html has some background why it's dropped.

#### Related issues

#12716


saschanaz:
A good reason not to add data and content for a non-implemented feature 😉

PR #12767, opened 2021-10-10
Update spec URL for CSS :has() selector

erosman:
#### Summary
Update the wrong URL

n/a

### DIFF
--- a/css/selectors/has.json
+++ b/css/selectors/has.json
@@ -5,7 +5,7 @@
         "__compat": {
           "description": "<code>:has()</code>",
           "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/:has",
-          "spec_url": "https://drafts.csswg.org/selectors/#relational",
+          "spec_url": "https://drafts.csswg.org/selectors-4/#relational",
           "support": {
             "chrome": {
               "version_added": false,


queengooborg:
It looks like their servers are normalized now though, as I mentioned above.  😛

erosman:
As soon as I posted, their server became operational.

Although both URLs are working now and both point to the **same document**, checking the main index page, (https://drafts.csswg.org/) , shows that they have separated Selectors into:

**Selectors**

- Selectors 3 (https://drafts.csswg.org/selectors-3/)
- Selectors 4 (https://drafts.csswg.org/selectors-4/)


It is my understanding that the new URL should be the proper one to use and the old URL has been kept active for backward compatibly with old links.

queengooborg:
Hmm...actually, I'm not sure about that.  I took another look at the Level 4 editor's draft, and when clicking the dropdown on the metadata, it states that the URL for that version does _not_ contain the `-4` in it.

Screenshot of the CSSWG Selectors Level 4 Editor's Draft

I've also noticed that the metadata refers to a "latest published version" that's hosted on W3.org, which contains a slightly older version, but one that's considered the Working Draft, and contains the `:has()` selector.  Perhaps that's what we should be linking to...?

Spec URLs are hard.  🙃

PR #12776, opened 2021-10-11
Sanitizer API updates

hamishwillee:
FF94 updates the Sanitizer API with support for methods `sanitizer.sanitizerFor()` and `Element.setHTML()` which are behind the same preference as earlier parts of the API: https://bugzilla.mozilla.org/show_bug.cgi?id=1716624

Further, I have checked and these are also supported in current Chrome 94. I couldn't find a bug report about when but it was probably Chrome 93 [(according to this](https://www.infoq.com/news/2021/10/wicg-html-sanitizer-api-mXSS/?topicPageSponsorship=5d18a24d-f16e-405a-95c4-f52cfcc38b9f&itm_source=presentations_about_development&itm_medium=link&itm_campaign=development)). Doesn't matter anyway since it is all under "preview".

Docs tracking - https://github.com/mdn/content/issues/9366

### DIFF
--- a/api/Sanitizer.json
+++ b/api/Sanitizer.json
@@ -6,7 +6,7 @@
         "spec_url": "https://wicg.github.io/sanitizer-api/#sanitizer-api",
         "support": {
           "chrome": {
-            "version_added": "preview",
+            "version_added": "93",
             "flags": [
               {
                 "type": "preference",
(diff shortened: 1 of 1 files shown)


jpmedley:
Looks good to me.

PR #12782, opened 2021-10-11
Mark font-synthesis:small-caps supported since Safari 10.1

foolip:
This is based on WebKit source history:
https://trac.webkit.org/changeset/209875/webkit
https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Configurations/Version.xcconfig?rev=209875

WebKit trunk version 603.1.17 suggests Safari 10.1 / iOS 10.3.

### DIFF
--- a/css/properties/font-synthesis.json
+++ b/css/properties/font-synthesis.json
@@ -78,10 +78,10 @@
                 "version_added": false
               },
               "safari": {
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
Seems like a good enough guess for now

PR #12818, opened 2021-10-12
Remove VRLayerInit API from BCD

queengooborg:
This PR removes `VRLayerInit` from BCD.  This feature is a dictionary, enum, or WebIDL typedef and should not be included in BCD.


ddbeck:
Fair enough, but was there any discussion about this approach? I'm surprised that the content PRs for this and https://github.com/mdn/browser-compat-data/pull/12766 were merged without discussion. I feel like there's a lot going on implicitly and I'd like to know where to refer to should questions arise (i.e., what do I say if someone tries to re-create the BCD for these pages?).

queengooborg:
There was some discussion when I had initially attempted to remove CryptoKeyPair from MDN web docs: https://github.com/mdn/content/pull/8846

PR #12830, opened 2021-10-13
Merge RTCConfiguration dict into RTCPeerConnection

queengooborg:
This PR removes `RTCConfiguration` from BCD.  This feature is a dictionary, enum, or WebIDL typedef and should not be included in BCD.


queengooborg:
I don't think that we have a set convention for this scenario, but in the MDN changes, I have made it so that the page for `setConfiguration()` has a link to the constructor to list the configuration options and their functions.  Thus, users will head to the constructor's documentation to find more information about these parameters.  Resultingly, I think it makes sense to just keep the BCD data here!

foolip:
I see, https://developer.mozilla.org/en-US/docs/Web/API/RTCPeerConnection/setConfiguration#parameters looks good!

api/RTCPeerConnection.json
```diff
@@ -210,6 +210,349 @@
...
+        "configuration_bundlePolicy_parameter": {
+          "__compat": {
+            "description": "<code>configuration.bundlePolicy</code> parameter",
+            "spec_url": "https://w3c.github.io/webrtc-pc/#dom-rtcconfiguration-bundlepolicy",
+            "support": {
+              "chrome": {
+                "version_added": "23"
+              },
+              "chrome_android": {
+                "version_added": "57"
```
foolip, line 223:
```suggestion
                "version_added": "42"
```

api/RTCPeerConnection.json
```diff
@@ -210,6 +210,349 @@
...
+            }
+          }
+        },
+        "configuration_certificates_parameter": {
+          "__compat": {
+            "description": "<code>configuration.certificates</code> parameter",
+            "spec_url": "https://w3c.github.io/webrtc-pc/#dom-rtcconfiguration-certificates",
+            "support": {
+              "chrome": {
+                "version_added": "23"
```
foolip, line 269: This was added in https://source.chromium.org/chromium/chromium/src/+/ef4ab6012fddc80992b8848cfc6700e731dc39b1 which maps to Chrome 48, but the feature was enabled in https://source.chromium.org/chromium/chromium/src/+/7318c1bfd0b0d45bc323bc4bcc45c98c4aeebfb6 which is Chrome 49. That also matches the `RTCCertificate` data.

```suggestion
                "version_added": "49"
```

api/RTCPeerConnection.json
```diff
@@ -210,6 +210,349 @@
...
+              },
+              "chrome_android": {
+                "version_added": "57"
```
foolip, line 272:
```suggestion
                "version_added": "49"
```

foolip (commented):
I don't trust the existing data here much at all. I double checked two, can you find the commits that added the rest in Chromium to double check?

api/RTCPeerConnection.json
```diff
@@ -210,6 +210,349 @@
...
+        "configuration_bundlePolicy_parameter": {
+          "__compat": {
+            "description": "<code>configuration.bundlePolicy</code> parameter",
+            "spec_url": "https://w3c.github.io/webrtc-pc/#dom-rtcconfiguration-bundlepolicy",
+            "support": {
+              "chrome": {
+                "version_added": "23"
+              },
+              "chrome_android": {
+                "version_added": "42"
```
foolip, line 223: I don't know where I got this from in https://github.com/mdn/browser-compat-data/pull/12830#discussion_r759189379, I'm pretty sure it was from some research though.

Can you re-research it and update Chrome and Chrome for Android to match?

foolip (commented):
I don't trust the original data at all here, so would like to see sources for at least all Chromium statements.

foolip:
@queengooborg can you turn this into a draft if it still needs research?

api/RTCPeerConnection.json
queengooborg: It appears this was added in https://source.chromium.org/chromium/chromium/src/+/70aed24e4ae5f04ec31709b36353bfaed01547ad, which dates Chrome 42 based upon feature freeze dates (Feb 6, 2015 for Chrome 42)!

foolip: Excellent! Can you update Chrome desktop data to match?

queengooborg:
It's quite possible that our existing data is incorrect, but I'm not yet sure how to check it and find out what the accurate versions are.  I think we can probably implement some form of test in the collector, but to do so, we'll need to fix the properties and where they're located.  I'll open up an issue to remind myself!

In summary: I'm lazy and would like to defer fixing the version numbers to future me... 😛

caugner:
> This feature is a dictionary, enum, or WebIDL typedef and should not be included in BCD.

@queengooborg Maybe this is obvious, but is this mentioned anywhere in the contribution or data docs?

caugner (changes requested):
IMO it would be good to update the PR title, as RTCConfiguration is not just removed, but merged/integrated into RTCPeerConnection.

queengooborg:
PR title has been updated and this is ready for review again!

foolip (approved):

foolip:
I've filed https://github.com/mdn/browser-compat-data/issues/16595 about data that's probably wrong here.

PR #12831, opened 2021-10-13
Remove RTCTrackEventInit API from BCD

queengooborg:
This PR removes `RTCTrackEventInit` from BCD.  This feature is a dictionary, enum, or WebIDL typedef and should not be included in BCD.


foolip:
Does the data of any of the parameters differ from the properties? If not, or if the differences is very small, they can be omitted.

queengooborg:
There doesn't look to be any difference between the parameters and the properties!  I'll go ahead and remove the newly-added subfeatures.

foolip (approved):

PR #12879, opened 2021-10-15
Remove api.XMLHttpRequest.readyState.constants from BCD

queengooborg:
Data about constants should not be included BCD:
https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#constants


foolip:
@queengooborg there's a conflict to resolve here.

queengooborg:
Realizing we don't need to make any content updates for this removal!

foolip (approved):

foolip:
> This feature is a dictionary, enum, or WebIDL typedef and should not be included in BCD.

I saw about to merge when I noticed this description isn't right. The reason this should be removed is https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#constants.

PR #12924, opened 2021-10-18
Mirror Samsung Internet from Chrome Android

AdaRoseCannon:
This PR mirrors the data for Samsung Internet using the mirroring script


AdaRoseCannon:
Oh that was the time the beta is released. We haven't publicly announced
the stable 16.0 release date yet as it depends on a few external factors
which we can't control.

PR #12931, opened 2021-10-19
Updated data for font-synthesis property

tursunova:
#### Summary

This PR sets Chromium version for font-synthesis CSS property and adds font-synthesis longhand

WARNING: font-synthesis property will be supported on Chrome 97 version, this CL has to be merged only after Chrome 97 version reaches the end of the [beta phase](https://chromiumdash.appspot.com/schedule)

Support in Chrome, Chrome for Android is claimed on [Chromestatus](https://www.chromestatus.com/feature/5640605355999232)
 
Longhand for font-synthesis property, i.e. [font-synthesis-weight](https://drafts.csswg.org/css-fonts-4/#font-synthesis-weight), [font-synthesis-style](https://drafts.csswg.org/css-fonts-4/#font-synthesis-style) and [font-synthesis-small-caps](https://drafts.csswg.org/css-fonts-4/#font-synthesis-small-caps) not supported on Safari ([link on bug](https://bugs.webkit.org/show_bug.cgi?id=232009)) and Firefox ([link on bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1724892))

### DIFF
--- a/css/properties/font-synthesis-small-caps.json
+++ b/css/properties/font-synthesis-small-caps.json
@@ -1,0 +1,58 @@
+{
+  "css": {
+    "properties": {
+      "font-synthesis-small-caps": {
+        "__compat": {
+          "spec_url": "https://drafts.csswg.org/css-fonts/#font-synthesis-small-caps",
+          "support": {
+            "chrome": {
+              "version_added": "97"
+            },
+            "chrome_android": {
+              "version_added": "97"
+            },
+            "edge": {
+              "version_added": false
+            },
+            "firefox": {
+              "version_added": false,
+              "notes": "<a href='https://bugzil.la/1724892'>bug 1724892</a>."
+            },
+            "firefox_android": {
+              "version_added": false,
+              "notes": "<a href='https://bugzil.la/1724892'>bug 1724892</a>."
+            },
+            "ie": {
+              "version_added": false
+            },
+            "opera": {
+              "version_added": false
+            },
+            "opera_android": {
+              "version_added": false
+            },
+            "safari": {
+              "version_added": false,
+              "notes": "<a href='https://webkit.org/b/232009'>bug 232009</a>."
+            },
+            "safari_ios": {
+              "version_added": false,
+              "notes": "<a href='https://webkit.org/b/232009'>bug 232009</a>."
+            },
+            "samsunginternet_android": {
+              "version_added": false
+            },
+            "webview_android": {
+              "version_added": false
+            }
+          },
+          "status": {
+            "experimental": true,
+            "standard_track": true,
+            "deprecated": false
+          }
+        }
+      }
+    }
+  }
+}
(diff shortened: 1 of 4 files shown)


css/properties/font-synthesis-small-caps.json
```diff
@@ -0,0 +1,58 @@
...
+          "spec_url": "https://drafts.csswg.org/css-fonts/#font-synthesis-small-caps",
+          "support": {
+            "chrome": {
+              "version_added": "97"
+            },
+            "chrome_android": {
+              "version_added": "97"
+            },
+            "edge": {
+              "version_added": false
```
ddbeck, line 15: The other Chromium-based browsers are all but certain to ship this feature at the same time. It'd be nice to see that reflected in the data here.

You can find information on the map from Chrome (Blink) to other browsers in the browsers data: https://github.com/mdn/browser-compat-data/tree/main/browsers. For example, Edge (and WebView) would be 97 and Opera would be 83. If the browsers data doesn't map to 97 or later, then leave it as false (e.g., Opera Android). It's OK that some of the releases are in the future, provided it's consistent with the `browsers` data.

This would apply throughout the PR, for Edge, Opera, and WebView Android.

ddbeck (changes requested):
Thanks for starting this challenging PR, @tursunova. It's tough to add new features like this, but you've gotten very close here. I have one last suggestion in a line comment below. Thank you again!

css/properties/font-synthesis-small-caps.json
tursunova: Sorry for the late reply and thank you for your comment, @ddbeck! 
Following your suggestions, I have changed versions for other Chromium-based browsers as well.

css/properties/font-synthesis-style.json
```diff
@@ -0,0 +1,58 @@
...
+            },
+            "samsunginternet_android": {
+              "version_added": false
+            },
+            "webview_android": {
+              "version_added": "97"
+            }
+          },
+          "status": {
+            "experimental": false,
```
ddbeck, line 50: Until there are multiple implementations, this should be experimental. See the [_Choosing an experimental status_ data guideline](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#choosing-an-experimental-status) for details.

```suggestion
            "experimental": true,
```

css/properties/font-synthesis-weight.json
```diff
@@ -0,0 +1,58 @@
...
+            },
+            "samsunginternet_android": {
+              "version_added": false
+            },
+            "webview_android": {
+              "version_added": "97"
+            }
+          },
+          "status": {
+            "experimental": false,
```
ddbeck, line 50:
```suggestion
            "experimental": true,
```

ddbeck (changes requested):
Two minor fixes to get this ready to merge. Thank you!

ddbeck (approved):
Thank you! 🎉

drott:
Super, thanks Munira!

PR #12991, opened 2021-10-21
Update speech media type support for Opera browser

pepelsbey:
#### Summary

`@media speech` introduced in Opera Presto engine is not supported in Chromium based Opera.

Support data states otherwise because it’s lacking `version_removed` keys.

Unfortunately I don’t see any [related tests in WPT](https://wpt.fyi/results/css/mediaqueries?label=experimental&label=master&aligned), but as a former Opera Software employee I can assure you that there were no plans in 2012–2013 during engine switch to implement any web platform APIs previously supported by Presto.

I don’t see it mentioned anywhere in [Opera version history](https://help.opera.com/en/opera-version-history/).

### DIFF
--- a/css/at-rules/media.json
+++ b/css/at-rules/media.json
@@ -1779,7 +1779,7 @@
               },
               "opera_android": {
                 "version_added": "10.1",
-                "version_removed": "15"
+                "version_removed": "14"
               },
               "safari": {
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Looks good. Thank you! 🎉

PR #13060, opened 2021-10-24
Add Safari versions for api.EventTarget.removeEventListener.type_listener_parameters_optional

queengooborg:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `removeEventListener.type_listener_parameters_optional` member of the `EventTarget` API.  I can't find any indication that these parameters were ever optional in the [commit history](https://github.com/WebKit/WebKit/blob/main/Source/WebCore/dom/EventTarget.idl).

### DIFF
--- a/api/EventTarget.json
+++ b/api/EventTarget.json
@@ -838,10 +838,10 @@
                 "version_added": true
               },
               "safari": {
-                "version_added": true
+                "version_added": false
               },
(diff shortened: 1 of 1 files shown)


queengooborg:
Closing this PR since we're talking about removing this feature.

PR #13101, opened 2021-10-26
Update note and bug link for navigator.onLine in WebView

foolip:


queengooborg (approved):
Definitely a way more helpful note, thank you!

PR #13107, opened 2021-10-26
Add Firefox versions for api.EventTarget.addEventListener.options_parameter.*

queengooborg:
This PR adds real values for Firefox and Firefox Android for the `addEventListener.options_parameter` members of the `EventTarget` API, based upon commit history and date.

Commit: https://hg.mozilla.org/mozilla-central/rev/5b7777972750aa3c57e6f6f6d4e4621072741617

### DIFF
--- a/api/EventTarget.json
+++ b/api/EventTarget.json
@@ -313,7 +313,7 @@
                   "version_added": "50"
                 },
                 "firefox_android": {
-                  "version_added": "49"
+                  "version_added": "50"
                 },
                 "ie": {
                   "version_added": false
--- a/api/EventTarget.json
+++ b/api/EventTarget.json
@@ -310,7 +310,7 @@
                   "version_added": "≤18"
                 },
                 "firefox": {
-                  "version_added": "49"
+                  "version_added": "50"
                 },
                 "firefox_android": {
                   "version_added": "49"
--- a/api/EventTarget.json
+++ b/api/EventTarget.json
@@ -255,10 +255,10 @@
                   "version_added": "≤18"
                 },
                 "firefox": {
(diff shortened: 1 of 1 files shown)


api/EventTarget.json
```diff
@@ -310,10 +310,10 @@
                   "version_added": "≤18"
                 },
                 "firefox": {
-                  "version_added": "50"
+                  "version_added": "49"
```
ddbeck, line 313: OK, I did some more digging and this definitely didn't ship in 49. See https://bugzilla.mozilla.org/show_bug.cgi?id=1287706.

```suggestion
                  "version_added": "50"
```

api/EventTarget.json
```diff
@@ -310,10 +310,10 @@
...
                 },
                 "firefox_android": {
-                  "version_added": "50"
+                  "version_added": "49"
```
ddbeck, line 316:
```suggestion
                  "version_added": "50"
```

ddbeck (changes requested):

api/EventTarget.json
```diff
@@ -255,10 +255,10 @@
                   "version_added": "≤18"
                 },
                 "firefox": {
-                  "version_added": true
+                  "version_added": "49"
```
foolip, line 258: Confirmed by testing http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=9840 in Firefox 48 and 49. One has to test `{ capture: false}` as well since if the options dictionary isn't supported at all, any object just gets interpreted as a boolean.

This also confirms that the `options_parameter` data above is correct.

foolip (approved):

PR #13108, opened 2021-10-26
Remove api.EventTarget.removeEventListener.type_listener_parameters_optional

queengooborg:
This PR removes `api.EventTarget.removeEventListener.type_listener_parameters_optional` from BCD.  This feature should actually be reversed to when it was made `required`, but tracking down when this was made required in Firefox is more trouble than I feel it's worth.  I feel that it's safe to simply represent these parameters as always being required by removing this feature.

~~Closes #13060.~~

### DIFF
--- a/api/EventTarget.json
+++ b/api/EventTarget.json
@@ -800,65 +800,6 @@
             }
           }
         },
-        "type_listener_parameters_optional": {
-          "__compat": {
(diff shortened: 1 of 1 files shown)


queengooborg:
I went ahead and merged #13143 so I'll be closing this PR.  #13143 contains some more details as to why this feature should be removed, and a part of it is backed by #13060 as well.

PR #13114, opened 2021-10-26
Remove irrelevant moz-prefixed sub-features of XHR responseType

foolip:
The latest of these was removed in Firefox 68, which is now more than 2
years ago.


queengooborg (approved):
No need for these irrelevant features any longer!

PR #13117, opened 2021-10-26
Add partial_implementation to Element members moved from HTMLElement

foolip:
This was already done in some places. Do it consistently in
Element.json, following this guideline:


queengooborg (approved):

PR #13127, opened 2021-10-26
Remove api.isSecureContext.considers_window_opener

queengooborg:
This PR removes `api.isSecureContext.considers_window_opener` from BCD.  This feature is marked as standard and experimental, but looking through the spec I cannot find any mention of an `opener` or anything relevant for this property.  It's possible this was added to a draft version of the spec, added in Firefox, but then removed from the spec.

Additionally, there is no documentation on on the MDN page that discusses what this feature means exactly.

### DIFF
--- a/api/_globals/isSecureContext.json
+++ b/api/_globals/isSecureContext.json
@@ -48,54 +48,6 @@
           "deprecated": false
         }
       },
-      "considers_window_opener": {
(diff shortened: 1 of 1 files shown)


foolip:
Thanks for checking! I agree this can be removed.

foolip (approved):

PR #13132, opened 2021-10-27
Update Blob support on Firefox for Android.

makotokato:
#### Summary
Gecko supports some new Blob APIs at 69, but BCD doesn't update this data for FxA due to Firefox → Fenix transition. So we should set that supported version is 79.

Pass WPT (http://wpt.live/FileAPI/idlharness.html etc)

#### Related issues
- https://bugzilla.mozilla.org/show_bug.cgi?id=1557121
- https://github.com/mdn/browser-compat-data/pull/4816/

### DIFF
--- a/api/Blob.json
+++ b/api/Blob.json
@@ -128,7 +128,7 @@
               "version_added": "69"
             },
             "firefox_android": {
-              "version_added": false
+              "version_added": "79"
             },
             "ie": {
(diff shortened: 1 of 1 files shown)


foolip (approved):

foolip:
Thanks @makotokato!

PR #12489, opened 2021-09-20
Adding @layer

rachelandrew:
I'm documenting `@layer`, this is the BCD. Currently, the Firefox implementation is behind a flag https://github.com/mdn/content/issues/8999

PR #13137, opened 2021-10-27
adding at layer rule

Rumyra:
#### Summary

Adding the `@layer` rule for CSS which is available under a flag in both Chrome (97) and Firefox (94)

It doesn't like the spec url I have added - I'm not sure how to resolve this

The only thing I got is this https://bugzilla.mozilla.org/show_bug.cgi?id=1699217 - but I've written some code which works in both browsers with the flags turned on

#### Related issues

https://github.com/mdn/content/issues/9369

### DIFF
--- a/css/at-rules/layer.json
+++ b/css/at-rules/layer.json
@@ -1,0 +1,77 @@
+{
+  "css": {
+    "at-rules": {
+      "layer": {
+        "__compat": {
+          "description": "<code>@layer</code>",
+          "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/@layer",
+          "spec_url": "https://drafts.csswg.org/css-cascade-5/#layering",
+          "support": {
+            "chrome": {
+              "version_added": "97",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "enable-cascade-layers",
+                  "value_to_set": "true"
+                }
+              ]
+            },
+            "chrome_android": {
+              "version_added": false
+            },
+            "edge": {
+              "version_added": false
+            },
+            "firefox": {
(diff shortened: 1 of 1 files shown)


css/at-rules/layer.json
```diff
@@ -0,0 +1,77 @@
+{
+  "css": {
+    "at-rules": {
+      "layer": {
+        "__compat": {
+          "description": "<code>@layer</code>",
+          "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/@layer",
+          "spec_url": "https://drafts.csswg.org/css-cascade-5/#layering",
```
queengooborg, line 8:
```suggestion
          "spec_url": "https://www.w3.org/TR/css-cascade-5/#layering",
```

css/at-rules/layer.json
```diff
@@ -0,0 +1,77 @@
...
+          "support": {
+            "chrome": {
+              "version_added": "97",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "enable-cascade-layers",
```
queengooborg, line 15:
```suggestion
                  "name": "#enable-cascade-layers",
```

Can we also copy this flag data over to Chrome Android, Edge, and Opera?

queengooborg (changes requested):
Thanks for your PR, apologies for the delay before reviewing!  Let's use the track release spec to make the browser spec linter happy.

queengooborg:
Wait, actually, I just noticed this is a duplicate of #12489.  I'm going to close this in favor of that PR.

Rumyra:
OK that's brill :D thanks @queengooborg

PR #13162, opened 2021-10-28
Improve descriptions of all_elements_selector entries

foolip:
Quote the strings, and add the first argument for the NS variant.


queengooborg (approved):
I'm liking these description updates, thanks!

PR #13172, opened 2021-10-29
Error constructor missing filename/linenumber parameters

hamishwillee:
[`Error` constructor](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Error/Error) specifies that the constructor takes arguments `message`, `options.cause`, `fileName`, `lineNumber`.

` fileName`, `lineNumber` are in [javascript/builtins/Error.json](https://github.com/mdn/browser-compat-data/blob/main/javascript/builtins/Error.json) as properties, but not listed as error constructor options - so they aren't visible in MDN on the error constructor page.

This reproduces the data for the properties under the error constructor. Note that it does NOT reproduce message, which is assumed to match the parent constructor function.

This is part of fixing #10159

### DIFF
--- a/javascript/builtins/Error.json
+++ b/javascript/builtins/Error.json
@@ -110,61 +110,6 @@
               "deprecated": false
             }
           },
-          "options_cause_parameter": {
-            "__compat": {
-              "description": "<code>options.cause</code> parameter",
-              "spec_url": "https://tc39.es/proposal-error-cause/#sec-errorobjects-install-error-cause",
-              "support": {
-                "chrome": {
-                  "version_added": "93"
-                },
-                "chrome_android": {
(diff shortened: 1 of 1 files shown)


evilpie:
It's also not obvious on MDN that `fileName` and `lineNumber` are Firefox non-standard extensions.

hamishwillee:
> It's also not obvious on MDN that fileName and lineNumber are Firefox non-standard extensions.

@evilpie Yes, but _this_ PR is the fix for that problem.

Specifically the fields are already marked as non-standard in MDN. With this change the compatibility table will be updated on the page, making it obvious that they are firefox-only.

ddbeck (approved):

PR #13178, opened 2021-10-29
Remove api.MessageEvent.source.MessageEventSource_type

queengooborg:
This PR removes the `MessageEventSource_type` subfeature of the `MessageEvent` API's `source` member.  This is a WebIDL typedef that defines a `WindowProxy` (formerly just `Window`), `MessagePort`, or `ServiceWorker`.  To the end web developer, this typedef is indistinguishable, thus there is no need to include this as a BCD feature.


foolip:
In https://github.com/mdn/browser-compat-data/pull/13073#issuecomment-953644382 I had some thoughts about what this entry means. Is there something about `ServiceWorker` support that should be noted on some other feature? I suspect not, but want to make sure it's noted and accounted for in this removal PR.

queengooborg:
There's nothing about `ServiceWorker` that we need to keep track of; the `source` property could first be a service worker all the way since service workers were implemented to begin with!

foolip:
Alright, let's drop this one!

PR #13192, opened 2021-10-30
Add IE/Edge versions for api.Document.createElementNS.options_parameter

queengooborg:
This PR adds real values for Internet Explorer and Edge for the `createElementNS.options_parameter` member of the `Document` API.  The data for `createElement.options_parameter` was set to `false` for IE and pre-Chromium Edge in https://github.com/mdn/browser-compat-data/pull/4975.  I doubt that this would be supported in `createElementNS` if `createElement` doesn't support it.

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -2213,7 +2213,7 @@
                 "notes": "For backwards compatibility, the <code>options</code> parameter can be an object or a string with the custom element tag name, although the string version is deprecated."
               },
               "edge": {
-                "version_added": "≤79",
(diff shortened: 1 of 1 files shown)


foolip (approved):
This is part of custom elements, and this change matches other relevant data:
https://developer.mozilla.org/en-US/docs/Web/API/Document/registerElement#browser_compatibility
https://developer.mozilla.org/en-US/docs/Web/API/CustomElementRegistry#browser_compatibility

PR #13195, opened 2021-10-31
Add Safari versions for api.Selection.collapse.node_parameter_nullable

queengooborg:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `collapse.node_parameter_nullable` member of the `Selection` API, based upon manual testing.

Test Code Used:
```js
try {
	var sel = window.getSelection();
	sel.collapse(null, 0);
} catch(e) {
	alert(e);
}
```

### DIFF
--- a/api/Selection.json
+++ b/api/Selection.json
@@ -274,10 +274,10 @@
                 "version_added": "26"
               },
               "safari": {
-                "version_added": true
+                "version_added": "10"
               },
               "safari_ios": {
-                "version_added": true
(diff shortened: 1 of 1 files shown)


foolip:
I'm not sure about this. Just like all arguments were originally optional, passing values of the wrong type was silently ignored.

This should be easy to test, checking if passing null has ever thrown an exception.

foolip (approved):

PR #13197, opened 2021-10-31
Add IE versions for api.HTMLCanvasElement.getContext.webgl_context.options_failIfMajorPerformanceCaveat_parameter

queengooborg:
This PR adds real values for Internet Explorer for the `getContext.webgl_context.options_failIfMajorPerformanceCaveat_parameter` member of the `HTMLCanvasElement` API.  This PR simply sets the value to "11", because honestly, I have no idea how I'd be able to create a test for this.  I don't think anyone would care about data before IE 11 nowadays as it is, and if anyone does, I feel that they could make the corrections themselves.

### DIFF
--- a/api/HTMLCanvasElement.json
+++ b/api/HTMLCanvasElement.json
@@ -483,7 +483,7 @@
                   "version_added": "41"
                 },
                 "ie": {
-                  "version_added": true
+                  "version_added": "11"
                 },
                 "opera": {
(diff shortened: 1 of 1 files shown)


foolip (approved):
Since the parent feature says 11, this is already what the data means. It's either 11 or false, and since someone set it to true at one point, let's keep trusting that.

Issue #13170, opened 2021-10-29
Complete nodejs data for api section


gilmoreorless:
After #6249 was completed we have all-real values for `nodejs` in the `javascript` section of the data. This is a similar high-level tracking issue to weed out the non-real values in the `api` section.

These are the non-real values I've found, and I'll progressively raise some PRs for them (data correct for version `4.0.8`).

### `true` values (31) [fixed in #13204, #13205, #13655, #13656]

- `api.Console.info.substitution_strings`
- `api.MessagePort.worker_support`
- `api.URL.URL`
- `api.URL.hash`
- `api.URL.host`
- `api.URL.hostname`
- `api.URL.href`
- `api.URL.origin`
- `api.URL.password`
- `api.URL.pathname`
- `api.URL.port`
- `api.URL.protocol`
- `api.URL.search`
- `api.URL.toJSON`
- `api.URL.username`
- `api.URLSearchParams.append`
- `api.URLSearchParams.delete`
- `api.URLSearchParams.entries`
- `api.URLSearchParams.get`
- `api.URLSearchParams.getAll`
- `api.URLSearchParams.has`
- `api.URLSearchParams.keys`
- `api.URLSearchParams.set`
- `api.URLSearchParams.toString`
- `api.URLSearchParams.values`
- `api.clearInterval`
- `api.clearTimeout`
- `api.setInterval`
- `api.setInterval.supports_parameters_for_callback`
- `api.setTimeout`
- `api.setTimeout.supports_parameters_for_callback`

### `null` values (11) [fixed in #12154, #14632, #14633, #14660, #14666, #14681, #14953]

- `api.Crypto.subtle.secure_context_required`
- `api.CryptoKey.secure_context_required`
- `api.EventTarget.addEventListener.options_parameter.options_passive_parameter_default_true_touch`
- `api.EventTarget.addEventListener.options_parameter.options_passive_parameter_default_true_wheel`
- `api.PerformanceEntry.duration`
- `api.PerformanceObserver.PerformanceObserver`
- `api.PerformanceObserver.worker_support`
- `api.SubtleCrypto.secure_context_required`
- `api.SubtleCrypto.worker_support`
- `api.TextDecoder.TextDecoder`
- `api.TextEncoder.encodeInto`

### Missing entry completely (31) [fixed in #13204, #13655, #13657, #14666, #14681, #14954]

[...]

PR #13204, opened 2021-10-31
Complete nodejs data for global timers

gilmoreorless:
#### Summary

Part of #13170. This fills in `true` and missing `nodejs` entries for global timers (`setTimeout` et al).

The basic methods were in the very first version, as listed in the history for each method at https://nodejs.org/dist/latest-v16.x/docs/api/timers.html

Worker threads in general were added to NodeJS in 10.5.0: https://nodejs.org/dist/latest-v16.x/docs/api/worker_threads.html


queengooborg (approved):
This looks good to me, thanks!

PR #13207, opened 2021-10-31
Add IE versions for api.FormData.append.filename_parameter

queengooborg:
This PR adds real values for Internet Explorer for the `append.filename_parameter` member of the `FormData` API.  I tried to write a manual test for this, but as it turns out, it's pretty much impossible to get the data back out of a `FormData` object in IE until it's been sent to a server.  Based upon the notes, I'm assuming that support for this parameter came when support for a file/blob was included, which seems to be IE 10.

Test code I tried to use:
```js
function createBlob() {
	var bytes = new Uint8Array(59);

	for(var i = 0; i < 59; i++) {
		bytes[i] = 32 + i;
	}

	return new Blob([bytes.buffer], {type: 'text/plain'});
}

var formData = new FormData();
var blob = createBlob();

formData.append('userpic', blob, 'myfile.txt');
```

### DIFF
--- a/api/FormData.json
+++ b/api/FormData.json
@@ -185,7 +185,7 @@
                 "version_added": "22"
               },
               "ie": {
-                "version_added": true
+                "version_added": "10"
               },
               "opera": {
                 "version_added": "≤15"


foolip (approved):
This seems like a good enough guess, thanks!

PR #13208, opened 2021-10-31
Add IE/Edge versions for api.PointerEvent.pointerType.fractional_coodinates

queengooborg:
This PR adds real values for Internet Explorer and Edge for the `pointerType.fractional_coodinates` member of the `PointerEvent` API, based upon manual testing.  A high DPI screen was simulated by zooming the page to 200%.

Test Code Used:
```js
document.addEventListener('pointerdown', function(event) {
	alert(event.pageX);
});
```

### DIFF
--- a/api/PointerEvent.json
+++ b/api/PointerEvent.json
@@ -688,11 +688,19 @@
                 "partial_implementation": true,
                 "notes": "<code>movementX</code> and <code>movementY</code> are not fractional, see <a href='https://crbug.com/802067'>Chromium bug 802067</a>."
               },
-              "edge": {
-                "version_added": "79",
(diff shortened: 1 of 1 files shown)


foolip:
This looks a bit suspect to me. Microsoft championed pointer events and were the first to ship it. Supporting high-resolution screens seems like it would have been an important goal, especially for something like stylus input.

Was this tested with a high-resolution screen, without any VMs involved?

queengooborg:
Okay, so apparently the dev console for Edge prints floating point numbers as integers if the decimal is zero.  I thought I checked for this before, apparently not -- I'll look into this some more!

foolip (approved):

Issue #12533, opened 2021-09-22
Is there a way to indicate "not applicable" for versions


hamishwillee:
Discussion came up [here](https://github.com/mdn/content/issues/9139) asking why HTTP `Digest` and `Want-Digest` headers are marked as "supported" when the browser does nothing with them. This is answered in  #4600:

> I have just used true here, since all browsers have to do to support it is include any request headers they are asked to include, and expose any response headers the server sends them.

I understand why this is done, but it is [misleading](https://github.com/mdn/content/issues/9139#issuecomment-924549071). 
What are our options?

I was thinking that it would be quite convenient to be able indicate that the whole support story is "Not applicable" - so you could keep the spec links, and possibly even have sub features, but not have to fill in the support section. Possibly with a note as to why it isn't applicable.

@ddbeck Thoughts?

reschke:
What makes things even worse is this:

  https://caniuse.com/mdn-http_headers_digest

It suggests that some browsers do things others do not, which evidently is not true.

queengooborg:
If there's a feature that doesn't really have compatibility data, such as dictionaries, typedefs, or in this case an HTTP header that browsers do nothing with, then the general consensus has been that it doesn't belong here in BCD.  Thus, I'd say we should probably remove this entry?

I opened up an issue on Yari discussing possible fallbacks for the spec URL when a feature isn't in BCD, see https://github.com/mdn/yari/issues/4704.

hamishwillee:
@queengooborg I chatted to @ddbeck about this on Thursday. While it is correct that there is no browser compatibility "story" it might be worth looking at other alternatives:
- Knowing that something has no compatibility story can be useful - so we don't revisit this every time someone else comes to do compatibility documentation.
- Having spec data here is useful. Further you can't necessarily easily just replace in yari, particularly when BCD breaks down to a lower level than MDN docs.

Daniel is going to have a chat with other team members offline to work though options (presumably including you).

LPardue:
Hey, I'm an editor on the Digest header draft. I agree with all of Julian's points here and the issue on the content repo. 

Digest and Want-Digest are request and response header and trailer fields I can think of dozens of permutations of how Digest could be used by client or server. The draft includes a handful https://httpwg.org/http-extensions/draft-ietf-httpbis-digest-headers.html#appendix-B

The browsers are using Digest headers for anything, so they can't support anything. Passing through headers isnt support, it is simply not blocking or dropping them.

[...]

PR #13219, opened 2021-11-01
Mark HTTP digest and want-digest support as null/unknown

hamishwillee:
This changes the `Digest` and `Want-Digest` HTTP header support from `true` to `null` for all platforms.

- `true` is incorrect because the browser does nothing with these headers.
- `false` is correct because there are things a browser could do with these headers. However if we used `false` the table would be empty, and we normally don't allow all-false data. 
-  `null` is a compromise, which essentially says "we don't know" (but at least we're not pretending it is supported in some way). It will force us to return to this question in the near future when we start enforcing better data quality on the HTTP section. At that time I am hoping we allow `false` and a note to be added as top level support option for all browsers so this can be explained.

There is a discussion on this in https://github.com/mdn/browser-compat-data/issues/12533 and https://github.com/mdn/content/issues/9139. The decision for now was that we set this as null and perhaps revisit later.

### DIFF
--- a/http/headers/digest.json
+++ b/http/headers/digest.json
@@ -7,40 +7,40 @@
           "spec_url": "https://datatracker.ietf.org/doc/html/draft-ietf-httpbis-digest-headers-05#section-3",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": null
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": null
             },
             "edge": {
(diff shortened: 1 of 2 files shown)


hamishwillee:
FYI @reschke et al. For now we're just going to be less incorrect, with the intent to return to this in the near future.

ddbeck (approved):

PR #13221, opened 2021-11-01
SubtleCrypto - match Edge to Chrome from v79

hamishwillee:
SubtleCrypto was present on Edge 12 but had a partial implementation on many methods. Since Edge updated to Blink on Edge 79 my ASSUMPTION is that at that point the methods match chromium. I have not tested that, but just updated edge to match chromium from v79.


queengooborg (approved):

PR #13206, opened 2021-10-31
Mark offsetX and offsetY as not experimental

Timmmm:
#### Summary
Mark offsetX and offsetY as not experimental.

#### Test results and supporting details
These are almost universally supported.

#### Related issues
https://github.com/mdn/content/issues/10120

PR #13232, opened 2021-11-02
MouseEvent - pageX, pageY, x, y not experimental

hamishwillee:
Follows on from #13206 - if the offsetX and offsetY are not experimental, neither are pageX, pageY, x, y. These are widely supported across many browsers in many versions.

The associated MDN docs were update in https://github.com/mdn/content/pull/10220

### DIFF
--- a/api/MouseEvent.json
+++ b/api/MouseEvent.json
@@ -971,7 +971,7 @@
             }
           },
           "status": {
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Looks good. Thank you! 🎉

PR #13236, opened 2021-11-02
SpeechSynthesisEvent.elapsedTime spec change ms to s

### LINKED ISSUE #12595: Issue with "SpeechSynthesisUtterance.onend": wrong unit for elapsedTime
BCD issue description:

While the spec states this is in seconds, it looks like only Safari returns a value in seconds. Both Chrome and Firefox currently erroneously return a value in milliseconds:

* https://bugs.chromium.org/p/chromium/issues/detail?id=1253036
* https://bugzilla.mozilla.org/show_bug.cgi?id=1732498

---

Original MDN issue description:

MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/SpeechSynthesisUtterance/onend

In the example, the elapsed time (event.elapsedTime) is printed as if it was milliseconds. But it's in second per specification and implementation.

#### Specific section or headline?
Examples

console.log('Utterance has finished being spoken after ' + event.elapsedTime + ' seconds.');

Specification: https://wicg.github.io/speech-api/#speechsynthesisevent-attributes

<details>

* Folder: `en-us/web/api/speechsynthesisutterance/onend`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/SpeechSynthesisUtterance/onend
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/api/speechsynthesisutterance/onend/index.html
* Last commit: https://github.com/mdn/content/commit/642d6eb7feb73e17cfc1ef61d3d09c2d200e65ca
* Document last modified: 2021-06-17T03:52:55.000Z

</details>

hamishwillee:
The Web Speech API spec recently changed the event property [`SpeechSynthesisEvent.elapsedTime`](https://developer.mozilla.org/en-US/docs/Web/API/SpeechSynthesisEvent/elapsedTime) from milliseconds to seconds:
- FF95 has made this change in https://bugzilla.mozilla.org/show_bug.cgi?id=1732498
- Chromium issue (not fixed) is https://bugs.chromium.org/p/chromium/issues/detail?id=1253036
- Safari 14.1.2 on macOS appears to do the right thing (seconds) according to this https://bugzilla.mozilla.org/show_bug.cgi?id=1732498#c0 . I have no information on when it changed but I've used 14.1 because it coincides fairly well with spec change and also mentions the [API in release notes](https://developer.apple.com/documentation/safari-release-notes/safari-14_1-release-notes)

There is a demo fiddle here https://jsfiddle.net/ybe1c6ot/ ([comes from this issue](https://github.com/mdn/content/pull/10179)).

To update this I have followed @ddbeck suggestion of having a subfeature for the deprecated behaviour named milliseconds, which has a version-removed matching the version in which browsers update. I also will not update the handlers and specific instances.

I also set the experimental flag `false`.

P.S.- Docs tracking: https://github.com/mdn/content/issues/10221

Fixes #12595

### DIFF
--- a/api/SpeechSynthesisEvent.json
+++ b/api/SpeechSynthesisEvent.json
@@ -44,7 +44,7 @@
           }
         },
         "status": {
-          "experimental": true,
+          "experimental": false,
           "standard_track": true,
           "deprecated": false
         }
@@ -141,7 +141,7 @@
             }
           },
           "status": {
-            "experimental": true,
+            "experimental": false,
             "standard_track": true,
             "deprecated": false
           }
@@ -238,9 +238,62 @@
             }
           },
           "status": {
-            "experimental": true,
+            "experimental": false,
             "standard_track": true,
             "deprecated": false
+          }
+        },
+        "milliseconds": {
+          "__compat": {
+            "description": "<code>elapsedTime</code> in milliseconds",
+            "support": {
+              "chrome": {
+                "version_added": "33"
+              },
+              "chrome_android": {
+                "version_added": "33"
+              },
+              "edge": {
+                "version_added": "14"
+              },
+              "firefox": {
+                "version_added": "49",
+                "version_removed": "95",
+                "partial_implementation": true,
+                "notes": "Windows and macOS only."
+              },
+              "firefox_android": {
+                "version_added": false
+              },
+              "ie": {
+                "version_added": false
+              },
+              "opera": {
+                "version_added": "21"
+              },
+              "opera_android": {
+                "version_added": false
+              },
+              "safari": {
+                "version_added": "7",
+                "version_removed": "14.1"
+              },
+              "safari_ios": {
+                "version_added": "7",
+                "version_removed": "14.5"
+              },
+              "samsunginternet_android": {
+                "version_added": "3.0"
+              },
+              "webview_android": {
+                "version_added": false
+              }
+            },
+            "status": {
+              "experimental": false,
+              "standard_track": false,
+              "deprecated": true
+            }
           }
         }
       },
@@ -287,7 +340,7 @@
             }
           },
           "status": {
-            "experimental": true,
+            "experimental": false,
             "standard_track": true,
             "deprecated": false
           }
@@ -336,7 +389,7 @@
             }
           },
           "status": {
-            "experimental": true,
+            "experimental": false,
             "standard_track": true,
             "deprecated": false
           }


api/SpeechSynthesisEvent.json
```diff
@@ -201,37 +201,53 @@
           "spec_url": "https://wicg.github.io/speech-api/#dom-speechsynthesisevent-elapsedtime",
           "support": {
             "chrome": {
-              "version_added": "33"
+              "version_added": "33",
+              "notes": "Units are miliseconds. See <a href='https://crbug.com/1253036'>bug 1253036</a>"
```
hamishwillee, line 205: So I think if something has not been updated it is correct to add a note and a bug link, if known.

api/SpeechSynthesisEvent.json
```diff
@@ -201,37 +201,53 @@
...
             },
             "edge": {
-              "version_added": "14"
-            },
-            "firefox": {
-              "version_added": "49"
-            },
+              "version_added": "14",
+              "notes": "Units are miliseconds. See <a href='https://crbug.com/1253036'>bug 1253036</a>"
```
hamishwillee, line 213: I'm aware of the fact the underlying engine changed, but this should be correct until chromium fixes the bug.

api/SpeechSynthesisEvent.json
```diff
@@ -201,37 +201,53 @@
...
+              {
+              "version_added": "49",
+              "version_removed": "95",
+              "notes": "Units are miliseconds."
+              }
+            ],
             "firefox_android": {
-              "version_added": "62"
+              "version_added": "62",
+              "notes": "From version 95 the units are seconds. Earlier versions return miliseconds. See <a href='https://bugzil.la/1732498'>bug 1732498</a>"
```
hamishwillee, line 228: This is what I am calling _Option 1_ for recording the change. Essentially just a note that reflects the change version.

Option 2 is to highlight this more clearly by having a separate version - see:

api/SpeechSynthesisEvent.json
```diff
@@ -201,37 +201,53 @@
...
             "edge": {
-              "version_added": "14"
-            },
-            "firefox": {
-              "version_added": "49"
-            },
+              "version_added": "14",
+              "notes": "Units are miliseconds. See <a href='https://crbug.com/1253036'>bug 1253036</a>"
+            },
+            "firefox": [
```
hamishwillee, line 215: This is _Option 2_ for recording the change - essentially create a new version and use notes in both to indicate the change. 

_Option 1_ [below](https://github.com/mdn/browser-compat-data/pull/13236/files#r740728356) is a single version with note. 

Option 3, not shown, would be some kind of subfeature.

api/SpeechSynthesisEvent.json
```diff
@@ -201,37 +201,53 @@
...
+              "version_added": "21",
+              "notes": "Units are miliseconds. See <a href='https://crbug.com/1253036'>bug 1253036</a>"
             },
             "opera_android": {
               "version_added": false
             },
             "safari": {
-              "version_added": "7"
+              "version_added": "7",
+              "notes": "Units are in miliseconds from at least version 14.1"
```
hamishwillee, line 242: Not sure what to do here. 14.1.2 is the version that is known to do the right thing, but using seconds might have appeared in an earlier version I guess. It can be tested using https://jsfiddle.net/ybe1c6ot/

api/SpeechSynthesisEvent.json
hamishwillee: PS NOTE, I am aware milliseconds is miss-spelled. I figure I'll fix that globally at the end. 

```suggestion
              "notes": "Units are milliseconds. See <a href='https://crbug.com/1253036'>bug 1253036</a>"
```

hamishwillee:
Note: Update to be deprecated subfeature - `milliseconds`. Just in the Event.

api/SpeechSynthesisEvent.json
```diff
@@ -238,10 +238,62 @@
...
+              "opera_android": {
+                "version_added": false
+              },
+              "safari": {
+                "version_added": "7",
+                "version_removed": "14.1"
+              },
+              "safari_ios": {
+                "version_added": "7",
+                "version_removed": "14.5"
```
hamishwillee, line 282: iOS version matching desktop 14.1

makotokato: Firefox for Linux and Firefox for Android is that `elapsedTime` is second even if old version.  [Bug 1732498](https://bugzilla.mozilla.org/show_bug.cgi?id=1732498) is Windows and macOS only.

api/SpeechSynthesisEvent.json
```diff
@@ -242,6 +242,58 @@
...
+              "edge": {
+                "version_added": "14"
+              },
+              "firefox": {
+                "version_added": "49",
+                "version_removed": "95"
+              },
+              "firefox_android": {
+                "version_added": "62",
+                "version_removed": "95"
```
makotokato, line 265: Firefox for Linux and Firefox for Android is that `elapsedTime` is second even if old version.  [Bug 1732498](https://bugzilla.mozilla.org/show_bug.cgi?id=1732498) is Windows and macOS only.

api/SpeechSynthesisEvent.json
```diff
@@ -242,6 +242,58 @@
...
+              },
+              "chrome_android": {
+                "version_added": "33"
+              },
+              "edge": {
+                "version_added": "14"
+              },
+              "firefox": {
+                "version_added": "49",
+                "version_removed": "95"
```
makotokato, line 261: Firefox for Linux is second even if old version. So you should add the note to explain that this is Windows and macOS implementation only.

ddbeck: @hamishwillee In this case, I'd suggest two support statements:

- The first, with a version added but no version removed, tagged as partial, with a note saying Linux-only.
- The second, with a version added and a version removed, tagged as partial, with a note specifying Windows and macOS.

api/SpeechSynthesisEvent.json
```diff
@@ -238,10 +238,63 @@
...
+            "support": {
+              "chrome": {
+                "version_added": "33"
+              },
+              "chrome_android": {
+                "version_added": "33"
+              },
+              "edge": {
+                "version_added": "14"
+              },
```
hamishwillee, line 258: @makotokato @ddbeck Thanks very much. 

@ddbeck I couldn't understand your suggestion in https://github.com/mdn/browser-compat-data/pull/13236/files#r743560481

But adding the information seemed pretty clear:
1. For FF desktop we support milliseconds from 49 to 95 partial, in Windows and macOS only. For Linux the implication is that the subfeature doesn't apply.
2. Firefox android only supports seconds forever. So for the milliseconds subfeature the support statement is just `false`.

Is that what you meant?

ddbeck: Yes. I misspoke in the original comment. This looks right. 👍

api/SpeechSynthesisEvent.json
```diff
@@ -238,10 +238,63 @@
...
+              "samsunginternet_android": {
+                "version_added": "3.0"
+              },
+              "webview_android": {
+                "version_added": false
+              }
+            },
+            "status": {
+              "experimental": false,
+              "standard_track": true,
```
ddbeck, line 294: This has been removed from the spec, so…

```suggestion
              "standard_track": false,
```

yarn.lock
```diff
@@ -0,0 +1,909 @@
+# THIS IS AN AUTOGENERATED FILE. DO NOT EDIT THIS FILE DIRECTLY.
```
ddbeck, line 1: Delete this file please!

ddbeck (changes requested):
A couple little fixes. Thank you!

yarn.lock
hamishwillee: YOiks. Done.

hamishwillee:
Thanks @ddbeck - makes sense. Should be good to merge now. FYI I squashed my own commits, just because I felt like hiding the screw up with adding yarn.lock :-0

ddbeck (approved):
Looks good. Thank you! 🎉

PR #13255, opened 2021-11-02
Mirror Chrome to Opera + Chrome/Samsung Internet/WebView Android

queengooborg:
This PR mirrors the data from Chrome Desktop to the other Chrome derivatives (particularly Chrome Android, Samsung Internet, and WebView Android), as well as some Opera data.  This PR focuses solely on changes that are straightforward and easy to review.

### DIFF
--- a/api/GamepadButton.json
+++ b/api/GamepadButton.json
@@ -15,9 +15,16 @@
               "version_removed": "34"
             }
           ],
-          "chrome_android": {
-            "version_added": true
-          },
+          "chrome_android": [
+            {
+              "version_added": "35"
+            },
+            {
+              "prefix": "webkit",
+              "version_added": "25",
+              "version_removed": "34"
+            }
+          ],
           "edge": {
             "version_added": "12"
           },
@@ -56,9 +63,16 @@
           "safari_ios": {
             "version_added": "10.3"
           },
-          "samsunginternet_android": {
-            "version_added": true
-          },
+          "samsunginternet_android": [
+            {
+              "version_added": "3.0"
+            },
+            {
+              "prefix": "webkit",
+              "version_added": "1.5",
+              "version_removed": "2.0"
+            }
+          ],
           "webview_android": {
             "version_added": false
           }
@@ -84,9 +98,16 @@
                 "version_removed": "34"
               }
             ],
-            "chrome_android": {
-              "version_added": true
-            },
+            "chrome_android": [
+              {
+                "version_added": "35"
+              },
+              {
+                "prefix": "webkit",
+                "version_added": "25",
+                "version_removed": "34"
+              }
+            ],
             "edge": {
               "version_added": "12"
             },
@@ -125,9 +146,16 @@
             "safari_ios": {
               "version_added": "10.3"
             },
-            "samsunginternet_android": {
-              "version_added": true
-            },
+            "samsunginternet_android": [
+              {
+                "version_added": "3.0"
+              },
+              {
+                "prefix": "webkit",
+                "version_added": "1.5",
+                "version_removed": "2.0"
+              }
+            ],
             "webview_android": {
               "version_added": false
             }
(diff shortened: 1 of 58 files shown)


api/GamepadButton.json
```diff
@@ -84,9 +98,16 @@
...
             ],
-            "chrome_android": {
-              "version_added": true
-            },
+            "chrome_android": [
+              {
+                "version_added": "35"
+              },
+              {
+                "prefix": "webkit",
```
foolip, line 106: I don't trust the data being mirrored here. It's not very likely that both the `GamepadButton` interface and the `pressed` property were webkit-prefixed. Most likely the prefix data should be removed from all sub-features here.

Backing out these changes and doing it separately would be best.

api/GamepadEvent.json
```diff
@@ -189,9 +196,16 @@
...
             },
-            "samsunginternet_android": {
-              "version_added": true
-            },
+            "samsunginternet_android": [
+              {
+                "version_added": "3.0"
+              },
+              {
+                "prefix": "webkit",
```
foolip, line 204: Same here, I don't trust the data being mirrored, this `gamepad` property probably wasn't prefixed.

api/HTMLMediaElement.json
```diff
@@ -2209,7 +2212,7 @@
               "version_added": "43"
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": "43"
```
foolip, line 2215: This copying incorrect data, a case of https://github.com/mdn/browser-compat-data/issues/7843. Can back out these changes you fix all of the "43" values in this file, updating both Chrome and mirrored data at the same time?

api/IDBKeyRange.json
```diff
@@ -50,12 +57,26 @@
...
-            "version_added": true
-          }
+          "samsunginternet_android": [
+            {
+              "version_added": "1.5"
+            },
+            {
+              "version_added": "1.5",
+              "version_removed": "7.0",
+              "prefix": "webkit"
```
foolip, line 67: This prefixed support should be on Opera as well. For consistency, can you add it there (my preference) or alternatively skip it here too?

api/IDBVersionChangeEvent.json
```diff
@@ -49,9 +55,15 @@
...
-          "samsunginternet_android": {
-            "version_added": true
-          },
+          "samsunginternet_android": [
+            {
+              "version_added": "1.5"
+            },
+            {
+              "version_added": "1.0",
+              "prefix": "webkit"
```
foolip, line 64: Can you add this for Opera too?

api/Metadata.json
```diff
@@ -38,7 +38,7 @@
             "version_added": false
           },
           "samsunginternet_android": {
-            "version_added": true,
+            "version_added": "1.0",
             "prefix": "webkit"
```
foolip, line 42: This is missing from Opera, but in this case the Opera data is false, so wrong in a more easily detectable way. It's fine to leave that wrong for now.

foolip (changes requested):
Found some things... backing out those changes and landing the rest would be most expedient.

api/GamepadButton.json
queengooborg: Opened #13269 to take care of the gamepad data!

api/HTMLMediaElement.json
queengooborg: This is really odd, running the collector on Chrome 1 shows support all the way back then, but the collector isn't updating this data...  I'll make a note of this in the related collector's issue.

foolip (approved):
I've used https://github.com/mdn/browser-compat-data/pull/6862 to look at the results in a more aggregated way and spot checked some things without finding any additional issues.

No doubt some of the data being mirrored here is correct, so if anyone finds this PR in the future, don't trust it any more than you would the Chrome data being mirrored.

PR #13256, opened 2021-11-02
Remove entry for Node.getFeature

teoli2003:
This method was only supported by Opera, and removed in… Opera 15.

It will not be mentioned on MDN either (the page doesn't exist, mention in Node is about to be removed)

Time to go…

### DIFF
--- a/api/Node.json
+++ b/api/Node.json
@@ -402,56 +402,6 @@
           }
         }
       },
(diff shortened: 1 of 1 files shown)


queengooborg (approved):
+1 for removal, no need to keep data for this feature!

PR #3971, opened 2019-04-24
Update css.properties.position.fixed/sticky

vinyldarkscratch:
I found `sticky` positioning a few moments ago, and immediately fell in love with it, so I [integrated it into my company's website](https://www.gooborg.com/team).  I decided to show it off to my boyfriend, but when he couldn't find out what I was talking about, we realized that `sticky` positioning simply isn't supported in Firefox ~~Android~~.  So, I decided to do some digging around to check on `fixed` and `sticky` positioning on all the browsers.

This PR reflects manual checks on the properties during my development of my studio's website.

PR #13259, opened 2021-11-02
menus.OnClickedData.srcUrl returned post-redirect URL

rebloor:
#### Summary

Add details for the `srcUrl` property to `menus.OnClickedData` including a note that between versions 59 and 94 it incorrectly returned the post-redirect URL rather than the `` value.

`npm test` executed with no errors

#### Related issues

Provide complementary browser compatibility data for [menus.OnClickedData.srcUrl returns  #10214](https://github.com/mdn/content/pull/10214) which addresses the documentation requirements for [Bug 1659155](https://bugzilla.mozilla.org/show_bug.cgi?id=1659155).

### DIFF
--- a/webextensions/api/menus.json
+++ b/webextensions/api/menus.json
@@ -454,7 +454,7 @@
                   {
                     "version_added": "59",
                     "version_removed": "94",
-                    "notes": "Returns a post-redirect URL rather than the raw <code>&lt;img src&gt;<code> value."
+                    "notes": "Returns a post-redirect URL rather than the raw value of the <code>src<code> attribute of the clicked element."
                   }
                 ],
                 "firefox_android": {
--- a/webextensions/api/menus.json
+++ b/webextensions/api/menus.json
@@ -438,6 +438,37 @@
               }
             }
           },
+          "srcUrl": {
+            "__compat": {
+              "support": {
+                "chrome": {
+                  "version_added": true
+                },
+                "edge": {
+                  "version_added": "14"
+                },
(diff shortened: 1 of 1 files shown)


webextensions/api/menus.json
```diff
@@ -438,6 +438,37 @@
...
+                  "version_added": "14"
+                },
+                "firefox": [
+                  {
+                    "version_added": "48"
+                  },
+                  {
+                    "version_added": "59",
+                    "version_removed": "94",
+                    "notes": "Returns a post-redirect URL rather than the raw <code>&lt;img src&gt;<code> value."
```
Rob--W, line 457: Not limited to ``, please make this more generic. See https://github.com/mdn/content/pull/10214#pullrequestreview-795673940

Rob--W (changes requested):

Rob--W (approved):
Once merged, double-check on MDN (after a week?) whether the rendered compat table matches with the intended message. I vaguely recall that the use of `version_removed` used to render as something different than what I meant.

rebloor:
@Rob--W I have a similar recollection, will check

rebloor:
@Rob--W 
> Once merged, double-check on MDN (after a week?) whether the rendered compat table matches with the intended message. I vaguely recall that the use of version_removed used to render as something different than what I meant.
There is still an inconsistency:
image
but confirm there is a Yari issue for this (Feature removals in BCD use a misleading version range [#3971](https://github.com/mdn/yari/issues/3971).)

PR #13263, opened 2021-11-02
Mirror Safari to Safari iOS

queengooborg:
This PR mirrors Safari to Safari iOS, focusing on the straightforward changes.

### DIFF
--- a/api/GeolocationPositionError.json
+++ b/api/GeolocationPositionError.json
@@ -67,7 +67,7 @@
           },
           "safari_ios": {
             "alternative_name": "PositionError",
-            "version_added": "5"
+            "version_added": "4.2"
           },
           "samsunginternet_android": [
             {
@@ -129,7 +129,7 @@
               "version_added": "5"
             },
             "safari_ios": {
-              "version_added": "5"
+              "version_added": "4.2"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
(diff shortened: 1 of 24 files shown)


api/GeolocationPositionError.json
```diff
@@ -67,7 +67,7 @@
           },
           "safari_ios": {
             "alternative_name": "PositionError",
-            "version_added": true
+            "version_added": "5"
```
foolip, line 70: Safari 5 should map to 4.2 using the mapping in https://github.com/mdn/browser-compat-data/pull/11299.

Can you (somehow) fix the mirroring script to match https://github.com/mdn/browser-compat-data/pull/11299 and update this PR?

foolip (changes requested):

api/GeolocationPositionError.json
queengooborg: I've updated this PR with a manual fix (`GeolocationPositionError` was the only affected file in this PR).  I'm hoping to get the other mirroring script changes landed first before opening more to prevent merge conflicts (particularly since some of my own PRs change the logic flow of the script a bit).

foolip:
@queengooborg are you sure that 5&rarr;4.2 was the only incorrect mapping here? It was the first that I noticed, but I didn't check everything. Is there a PR open to fix the mirroring script? Testing with that PR should reveal which mappings have changed, i.e., which versions to search for in this PR.

queengooborg:
That was the only incorrect mapping, yeah!  I don't have an open PR to fix the mirroring script's Safari iOS mapping yet.

foolip (approved):

PR #13272, opened 2021-11-03
Add Chromium versions for XMLHttpRequest API

queengooborg:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `XMLHttpRequest` API by mirroring the data.


api/XMLHttpRequest.json
```diff
@@ -1015,7 +1021,7 @@
                 "version_added": "31"
               },
               "chrome_android": {
-                "version_added": true
+                "version_added": "55"
```
foolip, line 1024: What data is this mirroring? I would expect "31" here.

queengooborg: This was to fix a consistency issue, but another merged PR already updates this data and sets it to "31", which seems more reasonable!

api/XMLHttpRequest.json
```diff
@@ -985,9 +985,15 @@
                 "version_removed": "15"
               }
             ],
-            "opera_android": {
-              "version_added": "42"
-            },
+            "opera_android": [
+              {
+                "version_added": "42"
```
foolip, line 990: This doesn't seem right, it doesn't match Opera 18 above. Some mirroring has gone wrong here.

api/XMLHttpRequest.json
```diff
@@ -1038,9 +1044,15 @@
                   "version_removed": "15"
                 }
               ],
-              "opera_android": {
-                "version_added": true
-              },
+              "opera_android": [
+                {
+                  "version_added": "42"
```
foolip, line 1049: This also doesn't add up with the above data.

api/XMLHttpRequest.json
```diff
@@ -1083,12 +1095,24 @@
...
+                  "version_added": "18"
+                },
+                {
+                  "version_added": "12",
+                  "version_removed": "15"
+                }
+              ],
+              "opera_android": [
+                {
+                  "version_added": "42"
```
foolip, line 1109: This also doesn't add up with the above data.

api/XMLHttpRequest.json
```diff
@@ -1131,12 +1155,24 @@
...
```
foolip, line 1169: This also doesn't add up with the above data.

api/XMLHttpRequest.json
```diff
@@ -1188,9 +1224,15 @@
                   "version_removed": "15"
                 }
               ],
-              "opera_android": {
-                "version_added": null
-              },
+              "opera_android": [
+                {
+                  "version_added": "42"
```
foolip, line 1229: This also doesn't add up with the above data.

foolip (changes requested):

api/XMLHttpRequest.json
queengooborg: Opera Android was originally set to "42" for this feature so I stuck with that.  However, I realize that doesn't really make sense for a feature like this one, so I'm replacing "42" with "18".

foolip (approved):

sophiaprincess002 (commented):
153a95d

PR #13300, opened 2021-11-05
Fixes #13299 - Firefox for Android doesn't support permissions.request()

### LINKED ISSUE #13299: webextensions.api.permissions.request - Not supported by Firefox for Android

[permissions.request()](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/permissions/request) says that it can be used in Firefox for Android. However, according to [Bugzilla #1601420](https://bugzilla.mozilla.org/show_bug.cgi?id=1601420) and [Fenix #16912](https://github.com/mozilla-mobile/fenix/issues/16912), `permissions.request()` is not implemented.

An "X" in the Firefox for Android column

Created my own experimental browser extension for Firefox for Android, which didn't work when requesting additional permissions, then found the existing github issues.

<details>

* Query: `webextensions.api.permissions.request`
* MDN URL: https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/permissions/request
* Report started: 2021-11-05T11:32:47.567Z

</details>

diafygi:
#### Summary

Correctly say that Firefox for Android doesn't support `browser.permissions.request()` (see #13299)

[permissions.request()](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/permissions/request) says that it can be used in Firefox for Android. However, according to [Bugzilla #1601420](https://bugzilla.mozilla.org/show_bug.cgi?id=1601420) and [Fenix #16912](https://github.com/mozilla-mobile/fenix/issues/16912), `permissions.request()` is not implemented.

#### Related issues

Fixes #13299

### DIFF
--- a/webextensions/api/permissions.json
+++ b/webextensions/api/permissions.json
@@ -174,7 +174,7 @@
                 ]
               },
               "firefox_android": {
-                "version_added": true
+                "version_added": false
               },
               "opera": {
                 "version_added": true
--- a/webextensions/api/permissions.json
+++ b/webextensions/api/permissions.json
@@ -174,7 +174,9 @@
                 ]
               },
               "firefox_android": {
-                "version_added": false
+                "version_added": true,
+                "partial_implementation": true,
+                "notes": "Since version 79, the UI to approve the permission request is missing (<a href='https://bugzil.la/1601420'>bug 1601420</a>)."
               },
               "opera": {
                 "version_added": true
--- a/webextensions/api/permissions.json
+++ b/webextensions/api/permissions.json
@@ -176,7 +176,7 @@
               "firefox_android": {
                 "version_added": true,
                 "partial_implementation": true,
-                "notes": "Since version 79, the UI to approve the permission request is missing (<a href='https://bugzil.la/1601420'>bug 1601420</a>)."
+                "notes": "From version 79, the user interface to approve the permission request is missing. See <a href='https://bugzil.la/1601420'>bug 1601420</a>."
               },
               "opera": {
                 "version_added": true


webextensions/api/permissions.json
```diff
@@ -174,7 +174,7 @@
                 ]
               },
               "firefox_android": {
-                "version_added": true
+                "version_added": false
```
Rob--W, line 177: The method is available and was supported until version 68.

The only issue is that there is no UI with a prompt. If an optional permission has already been granted, then there won't be a prompt. So the API can still be marked as supported, and you should use:
```suggestion
                "version_added": true,
                "partial_implementation": true,
                "notes": "Since version 79, the UI to approve the permission request is missing (<a href='https://bugzil.la/1601420'>bug 1601420</a>)."
```

Rob--W (changes requested):

webextensions/api/permissions.json
diafygi: I feel like the entire point of `permissions.request` is so the user is prompted if the permission if not already given, so saying the API is partially implemented as `true` is pretty misleading. Could we have `"version_added": false` and still have the partial implementation asterisk and note?

Rob--W: `version_added: false` implies that it was never supported. If any, `version_removed` should be added.

From the extension's point of view, the API is supported. It validates parameters and may sometimes even return something. There have been bugs where the API did not work as desired (see e.g. https://bugzilla.mozilla.org/show_bug.cgi?id=1433292#c14). In these cases, the API itself is still supported, but buggy. For that, the `notes` field fits perfectly.

diafygi: Thanks for the clarification!

Rob--W (approved):

webextensions/api/permissions.json
```diff
@@ -174,7 +174,9 @@
                 ]
               },
               "firefox_android": {
-                "version_added": true
+                "version_added": true,
+                "partial_implementation": true,
+                "notes": "Since version 79, the UI to approve the permission request is missing (<a href='https://bugzil.la/1601420'>bug 1601420</a>)."
```
ddbeck, line 179: There's not an actual problem here, just some style stuff to make this more consistent with other notes.

```suggestion
                "notes": "From version 79, the user interface to approve the permission request is missing. See <a href='https://bugzil.la/1601420'>bug 1601420</a>."
```

ddbeck (changes requested):
Thanks for this PR! Some style suggestions for the notes, before we merge. Thank you again!

ddbeck:
Also, I see this is your first PR. Welcome, @diafygi!

diafygi:
@ddbeck added suggested changes!

ddbeck (approved):
Thank you! 🎉

PR #13304, opened 2021-11-06
Add Safari iOS versions for TouchEvent API

queengooborg:
This PR adds real values for Safari iOS/iPadOS for the `TouchEvent` API, based upon commit history and date.

Commit: https://github.com/WebKit/WebKit/commit/ce64d715f62fe76b229b937dcf95f7b39baa6cc7 / https://github.com/WebKit/WebKit/commit/5f9e450ef87e941d63891f08a6ff986ea92e889e

The "3.2" came from the initial data for `api.TouchEvent.TouchEvent`.  `TouchEvent` was undefined in Safari iOS 3 but defined in Safari iOS 4, so this checks out.

### DIFF
--- a/api/TouchEvent.json
+++ b/api/TouchEvent.json
@@ -53,7 +53,7 @@
             "version_added": false
           },
           "safari_ios": {
-            "version_added": true
+            "version_added": "3.2"
           },
           "samsunginternet_android": {
             "version_added": "1.5"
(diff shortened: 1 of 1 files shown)


foolip:
> `TouchEvent` was undefined in Safari iOS 3 but defined in Safari iOS 4, so this checks out.

Thanks for double checking this! For touch events, I wouldn't trust source alone since this is an area where Safari has had patches on top of what's in WebKit trunk.

foolip (approved):

PR #13322, opened 2021-11-07
Add window.open popup feature and BarProp.visible value update

arai-a:
for https://github.com/whatwg/html/pull/6530

#### Summary

Add browser compat data for updated `windowFeatures` parameter behavior for `window.open`, and also `BarProp.visible` behavior, from the spec change in https://github.com/whatwg/html/pull/6530.

The behavior is tested by https://github.com/web-platform-tests/wpt/pull/28243

Firefox: fixed in 96 (currently Nightly) https://bugzilla.mozilla.org/show_bug.cgi?id=1701001
Chrome: fixed in 98 (current Canary) https://bugs.chromium.org/p/chromium/issues/detail?id=1192701
Safari: bug is filed https://bugs.webkit.org/show_bug.cgi?id=223751

#### Related issues

* https://github.com/mdn/content/pull/10339

### DIFF
--- a/api/BarProp.json
+++ b/api/BarProp.json
@@ -95,6 +95,54 @@
             "standard_track": true,
             "deprecated": false
           }
+        },
+        "returns_popup": {
+          "__compat": {
+            "description": "Returns <code>true</code> for non-popup windows",
+            "support": {
+              "chrome": {
+                "version_added": "98"
+              },
+              "chrome_android": {
+                "version_added": "98"
+              },
+              "edge": {
+                "version_added": "98"
+              },
+              "firefox": {
+                "version_added": "96"
+              },
+              "firefox_android": {
+                "version_added": "96"
+              },
+              "ie": {
+                "version_added": false
+              },
+              "opera": {
+                "version_added": false
+              },
+              "opera_android": {
+                "version_added": false
+              },
+              "safari": {
+                "version_added": false
+              },
+              "safari_ios": {
+                "version_added": false
+              },
+              "samsunginternet_android": {
+                "version_added": false
+              },
+              "webview_android": {
+                "version_added": "98"
+              }
+            },
+            "status": {
+              "experimental": false,
+              "standard_track": true,
+              "deprecated": false
+            }
+          }
         }
       }
     }
--- a/api/Window.json
+++ b/api/Window.json
@@ -5333,6 +5333,54 @@
             "deprecated": false
           }
         },
+        "features_parameter_popup": {
+          "__compat": {
+            "description": "<code>features</code> parameter accepts <code>\"popup\"</code> value",
+            "support": {
+              "chrome": {
+                "version_added": "98"
+              },
+              "chrome_android": {
+                "version_added": "98"
+              },
+              "edge": {
+                "version_added": "98"
+              },
+              "firefox": {
+                "version_added": "96"
+              },
+              "firefox_android": {
+                "version_added": "96"
+              },
+              "ie": {
+                "version_added": false
+              },
+              "opera": {
+                "version_added": false
+              },
+              "opera_android": {
+                "version_added": false
+              },
+              "safari": {
+                "version_added": false
+              },
+              "safari_ios": {
+                "version_added": false
+              },
+              "samsunginternet_android": {
+                "version_added": false
+              },
+              "webview_android": {
+                "version_added": "98"
+              }
+            },
+            "status": {
+              "experimental": false,
+              "standard_track": true,
+              "deprecated": false
+            }
+          }
+        },
         "once_per_event": {
           "__compat": {
             "description": "One <code>Window.open()</code> call per event",


api/BarProp.json
```diff
@@ -48,6 +48,54 @@
           "deprecated": false
         }
       },
+      "is_popup": {
```
ddbeck, line 51: I can't find this in the IDL for the `BarProp` interface, which makes me think this is not actually exposed to web developers directly. Instead, it seems to change the value of `visible`. Is that right?

If it is, I think the right way to go is to characterize this is as a change in the way `visible` works, rather than introducing a new feature entry here. I'm not sure how best to represent this change in Firefox 96, however. I think the core question is: before Firefox 96, what did `visible` return and what does it return from 96?

api/Window.json
```diff
@@ -5429,6 +5429,54 @@
               "deprecated": true
             }
           }
+        },
+        "popup_feature": {
```
ddbeck, line 5433: The situation here is a bit complex because the other features here aren't structured very conventionally. Sorry about that.

The way forward here is to rename this feature as `features_parameter_popup` and a description like `<code>features</code> parameter accepts <code>"popup"</code> value`(following [the convention documented in our data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#parameters-and-parameter-object-features)). You'll probably have to re-sort the file here, too.

(The MDN page wrongly identifies the parameter as `windowFeatures`, but let's overlook that for now.)

api/Window.json
```diff
@@ -5429,6 +5429,54 @@
...
+              },
+              "samsunginternet_android": {
+                "version_added": false
+              },
+              "webview_android": {
+                "version_added": false
+              }
+            },
+            "status": {
+              "experimental": false,
```
ddbeck, line 5475: Nearly all single implementation features are experimental. See the [_Choosing an experimental status_ data guideline](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#choosing-an-experimental-status) for details.

```suggestion
              "experimental": true,
```

ddbeck (changes requested):
Thanks for starting this PR, @arai-a. I don't completely understand this change, so I've got a few line comments to get a conversation started on how best to represent this information. Thank you again!

api/BarProp.json
arai-a: yes, the change is the value of `visible`.

https://html.spec.whatwg.org/multipage/window-object.html#dom-barprop-visible

The change is that:
  * before Firefox 96, `BarProp.value` was returning the actual visibility of each UI part (toolbar, menubar, etc), with exceptions that `scrollbars.visible` is always true
  * after Firefox 96, `BarProp.value` returns the negation of "is popup" of the top-level browsing context, regardless of the actual visibility of each UI parts

"is popup" of the top-level browsing context defaults to false, and is calculated from `features` parameter of `window.open`.


currently other browsers return different value for each, for `BarProp.value`.
Safari also returns the actual visibility of each UI parts, but with yet another exceptions.
Chrome returns whether each UI parts are enabled in `features` parameters of `window.open`, regardless of the actual UI parts visibility.

api/BarProp.json
```diff
@@ -96,6 +96,54 @@
             "deprecated": false
           }
         }
+      },
+      "visible_popup": {
```
ddbeck, line 100: Two changes needed here:

- This should be a subfeature of visible (that is, it should be nested one step deeper, as a sibling of the `__compat` above)
- Let's rename this to "returns_popup" (it's a specified as a getter, so I feel OK using this terminology)

```suggestion
      "returns_popup": {
```

api/BarProp.json
```diff
@@ -96,6 +96,54 @@
...
+        "__compat": {
+          "description": "<code>visible</code> represents popup or not",
```
ddbeck, line 102: This is really challenging to explain in just a couple of words! I'm going to suggest this. It doesn't tell the whole story, but I think it gets us pretty close. Open to further suggestions, if you like.

```suggestion
          "description": "Returns <code>true</code> for non-popup windows",
```

ddbeck (changes requested):
Thanks for the follow up, @arai-a. A couple of suggestions for you, but we're nearly there.

arai-a:
oh, wait, chromium have implemented this feature as well. so I'll reflect the change here.
give me a moment to gather information

arai-a:
Also, given now that 2 browsers have implemented and it's in the spec itself, I think it's fine to flip "experimental" to false

arai-a:
* Reflected chromium/edge support, from 98
 * Flipped `experimental` to `false`

https://groups.google.com/a/chromium.org/g/blink-dev/c/q6ybnmxxvpE
> Estimated milestones
> M98

ddbeck (approved):
Looks good. Thank you! 🎉

PR #13328, opened 2021-11-07
Add Chromium versions for api.HTMLInputElement.capture

queengooborg:
This PR adds real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `capture` member of the `HTMLInputElement` API, based upon commit history and date.

Commit: https://source.chromium.org/chromium/chromium/src/+/13bfc68b02ee2fe60514db8cb340b5aea49d8c8e

### DIFF
--- a/api/HTMLInputElement.json
+++ b/api/HTMLInputElement.json
@@ -244,7 +244,7 @@
               "version_added": false
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": "25"
             },
             "edge": {
               "version_added": false
@@ -271,10 +271,10 @@
               "version_added": true
             },
             "samsunginternet_android": {
-              "version_added": true
+              "version_added": "1.5"
             },
             "webview_android": {
-              "version_added": true
+              "version_added": "≤37"
             }
           },
           "status": {


foolip (commented):
Can you confirm this at least approximately with https://mdn-bcd-collector.appspot.com/tests/api/HTMLInputElement/capture? I don't really trust the source much for something that's controlled by a flag.

queengooborg:
Unfortunately I'm unable to run the collector on earlier Chrome Android versions yet -- I used the collector to confirm this was supported, but I'm only able to use the source code to determine the version number.

foolip:
I see. In that case, can you compare the data to other features which this attribute is related to, to at least ensure consistency?

queengooborg:
Actually, since there's no compatibility inconsistencies between the individual properties and the main feature itself, there's no reason to keep the subfeatures!

queengooborg:
Sadly this feature is the only one of its spec; there are no others like it.  This guesstimation is the best we can do... :(

foolip:
I found the place where the flag is enabled, and it was enabled for Android at the time the `capture` attribute was added:
https://source.chromium.org/chromium/chromium/src/+/13bfc68b02ee2fe60514db8cb340b5aea49d8c8e:third_party/WebKit/Source/WebKit/chromium/features.gypi;l=140-147

api/HTMLInputElement.json
```diff
@@ -244,7 +244,7 @@
               "version_added": false
             },
             "chrome_android": {
-              "version_added": true
+              "version_added": "25"
```
foolip, line 247: The commit date was 2012-05-10 and using https://www.chromium.org/developers/calendar that maps to Chrome 21:

```suggestion
              "version_added": "21"
```

foolip (changes requested):

api/HTMLInputElement.json
queengooborg: Ah great, thanks for tracking this down!  There was no Chrome release in between 18 and 25, so if the commit maps to Chrome 21, we'd bump it to 25.

foolip: Oh, right :)

foolip (approved):

PR #13337, opened 2021-11-07
Update Edge versions for RTCRtpSender API

queengooborg:
This PR updates and corrects the real values for Microsoft Edge for the `RTCRtpSender` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v4.0.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/RTCRtpSender

### DIFF
--- a/api/RTCRtpSender.json
+++ b/api/RTCRtpSender.json
@@ -12,7 +12,7 @@
             "version_added": "64"
           },
           "edge": {
-            "version_added": "12"
+            "version_added": "13"
           },
           "firefox": {
             "version_added": "34"
@@ -107,7 +107,7 @@
               "version_added": "66"
             },
             "edge": {
-              "version_added": "≤18"
+              "version_added": "79"
             },
             "firefox": {
               "version_added": "52"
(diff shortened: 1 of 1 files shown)


api/RTCRtpSender.json
```diff
@@ -107,7 +107,7 @@
               "version_added": "66"
             },
             "edge": {
-              "version_added": "≤18"
+              "version_added": "79"
```
foolip, line 110: This matches the data for `RTCDTMFSender` 👍

api/RTCRtpSender.json
```diff
@@ -355,7 +355,7 @@
               "version_added": "75"
             },
             "edge": {
-              "version_added": "12",
+              "version_added": "13",
               "version_removed": "79"
```
foolip, line 359: This `version_removed` has to be wrong, can you double check?

api/RTCRtpSender.json
```diff
@@ -565,7 +565,7 @@
...
```
foolip, line 569: This also looks wrong.

foolip (changes requested):

api/RTCRtpSender.json
queengooborg: Looks like these aren't covered by the collector, but they are indeed wrong.  Fixing!

foolip (approved):

PR #13353, opened 2021-11-08
FF95: Add CSS cursor support on Android

hamishwillee:
FF95 adds support for CSS [cursor](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor#browser_compatibility) on android. I've simply marked all of the FF for android fields as "95" rather than false.

The associated bug is https://bugzilla.mozilla.org/show_bug.cgi?id=1672609

Docs work tracked in https://github.com/mdn/content/issues/10145

### DIFF
--- a/css/properties/cursor.json
+++ b/css/properties/cursor.json
@@ -20,7 +20,7 @@
               "notes": "Starting in Firefox 67, the maximum size allowed for custom cursors is 32x32 pixels due to cursors being misused by certain malicious sites."
             },
             "firefox_android": {
-              "version_added": false
+              "version_added": "95"
             },
             "ie": {
               "version_added": "4",
@@ -68,7 +68,7 @@
                 "version_added": "1.5"
               },
               "firefox_android": {
-                "version_added": false
+                "version_added": "95"
               },
               "ie": {
                 "version_added": "10"
@@ -116,7 +116,7 @@
                 "version_added": "1.5"
               },
               "firefox_android": {
-                "version_added": false
+                "version_added": "95"
               },
               "ie": {
                 "version_added": "6"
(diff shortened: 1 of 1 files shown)


css/properties/cursor.json
```diff
@@ -359,7 +359,7 @@
                 "notes": "This cursor is only supported on macOS and Linux."
               },
               "firefox_android": {
-                "version_added": false
+                "version_added": "95"
```
hamishwillee, line 362: @makotokato This is the feature for `context-menu`. Above it says in chrome for android ` "notes": "This cursor is only supported on macOS and Linux."`

My _guess_ is that context menus are not supported on Android so actually all of the android platforms should be set to false, including this one for FF on Android. Do you happen to know?

hamishwillee:
[Using URL values for the cursor property](https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Basic_User_Interface/Using_URL_values_for_the_cursor_property) looks to be a VERY old doc covering compatibility back circa Firefox 2. Most of which is pointless or in [`cursor`](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor).

I plan to delete this. Is any of it useful at all? Can we capture the supported file types in BCD, and if so, what's the best way to do that? If not, then I suggest the table of supported stuff gets copied to [`cursor`](https://developer.mozilla.org/en-US/docs/Web/CSS/cursor)

css/properties/cursor.json
makotokato: Firefox Android supports `context-menu` cursor type. Windows isn't still yet (https://bugzilla.mozilla.org/show_bug.cgi?id=258960)

queengooborg (approved):
LGTM, thanks!

PR #13355, opened 2021-11-08
FF95 ships inputmode global attribute

hamishwillee:
FF95 ships the [input mode global attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/inputmode) in bug https://bugzilla.mozilla.org/show_bug.cgi?id=1205133 by making the `dom.forms.inputmode` always true.

This updates BCD data for affected elements against that key.

### DIFF
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -1540,16 +1540,22 @@
             "edge": {
               "version_added": "79"
             },
-            "firefox": {
-              "version_added": "77",
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "dom.forms.inputmode",
-                  "value_to_set": "true"
-                }
-              ]
-            },
+            "firefox": [
+              {
+                "version_added": "95",
+              },
+              {
+                "version_added": "77",
+                "version_removed": "95",
+                "flags": [
+                  {
+                    "type": "preference",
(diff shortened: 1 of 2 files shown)


html/global_attributes.json
```diff
@@ -1140,6 +1144,7 @@
               },
               {
                 "version_added": "68",
+                "version_removed": "79",
```
hamishwillee, line 1147: Note, this appears to have been an omission when the preference was disabled for android back in v79

api/HTMLElement.json
```diff
@@ -1540,16 +1540,22 @@
...
-              "version_added": "77",
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "dom.forms.inputmode",
-                  "value_to_set": "true"
-                }
-              ]
-            },
+            "firefox": [
```
hamishwillee, line 1543: Note, I can't find anything about `HTMLElement.inputMode` in the spec - am I missing anything? It "feels" like this perhaps is an old name for the new global attribute, and perhaps has no independent existance?

makotokato: https://html.spec.whatwg.org/#editing-2 has WebIDL defines

queengooborg (approved):

PR #13398, opened 2021-11-09
Add missing api.Device[Motion/Orientation]Event.requestPermission feature

queengooborg:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing `requestPermission` member of the DeviceMotionEvent and DeviceOrientationEvent APIs, populating the results using commit history and date.

Commit: https://github.com/WebKit/WebKit/commit/1978c01a68fb5d4a8d7a0c3becb685f3364150ae#diff-cc63b7286aced27c9467c2e4466e43c0ba3960ee6991e59cefc66b06d30ea2ce ([WebKit 611.1.1](https://github.com/WebKit/WebKit/blob/1978c01a68fb5d4a8d7a0c3becb685f3364150ae/Source/WebCore/Configurations/Version.xcconfig))

### DIFF
--- a/api/DeviceMotionEvent.json
+++ b/api/DeviceMotionEvent.json
@@ -245,6 +245,54 @@
           }
         }
       },
+      "requestPermission": {
+        "__compat": {
+          "spec_url": "https://w3c.github.io/deviceorientation/#dom-devicemotionevent-requestpermission",
+          "support": {
+            "chrome": {
+              "version_added": false
+            },
+            "chrome_android": {
+              "version_added": false
+            },
+            "edge": {
+              "version_added": false
+            },
+            "firefox": {
+              "version_added": false
+            },
+            "firefox_android": {
+              "version_added": false
+            },
+            "ie": {
+              "version_added": false
+            },
+            "opera": {
+              "version_added": false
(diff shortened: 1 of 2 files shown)


foolip:
Can you confirm this with real device testing? `requestPermission()` was in Source/WebCore/dom/DeviceOrientationOrMotionEvent.idl before https://github.com/WebKit/WebKit/commit/1978c01a68fb5d4a8d7a0c3becb685f3364150ae and also the `EnabledBySetting=DeviceOrientationPermissionAPI` extended attribute makes it impossible to have confidence based on commit history alone, as the settings for Safari aren't public.

foolip:
> settings for Safari aren't public

Actually maybe they are to a large degree, but in any case I don't know where to find them and have confidence in them :)

queengooborg:
Running the [collector's test](https://mdn-bcd-collector.appspot.com/tests/api/DeviceMotionEvent/requestPermission) in iOS 14 and 14.5 shows support was added in Safari 14.5!

foolip (approved):

PR #13424, opened 2021-11-10
Remove api.Range.insertNode.collapsed_ranges

foolip:
It's not clear what this sub-feature means from just the description,
and all we have is a guess:
https://github.com/mdn/browser-compat-data/pull/13075#issuecomment-954756548

The data matches the parent feature except for Firefox and Safari. This
probably means there's *something* that doesn't work the same before
Firefox 14 and Safari 4. This is far enough in the past that it's not
worth trying to salvage this sub-feature with more research.

### DIFF
--- a/api/Range.json
+++ b/api/Range.json
@@ -1002,54 +1002,6 @@
             "standard_track": true,
             "deprecated": false
           }
-        },
-        "collapsed_ranges": {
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
Agreed

PR #13443, opened 2021-11-10
IMP: update data for iOS safari for -webkit-scrollbar

### LINKED ISSUE #9967: ::-webkit-scrollbar - iOS removed support in 14+
Apparently all support for custom scrollbars has been removed from iOS 14+? https://developer.apple.com/forums/thread/670065

<details>

* Query: `css.selectors.-webkit-scrollbar-button`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar
* Report started: 2021-04-19T12:56:58.810Z

</details>

GrimLink:
#### Summary

This marks the removal of the CSS Selectors for the -webkit-scrollbar from iOS safari 14+
Since this does not work anymore on iOS safari 14+

Tested websites using an custom scrollbar like, [css-tricks.com](https://css-tricks.com/) and [fylgja.dev](https://fylgja.dev/components/scrollbar/), on iOS device

Sadly there is no mention of this in the official release notes for Safari.

Only this https://developer.apple.com/forums/thread/670065
And this issue found here -> https://github.com/mdn/browser-compat-data/issues/9967

#### Related issues

Fixes #9967

### DIFF
--- a/css/selectors/-webkit-scrollbar-button.json
+++ b/css/selectors/-webkit-scrollbar-button.json
@@ -36,7 +36,8 @@
               "version_added": "4"
             },
             "safari_ios": {
-              "version_added": "3"
+              "version_added": "3",
+              "version_removed": "14"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
--- a/css/selectors/-webkit-scrollbar-button.json
+++ b/css/selectors/-webkit-scrollbar-button.json
@@ -37,7 +37,7 @@
             },
             "safari_ios": {
               "version_added": "3",
-              "version_removed": "14"
+              "version_removed": "13"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
--- a/css/selectors/-webkit-scrollbar.json
+++ b/css/selectors/-webkit-scrollbar.json
@@ -37,7 +37,7 @@
             },
             "safari_ios": {
               "version_added": "3",
-              "version_removed": "14"
+              "notes": "As of Safari for iOS 13 only `display: none` will work, any other styles have no effect."
             },
             "samsunginternet_android": {
               "version_added": "1.0"
--- a/css/selectors/-webkit-scrollbar.json
+++ b/css/selectors/-webkit-scrollbar.json
@@ -36,7 +36,8 @@
               "version_added": "4"
             },
             "safari_ios": {
-              "version_added": "3"
+              "version_added": "3",
+              "version_removed": "14"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
--- a/css/selectors/-webkit-scrollbar.json
+++ b/css/selectors/-webkit-scrollbar.json
@@ -37,7 +37,7 @@
             },
             "safari_ios": {
               "version_added": "3",
-              "notes": "As of Safari for iOS 13 only `display: none` will work, any other styles have no effect."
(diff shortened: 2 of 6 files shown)


css/selectors/-webkit-scrollbar.json
```diff
@@ -36,7 +36,8 @@
               "version_added": "4"
             },
             "safari_ios": {
-              "version_added": "3"
+              "version_added": "3",
+              "version_removed": "14"
```
lukewarlow, line 40: It would be interesting to see if setting display: none on this property still works? I suspect it probably does?

ddbeck: I tested this in BrowserStack. `display: none` does work. We should probably not apply this change.

css/selectors/-webkit-scrollbar-button.json
```diff
@@ -36,7 +36,8 @@
               "version_added": "4"
             },
             "safari_ios": {
-              "version_added": "3"
+              "version_added": "3",
+              "version_removed": "14"
```
ddbeck, line 40: I tried the example on https://developer.mozilla.org/en-US/docs/Web/CSS/::-webkit-scrollbar in BrowserStack. It looks like this actually stopped working in version 13. Unless I'm missing something, it looks like this version number needs to be updated throughout the PR (with the exception noted at the end).

ddbeck (changes requested):
Thanks for this PR, @GrimLink! I see it's your first to the project—welcome!

I've got a couple of points for you to consider in line comments. Thanks again!

css/selectors/-webkit-scrollbar.json
GrimLink: Thx for testing it @ddbeck i have not had the chance yet.

css/selectors/-webkit-scrollbar-button.json
GrimLink: Yes on BrowserStack it also happens on version 13.4 but I suspect this also affects any iOS 13 version.
Funny nobody mentioned this anywhere yet for iOS 13.
I will update the PR to correct this 😉

GrimLink:
@ddbeck kan you check again, if the PR is good now?

css/selectors/-webkit-scrollbar.json
```diff
@@ -36,7 +36,8 @@
               "version_added": "4"
             },
             "safari_ios": {
-              "version_added": "3"
+              "version_added": "3",
+              "notes": "As of Safari for iOS 13 only `display: none` will work, any other styles have no effect."
```
ddbeck, line 40: Some style suggestions, for consistency with the way we write notes elsewhere (mostly, to be more concise, use HTML, and avoid a comma splice).

Also, a note in case anyone looks this up in the future: we didn't go with multiple support statements with `partial_implementation` because this is non-standard feature. It's unknowable what would constitute partial implementation, so we can only note the changes in behavior.

```suggestion
              "notes": "From Safari 13, only <code>display: none</code> works with this pseudo-element. Other styles have no effect."
```

ddbeck (changes requested):
One last suggestion for you, @GrimLink! Thank you!

ddbeck (approved):
Looks good. Thank you! 🎉

PR #13454, opened 2021-11-11
Add missing api.ImageData.colorSpace feature

queengooborg:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing `colorSpace` member of the ImageData API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v4.0.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/ImageData/colorSpace

### DIFF
--- a/api/ImageData.json
+++ b/api/ImageData.json
@@ -98,6 +98,54 @@
           }
         }
       },
+      "colorSpace": {
+        "__compat": {
+          "spec_url": "https://html.spec.whatwg.org/multipage/canvas.html#dom-imagedata-colorspace",
+          "support": {
+            "chrome": {
+              "version_added": "92"
+            },
+            "chrome_android": {
+              "version_added": "92"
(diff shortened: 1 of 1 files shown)


api/ImageData.json
```diff
@@ -98,6 +98,54 @@
...
+            },
+            "samsunginternet_android": {
+              "version_added": false
+            },
+            "webview_android": {
+              "version_added": "92"
+            }
+          },
+          "status": {
+            "experimental": false,
```
foolip, line 143: Since it's still Chromium-only:

```suggestion
            "experimental": true,
```

foolip (changes requested):

foolip (approved):

PR #13486, opened 2021-11-12
Remove calc()'s "<color> value support" entry

foolip:
This entry represents support for calc() inside color functions like
rgb(), and can be tested with `rgb(0, calc(64 + 64), 0)`:
http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=9798

The data for Firefox is correct, this didn't work before Firefox 59.

However, the data for Chrome and Edge is wrong, this was supported since
Chrome 16 (with `-webkit-calc()`) and at least since Edge 15.

Since this concerns only Firefox, make it a note instead.

### DIFF
--- a/css/types/calc.json
+++ b/css/types/calc.json
@@ -26,6 +26,7 @@
               {
                 "version_added": "16",
                 "notes": [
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
Thanks!

PR #13496, opened 2021-11-12
Add Edge versions for api.HTMLTrackElement.cuechange_event

queengooborg:
This PR adds real values for Microsoft Edge for the `cuechange_event` member of the `HTMLTrackElement` API based upon manual testing.  See https://github.com/mdn/browser-compat-data/pull/13042 for the test code.

### DIFF
--- a/api/HTMLTrackElement.json
+++ b/api/HTMLTrackElement.json
@@ -63,7 +63,7 @@
               "version_added": "25"
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "18"
             },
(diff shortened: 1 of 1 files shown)


foolip:
Can you test this manually? There are two "cuechange" events, one fired on `TextTrack` and the other on `HTMLTrackElement`, and they were *probably* introduced at the same time. But maybe not, and also I wouldn't bet that the `oncuechange` properties were added from the beginning.

foolip (approved):

PR #13514, opened 2021-11-14
Add Edge versions for FileSystem API

queengooborg:
This PR adds real values for Microsoft Edge for the `FileSystem` API, based upon manual testing.

Test Code Used: `'FileSystem' in self; 'WebKitFileSystem' in self;`

### DIFF
--- a/api/FileSystem.json
+++ b/api/FileSystem.json
@@ -13,11 +13,18 @@
             "version_added": "18",
             "alternative_name": "DOMFileSystem"
           },
-          "edge": {
-            "version_added": "14",
-            "prefix": "WebKit",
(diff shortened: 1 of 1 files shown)


foolip (changes requested):
Can you take a closer look at the prefixes over time for this? It makes no sense that Edge 14 would have added `WebKitFileSystem` if that wasn't in Chrome or Safari and Edge needed it for site compat.

Also Opera's lowercase "webkit" prefix is suspect, was it `webkitFileSystem` or `WebKitFileSystem`?

queengooborg:
Sorry for the notification spam!  Phone screen was covered in water...  😅

PR #13515, opened 2021-11-14
Add Edge versions for FileSystemDirectoryReader API

queengooborg:
This PR adds real values for Microsoft Edge for the `FileSystemDirectoryReader` API, based upon manual testing.

Test Code Used: `'FileSystemDirectoryReader' in self; 'DirectoryReader' in self; 'WebKitDirectoryReReader' in self;`

### DIFF
--- a/api/FileSystemDirectoryReader.json
+++ b/api/FileSystemDirectoryReader.json
@@ -13,10 +13,17 @@
             "version_added": "18",
             "alternative_name": "DirectoryReader"
           },
-          "edge": {
-            "version_added": "≤18",
-            "alternative_name": "WebKitDirectoryReader"
-          },
+          "edge": [
+            {
+              "version_added": "79",
+              "alternative_name": "DirectoryReader"
+            },
+            {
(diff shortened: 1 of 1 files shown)


api/FileSystemDirectoryReader.json
```diff
@@ -13,10 +13,17 @@
...
-          },
+          "edge": [
+            {
+              "version_added": "79",
+              "alternative_name": "DirectoryReader"
+            },
+            {
+              "version_added": "14",
+              "version_removed": "79",
+              "alternative_name": "WebKitDirectoryReader"
```
foolip, line 24: Which other browsers have had `WebKitDirectoryReader`? It seems unlikely that only Edge has ever had it.

foolip (changes requested):

queengooborg:
Closing this PR because it appears there's a lot regarding these APIs to dig into and fix.  I opened an issue on the collector repo to remind myself to write up custom tests for these.

PR #13537, opened 2021-11-15
`namespace CSS` is defined in CSSOM, set `spec_url` as such

saschanaz:
#### Summary

Somehow it currently points to CSS Conditional Rules.

### DIFF
--- a/api/CSS.json
+++ b/api/CSS.json
@@ -3,7 +3,7 @@
     "CSS": {
       "__compat": {
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/CSS",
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
Looks good to me given the CSSOM spec defines `namespace CSS` and CSS Conditional only defines `partial namespace CSS`. What do you think @sideshowbarker?

sideshowbarker (approved):

sideshowbarker:
> Looks good to me given the CSSOM spec defines `namespace CSS` and CSS Conditional only defines `partial namespace CSS`. What do you think @sideshowbarker?

Looks good to me too 👍

PR #13543, opened 2021-11-16
Update Edge versions for PushSubscription API

queengooborg:
This PR updates and corrects the real values for Microsoft Edge for the `PushSubscription` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v4.0.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/PushSubscription

Additional notes: support for service workers was implemented in Edge 17, so it wouldn't make sense for this interface to have any support before Edge 17.

### DIFF
--- a/api/PushSubscription.json
+++ b/api/PushSubscription.json
@@ -12,7 +12,7 @@
             "version_added": "42"
           },
           "edge": {
-            "version_added": "≤18"
+            "version_added": "17"
           },
           "firefox": {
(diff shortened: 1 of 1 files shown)


foolip (approved):
Matches https://developer.mozilla.org/en-US/docs/Web/API/PushManager#browser_compatibility 👍

PR #13559, opened 2021-11-17
Add Safari versions for MouseEvent API

queengooborg:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `MouseEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v4.0.0).

### DIFF
--- a/api/MouseEvent.json
+++ b/api/MouseEvent.json
@@ -132,10 +132,10 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "≤4"
+              "version_added": "3.1"
             },
             "safari_ios": {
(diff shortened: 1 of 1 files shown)


foolip (changes requested):
This doesn't seem plausible, I bet `MouseEvent` was supported since Safari 1. How does the test fail in Safari 1? Maybe it's not possible to create events, but they can be fired by the browser?

queengooborg:
Looks like all these features were implemented in the initial revision for Safari, fixed!

foolip (approved):

PR #13605, opened 2021-11-18
Add Safari versions for SVGRenderingIntent API

queengooborg:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `SVGRenderingIntent` API, based upon manual testing.

Test Code Used: `alert(SVGRenderingIntent);`

### DIFF
--- a/api/SVGRenderingIntent.json
+++ b/api/SVGRenderingIntent.json
@@ -34,10 +34,10 @@
             "version_removed": "32"
           },
           "safari": {
-            "version_added": "≤4"
+            "version_added": "3.1"
           },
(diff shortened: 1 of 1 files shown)


foolip (approved):
It's possible that `SVGRenderingIntent` was supported earlier but the interface object wasn't exposed, but this is a deprecated feature and added so far in the past that I won't dig deeper. Good enough!

queengooborg:
All other SVG support was added in Safari 3, so I didn't bother to dig any deeper -- it wouldn't make sense for this interface to be supported earlier than the rest of SVG!

PR #13629, opened 2021-11-19
compat-data-scheme: spec data clarification

hamishwillee:
This adds the information about what specs can be added, as a follow on to the discussion here: https://github.com/mdn/browser-compat-data/pull/13383#issuecomment-971825712 

This might not be the right place - e.g. it could be 
https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md or somewhere else.

However it would be good if it was somewhere.


Elchi3 (approved):
This works for now. @sideshowbarker and I still owe BCD proper spec_url guidelines and we can then decide where these should go.

PR #13655, opened 2021-11-20
Complete nodejs data for URL and URLSearchParams

gilmoreorless:
#### Summary

Part of #13170. This fills in `true` and missing `nodejs` values for `URL` and `URLSearchParams`.

Most of the data came from the "history" section of [the documentation](https://nodejs.org/dist/latest-v16.x/docs/api/url.html#class-url).

Local testing also revealed the oddity of `URL#searchParams` returning a dummy object in `7.0.0` before `URLSearchParams` was properly implemented in `7.5.0`.

### DIFF
--- a/api/URL.json
+++ b/api/URL.json
@@ -143,9 +143,15 @@
             "ie": {
               "version_added": false
             },
-            "nodejs": {
-              "version_added": true
-            },
+            "nodejs": [
+              {
+                "version_added": "10.0.0"
+              },
+              {
+                "version_added": "7.0.0",
+                "notes": "Has to be imported from the URL module."
+              }
+            ],
             "opera": {
               "version_added": "15"
             },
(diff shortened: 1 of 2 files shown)


api/URL.json
```diff
@@ -143,9 +143,15 @@
...
-            },
+            "nodejs": [
+              {
+                "version_added": "10.0.0"
+              },
+              {
+                "version_added": "7.0.0",
+                "notes": "Has to be imported from the URL module."
+              }
+            ],
```
gilmoreorless, line 154: This information for the `URL` constructor is copied from the top-level object.

api/URLSearchParams.json
```diff
@@ -89,16 +89,9 @@
...
-            "nodejs": [
-              {
-                "version_added": "7.10.0"
-              },
-              {
-                "version_added": "7.5.0",
-                "partial_implementation": true,
-                "notes": "Only string as <code>init</code> object supported."
-              }
-            ],
```
gilmoreorless, line 101: I removed the note because the information is covered by the later-added `USVString`, `record`, and `sequence` sub-properties.

Elchi3 (approved):
Thanks! 👍

PR #13669, opened 2021-11-21
Firefox for Android permissions.request can't be triggered by browserAction.onClicked

diafygi:
To include additional limitation on triggering `permissions.request()` in Firefox for Android

#### Summary
Add clarification that trying to call `permissions.request` from a [`browserAction.onClicked`](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/browserAction/onClicked) event in Firefox for Android will throw an error, despite it being a valid user interaction.

See https://github.com/mozilla-mobile/fenix/issues/22348 

#### Related issues
See https://github.com/mozilla-mobile/fenix/issues/22348

### DIFF
--- a/webextensions/api/permissions.json
+++ b/webextensions/api/permissions.json
@@ -195,7 +195,7 @@
               "firefox_android": {
                 "version_added": true,
                 "partial_implementation": true,
-                "notes": "From version 79, the user interface to approve the permission request is missing. See <a href='https://bugzil.la/1601420'>bug 1601420</a>."
+                "notes": "From version 79, the user interface to approve the permission request is missing (see <a href='https://bugzil.la/1601420'>bug 1601420</a>), and <code>browserAction.onClicked</code> is not recognized as a user action (see <a href='https://github.com/mozilla-mobile/fenix/issues/22348'>fenix bug 22348</a>)"
               },
               "opera": {
                 "version_added": true
--- a/webextensions/api/permissions.json
+++ b/webextensions/api/permissions.json
@@ -195,7 +195,10 @@
               "firefox_android": {
                 "version_added": true,
                 "partial_implementation": true,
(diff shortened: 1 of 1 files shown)


webextensions/api/permissions.json
```diff
@@ -195,7 +195,7 @@
               "firefox_android": {
                 "version_added": true,
                 "partial_implementation": true,
-                "notes": "From version 79, the user interface to approve the permission request is missing. See <a href='https://bugzil.la/1601420'>bug 1601420</a>."
+                "notes": "From version 79, the user interface to approve the permission request is missing (see <a href='https://bugzil.la/1601420'>bug 1601420</a>), and <code>browserAction.onClicked</code> is not recognized as a user action (see <a href='https://github.com/mozilla-mobile/fenix/issues/22348'>fenix bug 22348</a>)"
```
ddbeck, line 198: Since these are two separate issues, let's use an array of notes instead of a single note. You'll have an array of two elements (plus some style fixes):

1. `"From version 79, the user interface to approve the permission request is missing (see <a href='https://bugzil.la/1601420'>bug 1601420</a>)."`
2. `"<code>browserAction.onClicked</code> is not recognized as a user action (see <a href='https://github.com/mozilla-mobile/fenix/issues/22348'>mozilla-mobile/fenix#22348</a>)."`

Also, if this is a regression and you know what version it was introduced, it would be good to note that (e.g., `"From version X, <code>browserAction.onClicked</code> is not…"`). Otherwise, we can set it when it's resolved (e.g., `"Before version Y, …"`).

ddbeck (changes requested):
Thanks for this PR, @diafygi. Two suggestions for you in a line comment. Thank you again!

webextensions/api/permissions.json
diafygi: @ddbeck Thanks! I've updated the pull request with your suggested changes. I don't know if the onClicked bug is a regression, so I don't know when it was introduced.

ddbeck (approved):
Thank you! 🎉

PR #13682, opened 2021-11-22
Speech API is not (mosty) experimental

hamishwillee:
Following on from https://github.com/mdn/browser-compat-data/pull/13627 this looks at the rest of the API. 
Most of the interfaces have at least two implementations, and so are, by definition, not experimental. There are a few questions for some of them though:

- [x] [SpeechSynthesis](https://developer.mozilla.org/en-US/docs/Web/API/SpeechSynthesis) - not experimental - more than two implementations
- [x] [SpeechSynthesisErrorEvent](https://developer.mozilla.org/en-US/docs/Web/API/SpeechSynthesisErrorEvent) - not experimental - more than two implementations
- [x] [SpeechSynthesisVoice](https://developer.mozilla.org/en-US/docs/Web/API/SpeechSynthesisVoice) - not experimental - more than two implementations

Possible additional changes, pending questions:

- [x]  [SpeechRecognition](https://developer.mozilla.org/en-US/docs/Web/API/SpeechRecognition), [SpeechRecognitionEvent](https://developer.mozilla.org/en-US/docs/Web/API/SpeechRecognitionEvent), [SpeechGrammar](https://developer.mozilla.org/en-US/docs/Web/API/SpeechGrammar), [SpeechGrammarList](https://developer.mozilla.org/en-US/docs/Web/API/SpeechGrammarList) "_implemented with the vendor prefix: webkit_" in chrome and variants. These are supported in more than one browser, but with this caveat, is this "experimental"?
- [x] [SpeechRecognition](https://developer.mozilla.org/en-US/docs/Web/API/SpeechRecognition), [SpeechRecognitionAlternative](https://developer.mozilla.org/en-US/docs/Web/API/SpeechRecognitionAlternative), [SpeechRecognitionResult](https://developer.mozilla.org/en-US/docs/Web/API/SpeechRecognitionResult), [SpeechRecognitionResultList](https://developer.mozilla.org/en-US/docs/Web/API/SpeechRecognitionResultList), and a bunch of others are marked as false for Opera and OperaAndroid . Chrome went in blink 33, which corresponds to Opera 20. Is this likely an error? I plan on ignoring for now.

EDITED:
These ones reverted due to Daniel suspicions
- [x] [SpeechRecognitionAlternative](https://developer.mozilla.org/en-US/docs/Web/API/SpeechRecognitionAlternative) - not experimental - more than two implementations
- [x] [SpeechRecognitionResult](https://developer.mozilla.org/en-US/docs/Web/API/SpeechRecognitionResult) - not experimental - more than two implementations
- [x] [SpeechRecognitionResultList](https://developer.mozilla.org/en-US/docs/Web/API/SpeechRecognitionResultList) - not experimental - more than two implementations

### DIFF
--- a/api/SpeechRecognition.json
+++ b/api/SpeechRecognition.json
@@ -55,7 +55,7 @@
           }
         },
         "status": {
-          "experimental": true,
+          "experimental": false,
           "standard_track": true,
           "deprecated": false
         }
@@ -111,7 +111,7 @@
             }
           },
           "status": {
-            "experimental": true,
+            "experimental": false,
             "standard_track": true,
             "deprecated": false
           }
@@ -160,7 +160,7 @@
             }
           },
           "status": {
-            "experimental": true,
+            "experimental": false,
             "standard_track": true,
             "deprecated": false
           }
@@ -210,7 +210,7 @@
             }
           },
           "status": {
-            "experimental": true,
+            "experimental": false,
             "standard_track": true,
             "deprecated": false
           }
@@ -260,7 +260,7 @@
             }
           },
           "status": {
-            "experimental": true,
+            "experimental": false,
             "standard_track": true,
             "deprecated": false
           }
@@ -309,7 +309,7 @@
             }
           },
           "status": {
-            "experimental": true,
+            "experimental": false,
             "standard_track": true,
             "deprecated": false
           }
@@ -359,7 +359,7 @@
             }
           },
           "status": {
-            "experimental": true,
+            "experimental": false,
             "standard_track": true,
             "deprecated": false
           }
@@ -409,7 +409,7 @@
             }
           },
           "status": {
-            "experimental": true,
+            "experimental": false,
             "standard_track": true,
             "deprecated": false
           }
@@ -458,7 +458,7 @@
             }
           },
           "status": {
-            "experimental": true,
+            "experimental": false,
             "standard_track": true,
             "deprecated": false
           }
@@ -507,7 +507,7 @@
             }
           },
           "status": {
-            "experimental": true,
+            "experimental": false,
             "standard_track": true,
             "deprecated": false
           }
@@ -556,7 +556,7 @@
             }
           },
           "status": {
-            "experimental": true,
+            "experimental": false,
             "standard_track": true,
             "deprecated": false
           }
@@ -605,7 +605,7 @@
             }
           },
           "status": {
-            "experimental": true,
+            "experimental": false,
             "standard_track": true,
             "deprecated": false
           }
@@ -655,7 +655,7 @@
             }
           },
           "status": {
-            "experimental": true,
+            "experimental": false,
             "standard_track": true,
             "deprecated": false
           }
@@ -704,7 +704,7 @@
             }
           },
           "status": {
-            "experimental": true,
+            "experimental": false,
             "standard_track": true,
             "deprecated": false
           }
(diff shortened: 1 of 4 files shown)


ddbeck:
Answers, hopefully:

> [SpeechRecognition](https://developer.mozilla.org/en-US/docs/Web/API/SpeechRecognition), [SpeechRecognitionEvent](https://developer.mozilla.org/en-US/docs/Web/API/SpeechRecognitionEvent), [SpeechGrammar](https://developer.mozilla.org/en-US/docs/Web/API/SpeechGrammar), [SpeechGrammarList](https://developer.mozilla.org/en-US/docs/Web/API/SpeechGrammarList) "_implemented with the vendor prefix: webkit_" in chrome and variants. These are supported in more than one browser, but with this caveat, is this "experimental"?

That would no longer be experimental. From the guideline:

> If two or more engines support a feature behind a common prefix (such as `-webkit-` only), then set `experimental` to `false`.

So you can fix that too, if you like.

> [SpeechRecognition](https://developer.mozilla.org/en-US/docs/Web/API/SpeechRecognition), [SpeechRecognitionAlternative](https://developer.mozilla.org/en-US/docs/Web/API/SpeechRecognitionAlternative), [SpeechRecognitionResult](https://developer.mozilla.org/en-US/docs/Web/API/SpeechRecognitionResult), [SpeechRecognitionResultList](https://developer.mozilla.org/en-US/docs/Web/API/SpeechRecognitionResultList), and a bunch of others are marked as false for Opera and OperaAndroid . Chrome went in blink 33, which corresponds to Opera 20. Is this likely an error? I plan on ignoring for now.

That may be correct. I think you'd have to test it to find out for sure, but I think the API can rely on an external service (and does, for Chrome), so it may not be implemented in the derivatives.

hamishwillee:
Thanks @ddbeck ! This should be good to merge. Since you saw it last I updated [SpeechRecognition](https://developer.mozilla.org/en-US/docs/Web/API/SpeechRecognition), 

- I did not update
  -  [SpeechRecognitionEvent](https://developer.mozilla.org/en-US/docs/Web/API/SpeechRecognitionEvent) on further investigation it has it has two implementations but one uses webkit prefix, but the other does not.
  - [SpeechGrammar](https://developer.mozilla.org/en-US/docs/Web/API/SpeechGrammar) - it has two implementations, but the FF one is behind a flag
- [SpeechGrammarList](https://developer.mozilla.org/en-US/docs/Web/API/SpeechGrammarList)  - I was just wrong - no additional implementation.
- I'm not going to retest the android cases - I have no reason "not" to believe this.

ddbeck:
So, I started re-reviewing this and noticed something… odd. I wondered why only some of these were prefixed and some werent (e.g., `SpeechRecognition` versus `SpeechRecognitionAlternative`). I opened up Chrome and found that several of these don't appear to be `Window`-exposed interfaces, though I don't know if that means they're wholly unimplemented. The troublemakers seem to be:

- `SpeechRecognitionAlternative`
- `SpeechRecognitionResult`
- `SpeechRecognitionResultList`

They do exist (unprefixed) in Safari, however.

What do you think of 1) leaving them out of this PR and 2) opening an issue saying that these are super suspect and need more substantial testing.

hamishwillee:
@ddbeck Sounds like a good idea. I reverted the commits on the suspicious APIs in this PR so it should be good to re-review. New issue for those created in#13934

ddbeck (approved):
Looks good. Thank you! 🎉

PR #13466, opened 2021-11-11
Add missing api.PermissionStatus.name feature

queengooborg:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing `name` member of the PermissionStatus API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v4.0.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/PermissionStatus/name

_Check out the [collector's guide on how to review this PR](https://github.com/foolip/mdn-bcd-collector#reviewing-bcd-changes)._

PR #13683, opened 2021-11-23
PermissionStatus items other than name are not experimental

hamishwillee:
Other than it's new `name` property, `PermissionStatus` is used by more than 2 major browsers. This marks those as not experimental. As discussed in https://github.com/mdn/browser-compat-data/pull/13466#issuecomment-976016162

### DIFF
--- a/api/PermissionStatus.json
+++ b/api/PermissionStatus.json
@@ -44,7 +44,7 @@
           }
         },
         "status": {
-          "experimental": true,
+          "experimental": false,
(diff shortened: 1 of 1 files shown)


hamishwillee:
@queengooborg @ddbeck It seems to me that the rules for what is experimental are pretty clear. Perhaps "someone" should add a rule to check these?

Elchi3 (approved):

PR #13694, opened 2021-11-23
Remove the requestAnimationFrame() "Return value" subfeature

foolip:
This is already captured by the cancelAnimationFrame() entry.

The return value of requestAnimationFrame() is an integer which can be
used to cancel the callback with cancelAnimationFrame(). Judging by the
data, this entry must have been added to capture that
mozRequestAnimationFrame() returned undefined in Firefox 4-10, but this
by itself is not useful information.

Also, most readers could probably not decipher what this entry means.

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -6660,54 +6660,6 @@
             "standard_track": true,
             "deprecated": false
           }
-        },
-        "return_value": {
-          "__compat": {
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
Makes sense to me 👍

PR #13708, opened 2021-11-24
Add Edge versions for api.Window.languagechange_event

queengooborg:
This PR adds real values for Microsoft Edge for the `languagechange_event` member of the `Window` API.  The data was copied from its event handler counterpart.

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -2870,7 +2870,7 @@
               "version_added": "37"
             },
             "edge": {
-              "version_added": "≤79"
+              "version_added": "79"
             },
             "firefox": {
               "version_added": "32"


foolip:
I suspect this event was supported earlier. Can you do some manual testing?

queengooborg:
I think this is as best as we'll be able to get on our own.  I can't find any in-browser way to change preferred languages in Legacy Edge, and all tutorials I've looked up talk about Edgium.  The only one I found for Legacy Edge says you must change Windows' language.

Attempting to change the system language in Edge 14 yielded no results, and I can't test this in BrowserStack since the Settings app is restricted.

foolip:
> Attempting to change the system language in Edge 14 yielded no results

Alright, that's good enough for a test! Thanks!

foolip (approved):

PR #13714, opened 2021-11-24
Create bare JSON release assets (and upload to S3)

ddbeck:
#### Summary

This adds a bare JSON file, without the npm package wrapper, as an asset to each release on GitHub. It also uploads that asset to S3, for consumption by MDN itself.

GitHub Actions are really hard to test in advance. I tried it with [`act`](https://github.com/nektos/act) but it's tough to simulate all the pieces here. I _think_ it'll work, but I'm not sure. That said, everything that's new here will fail _after_ the known-good publishing process is done, so it shouldn't make things any worse, if it does fail.


Elchi3 (approved):
It reads good to me and it seems like we should just test this and correct things if needed.

PR #13695, opened 2021-11-23
Add feature inventory artifacts on push

ddbeck:
#### Summary

Create an artifact for every push that enumerates the features recorded for BCD.

#### Test results and supporting details

This will save considerable time for writing release notes or doing other automation where you need to know what features were added or removed in a given commit.

#### Related issues

This is a prerequisite to improving the release notes for https://github.com/mdn/browser-compat-data/issues/13229.

PR #13720, opened 2021-11-24
Diff features by fetching from GitHub Actions

ddbeck:
#### Summary

This substantially speeds up feature diffing, where the enumeration has already been done in GitHub Actions. Where possible, it tries to fetch from GitHub Actions, or falls back to using a worktree to enumerate directly.

You can see this in action with these examples:

- `node ./scripts/diff-features.js --format=json 80d976fb46e5ca453856adcab9bcf0b36a4bf140`, for a fallback case. This is typical for commits older than about 24 hours. On my machine, runs in about 10 seconds (with a full npm cache).
- `node ./scripts/diff-features.js --format=json 1409a8c6d17e7d420404bb1a0561cca431cba1b5`, for a GitHub-facilitated case. Runs in about 5 seconds.

#### Related issues

See https://github.com/mdn/browser-compat-data/pull/13695 for background.

### DIFF
--- a/scripts/diff-features.js
+++ b/scripts/diff-features.js
@@ -20,8 +20,8 @@ function main({ ref1, ref2, format }) {
     refB = `${ref1}`;
   }
 
-  const aSide = new Set(enumerateFeatures(refA));
-  const bSide = new Set(enumerateFeatures(refB));
+  let aSide = enumerate(refA);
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
Well done! I like the fallback option!

PR #13736, opened 2021-11-25
add Firefox 81 flex-direction fully supported

### LINKED ISSUE #13735: css.properties.flex-direction partial implementation for Firefox
Seems misleading, since even the referenced bug mentions it had been broken for `*-reverse` variants to version 81. Current ESR is 91, so it seems safe to mark it as fully supported.

https://github.com/mdn/browser-compat-data/blob/f14683313a65ca974f053e55636d09a16fe2f2e4/css/properties/flex-direction.json#L39
https://github.com/mdn/browser-compat-data/blob/f14683313a65ca974f053e55636d09a16fe2f2e4/css/properties/flex-direction.json#L53

myfonj:
Fixes #13735

Data from notes here that links https://bugzilla.mozilla.org/show_bug.cgi?id=1042151

### DIFF
--- a/css/properties/flex-direction.json
+++ b/css/properties/flex-direction.json
@@ -35,6 +35,9 @@
             ],
             "firefox": [
               {
+                "version_added": "81"
+              },
+              {
                 "version_added": "20",
                 "partial_implementation": true,
                 "notes": [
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
Thanks! 👍

PR #13741, opened 2021-11-26
Add Edge versions for api.IDBVersionChangeEvent.version

queengooborg:
This PR adds real values for Microsoft Edge for the `version` member of the `IDBVersionChangeEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v4.0.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/IDBVersionChangeEvent/version (will be added in collector v4.1)

### DIFF
--- a/api/IDBVersionChangeEvent.json
+++ b/api/IDBVersionChangeEvent.json
@@ -253,7 +253,7 @@
               "version_added": "18"
             },
             "edge": {
-              "version_added": "≤18"
+              "version_added": "79"
             },
             "firefox": {
               "version_added": "10"


api/IDBVersionChangeEvent.json
```diff
@@ -253,7 +253,7 @@
               "version_added": "18"
             },
             "edge": {
-              "version_added": "≤18"
+              "version_added": "79"
```
foolip, line 256: https://staging-dot-mdn-bcd-collector.appspot.com/tests/api/IDBVersionChangeEvent/version returns false on Edge 96. Has it been removed again?

foolip (changes requested):

api/IDBVersionChangeEvent.json
queengooborg: Looks like it's `false` for Chrome as well, hmm...

foolip: Looks like it's not in Chromium or Gecko source:
https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/indexeddb/idb_version_change_event.idl
https://github.com/mozilla/gecko-dev/blob/master/dom/webidl/IDBVersionChangeEvent.webidl

Can you check if it was ever there? If it was removed long enough ago, we can just remove this entry.

queengooborg: The test is returning false all the way back in Chrome 37, and even in Firefox 40.  That's enough evidence that this feature can be considered irrelevant and can be removed.  I'll open a PR to do so!

PR #13743, opened 2021-11-26
Add Edge versions for api.RTCIceCandidate.address

queengooborg:
This PR adds real values for Microsoft Edge for the `address` member of the `RTCIceCandidate` API, based upon manual testing.

Test Code Used: `var rtcic = new RTCIceCandidate(); rtcic.ip;`

### DIFF
--- a/api/RTCIceCandidate.json
+++ b/api/RTCIceCandidate.json
@@ -137,16 +137,9 @@
             "chrome_android": {
               "version_added": "74"
             },
-            "edge": [
-              {
-                "version_added": "79"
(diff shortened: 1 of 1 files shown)


foolip:
So the `ip` property was never there? Can you check where this data came from?

queengooborg:
It looks like the data for the `ip` feature/alt. name originally came from the wiki migration.  It was originally added with everything set to `true`, which further makes me think that it's incorrect.

foolip (approved):

PR #13748, opened 2021-11-26
Mention the existing Chromium attempt for certificate API in WebExtension

FranklinYu:
#### Summary
Add a note to mention the existing attempt, of implementing the [certificate API for WebExtension](https://developer.mozilla.org/en-US/docs/Mozilla/Add-ons/WebExtensions/API/webRequest#accessing_security_information) in Chromium.

Not applicable, since this is only adding a note.

#### Related issues
N/A

### DIFF
--- a/webextensions/api/webRequest.json
+++ b/webextensions/api/webRequest.json
@@ -64,7 +64,7 @@
             "support": {
               "chrome": {
                 "version_added": false,
-                "notes": "There was <a href='https://crrev.com/c/644858'>an attempt</a> to implement this."
+                "notes": "See <a href='https://crbug.com/628819'>bug 628819</a>."
               },
               "edge": {
                 "version_added": false
--- a/webextensions/api/webRequest.json
+++ b/webextensions/api/webRequest.json
@@ -1099,7 +1099,7 @@
             "support": {
               "chrome": {
                 "version_added": false,
-                "notes": "There was <a href='https://crrev.com/c/644858'>an attempt</a> to implement this."
+                "notes": "See <a href='https://crbug.com/628819'>bug 628819</a>."
               },
               "edge": {
                 "version_added": false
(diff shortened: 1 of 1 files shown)


webextensions/api/webRequest.json
```diff
@@ -63,7 +63,8 @@
             "mdn_url": "https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/CertificateInfo",
             "support": {
               "chrome": {
-                "version_added": false
+                "version_added": false,
+                "notes": "There was <a href='https://crrev.com/c/644858'>an attempt</a> to implement this."
```
ddbeck, line 67: I don't think this text works in the context of BCD. It's certainly true, but I don't think it's especially relevant to web developers who want to use this feature. Instead, we could link to [the corresponding bug](https://bugs.chromium.org/p/chromium/issues/detail?id=628819), so that if there's ever progress implementing this feature they'd be made aware of it more directly. We'd typically present it like this:

```suggestion
                "notes": "See <a href='https://crbug.com/628819'>bug 628819</a>."
```

webextensions/api/webRequest.json
```diff
@@ -1097,7 +1098,8 @@
             "mdn_url": "https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/API/webRequest/getSecurityInfo",
             "support": {
               "chrome": {
-                "version_added": false
+                "version_added": false,
+                "notes": "There was <a href='https://crrev.com/c/644858'>an attempt</a> to implement this."
```
ddbeck, line 1102:
```suggestion
                "notes": "See <a href='https://crbug.com/628819'>bug 628819</a>."
```

ddbeck (changes requested):
Thanks for this PR, @FranklinYu. A suggestion for you in line comments.

webextensions/api/webRequest.json
FranklinYu: I noticed that the bug also points to the Code Review link I used, so this would not lose any information. Thank you for finding the bug!

ddbeck (approved):
Looks good. Thank you! 🎉

PR #13764, opened 2021-11-27
Update Safari versions for api.Document.readystatechange_event

queengooborg:
This PR corrects values for Safari (Desktop and iOS/iPadOS) for the `readystatechange_event` member of the `Document` API, based upon manual testing.

Test Code Used:
```html
<div id="test">
	<button id="reload" type="button">Reload</button>
</div>

<script>
	var reload = document.getElementById('reload');

	reload.onclick = function() {
		window.location.reload(true);
	}

	document.addEventListener('readystatechange', function(event) {
		alert('readystatechange!');
	});
</script>
```

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -9669,10 +9669,10 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "≤4"
+              "version_added": "5.1"
             },
             "safari_ios": {
-              "version_added": "≤3"
+              "version_added": "5"
             },
             "samsunginternet_android": {
               "version_added": "1.0"


foolip (approved):
I thought this seems a bit late given that `document.readyState` is older, but confirmed by commit:
https://trac.webkit.org/changeset/69710/webkit
https://trac.webkit.org/browser/webkit/trunk/WebCore/Configurations/Version.xcconfig?rev=69710

This shows that Chrome is wrong, though.

foolip:
I've sent https://github.com/mdn/browser-compat-data/pull/13793 to fix the Chrome data.

PR #13778, opened 2021-11-28
Add Safari versions for api.HTMLMediaElement.waiting_event

queengooborg:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `waiting_event` member of the `HTMLMediaElement` API, based upon manual testing.

Test Code Used:
```html
<div id="test">
	<video id="video" controls width="250"></video>
</div>

<script>
	var video = document.getElementById('video');
	var videoSrc = '/queengooborg/static/rabbit320.mp4';
	// https://mdn.github.io/learning-area/html/multimedia-and-embedding/video-and-audio-content/rabbit320.mp4

	video.addEventListener('waiting', function() {
	  console.log('Waiting!');
	  // Triggered by loading the page, then disabling the connection to the server before playing the video
	});

	var source = document.createElement('source');
	source.setAttribute('src', videoSrc);
	source.setAttribute('type', 'video/mp4');

	video.appendChild(source);
</script>

```

### DIFF
--- a/api/HTMLMediaElement.json
+++ b/api/HTMLMediaElement.json
@@ -4112,10 +4112,10 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "≤4"
+              "version_added": "4"
             },
             "safari_ios": {
-              "version_added": "≤3"
+              "version_added": "3"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
--- a/api/HTMLMediaElement.json
+++ b/api/HTMLMediaElement.json
@@ -4112,7 +4112,7 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "4"
+              "version_added": "3.1"
             },
             "safari_ios": {
               "version_added": "3"


foolip:
The "waiting" event was in https://trac.webkit.org/changeset/27277/webkit, the initial commit support for `<video>`.

However, I don't think 1.3 is the right initial version. Can you check which version of Safari first had `<video>`, and then set the version to that?

foolip:
I think we can trust https://developer.apple.com/library/archive/documentation/AudioVideo/Conceptual/Using_HTML5_Audio_Video/Introduction/Introduction.html on this:
> Safari supports the `<video>` and `<audio>` media elements on iOS 3.0 and later and in Safari 3.1 and later on the desktop (Mac OS X and Windows). Support for these media elements allows Safari and other HTML5-compliant browsers to play the indicated source media without using a plug-in.

foolip:
Doing that update as a separate PR would be good. Updating audio.json and video.json too would be great.

foolip (approved):

PR #13805, opened 2021-11-29
Update supported platforms for BarcodeDetector in Chrome

### LINKED ISSUE #13752: BarcodeDetector will be supported on all platforms in Chromium
This is happening in https://chromium-review.googlesource.com/c/chromium/src/+/3290206, and I'm filing this issue as a reminder since it's not very likely to be fixed otherwise. Here's an example statement that needs to be updated:

https://github.com/mdn/browser-compat-data/blob/6248a009a5a78ed3b7d370c35fe8553a4238bf87/api/BarcodeDetector.json#L8-L12

reillyeon:
#### Summary
The BarcodeDetector interface has been supported on Chrome OS since Chrome 88.

https://groups.google.com/a/chromium.org/g/blink-dev/c/j-PLtssE5fo/m/Tj9HoFOIDAAJ

#### Related issues
Fixes #13752.

### DIFF
--- a/api/BarcodeDetector.json
+++ b/api/BarcodeDetector.json
@@ -5,10 +5,19 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/BarcodeDetector",
         "spec_url": "https://wicg.github.io/shape-detection-api/#barcode-detection-api",
         "support": {
-          "chrome": {
-            "version_added": "83",
-            "notes": "Before Chrome 98, this feature was only supported on macOS. In Chrome 98, support was added for Linux and Windows."
-          },
+          "chrome": [
+            {
+              "version_added": "88",
+              "partial_implementation": true,
+              "notes": "Supported on Chrome OS and macOS only."
+            },
+            {
+              "version_added": "83",
+              "version_removed": "88",
+              "partial_implementation": true,
+              "notes": "Supported on macOS only."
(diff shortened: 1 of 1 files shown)


ddbeck (changes requested):
Thanks for this PR, @reillyeon. I see it's your first to the project. Welcome!

A couple thoughts on this one:

1. The structure here will need to change. Rather than changing the note, we'll need to show the change with two support statements, as in this form:

   ```json
   [
     {
       "version_added": "98"
     },
     {
       "version_added": "83",
       "version_removed": "98",
       "notes": "Supported on macOS only."
     }
   ]
   ```

1. It's probably just a little early to land this PR. Historically, there's been a little reluctance to land Chrome Canary data, while beta tends to have a better track record of shipping as expected.

queengooborg:
Hi @reillyeon!  We haven't heard from you in a while -- do you plan to update this PR based upon the suggestions above?

reillyeon:
Yes, I was a little early in posting this change but plan to update it soon.

foolip (approved):
Thanks @reillyeon!

foolip:
I'll hold off merging and defer to @ddbeck since he requested changes in an earlier review.

ddbeck (approved):
Looks good. Thank you! 🎉

PR #13813, opened 2021-11-30
Add Safari versions for RTCRtpReceiver API

queengooborg:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `RTCRtpReceiver` API.  Honestly speaking, I have no clue how to test support for these features, and they're the last ones remaining that have true/null values.  This PR is simply filling in version numbers for the sake of completing the version numbers, and rely on the community to make changes if needed.

### DIFF
--- a/api/RTCRtpReceiver.json
+++ b/api/RTCRtpReceiver.json
@@ -235,10 +235,10 @@
                 "version_added": "43"
               },
               "safari": {
-                "version_added": null
+                "version_added": "12.1"
               },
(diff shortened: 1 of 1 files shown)


foolip (approved):
This was implemented in https://github.com/WebKit/WebKit/commit/e6d3663cfdb424aa1a684ad0d71169683f40f002 but I can't figure out from the source whether video tracks are supported. Let's just assume that they are until someone tells us otherwise.

PR #13820, opened 2021-12-01
Add a note of URL Constructor in old Edge

aleen42:
#### Summary

Add a note of URL Constructor in old Edge to specify a special behaviour.

```js
// Chrome, Firefox, Node, New Edge
new URL('', 'http://www.test.com?sid=1').searchParams.get('sid'); // => "1"
// Old Edge
new URL('', 'http://www.test.com?sid=1').searchParams.get('sid'); // => null
```

### DIFF
--- a/api/URL.json
+++ b/api/URL.json
@@ -124,7 +124,8 @@
               "version_added": "25"
             },
             "edge": {
-              "version_added": "12"
+              "version_added": "12",
+              "notes": "Before Edge 79, query arguments in the base URL argument are removed when calling the <code>URL</code> constructor."
             },
             "firefox": {
               "version_added": "26"


api/URL.json
```diff
@@ -124,7 +124,8 @@
               "version_added": "25"
             },
             "edge": {
-              "version_added": "12"
+              "version_added": "12",
+              "notes": "In old Edge, query arguments of base URL will be stripped when calling the <code>URL</code> constructor with an empty URL."
```
queengooborg, line 128:
```suggestion
              "notes": "Before Edge 79, query arguments in the base URL argument are removed when calling the <code>URL</code> constructor."
```

queengooborg (changes requested):
Thanks for your PR!  I can confirm this behavior in Edge 15 and 18 through BrowserStack.  I've made some suggestions to the note to include a specific version number (rather than using "old" and "new") as well as some other wording changes.

queengooborg (approved):
Thanks, looks good to me!

PR #13840, opened 2021-12-03
Samesite lax

hamishwillee:
FF96 sets a number of preferences true by default in https://bugzilla.mozilla.org/show_bug.cgi?id=1617609
- network.cookie.sameSite.laxByDefault
- network.cookie.sameSite.noneRequiresSecure
- network.cookie.sameSite.schemeful

This updates BCD to reflect the changes.

Other docs work for this is in https://github.com/mdn/content/issues/10857

### DIFF
--- a/http/headers/set-cookie.json
+++ b/http/headers/set-cookie.json
@@ -485,6 +485,7 @@
                   },
                   {
                     "version_added": "86",
+                    "version_removed": "89",
                     "flags": [
                       {
                         "type": "preference",
@@ -500,6 +501,7 @@
                   },
                   {
                     "version_added": "86",
+                    "version_removed": "89",
                     "flags": [
                       {
                         "type": "preference",
@@ -515,6 +517,7 @@
                   },
                   {
                     "version_added": "86",
+                    "version_removed": "89",
                     "flags": [
                       {
                         "type": "preference",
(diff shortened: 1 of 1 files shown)


http/headers/set-cookie.json
```diff
@@ -507,26 +519,38 @@
                     }
```
hamishwillee, line 519: Above here is edge, which adds support for schemeful behind a preference `#schemeful-same-site` in version 86. Now Chrome 89 removed this flag. Is it possible Edge did too, and this was just omitted?

I looked for the flag in edge:flags and nothing came up on Edge 96.

queengooborg: I'm pretty confident this would have been enabled by default in Edge 89 as well, and that the data just wasn't mirrored from Chrome.  Would you be willing to update this while we're here?

hamishwillee:
> I'm pretty confident this would have been enabled by default in Edge 89 as well, and that the data just wasn't mirrored from Chrome. Would you be willing to update this while we're here?

@queengooborg Done! I also aligned Opera (desktop) for the same reason.

https://caniuse.com/?search=Samesite%3DLax also indicates that SameSite=Lax is true by default on Opera for Android from version 64. Not sure if we can trust/use this data? Here we say "No". 
I "suspect" but don't know that perhaps Opera on Android might also support this (currently says "No"). The problem s

ddbeck:
> https://caniuse.com/?search=Samesite%3DLax also indicates that SameSite=Lax is true by default on Opera for Android from version 64.

In this case, caniuse is reporting from BCD (note the _Support data for this feature provided by_ bit at the bottom), so it's not really an independent data point. Some of caniuse relies on other data, but they won't have the MDN logo.

If this were my PR, I'd mirror Opera Android as well (though it doesn't support flags, so I'd leave out that statement).

http/headers/set-cookie.json
```diff
@@ -497,49 +511,73 @@
...
+                      {
+                        "type": "preference",
+                        "name": "#schemeful-same-site",
+                        "value_to_set": "Enabled"
+                      }
+                    ]
+                  }
+                ],
                 "opera_android": {
                   "version_added": false
```
hamishwillee, line 582: Add Opera on Android version for Blink 89

```suggestion
                  "version_added": ""63
```

hamishwillee:
Thanks @ddbeck @queengooborg . Done (updated as per the last comment, opera android supports schemeful samesite).

I think ready to merge.

ddbeck (approved):
Thank you! 🎉

PR #13841, opened 2021-12-03
Update screen.orientation.lock() support data since desktop browsers reject with NotSupportedError

makotokato:
#### Summary
In Gecko, screen orientation lock was implemented for old Firefox Android only (Fennec), and Fenix doesn't yet.
Also, Chromium has screen orientation lock method, but it always return rejected promise.

see related issues links.

#### Related issues
- Chrome desktop: https://bugs.chromium.org/p/chromium/issues/detail?id=1274012
- Chrome OS: https://bugs.chromium.org/p/chromium/issues/detail?id=1220784
- Firefox Android: https://bugzilla.mozilla.org/show_bug.cgi?id=1697647


saschanaz:
I guess the same applies to Opera too?

makotokato:
> I guess the same applies to Opera too?

Opera is same result as Chrome. I will update this.

PR #13879, opened 2021-12-06
Link Chrome and Safari bugs for initial-letter

foolip:

PR #13883, opened 2021-12-06
Fix WebKit bug URLs for initial-letter

foolip:
This was copypasta in https://github.com/mdn/browser-compat-data/pull/13879.


Elchi3 (approved):
Sorry for my sloppy review and thanks for catching this! 👍

PR #13896, opened 2021-12-06
FF96 Web Lock API supported

hamishwillee:
FF96 adds support for web locks API (in this release was done first in Nightly [bug1739233](https://bugzilla.mozilla.org/show_bug.cgi?id=1739233) and then release [bug 1740044](https://bugzilla.mozilla.org/show_bug.cgi?id=1740044).

Further, it seems to have first been implemented in FF93 behind pref ([bug 1725734](https://bugzilla.mozilla.org/show_bug.cgi?id=1725734)). Since android doesn't support preferences, I've indicated support there from FF96.

Associated with docs work in https://github.com/mdn/content/issues/10859

### DIFF
--- a/api/Lock.json
+++ b/api/Lock.json
@@ -14,11 +14,24 @@
           "edge": {
             "version_added": "79"
           },
-          "firefox": {
-            "version_added": false
-          },
+          "firefox": [
+            {
+              "version_added": "96"
+            },
+            {
+              "version_added": "93",
+              "version_removed": "96",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "dom.weblocks.enabled",
+                  "value_to_set": "true"
+                }
+              ]
+            }
+          ],
           "firefox_android": {
-            "version_added": false
+            "version_added": "96"
           },
           "ie": {
             "version_added": false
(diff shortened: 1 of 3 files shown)


saschanaz:
> Since android doesn't support preferences, I've indicated support there from FF96.

Is that specified in some policy? Nightly does support prefs, but sounds kinda reasonable to ignore it...

hamishwillee:
>> Since android doesn't support preferences, I've indicated support there from FF96.

> Is that specified in some policy? Nightly does support prefs, but sounds kinda reasonable to ignore it...

@saschanaz  The theory as per https://github.com/mdn/browser-compat-data/issues/9679 is that Firefox on Android does not support setting any flags. If this is not the case, perhaps this needs to be revisited. How do you set flags on Android nightly?

saschanaz:
Firefox Nightly supports about:config page, just as the desktop Firefox does. 👀

hamishwillee:
Thanks @saschanaz - lets see how this falls out with @ddbeck - because I suspect this might put some of our other data wrong.

ddbeck:
Firefox on Android is a weird case. Since only Betas and Nightlies can set flags, we shouldn't record flag data for numbered support statements (that is, for cases like this, where `"version_added": "96"`) because the flag data becomes functionally incorrect when that release reaches stable.

Though I suppose we _could_ record flags for features gated to Nightly, not yet riding to release (i.e., cases where `"version_added": "preview"`), since those statements would be correct continuously.

ddbeck (approved):
I think this looks good. Thank you! 🎉

PR #13902, opened 2021-12-07
IntersectionObserver is not experimental

hamishwillee:
Most of the methods on [IntersectionObserver](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver) are marked as experimental, but all are supported by multiple browser engines. This removes that tagging.

@ddbeck I'm actually here for some advice on whether I need to record a bug/compatibility issue for this API, and if so how. Specifically, the [`IntersectionObserver` constructor](https://developer.mozilla.org/en-US/docs/Web/API/IntersectionObserver/IntersectionObserver#parameters) can take an option object that specifies a rootmargin. The spec and chrome allow for an empty string to be specified for this option, but until [bug 1738791](https://bugzilla.mozilla.org/show_bug.cgi?id=1738791) was fixed in FF96 Firefox raised an exception.

Does this need to be recorded. If so, what's the best way?

### DIFF
--- a/api/IntersectionObserver.json
+++ b/api/IntersectionObserver.json
@@ -43,7 +43,7 @@
           }
         },
         "status": {
-          "experimental": true,
+          "experimental": false,
           "standard_track": true,
           "deprecated": false
         }
@@ -92,7 +92,7 @@
             }
           },
           "status": {
-            "experimental": true,
+            "experimental": false,
             "standard_track": true,
             "deprecated": false
           }
@@ -190,7 +190,7 @@
             }
           },
           "status": {
-            "experimental": true,
+            "experimental": false,
             "standard_track": true,
             "deprecated": false
           }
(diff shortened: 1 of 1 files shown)


ddbeck:
The immediate changes here look good. As for this parameter, I think I'd put a note of the style "Before Firefox 96, x throws y when condition z."

api/IntersectionObserver.json
```diff
@@ -64,7 +64,8 @@
               "version_added": "15"
             },
             "firefox": {
-              "version_added": "55"
+              "version_added": "55",
+              "notes": "Before version 96, the constructor throws a <code>DOMException</code> if the <code>options.rootMargin</code> option is passed an empty string (see <a href='https://bugzil.la/1738791'>bug 1738791</a>)."
```
hamishwillee, line 68: Thanks for advice @ddbeck Added note here. Should be ready for review.

ddbeck (approved):
Thank you! 🎉

Issue #12942, opened 2021-10-20
api.fetch - Nodejs not listed under "Browser Compatibility"


jeremylumanbailey:
#### What information was incorrect, unhelpful, or incomplete?
Browser compatibility

#### What did you expect to see?
Nodejs compatibility notes, it appears that the `fetch` API does not exist in Nodejs so it would make sense to list it as "no support". Deno appears to be listed, and Nodejs is usually listed on other docs, but it is not on this page.

#### Did you test this? If so, how?
Nodejs compatibility is not listed. 

<details>
<summary>MDN page report details</summary>

* Query: `api.fetch`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/Fetch_API/Using_Fetch
* Report started: 2021-10-20T03:11:05.024Z

</details>

ddbeck:
Thanks again for reporting this, @jeremylumanbailey. See https://github.com/mdn/browser-compat-data/pull/13919#issuecomment-993803660 for an explanation of why this is unlikely to change.

PR #13919, opened 2021-12-08
Add nodejs compatibility for fetch api

jeremylumanbailey:
#### Summary
fetch API does not exist in nodejs and instead has to use libraries such as [node-fetch](https://www.npmjs.com/package/node-fetch) in order to achieve the same result. 

This adds nodejs non-compatibility to reflect this on the MDN page.

Ran `npm test` without issue

#### Related issues
https://github.com/mdn/browser-compat-data/issues/12942

### DIFF
--- a/api/_globals/fetch.json
+++ b/api/_globals/fetch.json
@@ -27,6 +27,9 @@
           "ie": {
             "version_added": false
           },
+          "nodejs": {
+            "version_added": false
+          },
           "opera": {
             "version_added": "29"
           },
@@ -75,6 +78,9 @@
               "version_added": "39"
             },
             "ie": {
+              "version_added": false
+            },
+            "nodejs": {
               "version_added": false
             },
             "opera": {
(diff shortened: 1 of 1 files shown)


ddbeck:
Hi @jeremylumanbailey and thank you for this PR. I see it's your first for the project—welcome!

This PR alerted me to a problem: we don't document a certain convention for API data. Namely, because many, many Web APIs are unsupported in Node.js, we think it would add needless noise to include that fact in thousands of compat tables. Instead, we omit Node for most Web APIs (while including it in every JS feature that we track) and only include it for those APIs that (exceptionally) have a Node.js implementation. To fix the lack of documentation, I opened a new issue, https://github.com/mdn/browser-compat-data/issues/14009.

Unfortunately, that means this PR won't be merged. Although I'm not able to merge this, I want to thank you nonetheless for bringing this problem to light.

PR #13945, opened 2021-12-11
Require real values for Samsung Internet for api/

queengooborg:
We no longer have any nonreal values for Samsung Internet for the API folder -- let's keep it that way!


Elchi3 (approved):
Getting closer to "blockMany". Nice! 👍

PR #13974, opened 2021-12-13
HTTP Clear-Site-Data cache is behind pref

### LINKED ISSUE #13942: Issue with "Clear-Site-Data": Firefox doesn’t support the "cache" parameter
MDN URL: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Clear-Site-Data

Support information is unfortunately inaccurate for Firefox as support is not "full"

#### Specific section or headline?
Firefox support is not "full" :-(

Sadly Firefox has removed support for the "cache" option.
I filed a bug indicating that it doesn't work: https://bugzilla.mozilla.org/show_bug.cgi?id=1744611
But it was marked as "RESOLVED WONTFIX" because of this issue that forced its removal: https://bugzilla.mozilla.org/show_bug.cgi?id=1671182 based on this discussion of issues with supporting it: https://github.com/privacycg/storage-partitioning/issues/11#issuecomment-662932069

Created a test case, filed a bug, had it rejected as the feature has been intentionally blocked (due to apps sharing a domain/subdomain) :-(

<details>

* Folder: `en-us/web/http/headers/clear-site-data`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Clear-Site-Data
* GitHub URL: https://github.com/mdn/content/blob/main/files/en-us/web/http/headers/clear-site-data/index.md
* Last commit: https://github.com/mdn/content/commit/abf2061ec81ba17f987ecea93747085bde382654
* Document last modified: 2021-10-27T03:40:13.000Z

</details>

hamishwillee:
Fixes #13942

The `cache` was removed behind a pref in FF94: https://bugzilla.mozilla.org/show_bug.cgi?id=1671182

The fix marks versions support as partial. It also removes the old version and adds a (partial) support behind a pref for FF94.
Note that the reasons why support is partial were not detailed so I have not added notes about that. Might reasonably add bug link if you wished?

### DIFF
--- a/http/headers/clear-site-data.json
+++ b/http/headers/clear-site-data.json
@@ -66,7 +66,6 @@
               "firefox": [
                 {
                   "version_added": "94",
-                  "partial_implementation": true,
                   "flags": [
                     {
                       "type": "preference",
@@ -77,14 +76,12 @@
                 },
                 {
                   "version_added": "63",
-                  "version_removed": "94",
-                  "partial_implementation": true
+                  "version_removed": "94"
                 }
               ],
               "firefox_android": {
                 "version_added": "63",
-                "version_removed": "94",
-                "partial_implementation": true
+                "version_removed": "94"
               },
               "ie": {
                 "version_added": false
--- a/http/headers/clear-site-data.json
+++ b/http/headers/clear-site-data.json
@@ -70,7 +70,7 @@
                   "flags": [
                     {
                       "type": "preference",
-                      "name": "#privacy.clearsitedata.cache.enabled",
+                      "name": "privacy.clearsitedata.cache.enabled",
                       "value_to_set": "true"
                     }
                   ]
--- a/http/headers/clear-site-data.json
+++ b/http/headers/clear-site-data.json
@@ -63,11 +63,28 @@
               "edge": {
                 "version_added": "≤79"
               },
-              "firefox": {
-                "version_added": "63"
-              },
+              "firefox": [
+                {
+                  "version_added": "94",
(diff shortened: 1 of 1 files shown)


http/headers/clear-site-data.json
```diff
@@ -63,11 +63,28 @@
...
-                "version_added": "63"
-              },
+              "firefox": [
+                {
+                  "version_added": "94",
+                  "partial_implementation": true,
+                  "flags": [
+                    {
+                      "type": "preference",
+                      "name": "#privacy.clearsitedata.cache.enabled",
```
ddbeck, line 73: Firefox doesn't use the hash notation

```suggestion
                      "name": "privacy.clearsitedata.cache.enabled",
```

ddbeck (changes requested):
Thanks, Hamish. One nit in a line comment.

 I think for the partials here, either we need to specify what's incomplete about the implementation. Looking at things you've linked to (and the BCD bug besides), I'm not sure there's anything obviously incomplete about it, so maybe just get rid of the partial here?

hamishwillee:
Thanks! I've made the change you suggested. You're right that it is not clear why this is "partial" or not. The bugs do mention that this is moved behind the preference because the spec is unclear and perhaps they went too early with the implementation. From which I infer our implementation is "somewhat complete". Anyway, as we can't characterise that, happy to remove the markup.

ddbeck (approved):
Looks good. Thank you! 🎉

KakaNguyen86vn (approved):

PR #13987, opened 2021-12-14
FF96 ElementInternals.willValidate behind preference

hamishwillee:
Fixes #13976

[Late breaking news](https://github.com/mdn/content/pull/11165#issuecomment-992257770) is that this API is behind a preference. I've added the preference for FF96 desktop, and set FF96 android to false as per normal policy (since it has no flag setting API on releases other than nightly).

### DIFF
--- a/api/ElementInternals.json
+++ b/api/ElementInternals.json
@@ -573,10 +573,17 @@
               "version_added": "79"
             },
             "firefox": {
-              "version_added": "96"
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
You probably want to set dev-doc-needed on https://bugzilla.mozilla.org/show_bug.cgi?id=1745378 to be informed when it ships

PR #13989, opened 2021-12-14
HTTP Sec-Ch-UA-Bitness header in Chrome 93

hamishwillee:
The header is in the spec, and has an implementation in from Chrome 93. I couldn't find an announcement of that version, but it is live on my local version and browserstack testing indicates first version it went in was 93. There is no corresponding opera on android or samsunginternet version released yet.

Note it is mentioned as planned for delivery in https://chromestatus.com/feature/5733498725859328

### DIFF
--- a/http/headers/sec-ch-ua-bitness.json
+++ b/http/headers/sec-ch-ua-bitness.json
@@ -1,0 +1,56 @@
+{
+  "http": {
+    "headers": {
+      "Sec-CH-UA-Bitness": {
+        "__compat": {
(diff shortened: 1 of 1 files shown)


hamishwillee:
PS Does MDN have a team login to browserstack I might use? For this I used up my free 1 minute of testing :-)

ddbeck (approved):
Looks good. Thank you! 🎉

re: BrowserStack: would you email me from your preferred email address for this? I can send you an invite.

PR #13990, opened 2021-12-14
HTTP Sec-CH-UA-Full-Version header deprecated

hamishwillee:
`Sec-CH-UA-Full-Version` header is deprecated; replaced by `Sec-CH-UA-Full-Version-List`. See spec here: https://wicg.github.io/ua-client-hints/#sec-ch-ua-full-version

It is still supported by Chrome.


ddbeck (approved):
Looks good. Thank you! 🎉

PR #14003, opened 2021-12-14
Require real values for Opera for api/

queengooborg:
After this, only one more browser (Firefox Android) before we can use `blockMany`!


Elchi3 (approved):
Nice! 🎉

PR #14008, opened 2021-12-14
Unmark crossOriginIsolated as experimental

lukewarlow:
#### Summary

crossOriginIsolated is no longer experimental, this PR updates the data to reflect that.

Based on existing support data it's available in all three browser engines.


ddbeck (approved):
Thank you! 🎉

PR #14021, opened 2021-12-15
Add IE/Edge versions for api.Document.copy_event

queengooborg:
This PR adds real values for Internet Explorer and Edge for the `copy_event` member of the `Document` API, based upon manual testing.

Test Code Used:
```html
<div id="test">
	<p>Copy me!</p>
</div>

<script>
	document.addEventListener('copy', function() {
		alert('Copied!');
	});
</script>
```

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -1750,7 +1750,7 @@
               "version_added": "18"
             },
             "edge": {
-              "version_added": "≤18"
+              "version_added": "12"
             },
             "firefox": {
               "version_added": "6"
@@ -1759,7 +1759,7 @@
               "version_added": "6"
             },
             "ie": {
-              "version_added": false
+              "version_added": "9"
             },
             "opera": {
               "version_added": "15"


api/Document.json
```diff
@@ -1759,7 +1759,7 @@
               "version_added": "6"
             },
             "ie": {
-              "version_added": false
+              "version_added": "9"
```
foolip, line 1762: Is this simply because `addEventListener` in the test didn't work before IE9?

queengooborg: I tried this with `attachEvent` as well but had no luck there.  I have a hunch that there's something different about `attachEvent` that I'm not aware of, but if that's the case, I recall we talked about treating it as a lack of support.

foolip: We discussed it at some length in https://github.com/mdn/browser-compat-data/pull/12907, but that ultimately landed on IE9. Let's just go with that, live a little and risk being wrong about decade old browser release :)

foolip (approved):

Issue #13977, opened 2021-12-13
css.properties.background-clip - supported unprefixed in Safari 14


devongovett:
#### What information was incorrect, unhelpful, or incomplete?

`background-clip: text` appears to work unprefixed in Safari 14.1 and above. I confirmed it does not work in 13.x. Was not able to test 14 below `.1`.

#### What did you expect to see?

Should be green

#### Did you test this? If so, how?

Locally, in Safari 14.1 and Safari 15, and in Browserstack in Safari 13.x.

<details>
<summary>MDN page report details</summary>

* Query: `css.properties.background-clip`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/CSS/background-clip
* Report started: 2021-12-13T04:22:43.378Z

</details>

devongovett:
Example: https://codepen.io/devongovett/pen/OJxbQBZ

PR #14030, opened 2021-12-15
Correct Safari support for background-clip

lukewarlow:
#### Summary

This PR updates Safari support data for the background-clip property to show it's supported as of Safari 14.

Based on issue it definitely wasn't supported before Safari 14. And definitely was from at least Safari 14.1. 

I have verified it works on iOS 14.5 and Safari 15.2 personally.

#### Related issues

Fixes https://github.com/mdn/browser-compat-data/issues/13977

### DIFF
--- a/css/properties/background-clip.json
+++ b/css/properties/background-clip.json
@@ -82,16 +82,26 @@
                 "notes": "Opera accepts alternate synonyms to its values: <code>padding</code>, <code>border</code>, and <code>content</code>."
               }
             ],
-            "safari": {
-              "version_added": "3",
-              "prefix": "-webkit-",
-              "notes": "Safari accepts alternate synonyms to its values: <code>padding</code>, <code>border</code>, and <code>content</code>."
-            },
-            "safari_ios": {
-              "version_added": "1",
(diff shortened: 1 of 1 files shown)


lukewarlow:
I've looked but can't find any webkit commits to help confirm whether it was 14 or 14.1 that it was added in.

ddbeck:
@queengooborg do you know if there's an easy way to test 14 versus 14.1?

If not, I'm OK with merging as-is.

queengooborg:
SauceLabs has both versions available, that's the easiest way I can think of?

queengooborg:
There's a little trick to it, yeah -- my apologies, I thought I wrote up how to do so but apparently not...  Alas, I just found out this won't work for Safari 14 + 15, but I'll leave the trick here anyways:

In SauceLabs, when selecting Safari 10 through 13, the "OS Version" dropdown will include two different macOS versions.  Simply select the older of the two, and you're good to go!

ddbeck (approved):
OK, since I can't directly test this to be sure it wasn't later, I'm going to merge it as written. Thank you, @lukewarlow! 🎉

PR #14035, opened 2021-12-16
Documenting the titlebar-area-* env variables

captainbrosset:
#### Summary
The CSS `env()` function was introduce for the `safe-area-inset-*` variables only. But these are no longer the only ones to exist.
Since I'm adding some MDN docs about the new `titlebar-area-*` vars, I'd like the BCD data to give more specific compat data about those.
So I introduced a section for the `safe-inset-area-*` vars (copied the same compat data from the `env()` section, since I think both were introduced at the same time), and I added a section for the `titlebar-area-*` vars.
I got the chrome version numbers by asking the engineer who worked on them, then used the same number for Edge (since they match), and derived the Opera versions from  https://github.com/mdn/browser-compat-data/blob/main/browsers/opera.json (and the other one for android).

I ran the `npm run test` command to make sure the json structure was right, and spec urls compliant.
You can see the chromium CLs where these variables were introduced here:
https://chromium-review.googlesource.com/c/chromium/src/+/2545685
https://chromium-review.googlesource.com/c/chromium/src/+/2826512
https://chromium-review.googlesource.com/c/chromium/src/+/2886844
https://chromium-review.googlesource.com/c/chromium/src/+/2938404

#### Related issues
This is related to this umbrella issue https://github.com/mdn/content/issues/10539
And I'm also adding MDN docs for these in https://github.com/mdn/content/pull/10622


css/properties/custom-property.json
```diff
@@ -112,6 +112,109 @@
               "standard_track": true,
               "deprecated": false
             }
+          },
+          "safe-area-inset": {
```
ddbeck, line 116: Would you mind expanding these to separate features for each keyword? (See my other comments first, however, to simplify the copy-and-paste.)

css/properties/custom-property.json
```diff
@@ -112,6 +112,109 @@
...
+                }
+              },
+              "status": {
+                "experimental": false,
+                "standard_track": true,
+                "deprecated": false
+              }
+            }
+          },
+          "titlebar-area": {
```
ddbeck, line 166: Some goes for these keywords

css/properties/custom-property.json
```diff
@@ -112,6 +112,109 @@
...
+            "__compat": {
+              "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/env()",
+              "spec_url": "https://wicg.github.io/window-controls-overlay/#title-bar-area-env-variables",
+              "description": "Window Controls Overlay variables <code>titlebar-area-x</code>, <code>titlebar-area-y</code>, <code>titlebar-area-width</code>, and <code>titlebar-area-height</code>",
+              "support": {
+                "chrome": {
+                  "version_added": "92",
+                  "notes": "Linux support was added in 93."
```
ddbeck, line 174: OS limitations like this require `partial_implementation`. See https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#operating-system-limitations-imply-partial_implementation In this case it means that you'll need two support statements:

1. From Version 93, no notes or partial.
2. From version 92 to 93, with a note and partial.

Also, typically, notes are written in the present tense (as if from the perspective of the version(s) to which it applies). Something like this: "Before version 93, Linux is unsupported."

ddbeck (changes requested):
Thanks for this PR, @captainbrosset. A few suggestions for you in line comments. Thank you again!

css/properties/custom-property.json
captainbrosset: Ah I see, so that they each show up individually in the compat tables. Makes sense.

css/properties/custom-property.json
captainbrosset: Will do.

css/properties/custom-property.json
captainbrosset: Awesome. Thanks for the tip. Now I know how to do this properly.

captainbrosset:
Thanks for the review @ddbeck . All comments should be addressed now.

ddbeck (approved):
Thank you! 🎉

PR #14039, opened 2021-12-16
Add Edge versions for api.RTCPeerConnection.track_event

queengooborg:
This PR adds real values for Microsoft Edge for the `track_event` member of the `RTCPeerConnection` API.  The data was copied from its event handler counterpart.


foolip:
Can you update `RTCTrackEvent` data at the same time? It should match, hopefully.

foolip (approved):

PR #14048, opened 2021-12-16
Add spec_url to HTTP Link header

mb:
A suggestion to partly replace mdn/content#11121 by adding the spec data to this repository.

I'm fine with both ways of doing it. Maybe removing the `support`-attribute in the json is also an option instead of setting all platforms to `null`.

### DIFF
--- a/http/headers/link.json
+++ b/http/headers/link.json
@@ -1,0 +1,55 @@
+{
+  "http": {
+    "headers": {
+      "Link": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/HTTP/Headers/Link",
+          "support": {
+            "chrome": {
+              "version_added": null
+            },
+            "chrome_android": {
+              "version_added": null
+            },
+            "edge": {
+              "version_added": null
+            },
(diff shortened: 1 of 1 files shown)


mb:
I don't understand why the test is failing. Error message:

```diff
        spec_urls only use allow listed hosts by w3c/browser-specs (and our exception list):

      AssertionError [ERR_ASSERTION]: Invalid specification host(s) found. Try a more current specification URL and/or
      check if the specification URL is listed in https://github.com/w3c/browser-specs.
      + expected - actual

      -[
      -  "https://www.rfc-editor.org/rfc/rfc8288#section-3"
      -]
      +[]
```

but rfc-editor.org is used in all other `http/headers` files with rfcs as spec, e.g. https://github.com/mdn/browser-compat-data/blob/2134aa3a63c198b50791152d67699dcac081d13c/http/headers/forwarded.json#L7

ddbeck:
Thanks for this PR, @mb. mdn/content#11121 seems to have taken care of the purpose of this PR so I'm going to close it. That said, I saw this was your first to the project—welcome!

PR #14059, opened 2021-12-17
Update mdn_urls for HTMLInputElement api

AnilSeervi:
#### Summary

- Fixes `mdn_url` those lead to 404
#### Related issues

Fixes mdn/content#11248

### DIFF
--- a/api/HTMLInputElement.json
+++ b/api/HTMLInputElement.json
@@ -191,7 +191,7 @@
       },
       "autocomplete": {
         "__compat": {
-          "mdn_url": "https://developer.mozilla.org/en-US/docs/Web/HTML/Element/input#attr-autocomplete",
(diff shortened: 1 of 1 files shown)


queengooborg:
Hey there, thank you for your PR!

This was a good attempt to resolve the issue, but we want to make sure that we're linking to API-specific pages for our API BCD, as there are compatibility (and sometimes functionality) differences between the API member and its HTML attribute counterpart.  If you're running into 404-ing URLs in BCD, we'd prefer to simply remove the URL, as it's a way we can identify what MDN pages still need to be written.

If you'd like to perform the URL removal in this PR, please feel free to!

AnilSeervi:
I couldn't find API-specific pages for any of the 404-ing urls so removed all the 404 urls as u suggested :)

PR #14071, opened 2021-12-18
Remove MediaSessionAction[Details] from BCD

### LINKED ISSUE #4557: Should MediaSession.setActionHandler have subentries for actions?
Chrome is [adding `stop` in 77](https://www.chromestatus.com/feature/6140125320839168). (There's a [whole list](https://w3c.github.io/mediasession/#enumdef-mediasessionaction) in the spec.) What I'm really asking is what's the preferred way to note this in BCD?

queengooborg:
This PR removes `MediaSessionAction` and `MediaSessionActionDetails` from BCD.  These features are a dictionary, enum, or WebIDL typedef and should not be included in BCD.

Closes #4557.


hamishwillee:
@queengooborg When you are removing these dictionaries, are you also "generally" checking that the data for the things that use them is correct? There is no issue if the caller that takes this object has matching versions, but I ran into one case (RTCIce something) where the versions did not match - so we were effectively throwing away information about options to a method changing over time.

Note, I don't think it matters for this case.

queengooborg:
I am, yes!  I'm giving a quick glance at the versions of each subfeature, and if they all match with the method's version, I'm refraining from creating subfeatures for the appropriate method.

hamishwillee:
Thanks for confirming :-)

ddbeck (approved):
Thank you! 🎉

PR #14077, opened 2021-12-20
@media type speech is deprecated

hamishwillee:
In the [Media Queries Level 4](https://drafts.csswg.org/mediaqueries/#valdef-media-speech) editors draft the @media option of speech is listed as deprecated. 

If this goes through, will update matching docs as per https://github.com/mdn/content/issues/11317

### DIFF
--- a/css/at-rules/media.json
+++ b/css/at-rules/media.json
@@ -1751,56 +1751,6 @@
             }
           }
         },
-        "speech_type": {
-          "__compat": {
-            "description": "<code>speech</code> media type",
-            "support": {
-              "chrome": {
-                "version_added": false
-              },
-              "chrome_android": {
-                "version_added": false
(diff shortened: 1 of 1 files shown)


ddbeck (changes requested):
This is unsupported and has been for years and meets [our irrelevance criteria](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-features). How about we remove it instead?

Edited to add: we'd want to scrub it from the MDN page as well.

hamishwillee:
@ddbeck Yes, that makes more sense. Done. Follow up work in MDN done here: https://github.com/mdn/content/pull/11999

ddbeck (approved):
Thank you! 🎉

PR #14095, opened 2021-12-21
More compat data for WindowControlsOverlay

captainbrosset:
#### Summary
The Window Controls Overlay feature is actually available starting with Chrome and Edge 98 (which is when it's out of origin trial). We had originally documented 97, and I'm updating the version here.

I'm also adding the data for the WindowControlsOverlay JS interface. The reason for this is that I'm writing some reference documentation for it in https://github.com/mdn/content/pull/11368.

#### Related issues
See https://github.com/mdn/content/pull/11368 and https://github.com/mdn/content/issues/10539 for more context on this change.


api/WindowControlsOverlay.json
```diff
@@ -0,0 +1,201 @@
...
+            "version_added": false
+          }
+        },
+        "status": {
+          "experimental": true,
+          "standard_track": true,
+          "deprecated": false
+        }
+      },
+      "geometrychange_event": {
```
ddbeck, line 51: Is this right? The linked spec seems to say the event is "change" (implying `change_event` here). I haven't actually used the API though, so it's possible the actual implementation is different

api/WindowControlsOverlayGeometryChangeEvent.json
```diff
@@ -0,0 +1,201 @@
...
+            },
+            "samsunginternet_android": {
+              "version_added": false
+            },
+            "webview_android": {
+              "version_added": false
+            }
+          },
+          "status": {
+            "experimental": false,
```
ddbeck, line 95: For consistency with the parent feature

```suggestion
            "experimental": true,
```

ddbeck (changes requested):
Thanks for the PR. A few suggestions for your, @captainbrosset.

Also, for posterity, would you mind linking to a Chrome Status entry or bug or something that shows that this shipped in the version noted? Nice to have when revisiting this data in the future. Thank you again!

api/WindowControlsOverlay.json
```diff
@@ -0,0 +1,201 @@
...
+          "webview_android": {
+            "version_added": false
+          }
+        },
+        "status": {
+          "experimental": true,
+          "standard_track": true,
+          "deprecated": false
+        }
+      },
```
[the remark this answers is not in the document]
captainbrosset: No, you can access the only instance via navigator.windowControlsOverlay, but can't construct new instances.

api/WindowControlsOverlay.json
```diff
@@ -0,0 +1,201 @@
...
+      "geometrychange_event": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/WindowControlsOverlay/ongeometrychange",
```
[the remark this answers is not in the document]
captainbrosset: Ah, thanks. Indeed, I changed this in my mdn/content PR, but forgot to change the URL here. Doing it now.

api/WindowControlsOverlay.json
captainbrosset: Yeah, there was the same confusion with the event name on the MDN PR. But testing live in chromium shows that the event name is geometrychange. So I think we should keep it like it is here.

api/WindowControlsOverlayGeometryChangeEvent.json
captainbrosset: Done

captainbrosset:
Here is the chromestatus.com feature entry: https://chromestatus.com/feature/5741247866077184

captainbrosset:
@ddbeck thanks for the review. I think I addressed all of your comments.

api/WindowControlsOverlay.json
ddbeck: Agreed.

ddbeck (approved):
Looks great. Thank you! 🎉

PR #14128, opened 2021-12-24
Update svg: paint-order support in firefox

smalllong:
#### Summary

Update svg: paint-order support in firefox
reference: https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/21#svg

### DIFF
--- a/svg/attributes/paint-order.json
+++ b/svg/attributes/paint-order.json
@@ -16,7 +16,7 @@
               "version_added": "≤79"
             },
             "firefox": {
-              "version_added": true
+              "version_added": "21"
             },
             "firefox_android": {
               "version_added": true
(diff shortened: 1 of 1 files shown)


svg/attributes/paint-order.json
```diff
@@ -16,7 +16,7 @@
               "version_added": "≤79"
             },
             "firefox": {
-              "version_added": true
+              "version_added": "21"
             },
             "firefox_android": {
               "version_added": true
```
queengooborg, line 22:
```suggestion
              "version_added": "21"
```

queengooborg (changes requested):
Nice find, thanks!  May we also set Firefox Android to the same version number?

svg/attributes/paint-order.json
smalllong: I don't know about Android version so please leave it unchanged

ddbeck (approved):
I can't find any evidence that Android was unsupported (Firefox desktop and Android are typically the same). I'm going to apply the suggested change and merge it. Thank you, @smalllong! 🎉

PR #14129, opened 2021-12-24
Add __meta property to BCD build

### LINKED ISSUE #7358: Expose BCD version in exported object?
While one can get the BCD version in-script by running `require(require.resolve('@mdn/browser-compat-data/package.json')).version`, I was thinking that we should perhaps expose the version ourselves in our exported object?  One could then get it via `bcd.__version` (or something along those lines).

queengooborg:
This PR fixes #7358 by adding a `__meta` property to the BCD root, containing the build's version number.  This allows for easier access of the version number for NodeJS users, and straight-up allows access for Deno/browser/Python/etc. users entirely.


foolip:
This change is not without risk and I think would require at least a minor version bump. I am fairly skeptical of making this change unless there are multiple requests from BCD consumers.

queengooborg:
I think that we should include this in a major semver bump actually, since it is a potentially breaking change.  In doing so, however, this might help our consumers understand that keys starting with `__` are special in BCD, such as compatibility info or otherwise.

I think if we are to do this, we should land it around the same time we start releasing ESM builds.

caugner (commented):
LGTM, but ...

1. ... wouldn't it make sense to wrap the version in a `__meta` object (with one key `version`), so that we can extend this in the future?
2. ... this is probably a breaking change for consumers that just loop over the data (and this makes me wonder if we have other breaking changes in the pipeline - possibly among your open PRs - that should be bundled with this).

caugner (approved):
LGTM

queengooborg:
I'm going to hold off on merging this for the time being since it is a major semver bump.  I'd like to get other changes merged alongside this one before the next release.

caugner:
@queengooborg Would it make sense to create a `v5` branch for this?

queengooborg:
Maybe so, yeah -- my plan was originally to hold off on a v5 release until #16232 is merged (and maybe the ESLint upgrades), but also to hold off so we get a v4.2.0 release... :D

PR #14130, opened 2021-12-24
Remove mesh SVG element from BCD

### LINKED ISSUE #7413: Remove SVG <mesh> element?
#6941 removed `SVGMeshElement` as unsupported. The [data for `<mesh>`](https://github.com/mdn/browser-compat-data/blob/master/svg/elements/mesh.json) is all `null` values. Can we confirm that it's non-existent and remove this feature?

queengooborg:
This PR removes the irrelevant `mesh` SVG element as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-features). The lack of current support has been confirmed by the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v4.0.0) (by checking the `SVGMeshElement` API), even if the current BCD suggests support.

Fixes #7413.


foolip:
There are references to this in https://developer.mozilla.org/en-US/docs/Web/SVG/Element and perhaps other places, so a content cleanup seems to be in order.

queengooborg:
Content cleanup on its way!  The dedicated page for this element, however, is already removed, so while a content update is in order, it's not blocking this PR.

foolip (approved):

PR #14172, opened 2021-12-27
Update Chromium versions for ANGLE_instanced_arrays API

queengooborg:
This PR updates and corrects the real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `ANGLE_instanced_arrays` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v4.0.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/ANGLE_instanced_arrays

### DIFF
--- a/api/ANGLE_instanced_arrays.json
+++ b/api/ANGLE_instanced_arrays.json
@@ -6,10 +6,10 @@
         "spec_url": "https://www.khronos.org/registry/webgl/extensions/ANGLE_instanced_arrays/",
         "support": {
           "chrome": {
-            "version_added": "32"
+            "version_added": "30"
           },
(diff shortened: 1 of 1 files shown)


queengooborg:
Note: the difference seems to be in OS support -- macOS supports this interface before Windows does.  I updated the Android data as well.  Although there's a strong chance the Android data is correct, it's a tad too challenging to test at the moment and I don't have a testing setup in place for it yet.

foolip:
To avoid changing this back and forth based on collector results from different OSes, can you add a `partial_implementation` statement first?

foolip (approved):

PR #14174, opened 2021-12-27
Update Firefox versions for AnimationEvent API

queengooborg:
This PR updates and corrects the real values for Firefox and Firefox Android for the `AnimationEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v4.0.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/AnimationEvent

### DIFF
--- a/api/AnimationEvent.json
+++ b/api/AnimationEvent.json
@@ -29,10 +29,10 @@
             "version_added": "12"
           },
           "firefox": {
-            "version_added": "6"
+            "version_added": "5"
           },
           "firefox_android": {
-            "version_added": "6"
+            "version_added": "5"
           },
           "ie": {
             "version_added": "10"
@@ -197,10 +197,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "6"
+              "version_added": "5"
             },
             "firefox_android": {
-              "version_added": "6"
+              "version_added": "5"
             },
             "ie": {
               "version_added": "10"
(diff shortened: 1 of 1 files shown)


foolip:
What are the events that use this interface? Should their data also be updated?

queengooborg:
I'm not 100% positive what events are returning as `AnimationEvent`, but I'm pretty sure it's the `onanimation*` events.  Unfortunately though, I can't find any support for them prior to Firefox 54, prefixed or not.  I'll have to look into this more.

It appears that the original "6" came from the wiki migration, but looking through, Firefox 5 seems to match up with the data for the CSS `@keyframes` rule.

foolip:
I guess this is more likely correct. Other entries that are marked Firefox 5:
```
css.properties.animation-delay (-moz- prefix)
css.properties.animation-direction (-moz- prefix)
css.properties.animation-duration (-moz- prefix)
css.properties.animation-fill-mode (-moz- prefix)
css.properties.animation-iteration-count (-moz- prefix)
css.properties.animation-name (-moz- prefix)
css.properties.animation-play-state (-moz- prefix)
css.properties.animation-timing-function (-moz- prefix)
css.properties.animation (-moz- prefix)
svg.attributes.style.class.animation_support
```

foolip (approved):

foolip:
I've sent https://github.com/mdn/browser-compat-data/pull/15262 as a consistency fix.

PR #14188, opened 2021-12-27
Update all browsers versions for Blob API

queengooborg:
This PR updates and corrects the real values for all browsers for the `Blob` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v4.0.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Blob

### DIFF
--- a/api/Blob.json
+++ b/api/Blob.json
@@ -41,7 +41,7 @@
             "version_added": "11"
           },
           "safari": {
-            "version_added": "5.1"
+            "version_added": "6"
           },
           "safari_ios": {
             "version_added": "6"
@@ -209,7 +209,7 @@
               "version_added": "11"
             },
             "safari": {
-              "version_added": "5.1"
+              "version_added": "6"
             },
             "safari_ios": {
               "version_added": "6"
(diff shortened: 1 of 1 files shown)


api/Blob.json
```diff
@@ -98,10 +98,10 @@
               "version_added": "12"
             },
             "safari": {
-              "version_added": "8"
+              "version_added": "6"
```
foolip, line 101: The existing data is paired Safari 5.1 / iOS 6. Should this be too, or is the existing data wrong?

queengooborg: It appeared that Safari's data wasn't getting updated, but that's because our tests are returning `null` since neither `Blob` nor `BlobBuilder` are defined.  Updating all the data now!

foolip: I think there was `WebKitBlob` originally, can you test some prefixed variants?

foolip (approved):

api/Blob.json
foolip: I did some testing, no `WebKitBlob` in Safari 5.1, and also no `WebKitBlobBuilder` in Safari 6. Maybe that was never shipped even though it shows up in WebKit source history a lot.

PR #14202, opened 2021-12-27
Update all browsers versions for CanvasPattern API

queengooborg:
This PR updates and corrects the real values for all browsers for the `CanvasPattern` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v4.0.0).

### DIFF
--- a/api/CanvasPattern.json
+++ b/api/CanvasPattern.json
@@ -72,10 +72,10 @@
               "version_added": false
             },
             "opera": {
-              "version_added": "55"
+              "version_added": "9"
             },
             "opera_android": {
-              "version_added": "48"
+              "version_added": "10.1"
             },
             "safari": {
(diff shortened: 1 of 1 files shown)


api/CanvasPattern.json
```diff
@@ -72,10 +72,10 @@
               "version_added": false
             },
             "opera": {
-              "version_added": "9"
```
foolip, line 75: Did you check that this wasn't in Presto? The update script bails with data like that (multiple ranges) so if it was it wouldn't make the right edit.

queengooborg: Whoops, this change wasn't meant to be here -- reverted!

foolip (approved):
This is more consistent with the `createPattern` data than before.

PR #14209, opened 2021-12-27
Update Firefox versions for Crypto API

queengooborg:
This PR updates and corrects the real values for Firefox and Firefox Android for the `Crypto` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v4.0.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Crypto

### DIFF
--- a/api/Crypto.json
+++ b/api/Crypto.json
@@ -26,10 +26,10 @@
             "version_added": "12"
           },
           "firefox": {
-            "version_added": "26"
+            "version_added": "1"
           },
           "firefox_android": {
-            "version_added": "26"
+            "version_added": "4"
           },
           "ie": {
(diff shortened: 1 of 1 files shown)


api/Crypto.json
```diff
@@ -26,10 +26,10 @@
             "version_added": "12"
           },
           "firefox": {
-            "version_added": "26"
+            "version_added": "1"
```
foolip, line 29: This makes the data look weird. Was there nothing on the `window.crypto` object between Firefox 1 and 26? Or only things that aren't in BCD? Maybe just match the version of the earliest subfeature?

foolip (approved):

PR #14211, opened 2021-12-27
Update all browsers versions for CustomEvent API

queengooborg:
This PR updates and corrects the real values for all browsers for the `CustomEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v4.0.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/CustomEvent

### DIFF
--- a/api/CustomEvent.json
+++ b/api/CustomEvent.json
@@ -6,7 +6,7 @@
         "spec_url": "https://dom.spec.whatwg.org/#interface-customevent",
         "support": {
           "chrome": {
-            "version_added": "15"
+            "version_added": "5"
           },
           "chrome_android": {
             "version_added": "18"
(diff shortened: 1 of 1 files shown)


foolip (commented):
How about the `CustomEvent` constructor? Can you double check that wasn't supported in Chrome 5, so that this isn't a problem with the custom test?

foolip:
@queengooborg can you describe the testing you did for the constructor? I was actually expecting you to find that it *wasn't* supported in Chrome 5, since most event constructors were added more recently.

queengooborg:
I'm not sure what I had done last time, but I re-tested it with the collector now and found the original data is accurate!

foolip (approved):

PR #14214, opened 2021-12-27
Update all browsers versions for DOMRectList API

queengooborg:
This PR updates and corrects the real values for all browsers for the `DOMRectList` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v4.0.0).

Note: this interface is never accessed by name, so the alt. names are pointless to record.  This interface is only accessed via api.Element.getClientRects().

### DIFF
--- a/api/DOMRectList.json
+++ b/api/DOMRectList.json
@@ -4,120 +4,42 @@
       "__compat": {
         "spec_url": "https://drafts.fxtf.org/geometry/#DOMRectList",
         "support": {
-          "chrome": [
-            {
-              "version_added": "61"
-            },
-            {
-              "alternative_name": "ClientRectList",
-              "version_added": "2",
-              "version_removed": "61"
-            }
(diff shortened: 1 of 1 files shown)


foolip:
@queengooborg `DOMRect` also has these aliases and I think the two interfaces should be treated the same way. I agree that the name is pointless, but `DOMRect` does have a bunch of things, notably a constructor, which were added at the same time the name changed. I think the data will look mysterious without the previous name, and think we should keep them here.

If there are other cases like this, maybe we should have a data guidelines.

caugner (commented):
@foolip has made a valid point about consistency with DOMRect:

https://github.com/mdn/browser-compat-data/blob/e054f8994e0ec360857d07e86a2ce4d8f324d22f/api/DOMRect.json#L1-L13

@queengooborg What do you think?

PR #14219, opened 2021-12-27
Update Opera versions for DataTransferItem API

queengooborg:
This PR updates and corrects the real values for Opera and Opera Android for the `DataTransferItem` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v4.0.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/DataTransferItem

### DIFF
--- a/api/DataTransferItem.json
+++ b/api/DataTransferItem.json
@@ -27,7 +27,7 @@
             "version_added": "12"
           },
           "opera_android": {
-            "version_added": false
+            "version_added": "12"
           },
           "safari": {
             "version_added": "5.1"
@@ -75,7 +75,7 @@
               "version_added": "12"
             },
             "opera_android": {
-              "version_added": false
+              "version_added": "12"
             },
             "safari": {
               "version_added": "5.1"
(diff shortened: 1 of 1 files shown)


foolip:
Is this based on testing the real Opera 12 on Android?

queengooborg:
I don't have older browser testing for Android-based browsers set up just yet, so instead, this is based upon testing in the latest Opera Android and then mirroring from Opera Desktop.  I do plan on running tests for every Android browser version in the future!

foolip:
Can you make sure that the PRs say when the data is based on mirroring vs. testing? When it's based on testing I wouldn't doubt it at all, but if mirroring one has to think about assumptions. Here, I think there's a chance that `DataTransferItem` wasn't supported on Opera mobile, because it's part of a drag-and-drop API that I suspect doesn't work on mobile at all.

queengooborg:
I'll replace Opera Android's data with a range then!

foolip (approved):

PR #14233, opened 2021-12-28
Update Edge versions for EXT_disjoint_timer_query API

queengooborg:
This PR updates and corrects the real values for Microsoft Edge for the `EXT_disjoint_timer_query` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v4.0.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/EXT_disjoint_timer_query

### DIFF
--- a/api/EXT_disjoint_timer_query.json
+++ b/api/EXT_disjoint_timer_query.json
@@ -16,7 +16,7 @@
             "notes": "Removed due to the <a href='https://www.vusec.net/projects/glitch/'>GLitch exploit</a>."
           },
           "edge": {
-            "version_added": false
+            "version_added": "80"
           },
           "firefox": {
             "version_added": "51",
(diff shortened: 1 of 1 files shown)


api/EXT_disjoint_timer_query.json
```diff
@@ -16,7 +16,7 @@
             "notes": "Removed due to the <a href='https://www.vusec.net/projects/glitch/'>GLitch exploit</a>."
           },
           "edge": {
-            "version_added": false
+            "version_added": "80"
```
foolip, line 19: Are you sure about this? Did Edge decide to enable this extension, but Chrome still hasn't?

queengooborg: Doing some more digging into this, it looks like it's enabled in macOS on both Chrome and Edge, but not on Windows. Doing some quick testing, it seems that it was removed due to the GLitch exploit as mentioned, and then added back only for macOS in Chromium 70. I'll update the data accordingly!

foolip (approved):

PR #14236, opened 2021-12-28
Update Chrome/Safari versions for Element API

queengooborg:
This PR updates and corrects the real values for Chrome and Safari for the `Element` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v4.0.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Element


foolip (changes requested):
I've sent https://github.com/mdn/browser-compat-data/pull/14408 to remove the notes here. That will cause conflicts.

Not all of the "33" values have been changed to "32", but also I'm pretty sure `innerHTML` and `outerHTML` were around long before Chrome 32. Maybe they were on `HTMLElement` and we need `partial_implementation` entries here? Can you do some testing and edit to match https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#apis-moved-on-the-prototype-chain?

foolip:
@queengooborg I marked https://github.com/mdn/browser-compat-data/pull/14408 as draft since it requires research. Do you want to do that to unblock this?

caugner:
Don't hesitate to mark as ready for review again, but I thought "draft" better mirrors the "not ready" status.

api/Element.json
```diff
@@ -2667,7 +2667,7 @@
               "version_added": "10.1"
             },
             "safari": {
-              "version_added": "6"
+              "version_added": "3.1"
```
foolip, line 2670: After https://github.com/mdn/browser-compat-data/pull/17226 this is the only change that will remain of this PR. This change seems correct and is supported by source archeology:
https://github.com/WebKit/WebKit/commit/74c8babf4d4580c0a65ec8cda910c28d8bb9e35e
https://github.com/WebKit/WebKit/blob/74c8babf4d4580c0a65ec8cda910c28d8bb9e35e/WebCore/Configurations/Version.xcconfig

That's WebKit trunk version 525.1, mapping to Safari 3.1.

foolip (changes requested):
Can you revert the changes other than `getElementsByClassName`?

foolip:
@queengooborg are you making the final change in some other PR?

queengooborg:
Yeah, I'll be revisiting this in a later PR with newer collector results!

foolip:
@queengooborg then I'd have to do the review again unless I remember this PR. I'll create a PR with the evidence I found.

PR #14246, opened 2021-12-28
Update all browsers versions for File API

queengooborg:
This PR updates and corrects the real values for all browsers for the `File` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v4.0.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/File

### DIFF
--- a/api/File.json
+++ b/api/File.json
@@ -145,10 +145,10 @@
               "version_added": false
             },
             "opera": {
-              "version_added": "16"
+              "version_added": "≤15"
             },
             "opera_android": {
-              "version_added": "16"
+              "version_added": "≤14"
             },
             "safari": {
               "version_added": "10"
@@ -197,7 +197,7 @@
               "version_added": "10"
             },
             "opera": {
-              "version_added": "16"
+              "version_added": "≤15"
             },
             "opera_android": {
               "version_added": false
@@ -251,10 +251,10 @@
               "version_added": "10"
             },
             "opera": {
-              "version_added": "16"
+              "version_added": "≤15"
             },
             "opera_android": {
-              "version_added": "16"
+              "version_added": "≤14"
             },
             "safari": {
               "version_added": "7"
@@ -303,7 +303,7 @@
               "version_added": "10"
             },
             "opera": {
-              "version_added": "16"
+              "version_added": "≤15"
             },
             "opera_android": {
               "version_added": false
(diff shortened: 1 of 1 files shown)


api/File.json
```diff
@@ -6,7 +6,7 @@
         "spec_url": "https://w3c.github.io/FileAPI/#file-section",
         "support": {
           "chrome": {
-            "version_added": "13"
+            "version_added": "1"
```
foolip, line 9: Shouldn't some of the members also be supported in Chrome 1?

foolip (changes requested):

api/File.json
queengooborg: It looks like pretty much all the members are actually not defined in Chrome 13, even.  In these PRs, though, I'm only focusing on changes where the version number reports earlier than what BCD says, and leaving the other direction for another time -- I'll be sure to look into the members in the next wave!

foolip: If none of the members were present in Chrome 1, then there's a good chance that the `File` interface didn't actually work and should be `partial_implementation` and/or have some notes. I don't think we should update part of the data and leave it less consistent based just on a `"File" in self` test without understanding if the inconsistency is correct or not.

api/File.json
```diff
@@ -148,7 +148,7 @@
               "version_added": "16"
             },
             "opera_android": {
-              "version_added": false
+              "version_added": "16"
```
foolip, line 151: Can you re-mirror all of the Opera data? It's inconsistent, Chrome/Android 13/18 versions don't map to the same Opera versions on different features.

foolip (changes requested):

api/File.json
```diff
@@ -197,7 +197,7 @@
               "version_added": "10"
             },
             "opera": {
-              "version_added": "16"
+              "version_added": "≤15"
             },
             "opera_android": {
               "version_added": false
```
foolip, line 203:
```suggestion
              "version_added": "≤14"
```

I guess?

api/File.json
```diff
@@ -303,7 +303,7 @@
...
```
foolip, line 309:
```suggestion
              "version_added": "≤14"
```

foolip (changes requested):

foolip (approved):

PR #14257, opened 2021-12-28
Update all versions for Headers API

queengooborg:
This PR updates and corrects the real values for all browsers for the `Headers` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v4.0.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Headers

### DIFF
--- a/api/Headers.json
+++ b/api/Headers.json
@@ -429,7 +429,7 @@
               "version_added": "42"
             },
             "chrome_android": {
-              "version_added": "45"
+              "version_added": "42"
             },
             "deno": {
               "version_added": "1.0"
@@ -447,10 +447,10 @@
               "version_added": false
             },
             "opera": {
-              "version_added": "32"
+              "version_added": "29"
             },
             "opera_android": {
-              "version_added": "32"
+              "version_added": "29"
             },
             "safari": {
               "version_added": "10.1"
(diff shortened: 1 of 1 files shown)


api/Headers.json
```diff
@@ -213,7 +213,7 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Headers/entries",
           "support": {
             "chrome": {
-              "version_added": "45"
+              "version_added": "42"
```
foolip, line 216: Should this be mirrored to Chrome Android too?

api/Headers.json
```diff
@@ -21,7 +21,7 @@
             "version_added": "39"
           },
           "firefox_android": {
-            "version_added": "44"
+            "version_added": "39"
```
foolip, line 24: Is this based on testing Firefox for Android or mirroring? I wouldn't trust mirroring alone, maybe the data is different on purpose currently.

foolip (changes requested):

api/Headers.json
```diff
@@ -580,7 +580,7 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Headers/values",
           "support": {
             "chrome": {
-              "version_added": "45"
+              "version_added": "42"
```
foolip, line 583: Should this change also be mirrored to all other Chromium based browsers?

foolip (approved):

PR #14262, opened 2021-12-28
Update all browsers versions for ImageData API

queengooborg:
This PR updates and corrects the real values for all browsers for the `ImageData` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v4.0.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/ImageData

### DIFF
--- a/api/ImageData.json
+++ b/api/ImageData.json
@@ -15,10 +15,10 @@
             "version_added": "12"
           },
           "firefox": {
-            "version_added": "14"
+            "version_added": "3.5"
           },
           "firefox_android": {
-            "version_added": "14"
+            "version_added": "4"
           },
           "ie": {
             "version_added": "9"
@@ -55,10 +55,10 @@
           "spec_url": "https://html.spec.whatwg.org/multipage/canvas.html#dom-imagedata-dev",
           "support": {
             "chrome": {
-              "version_added": "42"
+              "version_added": "36"
             },
             "chrome_android": {
-              "version_added": "42"
+              "version_added": "36"
             },
             "edge": {
               "version_added": "14"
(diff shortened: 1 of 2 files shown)


api/ImageData.json
```diff
@@ -15,10 +15,10 @@
             "version_added": "12"
           },
           "firefox": {
-            "version_added": "14"
+            "version_added": "3.5"
```
foolip, line 18: This is still later than the api.CanvasRenderingContext2D.createImageData entry. They should match, so can you check which is correct?

api/ImageData.json
```diff
@@ -55,10 +55,10 @@
           "spec_url": "https://html.spec.whatwg.org/multipage/canvas.html#dom-imagedata-dev",
           "support": {
             "chrome": {
-              "version_added": "42"
+              "version_added": "36"
```
foolip, line 58: Should this be mirrored to Opera too?

api/ImageData.json
```diff
@@ -161,10 +161,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "14"
+              "version_added": "3.5"
```
foolip, line 164: This and the following features should probably match the parent feature if that changes to match `createImageData`.

foolip (changes requested):

api/ImageData.json
queengooborg: Our test uses that function to create an `ImageData` instance, and it appears it isn't defined in Firefox 3.  Will correct that data in this PR!

foolip (approved):

PR #14273, opened 2021-12-28
Update Firefox versions for MessageEvent API

queengooborg:
This PR updates and corrects the real values for Firefox and Firefox Android for the `MessageEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v4.0.0).

### DIFF
--- a/api/MessageEvent.json
+++ b/api/MessageEvent.json
@@ -73,10 +73,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "4"
+              "version_added": "26"
             },
             "firefox_android": {
-              "version_added": "4"
+              "version_added": "26"
             },
             "ie": {
               "version_added": "9"
--- a/api/MessageEvent.json
+++ b/api/MessageEvent.json
@@ -183,7 +183,7 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "4"
+              "version_added": "3"
             },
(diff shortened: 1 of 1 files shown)


api/MessageEvent.json
```diff
@@ -18,7 +18,7 @@
             "version_added": "12"
           },
           "firefox": {
-            "version_added": "4"
+            "version_added": "3"
```
foolip, line 21: Some features remain as "4" which could be correct but seems suspect. Can you double check them?

foolip:
@queengooborg the constructor still remains set to "4". Is that correct?

api/MessageEvent.json
queengooborg: Looks like they're all accurate.  Ran the collector in Firefox 3.6 on BrowserStack and double-checked everything still set to Firefox 4!

foolip: The constructor is the only thing still set to "4" and I've confirmed that it *doesn't* work in Firefox 4, it was added later. Can you find the version with http://mdn-bcd-collector.appspot.com/tests/api/MessageEvent/MessageEvent?

queengooborg: Ah, didn't test the other way -- fixed!

foolip (approved):

PR #14274, opened 2021-12-28
Update Chromium versions for MessagePort API

queengooborg:
This PR updates and corrects the real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `MessagePort` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v4.0.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/MessagePort

### DIFF
--- a/api/MessagePort.json
+++ b/api/MessagePort.json
@@ -131,7 +131,7 @@
           "spec_url": "https://html.spec.whatwg.org/multipage/indices.html#event-message",
           "support": {
             "chrome": {
-              "version_added": "4"
+              "version_added": "1"
             },
             "chrome_android": {
               "version_added": "18"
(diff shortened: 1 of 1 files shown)


api/MessagePort.json
```diff
@@ -6,7 +6,7 @@
         "spec_url": "https://html.spec.whatwg.org/multipage/web-messaging.html#message-ports",
         "support": {
           "chrome": {
-            "version_added": "4"
+            "version_added": "1"
```
foolip, line 9: Should the "message_event" and "worker_support" entries also be 1? I'm pretty sure at least the event should be.

foolip (changes requested):

api/MessagePort.json
queengooborg: Updated!  Worker support in general was added in Chrome 4, so that should remain as "4".

foolip (approved):

PR #14287, opened 2021-12-29
Update Chromium/Firefox versions for Navigator API

queengooborg:
This PR updates and corrects the real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) and Firefox (Desktop/Android) for the `Navigator` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v4.0.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Navigator

### DIFF
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -1306,7 +1306,7 @@
               "version_added": "29"
             },
             "firefox_android": {
-              "version_added": false
+              "version_added": "32"
             },
             "ie": {
               "version_added": false
@@ -1893,7 +1893,7 @@
               "version_added": "68"
             },
             "chrome_android": {
-              "version_added": false
+              "version_added": "68"
             },
             "edge": {
               "version_added": "79"
@@ -1911,7 +1911,7 @@
               "version_added": "55"
             },
             "opera_android": {
-              "version_added": false
+              "version_added": "48"
             },
             "safari": {
               "version_added": false
@@ -1920,7 +1920,7 @@
               "version_added": false
             },
             "samsunginternet_android": {
-              "version_added": false
+              "version_added": "10.0"
             },
             "webview_android": {
               "version_added": "68"
@@ -2363,10 +2363,10 @@
               "version_added": false
             },
             "opera": {
-              "version_added": false
+              "version_added": "60"
             },
             "opera_android": {
-              "version_added": false
+              "version_added": "43"
             },
             "safari": {
               "version_added": "15"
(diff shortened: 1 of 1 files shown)


api/Navigator.json
```diff
@@ -3411,15 +3411,15 @@
           "spec_url": "https://w3c.github.io/badging/#setappbadge-method",
           "support": {
             "chrome": {
-              "version_added": "83",
+              "version_added": "81",
```
foolip, line 3414: Should the `clearAppBadge` data also be updated, or weren't the shipped at the same time?

api/Navigator.json
```diff
@@ -1306,7 +1306,7 @@
               "version_added": "29"
             },
             "firefox_android": {
-              "version_added": false
+              "version_added": "32"
```
foolip, line 1309: Is this based on testing Firefox for Android 29-32? Or why not 29 here?

api/Navigator.json
```diff
@@ -3411,15 +3411,15 @@
...
           "support": {
             "chrome": {
-              "version_added": "83",
+              "version_added": "81",
               "partial_implementation": true,
               "notes": "Windows and macOS only."
             },
             "chrome_android": {
-              "version_added": false
+              "version_added": "81"
```
foolip, line 3419: Is this based on testing Chrome for Android? It wouldn't be surprising if this API was shipped at a different time on desktop platforms and Android.

api/Navigator.json
```diff
@@ -4303,7 +4303,7 @@
               "version_added": "8.0"
             },
             "webview_android": {
-              "version_added": false
+              "version_added": "63"
```
foolip, line 4306: Although I don't think WebView can be controlled by WebDriver, the `navigator.webdriver` property should exist and return false, so this seems right.

api/Navigator.json
queengooborg: This was, yes -- the rest of the Android browsers were mirrored, but testing those in their latest versions showed support.

api/Navigator.json
queengooborg: Support for the Gamepad API was added in Firefox Android 32 per our data, so this version number is offset to match that of the Gamepad interface.

foolip (approved):
Before landing, can you update the description to reflect that Firefox data is also updated?

queengooborg:
Done!

foolip (approved):

PR #14289, opened 2021-12-29
Update all browsers versions for Node API

queengooborg:
This PR updates and corrects the real values for all browsers for the `Node` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v4.0.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Node

### DIFF
--- a/api/Node.json
+++ b/api/Node.json
@@ -383,7 +383,7 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "≤4"
+              "version_added": "1"
             },
             "safari_ios": {
               "version_added": "1"
@@ -899,7 +899,7 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "≤4"
+              "version_added": "1"
             },
             "safari_ios": {
               "version_added": "1"
@@ -1095,7 +1095,7 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "≤4"
+              "version_added": "1"
             },
             "safari_ios": {
               "version_added": "1"
@@ -1193,7 +1193,7 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "≤4"
+              "version_added": "1"
             },
             "safari_ios": {
               "version_added": "1"
@@ -1305,7 +1305,7 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "≤4"
+              "version_added": "1"
             },
             "safari_ios": {
               "version_added": "1"
@@ -1455,7 +1455,7 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "≤4"
+              "version_added": "1"
             },
             "safari_ios": {
               "version_added": "1"
--- a/api/Node.json
+++ b/api/Node.json
@@ -135,10 +135,10 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "7"
+              "version_added": "≤4"
             },
             "safari_ios": {
-              "version_added": "7"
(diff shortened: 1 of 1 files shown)


api/Node.json
```diff
@@ -383,10 +383,10 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "7"
+              "version_added": "≤4"
```
foolip, line 386:
```suggestion
              "version_added": "1"
```

Source: https://github.com/WebKit/WebKit/blob/b923c2f4eb7804aae65d67b4c3061d6435acb10d/WebCore/khtml/ecma/kjs_dom.cpp#L168-L169

api/Node.json
```diff
@@ -899,10 +899,10 @@
...
```
foolip, line 902:
```suggestion
              "version_added": "1"
```

Source: https://github.com/WebKit/WebKit/blob/b923c2f4eb7804aae65d67b4c3061d6435acb10d/WebCore/khtml/ecma/kjs_dom.cpp#L170-L171

api/Node.json
```diff
@@ -1095,10 +1095,10 @@
...
```
foolip, line 1098:
```suggestion
              "version_added": "1"
```

Source: https://github.com/WebKit/WebKit/blob/b923c2f4eb7804aae65d67b4c3061d6435acb10d/WebCore/khtml/ecma/kjs_dom.cpp#L156-L157

api/Node.json
```diff
@@ -1193,10 +1193,10 @@
...
```
foolip, line 1196:
```suggestion
              "version_added": "1"
```

Source: https://github.com/WebKit/WebKit/blob/b923c2f4eb7804aae65d67b4c3061d6435acb10d/WebCore/khtml/ecma/kjs_dom.cpp#L158-L159

api/Node.json
```diff
@@ -1305,10 +1305,10 @@
...
```
foolip, line 1308:
```suggestion
              "version_added": "1"
```

Source: https://github.com/WebKit/WebKit/blob/b923c2f4eb7804aae65d67b4c3061d6435acb10d/WebCore/khtml/ecma/kjs_dom.cpp#L184-L185

api/Node.json
```diff
@@ -1455,10 +1455,10 @@
...
```
foolip, line 1458:
```suggestion
              "version_added": "1"
```

Source: https://github.com/WebKit/WebKit/blob/b923c2f4eb7804aae65d67b4c3061d6435acb10d/WebCore/khtml/ecma/kjs_dom.cpp#L172-L173

api/Node.json
```diff
@@ -135,10 +135,10 @@
...
```
foolip, line 138: I can't figure out from source when this was added, so leaving as a range seems fine.

api/Node.json
```diff
@@ -135,10 +135,10 @@
...
             },
             "safari_ios": {
-              "version_added": "7"
+              "version_added": "1"
```
foolip, line 141: Is this based on testing iOS 1? Safari ≤4 + iOS 1 is an unusual pairing of versions.

foolip (changes requested):

api/Node.json
queengooborg: The generated range was actually "≤3" since that's the earliest version the collector will run in, and Safari 3 = iOS 1, but since we don't have a "≤3" range I had to bump it up to "≤4".

foolip: I see, that explains all of the recent Safari ≤4 + iOS 1 changes.

foolip (approved):

PR #14291, opened 2021-12-29
Update all browsers versions for OfflineAudioContext API

queengooborg:
This PR updates and corrects the real values for all browsers for the `OfflineAudioContext` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v4.0.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/OfflineAudioContext

### DIFF
--- a/api/OfflineAudioContext.json
+++ b/api/OfflineAudioContext.json
@@ -430,10 +430,10 @@
               "version_added": false
             },
             "opera": {
-              "version_added": "36"
+              "version_added": "28"
             },
             "opera_android": {
-              "version_added": "36"
+              "version_added": "28"
             },
             "safari": {
               "version_added": "9"
@@ -590,10 +590,10 @@
               "version_added": false
             },
             "opera": {
-              "version_added": "36"
+              "version_added": "28"
             },
             "opera_android": {
-              "version_added": "36"
+              "version_added": "28"
             },
             "safari": {
               "version_added": "9"
--- a/api/OfflineAudioContext.json
+++ b/api/OfflineAudioContext.json
@@ -134,10 +134,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "53"
+              "version_added": "25"
             },
             "firefox_android": {
-              "version_added": "53"
+              "version_added": "25"
             },
             "ie": {
               "version_added": false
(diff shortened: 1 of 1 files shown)


foolip (changes requested):
Can you mirror data for Opera?

api/OfflineAudioContext.json
```diff
@@ -134,10 +134,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "53"
+              "version_added": "25"
```
foolip, line 137: Yep, the previous data didn't make any sense this the constructor is the entry point to this API.

api/OfflineAudioContext.json
```diff
@@ -412,40 +412,40 @@
...
-              "version_added": "36"
+              "version_added": "28"
             },
             "opera_android": {
-              "version_added": "36"
+              "version_added": "28"
             },
             "safari": {
-              "version_added": "14.1"
+              "version_added": "9"
```
foolip, line 439: Can you check if the same change should be made for `startRendering`? That has claims of prefixed support right now which I'm pretty sure are wrong, and probably get in the way of the collector fixing the data.

queengooborg: Just checked the data for `startRendering`, looks like it was in need of updating as well!  Double-checked the prefix data too!

foolip: Do you mean that it really was called `webkitStartRendering` at one point? There are zero hits for that word on Google, can you describe the testing you did? I expect the interface but none of its members to have been prefixed.

queengooborg: That's correct -- I used the following code to test support with this prefix (in a console window open on mdn-bcd-collector):

```js
var instance = reusableInstances.offlineAudioContext;
'webkitStartRendering' in instance;
```

foolip: That just sounds too weird to be true, so I checked more.

`git log -S webkitStartRendering --follow -- Source/WebCore/Modules/webaudio/OfflineAudioContext.idl` doesn't find this in the history of WebKit, and testing Safari 8 on BrowserStack, I find that `webkitOfflineAudioContext.prototype.startRendering` is there and that `webkitOfflineAudioContext.prototype.webkitStartRendering` isn't. Both of those findings contradict the data.

Can you double-tripple check this? What is going on?

queengooborg: Alright, now I'm just as confused as you are on this.  Now when I try that code, I find that the prefixed version is undefined.  Hrm...

I'm just going to proceed by removing the prefix from the data and just stating this was available unprefixed since the API was available.

foolip (approved):

PR #14295, opened 2021-12-29
Update Chrome versions for iframe sandbox-allow-modals

felixh10r:
#### Summary
allow-modals support is currently listed as unknown, this PR will add known support for Chrome and related browsers.

https://chromestatus.com/feature/4747009953103872
https://googlechrome.github.io/samples/block-modal-dialogs-sandboxed-iframe/index.html


html/elements/iframe.json
```diff
@@ -912,10 +912,10 @@
...
               "chrome": {
-                "version_added": null
+                "version_added": "46"
               },
               "chrome_android": {
-                "version_added": null
+                "version_added": "46"
               },
               "edge": {
                 "version_added": null
```
ddbeck, line 921: This has probably reached some of the other Chrome-derived browsers.

```suggestion
                "version_added": "≤79"
```

And below this, outside of GitHub's suggestion range:

- Opera 33
- Opera for Android 33
- Samsung Internet 5.0

ddbeck (changes requested):
Thanks for this PR, @felixh10r. I see it's your first to the project—welcome! A few suggestions in the line comment. Thanks again!

html/elements/iframe.json
felixh10r: Thanks for the spot-on suggestions, I've added the respective versions.

ddbeck (approved):
Thank you! 🎉

Issue #3068, opened 2018-11-09
Need guidance for callback functions.


jpmedley:
Such as this: https://drafts.csswg.org/resize-observer-1/#resize-observer-callback

There are currently no callbacks listed in BCD and no guidance in the help on how to do them.

vinyldarkscratch:
Looking around, I found that BCD actually *does* have some callbacks within the data, for example [`RTCSessionDescriptionCallback`](https://github.com/mdn/browser-compat-data/blob/master/api/RTCSessionDescriptionCallback.json).  It appears we're essentially treating them like separate APIs.

vinyldarkscratch:
Closing this issue since it's stale.  Happy to reopen or create a new issue for work on the specific callback!

ddbeck:
I'm reopening this. Even if that's how we mean to deal with callbacks, we still don't have documentation that tells contributors how to add this data (it's not obvious to me, for example, how you would name a callback feature or where it should be structured in BCD).

vinyldarkscratch:
Ah, I see what you mean, good point!

vinyldarkscratch:
I know this is a pretty stale issue, but since I had last commented, we've removed all of the callbacks from BCD, as well as mdn/content, so we aren't documenting them anymore (especially since the interface attribute that utilizes it will cover its data).

I'm wondering if we should document what BCD won't include in the data guidelines?

jpmedley:
We have?! This particular callback is still supported in Chrome. (At least, it's still in the IDL.) We have 144 more.

jpmedley:
It seems like we shouldn't remove things from MDN or BCD ahead of browsers deprecating and removing them.

vinyldarkscratch:
There are many WebIDL features that don't make sense to document separately from the interface features that utilize them, including typedefs, enums, and callbacks. These are never directly exposed to web developers, and will never have their own compatibility separate from the interfaces.

queengooborg:
Considering we've removed all of the callbacks and are working to remove non-interface data from the API folder (#6810), should we close this issue @ddbeck?  Or should we leave it open as a reminder to update our documentation?

ddbeck:
@queengooborg We haven't quite removed all callbacks. But I've written up a draft guideline that ought to close this issue, if it lands. See https://github.com/mdn/browser-compat-data/issues/3068.

PR #14302, opened 2021-12-29
Add proposal for a no-callbacks data guideline

ddbeck:
#### Summary

Add a guideline to codify that we don't include non-exposed callback interfaces as standalone interfaces in BCD.

This copies heavily from the [mixin guideline](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#mixins).

There's one iffy thing here: the example provided is one we don't actually follow in practice. Maybe there's a better example?

#### Related issues

Fixes https://github.com/mdn/browser-compat-data/issues/3068.


queengooborg (approved):
LGTM, very descriptive!

foolip:
Web IDL callback interfaces can be observable to web developers, most easily if the interface has constants on it, as in the case of `NodeFilter`. Should https://developer.mozilla.org/en-US/docs/Web/API/NodeFilter really be removed?

Callback functions don't have names so I don't think a guideline to prevent them from being included in BCD is necessary. Or are there any?

ddbeck:
> Web IDL callback interfaces can be observable to web developers, most easily if the interface has constants on it, as in the case of `NodeFilter`. Should https://developer.mozilla.org/en-US/docs/Web/API/NodeFilter really be removed?

No, that's not intentional. I didn't know that there were any exposed callback interfaces (and, so far, I've only seen `NodeFilter`). Though it's still an open question whether we need a guideline for this, I've revised the text to exclude `[Exposed]` interfaces. I suppose another approach might be a more general guideline: no unexposed (non-exposed? not sure on terminology here) names in `api`. 

> Callback functions don't have names so I don't think a guideline to prevent them from being included in BCD is necessary. Or are there any?

Historically, there were some, though I don't know if every one has been purged. Resolving everything in https://github.com/mdn/browser-compat-data/issues/6810 might settle that question. In any case, we don't add them _now_ but haven't really said so explicitly.

queengooborg:
Doing a quick search through the [`@webref/idl`](https://www.npmjs.com/package/@webref/idl) package, it appears that there is a total of three callback interfaces: `EventListener`, `NodeFilter`, and `XPathNSResolver`.  Only the first two are in BCD, and only one of them, `NodeFilter`, is exposed.  I feel like it's also worth pointing out that the mdn-bcd-collector project doesn't generate tests for these callback interfaces, as the WebIDL parser ([`webidl2` by W3C](https://www.npmjs.com/package/webidl2)) doesn't consider them to be interfaces.

Personally, I'm voting for just removing all callback data.  I feel that their functionality is sufficiently represented by the [features they're used in](https://developer.mozilla.org/en-US/docs/Web/API/NodeIterator/filter), and otherwise [well documented on MDN web docs](https://developer.mozilla.org/en-US/docs/Web/API/NodeIterator/whatToShow).

Elchi3 (approved):
The guideline as is works for me.

I think I'm with @queengooborg on this one. So, I also think we could go further and just don't record callbacks compat data of any kind.

ddbeck:
@foolip did you have any more thoughts on this? Not sure if I overcame your objections or not.

ddbeck:
For `NodeFilter`, I think we can make a judgment that it's sufficiently dict-like to remove it, even though it's not _technically_ one. Ostensibly, this new guideline would address it, but if it quacks like a duck…

For `EventListener`, I think this guideline speaks directly to it. We have data for [the `EventListener` "interface"](https://github.com/mdn/browser-compat-data/blob/main/api/EventListener.json). But it's not exposed. If there's anything interesting to be reported about the `EventListener` callback, then it should be as a subfeature of `addEventListener`, not as `api.EventListener` alone.

That said, I'd be open to fixing those two cases and moving on without a guideline entirely, if those are the only two cases that this guideline would tend to.

queengooborg:
I opened two PRs to remove `NodeFilter` and `EventListener` in #15783 and #15785, taking care of the two remaining callbacks in our data!  I do still think we should add a guideline for callbacks, as there are many other callbacks we have removed in the past.

queengooborg:
Question for everyone: what do we need to do to move this PR forward?  I'm not quite certain what we'd need to do next, if anything.

foolip (approved):

PR #14320, opened 2021-12-29
Update all browsers versions for RadioNodeList API

queengooborg:
This PR updates and corrects the real values for all browsers for the `RadioNodeList` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v4.0.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/RadioNodeList

### DIFF
--- a/api/RadioNodeList.json
+++ b/api/RadioNodeList.json
@@ -6,40 +6,40 @@
         "spec_url": "https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#radionodelist",
         "support": {
           "chrome": {
-            "version_added": "34"
+            "version_added": "1"
           },
           "chrome_android": {
-            "version_added": "34"
+            "version_added": "18"
           },
           "edge": {
             "version_added": "12"
           },
           "firefox": {
-            "version_added": "33"
+            "version_added": "1.5"
           },
           "firefox_android": {
-            "version_added": "33"
+            "version_added": "4"
           },
           "ie": {
-            "version_added": "9"
+            "version_added": "8"
           },
           "opera": {
-            "version_added": "21"
+            "version_added": "≤12.1"
           },
           "opera_android": {
-            "version_added": "21"
+            "version_added": "≤12.1"
           },
           "safari": {
-            "version_added": "10"
+            "version_added": "≤4"
           },
           "safari_ios": {
-            "version_added": "9"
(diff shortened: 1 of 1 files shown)


api/RadioNodeList.json
```diff
@@ -6,40 +6,40 @@
         "spec_url": "https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#radionodelist",
         "support": {
           "chrome": {
-            "version_added": "34"
+            "version_added": "1"
```
foolip, line 9: Since this interface has `value` as the only member, I don't think it's meaningful to say that it's supported before the `value` property. The custom test for this will return true regardless of what kind of object `form.elements.namedItem('radio')` returns. In particular if it's a `NodeList` I don't think that should count as supported.

api/RadioNodeList.json
```diff
@@ -54,10 +54,10 @@
           "spec_url": "https://html.spec.whatwg.org/multipage/common-dom-interfaces.html#dom-radionodelist-value",
           "support": {
             "chrome": {
-              "version_added": "34"
+              "version_added": "21"
```
foolip, line 57: Looks like this is where this was all implemented:
https://trac.webkit.org/changeset/115009/webkit
https://trac.webkit.org/browser/webkit/trunk/Source/WebCore/Configurations/Version.xcconfig?rev=115009

WebKit 536.9 would suggest Chrome 20, but if tests say 21 I trust that.

Can you double check the Safari versions? I think they should both be 6.

foolip (changes requested):

api/RadioNodeList.json
queengooborg: It appears that this wasn't present in Safari 6, but is in Safari 8.  I haven't confirmed Safari 7 support (my VMs aren't working properly at the moment), but I'll set it to "7" based upon the timing of the commit.

foolip: 7 sounds good!

foolip:
Some fixes are still needed here. Can you copy the subfeature data to the parent feature and see what differences there are? See https://github.com/mdn/browser-compat-data/pull/14320#discussion_r777926744 for my thinking on this.

foolip (approved):

PR #14340, opened 2021-12-30
Update Chromium versions for api.ServiceWorkerContainer.onmessageerror

queengooborg:
This PR updates and corrects the real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `onmessageerror` member of the `ServiceWorkerContainer` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v4.0.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/ServiceWorkerContainer/onmessageerror

### DIFF
--- a/api/ServiceWorkerContainer.json
+++ b/api/ServiceWorkerContainer.json
@@ -406,14 +406,13 @@
           "spec_url": "https://w3c.github.io/ServiceWorker/#dom-serviceworkerglobalscope-onmessageerror",
           "support": {
             "chrome": {
-              "version_added": false
+              "version_added": "80"
             },
             "chrome_android": {
-              "version_added": false
+              "version_added": "80"
             },
             "edge": {
-              "version_added": "17",
-              "version_removed": "79"
+              "version_added": "17"
             },
             "firefox": {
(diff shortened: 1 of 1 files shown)


api/ServiceWorkerContainer.json
```diff
@@ -406,14 +406,13 @@
...
-              "version_added": false
+              "version_added": "80"
             },
             "chrome_android": {
-              "version_added": false
+              "version_added": "80"
             },
             "edge": {
-              "version_added": "17",
-              "version_removed": "79"
```
foolip, line 416: Was it really in 79, or actually missing in just that one version as the Chrome data would suggest?

foolip: I've run https://mdn-bcd-collector.appspot.com/tests/api/ServiceWorkerContainer/onmessageerror in Edge 16, 17, 18, 79 and 80 to confirm. It was added in 17, not in 79, and then in 80 again as one would expect. I'll update the data here.

foolip (approved):

PR #14353, opened 2021-12-30
Update WebView versions for USB API

queengooborg:
This PR updates and corrects the real values for WebView Android for the `USB` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v4.0.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/USB

### DIFF
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -3821,7 +3821,7 @@
             },
             "webview_android": {
               "version_added": false,
-              "notes": "WebView exposes this member, but does not support WebUSB. See <a href='https://crbug.com/933055'>Chromium bug 933055</a>."
+              "notes": "WebView exposes <code>navigator.usb</code>, but does not support WebUSB. See <a href='https://crbug.com/933055'>Chromium bug 933055</a>."
             }
           },
           "status": {
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -3820,7 +3820,8 @@
               "version_added": "8.0"
             },
             "webview_android": {
-              "version_added": "61"
+              "version_added": false,
+              "notes": "WebView exposes this member, but does not support WebUSB. See <a href='https://crbug.com/933055'>Chromium bug 933055</a>."
             }
           },
           "status": {
--- a/api/USB.json
+++ b/api/USB.json
@@ -88,8 +88,7 @@
               "version_added": "8.0"
             },
             "webview_android": {
-              "version_added": false,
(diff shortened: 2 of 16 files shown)


foolip (changes requested):
It seems like even though the interfaces are exposed, USB isn't supported in WebView. Can you link https://crbug.com/933055 in notes to avoid making this change again?

api/USB.json
```diff
@@ -87,7 +88,8 @@
               "version_added": "8.0"
             },
             "webview_android": {
-              "version_added": false
+              "version_added": false,
+              "notes": "WebView exposes this interface, but does not support WebUSB. See <a href='https://crbug.com/933055'>Chromium bug 933055</a>."
```
foolip, line 92: No need to add the note for every subfeature. It also becomes inaccurate, since `getDevices` isn't an interface.

foolip (changes requested):
Can you also check and update the "usb" entries in Navigator.json, WorkerNavigator.json and feature-policy.json? That data is probably mirrored and wrong.

api/WorkerNavigator.json
```diff
@@ -999,7 +999,8 @@
               "version_added": "10.0"
             },
             "webview_android": {
-              "version_added": "70"
+              "version_added": false,
+              "notes": "WebView exposes this interface, but does not support WebUSB. See <a href='https://crbug.com/933055'>Chromium bug 933055</a>."
```
foolip, line 1003:
```suggestion
              "notes": "WebView exposes <code>navigator.usb</code>, but does not support WebUSB. See <a href='https://crbug.com/933055'>Chromium bug 933055</a>."
```

api/Navigator.json
```diff
@@ -3820,7 +3820,8 @@
               "version_added": "8.0"
             },
             "webview_android": {
-              "version_added": "61"
+              "version_added": false,
+              "notes": "WebView exposes this member, but does not support WebUSB. See <a href='https://crbug.com/933055'>Chromium bug 933055</a>."
```
foolip, line 3824:
```suggestion
              "notes": "WebView exposes <code>navigator.usb</code>, but does not support WebUSB. See <a href='https://crbug.com/933055'>Chromium bug 933055</a>."
```

foolip (changes requested):
There was one incorrect "this interface" left, suggest changing both of them to just use the name of the API, since it's short.

foolip (approved):

PR #14374, opened 2021-12-31
Add a note of Blob Constructor in IE10

aleen42:
#### Summary

Add a note of Blob Constructor in IE10 to specify a special behaviour.

```js
new Blob([new Uint8Array(0)]); // => "InvalidStateError"
```

#### Workaround

```js
var bytes = new Uint8Array(0);

try {
    new Blob([bytes]); // => "InvalidStateError"
} catch (ignore) {
    var blobBuilder = new MSBlobBuilder();
    blobBuilder.append(bytes.buffer);
    blobBuilder.getBlob();
}
```

### DIFF
--- a/api/Blob.json
+++ b/api/Blob.json
@@ -86,7 +86,8 @@
               "notes": "Before Firefox 16, the second parameter, when set to <code>null</code> or <code>undefined</code>, leads to an error instead of being handled as an empty dictionary."
             },
             "ie": {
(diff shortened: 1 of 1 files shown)


estelle:
Instead of https://github.com/mdn/content/pull/11622/files

ddbeck (approved):
Looks good. Thank you! 🎉

PR #14408, opened 2022-01-03
Remove notes about innerHTML/outHTML prototype move

foolip:
This is a case of "When members are moved down the prototype chain" and
shouldn't have any notes, per our guideline:


foolip:
Converting to draft because after https://github.com/mdn/browser-compat-data/pull/14236#pullrequestreview-842861955 I suspect that there *should* be notes, but about `HTMLElement`, not `Node`.

foolip:
Closing in favor of https://github.com/mdn/browser-compat-data/pull/17226.

PR #14437, opened 2022-01-05
Mark transitioncancel event shipping in Chromium 74

foolip:
The data was already updated for HTMLElement, but the event bubbles and
so these other entries should also be updated.

Note that ontransitioncancel was added later, and that data is correct,
but will have to be merged according to this guideline:
https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#dom-events-eventname_event

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -10844,13 +10844,13 @@
           "spec_url": "https://drafts.csswg.org/css-transitions/#transitioncancel",
           "support": {
             "chrome": {
-              "version_added": false
+              "version_added": "74"
             },
             "chrome_android": {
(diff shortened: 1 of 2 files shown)


foolip:
@Elchi3 I've also tried fixing up other transition events and it's all really complicated. I'm thinking that maybe it's best to do all four events in a single PR and research it once. I think they should all be under `Element`, not `HTMLElement`. When you get to transition events in your event cleanup, can you ping me? The existing data seems pretty wrong in places.

sideshowbarker (approved):

PR #14445, opened 2022-01-05
Fix Chromium data for devicemotion event interfaces

foolip:
This was all added in WebKit trunk 534.5 and 534.9:
https://trac.webkit.org/changeset/64845/webkit
https://trac.webkit.org/changeset/68252/webkit

It wasn't enabled then, but because it all went together, it can all be
assumed to match the ondevicemotion property being exposed. That was
confirmed to be Chrome 31 with this test:
http://mdn-bcd-collector.appspot.com/tests/api/Window/ondevicemotion

The source of Chrome 11 is here:
https://github.com/mdn/browser-compat-data/pull/5834

The description says Chrome 31, so this must have been a copy-paste
error from the IE version.

Also update the Samsung Internet and WebView versions based on
conservative mirroring. It's possible this was supported earlier, but
the current version numbers also seem to be based on mirroring, not
testing:
https://github.com/mdn/browser-compat-data/pull/6838

### DIFF
--- a/api/DeviceMotionEvent.json
+++ b/api/DeviceMotionEvent.json
@@ -6,10 +6,10 @@
         "spec_url": "https://w3c.github.io/deviceorientation/#devicemotion",
         "support": {
           "chrome": {
-            "version_added": "11"
+            "version_added": "31"
           },
           "chrome_android": {
-            "version_added": "18"
+            "version_added": "31"
           },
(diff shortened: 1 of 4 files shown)


queengooborg (approved):
LGTM -- I also see some removal of redundant notes in here, which I'm all for!

PR #14448, opened 2022-01-05
Mark onsecuritypolicyviolation as supported in Chromium 97

foolip:
Source: https://storage.googleapis.com/chromium-find-releases-static/bec.html#bec147a63f30ec6e1157bc591c4cdb4998387d00

Also confirmed by logging `window.onsecuritypolicyviolation` in Chrome 97.


foolip:
@Elchi3 note that this is yet another case of the event and handler coming apart.

queengooborg (approved):

PR #14451, opened 2022-01-06
IE does not also support `visibility: unset`

aleen42:
#### Summary

In addition to `visibility: initial`, `visibility: unset` is also not supported under IE.

I have tested under IE11.

image

### DIFF
--- a/css/properties/visibility.json
+++ b/css/properties/visibility.json
@@ -25,6 +25,7 @@
               "version_added": "4",
               "notes": [
                 "Internet Explorer doesn't support <code>visibility: initial</code>.",
+                "Internet Explorer doesn't support <code>visibility: unset</code>.",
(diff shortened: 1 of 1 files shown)


css/properties/visibility.json
```diff
@@ -24,7 +24,7 @@
             "ie": {
               "version_added": "4",
               "notes": [
-                "Internet Explorer doesn't support <code>visibility: initial</code>.",
+                "Internet Explorer doesn't support <code>visibility: initial</code> and <code>visibility: unset</code>.",
```
ddbeck, line 27: Would you mind inserting this as a new note? I'd make this three total entries in the array, rather than two.

Technically, we ought to have separate subfeatures for each of the named values, but breaking up the note respects that fact a little better in the mean time.

ddbeck (changes requested):
One suggestion for you. Thank you!

Yaliang0802 (approved):

ddbeck (approved):
Looks good. Thank you! 🎉

PR #14453, opened 2022-01-07
ScreenOrientation.lock - partially supported from 69/79

### LINKED ISSUE #8524: api.ScreenOrientation.lock - doesn't work with Firefox Android 84.1.2
The lock() property of ScreenOrientation for Firefox Android doesn't seem to work, at least on old devices / OS.

The locking of the screen of the device in landscape-primary orientation (or another orientation) when ``ScreenOrientation.lock("chosen-orientation")`` is called in fullscreen mode. 

I've tested with emscripten using different methods (deprecated methods too) and with a simple script that doesn't seem to work too :  

```
function lock (orientation) {
  if (document.documentElement.requestFullscreen) {
    document.documentElement.requestFullscreen();
  } else if (document.documentElement.mozRequestFullScreen) {
    document.documentElement.mozRequestFullScreen();
  } else if (document.documentElement.webkitRequestFullscreen) {
    document.documentElement.webkitRequestFullscreen();
  } else if (document.documentElement.msRequestFullscreen) {
    document.documentElement.msRequestFullscreen();
  }

  screen.orientation.lock(orientation);
}

function unlock () {
  screen.orientation.unlock();

  if (document.exitFullscreen) {
    document.exitFullscreen();
  } else if (document.webkitExitFullscreen) {
    document.webkitExitFullscreen();
  } else if (document.mozCancelFullScreen) {
    document.mozCancelFullScreen();
  } else if (document.msExitFullscreen) {
    document.msExitFullscreen();
  }
}
```

**Hardware and OS version :**
- Samsung Galaxy Tab (SM-T533) running Android Lolipop (Android 5.1.1)
- Samsung Galaxy S4 (GT-I9505) running CyanogenMod 12.1 (Android 5.1.1)

**Firefox version :**
- Fifrefox 84.1.2
- Firefox nightly 210103 (build \#2015785225)

I don't have any problem with chrome based browsers.

<details>

* Query: `api.ScreenOrientation.lock`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/ScreenOrientation/lock
* Report started: 2021-01-04T01:20:07.765Z

</details>

hamishwillee:
Fixes #8524

As discussed in https://github.com/mdn/browser-compat-data/issues/8524#issuecomment-1004476882 [ScreenOrientation.lock()](https://developer.mozilla.org/en-US/docs/Web/API/ScreenOrientation/lock) was broken by Fenix rewrite or as part of the Fennec code removal project at around FFv69. As the first release of Android after 68 is FF 79 I have marked removed in that version, and partially supported afterwards - indicating the API exists but does not work. 

I have left desktop unchanged - it does not appear to behave differently in those versions. i.e. you can't set desktop orientation anyway, but this is expected.

This relates to docs work in https://github.com/mdn/content/issues/11594 . There is a fix for the bug but it requires a preference, and so is not recorded here.

### DIFF
--- a/api/ScreenOrientation.json
+++ b/api/ScreenOrientation.json
@@ -111,17 +111,9 @@
             "edge": {
               "version_added": "79"
             },
-            "firefox": [
-              {
-                "version_added": "69",
-                "partial_implementation": true,
-                "notes": "The API exists but does not lock the screen orientation."
-              },
-              {
-                "version_added": "43",
-                "version_removed": "69"
-              }
-            ],
+            "firefox": {
+              "version_added": "43"
+            },
             "firefox_android": [
               {
                 "version_added": "79",
--- a/api/ScreenOrientation.json
+++ b/api/ScreenOrientation.json
@@ -123,8 +123,9 @@
             "firefox_android": [
               {
                 "version_added": "79",
+                "version_removed": "97",
                 "partial_implementation": true,
-                "notes": "The API exists but does not lock the screen orientation."
+                "notes": "The API exists but returns <code>NS_ERROR_UNEXPECTED</code>."
               },
               {
                 "version_added": "43",
--- a/api/ScreenOrientation.json
+++ b/api/ScreenOrientation.json
@@ -111,12 +111,28 @@
             "edge": {
               "version_added": "79"
             },
-            "firefox": {
-              "version_added": "43"
-            },
-            "firefox_android": {
-              "version_added": "43"
-            },
+            "firefox": [
+              {
+                "version_added": "69",
+                "partial_implementation": true,
+                "notes": "The API exists but does not lock the screen orientation."
+              },
+              {
+                "version_added": "43",
+                "version_removed": "69"
+              }
+            ],
+            "firefox_android": [
+              {
+                "version_added": "79",
+                "partial_implementation": true,
+                "notes": "The API exists but does not lock the screen orientation."
+              },
+              {
+                "version_added": "43",
+                "version_removed": "79"
+              }
+            ],
             "ie": {
               "version_added": false
             },
--- a/api/ScreenOrientation.json
+++ b/api/ScreenOrientation.json
@@ -103,16 +103,22 @@
           "spec_url": "https://w3c.github.io/screen-orientation/#dom-screenorientation-lock",
           "support": {
             "chrome": {
-              "version_added": "38"
+              "version_added": "38",
+              "partial_implementation": true,
+              "notes": "Always throws <code>NotSupportedError</code>."
             },
             "chrome_android": {
               "version_added": "38"
             },
             "edge": {
-              "version_added": "79"
+              "version_added": "79",
+              "partial_implementation": true,
+              "notes": "Always throws <code>NotSupportedError</code>."
             },
             "firefox": {
-              "version_added": "43"
+              "version_added": "43",
+              "partial_implementation": true,
+              "notes": "Always throws <code>NotSupportedError</code>."
             },
             "firefox_android": [
               {
@@ -129,7 +135,9 @@
               "version_added": false
             },
             "opera": {
-              "version_added": "25"
+              "version_added": "25",
+              "partial_implementation": true,
+              "notes": "Always throws <code>NotSupportedError</code>."
             },
             "opera_android": {
               "version_added": "25"


makotokato:
Although I am sending PR for related same issue (https://github.com/mdn/browser-compat-data/pull/13841), until Firefox 68 for Android (Fennec), screen orientation API is supported and it will work. But Fenix didn't (due to e10s support etc)

And even if Firefox desktop (from 43), there is this API (`lock`) is implemented, but it always throws `NotSupportedError`.

hamishwillee:
@makotokato Thanks for that confirmation.

@ddbeck The lock() API is implemented on desktop, in that the method exists, but not "enabled" (it will always throw `NotSupportedError`). The spec explicitly states that a browser might choose to work in this way. So technically BCD is correct about support on desktop.  I find it a bit confusing but I am not sure the right way to represent this in BCD.

The API actually is actually even worse than this. The spec allows for a browser to have its own rules about the state in which it will allow locking (e.g. FF and Chrome both require fullScreenLock on the document context) so it is possible that the same code will not work on two browsers because they have different conditions. Even worse, there is no way to query if lock is supported on a platform (like desktop) or what the conditions for lock are, or if the conditions for lock are met.

api/ScreenOrientation.json
```diff
@@ -114,9 +114,17 @@
...
-            },
+            "firefox_android": [
+              {
+                "version_added": "79",
+                "partial_implementation": true,
+                "notes": "The API exists but does not lock the screen orientation."
+              },
+              {
+                "version_added": "43",
+                "version_removed": "79"
```
hamishwillee, line 125: Note, Fennec last version was FF68. The next version is FF79, which uses Fenix and does not support the API on android. So FF79 is the version in which support was removed.

hamishwillee:
@ddbeck Thanks for the advice - updated.

Fixes #13841

api/ScreenOrientation.json
```diff
@@ -103,25 +103,41 @@
...
-            },
+              "version_added": "43",
+              "partial_implementation": true,
+              "notes": "Always throws <code>NotSupportedError</code>."
+            },
+            "firefox_android": [
+              {
+                "version_added": "79",
+                "partial_implementation": true,
+                "notes": "The API exists but does not lock the screen orientation."
```
ddbeck, line 127: Is this a different behavior than throwing `NotSupportedError`?

makotokato: `NotSupportedError` is thrown on Firefox Android 97+.  Android 79-96 will throw `NS_ERROR_UNEXPETED`

hamishwillee: > Is this a different behavior than throwing NotSupportedError?

Sorry for the late response - just back from holiday.

@ddbeck What I was saying here is that in my testing of FF **android** prior to FF96 I could call the API but the orientation would simply not lock - no error or anything. This differs from the desktop version which throws `NotSupportedError` to indicate that locking is not supported.

@makotokato Are you saying that my testing was incorrect - there is a thrown error in **FF Android 79-96** of `NS_ERROR_UNEXPECTED` if you tried to use the API on android?

@ddbeck If so, in that case wouldn't we just assume that this is the default case for anything not supported (i.e. "equivalent" to `NotSupportedError` and simply remove this support statement. Then it would look like this was supported from 43 and removed in 79, which is in fact the way users will see it.

Otherwise I can put this error in.

> `NotSupportedError` is thrown on Firefox Android 97+.

Presumably only if the preference is not enabled.

makotokato: > Are you saying that my testing was incorrect - there is a thrown error in FF Android 79-96 of NS_ERROR_UNEXPECTED if you tried to use the API on android?

I tested this by https://makotokato.github.io/demo/orientation.html. Firefox 96 throws `NS_ERROR_UNEXPECTED`. Firefox (beta) 97+ throws `NotSupportedError` if pref isn't set.

hamishwillee: Thanks @makotokato. So @ddbeck Above, question is whether we try to capture that we get `NS_ERROR_UNEXPECTED` or we just pretend that support was removed from FF79 (ie mark that as version things were removed).

ddbeck: Unfortunately, convention requires that we represent this API as partially supported. Setting a `version_removed` can confuse things a bit, since a simple feature detection will show that the feature exists (i.e., `"lock" in ScreenOrientation.prototype` returns `true`) though it doesn't actually _do_ anything.

ddbeck (changes requested):

api/ScreenOrientation.json
hamishwillee: @ddbeck I don't understand the action you are asking me to take here ^^^ with `lock`. 

This is how I see it - but I could be wrong - I've lost track over time and I'm not sure I saw the preference first time around. @makotokato would appreciate your confirmation.

Firefox **desktop** looks as below
- The feature has existed since 43 but always throws `NotSupportedError` on desktop. 
- There is a pref (`dom.screenorientation.allow-lock`) which was introduced at v97 (I think), but this is _irrelevant_ because whether true or false you get `NotSupportedError` on desktop

```json
            "firefox": {
              "version_added": "43",
              "notes": "Always throws <code>NotSupportedError</code>."
            },
```

Firefox android
-  has existed since FF43 and worked in the early versions..
- In v79 (first version on new UI framework) it was broken. From FF79 to FF96 it returns NS_ERROR_UNEXPECTED 
- FF97 the API was fixed if `dom.screenorientation.allow-lock` is true. If not set, it returns `NotSupportedError`.
- But as per MDN standard, the pref is considered not to exist on Android.

This is what the definition looks like right now. 
```json
            "firefox_android": [
              {
                "version_added": "79",
                "notes": "The API exists but does not lock the screen orientation."
              },
              {
                "version_added": "43",
                "version_removed": "79"
              }
            ],
```
I think this might be wrong. It is supposed to indicate that this API works from 43 to 79. Then from 79 it should indicate that it exists but doesn't work correctly. Then in FF97 it should appear removed, because the preference cannot be set. Is this what you meant?

```json
            "firefox_android": [
              {
                "version_added": "79",
                "version_removed": "97",
                "notes": "The API exists but returns <code>NS_ERROR_UNEXPECTED</code> ."
              },
              {
                "version_added": "43",
                "version_removed": "79"
              }
            ],
```

ddbeck: Yes, this last code block would be correct. Sorry for misunderstanding before.

hamishwillee: No worries. I don't think I'd got to the point where this could be put succinctly before :-)

hamishwillee:
@ddbeck Updated as discussed. Should be good to merge!

ddbeck (approved):
Looks good. Thank you! 🎉

PR #14455, opened 2022-01-07
Update the compatible version of `dblclick` events under IE

aleen42:
#### Summary

I have tested `dblclick` event with the following snippet is supported down to IE8, so update the compatible version.

```js
document.body.addEventListener('dblclick', function () { console.log('double clicked') });

// IE8
document.body.attachEvent('ondblclick', function () { console.log('double clicked') });
```

#### Test Snapshots

- IE10:

image

- IE9: 

image

- IE8:

image

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -2429,7 +2429,7 @@
               "version_added": "6"
             },
             "ie": {
-              "version_added": "11"
+              "version_added": "8"
             },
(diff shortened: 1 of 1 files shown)


queengooborg (approved):
Looks good, I've confirmed a lack of support in IE7!

PR #14456, opened 2022-01-07
Add edge 99

beaufortfrancois:
This PR adds the version 99 of the Edge browser.

@foolip Please review.


foolip (approved):
I can confirm that my Edge Canary is already at version 99.

PR #14461, opened 2022-01-07
Put the <dialog> cancel event behind a flag in Firefox

foolip:
The version number (78) came from here:
https://github.com/mdn/browser-compat-data/pull/6433

But <dialog> itself isn't shipped yet, so fix up the data to match other
subfeatures.


foolip:
I noticed this because it looks like a feature supported by Chrome+Firefox but not Safari, but that wasn't the case.

queengooborg (approved):

PR #4645, opened 2019-08-19
Add Safari versions for Gamepad API

vinyldarkscratch:
This PR fixes #4589 by adding Safari versions for the Gamepad API.

PR #14465, opened 2022-01-07
Mark Gamepad's id property as supported in Safari

foolip:
This was missed in https://github.com/mdn/browser-compat-data/pull/4645 and it's not clear why.

The id attribute has been in the IDL/C++ in WebKit for very long:
https://trac.webkit.org/changeset/170309/webkit
https://trac.webkit.org/changeset/170465/webkit

### DIFF
--- a/api/Gamepad.json
+++ b/api/Gamepad.json
@@ -410,10 +410,10 @@
               "version_added": "14"
             },
             "safari": {
-              "version_added": false
+              "version_added": "10.1"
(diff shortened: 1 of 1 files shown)


queengooborg (approved):
Huh, not sure why this one was missed in the original PR -- I'm guessing that I was only testing the prototype and this property wasn't present on it at that time.  Thanks for the fix!

PR #14481, opened 2022-01-09
Make HTML element "name" property statuses consistent

### LINKED ISSUE #6683: api.Element.name entry doesn't make sense as used on MDN
https://developer.mozilla.org/en-US/docs/Web/API/Element/name says:

> name gets or sets the name property of an element in the DOM. It only applies to the following elements: &lt;a>, &lt;applet>, &lt;button>, &lt;form>, &lt;frame>, &lt;iframe>, &lt;img>, &lt;input>, &lt;map>, &lt;meta>, &lt;object>, &lt;param>, &lt;select>, and &lt;textarea>.
>
> Note: The name property doesn't exist for other elements; unlike tagName and nodeName, it is not a property of the Node, Element or HTMLElement interfaces.

https://developer.mozilla.org/en-US/docs/Web/API/Element/name#Browser_compatibility however is not a useful table for this page, because BCD correctly reflects that there's no `name` attribute on `Element` in any browser.

@ddbeck how should one handle a case like this? Should the MDN page instead link to the tables for the elements that do support it?

queengooborg:
This PR makes the statuses for the `name` property of the HTML element APIs consistent.  All of these entries are listed in the current HTML spec, so they are standard and non-deprecated.  Fixes #6683.

### DIFF
--- a/api/HTMLAnchorElement.json
+++ b/api/HTMLAnchorElement.json
@@ -285,7 +285,7 @@
           "status": {
             "experimental": false,
             "standard_track": true,
-            "deprecated": false
+            "deprecated": true
           }
         }
       },
--- a/api/HTMLAnchorElement.json
+++ b/api/HTMLAnchorElement.json
@@ -284,8 +284,8 @@
           },
           "status": {
             "experimental": false,
-            "standard_track": false,
-            "deprecated": true
+            "standard_track": true,
+            "deprecated": false
           }
         }
       },
--- a/api/HTMLEmbedElement.json
+++ b/api/HTMLEmbedElement.json
@@ -238,7 +238,7 @@
           "status": {
             "experimental": false,
             "standard_track": true,
-            "deprecated": false
+            "deprecated": true
           }
         }
       },
--- a/api/HTMLEmbedElement.json
+++ b/api/HTMLEmbedElement.json
@@ -238,7 +238,7 @@
           "status": {
             "experimental": false,
             "standard_track": true,
-            "deprecated": true
+            "deprecated": false
           }
         }
       },
--- a/api/HTMLFrameElement.json
+++ b/api/HTMLFrameElement.json
@@ -379,7 +379,7 @@
           "status": {
             "experimental": false,
             "standard_track": true,
-            "deprecated": false
+            "deprecated": true
           }
         }
       },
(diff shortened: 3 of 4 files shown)


api/HTMLAnchorElement.json
```diff
@@ -284,8 +284,8 @@
           },
           "status": {
             "experimental": false,
-            "standard_track": false,
-            "deprecated": true
+            "standard_track": true,
+            "deprecated": false
```
ddbeck, line 288: The IDL for this has the `// also has obsolete members` comment link. It's both standard track and deprecated.

https://html.spec.whatwg.org/multipage/text-level-semantics.html#htmlanchorelement

```suggestion
            "deprecated": true
```

api/HTMLFrameElement.json
```diff
@@ -379,7 +379,7 @@
           "status": {
             "experimental": false,
             "standard_track": true,
-            "deprecated": true
+            "deprecated": false
```
ddbeck, line 382: The entirety of the `<frame>` element and its API is deprecated, including this property. Don't bring it back 🙅‍♂️

```suggestion
            "deprecated": true
```

api/HTMLImageElement.json
```diff
@@ -831,7 +831,7 @@
           "status": {
             "experimental": false,
             "standard_track": true,
-            "deprecated": true
+            "deprecated": false
```
ddbeck, line 834: This is specified as non-conforming: https://html.spec.whatwg.org/multipage/obsolete.html#attr-img-name

```suggestion
            "deprecated": true
```

api/HTMLEmbedElement.json
```diff
@@ -238,7 +238,7 @@
           "status": {
             "experimental": false,
             "standard_track": true,
-            "deprecated": true
+            "deprecated": false
```
ddbeck, line 241: The IDL for this has the `// also has obsolete members` comment link too.

```suggestion
            "deprecated": true
```

ddbeck (changes requested):
All ~~but one~~ of these are correctly marked as deprecated. One should be standard track but isn't. See line comments for details.

ddbeck (approved):
Looks good. Thank you! 🎉

PR #14484, opened 2022-01-09
Add compat data for :current css pseudo class

dragozin:
There is an error `No specification found` in the [Specifications section](https://developer.mozilla.org/en-US/docs/Web/CSS/:current#specifications)

### DIFF
--- a/css/selectors/current.json
+++ b/css/selectors/current.json
@@ -1,0 +1,56 @@
+{
+  "css": {
+    "selectors": {
+      "current": {
+        "__compat": {
+          "description": "<code>:current</code>",
+          "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/:current",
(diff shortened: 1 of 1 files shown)


ddbeck (changes requested):
Hi @dragozin and thanks for this PR. I see it's your first to the project—welcome!

There's a key question here: have you tested to see if this is supported in any browser (even behind a flag or in a development version)? If it is, then the data can reflect that and we can merge it. If not, then the feature isn't mature enough for inclusion on MDN yet (and the corresponding page would need to be deleted).

If you can shed any light on the status of this feature, that would help us decide what to do next. Thank you again!

ddbeck:
I couldn't find any evidence that this feature is supported. I've decided to drop the content, rather than add to BCD in this case. See https://github.com/mdn/content/pull/14931. In any case, thank you, @dragozin!

PR #14486, opened 2022-01-10
FF97 supports cap and ic units for CSS lengths

hamishwillee:
Support was added in https://bugzilla.mozilla.org/show_bug.cgi?id=1531223 (ic) and https://bugzilla.mozilla.org/show_bug.cgi?id=1702924 (cap)

Associated docs work can be tracked in: https://github.com/mdn/content/issues/11592

### DIFF
--- a/css/types/length-percentage.json
+++ b/css/types/length-percentage.json
@@ -112,10 +112,10 @@
                 "version_added": false
               },
               "firefox": {
-                "version_added": false
+                "version_added": "97"
               },
               "firefox_android": {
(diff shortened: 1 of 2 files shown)


ddbeck (changes requested):
One minor thing here: can you toggle these to `experimental: true`? So far, they're single implementation.

hamishwillee:
By your command :-) Done!

ddbeck (approved):
Looks good. Thank you! 🎉

PR #14488, opened 2022-01-10
CSS hyphenate-character property - add

hamishwillee:
FF97 adds support for the CSS `hyphenate-character` property https://bugzilla.mozilla.org/show_bug.cgi?id=1746187 behind a preference. I've added that for FF97, but not for FF97 on android.

Now Chrome supports this with a -webkit extension, so I added that as well. BUT I am not sure exactly what versions to put. 
I found it works fine to specify a value (`=`) say for back to Chrome 45, but in Chrome 44 the CSS doesn't seem to understand the "=" char - so the rendered character appears as a box. In other words it isn't clear if it doesn't work or if = just doesn't work.

image

Anyway, Safari has the same problem on current versions as the very old chrome, so I put that as false.

I just extended the other things to match chrome versions if appropriate. 

Thoughts on next steps?

Rest of work for this can be tracked in  https://github.com/mdn/content/issues/11596

@ddbeck Would appreciate your advice on this one.

### DIFF
--- a/css/properties/hyphenate-character.json
+++ b/css/properties/hyphenate-character.json
@@ -8,11 +8,11 @@
           "support": {
             "chrome": {
               "prefix": "-webkit-",
-              "version_added": "37"
+              "version_added": "6"
             },
             "chrome_android": {
               "prefix": "-webkit-",
-              "version_added": "37"
+              "version_added": "18"
             },
             "edge": {
               "prefix": "-webkit-",
@@ -35,26 +35,28 @@
               "version_added": false
             },
             "opera": {
-              "version_added": "24",
-              "prefix": "-webkit-"
+              "prefix": "-webkit-",
+              "version_added": "15"
             },
             "opera_android": {
               "prefix": "-webkit-",
-              "version_added": "24"
+              "version_added": "14"
             },
             "safari": {
-              "version_added": false
+              "prefix": "-webkit-",
+              "version_added": "5.1"
             },
             "safari_ios": {
-              "version_added": false
+              "prefix": "-webkit-",
+              "version_added": "5"
             },
             "samsunginternet_android": {
               "prefix": "-webkit-",
-              "version_added": "3.0"
+              "version_added": "1.0"
             },
             "webview_android": {
               "prefix": "-webkit-",
-              "version_added": "37"
+              "version_added": "≤37"
             }
           },
           "status": {
--- a/css/properties/hyphenate-character.json
+++ b/css/properties/hyphenate-character.json
@@ -8,11 +8,11 @@
           "support": {
             "chrome": {
               "prefix": "-webkit-",
-              "version_added": "45"
+              "version_added": "37"
             },
             "chrome_android": {
               "prefix": "-webkit-",
-              "version_added": "45"
+              "version_added": "37"
             },
             "edge": {
               "prefix": "-webkit-",
@@ -35,12 +35,12 @@
               "version_added": false
             },
             "opera": {
-              "version_added": "44",
+              "version_added": "24",
               "prefix": "-webkit-"
             },
             "opera_android": {
               "prefix": "-webkit-",
-              "version_added": "32"
+              "version_added": "24"
             },
             "safari": {
               "version_added": false
@@ -50,11 +50,11 @@
             },
             "samsunginternet_android": {
               "prefix": "-webkit-",
-              "version_added": "5.0"
+              "version_added": "3.0"
             },
             "webview_android": {
               "prefix": "-webkit-",
-              "version_added": "45"
+              "version_added": "37"
             }
           },
           "status": {
--- a/css/properties/hyphenate-character.json
+++ b/css/properties/hyphenate-character.json
@@ -36,8 +36,7 @@
             },
             "opera": {
               "version_added": "44",
-              "partial_implementation": true,
-              "notes": "<code>auto</code> value is only supported on macOS and Android."
+              "prefix": "-webkit-"
             },
             "opera_android": {
               "prefix": "-webkit-",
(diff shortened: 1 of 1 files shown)


css/properties/hyphenate-character.json
```diff
@@ -0,0 +1,69 @@
...
+              "prefix": "-webkit-",
+              "version_added": "5.0"
+            },
+            "webview_android": {
+              "prefix": "-webkit-",
+              "version_added": "45"
+            }
+          },
+          "status": {
+            "experimental": false,
```
hamishwillee, line 61: False because of the webkit prefix thingy

ddbeck:
Is the character oddity from the example in https://github.com/mdn/content/pull/11875? I think it might be an issue with the character, not the CSS. See https://github.com/mdn/content/pull/11875/files#r782305961

hamishwillee:
@ddbeck Yes it is. With the * instead of = char it works back to Chrome v37, which is as far as browserstack will let me go back. I've put v37 and associated versions in place, but not sure what to do now if we can't test earlier?

I won't get to this again until back from holiday.

css/properties/hyphenate-character.json
```diff
@@ -0,0 +1,69 @@
...
+  "css": {
+    "properties": {
+      "hyphenate-character": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/hyphenate-character",
+          "spec_url": "https://drafts.csswg.org/css-text-4/#propdef-hyphenate-character",
+          "support": {
+            "chrome": {
+              "prefix": "-webkit-",
+              "version_added": "37"
```
ddbeck, line 11: Finally, a chance to exercise my ancient browser research skills. Here's how I did it:

1. If you switch operating systems, you can go back to Chrome 15 or so on BrowserStack. It works there, which gets us in the range of Chromium releases which used WebKit (before Blink emerged). Range narrowed to Chrome 1-15.
2. I tested it in older Safaris to rough things out a little more. It shows up in Safari 5.1 (WebKit ~534.48). This puts us in the range of Chrome 1-12, based on WebKit version mapping (i.e., it wouldn't have shown up any later than the Chrome that ships WebKit 534.48 or greater).
3. I searched [webkit/webkit](https://github.com/webkit/webkit) to find the changset that added it: https://trac.webkit.org/changeset/61548/webkit. This dates the change to circa June 2010, narrowing things to Chrome 6-12.
4. A comment in [the associated WebKit bug](https://bugs.webkit.org/show_bug.cgi?id=10228) mentions a Chromium performance regression. This suggests that Chrome was updating WebKit versions pretty promptly. Based on the date, I think we can safely assume Chrome 6. But just to be sure, the we can compare [WebKit tags](https://trac.webkit.org/browser/webkit/tags) to the revision number, which places this around WebKit 534.1, which also aligns with Chrome 6.

```suggestion
              "version_added": "6"
```

Updating the rest of the values is an exercise in version mapping.

css/properties/hyphenate-character.json
```diff
@@ -0,0 +1,69 @@
...
+            },
+            "chrome_android": {
+              "prefix": "-webkit-",
+              "version_added": "37"
```
ddbeck, line 15:
```suggestion
              "version_added": "18"
```

css/properties/hyphenate-character.json
```diff
@@ -0,0 +1,69 @@
...
+            },
+            "firefox_android": {
+              "version_added": false
+            },
+            "ie": {
+              "version_added": false
+            },
+            "opera": {
+              "version_added": "24",
+              "prefix": "-webkit-"
```
ddbeck, line 39:
```suggestion
              "prefix": "-webkit-",
              "version_added": "15"
```

css/properties/hyphenate-character.json
```diff
@@ -0,0 +1,69 @@
...
+            },
+            "opera_android": {
+              "prefix": "-webkit-",
+              "version_added": "24"
```
ddbeck, line 43:
```suggestion
              "version_added": "14"
```

css/properties/hyphenate-character.json
```diff
@@ -0,0 +1,69 @@
...
+            },
+            "safari": {
+              "version_added": false
```
ddbeck, line 46:
```suggestion
              "prefix": "-webkit-",
              "version_added": "5.1"
```

css/properties/hyphenate-character.json
```diff
@@ -0,0 +1,69 @@
...
+            },
+            "safari_ios": {
+              "version_added": false
```
ddbeck, line 49:
```suggestion
              "prefix": "-webkit-",
              "version_added": "5"
```

css/properties/hyphenate-character.json
```diff
@@ -0,0 +1,69 @@
...
+            },
+            "samsunginternet_android": {
+              "prefix": "-webkit-",
+              "version_added": "3.0"
```
ddbeck, line 53:
```suggestion
              "version_added": "1.0"
```

css/properties/hyphenate-character.json
```diff
@@ -0,0 +1,69 @@
...
+            },
+            "webview_android": {
+              "prefix": "-webkit-",
+              "version_added": "37"
```
ddbeck, line 57:
```suggestion
              "version_added": "≤37"
```

ddbeck (changes requested):

hamishwillee:
Thanks @ddbeck - useful research tips. I'm going to hope the versions are always in current browserstack test versions :-)

I accepted all your changes.

ddbeck (approved):
Looks good. Thank you! 🎉

PR #14492, opened 2022-01-10
Split first baseline and last baseline entries

foolip:
The 3 properties `align-content`, `align-items` and `align-self` all
take the 3 values `baseline`, `first baseline` and `last baseline`.
However, support for `last baseline` doesn't match support for
`first baseline`, so the entries need to be split to represent this.

There was no entry for `align-items: baseline`, so add that.

Tests:
http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=9948
http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=9955

The following was found based on testing:

- Chrome 21 got support for `-webkit-align-items: baseline` and
  `-webkit-align-self: baseline`.
- Chrome 29 got support for the unprefixed `align-items: baseline` and
  `align-self: baseline`.
- Chrome 57 got support for `align-content: baseline`.
- Chrome 59 got support for `first baseline` and `last baseline` for all
  3 properties.
- Chrome 86 dropped support for `last baseline` for all 3 properties,
  because it was only recognized but not actually supported. See:
  https://storage.googleapis.com/chromium-find-releases-static/af5.html#af5e59b01391ffa6be2a37d7958df22821a4a18d

- Edge 18 support all 9 combinations of properties and values.

- Firefox 28 had support for `align-items: baseline` and
  `align-self: baseline`. They were supported in Firefox 27 too, but the
  parent feature here is set to 28 so clamp to that.
- Firefox 45 got support for `align-content: baseline`.
- Firefox 52 got support for `first baseline` and `last baseline` for
  all 3 properties.

- Safari 7.1 had support for `-webkit-align-items: baseline` and
  `-webkit-align-self: baseline`, but Safari 6 did not.
- Safari 9.1 had support for `-webkit-align-content: baseline`, but
  Safari 8 did not.
- Safari 11 got support for `first baseline` and `last baseline` for all
  3 properties.

In order to avoid researching when Edge first got support and whether it
was partial, use ≤79 ranges to represent that it could be supported
earlier.

### DIFF
--- a/css/properties/align-content.json
+++ b/css/properties/align-content.json
@@ -119,7 +119,7 @@
                   "version_added": "57"
                 },
                 "edge": {
-                  "version_added": "79"
+                  "version_added": "≤79"
                 },
                 "firefox": {
                   "version_added": "45"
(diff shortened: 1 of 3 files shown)


queengooborg (approved):
This is LGTM, thank you for digging into this!

PR #14496, opened 2022-01-10
Change Firefox version number supporting scrollbar-gutter

aethanyc:
scrollbar-gutter is shipped in Firefox 97 in
https://bugzilla.mozilla.org/show_bug.cgi?id=1715112

### DIFF
--- a/css/properties/scrollbar-gutter.json
+++ b/css/properties/scrollbar-gutter.json
@@ -40,12 +40,10 @@
               "version_added": "94"
             },
             "firefox": {
-              "version_added": false,
-              "notes": "See <a href='https://bugzil.la/1715112'>bug 1715112</a>."
+              "version_added": "97"
             },
             "firefox_android": {
-              "version_added": false,
-              "notes": "See <a href='https://bugzil.la/1715112'>bug 1715112</a>."
+              "version_added": "97"
             },
             "ie": {
               "version_added": false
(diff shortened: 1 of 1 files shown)


css/properties/scrollbar-gutter.json
```diff
@@ -40,11 +40,11 @@
               "version_added": "94"
             },
             "firefox": {
-              "version_added": false,
+              "version_added": "97",
               "notes": "See <a href='https://bugzil.la/1715112'>bug 1715112</a>."
```
ddbeck, line 44:
```suggestion
              "version_added": "97"
```

css/properties/scrollbar-gutter.json
```diff
@@ -40,11 +40,11 @@
...
             },
             "firefox_android": {
-              "version_added": false,
+              "version_added": "97",
               "notes": "See <a href='https://bugzil.la/1715112'>bug 1715112</a>."
```
ddbeck, line 48:
```suggestion
              "version_added": "97"
```

ddbeck (changes requested):
Thanks for this PR, @aethanyc! A couple of suggestions for you. First, we ought to remove the now uninteresting bug link. Also, since there are two implementations, I suggest that you set `experimental: false`. Thanks again!

aethanyc:
@ddbeck Thank you for the review. I updated the commit to address your comments.

ddbeck (approved):
Looks good. Thank you! 🎉

PR #14498, opened 2022-01-11
FF97 CSS basic-shape path() function enabled by default

hamishwillee:
FF97 ships the bug [Ship d property and make SVG path d attribute as the presentation attribute](https://bugzilla.mozilla.org/show_bug.cgi?id=1744599) which was implemented behind pref in FF91 https://bugzilla.mozilla.org/show_bug.cgi?id=1340422

My understanding is that essentially this means you can now call `path()` on the `d` property of an SVG path element in CSS. There is a cool little demo here: https://codepen.io/airen/pen/wgNEQo

So what I have done here is for `path()` added a NOTE about the introduction of the feature in FF91 and a new version for FF97 that removes the preference. For FF android the preference does not work from FF79 so I just put that as the end version.

> **NOTE:** I am not 100% certain about the partial implementation, but I will query that separately. This could go in as "no worse". 

> NOTE: Chrome also supports this in CSS now too, so that would have to be tested.

--- 
As a separate PR to follow ...

In addition, I THINK it would be useful to update the [`d` attribute](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/d) to indicate that from FF97 it is now a "presentation attribute", which is what allows it to be modified in CSS and set to the [`path()`](https://developer.mozilla.org/en-US/docs/Web/CSS/path()) or `none`. Not sure right way to do this - subfeature "supports_css_path()" or similar. Or would we do this as a "reversed" one - e.g. not_supports_css_path. Note that current chrome does do the right thing.  @ddbeck Advice?

Further, any reason not to add these spec URLs that are currently manually in the page here https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/d#specifications ?

### DIFF
--- a/css/properties/clip-path.json
+++ b/css/properties/clip-path.json
@@ -326,7 +326,7 @@
               ],
               "firefox_android": [
                 {
-                  "version_added": "71"
+                  "version_added": "79"
                 },
                 {
                   "version_added": "64",
--- a/css/properties/clip-path.json
+++ b/css/properties/clip-path.json
@@ -315,6 +315,7 @@
                 },
                 {
                   "version_added": "64",
+                  "version_removed": "71",
                   "flags": [
                     {
                       "type": "preference",
@@ -330,6 +331,7 @@
                 },
                 {
                   "version_added": "64",
+                  "version_removed": "79",
                   "flags": [
                     {
                       "type": "preference",
--- a/css/properties/clip-path.json
+++ b/css/properties/clip-path.json
@@ -314,7 +314,7 @@
                   "version_added": "71"
                 },
                 {
-                  "version_added": "63",
+                  "version_added": "64",
                   "flags": [
                     {
                       "type": "preference",
@@ -324,16 +324,21 @@
                   ]
                 }
               ],
-              "firefox_android": {
-                "version_added": "63",
-                "flags": [
-                  {
-                    "type": "preference",
-                    "name": "layout.css.clip-path-path.enabled",
-                    "value_to_set": "true"
-                  }
-                ]
-              },
+              "firefox_android": [
+                {
+                  "version_added": "71"
+                },
+                {
+                  "version_added": "64",
+                  "flags": [
+                    {
+                      "type": "preference",
+                      "name": "layout.css.clip-path-path.enabled",
+                      "value_to_set": "true"
+                    }
+                  ]
+                }
+              ],
               "ie": {
                 "version_added": false
               },
--- a/css/types/basic-shape.json
+++ b/css/types/basic-shape.json
@@ -270,46 +270,24 @@
                 "partial_implementation": true,
                 "notes": "Only supported on the <code>offset-path</code> property."
               },
-              "firefox": [
-                {
-                  "version_added": "97",
-                  "partial_implementation": true,
-                  "notes": "It is supported on the <code>d</code> SVG presentation attribute, <code>clip-path</code> and <code>offset-path</code> properties."
-                },
-                {
-                  "version_added": "63",
-                  "version_removed": "97",
-                  "partial_implementation": true,
-                  "notes": "From version 91 it can be used on the <code>d</code> SVG presentation attribute. For earlier versions it is only supported on the <code>clip-path</code> and <code>offset-path</code> properties.",
-                  "flags": [
-                    {
-                      "type": "preference",
-                      "name": "layout.css.clip-path-path.enabled",
-                      "value_to_set": "true"
-                    }
-                  ]
-                }
-              ],
-              "firefox_android": [
-                {
-                  "version_added": "97",
-                  "partial_implementation": true,
-                  "notes": "It is supported on the <code>d</code> SVG presentation attribute, <code>clip-path</code> and <code>offset-path</code> properties."
-                },
-                {
-                  "version_added": "63",
-                  "version_removed": "79",
-                  "partial_implementation": true,
-                  "notes": "Only supported on the <code>clip-path</code> and <code>offset-path</code> properties.",
-                  "flags": [
-                    {
-                      "type": "preference",
-                      "name": "layout.css.clip-path-path.enabled",
-                      "value_to_set": "true"
-                    }
-                  ]
-                }
-              ],
+              "firefox": {
+                "version_added": "63",
+                "partial_implementation": true,
+                "notes": [
+                  "From version 63 it is supported on the <code>offset-path</code> property.",
+                  "From version 71 it is supported on the <code>clip-path</code> property (from version 64 to 71 it is supported behind the preference <code>layout.css.clip-path-path.enabled</code>).",
+                  "From version 97 it can be used on the <code>d</code> SVG presentation attribute (from version 91 to 97 it is supported behind the preference <code>layout.css.d-property.enabled</code>)."
+                ]
+              },
+              "firefox_android": {
+                "version_added": "63",
+                "partial_implementation": true,
+                "notes": [
+                  "From version 63 it is supported on the <code>offset-path</code> property.",
+                  "From version 79 it is supported on the <code>clip-path</code> property (from version 64 to 79 it is supported behind the preference <code>layout.css.clip-path-path.enabled</code>).",
+                  "From version 97 it can be used on the <code>d</code> SVG presentation attribute."
+                ]
+              },
               "ie": {
                 "version_added": false
               },
--- a/css/types/basic-shape.json
+++ b/css/types/basic-shape.json
@@ -270,24 +270,67 @@
                 "partial_implementation": true,
                 "notes": "Only supported on the <code>offset-path</code> property."
               },
-              "firefox": {
-                "version_added": "63",
-                "partial_implementation": true,
-                "notes": [
-                  "From version 63 it is supported on the <code>offset-path</code> property.",
-                  "From version 71 it is supported on the <code>clip-path</code> property (from version 64 to 71 it is supported behind the preference <code>layout.css.clip-path-path.enabled</code>).",
-                  "From version 97 it can be used on the <code>d</code> SVG presentation attribute (from version 91 to 97 it is supported behind the preference <code>layout.css.d-property.enabled</code>)."
-                ]
-              },
-              "firefox_android": {
-                "version_added": "63",
-                "partial_implementation": true,
-                "notes": [
-                  "From version 63 it is supported on the <code>offset-path</code> property.",
-                  "From version 79 it is supported on the <code>clip-path</code> property (from version 64 to 79 it is supported behind the preference <code>layout.css.clip-path-path.enabled</code>).",
-                  "From version 97 it can be used on the <code>d</code> SVG presentation attribute."
-                ]
-              },
+              "firefox": [
+                {
+                  "version_added": "63",
+                  "partial_implementation": true,
+                  "notes": [
+                    "From version 97, the <code>d</code> SVG presentation attribute supports <code>path()</code>.",
+                    "From version 71, the <code>clip-path</code> property supports <code>path()</code>.",
+                    "Before version 71, only the <code>offset-path</code> property supports <code>path()</code>."
+                  ]
+                },
+                {
+                  "version_added": "91",
+                  "version_removed": "97",
+                  "partial_implementation": true,
+                  "flags": [
+                    {
+                      "type": "preference",
+                      "name": "layout.css.d-property.enabled",
+                      "value_to_set": "true"
+                    }
+                  ],
+                  "notes": "The <code>d</code> SVG presentation attribute, <code>clip-path</code> CSS property, and <code>offset-path</code> CSS property support <code>path()</code>."
+                },
+                {
+                  "version_added": "64",
+                  "version_removed": "71",
+                  "partial_implementation": true,
+                  "flags": [
+                    {
+                      "type": "preference",
+                      "name": "layout.css.clip-path-path.enabled",
+                      "value_to_set": "true"
+                    }
+                  ],
+                  "notes": "The <code>clip-path</code> and <code>offset-path</code> properties support <code>path()</code>."
+                }
+              ],
+              "firefox_android": [
+                {
+                  "version_added": "63",
+                  "partial_implementation": true,
+                  "notes": [
+                    "From version 97, the <code>d</code> SVG presentation attribute supports <code>path()</code>.",
+                    "From version 79, the <code>clip-path</code> property supports <code>path()</code>.",
+                    "Before version 79, only the <code>offset-path</code> property supports <code>path()</code>."
+                  ]
+                },
+                {
+                  "version_added": "64",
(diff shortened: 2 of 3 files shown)


css/types/basic-shape.json
```diff
@@ -270,30 +270,46 @@
...
-                    "name": "layout.css.clip-path-path.enabled",
-                    "value_to_set": "true"
-                  }
-                ]
-              },
+              "firefox": [
+                {
+                  "version_added": "97",
+                  "partial_implementation": true,
+                  "notes": "It is supported on the <code>d</code> SVG presentation attribute, <code>clip-path</code> and <code>offset-path</code> properties."
```
ddbeck, line 277: These notes (the originals included here) seem sketchy. `d` and `clip-path` are both SVG presentation attributes, right? And `offset-path` is a CSS property? Either we should make the distinction for each or for none, though I don't know which is right (since I'm a bit ignorant of SVG).

css/types/basic-shape.json
```diff
@@ -270,30 +270,46 @@
...
+                },
+                {
+                  "version_added": "63",
+                  "version_removed": "97",
+                  "partial_implementation": true,
+                  "notes": "From version 91 it can be used on the <code>d</code> SVG presentation attribute. For earlier versions it is only supported on the <code>clip-path</code> and <code>offset-path</code> properties.",
```
ddbeck, line 283: It's not clear whether `d` is the _only_ supported attribute from version 91. Here we need two notes that stand alone, I think.

1. From version 91…
2. Before version 91…

ddbeck (changes requested):
> In addition, I THINK it would be useful to update the [`d` attribute](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/d) to indicate that from FF97 it is now a "presentation attribute", which is what allows it to be modified in CSS and set to the [`path()`](https://developer.mozilla.org/en-US/docs/Web/CSS/path()) or `none`. Not sure right way to do this - subfeature "supports_css_path()" or similar. Or would we do this as a "reversed" one - e.g. not_supports_css_path. Note that current chrome does do the right thing. @ddbeck Advice?

I'm not sure, honestly. I don't know SVG that well. What kind of attribute was it before? Was it specified as that? Might be easier to have this discussion on a draft PR, though, since you've already got an idea and we may as well start from there.

> Further, any reason not to add these spec URLs that are currently manually in the page here https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/d#specifications ?

None that I can see. Proceed!

One extra thing: the MDN URL points to a now-non-existent fragment. I'd like to see that go away.

css/types/basic-shape.json
hamishwillee: Both offset-path and clip-path are CSS properties:

As I understand it,
- Generally CSS can only modify CSS properties. From v63 the `path()` method could be called on just 2 of these properties `clip-path` and `offset-path`
- "Generally" you can't set the value of SVG attributes using CSS - the set of SVG attributes that you can set using CSS are called [presentation attributes](https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/Presentation#attr-d). `d` is one of these, so you should be able to set the value of it using `path()`. However you can't do that until FF97.
- Now there are other presentation attributes than `d`, but `d` is the only one that you can set with `path()` and path is what we are talking about here.

So from v63 the story was :

> "notes": "It is supported on <code>clip-path</code> and <code>offset-path</code> CSS properties."

The v97 and later story is:

> "notes": "It is supported on <code>clip-path</code> and <code>offset-path</code> CSS properties."
> It is supported on  the <code>d</code> SVG presentation attribute.

Hope that makes sense. 

I think you are suggesting I modify this as:
```suggestion
                  "notes": [
                    "It is supported on the <code>d</code> SVG presentation attribute",
                    "it is supported on the  <code>clip-path</code> and <code>offset-path</code> CSS properties."
                  ]
```

hamishwillee:
This got a bit messier. Just found out 71 was where the clip-path stuff was enabled by default https://bugzilla.mozilla.org/show_bug.cgi?id=1488530. Let me have another go.

ddbeck:
Ugh, this is a bit messy; I think it's an artifact of how we've recorded CSS functions in BCD (which is to say, abstractly—like mixins, instead of concrete interfaces). I think you've got the right idea, but perhaps we need to go a step further:

* Use support statements on `path()` feature to show partial support, with notes showing the introduction of each supported property (basically, continue down the path started in this PR)
* Add/update/reconcile subfeatures of

  - `css.properties.clip-path`
  - `css.properties.offset-path`
  - `svg.elements.path.d`

  to show support for `path()` on those specific items. Some of them already have a `path()` subfeature, so it's just a matter of making sure the numbers are consistent.

lucalves (approved):

css/properties/clip-path.json
```diff
@@ -314,7 +314,22 @@
                   "version_added": "71"
                 },
                 {
-                  "version_added": "63",
+                  "version_added": "64",
```
hamishwillee, line 317: @ddbeck FYI
- changes version added to 64, based on tracking flag in bug https://bugzilla.mozilla.org/show_bug.cgi?id=1487838
- Mirrored the release version for FF for Android.

css/types/basic-shape.json
```diff
@@ -273,25 +273,19 @@
...
                 "version_added": "63",
                 "partial_implementation": true,
-                "notes": "Only supported on the <code>clip-path</code> and <code>offset-path</code> properties.",
-                "flags": [
-                  {
-                    "type": "preference",
-                    "name": "layout.css.clip-path-path.enabled",
-                    "value_to_set": "true"
-                  }
+                "notes": [
```
hamishwillee, line 276: So these are now all notes as suggested. Note that using this approach we lose the preference as flags.

hamishwillee:
@ddbeck Re https://github.com/mdn/browser-compat-data/pull/14498#issuecomment-1015523620 - what I have done"

1. `css.properties.offset-path` - was already correct.
2. `css.properties.clip-path` - updated to match bugzilla version and match FF desktop and android builds to closest versions
3. `svg.elements.path.d` - added `svg.element.path.d.path` - Chrome and Safari versions tested on browserstack. Everything else mirrored from Chrome.
4. Added all the information about version support/preferences as notes for basic-shapes BCD. Not entirely sure of this because  by doing this we can't have the original flags information (as there are a bunch of flags that change this and they affect different versions).
5. The chrome info for basic-shape.path says ""notes": "Only supported on the <code>offset-path</code> property."" but I know that this is also supported from v52 on `d`. Perhaps add a note for that too in the same way as for firefox?

css/types/basic-shape.json
ddbeck: Oh, interesting! Writing the notes shows a much more complex story than I gleaned from the PR commentary. Here's how I'd represent this:

```json
[
  {
    "version_added": "63",
    "notes": [
      "From version 97, the <code>d</code> SVG presentation attribute supports <code>path()</code>.",
      "From version 71, the <code>clip-path</code> property supports <code>path()</code>.",
      "Before version 71, only the <code>offset-path</code> property supports <code>path()</code>."
    ]
  },
  {
    "version_added": "91",
    "version_removed": "97",
    "flags": [
      {
        "type": "preference",
        "name": "layout.css.d-property.enabled",
        "value_to_set": "true"
      }
    ],
    "notes": [
      "The <code>d</code> SVG presentation attribute, <code>clip-path</code> CSS property, and <code>offset-path</code> CSS property support <code>path()</code>."
    ]
  },
  {
    "version_added": "64",
    "version_removed": "71",
    "flags": [
      {
        "type": "preference",
        "name": "layout.css.clip-path-path.enabled",
        "value_to_set": "true"
      }
    ],
    "notes": [
      "The <code>clip-path</code> and <code>offset-path</code> properties support <code>path()</code>."
    ]
  }
]
```

There's a lot going here! Some highlights:

- Unflagged support goes first, even though the flag stuff is more recent. The idea here is that the most expansive support goes first, not necessarily the most recent.
- Notes are sorted with most recent first, since the other notes should become increasingly less interesting as time goes on.
- The notes attached to the flagged statements don't need to specify version numbers, since it applies to the whole range covered by that statement.
- The notes have been reworded to lose the ambiguous "it" which had me completely lost before long.

All that said, I'm not sure if this information will copy over to the properties themselves very well. I'd be okay with dropping the older of the two flagged statements (in following [this guideline](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-flag-data)), if making everything reconcile is easier that way.

svg/elements/path.json
```diff
@@ -94,6 +94,67 @@
               "standard_track": true,
               "deprecated": false
             }
+          },
+          "path": {
+            "__compat": {
+              "description": "<code>path()</code> function can be used to set value in CSS",
```
ddbeck, line 100: Does this mean something different than "you can use `path()` in the `d` attribute"? If not…

```suggestion
              "description": "<code>path()</code>",
```

If I've misunderstood, then help me out here because I'm not sure what distinction the current description text is making. Maybe give me an example?

ddbeck (changes requested):
More suggestions for you, @hamishwillee. This has been a tough one. Forgive me. 🙏

svg/elements/path.json
hamishwillee: @ddbeck In summary, I added that description because everything else is about "static" SVG. The uninformed seeing `path()` might assume it is something that you specify in an SVG file. Actually it is something you specify in CSS to modify the SVG - as shown in https://developer.mozilla.org/en-US/docs/Web/SVG/Attribute/d#using_d_as_a_css_property

I've now added an MDN URL to the function to which will help here. 

Upshot, if you think the distinction is something for MDN rather than BCD I'm OK with that - either add a further response or press that commit button!

css/types/basic-shape.json
hamishwillee: @ddbeck Thanks very much. I don't find anywhere near as easy to understand in this context, but I guess it is easier when someone sees the rendered version - as they can look at the first entry and work out the first versions where support exists. 

I have copied this in. No need to drop the older statements. The properties themselves are much easier to document because they have a smaller context to work in. See `clip-path` in top page above. 

I'll add a few comments against the new version :-)

css/types/basic-shape.json
```diff
@@ -270,30 +270,67 @@
...
+                  "version_removed": "97",
+                  "partial_implementation": true,
+                  "flags": [
+                    {
+                      "type": "preference",
+                      "name": "layout.css.d-property.enabled",
+                      "value_to_set": "true"
+                    }
+                  ],
+                  "notes": "The <code>d</code> SVG presentation attribute, <code>clip-path</code> CSS property, and <code>offset-path</code> CSS property support <code>path()</code>."
```
hamishwillee, line 294: @ddbeck FYI only, I find these notes a little confusing because offset-path and clip-path are fully supported - it is only `d` use of path that behind the pref. I guess the theory is that they understand all these things are available if they enable the preference.

css/types/basic-shape.json
```diff
@@ -270,30 +270,67 @@
...
+                  "version_added": "63",
+                  "partial_implementation": true,
+                  "notes": [
+                    "From version 97, the <code>d</code> SVG presentation attribute supports <code>path()</code>.",
+                    "From version 79, the <code>clip-path</code> property supports <code>path()</code>.",
+                    "Before version 79, only the <code>offset-path</code> property supports <code>path()</code>."
+                  ]
+                },
+                {
+                  "version_added": "64",
```
hamishwillee, line 321: Note, I copied the pattern for FF android too. Note though that only the earlier preference applies - after v79 the settings UI disappeared.

css/types/basic-shape.json
ddbeck: That's a fair point. We can revise this to be a little more explicit.

```suggestion
                  "notes": "With the preference enabled, the <code>d</code> SVG presentation attribute, <code>clip-path</code> CSS property, and <code>offset-path</code> CSS property support <code>path()</code>."
```

svg/elements/path.json
ddbeck: This is still an outstanding question, or it got addressed elsewhere and it got lost in the shuffle. An answer or a pointer in the right direction would be helpful here.

css/types/basic-shape.json
```diff
@@ -270,30 +270,67 @@
...
+                  "version_removed": "71",
+                  "partial_implementation": true,
+                  "flags": [
+                    {
+                      "type": "preference",
+                      "name": "layout.css.clip-path-path.enabled",
+                      "value_to_set": "true"
+                    }
+                  ],
+                  "notes": "The <code>clip-path</code> and <code>offset-path</code> properties support <code>path()</code>."
```
ddbeck, line 307:
```suggestion
                  "notes": "With the preference enabled, the <code>clip-path</code> and <code>offset-path</code> properties support <code>path()</code>."
```

css/types/basic-shape.json
```diff
@@ -270,30 +270,67 @@
...
+                  "version_removed": "79",
+                  "partial_implementation": true,
+                  "flags": [
+                    {
+                      "type": "preference",
+                      "name": "layout.css.clip-path-path.enabled",
+                      "value_to_set": "true"
+                    }
+                  ],
+                  "notes": "The <code>clip-path</code> and <code>offset-path</code> properties support <code>path()</code>."
```
ddbeck, line 331:
```suggestion
                  "notes": "With the preference enabled, the <code>clip-path</code> and <code>offset-path</code> properties support <code>path()</code>."
```

ddbeck (changes requested):
Getting pretty close here. Thanks again, Hamish!

svg/elements/path.json
hamishwillee: @ddbeck I "definitely" wrote an answer for this - but as you say, not present.

My thinking was that everything else about this definition indicates support for SVG compatibility - so `svg.elements.path.d` indicates that you can set the SVG path in the `d` element. Now we have added `svg.elements.path.d.path` which is a function that allows you to set this SVG value in CSS. 

That seemed really unexpected to me: enough that clarifying with a proper description seemed sensible. Otherwise you might read this as SVG could call the path() function. 

Of course for BCD you have to read the docs to understand what the compatibility means, so this is not an unreasonable suggestion.

SO. Please self-commit and merge if you still think this should happen, or just merge if you think my argument is compelling :-)

hamishwillee:
@ddbeck I think "this" is ready. Your question answered in https://github.com/mdn/browser-compat-data/pull/14498/files#r809627789

svg/elements/path.json
hamishwillee: That is correct! I'll commit that.

hamishwillee:
I merged your commit. Should be good to go now.

ddbeck (approved):
Looks good. Thank you! 🎉

hamishwillee:
Thanks @ddbeck .- feels like a long time coming!

PR #14506, opened 2022-01-11
Add a note of URL Constructor in Chrome under MacOS

aleen42:
#### Summary

Recently, I found that Chrome had different behaviours when constructing URL with following patter:

```js
new URL('file:///C:/test.png') // => OK
new URL('file://C:/test.png') // => Throw Error: Invalid URL
```

In Windows, both are ok.

#### Polyfill

A possible workaround: https://github.com/aleen42/core-web/blob/main/lib/stable/URL/index.js

### DIFF
--- a/api/URL.json
+++ b/api/URL.json
@@ -123,7 +123,8 @@
           "description": "<code>URL()</code> constructor",
           "support": {
             "chrome": {
-              "version_added": "19"
+              "version_added": "19",
+              "notes": "Under MacOS, Chrome failed to construct URL with given <code>file://</code>."
             },
             "chrome_android": {
               "version_added": "25"


queengooborg:
Hey @aleen42, thanks for your PR!  It's odd that Chrome has such a discrepancy with the file URLs, where it throws an error with only two slashes on macOS, where the other browsers I've tested (Firefox and Safari) work just fine without that third slash on macOS.  However, reading through the [spec for URL strings](https://url.spec.whatwg.org/#relative-url-string) (and spending a good chunk of time trying to read through), I _don't_ think that the omitted third slash is actually a spec-valid URL string.

Please feel free to skip over the next two paragraphs (which took way too long to write, haha) if you'd like to hear the short and simple!

In the spec, it explains the variants of URL strings and how each differs based upon the schema or local vs. remote connections.  For our purposes, we're only concerned with one variant: a [scheme-relative-file-URL string](https://url.spec.whatwg.org/#scheme-relative-file-url-string).  The variant is defined as "//", followed by either: a [valid host string](https://url.spec.whatwg.org/#valid-host-string) (like `localhost` -- which actually works I've found), optionally followed by a [path-absolute-non-Windows-file-URL string](https://url.spec.whatwg.org/#path-absolute-non-windows-file-url-string), or...and this is the one that applies to us...a [path-absolute-URL string](https://url.spec.whatwg.org/#path-absolute-url-string).  It is stated that a path-absolute-URL string "**must be U+002F (/)** followed by a path-relative-URL string". (The non-Windows absolute path is just how it sounds: same thing, but Windows drive letters aren't allowed.)

What this is all saying is that after the double slashes, there **must** be a slash to begin the actual file path.  This is partially represented in the spec's [URL parsing examples](https://url.spec.whatwg.org/#example-url-parsing) -- the two local filepaths use three slashes for Windows paths, and the only one with two contains a host string (`localhost`).

All of this is a big explanation to basically say: I don't think we should document this discrepancy.  If anything, Chrome on macOS is seemingly properly conforming to the spec, as weird as it is that Chrome on Windows isn't, nor is any browser.  I'll keep this PR open, however, to allow for further discussion!

aleen42:
@queengooborg Thanks for the research.

ddbeck:
Interesting stuff, @queengooborg. With all the background, I agree with your recommendation that we don't need to cover this. It sounds like what we have here is a case where, with the exception of Chrome on macOS, browsers are more permissive about such URLs that the spec requires (also, I checked to see if there were any bugs relating to this and turned up https://bugs.chromium.org/p/chromium/issues/detail?id=1238236, which makes it seem somewhat intentional). While that's an interesting _difference_, I'm not sure there's much in the way of incompatibility to talk about here—it appears that all platforms correctly parse valid URLs.

With all that said, it doesn't look like any more discussion is needed. I'm going to close this PR. Thank you, @aleen42 and @queengooborg for getting into the details here.

PR #14577, opened 2022-01-18
Set ESR status on correct version

nchevobbe:
Firefox 60 isn't ESR anymore, it's Firefox 91 those days


ddbeck (approved):
Thanks for this PR. I see it's your first to the project—welcome! Looks good. 🎉

PR #14600, opened 2022-01-20
Save-Data renamed to Sec-CH-Save-Data

sideshowbarker:
Related MDN change: https://github.com/mdn/content/pull/12210

Related: https://github.com/WICG/savedata/pull/3
Related: https://github.com/WICG/client-hints-infrastructure/issues/76


sideshowbarker:
> Given Chromiums shipped this (I assume with the old name), the compat story should be more complex now and `alternative_name` should be used, right?

Yes. But actually I realize now that it’d be premature to merge this — because the spec change hasn’t been implemented in the Chromium sources yet. So I’ll either write a browser patch for it myself, or else just raise a Chromium bug for it.

hamishwillee:
Yeah, I didn't do this update because there are no implementations and there is some discussion about how the switchover will be managed. The concern being that this is a low entropy hint (always sent) and they don't want to have to send both versions. 

In addition to Save-Data perhaps being an alternative name, we might also need  `Sec-CH-Prefers-Reduced-Data`. It probably isn't implemented but it sounds like an alias: 
https://github.com/WICG/client-hints-infrastructure/issues/90#issuecomment-979758562

PR #14609, opened 2022-01-20
Update a bunch of Permissions API spec URLs

sideshowbarker:


Elchi3 (approved):
We still need to do the renaming of Feature-Policy to Permissions-Policy 😮 
Thanks for this, though

PR #14629, opened 2022-01-21
Add Cloudflare Workers to MDN compat table

jkup:
#### Summary
Hi folks!

I'm hoping you'll consider adding the Cloudflare Workers runtime to the server compatibility list. Like Node.js and Deno, the Cloudflare Workers runtime uses Google's v8 engine. We have been hard at work adding support for standard APIs and think MDN would be a great resource for people to check on our progress.

I followed along with the Deno PR and thought this would make a good start adding support for two fully-support APIs.

#### Note

I wanted to note that, if you all feel we'd make a good fit, we'd open PRs for all of the APIs that we support. We don't just support `AbortController` and `AbortSignal`!

### DIFF
--- a/api/AbortController.json
+++ b/api/AbortController.json
@@ -11,6 +11,9 @@
           "chrome_android": {
             "version_added": "66"
           },
+          "cloudflare_workers": {
+            "version_added": "2021-09-24"
+          },
           "deno": {
             "version_added": "1.0"
           },
@@ -79,6 +82,9 @@
             },
             "chrome_android": {
               "version_added": "66"
+            },
+            "cloudflare_workers": {
+              "version_added": "2021-09-24"
             },
             "deno": {
               "version_added": "1.0"
@@ -149,6 +155,9 @@
             "chrome_android": {
               "version_added": "66"
             },
+            "cloudflare_workers": {
+              "version_added": "2021-09-24"
+            },
             "deno": {
               "version_added": "1.0"
             },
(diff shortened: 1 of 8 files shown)


jasnell (approved):
LGTM as far as the workers version details are concerned

ddbeck:
Hi @jkup and thanks for this PR. I see it's your first to the project—welcome!

Have you had a chance to look at https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#addition-of-browsers? It's got some more detail about how we decide whether to add a browser or engine to BCD. I'm going to focus on the first point, but we'll need to establish all of them to move ahead.

The key thing here is to to establish that there's a data consumer interest in adding a new browser or engine to the data set. To talk about adding Cloudflare Workers to compat tables, we'd need to have a separate discussion on the mdn/content. In the past, when adding to the compat tables originated outside the MDN team, we've shared a questionnaire to help understand how the project serves MDN's audience, with questions about audience size, availability, etc. That said, if you know there's another BCD consumer interested in using such data (and could get them to chime in here), we don't need to block on the MDN question.

Does that help understand the path toward inclusion here? I know this is not super well documented, so I'm happy to answer any questions.

jkup:
@ddbeck Thank you so much! That's very helpful. I just want to make sure I understand correctly but it sounds like my best next step would be to file an issue over on https://github.com/mdn/content, is that right?

queengooborg:
Hey @jkup, thank you for your PR, however I don't feel that we can add Cloudflare Workers to BCD.  We unfortunately do not have the resources available to maintain server engines like Cloudflare Workers, especially as Daniel has left MDN Web Docs.  Thank you in advance for understanding!

PR #14633, opened 2022-01-21
Add nodejs version for api.TextDecoder.TextDecoder

vabushkevich:
#### Summary

This PR adds the real value for Node.js for the `TextDecoder` constructor.

Changelog for adding: https://nodejs.org/en/blog/release/v8.3.0/

> The WHATWG Encoding Standard (TextDecoder and TextEncoder) has been implemented as an experimental feature

Changelog for making the API global: https://nodejs.org/en/blog/release/v11.0.0/

> The WHATWG TextEncoder and TextDecoder are now globals

History section in the documentation: https://nodejs.org/docs/latest-v16.x/api/util.html#new-textdecoderencoding-options

#### Related issues

Part of #13170.

### DIFF
--- a/api/TextDecoder.json
+++ b/api/TextDecoder.json
@@ -116,9 +116,15 @@
             "ie": {
               "version_added": false
             },
-            "nodejs": {
-              "version_added": null
-            },
+            "nodejs": [
+              {
+                "version_added": "11.0.0"
+              },
+              {
+                "version_added": "8.3.0",
+                "notes": "Exported from the <code>util</code> module but not globally available."
+              }
+            ],
             "opera": {
               "version_added": "25"
             },


api/TextDecoder.json
```diff
@@ -116,9 +116,15 @@
...
-            "nodejs": {
-              "version_added": null
-            },
+            "nodejs": [
+              {
+                "version_added": "11.0.0"
+              },
+              {
+                "version_added": "8.3.0",
+                "notes": "Exported from the <code>util</code> module but not globally available."
```
queengooborg, line 125:
```suggestion
                "notes": "Available as a part of the <code>util</code> module."
```

queengooborg (changes requested):
Looking good, thanks!  Just got a small suggestion for the wording of the note:

api/TextDecoder.json
queengooborg: I realized this note was copied from that of the rest of `TextDecoder` and `TextEncoder`.  I'm going to open a PR to update this in the other locations!

ddbeck: It looks like the parent feature also uses `partial_implementation`. I'm wondering if the right thing to do here is to do the same as the parent, but revisit this to use `alternative_name` instead, which I think would more accurately capture what's happening.

ddbeck: (I'll defer to @queengooborg whether the current changes satisfy though. A little _more_ inconsistency is probably not a huge problem here.)

queengooborg: We should certainly update the child features to mark them as `partial_implementation`, but I think we can perform that in a separate PR!

queengooborg (approved):

PR #14660, opened 2022-01-25
Complete nodejs data for EventTarget

vabushkevich:
#### Summary

This PR adds the last missing real values for Node.js for the `EventTarget` API.

[The documentation](https://nodejs.org/docs/latest-v16.x/api/events.html#eventtargetaddeventlistenertype-listener-options) says that `options.passive` is `false` by default without notes about any special cases. I also looked at the [changelogs](https://github.com/nodejs/node/tree/master/doc/changelogs) and didn't find any notes about `options.passive`.

#### Related issues

Part of #13170.

### DIFF
--- a/api/EventTarget.json
+++ b/api/EventTarget.json
@@ -428,7 +428,7 @@
                   "version_added": false
                 },
                 "nodejs": {
-                  "version_added": null
+                  "version_added": false
                 },
                 "opera": {
                   "version_added": "42"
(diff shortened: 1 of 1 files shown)


queengooborg (approved):
Thank you for your PR, and welcome to BCD!  🎉   I have no clue how a user would be able to obtain a touch or wheel event in Node.js, so seeing "default is false" in the documentation is good enough for me.  👍

PR #14691, opened 2022-01-27
@layer removed from behind a pref in FF

Rumyra:
#### Summary

@layer is now available by default in FF97

~Weirdly done this through gh ui because tests are failing locally and I can't see why - interested to see what happens here~ found it, switching contexts helped

see https://bugzilla.mozilla.org/show_bug.cgi?id=1699215

#### Related issues

Part of https://github.com/mdn/content/issues/9369

### DIFF
--- a/css/at-rules/layer.json
+++ b/css/at-rules/layer.json
@@ -18,28 +18,40 @@
             "edge": {
               "version_added": false
             },
-            "firefox": {
-              "version_added": "94",
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "layout.css.cascade-layers.enabled",
-                  "value_to_set": "true"
-                }
-              ],
-              "notes": "See Firefox <a href='https://bugzil.la/1699217'>bug 1699217</a>."
-            },
-            "firefox_android": {
-              "version_added": "94",
-              "flags": [
-                {
-                  "type": "preference",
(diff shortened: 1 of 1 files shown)


css/at-rules/layer.json
```diff
@@ -18,28 +18,40 @@
...
+                "flags": [
+                  {
+                    "type": "preference",
+                    "name": "layout.css.cascade-layers.enabled",
+                    "value_to_set": "true"
+                  }
+                ],
+                "notes": "See Firefox <a href='https://bugzil.la/1699217'>bug 1699217</a>."
+              }
+            ],
```
queengooborg, line 54: Fenix (Firefox Android) doesn't actually support flags -- can we remove the data regarding the flag and simplify it to just the following?

```suggestion
            "firefox_android": {
              "version_added": "97"
            },
```

queengooborg (changes requested):
Looks good, thanks!  Just one quick update regarding Firefox Android:

css/at-rules/layer.json
Rumyra: Yeh absolutely - and thanks @queengooborg - I'll remember that in future :)

ddbeck (approved):
Looks good. Thank you! 🎉

PR #14696, opened 2022-01-27
Mark Opera Android 66 as current

lukewarlow:
#### Summary

Add Opera Android 66 and beta 67 and corresponding data

Version 66:

Version 67:


Elchi3 (approved):
Thanks! 👍 

I'm assuming you used some sort of script (or the mirror script we have in the repo) to catch all these. I skimmed the diff and it makes sense but I didn't check if this is everything.

PR #14757, opened 2022-01-31
Add Safari 15.4 support for Dialog element and ::backdrop

jdatapple:
#### Summary
Safari 15.4 beta supports the `<dialog>` element and the CSS `::backdrop` selector.

See [Safari 15.4 Beta Release Notes](https://developer.apple.com/documentation/safari-release-notes/safari-15_4-release-notes)

### DIFF
--- a/api/HTMLDialogElement.json
+++ b/api/HTMLDialogElement.json
@@ -46,10 +46,10 @@
             "version_added": "24"
           },
           "safari": {
-            "version_added": false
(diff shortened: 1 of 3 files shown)


Elchi3 (changes requested):
This should also update https://github.com/mdn/browser-compat-data/blob/main/html/elements/dialog.json

Elchi3 (approved):

PR #14760, opened 2022-01-31
Corrected Safari 15.4 support for the FileSystem API

jdatapple:
#### Summary
Fixed Safari/Safari Technology Preview support information for FileSystemHandle, FileSystemWritableFileStream and StorageManager.

See [Safari 15.4 Beta Release Notes](https://developer.apple.com/documentation/safari-release-notes/safari-15_4-release-notes)

### DIFF
--- a/api/FileSystemHandle.json
+++ b/api/FileSystemHandle.json
@@ -225,7 +225,7 @@
               "version_added": false
             },
             "safari": {
-              "version_added": "preview"
+              "version_added": false
             },
             "safari_ios": {
               "version_added": false
@@ -274,7 +274,7 @@
               "version_added": false
             },
             "safari": {
-              "version_added": "preview"
+              "version_added": false
             },
             "safari_ios": {
               "version_added": false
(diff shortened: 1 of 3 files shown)


Elchi3 (commented):
The commit message is "Added Safari 15.4 support for the FileSystem API" but then you change the values to `false`. So, did you mean to say that Safari no longer supports the FileSystem API or is the change to `false` wrong?

gsnedders:
The commit message seems to be wrong.

Looking at https://github.com/WebKit/WebKit/blob/main/Source/WebCore/Modules/filesystemaccess/FileSystemHandle.idl and its history makes it look like there never was support for those in api/FileSystemHandle.json.

I don't see any evidence for FileSystemWritableFileStream ever being supported.

https://github.com/WebKit/WebKit/blob/main/Source/WebCore/Modules/storage/StorageManager.idl has estimate commented out.

So this commit seems accurate aside from the commit message.

jdatapple:
I've corrected the commit message.

Elchi3 (approved):
Thanks for the clarification! 👍

PR #14775, opened 2022-02-01
Remove "Streaming response body" compat data

evilpie:
I think this must be referring to Response.body, which should be supported since 65: https://developer.mozilla.org/en-US/docs/Web/API/Response/body.

#### Summary
"Streaming response body" feature of fetch is supported in Firefox

https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/65
> Readable Streams have been enabled by default

### DIFF
--- a/api/_globals/fetch.json
+++ b/api/_globals/fetch.json
@@ -222,13 +222,7 @@
               "version_added": "14"
             },
             "firefox": {
-              "version_added": "64",
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "javascript.options.streams"
-                }
-              ]
+              "version_added": "65",
             },
             "firefox_android": {
               "version_added": false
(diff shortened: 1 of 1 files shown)


saschanaz:
Maybe it should be removed in favor of Response.body in that case? 🤔

queengooborg:
I'm wondering if support was also enabled by default in Firefox Android at the same time?

evilpie:
> Maybe it should be removed in favor of Response.body in that case? thinking

That would be a good idea as well.

> I'm wondering if support was also enabled by default in Firefox Android at the same time?

Yes, why shouldn't it?

evilpie:
@queengooborg Could you merge this?

ddbeck (approved):
Yeah, this feels like a pretty clear case of unwanted duplication. Thank you! 🎉

PR #14821, opened 2022-02-04
FF98 WebVR support disabled by default

hamishwillee:
WebVR has been disabled by default in FF98 - https://bugzilla.mozilla.org/show_bug.cgi?id=1750902 . Other docs work for this in https://github.com/mdn/content/issues/12582

So Web VR is apparently out of the standards, superseded by WebXR. It was implemented in Chrome Android until v80 and then removed. It is still marked as supported in Samsung Internet. It was implemented on FF in Windows and MacOS from around v55/64, but only on nightly and dev for other platforms.
This change turns it off by default on all platforms. 

I have made the changes just to [VRDisplay](https://developer.mozilla.org/en-US/docs/Web/API/VRDisplay). Once we confirm the right approach I'll role them out widely for the API within this same PR. 

@ddbeck There are questions inline to test changes. 
MAIN question is whether you would instead delete this and all associated MDN documents? I am assuming you don't want to because this will still be supported on Samsung internet.

### DIFF
--- a/api/VRDisplay.json
+++ b/api/VRDisplay.json
@@ -130,7 +130,8 @@
               }
             ],
             "firefox_android": {
-              "version_added": "55"
+              "version_added": "55",
+              "version_removed": "98"
             },
             "ie": {
               "version_added": false
@@ -211,7 +212,8 @@
               }
             ],
             "firefox_android": {
-              "version_added": "55"
+              "version_added": "55",
+              "version_removed": "98"
             },
             "ie": {
               "version_added": false
@@ -292,7 +294,8 @@
               }
             ],
             "firefox_android": {
-              "version_added": "55"
+              "version_added": "55",
+              "version_removed": "98"
             },
             "ie": {
               "version_added": false
@@ -373,7 +376,8 @@
               }
             ],
             "firefox_android": {
-              "version_added": "55"
+              "version_added": "55",
+              "version_removed": "98"
             },
             "ie": {
               "version_added": false
@@ -454,7 +458,8 @@
               }
             ],
             "firefox_android": {
-              "version_added": "55"
+              "version_added": "55",
+              "version_removed": "98"
             },
             "ie": {
               "version_added": false
@@ -535,7 +540,8 @@
               }
             ],
             "firefox_android": {
-              "version_added": "55"
+              "version_added": "55",
+              "version_removed": "98"
             },
             "ie": {
               "version_added": false
@@ -616,7 +622,8 @@
               }
             ],
             "firefox_android": {
-              "version_added": "55"
+              "version_added": "55",
+              "version_removed": "98"
             },
             "ie": {
               "version_added": false
@@ -697,7 +704,8 @@
               }
             ],
             "firefox_android": {
-              "version_added": "55"
+              "version_added": "55",
+              "version_removed": "98"
             },
             "ie": {
               "version_added": false
@@ -778,7 +786,8 @@
               }
             ],
             "firefox_android": {
-              "version_added": "55"
+              "version_added": "55",
+              "version_removed": "98"
             },
             "ie": {
               "version_added": false
@@ -846,7 +855,8 @@
               }
             ],
             "firefox_android": {
-              "version_added": "55"
+              "version_added": "55",
+              "version_removed": "98"
             },
             "ie": {
               "version_added": false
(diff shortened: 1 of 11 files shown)


api/VRDisplay.json
```diff
@@ -64,7 +77,7 @@
         },
         "status": {
           "experimental": true,
-          "standard_track": true,
+          "standard_track": false,
```
hamishwillee, line 80: @ddbeck I think this is no longer standards track, so changed `false` everywhere. OK?

api/VRDisplay.json
```diff
@@ -27,17 +27,30 @@
...
             {
               "version_added": "64",
+              "version_removed": "98",
               "notes": "macOS support was enabled in Firefox 64."
             }
           ],
           "firefox_android": {
-            "version_added": "55"
+            "version_added": "55",
+            "version_removed": "98"
```
hamishwillee, line 53: @ddbeck Not sure what to do for android. It is definitely not present from FF98 as it is hidden behind a preference. However my understanding is that it was only enabled by default for Windows/macOS, and only for nightly/dev on other platforms, including Android. 

My **guess** is that it was never supported on android so this should be set false. However it is possible it was initially supported and then moved to only being enabled in dev/nightly -  in which case we would need to find the specific version. 

Thoughts?  [Note, have only made this change here, pending discussion]

api/VRDisplay.json
```diff
@@ -27,17 +27,30 @@
...
+                {
+                  "type": "preference",
+                  "name": "dom.vr.enabled",
+                  "value_to_set": "true"
+                }
+              ]
+            },
             {
               "version_added": "55",
+              "version_removed": "98",
```
hamishwillee, line 42: @ddbeck Note, there are other ways that the older support might have been indicated. I have chosen just to mark these as removed because there isn't much point playing with an API that is "on the way out".

api/VRDisplay.json
ddbeck: Yes, this is right. 👍

api/VRDisplay.json
ddbeck: Hmm. I'm not sure. Some poorly directed searching shows it was available in Nightly at some point, but I don't know whether it was ever enabled. I'm going to ask for some advice here:

@queengooborg have you ever tried any WebVR stuff out on a mobile browser? The collector doesn't cover it, but maybe you've got a recommendation on how to go about testing this? (I don't want or care to do a comprehensive test of the API, just a quick reality check on whether it was supported in Firefox for Android or not)

api/VRDisplay.json
hamishwillee: As discussed in 1:1, there are many ways to indicate the legacy partial support in BCD. Rather than using any of these I plan to just follow the pattern of the existing BCD, because it isn't worth messing with that for an API that is on the way out.
Resolving

api/VRDisplay.json
hamishwillee: FYI, As discussed in 1:1 I am going to roll it out like this everywhere on the assumption the original info is correct. This will  be better than we have now, going forward. If @queengooborg or someone else finds we're wrong we can adjust, just as we are doing now.

hamishwillee:
@ddbeck API is marked up as discussed in first case (still looks OK to me). The android stuff is marked as removed - I'm assuming original info was correct. Ready for review.

api/VRDisplay.json
ddbeck: Yes, that's fine. Seems to have only been applied in this one place, however. Want to push another commit that does this to all of Firefox for Android?

ddbeck: On further review, it appears that the top-level interfaces have been updated, but not their descendants. This is the cause of the linter failure; the descendants are inconsistent with the parent features.

ddbeck (changes requested):
Thanks, Hamish. I think the remaining work is to resolve the inconsistencies for Android. See the two comments in that conversation.

api/VRDisplay.json
hamishwillee: Arrrg. Thanks. Fixed.

hamishwillee:
@ddbeck Thanks for the debug. Subfeatures fixed so should be good to go now.

ddbeck (approved):
Thank you! 🎉

PR #14858, opened 2022-02-06
Mirror Firefox -> Firefox Android WebGL data

queengooborg:
This PR adds real values to Firefox Android for the WebGL data.  The data is verified using the mdn-bcd-collector project to check for support, and then version numbers come from mirroring the data.

I found a [bug that includes a commen](https://bugzilla.mozilla.org/show_bug.cgi?id=647802#c3)t talking about a WebGL feature bugging out in Firefox 4 beta.  Based upon this comment, I believe that WebGL support was introduced to Firefox Android since the very beginning.

### DIFF
--- a/api/ANGLE_instanced_arrays.json
+++ b/api/ANGLE_instanced_arrays.json
@@ -18,7 +18,7 @@
             "version_added": "47"
           },
           "firefox_android": {
-            "version_added": true
+            "version_added": "47"
           },
           "ie": {
             "version_added": "11"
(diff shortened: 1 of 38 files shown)


Elchi3 (approved):
Thanks @queengooborg 👍 

I'm skeptical when it comes to the WebGL extensions as support there is often quite platform- and even hardware-specific, but I see we've also mirrored other mobile browsers already. 

Generally, I think, all the true -> version number changes are okay with me, more doubt with the null or false -> version number changes. However, I think merging this altogether is fine and the data could always be refined if we have new information.

PR #14874, opened 2022-02-07
FF97 WebMidi support - desktop

hamishwillee:
FF97 added support for [WebMidi API](https://developer.mozilla.org/en-US/docs/Web/API/Web_MIDI_API) backends for desktop platforms in https://bugzilla.mozilla.org/show_bug.cgi?id=836897 behind a preference: Linux, MacOS, Windows.
(Android does not have a backend yet, but is by definition `false` for support because of the preference). 

This just updates [MIDIInputMap](https://developer.mozilla.org/en-US/docs/Web/API/MIDIInputMap) - I will roll out to other parts if this is agreed good.

@ddbeck There are a few nuances. 
- The API consists of a front end and platform-specific backends. The front end including preference has been around since FF60 according to browserstack. This allows you to read files and output to "local synth" (your computer speakers) - e.g. run this kind of page: https://blog.karimratib.me/demos/sheetplayer/. 
  - This is not considered "real" midi by the team (see https://bugzilla.mozilla.org/show_bug.cgi?id=836897#c119) - it is referred to as a "dummy implementation". It doesn't look all that dummy to me, but I guess that is the team's call. 
  - My concern is that someone trying this on FF60 to 97 will see the preference, and see that they can call the API. So they will think we got BCD wrong. **Should we perhaps add another note**: "From version 60 to version 96 a dummy implementation is provided; you can call the API but not route MIDI to an output"?
- There are two other prefs which we can ignore for BCD:
  - `dom.webmidi.gated` causes the Web MIDI implementation to be gated by a new add-on based security policy. But this is on by default from this first real version 97, so IMO does not need to be mentioned.
  - `midi.tested` enables some behavior required for testing - can be ignored.

### DIFF
--- a/api/MIDIAccess.json
+++ b/api/MIDIAccess.json
@@ -15,7 +15,14 @@
             "version_added": "79"
           },
           "firefox": {
-            "version_added": false
+            "version_added": "97",
+            "flags": [
+              {
+                "type": "preference",
+                "name": "dom.webmidi.enabled",
+                "value_to_set": "true"
+              }
+            ]
           },
           "firefox_android": {
             "version_added": false
@@ -63,7 +70,14 @@
               "version_added": "79"
             },
             "firefox": {
-              "version_added": false
+              "version_added": "97",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "dom.webmidi.enabled",
+                  "value_to_set": "true"
+                }
+              ]
             },
             "firefox_android": {
               "version_added": false
@@ -112,7 +126,14 @@
               "version_added": "79"
             },
             "firefox": {
-              "version_added": false
+              "version_added": "97",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "dom.webmidi.enabled",
+                  "value_to_set": "true"
+                }
+              ]
             },
             "firefox_android": {
               "version_added": false
@@ -161,7 +182,14 @@
               "version_added": "79"
             },
             "firefox": {
-              "version_added": false
+              "version_added": "97",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "dom.webmidi.enabled",
+                  "value_to_set": "true"
+                }
+              ]
             },
             "firefox_android": {
               "version_added": false
@@ -210,7 +238,14 @@
               "version_added": "79"
             },
             "firefox": {
-              "version_added": false
+              "version_added": "97",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "dom.webmidi.enabled",
+                  "value_to_set": "true"
+                }
+              ]
             },
             "firefox_android": {
               "version_added": false
@@ -259,7 +294,14 @@
               "version_added": "79"
             },
             "firefox": {
-              "version_added": false
+              "version_added": "97",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "dom.webmidi.enabled",
+                  "value_to_set": "true"
+                }
+              ]
             },
             "firefox_android": {
               "version_added": false
--- a/api/MIDIConnectionEvent.json
+++ b/api/MIDIConnectionEvent.json
@@ -15,7 +15,14 @@
             "version_added": "79"
           },
           "firefox": {
-            "version_added": false
+            "version_added": "97",
+            "flags": [
+              {
+                "type": "preference",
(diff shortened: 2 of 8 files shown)


ddbeck:
> The API consists of a front end and platform-specific backends

Can you elaborate on this point? Are there separate interfaces or methods? I'm not clear on what the distinction is between a "dummy" implementation and a complete one. Or to put this another way, at what point would a developer discover that the API doesn't do the thing that they expect (i.e., using some MIDI output)?

hamishwillee:
@ddbeck  Re https://github.com/mdn/browser-compat-data/pull/14874#issuecomment-1032608442

> Can you elaborate on this point? Are there separate interfaces or methods?

No, the public API (or "front end") is the same for all cases. The back end is the platform-specific implementation. 

> I'm not clear on what the distinction is between a "dummy" implementation and a complete one. 

So my understanding is that the MIDI interface allows you to define midi inputs and outputs. These might be defined in software or be links to hardware. The dummy implementation only covers the software input/outputs, while the new plugins define the integration to MIDI hardware - ie you can iterate through connected midi hardware.

So if you look at https://blog.karimratib.me/demos/sheetplayer/ the output here is local synth, which means this plays on your computer as an output. That is the dummy implementation - it can't actually integrate with Midi hardware. With the "back ends" this would show a list of possible hardware outputs too. 

image

> Or to put this another way, at what point would a developer discover that the API doesn't do the thing that they expect (i.e., using some MIDI output)?

When they try it with real MIDI hardware. 

My first impression was that we have some kind of basic implementation in FF60 and then plugin support for midi hardware in FF97. However when I asked in https://bugzilla.mozilla.org/show_bug.cgi?id=836897#c119 they called it a "dummy". So the team at least don't think of this as delivering anything useful. 

That's why I characterised this as introduced in FF97. But of course we still want people to not be surprised by the fact that the API appears to "work".

I could alternatively make FF60 be a partial implementation on the outputs (I will have to check inputs). Kind of depends whether I respect the term "dummy" to mean "we don't consider this an implementation". Happy to be lead by you if you have an opinion.

infojunkie:
Small clarification regarding
> With the "back ends" this would show a list of possible hardware outputs too.
Not only hardware outputs, but other software MIDI input ports to which a MIDI output can be connected. Examples:

- [TiMidity++](http://timidity.sourceforge.net/), a software synthesizer which is widely available on Linux platforms.
- [DX7 Synth](http://mmontag.github.io/dx7-synth-js/), a Yamaha DX7 emulator that uses Wed MIDI to receive its input.

hamishwillee:
@infojunkie Thanks so much! Could you further characterise what local synth does vs the full implementation - It "sounds" like:
- the dummy implementation defines a single MIDI output - which is a local synth that outputs to the computer audio system.
- What about the inputs to WebMIDI? My "assumption" is that the dummy implementation can't hook into MIDI input hardware - but what can it do ?

infojunkie:
The "local synth" implementation does not deal with the Web MIDI API at all. It uses Web Audio to produce the sound. So to be precise, "local synth" is not a MIDI output. The reason It works on FF60 is because FF has supported Web Audio - a completely different subsystem - for a long time.

Also, I should explain that this test page was not created as a testbed for FF Web MIDI support. It's simply a demo for various capabilities of the Web audio stack, including Web Audio, Web MIDI, music engraving, etc. I would NOT recommend mentioning this page in offical FF documentation - among other things, the code is purely of demo quality and it is outdated by now. That's why I was suggesting creating a Web MIDI-specific test page - maybe even within MDN if this makes sense.

ddbeck:
OK, after a bit of discussion today, Hamish and I decided to either a) not acknowledge the pre-Firefox 97 implementation or b) acknowledge it in a note (as in, "Before Firefox 97, an incomplete implementation was available behind this flag. See bug whatever for details"), but not as a support statement. I'll leave it to Hamish to decide whether a or b is preferable, based on how many places would need this note.

hamishwillee:
Thanks @infojunkie for the information. 
@ddbeck I have decided to go with option "a" - not talking about the dummy implementation. That means the current docs here are correct for FF97. The "desire" is for these to go into FF98 but I will do that as a separate update if it happens.

hamishwillee:
This should be good to go @ddbeck

ddbeck (approved):
Looks good. Thank you! 🎉

PR #14876, opened 2022-02-08
Ff98 create image bitmap options

hamishwillee:
FF98 adds support for [createImageBitmap()](https://developer.mozilla.org/en-US/docs/Web/API/createImageBitmap#parameters) parameter `option` properties:  `colorSpaceConversion`, `resizeWidth` and `resizeHeight` in https://bugzilla.mozilla.org/show_bug.cgi?id=1748868 and https://bugzilla.mozilla.org/show_bug.cgi?id=1733559

As `resizeQuality` is not supported by FF but is by Chrome, I've removed the subfeature `resizeWidth_resizeHeight_resizeQuality`, spitting it into three options subfeatures - e.g. `options_resizeHeight_parameter`. I also added  options subfeatures subfeatures for colorSpaceConversion`, `resizeWidth` and `resizeHeight`.

The remaining options are `imageOrientation` and `premultiplyAlpha`. These were added to the initial options of both FF and chromium. In theory users should assume this, and these would not need an entry. However in this case I have added one for both because clearly it is quite possible for all of these to be added in different versions, and otherwise this is not obvious. Should I delete these two?

This is part of https://github.com/mdn/content/issues/12577


ddbeck (changes requested):
My only suggestion here is structural: I can only find one other case of `options` being represented as a feature with its members as subfeatures of `options`. That is:

* method
  * options
    * member 1
    * member 2

The more usual thing is that the options get flattened:

* method
  * options member 1
  * options member 2

Since you've expanded this to represent all of the options, I think that's the thing we ought to do here too. I'd suggest promoting the new features a level, then deleting the vestigial `options_parameter` feature.

hamishwillee:
Thanks. Done!

ddbeck (approved):
Looks good. Thank you! 🎉

PR #14877, opened 2022-02-08
FF98 HTMLElement.outerText - add support and mark as standard

hamishwillee:
[`HTMLElement.outerText`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/outerText) became part of the spec in https://github.com/whatwg/html/pull/6653 (Sep. 2021) and can now be found here: https://html.spec.whatwg.org/multipage/dom.html#the-innertext-idl-attribute.

Support was added in FF98 in https://bugzilla.mozilla.org/show_bug.cgi?id=1709790

This adds the FF support statements, adds spec-url links, and adds it as being on the standard track.

### DIFF
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -2552,6 +2552,7 @@
       "outerText": {
         "__compat": {
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLElement/outerText",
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Looks good. Thank you! 🎉

PR #14891, opened 2022-02-08
Fix spec_url of "javascript.operators.right_shift"

generalmimon:
#### Summary

The specification URL was linking to the section of the bitwise OR `|` operator,
apparently due to a copy-paste error. The unsigned right shift is correct, see
https://github.com/mdn/browser-compat-data/blob/6f71796dc415c5562e857be2312cc79ed6335422/javascript/operators/unsigned_right_shift.json#L8

I tested the URL https://tc39.es/ecma262/multipage/ecmascript-language-expressions.html#sec-signed-right-shift-operator that I'm linking, it works.

#### Related issues

I didn't find any - no one seems to have noticed yet.


Elchi3 (approved):
Nice catch! Thanks! 👍
Welcome to mdn-browser-compat-data 🎉

PR #14926, opened 2022-02-10
Add yarn.lock to .gitignore

hamishwillee:
This adds `yarn.lock` to the `.gitignore` so that it doesn't accidentally get included in commits (yarn works, but is not the supported package manager). As discussed @ddbeck


ddbeck (approved):
Thank you! 🎉

PR #14963, opened 2022-02-14
Adapt AudioScheduledSourceNode ended event to new structure

queengooborg:
This PR adapts the `ended` event of the AudioScheduledSourceNode API to the new guidelines.  Part of work for #14578.


queengooborg:
The redirect has actually already been performed!

Elchi3 (approved):
Thanks, I see it now :)

PR #14981, opened 2022-02-14
FF98 registerProtcolHandler() supports ftp, sftp, ftps

hamishwillee:
FF98 adds support for new protocols in [navigator.registerProtcolHandler()](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/registerProtocolHandler): `ftp`, `sftp`, `ftps` - https://bugzilla.mozilla.org/show_bug.cgi?id=1705202

This adds information about the new addition as a NOTE. @ddbeck not sure if this is right approach - please see note inline

### DIFF
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -4260,7 +4260,7 @@
             "status": {
               "experimental": false,
               "standard_track": true,
-              "deprecated": false
+              "deprecated": true
             }
           }
         }
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -3022,6 +3022,54 @@
             "deprecated": false
           }
         },
+        "scheme_parameter_bitcoin": {
+          "__compat": {
+            "description": "<code>scheme</code> parameter supports <code>bitcoin</code>",
+            "support": {
+              "chrome": {
+                "version_added": "28"
+              },
+              "chrome_android": {
+                "version_added": false
+              },
+              "edge": {
+                "version_added": "79"
+              },
+              "firefox": {
+                "version_added": "2"
+              },
+              "firefox_android": {
+                "version_added": "4"
+              },
+              "ie": {
+                "version_added": false
+              },
+              "opera": {
+                "version_added": "15"
+              },
+              "opera_android": {
+                "version_added": false
+              },
+              "safari": {
+                "version_added": false
+              },
+              "safari_ios": {
+                "version_added": false
+              },
+              "samsunginternet_android": {
+                "version_added": false
+              },
+              "webview_android": {
+                "version_added": false
+              }
+            },
+            "status": {
+              "experimental": false,
+              "standard_track": true,
+              "deprecated": false
+            }
+          }
+        },
         "scheme_parameter_ftp": {
           "__compat": {
             "description": "<code>scheme</code> parameter supports <code>ftp</code>",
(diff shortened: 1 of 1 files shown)


api/Navigator.json
```diff
@@ -2991,10 +2991,12 @@
               ]
             },
             "firefox": {
-              "version_added": "2"
+              "version_added": "2",
+              "notes": "Firefox 98 adds support for new safelisted schemes: <code>ftp</code>, <code>sftp</code>, <code>ftps</code>. All other safelisted schemes listed in the specification are supported, as is the custom scheme protocol format."
```
hamishwillee, line 2995: FF tracks the spec - it supports all [safelisted schemes](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/registerProtocolHandler#permitted_schemes) and the standard format of the custom schemes as defined in the spec. The schemes listed here were just added to the spec, and this reflects the addition to FF in FF98

I added a note because that follows the pattern of the other browsers - which use notes to indicate the small subset of the scheme that they support, if any. 

The other way to do this is to create subfeatures for all the schemes and for the title attribute, and indicate precisely which schemes are supported. I assume this was not done originally because there are 21 schemes. However that seems more accurate/reliable. 
@ddbeck Should I move to that approach or do you see it as "not worth the effort"

ddbeck: You're right. That approach would be much more useful to readers: knowing which browsers support the one scheme you want to use seems like an obvious use case. Would you mind breaking them out?

api/Navigator.json
```diff
@@ -2974,21 +2974,13 @@
           "support": {
             "chrome": {
               "version_added": "13",
-              "notes": [
-                "Allowed schemes include <code>mailto</code>, <code>mms</code>, <code>nntp</code>, <code>rtsp</code>, and <code>webcal</code>. Custom protocols must be prefixed with <code>web+</code>.",
```
hamishwillee, line 2978: Note, 
- schemes now captured as subfeatures. Though might need to add rtsp back if you say (this note indicated supported from v13, but tests show not supported in v15)
- Custom protocol note not needed - that is a spec requirement so this doesn't indicate an incompatibility.

Sammygirl30 (approved):

ddbeck:
Great, thank you, Hamish. On your two questions:

> Chrome states support for `rtsp` from v13. But this does not appear to be supported in  v15).

If this is no longer supported now, we can safely ignore it (it would've fallen out due to the irrelevance guideline).

> Chrome added a few other non-standard cases in https://chromestatus.com/feature/4776602869170176 - do we ignore, or add and mark as non-standard (that's my inclination).

Exactly right. Add and mark as non-standard.

api/Navigator.json
```diff
@@ -3039,16 +3030,1216 @@
...
+              },
+              "samsunginternet_android": {
+                "version_added": false
+              },
+              "webview_android": {
+                "version_added": false
+              }
+            },
+            "status": {
+              "experimental": false,
```
ddbeck, line 3109: The single implementations are going to be experimental. Would you mind double checking each of these?

```suggestion
              "experimental": true,
```

api/Navigator.json
```diff
@@ -3039,16 +3030,1216 @@
...
```
ddbeck, line 3157:
```suggestion
              "experimental": true,
```

api/Navigator.json
```diff
@@ -3039,16 +3030,1216 @@
...
+              "standard_track": true,
+              "deprecated": false
+            }
+          }
+        },
+        "title_parameter_required": {
+          "__compat": {
+            "description": "Obsolete <code>title</code> parameter required",
```
ddbeck, line 4221:
```suggestion
            "description": "<code>title</code> parameter required",
```

Also, the "deprecated" status below should be toggled.

ddbeck (changes requested):
A few little things to fix up.

api/Navigator.json
```diff
@@ -3039,16 +3030,1216 @@
...
+              },
+              "samsunginternet_android": {
+                "version_added": false
+              },
+              "webview_android": {
+                "version_added": false
+              }
+            },
+            "status": {
+              "experimental": false,
```
hamishwillee, line 3733:
```suggestion
              "experimental": true,
```

api/Navigator.json
hamishwillee: Cheers. You got all except one. Fixed.

api/Navigator.json
```diff
@@ -3072,6 +3063,1638 @@
               "deprecated": false
             }
           }
+        },
+        "scheme_parameter_cabal": {
```
hamishwillee, line 3067: FYI @ddbeck This is the newly added non-standard chromium extension for "cabal" added in https://chromestatus.com/feature/4776602869170176 - just added to ease review - these are all the same versions, only the name of the subfeature and the description are changed.

hamishwillee:
Should be good to go now thanks @ddbeck

ddbeck (approved):
Looks good. Thank you! 🎉

PR #14991, opened 2022-02-15
Deno 1.19 is released

lucacasonato:
We're not actually releasing until Thursday, but the new features are pretty much locked in now.


ddbeck (approved):
Thank you! (If there are significant changes between now and release—or a delay—I trust you'll be back to make corrections.) 🎉

PR #15000, opened 2022-02-16
Add missing StyleMedia API

queengooborg:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing `StyleMedia` API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v5.0.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/StyleMedia

### DIFF
--- a/api/StyleMedia.json
+++ b/api/StyleMedia.json
@@ -77,7 +77,7 @@
               "version_added": "5"
             },
             "safari_ios": {
-              "version_added": "5"
+              "version_added": "4"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
@@ -124,7 +124,7 @@
               "version_added": "5"
             },
             "safari_ios": {
-              "version_added": "5"
+              "version_added": "4"
             },
             "samsunginternet_android": {
               "version_added": "1.0"
--- a/api/StyleMedia.json
+++ b/api/StyleMedia.json
@@ -31,7 +31,7 @@
             "version_added": "5"
           },
           "safari_ios": {
-            "version_added": "5"
+            "version_added": "4"
           },
           "samsunginternet_android": {
             "version_added": "1.0"
--- a/api/StyleMedia.json
+++ b/api/StyleMedia.json
@@ -42,8 +42,8 @@
         },
         "status": {
           "experimental": false,
-          "standard_track": true,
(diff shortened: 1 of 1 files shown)


api/StyleMedia.json
```diff
@@ -0,0 +1,145 @@
...
+              "version_added": "1.0"
+            },
+            "webview_android": {
+              "version_added": "≤37"
+            }
+          },
+          "status": {
+            "experimental": false,
+            "standard_track": true,
+            "deprecated": false
```
gsnedders, line 92: Not standards track, is deprecated, no?

api/StyleMedia.json
```diff
@@ -0,0 +1,145 @@
...
```
gsnedders, line 139: ditto

api/StyleMedia.json
```diff
@@ -0,0 +1,145 @@
...
+            "version_added": "15"
+          },
+          "opera_android": {
+            "version_added": "14"
+          },
+          "safari": {
+            "version_added": "5"
+          },
+          "safari_ios": {
+            "version_added": "5"
```
foolip, line 34: This does not match https://github.com/mdn/browser-compat-data/blob/main/api/Window.json#L9299-L9345

Which is correct?

foolip (changes requested):
Can you sync with api.Window.styleMedia? In particular, the statuses of that existing feature need updating.

api/StyleMedia.json
queengooborg: Fixed!  "4" is correct (based on a quick test in BrowserStack).

api/StyleMedia.json
```diff
@@ -0,0 +1,145 @@
...
+            "version_added": "15"
+          },
+          "opera_android": {
+            "version_added": "14"
+          },
+          "safari": {
+            "version_added": "5"
+          },
+          "safari_ios": {
+            "version_added": "4"
```
foolip, line 34: Should the subfeatures also be 4?

foolip (changes requested):

foolip (approved):

PR #15020, opened 2022-02-17
Fix structure of `dump()` data

ddbeck:
#### Summary

Show unflagged existence of `dump()` (i.e., that `"dump" in window` is `true`), which is implied by the existing notes.

#### Related issues

Post-merge suggestions for https://github.com/mdn/browser-compat-data/pull/15014 by @wbamberg.

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -1880,28 +1880,40 @@
             "edge": {
               "version_added": false
             },
-            "firefox": {
-              "version_added": "1.5",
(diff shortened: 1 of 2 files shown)


Elchi3 (approved):
Thanks Daniel, this makes sense to me!

(and if we ever lint for irrelevant features behind a flag only, this will not fall into that category since it is exposed even if only partially working)

PR #15045, opened 2022-02-19
Fix incorrectly nested attributes of <output>

alattalatta:
#### Summary
image

This isn't right. There are `form` and `name` attributes for `<output>`, but no `for.form` and `form.name` something.


Elchi3 (approved):
Nice catch! 👍

PR #15046, opened 2022-02-19
Adapt Worker/Message API to new events structure

queengooborg:
This PR adapts the DedicatedWorkerGlobalScope, ServiceWorkerContainer, Window[EventHandlers], and Worker APIs to conform to the new events structure.  Part of work for #14578.


lucalves (approved):

api/ServiceWorkerContainer.json
```diff
@@ -350,115 +350,6 @@
...
-            }
-          },
-          "status": {
-            "experimental": true,
-            "standard_track": true,
-            "deprecated": false
-          }
-        }
-      },
-      "onmessageerror": {
```
ddbeck, line 403: We've got `message_event` but no `messageerror_event`. What's the story here?

Also, it seems like the notes for Safari ought to be represented _somewhere_.

api/WindowEventHandlers.json
```diff
@@ -293,106 +293,6 @@
           }
         }
       },
-      "onmessage": {
```
ddbeck, line 296: Another case where I'm not sure where this data went. What feature does this map to? Or if it doesn't exist, why isn't it added, based on this?

(My hunch here is that this PR is both demixing and consolidating events, but that's rather difficult to understand and verify from the diff alone.)

ddbeck (commented):
A few questions here. Thank you!

api/ServiceWorkerContainer.json
queengooborg: Oops, my bad -- I've added it now!

api/WindowEventHandlers.json
queengooborg: For the `WindowEventHandlers` mixin, they have been merged into `Window`!

PR #15052, opened 2022-02-19
IDBDatabaseException is no longer supported

bershanskiy:
#### Summary

Chrome and Edge no longer support IDBDatabaseException.

Opened the console and tried to get `webkitIDBDatabaseException` and `IDBDatabaseException` but instead saw an error:
> Uncaught ReferenceError: webkitIDBDatabaseException is not defined

#### Related issues

Corresponding content PR: [Remove obsolete IDBDatabaseException](https://github.com/mdn/content/pull/13193)
Very vaguely related content PR: [Remove sync version of IDB API which never existed](https://github.com/mdn/content/pull/13139)


Elchi3 (changes requested):
We don't use `true` values anymore since everything has a version number now :)

https://bugs.chromium.org/p/chromium/issues/detail?id=129766&q=IDBDatabaseException&can=1 hints that it was removed many years ago, so I think this whole file should be removed as well as the content on MDN. It meets the criteria for being irrelevant. https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-features

bershanskiy:
@Elchi3 Ah, that makes sense. I removed the file now.

Elchi3 (approved):

PR #15062, opened 2022-02-20
Fix Chalk expression

bershanskiy:
#### Summary

Fix Chalk expression

Prior to this commit this expression has an extra `}` which causes Chalk to throw:
```
chalk`{red → {bold ${relPath}} - {bold ${browser}} no longer accepts} {bold ${statement.version_removed}} as a value}`,
                                                                    ^
```
Removing this `}` makes chalk print actual error properly. Also, this PR makes lines 71 and 66 consistent.


ddbeck (approved):
Thank you! 🎉

PR #15085, opened 2022-02-22
Adapt api.Document.onreadystatechange to new events structure

queengooborg:
This PR adapts the readystatechange event of the Document API to conform to the new events structure.

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -7350,54 +7350,6 @@
           }
         }
       },
-      "onreadystatechange": {
-        "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/Document/onreadystatechange",
-          "support": {
-            "chrome": {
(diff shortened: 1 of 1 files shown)


ddbeck:
As far as I can tell, there's no corresponding `api.Document.readystatechange_event`. Perhaps you didn't push all the changes to this PR?

queengooborg:
There is -- no modifications were made to it because the data between the event handler and the event match!

Elchi3 (approved):
I see the readystatechange_event, so this seems fine to me 👍

ddbeck:
Ah, for some reason GitHub's code search never turned up that feature (I have no idea why). Grepping turned up the feature. Sorry about that. (Though, for reviewer's convenience, it'd be nice to link to the related features, when they're not part of the diff.)

PR #15159, opened 2022-02-28
Adapt api.HTMLInputElement.onsearch to new events structure

queengooborg:
This PR adapts the search event of the HTMLInputElement API to conform to the new events structure.

Note: there are no MDN pages associated with these events, so there will be no corresponding content PR. Any broken MDN URLs in BCD have been removed.

### DIFF
--- a/api/HTMLInputElement.json
+++ b/api/HTMLInputElement.json
@@ -1468,53 +1468,6 @@
           }
         }
       },
-      "onsearch": {
-        "__compat": {
-          "support": {
-            "chrome": {
-              "version_added": "1"
-            },
-            "chrome_android": {
-              "version_added": "18"
-            },
-            "edge": {
-              "version_added": "79"
-            },
-            "firefox": {
-              "version_added": false
-            },
-            "firefox_android": {
-              "version_added": false
-            },
-            "ie": {
-              "version_added": false
-            },
-            "opera": {
-              "version_added": "15"
-            },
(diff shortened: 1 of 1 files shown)


foolip:
This just removes the entry but the description doesn't say so. Was there a mistake?

queengooborg:
It wasn't a mistake, no -- there was simply nothing that needed to be added to the existing `search_event` member!

foolip:
I see, in that case it would be helpful to mention that in the description.

api/HTMLInputElement.json
```diff
@@ -1468,53 +1468,6 @@
...
-              "version_added": false
-            },
-            "opera": {
-              "version_added": "15"
-            },
-            "opera_android": {
-              "version_added": "14"
-            },
-            "safari": {
-              "version_added": "1.3"
```
foolip, line 1499: This doesn't match the version in the "search_event" entry that remains. Can you check which is correct?

queengooborg: Ack, this one slipped through -- the "1.3" came from tests I did a while back, and the "2" came from https://github.com/mdn/browser-compat-data/pull/11755#issuecomment-905677809.  I'm going to stick with "2" on this one for the same reasons.

foolip: Alright, let's stick with Safari 2!

foolip (approved):

PR #15170, opened 2022-02-28
Adapt MediaStream API to new events structure

queengooborg:
This PR adapts the MediaStream API to conform to the new events structure.

### DIFF
--- a/api/MediaStream.json
+++ b/api/MediaStream.json
@@ -184,7 +184,8 @@
       },
       "active": {
         "__compat": {
-          "description": "<code>active</code> event",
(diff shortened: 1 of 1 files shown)


Elchi3 (changes requested):
It is a bit sad the CI didn't fail here. It looks like there are duplicate entries.

Elchi3 (changes requested):
The diff still shows changes to the `active` property and that seems unrelated to the event work. Can you explain why you made those changes?

Elchi3 (approved):

PR #15185, opened 2022-02-28
Adapt PresentationConnection API to new events structure

queengooborg:
This PR adapts the PresentationConnection API to conform to the new events structure.

Note: there are no MDN pages associated with this event, so there will be no corresponding content PR. Any broken MDN URLs in BCD have been removed.

### DIFF
--- a/api/PresentationConnection.json
+++ b/api/PresentationConnection.json
@@ -194,6 +194,134 @@
           }
         }
       },
+      "close_event": {
+        "__compat": {
+          "description": "<code>close</code> event",
+          "support": {
+            "chrome": {
+              "version_added": "50"
+            },
+            "chrome_android": {
+              "version_added": "50"
+            },
+            "edge": {
+              "version_added": "79"
+            },
+            "firefox": {
+              "version_added": "51",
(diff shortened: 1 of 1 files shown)


api/PresentationConnection.json
```diff
@@ -194,16 +194,15 @@
...
-      "id": {
+      "close_event": {
         "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/PresentationConnection/id",
-          "spec_url": "https://w3c.github.io/presentation-api/#dom-presentationconnection-id",
+          "description": "<code>close</code> event",
           "support": {
             "chrome": {
-              "version_added": "47"
+              "version_added": "50"
```
foolip, line 202: This doesn't match the parent interface and made me suspect that on* properties were missing while the events were actually supported, but per https://source.chromium.org/chromium/chromium/src/+/5458a1b5f82ef7a0d01fca7c7352cd7fe7ec5d9a the events and the properties were added together. `PresentationConnectionCloseEvent` is also marked as Chrome 50.

foolip (approved):

PR #15195, opened 2022-02-28
Adapt SVGAnimationElement API to new events structure

queengooborg:
This PR adapts the SVGAnimationElement API to conform to the new events structure.

Notice: since the event is named `xxxEvent` when using `addEventListener()`, the merge on this one is a special case.  I chose to treat the events as if they're simply called `begin` or `end` rather than `beginEvent` or `endEvent`, and added a note stating that the event is called the latter when using `addEventListener()`.  Please let me know if this seems like the right way to go about this before I proceed with the content updates!

### DIFF
--- a/api/SVGAnimationElement.json
+++ b/api/SVGAnimationElement.json
@@ -48,6 +48,56 @@
           "deprecated": false
         }
       },
+      "begin_event": {
+        "__compat": {
+          "description": "<code>begin</code> event",
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/SVGAnimationElement/begin_event",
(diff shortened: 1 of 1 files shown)


ddbeck:
My take on this is that we should use the name you would use in `addEventListener()`. As written, the PR treats the `on` property as the canonical name for the event, which runs counter the spec. And part of the reason to restructure events is to de-emphasize those properties. Instead, I think we we ought to go the other way around: use `beginEvent` and use the notes to flag the name divergence for the `on` handler (and its availability).

Elchi3 (approved):
So strange!
Agree with Daniel and the notes you've added look good to me.

PR #15197, opened 2022-03-01
Adapt api.ScreenOrientation.onchange to new events structure

queengooborg:
This PR adapts the change event of the ScreenOrientation API to conform to the new events structure.


queengooborg:
Removing the "needs content update" label because the related page is a stub and is getting removed.

PR #15230, opened 2022-03-02
Adapt api.Document.onpointerlock[change/error] to new events structure

queengooborg:
This PR adapts the pointerlockchange and pointerlockerror event of the Document API to conform to the new events structure.

The data for Chromium was updated to adhere to the event handler data's changes in #6836.  Prefixed data was left alone.

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -7226,130 +7226,6 @@
           }
         }
       },
-      "onpointerlockchange": {
-        "__compat": {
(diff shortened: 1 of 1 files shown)


queengooborg:
Removing the "needs content update" label because there's only event pages and no event handler pages, so content updates may occur independently.

Elchi3 (approved):

PR #15239, opened 2022-03-02
Adapt api.Window.onuserproximity to new events structure

queengooborg:
This PR adapts the userproximity event of the Window API to conform to the new events structure.

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -4513,82 +4513,6 @@
           }
         }
       },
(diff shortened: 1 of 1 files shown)


queengooborg:
Turns out, there's no MDN page to update, so no content PR is needed.

Elchi3 (approved):
This can hopefully be removed as irrelevant at some point.

PR #15253, opened 2022-03-03
Add data for `:local-link` css selector

### LINKED ISSUE #13029: :local-link page missing Specifications, Browser compatibility
There is no Specifications, Browser compatibility on `:local-link` page.

MDN page: 
https://developer.mozilla.org/en-US/docs/Web/CSS/:local-link

Info: CSS Selector Level 4 » Local link pseudo-class
https://css4-selectors.com/selector/css4/local-link-pseudo-class/

binyamin:
#### Summary

Added spec URL and  browser data for the `:local-link` CSS selector.

I'm not aware of any applicable tests. Although, I did use the JSON schema when adding the data.

- [No WebKit support](https://webkit.org/status/#specification-css-selectors-level-4)
- [Open Firefox bug](https://bugzil.la/1654861)
- [Relevant Chromium bug](https://bugs.chromium.org/p/chromium/issues/detail?id=747818)
- [Edge status page](https://developer.microsoft.com/en-us/microsoft-edge/status/cssselectorslevel4/)
- For Internet Explorer, I put `version_added` as `null`. Though other browsers don't supports this feature, so I can take a guess....

#### Related issues

Closes #13029

### DIFF
--- a/css/selectors/local-link.json
+++ b/css/selectors/local-link.json
@@ -1,0 +1,58 @@
+{
+  "css": {
+    "selectors": {
+      "local-link": {
+        "__compat": {
+          "description": "<code>:local-link</code>",
+          "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/:local-link",
+          "spec_url": "https://drafts.csswg.org/selectors/#local-link-pseudo",
+          "support": {
+            "safari": {
+              "version_added": false
+            },
+            "safari_ios": {
+              "version_added": false
+            },
+            "firefox": {
+              "version_added": false,
+              "notes": "See <a href='https://bugzil.la/1654861'>bug 1654861</a>."
+            },
+            "firefox_android": {
+              "version_added": false,
+              "notes": "See <a href='https://bugzil.la/1654861'>bug 1654861</a>."
+            },
+            "edge": {
+              "version_added": false
+            },
+            "ie": {
+              "version_added": null
+            },
+            "chrome": {
+              "version_added": false
+            },
+            "opera": {
+              "version_added": false
+            },
+            "chrome_android": {
+              "version_added": false
+            },
+            "opera_android": {
+              "version_added": false
+            },
+            "webview_android": {
+              "version_added": false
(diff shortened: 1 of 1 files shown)


queengooborg:
Thanks for your PR, and welcome to BCD!  This is all looking good to me, however we're currently working on deciding whether we want to allow for all-false features (AKA, features with no implementation in any stable browser).  I'm going to mark this as "not ready" in the meantime.

queengooborg:
> If there was a way to separate the specification from the compatibility data, that might solve the problem.

There is actually!  On the content pages, you can add a `spec-url: <url>` in the page metadata at the top of the Markdown files.  This allows for linking to a specification in cases where there is no BCD.  That should solve the page problem until we come to a conclusion on all-false features!

queengooborg:
Hey again, thank you for your patience on this! We have decided not to include features that have no compatibility in BCD. A spec may implement a feature, but then rename or remove it before any browsers implement it, thus leaving a leftover statement in BCD.

Would you be down to research if any browsers support this feature now? If not, I'd be happy to look into it and follow up!

binyamin:
@queengooborg No browsers support this. I _did_ add the spec URL in https://github.com/mdn/content/pull/17445.

queengooborg:
Thanks for following up on this, @binyamin!  In that case, I'm going to go ahead and close this issue -- happy to add this in once a browser has support, or at the very least, an open bug describing intent to implement!

binyamin:
@queengooborg See the top comment. Both Firefox and Chromium have open bugs.

queengooborg:
I thought I had replied to this thread, my bad!

From what it looks like for both Chrome's and Firefox's bugs, they are both "unconfirmed" and there have been no signals from browser engineers in regards to implementing the feature.  What I mean by "an open bug describing intent to implement" is a bug where a browser engineer states they will be implementing the feature or are in the process of doing so.  A bug that has been opened but has no signs of positive engagement isn't quite enough to add a feature to BCD.

PR #15254, opened 2022-03-03
Add note about safari's event restrictions

rayfoss:
#### Summary
Safari will not allow calling from an HTMLSelectElement's change event, Firefox and Chrome do allow it.

- https://webkit.org/blog/10247/new-webkit-features-in-safari-13-1/

Testing within an iFrame is disallowed by all vendors, hence making a straight forward reproduction is difficult.

```ts
 const input = document.querySelector(
    '#myForm select[name="turtles"]'
  ) as HTMLSelectElement
  if (input === null) throw new Error('input element missing')

  input.addEventListener('change', (event: any) => {
    if (typeof event.target.value !== 'string')
      throw new error('Selection is not a string')
  navigator.clipboard
    .writeText('Tuhtles')
    .then(() => {
      console.log('wrote to clipboard')
    })
    .catch((e) => {
      console.log('Writing to clipboard failed.')
    })
  })
```

#### Related issues

- https://developer.apple.com/forums/thread/691873

### DIFF
--- a/api/Clipboard.json
+++ b/api/Clipboard.json
@@ -396,7 +396,7 @@
             },
             "safari": {
               "version_added": "13.1",
-              "notes": "Must be called within user gesture event handlers like pointerdown or pointerup"
+              "notes": "Must be called within user gesture event handlers such as <code>pointerdown</code> or <code>pointerup</code>."
             },
             "safari_ios": {
               "version_added": "13.4"
--- a/api/Clipboard.json
+++ b/api/Clipboard.json
@@ -399,7 +399,8 @@
               "notes": "Must be called within user gesture event handlers such as <code>pointerdown</code> or <code>pointerup</code>."
             },
             "safari_ios": {
-              "version_added": "13.4"
+              "version_added": "13.4",
+              "notes": "Must be called within user gesture event handlers such as <code>pointerdown</code> or <code>pointerup</code>."
             },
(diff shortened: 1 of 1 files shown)


api/Clipboard.json
```diff
@@ -395,7 +395,8 @@
               "version_added": "47"
             },
             "safari": {
-              "version_added": "13.1"
+              "version_added": "13.1",
+              "notes": "Must be called within user gesture event handlers like pointerdown or pointerup"
```
ddbeck, line 399: A small bit of copy editing here ("such as" since I think these are included, not just similar to; code format)

```suggestion
              "notes": "Must be called within user gesture event handlers such as <code>pointerdown</code> or <code>pointerup</code>."
```

api/Clipboard.json
```diff
@@ -395,7 +395,8 @@
...
             },
             "safari_ios": {
               "version_added": "13.4"
```
ddbeck, line 402: Could you apply the note here too? It's a near certainty that iOS has a similar user gesture requirement.

ddbeck (changes requested):
Thank you for this PR. I see it's your first to the project—welcome! I have a couple minor suggestions in line comments for your consideration. Thanks again!

ddbeck (approved):
Looks good. Thank you! 🎉

PR #15257, opened 2022-03-04
FF99 WebMIDI API released

hamishwillee:
Web MIDI API ships by preference switch in FF99: https://bugzilla.mozilla.org/show_bug.cgi?id=1752906. 

This PR updates all the features with release version. Also this makes the API no longer experimental. Other docs work tracked in https://github.com/mdn/content/issues/12792#issuecomment-1031105153

### DIFF
--- a/api/MIDIAccess.json
+++ b/api/MIDIAccess.json
@@ -14,16 +14,22 @@
           "edge": {
             "version_added": "79"
           },
-          "firefox": {
-            "version_added": "97",
-            "flags": [
-              {
-                "type": "preference",
-                "name": "dom.webmidi.enabled",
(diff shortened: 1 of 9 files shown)


queengooborg (changes requested):
Thanks for your PR!  I see that Android support is still disabled as suggested by https://bugzilla.mozilla.org/show_bug.cgi?id=1752906#c1.

I noticed that `MIDIInput` was left marked as experimental, can we update that?

hamishwillee:
Thanks @queengooborg - I've fixed up `MIDIInput`. As you say, Android support is still false.

queengooborg (approved):

PR #15272, opened 2022-03-05
Add nodejs for fetch

shisama:
#### Summary

Node.js has started supporting `fetch()` with the `--experimental-fetch` flag since version 17.5.0.
https://nodejs.org/en/blog/release/v17.5.0/

### DIFF
--- a/browsers/nodejs.json
+++ b/browsers/nodejs.json
@@ -366,6 +366,13 @@
           "status": "current",
           "engine": "V8",
           "engine_version": "9.6"
+        },
+        "17.5.0": {
+          "release_date": "2022-02-10",
+          "release_notes": "https://nodejs.org/en/blog/release/v17.5.0/",
+          "status": "current",
+          "engine": "V8",
+          "engine_version": "9.6"
         }
       }
     }
--- a/browsers/nodejs.json
+++ b/browsers/nodejs.json
@@ -370,7 +370,7 @@
         "17.5.0": {
           "release_date": "2022-02-10",
           "release_notes": "https://nodejs.org/en/blog/release/v17.5.0/",
-          "status": "current",
+          "status": "esr",
           "engine": "V8",
           "engine_version": "9.6"
         },
--- a/browsers/nodejs.json
+++ b/browsers/nodejs.json
@@ -373,6 +373,13 @@
           "status": "current",
           "engine": "V8",
           "engine_version": "9.6"
+        },
+        "18.0.0": {
+          "release_date": "2022-04-19",
+          "release_notes": "https://nodejs.org/en/blog/release/v18.0.0/",
+          "status": "current",
+          "engine": "V8",
+          "engine_version": "10.1"
         }
       }
     }
--- a/browsers/nodejs.json
+++ b/browsers/nodejs.json
@@ -363,7 +363,7 @@
         "17.3.0": {
           "release_date": "2021-12-17",
           "release_notes": "https://nodejs.org/en/blog/release/v17.3.0/",
-          "status": "current",
+          "status": "retired",
           "engine": "V8",
           "engine_version": "9.6"
         },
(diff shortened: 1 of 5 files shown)


ddbeck (commented):
Thanks for opening this PR—it's an exciting development!

One thing I'm wondering about is whether the subfeatures of `fetch` are supported, as well as `Headers`, `Request`, and `Response` objects. Are these not yet supported behind the flag?

shisama:
@ddbeck Thanks for your comment. Node.js supported [`Headers`](https://nodejs.org/dist/latest-v17.x/docs/api/globals.html#class-headers), [`Request`](https://nodejs.org/dist/latest-v17.x/docs/api/globals.html#request), and [`Response`](https://nodejs.org/dist/latest-v17.x/docs/api/globals.html#response) objects exposed as global objects. They are compatible with browser apis. It is available to use them with `--experimental-fetch` flag since Node.js v17.5.0 the version of which the `fetch()` was added in Node.js.

ddbeck:
Great. Would you like to update the data in these files as well, so that all of the `fetch`-related data and can be mutually consistent?

- https://github.com/mdn/browser-compat-data/blob/main/api/Headers.json
- https://github.com/mdn/browser-compat-data/blob/main/api/Request.json
- https://github.com/mdn/browser-compat-data/blob/main/api/Response.json

Rexogamer:
These are also now enabled by default [in Node 18](https://nodejs.org/en/blog/release/v18.0.0)

queengooborg:
Hey @shisama, it looks like this PR is still waiting on you to make changes (as well as rebase).  Do you still plan to work on this PR?

shisama:
@queengooborg Sorry. I am going to work this week. Thank you for your mention.

shisama:
@ddbeck @queengooborg 
Sorry for my late work. I updated the data for Headers, Request, and Response.
By the way, Node.js v18 enables them by default so we can use them without `--experimental-fetch` flag.
Should the json files be updated like the following?

```diff
           "nodejs": {
-            "version_added": "17.5.0",
-            "flags": [
-              {
-                "name": "--experimental-fetch",
-                "type": "runtime_flag"
-              }
-            ]
+            "version_added": "18.0.0"
           },
```

queengooborg:
Thanks for updating the PR!  I would say that we should leave the flag data in, but include a second statement for the feature being enabled by default -- in other words, something like this:

```js
"nodejs": [
  {
    "version_added": "18.0.0"
  },
  {
    // ...flag data...
  }
]
```

shisama:
@queengooborg Thanks! I updated the files.

browsers/nodejs.json
```diff
@@ -366,6 +366,20 @@
           "status": "current",
```
queengooborg, line 366:
```suggestion
          "status": "retired",
```

browsers/nodejs.json
```diff
@@ -366,6 +366,20 @@
...
           "engine": "V8",
           "engine_version": "9.6"
+        },
+        "17.5.0": {
+          "release_date": "2022-02-10",
+          "release_notes": "https://nodejs.org/en/blog/release/v17.5.0/",
+          "status": "current",
```
queengooborg, line 373:
```suggestion
          "status": "esr",
```

queengooborg (changes requested):
Thank you!  Just a couple of small changes to the browser data, but otherwise looking good to me!

queengooborg (approved):
Thanks for the quick fixes, LGTM!

shisama:
@queengooborg Thank you. I applied your suggestions.

vsn4ik:
`fetch` also supported (w/ flag) since version 16.15.0.
https://github.com/nodejs/node/releases/tag/v16.15.0

queengooborg:
Thanks for the info, @vsn4ik!  Would you mind opening up a pull request to add this to our data, or an issue?

vsn4ik:
@queengooborg I'll create a pull request today.

PR #15281, opened 2022-03-06
offset-anchor

NiedziolkaMichal:
#### Summary
Chromium browsers don't support offset-anchor, while mobile firefox does.

I tested it on desktor Chrome, mobile Chrome, desktop Edge, desktor Firefox, mobile Firefox, mobile Samsung Browser.
On desktop browsers I checked @support(offset-anchor: auto) and examples in [offset-anchor](https://developer.mozilla.org/en-US/docs/Web/CSS/offset-anchor). On mobile browsers I checked just the examples.

I verified that mobile Firefox works, but I am not sure since which version. I think it's 72, because of [release notes](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/72#css).

#### Related issues

Issue [#1826](https://github.com/mdn/interactive-examples/issues/1826)

### DIFF
--- a/css/properties/offset-anchor.json
+++ b/css/properties/offset-anchor.json
@@ -32,7 +32,7 @@
               }
             ],
             "firefox_android": {
-              "version_added": "72"
+              "version_added": "79"
             },
             "ie": {
               "version_added": false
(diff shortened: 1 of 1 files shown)


NiedziolkaMichal:
Since 72 was shown to be invalid firefox android version, I have changed it to 79, which seems to be next available number after 72.

ddbeck (approved):
Good discovery! Thank you. 🎉 

(Also, I see that this is one of your first PRs—welcome!)

PR #15292, opened 2022-03-07
FF98 Navigator.pdfViewerEnabled - supported

hamishwillee:
FF98 adds support for `Navigator.pdfViewerEnabled` in https://bugzilla.mozilla.org/show_bug.cgi?id=1720353
Other docs work for this tracked in https://github.com/mdn/content/issues/13370

Also adds subfeature for `Navigator.plugins` - see comment inline "for discussion". When that resolves I will copy the pattern for `Navigator.mimeTypes`.

### DIFF
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -2817,6 +2817,54 @@
             "standard_track": true,
             "deprecated": true
           }
+        },
+        "fixed_return_value": {
+          "__compat": {
+            "description": "Returns hard coded list of plugins or none",
+            "support": {
+              "chrome": {
+                "version_added": "94"
+              },
+              "chrome_android": {
+                "version_added": "94"
+              },
+              "edge": {
+                "version_added": "94"
+              },
+              "firefox": {
+                "version_added": "99"
+              },
+              "firefox_android": {
+                "version_added": "99"
+              },
+              "ie": {
+                "version_added": false
+              },
+              "opera": {
+                "version_added": "80"
+              },
+              "opera_android": {
+                "version_added": "66"
+              },
+              "safari": {
+                "version_added": false
+              },
+              "safari_ios": {
+                "version_added": false
+              },
+              "samsunginternet_android": {
+                "version_added": false
+              },
+              "webview_android": {
+                "version_added": "94"
+              }
+            },
+            "status": {
+              "experimental": true,
+              "standard_track": true,
+              "deprecated": false
+            }
+          }
         }
       },
       "presentation": {
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -2459,6 +2459,62 @@
             "standard_track": true,
             "deprecated": true
           }
+        },
+        "returns_plugin_type": {
+          "__compat": {
+            "description": "Returns mime types from plugins rather than hard coded PDF values",
+            "support": {
+              "chrome": {
+                "version_removed": "94",
+                "version_added": "1"
+              },
+              "chrome_android": {
+                "version_removed": "94",
+                "version_added": "18"
+              },
+              "edge": {
+                "version_removed": "94",
+                "version_added": "12"
+              },
+              "firefox": {
+                "version_removed": "99",
+                "version_added": "1"
+              },
+              "firefox_android": {
+                "version_removed": "99",
+                "version_added": "4"
+              },
+              "ie": {
+                "version_added": "4"
+              },
+              "opera": {
+                "version_removed": "80",
+                "version_added": "≤12.1"
+              },
+              "opera_android": {
+                "version_removed": "66",
+                "version_added": "≤12.1"
+              },
+              "safari": {
+                "version_added": "1"
+              },
+              "safari_ios": {
+                "version_added": "1"
+              },
+              "samsunginternet_android": {
+                "version_added": "1.0"
+              },
+              "webview_android": {
+                "version_removed": "94",
+                "version_added": "1"
+              }
+            },
+            "status": {
+              "experimental": true,
+              "standard_track": true,
+              "deprecated": false
+            }
+          }
         }
       },
       "mozIsLocallyAvailable": {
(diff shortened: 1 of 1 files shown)


api/Navigator.json
```diff
@@ -2816,6 +2817,54 @@
             "standard_track": true,
             "deprecated": true
           }
+        },
+        "fixed_return_value": {
```
hamishwillee, line 2821: @ddbeck As part of this spec change the deprecated `Navigator.plugins` and `Navigator.mimeTypes` properties were changed from returning "real" plugins (including possibly Adobe flash) to returning hard coded values.

Essentially these are both  long deprecated but some code is still using it to determine if inline PDF viewing is supported, by checking for particular mime types being supported and particular mime type names. This is unreliable and exposes fingerprinting information so the spec made two changes:
- add `pdfViewerEnabled` to indicate reliably inline viewing is supported
- make these old properties return a standard set of plugins if PDF viewing is enabled and an empty object if not -  mirroring `pdfViewerEnabled`. The idea being that old sites would continue to work, but actually be more reliable since the set of strings they use are accurately set. Eventually though we hope all will migrate to  `pdfViewerEnabled

Anyway, I have indicated this using a subfeature, currently just for `Navigator.plugins`. Wanted to check if this is the best way. We might alternatively have a subfeature that is expiring instead - ie one that indicates "plugins actually exist" rather than "plugin values are hard coded". Or I could just add a note for all cases indicating the version.

Would appreciate your guidance.

ddbeck: > We might alternatively have a subfeature that is expiring instead - ie one that indicates "plugins actually exist" rather than "plugin values are hard coded". Or I could just add a note for all cases indicating the version.

Yes, I think inverting things is the way to go. Ideally, in a few years, we'll end up in a situation where there hard-coded behavior is the only one that matters, at which point we can get rid of any mention of the historic behavior.

lucalves (approved):

api/Navigator.json
```diff
@@ -2816,6 +2873,62 @@
             "standard_track": true,
             "deprecated": true
           }
+        },
+        "returns_plugins": {
```
hamishwillee, line 2877: This is the "inverted" subfeature indicating when Navigator.plugins returns actual plugins. Following the version_removed it returns hard coded values used to mock the `Navigator.pdfViewerEnabled` property.

api/Navigator.json
```diff
@@ -2459,6 +2459,62 @@
             "standard_track": true,
             "deprecated": true
           }
+        },
+        "returns_plugin_type": {
```
hamishwillee, line 2463: This is the "inverted" subfeature indicating when `Navigator.mimeTypes` returns mime types from installed plugins. Following the version_removed it returns hard coded values used to mock the `Navigator.pdfViewerEnabled` property.

hamishwillee:
Thanks @ddbeck - I've inverted the subfeatures indicating the point of hard coding plugins and mimeTypes. Ready to review again.

sideshowbarker (approved):

sideshowbarker:
oops, I hadn’t meant to merge this before @ddbeck had a chance to re-review

hamishwillee:
Oops. Do I need to somehow revert this?

sideshowbarker:
> Oops. Do I need to somehow revert this?

I guess maybe we can wait to see if @ddbeck had any further catches, and do a follow-up PR with any fixes needed

api/Navigator.json
```diff
@@ -2459,6 +2459,62 @@
...
+              "webview_android": {
+                "version_removed": "94",
+                "version_added": "1"
+              }
+            },
+            "status": {
+              "experimental": true,
+              "standard_track": true,
+              "deprecated": false
+            }
```
ddbeck, line 2516: Status ought to be:

* Experimental: false (weird case of a single implementation, but an old one that's unlikely to change)
* Standard track: false
* Deprecated: true

api/Navigator.json
```diff
@@ -2459,6 +2459,62 @@
             "standard_track": true,
             "deprecated": true
           }
+        },
+        "returns_plugin_type": {
+          "__compat": {
+            "description": "Returns mime types from plugins rather than hard coded PDF values",
```
ddbeck, line 2465:
```suggestion
            "description": "Returns MIME types from plugins rather than hard-coded PDF values",
```

api/Navigator.json
```diff
@@ -2816,6 +2873,62 @@
             "standard_track": true,
             "deprecated": true
           }
+        },
+        "returns_plugins": {
+          "__compat": {
+            "description": "Returns plugins rather than hard coded PDF-plugin values",
```
ddbeck, line 2879:
```suggestion
            "description": "Returns plugins rather than hard-coded PDF plugin values",
```

api/Navigator.json
```diff
@@ -2816,6 +2873,62 @@
...
+              "webview_android": {
+                "version_removed": "94",
+                "version_added": "1"
+              }
+            },
+            "status": {
+              "experimental": true,
+              "standard_track": true,
+              "deprecated": false
+            }
```
ddbeck, line 2930: Status ought to be:

- Experimental: false
- Standard track: false
- Deprecated: true

ddbeck (commented):
Could you fix some issues in a follow up PR? No need to revert, I think; a direct fix is good.

api/Navigator.json
hamishwillee: Fixed in #15361

api/Navigator.json
hamishwillee: Fixed in #15361

api/Navigator.json
hamishwillee: Fixed in #15361

api/Navigator.json
hamishwillee: Fixed in #15361

hamishwillee:
Thanks, all fixed in #15361

PR #15256, opened 2022-03-04
FF99 Navigator.requestMIDIAccess() - add secure context subfeature

hamishwillee:
FF99 adds secure context to [`Navigator.requestMIDIAccess()`](https://developer.mozilla.org/en-US/docs/Web/API/Navigator/requestMIDIAccess) in https://bugzilla.mozilla.org/show_bug.cgi?id=1757153

All the other browsers  are set to false. Note that this might have been added to [chromium here in M77](https://chromestatus.com/feature/5138066234671104). I am querying, but [indications](https://groups.google.com/a/chromium.org/g/blink-dev/c/_2XZt3yInCI/m/tk-f_vPwAQAJ) seems to be that there might just be a console warning.

Related docs work tracked in https://github.com/mdn/content/issues/12792#issuecomment-1031105153

PR #15301, opened 2022-03-07
Navigator.requestMIDIAccess - chrome secure context info

hamishwillee:
Chromium based browsers also require secure context for `Navigator.requestMIDIAccess`. From https://www.chromium.org/developers/design-documents/web-midi/ it looks like the sysex permission was required from M43 (first release). 
Note that for chrome there was some "non gated" use allowed prior to version 82 as well, but I don't think we need to draw this distinction (see https://chromestatus.com/feature/5138066234671104).

This follows on from #15256

### DIFF
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -4870,13 +4870,13 @@
             "description": "Secure context required",
             "support": {
               "chrome": {
-                "version_added": false
+                "version_added": "43"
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Looks good. Thank you! 🎉

PR #15303, opened 2022-03-08
FF99 NavigationPreloadManager - behind pref and in nightly

hamishwillee:
FF99 now supports [NavigationPreloadManager](https://developer.mozilla.org/en-US/docs/Web/API/NavigationPreloadManager) in nightly - see https://bugzilla.mozilla.org/show_bug.cgi?id=1750515

FF also has this behind a preference. The interface and preference were [added in FF91](https://bugzilla.mozilla.org/show_bug.cgi?id=1564235) but it looks like it only became useful in FF97 with fetch integration: https://bugzilla.mozilla.org/show_bug.cgi?id=1577346

I have therefore added for desktop a preference introduced in 97 and a preview release. Nothing appears in android, so I set that to false (i.e no pref settable, and there is no nightly build of android).

Related docs work can be tracked in https://github.com/mdn/content/issues/13367

### DIFF
--- a/api/NavigationPreloadManager.json
+++ b/api/NavigationPreloadManager.json
@@ -14,13 +14,23 @@
           "edge": {
             "version_added": "18"
           },
-          "firefox": {
-            "version_added": false,
-            "notes": "Implementation tracked in <a href='https://bugzil.la/1290958'>bug 1290958</a>."
-          },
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Thank you! 🎉

PR #15305, opened 2022-03-08
Fix spec URL of repeating-conic-gradient

NiedziolkaMichal:
#### Summary

Current URL points to draft that doesn't have conic gradients. I fixed link to be the same as conic-gradient().


queengooborg (approved):
Thanks for the quick fix, LGTM!

PR #2103, opened 2018-05-21
Add compat data for Screen

bunnybooboo:
https://developer.mozilla.org/en-US/docs/Web/API/Screen

PR #15320, opened 2022-03-10
screen.lockOrientation/unlockOrientation don't support all version Blink-based browsers

makotokato:
#### Summary
`screen.lockOrientation` and `screen.unlockOrientation` was prefixed API and both become `screen.orientation.lock` and `screen.orientation.unlock` as standard. IE, old Edge (EdgeHTML) and Gecko support it with prefixed, but other don't.

As long as I check Blink's code, although it isn't implemented, BCD says "supported".

This data is added by https://github.com/mdn/browser-compat-data/pull/2103, but this change is incorrect for these APIs. So we should update it.

(I guess that Chrome added `screen.orientation` by 38, so anyone copies the version by mistake)

Using https://mdn-bcd-collector.appspot.com/tests/api/Screen with https://browserstack.com

### DIFF
--- a/api/Screen.json
+++ b/api/Screen.json
@@ -399,13 +399,14 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Screen/lockOrientation",
           "support": {
             "chrome": {
-              "version_added": "38"
+              "version_added": false
             },
             "chrome_android": {
-              "version_added": "38"
+              "version_added": false
             },
             "edge": {
               "version_added": "12",
+              "version_removed": "79",
               "prefix": "ms"
             },
             "firefox": {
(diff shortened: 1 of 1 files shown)


Destop-Trang1801:
Tải Outlook for iOS<https://aka.ms/o0ukef>
Từ: Daniel D. Beck ***@***.***>
Đã gửi: Thursday, March 10, 2022 10:02:23 PM
Đến: mdn/browser-compat-data ***@***.***>
Cc: Subscribed ***@***.***>
Chủ đề: Re: [mdn/browser-compat-data] screen.lockOrientation/unlockOrientation don't support all version Blink-based browsers (PR #15320)


@ddbeck approved this pull request.

Nice catch. Thank you! 🎉

PR #15332, opened 2022-03-11
Remove incorrect version for `font-variation-settings`

dlibby-:
#### Summary
Removes reported support for font-variation-settings in Chromium-based browsers.

In Chromium browsers, `font-variation-settings` is supported as a normal CSS property, but not within `@font-face` at the moment

See https://bugs.chromium.org/p/chromium/issues/detail?id=443467.

### DIFF
--- a/css/at-rules/font-face.json
+++ b/css/at-rules/font-face.json
@@ -613,13 +613,13 @@
             "spec_url": "https://drafts.csswg.org/css-fonts/#font-rend-desc",
             "support": {
               "chrome": {
-                "version_added": "62"
+                "version_added": false
               },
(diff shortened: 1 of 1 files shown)


ddbeck (changes requested):
Thanks for this PR, @dlibby-. Would you mind setting Opera and Opera Android to `false` as well? These are also likely affected by the lack of Chromium support.

dlibby-:
@ddbeck - thanks for catching that! Fixed now (sorry for the delay, I was out of office last week).

ddbeck (approved):
Looks good. Thank you! 🎉

PR #15341, opened 2022-03-11
Update background-attachment: fixed support in Firefox

smalllong:
I've tried it on Firefox 2 on macOS, the 'fixed' value is working so it's already being supported in the very early versions. It's also proved here: https://github.com/Fyrd/caniuse/blob/main/features-json/background-attachment.json


queengooborg (approved):
Makes sense to me, LGTM!  👍

Issue #15300, opened 2022-03-07
api.WindowClient.navigate - is not supported in safari


jimmywarting:
#### What information was incorrect, unhelpful, or incomplete?
it isn't supported in safari.

#### What did you expect to see?
x as in "not supported"

#### Did you test this? If so, how?
yep, i tested it in safari Version 15.3 (17612.4.9.1.8) and was faced with an unhandled promise.

Unhandled Promise Rejection: NotSupportedError: windowClient.navigate() is not yet supported

<details>
<summary>MDN page report details</summary>

* Query: `api.WindowClient.navigate`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/WindowClient/navigate
* Report started: 2022-03-07T14:04:04.007Z

</details>

PR #15344, opened 2022-03-11
Fix Safari support for api.WindowClient.navigate

MrBrain295:
#### Summary

Fix: #15300
#### Related issues

#15300

### DIFF
--- a/api/WindowClient.json
+++ b/api/WindowClient.json
@@ -185,7 +185,9 @@
               "notes": "This method exists, but always throws <code>NotSupportedError</code>."
             },
             "safari_ios": {
-              "version_added": "11.3"
+              "version_added": "11.3",
+              "partial_implementation": true,
+              "notes": "This method exists, but always throws <code>NotSupportedError</code>."
             },
             "samsunginternet_android": {
               "version_added": "4.0"
--- a/api/WindowClient.json
+++ b/api/WindowClient.json
@@ -180,7 +180,7 @@
               "version_added": "36"
             },
             "safari": {
-              "version_added": "11.1"
+              "version_added": false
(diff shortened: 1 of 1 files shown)


api/WindowClient.json
```diff
@@ -180,7 +180,7 @@
               "version_added": "36"
             },
             "safari": {
-              "version_added": "11.1"
+              "version_added": false
```
ddbeck, line 183: Because the method exists, we have to use the `partial_implementation` flag instead of marking it as unsupported (see [this guideline](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#non-functional-defined-names-imply-partial_implementation) for background information). How about this instead:

```suggestion
              "version_added": "11.1",
              "partial_implementation": true,
              "notes": "This method exists, but always throws <code>NotSupportedError</code>."
```

api/WindowClient.json
```diff
@@ -180,7 +180,7 @@
...
             },
             "safari_ios": {
               "version_added": "11.3"
```
ddbeck, line 186: I suspect this is also true of Safari for iOS, so we should apply similar changes here:

```suggestion
              "version_added": "11.3",
              "partial_implementation": true,
              "notes": "This method exists, but always throws <code>NotSupportedError</code>."
```

ddbeck (changes requested):
Thank you for this PR. I see it's your first to the project—welcome!

I investigated this and it turns out the story is a little more complicated. The API exists, but it doesn't actually do anything. I've suggested changes in line comments below.

ddbeck (approved):
Thank you! 🎉

PR #15363, opened 2022-03-14
FF99 NetworkInformation API pref disabled on Android 

hamishwillee:
[Network Information API](https://developer.mozilla.org/en-US/docs/Web/API/Network_Information_API) was enabled on Android but hidden behind a pref in desktop. Because it "leaks" too much information that pref was disabled by default on all platforms: https://bugzilla.mozilla.org/show_bug.cgi?id=1637922

Because FF on Android does not expose prefs, this is the same as disabling the feature. This update does that.

Feature was already "experimental" so did not change that. Feature is probably deprecated or being revisited, but there is nothing formal for that, so have not changed the marking.


api/NetworkInformation.json
```diff
@@ -15,10 +15,18 @@
             "version_added": "79"
           },
           "firefox": {
-            "version_added": false
+            "version_added": "31",
```
hamishwillee, line 18: FYI, this was an omission as far as I can tell. I have added versions for desktop behind flag for anything defined in Android.

hamishwillee:
@ddbeck This data indicates that the API is on the standards track, but as pointed out in https://bugzilla.mozilla.org/show_bug.cgi?id=1637922#c22 , this is a WICG Draft Community Group Report https://wicg.github.io/netinfo - so my assumption is that this is not standards track right?

queengooborg:
WICG is a gray area in terms of whether it's considered standard track or not, depending on who you ask.  In BCD, we have treated WICG as if it is a standard track, and WICG spec URLs are [included in w3c/browser-specs](https://github.com/w3c/browser-specs/blob/main/specs.json), so I'd say we should set `standard_track` to `true`!

hamishwillee:
@queengooborg 
OK, so it looks like someone merged in the updated structure for events. I have therefore removed onchange_event and moved the changes into `typechange_event` - is that the right thing to do?

> WICG is a gray area in terms of whether it's considered standard track or not, depending on who you ask. In BCD, we have treated WICG as if it is a standard track, and WICG spec URLs are [included in w3c/browser-specs](https://github.com/w3c/browser-specs/blob/main/specs.json), so I'd say we should set standard_track to true!

Cool. That's what it is except for `typechange_event` (which I did not add) - so my understanding is that if I change that one to also be standard track true, we're good to go?

ddbeck:
> Cool. That's what it is except for `typechange_event` (which I did not add) - so my understanding is that if I change that one to also be standard track true, we're good to go?

I think this is correct. Another way to look at this is that a spec URL implies standard track is `true`.

hamishwillee:
Great, then this should now be ready to merge!

annevk:
Based on what is WICG standards track though? I don't think any implementer considers it more than incubation which is literally part of its name.

ddbeck:
@annevk: The short version is that we've deferred to the [w3c/browser-specs selection criteria](https://github.com/w3c/browser-specs#spec-selection-criteria) to set this value.

The long version is that, for historic reasons, the status property `standard_track` is somewhat misleadingly named. We document the `standard_track` key as "part of an active specification or specification process." In practice, a `true` value means something more like "specified and not overtly proprietary."

Admittedly, a boolean value is a blunt instrument. In the future, we'd like to drop the `standard_track` property entirely and use the specification URLs as a richer replacement. As a transitional convention, we encourage contributors to BCD to set `standard_track` to `true` whenever there's a spec URL, which we validate against w3c/browser-specs. This means that some pre-standardization specifications are marked as "standard track" even when the story about them is somewhat more complex (though I'm skeptical that web developers generally—the audience for the information contained in BCD—make much of the distinction).

annevk:
Well, but that repository explicitly says "The spec is being developed by a well-known standardization or pre-standardization group." so saying they're all standards track doesn't follow. That's at least a rather big mismatch with how we view these things at Mozilla.

PR #15368, opened 2022-03-14
cascade layers shipped in Chrome 99

rachelandrew:
Cascade layers `@layer` shipped in Chrome 99, I also marked this as experimental: false now it has landed in three browsers.

https://chromestatus.com/feature/5663362835808256


Elchi3 (approved):

captainbrosset:
Microsoft Edge also supports `@layer`. It got it from Chromium and, I need to check, it probably made it to Edge 99 too.
We'll make a PR to update this soon.

PR #15380, opened 2022-03-15
fix(DateTimeFormat): split nodejs entry

caugner:
#### Summary

DateTimeFormat was partially supported in Node.js 0.12 and fully supported in Node.js 13, but this is not reflected in the data.

(_Not tested._)

#### Related issues

Noticed here: https://github.com/mdn/yari/pull/5557#issuecomment-1067812834


caugner:
I agree with the concerns that splitting the entry up gives the false impression that `DateTimeFormat` was not supported at all before 13.0, so I will close this PR. :)

PR #15405, opened 2022-03-16
Delete HTTP headers Cookie2 and Set-Cookie2

bershanskiy:
#### Summary

Delete HTTP headers `Cookie2` and `Set-Cookie2`

These headers are obsolete and have not been used anywhere in a while now. For context, proposed standard [RFC 6265](https://datatracker.ietf.org/doc/html/rfc6265) which obsoleted them was published in April 2011.

#### Related issues

https://github.com/mdn/content/pull/13965


sideshowbarker:
Yeah, as I commented over at https://github.com/mdn/content/pull/13965#issuecomment-1069724728, these are essentially just cruft and it’s hard to imagine anybody having any need for them at all at this point

PR #15422, opened 2022-03-18
Update spec URLs for WebSocket features

sideshowbarker:
The WebSocket section was taken out of the HTML spec and moved to its own spec at https://websockets.spec.whatwg.org/

This change also adds the following previously-missing MDN URLs:

- https://developer.mozilla.org/docs/Web/API/CloseEvent/code
- https://developer.mozilla.org/docs/Web/API/CloseEvent/reason
- https://developer.mozilla.org/docs/Web/API/CloseEvent/wasClean


Elchi3 (approved):
👍

PR #15460, opened 2022-03-21
Make GitHub recognize the repo sources are largely JSON

sideshowbarker:
This change causes GitHub to recognize that the repo sources are currently 99.4% JSON and 0.6% JavaScript.

Otherwise, without this change, the GitHub UI incorrectly shows the repo sources as being 100% JavaScript.

See https://github.com/github/linguist/blob/master/docs/overrides.md#detectable And see the similar change (but for Markdown) merged into mdn/content: https://github.com/mdn/content/pull/14123


Elchi3 (approved):
This is truly awesome! 🥳

PR #15475, opened 2022-03-22
Remove left & right from align-*

NiedziolkaMichal:
#### Summary
Values left & right were removed from properties [align-items](https://developer.mozilla.org/en-US/docs/Web/CSS/align-items), [align-self](https://developer.mozilla.org/en-US/docs/Web/CSS/align-self) & [align-content](https://developer.mozilla.org/en-US/docs/Web/CSS/align-content), because they are not supported and don't apply here.

[Specification ](https://drafts.csswg.org/css-align/#positional-values)defines that those values apply only to "justify-*" properties. On Firefox & Chrome they are not recognized in align-items, align-self & align-content.


queengooborg:
Hey @NiedziolkaMichal, thank you for your PR!  I'd just like to clarify the compatibility story with you.  Have you tested the values in Chrome 93 to see if they had functioned at some point, but have since been removed?  If that's the case, then we should add `version_removed` statements like how Firefox's data is, rather than remove the features as it doesn't quite meet the requirements for the [irrelevant features removal guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-features).  Otherwise, if the data is plain wrong and support was never added in Chrome, then removal seems like the right call!

NiedziolkaMichal:
@queengooborg I did now. They are not recognized by Chrome 93.

queengooborg:
Thank you, I just confirmed the same as well, and removal seems good to me!  I'm marking this as "needs content update" because the live sample on the webpages should be updated as well.

This PR also needs a merge from the latest `main`, could you take care of that real quick?

NiedziolkaMichal:
@queengooborg Both things are done. I made PR [#14861](https://github.com/mdn/content/pull/14861) to remove left & right from examples.

queengooborg (approved):
Thank you, LGTM!  👍

PR #15521, opened 2022-03-25
App badging not supported on WebView.

jpmedley:
#### Summary
The badging APIs are NOT supported on WebView. 

They require an installation method that is not available on that platform. Also, here's the [Chrome Status entry](https://chromestatus.com/feature/6068482055602176). Please note that [not-webview-exposed.txt](https://source.chromium.org/chromium/chromium/src/+/main:android_webview/tools/system_webview_shell/test/data/webexposed/not-webview-exposed.txt) is not correct in this case.

### DIFF
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -768,7 +768,7 @@
               "version_added": "13.0"
             },
             "webview_android": {
-              "version_added": "81"
+              "version_added": false
             }
           },
           "status": {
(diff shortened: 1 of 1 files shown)


jpmedley:
cc: @petele

lucalves (approved):

queengooborg (approved):
It looks like the way this is unexposed on WebView is messing with the mdn-bcd-collector's feature detection.  There's apparently still members on the Navigator API, so `'setAppBadge' in instance` will return true, even though `setAppBadge` is false.  I'll make note of this and update the collector's detection!

PR #15524, opened 2022-03-25
Add revert-layer to bcd table and add spec url

dipikabh:
#### Summary

A new global keyword was introduced in FF97, the `revert-layer` keyword.

- Added the spec url
- Added BCD table for `revert-layer`.
    - Currently, the table only contains information about support in Firefox.
    - Information about support in other browsers and the specific versions is not known at this point, so leaving them as `false` for now.
    - Also, `standard_track` has been kept as `true` for now - hopefully information about support in other browsers will be available soon.

https://bugzilla.mozilla.org/show_bug.cgi?id=1699220
https://drafts.csswg.org/css-cascade-5/#revert-layer

#### Related issues

Content issue tracking this work: https://github.com/mdn/content/issues/13373
PR for the new `revert-layer` page: https://github.com/mdn/content/pull/14287

### DIFF
--- a/css/types/global_keywords.json
+++ b/css/types/global_keywords.json
@@ -209,6 +209,55 @@
             }
           }
         },
+        "revert_layer": {
+          "__compat": {
+            "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/revert-layer",
+            "spec_url": "https://drafts.csswg.org/css-cascade-5/#revert-layer",
+            "support": {
+              "chrome": {
+                "version_added": false
+              },
+              "chrome_android": {
+                "version_added": false
+              },
+              "edge": {
+                "version_added": false
+              },
+              "firefox": {
+                "version_added": "97"
+              },
+              "firefox_android": {
+                "version_added": "97"
+              },
+              "ie": {
+                "version_added": false
+              },
+              "opera": {
+                "version_added": false
+              },
+              "opera_android": {
+                "version_added": false
+              },
+              "safari": {
+                "version_added": false
+              },
+              "safari_ios": {
+                "version_added": false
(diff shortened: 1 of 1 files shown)


css/types/global_keywords.json
```diff
@@ -209,6 +209,55 @@
             }
           }
         },
+        "revert_layer": {
```
queengooborg, line 212: We don't replace hyphens with underscores; this allows us to provide less conversion work for consumers!

```suggestion
        "revert-layer": {
```

queengooborg (changes requested):
Thank you for your PR!  I just have one change to request, but LGTM overall!

css/types/global_keywords.json
```diff
@@ -209,6 +209,55 @@
...
+              },
+              "samsunginternet_android": {
+                "version_added": false
+              },
+              "webview_android": {
+                "version_added": false
+              }
+            },
+            "status": {
+              "experimental": false,
```
lukewarlow, line 255: Should be "experimental" true as only one engine implements it currently (as far as the data is concerned)

queengooborg: Good catch -- yes, this as well!

dipikabh: Thanks, @lukewarlow. I have fixed this and also changed standard_track to false.

I had deliberately kept "standard_track" true and "experimental" false even though I only know about support in FF, hoping that info about support in other browsers will be available soon. But I guess it is better to keep it this way with experimental true and change it when we actually update other browsers in this table. Thanks!

queengooborg (approved):
Thanks for applying changes, LGTM!

PR #15552, opened 2022-03-28
Remove notes linking to polyfills

### LINKED ISSUE #6762: Remove polyfills from BCD?
I remember that there was some [discussion on removing polyfills from MDN](https://discourse.mozilla.org/t/mdn-rfc-001-mdn-wiki-pages-shouldnt-be-a-distributor-of-polyfills/24500), but it doesn't look like there was an issue opened here.  When going through the data, I noticed that there's a few notes that mention polyfills.  However, I'm thinking we should remove them.  While it's nice to have polyfills, I don't think that we should include polyfills in BCD, since they may become old and/or abandoned, or simply redundant as browsers implement support.

queengooborg:
This PR fixes #6762 by removing the remaining notes linking to polyfills.

### DIFF
--- a/webextensions/api/contentScripts.json
+++ b/webextensions/api/contentScripts.json
@@ -7,12 +7,10 @@
             "mdn_url": "https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/API/contentScripts/RegisteredContentScript",
             "support": {
               "chrome": {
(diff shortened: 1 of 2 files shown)


rebloor (approved):
Looks good to me, thanks

rebloor (approved):

PR #15567, opened 2022-03-29
FF100 Remove and deprecate Large-Allocation HTTP header

hamishwillee:
FF100 has removed the non-standard [Large-Allocation](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Large-Allocation) HTTP header in https://bugzilla.mozilla.org/show_bug.cgi?id=1598759. 

I've recorded the removal. As this was the only browser using it, I have also marked the API as deprecated.

### DIFF
--- a/http/headers/large-allocation.json
+++ b/http/headers/large-allocation.json
@@ -15,7 +15,8 @@
               "version_added": false
             },
             "firefox": {
-              "version_added": "53"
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
I think I documented this when it shipped. Well well :)

PR #15568, opened 2022-03-29
Remove all flag data for Opera Android

### LINKED ISSUE #7354: Opera for Android appears to not support flags

queengooborg:
This PR fixes #7354 by replacing all of Opera Android's flag data with `false` statements using the substitution regex below, followed by 
marking Opera as not accepting flags to prevent regressions.  A consistency follow-up is then performed, along with a manual removal of a 
few flag entries too challenging to remove with a regex script.

Regex Used for Mass Replacement: `"opera_android": {\n(\s+)"version_added": (.*),\n\s+(?:"version_removed": (.*),\n\s+)?"flags": 
\[([^\]]*\n)*(\s+)\]\n(\s+)}` 
-> `"opera_android": {\n$1"version_added": false\n$5}`


javascript/statements.json
```diff
@@ -331,27 +331,9 @@
...
             ],
-            "opera_android": [
-              {
-                "version_added": "36"
-              },
-              {
-                "version_removed": "36",
-                "version_added": "29",
-                "notes": "Strict mode is required."
-              },
```
ddbeck, line 342: Dropping this looks suspicious or out of scope for this PR.

javascript/classes.json
```diff
@@ -828,27 +774,9 @@
...
             ],
-            "opera_android": [
-              {
-                "version_added": "36"
-              },
-              {
-                "version_removed": "36",
-                "version_added": "29",
-                "notes": "Strict mode is required."
-              },
```
ddbeck, line 839: Dropping this looks suspicious or out of scope for this PR

javascript/classes.json
```diff
@@ -427,27 +391,9 @@
...
```
ddbeck, line 438: Dropping this looks suspicious or out of scope for this PR

javascript/classes.json
```diff
@@ -266,27 +248,9 @@
...
```
ddbeck, line 277: Dropping this looks suspicious or out of scope for this PR

javascript/classes.json
```diff
@@ -106,27 +106,9 @@
...
           ],
-          "opera_android": [
-            {
-              "version_added": "36"
-            },
-            {
-              "version_removed": "36",
-              "version_added": "29",
-              "notes": "Strict mode is required."
-            },
```
ddbeck, line 117: Dropping this looks suspicious or out of scope for this PR

css/properties/padding-inline.json
```diff
@@ -76,14 +76,7 @@
...
             "opera_android": {
-              "version_added": "48",
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "enable-experimental-web-platform-features",
-                  "value_to_set": "enabled"
-                }
-              ]
+              "version_added": false
```
ddbeck, line 79: Looks like a case of incomplete mirroring 🤨

css/properties/padding-block.json
```diff
@@ -76,14 +76,7 @@
...
             "opera_android": {
-              "version_added": "48",
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "enable-experimental-web-platform-features",
-                  "value_to_set": "enabled"
-                }
-              ]
+              "version_added": false
```
ddbeck, line 79: Looks like another incomplete mirroring

css/properties/margin-inline.json
```diff
@@ -76,14 +76,7 @@
...
             "opera_android": {
-              "version_added": "48",
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "enable-experimental-web-platform-features",
-                  "value_to_set": "enabled"
-                }
-              ]
+              "version_added": false
```
ddbeck, line 79: And another. I'm stopping with this file here

ddbeck (changes requested):
I reviewed a few of these files (JS, HTML, and the CSS files where I've commented) and I'm not sure a mechanical removal is going to be a good approach here. While some of the removals look right, a bunch highlight other issues, particularly bad mirroring and perhaps over-aggressive statement removal (I think that's because you've used a regex, rather than attacking the data structurally, but I'm not certain about that).

I'm not sure what the right next step is here. Maybe the thing to do here is to slice this up more finely (e.g., by folder) and closely read each change, to make sure that you're only doing the removals that get us _closer_ to reality (i.e., showing less support when there's genuinely less support) than merely removing flagged statements. I'm skeptical that a bulk removal is going to produce very good results, unfortunately.

javascript/statements.json
queengooborg: Digging into this some more, I think I see how the data is supposed to be read.  In #6012, this was restructured from a note into multiple segments, however when I was reading through the data, I thought that the compat sections meant that both strict mode _and_ enabling the flags were required, not either-or.  Luckily though, this data falls under the scope of irrelevant flags, so I'll open a separate PR to take care of them for all of Chromium!

queengooborg: Deferred to #15643.

foolip:
@queengooborg can you split into smaller PRs and self-review as suggested by @ddbeck?

In particular if any Opera Android data ends up not matching Chrome Android, that would be good to investigate.

queengooborg:
I've separated the changes that would need even slightly more review into #15997, #15998, #15999, #16000, #16001, and #16002.  The rest of the changes are non-trivial and it wouldn't make much sense to separate into multiple PRs.

foolip:
@queengooborg I've approved all the PRs, can you rebase this after merging them?

PR #15574, opened 2022-03-29
Add specific engine versions for Safari 15.4

queengooborg:
This PR adds the minor and patch semver parts to the engine version for Safari 15.4.  The data comes from Safari itself.


ddbeck (approved):
Thanks!

PR #15576, opened 2022-03-29
Remove revert row from BCD table for all property

dipikabh:
#### Summary

The [BCD table on `all` property page](https://developer.mozilla.org/en-US/docs/Web/CSS/all#browser_compatibility) also contains a row for `revert` value.

The [BCD table for the global revert value](https://developer.mozilla.org/en-US/docs/Web/CSS/revert#browser_compatibility) is already maintained separately in this file: https://github.com/mdn/browser-compat-data/blob/main/css/types/global_keywords.json

This PR removes the `revert` row from `all` property's BCD table to avoid maintaining the same information in two places.

### DIFF
--- a/css/properties/all.json
+++ b/css/properties/all.json
@@ -48,54 +48,6 @@
             "standard_track": true,
             "deprecated": false
           }
-        },
-        "revert": {
-          "__compat": {
-            "description": "<code>revert</code>",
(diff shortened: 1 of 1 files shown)


ddbeck (approved):
Looks good. Thank you! 🎉

PR #15595, opened 2022-03-30
Safari on iOS does not support ::selection

cletusw:
#### Summary
Safari on iOS does not support ::selection

Verified in iOS Simulator - iPhone 13 - iOS 15.2 using https://quirksmode.org/css/selectors/selection.html

Also corroborated by https://caniuse.com/css-selection

### DIFF
--- a/css/selectors/selection.json
+++ b/css/selectors/selection.json
@@ -47,7 +47,7 @@
               "version_added": "1.1"
             },
             "safari_ios": {
-              "version_added": "1"
+              "version_added": false
             },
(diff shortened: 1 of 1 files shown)


queengooborg (approved):
Hi there, thank you for your PR!  The "1" had originally come from mirroring from Safari Desktop, but running that test in multiple versions of Safari iOS (15.3, 15, 10, and 7), and can confirm a lack of support.  This is LGTM!

Welcome to BCD!  🎉

PR #15648, opened 2022-04-04
Drop unneeded “index.html” from WebAssembly Exception Handling spec URLs

sideshowbarker:


Elchi3 (approved):
Ah yes, good catch! 👍

PR #15654, opened 2022-04-04
Remove irrelevant Firefox flag data for Element API

queengooborg:
This PR removes irrelevant flag data for Firefox and Firefox Android for the `Element` API as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-flag-data).

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -4145,36 +4145,12 @@
             "edge": {
               "version_added": "79"
             },
-            "firefox": [
-              {
-                "version_added": "59"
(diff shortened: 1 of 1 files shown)


queengooborg:
Notice: this removes flags for pre-79 releases of Firefox Android.  Fenix 79 was released in July 2020, so this is PR is a little bit early, but I feel that it's close enough to the two year mark.

PR #15656, opened 2022-04-04
Remove irrelevant Firefox flag data for HTMLElement API

queengooborg:
This PR removes irrelevant flag data for Firefox and Firefox Android for the `HTMLElement` API as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-flag-data).

### DIFF
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -2615,36 +2615,12 @@
                 "alternative_name": "mspointercancel"
               }
             ],
-            "firefox": [
-              {
-                "version_added": "59"
-              },
-              {
-                "version_added": "29",
-                "flags": [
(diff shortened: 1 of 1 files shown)


ddbeck:
Does this remove _all_ instances of `dom.w3c_pointer_events.enabled` prefs in the data? That is, is there any way the removal of the flags here could lead to a situation where our data (never mind the presentation) would invite someone to turn off a feature without being able to learn that it affects other features too?

queengooborg:
It does not, you bring up a good point.  I'll open up another PR to take care of the flag as a whole instead of file-by-file!

Issue #5796, opened 2020-03-08
Tracking WebAssembly features


petamoriken:
WebAssembly features implemented in browsers is unknown. Fortunately, [the WebAssembly feature detection library](https://github.com/GoogleChromeLabs/wasm-feature-detect) is available and it can be tracked.

FYI: https://github.com/WebAssembly/proposals/issues/21

gsnedders:
The biggest part of this is we need to decide how to model many of these extensions (i.e., what granularity we want… do we want to have a compat entry for each [instruction](https://webassembly.github.io/spec/core/appendix/index-instructions.html)? or do we want to group these by high-level extensions like "threads"?).

MendyBerger:
@gsnedders think tracking hi level extensions makes the most sense, browsers have for the most part been adding whole new extensions at once, not instruction by instruction.
In the case where a browser does add only parts of an extension it can be marked with `partial_implementation`.

Thoughts?

MendyBerger:
@Queengooborg what do you think? (I was told that you're the one that can make a decision here)

Feel free to assign to me if you agree with my previous comment.

queengooborg:
This sounds great and would certainly be a big help in keeping BCD auto-maintained!  Perhaps this is something we can integrate into the collector project?  https://github.com/foolip/mdn-bcd-collector

MendyBerger:
@queengooborg what about making them part of BCD? Think it would make sense to track these features in here, it's less than 20 features altogether.

PR #15668, opened 2022-04-05
FF100 supports WebAssembly.Tag and .Exception

hamishwillee:
FF100 supports `WebAssembly.Tag` and `WebAssembly.Exception` 
- went to nightly in FF98 behind pref https://bugzilla.mozilla.org/show_bug.cgi?id=1750040
- shipped FF100 https://bugzilla.mozilla.org/show_bug.cgi?id=1759217
- meta bug on feature https://bugzilla.mozilla.org/show_bug.cgi?id=1695715

Very pleased Safari got in first :-) https://github.com/mdn/browser-compat-data/pull/14910

Docs work associated with this can be tracked in https://github.com/mdn/content/issues/14392

### DIFF
--- a/javascript/builtins/webassembly/Exception.json
+++ b/javascript/builtins/webassembly/Exception.json
@@ -19,12 +19,38 @@
               "edge": {
                 "version_added": false
               },
-              "firefox": {
-                "version_added": false
-              },
-              "firefox_android": {
-                "version_added": false
-              },
+              "firefox": [
+                {
+                  "version_added": "100"
+                },
+                {
+                  "version_added": "98",
+                  "version_removed": "100",
+                  "flags": [
+                    {
+                      "type": "preference",
+                      "name": "javascript.options.wasm_exceptions",
+                      "value_to_set": "true"
+                    }
+                  ]
+                }
+              ],
+              "firefox_android": [
+                {
+                  "version_added": "100"
+                },
+                {
+                  "version_added": "98",
+                  "version_removed": "100",
+                  "flags": [
+                    {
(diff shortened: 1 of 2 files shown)


hamishwillee:
@ddbeck I'm still getting my head around WASM, and WASM exception handling, but my understanding is that these are the "JavaScript" interfaces to WASM. 

As an aside, should BCD capture the WASM-internal compatibility?

There is a whole lot more to WASM that is internal to WASM modules. For example, the [exception handling explainer overview](https://github.com/WebAssembly/exception-handling/blob/main/proposals/exception-handling/Exceptions.md) talks about the try, catch, delegate, rethrow etc, which are code you can write inside wasm or which is generated for you. All of this stuff came into firefox in FF100 (by default) and different versions for Safari etc. The question is, how should the presence of that stuff be covered in BCD?

The docs seem to just capture the behaviour of WASM internals - see peers of https://developer.mozilla.org/en-US/docs/WebAssembly/Concepts - without any particular versioning.

ddbeck (approved):

ddbeck:
BCD doesn't cover WebAssembly directly. https://github.com/mdn/browser-compat-data/issues/5796 raises this as a possibility, but it's never been taken up. Mostly, there's been little demand for it (I suspect wasm feature support is most interesting to compiler authors, rather than web developers directly), so we've never really put the work into figuring out how to represent that data.

Before we'd consider covering wasm's features, we'd need to establish that there's reader demand for this stuff. If you wanted to see this move forward, I'd start looking for evidence that web developers are being caught out by wasm compat issues.

PR #13905, opened 2021-12-07
Make it easier to pipe release-pulls to other tools

ddbeck:
#### Summary

Add some more options and logic to `scripts/release-pulls.js` to facilitate more release automation stuff.

#### Test results and supporting details

For example, I can run something like this:

```
gh pr list --search $(node ./scripts/release-pulls.js -q --query-only)
```

to search the PRs in a pending release. Or use `gh`'s `--json` and `--jq` options to extract useful bits of PR metadata for a release.

PR #15674, opened 2022-04-05
Generate release notes

ddbeck:
#### Summary

Add a script to generate release notes.

You can demonstrate this by running `node ./scripts/release-notes.js v4.1.13 v4.1.14` to generate release notes for v4.1.14, or running `node ./scripts/release-notes.js` to see what hypothetical release notes would be for the next release.

There are quite a few changes here; I'll do a self-review to help guide things here.

#### Related issues

This builds on some release automation introduced in https://github.com/mdn/browser-compat-data/pull/13720 and https://github.com/mdn/browser-compat-data/pull/13905 and others.


scripts/diff-features.js
```diff
@@ -4,6 +4,16 @@ const fs = require('fs');
...
+  const results = diff({ ref1, ref2, github });
+
+  if (format === 'markdown') {
+    printMarkdown(results);
+  } else {
+    console.log(JSON.stringify(results, undefined, 2));
+  }
+}
+
+function diff({ ref1, ref2, github }) {
```
ddbeck, line 16: I've extracted the actual `diff` portion here to an exported function, so it can be used in the `release-notes` script.

scripts/diff-features.js
```diff
@@ -68,7 +74,7 @@ function getEnumerationFromGithub(ref) {
     encoding: 'utf-8',
   }).trim();
   const workflowRun = execSync(
-    `gh api /repos/:owner/:repo/actions/workflows/${ENUMERATE_WORKFLOW}/runs?per_page=100 --jq '.workflow_runs[] | select(.head_sha=="${hash}") | .id'`,
+    `gh api /repos/:owner/:repo/actions/workflows/${ENUMERATE_WORKFLOW}/runs?per_page=100 --jq '[.workflow_runs[] | select(.head_sha=="${hash}") | .id] | first'`,
```
ddbeck, line 77: This wasn't required, but it was a bug that slowed testing of the release notes generation.

Sometimes, a commit can have multiple workflow runs (e.g., on merge and again on being tagged) and the original `jq` query assumed there would be only one, which caused this function to fall back to the slower diffing process.

I've fixed it for my convenience. Happy to move this to a separate PR, if desired.

scripts/diff-features.js
```diff
@@ -127,36 +133,38 @@ function printMarkdown({ added, removed }) {
...
-        description: "Don't fetch artifacts from GitHub.",
-      })
-      .example('$0', 'compare HEAD to parent commmit')
-      .example('$0 176d4ed', 'compare 176d4ed to its parent commmit')
-      .example('$0 topic-branch main', 'compare a branch to main');
-  },
-);
-
 if (require.main === module) {
+  const { argv } = yargs.command(
```
ddbeck, line 137: I've moved the yargs stuff into the `require.main == module` gating so that this module could be `required` without clobbering yargs in the `require`-ing module.

scripts/release-notes.js
```diff
@@ -0,0 +1,127 @@
...
+  const searchDetails = {
+    limit: 1000, // As many PRs as GitHub will allow
+    search: `${buildQuery(end, start, false)}`,
+    json: `number,url,mergeCommit`,
+    jq: '[.[] | { mergeCommit: .mergeCommit.oid, number: .number, url: .url }]', // Flatten the structure provided by GitHub
+  };
+  const args = Object.entries(searchDetails)
+    .map(([key, value]) => `--${key}='${value}'`)
+    .join(' ');
+  const command = `gh pr list ${args}`;
```
ddbeck, line 66: So we shell out to GitHub here. Perhaps this ought to use the GitHub API package, but shelling out has some advantages: we don't have to do pagination ourselves and we don't have to deal with auth. Plus, I had already written a shell script to do a lot of this stuff and this was easy to get up and running quickly.

scripts/release-pulls.js
```diff
@@ -1,4 +1,8 @@
-const { getRefDate, releaseYargsBuilder } = require('./release-utils');
+const {
+  buildQuery,
```
ddbeck, line 2: I moved `buildQuery` to `release-utils` to reuse it in `release-notes`.

scripts/release-utils.js
```diff
@@ -4,6 +4,22 @@ function exec(command) {
   return execSync(command, { encoding: 'utf8' }).trim();
 }
 
+function requireGitHubCLI() {
```
ddbeck, line 7: If we're going to shell out to `gh`, I thought it would at least be friendly to look before we leap.

scripts/release-notes.js
```diff
@@ -0,0 +1,126 @@
...
+
+function main(argv) {
+  const { startVersionTag, endVersionTag } = argv;
+
+  requireGitHubCLI();
+
+  const allAdds = [];
+  const allRemoves = [];
+
+  console.error(
```
queengooborg, line 17: I noticed you're using console errors vs. console logs throughout this script, any particular reason why?

queengooborg (commented):
Thanks for this Daniel, this script is looking really wonderful!  I just have one question, but otherwise this is LGTM!

scripts/release-notes.js
ddbeck: I've used `console.log()` for the actual release notes output (see lines 51-53), but all the status messages (and actual errors) are written to stderr to facilitate redirection. For instance, you could run `node ./scripts/release-notes.js | pbcopy` (on macOS) to copy the output directly the clipboard, without cluttering the results with status messages.

queengooborg: Gotcha, that makes sense!

queengooborg (approved):
LGTM! 👍

PR #15709, opened 2022-04-07
Remove support for content-visibility for Safari 15.4

jensimmons:
Safari 15.4 does not support content-visibility. This was an error. Thanks.

You can see that it's _not_ listed in the official release notes:
https://developer.apple.com/documentation/safari-release-notes/safari-15_4-release-notes

I'm opening this PR after talking to WebKit engineers familiar with the matter.

### DIFF
--- a/css/properties/content-visibility.json
+++ b/css/properties/content-visibility.json
@@ -31,10 +31,10 @@
               "version_added": "60"
             },
             "safari": {
-              "version_added": "15.4"
+              "version_added": false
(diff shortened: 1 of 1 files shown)


queengooborg (approved):
Hey Jen, thank you for your PR!  I just did a quick test in Safari 15.4 and can confirm that the property isn't supported.  Welcome to BCD, we're glad to have you here!  🎉

PR #15723, opened 2022-04-07
Remove css.properties.display.xul* from BCD

queengooborg:
This PR removes the irrelevant `xul*` members of the `display` CSS property as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-features).  These features are only available in Firefox UI.


Elchi3 (approved):
Great PR, love to see these go. 🎉

PR #15727, opened 2022-04-07
Remove irrelevant Firefox Android flag data for document CSS @rule

queengooborg:
This PR removes irrelevant flag data for Firefox Android for the `document` CSS @rule as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-flag-data).

### DIFF
--- a/css/at-rules/document.json
+++ b/css/at-rules/document.json
@@ -82,22 +82,10 @@
               "edge": {
                 "version_added": false
               },
-              "firefox": [
-                {
-                  "version_added": "61",
-                  "flags": [
-                    {
-                      "type": "preference",
-                      "name": "layout.css.moz-document.content.enabled",
-                      "value_to_set": "true"
-                    }
-                  ]
-                },
-                {
-                  "version_added": "6",
-                  "version_removed": "61"
-                }
-              ],
+              "firefox": {
+                "version_added": "6",
+                "version_removed": "61"
+              },
               "firefox_android": {
                 "version_added": "6",
                 "version_removed": "61"
--- a/css/at-rules/document.json
+++ b/css/at-rules/document.json
@@ -24,18 +24,6 @@
               },
               {
                 "prefix": "-moz-",
-                "version_added": "61",
-                "flags": [
-                  {
-                    "type": "preference",
-                    "name": "layout.css.moz-document.content.enabled",
-                    "value_to_set": "true"
-                  }
-                ],
-                "notes": "To use values other than an empty <code>url-prefix()</code>, set this browser flag to true."
-              },
-              {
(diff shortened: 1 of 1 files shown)


css/at-rules/document.json
```diff
@@ -34,25 +34,11 @@
...
-                "prefix": "-moz-",
-                "version_added": "61",
-                "flags": [
-                  {
-                    "type": "preference",
-                    "name": "layout.css.moz-document.content.enabled",
-                    "value_to_set": "true"
-                  }
-                ],
-                "notes": "Disabled by default in web pages, except for an empty <code>url-prefix()</code> value, which is supported due to its <a href='https://css-tricks.com/snippets/css/css-hacks-targeting-firefox/'>use in Firefox browser detection</a>. Still supported in user stylesheets."
```
foolip, line 48: This note suggests there's still partial support. Can you investigate and rewrite the note if needed?

foolip (changes requested):

css/at-rules/document.json
```diff
@@ -38,14 +38,8 @@
...
+                "partial_implementation": true,
+                "notes": "Only supports an empty <code>url-prefix()</code> value, which is supported due to its <a href='https://css-tricks.com/snippets/css/css-hacks-targeting-firefox/'>use in Firefox browser detection</a>. Still supported in user stylesheets."
```
foolip, line 42: Are you sure this quirk isn't also on Firefox Android?

queengooborg: These are the notes for Firefox Android actually!

foolip: Oh right, I saw the next chunk in the diff and thought this was desktop.

Looking at the full diff I have another ask: we should capture that there's partial support in desktop too with the same statement and note. I think that it should be possible to remove all flag data from BCD and the result should be what's supported in default configs, but that isn't true if support is hidden in notes of flag entries.

css/at-rules/document.json
```diff
@@ -94,11 +82,25 @@
...
-              "firefox": {
-                "version_added": "6"
-              },
+              "firefox": [
+                {
+                  "version_added": "61",
+                  "flags": [
+                    {
+                      "type": "preference",
+                      "name": "layout.css.moz-document.content.enabled",
```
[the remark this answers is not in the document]
queengooborg: Considering this feature has been behind a flag for so long, it's probably not going to be enabled by default.  Since I'm apparently adding the flag here, I'm going to keep it removed for simplicity.

foolip (approved):

PR #15729, opened 2022-04-07
Remove irrelevant Firefox Android flag data for -moz-context-properties CSS property

queengooborg:
This PR removes irrelevant flag data for Firefox Android for the `-moz-context-properties` CSS property as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-flag-data).

### DIFF
--- a/css/properties/-moz-context-properties.json
+++ b/css/properties/-moz-context-properties.json
@@ -27,14 +27,8 @@
             },
             "firefox_android": {
               "version_added": "55",
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "svg.context-properties.content.enabled",
-                  "value_to_set": "true"
-                }
-              ],
-              "notes": "With the preference set to <code>false</code>, the property still works with SVGs via <code>chrome://</code> or <code>resource://</code> URLs"
+              "partial_implementation": true,
+              "notes": "The property only works with SVGs via <code>chrome://</code> or <code>resource://</code> URLs."
             },
             "ie": {
               "version_added": false


css/properties/-moz-context-properties.json
```diff
@@ -27,14 +27,8 @@
...
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "svg.context-properties.content.enabled",
-                  "value_to_set": "true"
-                }
-              ],
-              "notes": "With the preference set to <code>false</code>, the property still works with SVGs via <code>chrome://</code> or <code>resource://</code> URLs"
+              "partial_implementation": true,
+              "notes": "The property only works with SVGs via <code>chrome://</code> or <code>resource://</code> URLs."
```
foolip, line 31: Why this note and partial implementation? Seems like the feature works in some non-web-exposed contexts despite the flag, but with the flag removed is there anything to say still?

foolip (changes requested):

css/properties/-moz-context-properties.json
queengooborg: I guess it's not helpful to say that a feature is supported only in internal contexts, so we should probably just set this to "false"?

foolip: I think so, yes. But can you search for other comments involving chrome:// or similar and see if there's a lot of them? If there are I think we should remove them all in one PR and have a clear justification, for future reference.

queengooborg: It appears this is the only instance containing `resource://` or `chrome://` (besides references to `chrome://flags` for Google Chrome) -- removal inbound!

queengooborg: I opened #15870 to remove this property altogether.

PR #15735, opened 2022-04-07
Remove irrelevant Firefox Android flag data for a HTML element

queengooborg:
This PR removes irrelevant flag data for Firefox Android for the `a` HTML element as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-flag-data).

### DIFF
--- a/html/elements/a.json
+++ b/html/elements/a.json
@@ -470,15 +470,7 @@
                 ]
               },
               "firefox_android": {
-                "version_added": true,
-                "version_removed": "79",
-                "flags": [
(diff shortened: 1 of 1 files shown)


lucalves (approved):

Elchi3 (approved):
The important note to make here is that Firefox Android 79+ supports no flags (https://github.com/mdn/browser-compat-data/issues/9679). With that in mind, this PR makes sense.

PR #15737, opened 2022-04-07
Remove irrelevant Firefox flag data for link HTML element

queengooborg:
This PR removes irrelevant flag data for Firefox and Firefox Android for the `link` HTML element as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-flag-data).

### DIFF
--- a/html/elements/link.json
+++ b/html/elements/link.json
@@ -1142,19 +1142,23 @@
                     "version_added": "85"
                   },
                   {
-                    "version_added": "78",
-                    "flags": [
-                      {
-                        "type": "preference",
-                        "name": "network.preload",
-                        "value_to_set": "true"
-                      }
-                    ]
+                    "partial_implementation": true,
+                    "version_added": "56",
+                    "version_removed": "57",
+                    "notes": "Disabled due to various web compatibility issues (e.g. <a href='https://bugzil.la/1405761'>bug 1405761</a>)."
                   }
                 ],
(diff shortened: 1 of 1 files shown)


html/elements/link.json
```diff
@@ -1150,35 +1150,11 @@
                         "value_to_set": "true"
                       }
                     ]
-                  },
-                  {
-                    "partial_implementation": true,
```
foolip, line 1155: If this was removed by the script there's a bug, this isn't flag data at all.

foolip (changes requested):

html/elements/link.json
queengooborg: This one was a manual change; it's not particularly useful information to record that one single release had support in the past.

foolip: I agree it's not particularly useful, but still more useful than the flag data that we're still keeping.

Maybe we could have a guideline about removing failed attempts to ship that only lasted one release far in the past, but I think I'd still prefer to keep the data, since the collector could potentially re-discover this and then we'd have to investigate.

foolip: Oh, I missed that you also removed the same thing for Firefox Android below. I have the same feedback there &ndash; let's remove the flag data only.

foolip (approved):

PR #15752, opened 2022-04-10
Fix broken HTML in notes

queengooborg:
This PR fixes the broken HTML currently present in the notes.  These were all caught by the linter introduced in #5591.

### DIFF
--- a/webextensions/api/management.json
+++ b/webextensions/api/management.json
@@ -326,7 +326,7 @@
               },
               "firefox_android": {
                 "version_added": "55",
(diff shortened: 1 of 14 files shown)


webextensions/api/management.json
```diff
@@ -326,7 +326,7 @@
               },
               "firefox_android": {
                 "version_added": "55",
-                "notes": "Before version 56, only extensions whose <code>'type'</code> is <code>'theme'<code> are supported."
+                "notes": "Before version 56, only extensions whose <code>type</code> is <code>'theme'</code> are supported."
```
foolip, line 329: Drive by removal of the first quotes here makes sense.

foolip (approved):

PR #15760, opened 2022-04-11
Add opera 71 support

keithamus:
#### Summary

I've tested Opera 71 against https://github.github.com/browser-support/ which uses MDN compat data to generate feature tables, but also has live tests for the currently running browser.

I've noticed that Opera 71 supports both `Promise.any` and `AggregateError`.

 The two tests that pass are:

```js
'any' in Promise && typeof Promise.any === 'function'
```

```js
typeof globalThis.AggregateError === 'function'
```

You can verify this by visiting https://github.github.com/browser-support/ which runs these tests next to their respective table headings. The result of visiting that site in Opera 71 should be:

a screenshot showing Opera visiting the browser-support page

As you can see the `AggregateError` and `Promise.any` pips are green, denoting that these tests returned a true on this browser.

#### Related issues

https://github.com/github/browser-support/pull/16 might also be relevant.

### DIFF
--- a/javascript/builtins/AggregateError.json
+++ b/javascript/builtins/AggregateError.json
@@ -34,7 +34,7 @@
               "version_added": "71"
             },
             "opera_android": {
-              "version_added": false
+              "version_added": "60"
             },
             "safari": {
               "version_added": "14"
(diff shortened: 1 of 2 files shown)


queengooborg (changes requested):
Hey there, thank you for your PR, we appreciate you taking the time to share your research with this project as well!

This makes sense to me; support was added in Chrome 85 which matches Opera 71.  May we also update Opera Android?  That should be "60" based upon the Chromium version/

queengooborg (approved):
Thank you very much, LGTM!

PR #15768, opened 2022-04-11
Update readme

queengooborg:
This PR adds additional information to the readme to clarify usage in several contexts, including NodeJS CJS + ESM, Deno, directly within web browsers, and even other programming languages.

Blocked by #16232 which allows for ESM imports using type assertions.


ddbeck:
Marked as not ready, since merging https://github.com/mdn/browser-compat-data/pull/15774 will block this.

Also, I'm 👎 on blessing the use of a CDN here. Adding it to the readme suggests that it's something we support. We can control what's on npm, but not unrelated services that consume it. If unpkg has some problem (e.g., it's slow to pick up new releases), I don't think BCD itself should invite issues about it.

queengooborg:
I can understand the reaction towards CDNs, but I don't know of any other solutions for Deno users (and in-browser ones).  Since Deno doesn't have a package manager and simply relies on web URLs, they recommend using a CDN to import NodeJS modules.  I think that we should provide some instructions to Deno users to show we're thinking about them and their growing user base.

We could also include links to a few different CDNs?

queengooborg:
@lucacasonato, may I get your feedback on this?

lucacasonato:
Also it's worth noting that unpkg.com is really just an NPM proxy. It downloads tarballs from NPM on demand, unzips them, and serves the individual file that was requested (there is some caching logic too). It's not like this is some alternate registry, or like BCD is a very security sensitive package that would be very sensitive to being tampered with.

queengooborg:
I'm going to bundle these changes into #16232 which introduces a wrapper script as well.

Issue #15771, opened 2022-04-12
NodeIterator.expandEntityReferences, TreeWalker.expandEntityReferences - remove the entries from BCD


OnkarRuikar:
The PR https://github.com/mdn/content/pull/14922 removed following WebAPI property pages from `mdn/content`:
- NodeIterator.expandEntityReferences
- TreeWalker.expandEntityReferences

These expandEntityReferences properties have been dropped from the specification long time ago. And are deprecated.
So they are no longer needed in BCD and need to be removed.

### Can you link to any release notes, bugs, pull requests, or MDN pages related to this?
In Firefox release 21 they got removed:
https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/21#dom

It has been instructed here: https://github.com/mdn/content/pull/14856#issuecomment-1095663979

PR #15789, opened 2022-04-13
Remove deprecated param `expandEntityReferences` from Document.createTreeWalker()

OnkarRuikar:
Completes the fix for #15771

#### Summary
The expandEntityReferences parameter has been dropped from the specification long time ago. And is deprecated.

The mdn/content page doesn't use the parameter. It's been lingering in BCD table.
https://developer.mozilla.org/en-US/docs/Web/API/Document/createTreeWalker

In Firefox release 21 it got removed:
https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/21#dom

It has been instructed here: https://github.com/mdn/content/pull/14856#issuecomment-1095663979

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -2912,61 +2912,6 @@
             "deprecated": false
           }
         },
-        "expandEntityReferences_parameter": {
-          "__compat": {
-            "description": "<code>expandEntityReferences</code> parameter",
-            "support": {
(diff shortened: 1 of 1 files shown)


queengooborg (approved):
LGTM, thanks!

PR #15790, opened 2022-04-13
Update the javascript/operators/await spec URL

sideshowbarker:
The spec URLs for the ES spec should have not just an anchor but also a filename for the multipage HTML filename the anchor is in. (Browsers do a client-side redirect of the links, so that the user ends up in the right file anyway. But having the filename avoids that unnecessary redirect.


queengooborg (approved):
Good catch, thank you!

PR #15793, opened 2022-04-13
Safari supports Subresource Integrity

### LINKED ISSUE #5430: http.Content-Security-Policy.require-sri-for - Supported in Safari iOS
`Require-Sri-For` HTTP Header functions is available in iOS. This needs to be updated.

EDIT; Additional Information:

https://webkit.org/status/#feature-subresource-integrity

queengooborg:
This PR fixes #5430 by copying the data from CanIUse: https://caniuse.com/subresource-integrity

### DIFF
--- a/http/headers/content-security-policy.json
+++ b/http/headers/content-security-policy.json
@@ -1169,10 +1169,10 @@
                   "version_added": "41"
                 },
                 "safari": {
-                  "version_added": false
+                  "version_added": "11"
                 },
                 "safari_ios": {
-                  "version_added": false
+                  "version_added": "11.3"
                 },
                 "samsunginternet_android": {
                   "version_added": "6.0"


foolip:
Have you been able to confirm in some way this is correct? I can't find any "require-sri-for" in https://bugs.webkit.org/ and https://caniuse.com/subresource-integrity is about https://w3c.github.io/webappsec-subresource-integrity/ which is a bunch of other bits.

foolip:
Also `git grep -i require-sri-for` in a checkout of https://github.com/WebKit/WebKit has no matches. Seems to me this isn't supported.

queengooborg:
I haven't really looked into this too much, but it sounds like this isn't supported.  I'm going to close this PR and look into it again later on.

PR #15798, opened 2022-04-13
Collapse `window.inner{Width,Height}` notes into one support statement

ddbeck:
#### Summary

The Firefox notes for `window.innerWidth` and `window.innerHeight` don't really need their own statements. This collapses things into something a little less complicated. 

#### Related issues

It also partially addresses an issue raised in https://github.com/mdn/yari/issues/5945

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -3155,7 +3155,7 @@
             },
             "firefox_android": {
               "version_added": "4",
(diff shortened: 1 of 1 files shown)


queengooborg (approved):
LGTM, thanks!

PR #15806, opened 2022-04-13
Fix statuses for Deno releases

queengooborg:
This PR fixes the status data for Deno releases.  Multiple releases are marked as "current", which can't be true.  Caught by the linter in 
#6167.

Pinging @lucacasonato for review.


lucacasonato (approved):
My bad, I don't know how I managed to mess this up 4x in a row. 😆

queengooborg:
Haha, it's all good, it gets the best us!  😛

PR #15814, opened 2022-04-13
HTTP Referer header length limit in Safari

bershanskiy:
This is a follow-up to commit 1e966e53ce67eb46f5e633ccd949d9a0105a981c and issue #6130.

#### Summary
Safari 14 limits Referer header to 4096 bytes.

This change landed in WebKit commit
https://github.com/WebKit/WebKit/commit/12b8dc73bed78248eb6697cc3c9088d1bd880dbe
which was first released in Safari 14, according to [browsers/safari.json](https://github.com/mdn/browser-compat-data/blob/main/browsers/safari.json), [browsers/safari_ios.json](https://github.com/mdn/browser-compat-data/blob/main/browsers/safari_ios.json).

#### Related issues

#6125
#6130

### DIFF
--- a/http/headers/referer.json
+++ b/http/headers/referer.json
@@ -78,10 +78,10 @@
                 "version_added": "55"
               },
               "safari": {
-                "version_added": false
+                "version_added": "14"
               },
               "safari_ios": {
-                "version_added": false
+                "version_added": "14"
               },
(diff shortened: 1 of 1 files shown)


queengooborg (approved):
Thank you for following this one up!

PR #15817, opened 2022-04-13
Add Chrome/Safari versions for text-size-adjust CSS property

queengooborg:
This PR adds real values for Chrome and Safari for the `text-size-adjust` CSS property.  There are `version_added: false` statements referring to a bug when the prefix is applied, however the bug was fixed a long time ago and can now be considered irrelevant.

### DIFF
--- a/css/properties/text-size-adjust.json
+++ b/css/properties/text-size-adjust.json
@@ -6,16 +6,9 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/text-size-adjust",
           "spec_url": "https://drafts.csswg.org/css-size-adjust/#adjustment-control",
           "support": {
-            "chrome": [
-              {
-                "version_added": "54"
-              },
-              {
-                "prefix": "-webkit-",
-                "version_added": false,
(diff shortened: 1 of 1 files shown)


css/properties/text-size-adjust.json
```diff
@@ -62,16 +55,9 @@
...
-                "version_added": false
-              },
-              {
-                "prefix": "-webkit-",
-                "version_added": false,
-                "notes": "Instead of ignoring the <code>-webkit-text-size-adjust</code> property, a bug prevents desktop Safari users from zooming in or out. The bug was fixed after Safari 5."
-              }
-            ],
+            "safari": {
+              "version_added": false
```
foolip, line 59: Have you verified that this isn't supported in Safari macOS? But it is supported in Safari iOS and Chrome macOS? It seems like unusual situation.

queengooborg: I just checked, and that does seem to be the case:

Screen Shot 2022-05-04 at 15 26 48
(top is my laptop, bottom is my iPad)

gsnedders: > Have you verified that this isn't supported in Safari macOS?

That's correct.

queengooborg:
Note: this is needed to land #15816.

foolip (approved):

PR #15825, opened 2022-04-14
Update Contact Picker API data

lukewarlow:
#### Summary

Marked as standard track

Removed incorrect opera support data

Add Safari iOS flag data

The opera data is clearly an error when looking at all the data as a whole and I did double check that it wasn't exposed.

Safari I've tested on latest version.

https://firt.dev/ios-14.5#contact-picker-api and this article suggests 14.5 (which fits with what I remember)

While the flag is available on desktop and is exposed the API always throws errors for me so I've left that as false.

### DIFF
--- a/api/ContactsManager.json
+++ b/api/ContactsManager.json
@@ -24,7 +24,7 @@
             "version_added": false
           },
           "opera": {
-            "version_added": "57"
+            "version_added": false
           },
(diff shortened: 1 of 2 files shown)


queengooborg (approved):
LGTM!  👍

PR #15834, opened 2022-04-14
Add overflow:overlay property compatibility

podrezo:
#### Summary
Add overflow:overlay property compatibility. This was requested in https://github.com/mdn/content/pull/14435

Tested the compatibility on Safari on Mac; the behavior is exactly described here:

https://caniuse.com/?search=overlay

### DIFF
--- a/css/properties/overflow.json
+++ b/css/properties/overflow.json
@@ -119,54 +119,6 @@
             }
           }
         },
-        "overlay": {
-          "__compat": {
-            "description": "<code>overlay</code> value",
-            "support": {
-              "chrome": {
-                "version_added": "15"
-              },
-              "chrome_android": {
-                "version_added": "100"
-              },
-              "edge": {
-                "version_added": "79"
-              },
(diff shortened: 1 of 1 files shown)


wbamberg:
@podrezo , thanks for this PR!

It's getting a few [CI errors](https://github.com/mdn/browser-compat-data/runs/6030797632?check_suite_focus=true):

* subfeatures need to be listed alphabetically (apparently `npm run fix` will sort this out)
* `4` is not a valid version number for `samsunginternet_android`, but apparently `4.0` is allowed
* subfeature support must not be earlier than the base feature support (i.e. you can't say that a particular property value was supported _after_ the property itself was supported).

You can run `npm run test` locally before pushing your changes, so you'll know whether CI is likely to work.

podrezo:
Sorry about that @wbamberg , fixed now I think.

wbamberg (approved):
👍 this looks good to me @podrezo , thank you!

PR #15845, opened 2022-04-15
Remove obsolete SVG element <animateColor>

bershanskiy:
#### Summary

Remove data about obsolete SVG element [`<animateColor>`](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/animateColor). Modern browsers support [`<animate>`](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/animate) instead.

According to BCD, no modern browser implements this feature. Also, [the MDN demo does not work](https://developer.mozilla.org/en-US/docs/Web/SVG/Element/animateColor#result).

#### Related issues
https://github.com/mdn/content/pull/15027


queengooborg (approved):
👍

PR #15851, opened 2022-04-15
Remove irrelevant <textarea> attribute autocapitalize

bershanskiy:
#### Summary

Remove irrelevant/odd data about `<textarea>` attribute `autocapitalize`. Note that there is a dedicated page about [`autocapitalize` global attribute](https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/autocapitalize).

Create `<textarea autocapitalize="...">` with any value `off`, `none`, `on`, `sentences`, `words`, or `characters` and notice that nothing changes.

#### Related issues

Vaguely related: https://github.com/mdn/content/pull/15034

### DIFF
--- a/html/elements/textarea.json
+++ b/html/elements/textarea.json
@@ -60,53 +60,6 @@
             "deprecated": false
           }
         },
-        "autocapitalize": {
-          "__compat": {
-            "support": {
-              "chrome": {
-                "version_added": false
-              },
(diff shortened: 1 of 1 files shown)


queengooborg (approved):
Makes sense to me, no reason to keep a separate instance for a specific element when we already have a global one!

PR #15864, opened 2022-04-17
Update Chromium versions for api.CSSStyleDeclaration.length

queengooborg:
This PR updates and corrects the real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `length` member of the `CSSStyleDeclaration` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v5.0.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/CSSStyleDeclaration/length

### DIFF
--- a/api/CSSStyleDeclaration.json
+++ b/api/CSSStyleDeclaration.json
@@ -365,7 +365,7 @@
           "spec_url": "https://drafts.csswg.org/cssom/#dom-cssstyledeclaration-length",
           "support": {
             "chrome": {
-              "version_added": "44"
+              "version_added": "1"
             },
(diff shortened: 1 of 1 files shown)


foolip (changes requested):
Should the Safari data also be updated? It seems likely it's also wrong...

queengooborg:
It probably does need to be updated, but I currently don't have access to older Safari versions...

foolip (approved):

foolip:
OK, let's leave Safari wrong for now :)

PR #15872, opened 2022-04-18
Scripting API data

rebloor:
#### Summary

This is the browser compatibility data for the Scripting API and an update to the Tabs API to indicate that executeScript, insertCSS, and removeCSS are not available in Manifest V3.

Supported versions for Safari to be confirmed.

Find compatibility data in support of https://github.com/mdn/content/pull/14725


webextensions/api/scripting.json
```diff
@@ -0,0 +1,393 @@
...
+  "webextensions": {
+    "api": {
+      "scripting": {
+        "ContentScriptFilter": {
+          "__compat": {
+            "mdn_url": "https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/API/scripting/ContentScriptFilter",
+            "support": {
+              "chrome": {
+                "version_added": "88",
+                "notes": "Available for use in Manifest V3 only."
```
queengooborg, line 11: I see this note is repeated a lot.  I think this would be better to put on the corresponding MDN article, rather than in notes, since this seems to be consistent for all browsers?

rebloor: If we could put a note on the feature itself, that would be ideal, otherwise, we really need to include these notes so that the information isn't disconnected.

queengooborg: I do think that the information is helpful, but generally notes in BCD are meant to describe a difference in implementation that may affect (cross-)browser compatibility, such as variation from the spec.  This is tough because web extensions don't really have a "spec" to refer to in the first place, but the way I see it is: if this behavior is consistent across all browser engines, then it should be documented in mdn/content as a part of the feature rather than a BCD note to describe uncommon behavior.

rebloor: Yes, from some of the other changes I'm seeing that this is not an appropriate or sustainable approach. I'm going to revisit all the compatibility data changes when I've completed the core content. Hopefully early next week.

webextensions/api/scripting.json
```diff
@@ -0,0 +1,393 @@
...
+              },
+              "safari_ios": {
+                "version_added": null,
+                "notes": "Available for use in Manifest V3 only."
+              }
+            }
+          }
+        },
+        "InjectionResult": {
```
Rob--W, line 40: InjectionResult doesn't need an entry either as it's part of `executeScript`, and in fact the latest version of https://github.com/mdn/content/pull/14725 doesn't list the article either.

webextensions/api/scripting.json
```diff
@@ -0,0 +1,393 @@
...
+                "version_added": "101",
+                "notes": "Available for use in Manifest V3 only."
+              },
+              "opera": {
+                "version_added": "74",
+                "notes": "Available for use in Manifest V3 only."
+              },
+              "safari": {
+                "version_added": "15.4",
+                "notes": "Available for use in Manifest V3 only."
```
Rob--W, line 101: In Safari, InjectionTarget (describing a parameter of executeScript/insertCSS/removeCSS) is supported in MV2 and MV3 (since Safari 15.4, including `safari_ios`).

webextensions/api/scripting.json
```diff
@@ -0,0 +1,393 @@
...
+                "version_added": "101",
+                "notes": "Available for use in Manifest V3 only."
+              },
+              "opera": {
+                "version_added": "74",
+                "notes": "Available for use in Manifest V3 only."
+              },
+              "safari": {
+                "version_added": null,
+                "notes": "Available for use in Manifest V3 only."
```
Rob--W, line 136: Safari doesn't support `RegisteredContentScript`, so no note needed about MV3 (if supported, it would likely be available in MV2 too).

This also applies to `getRegisteredContentScripts`, `registerContentScripts`, `removeContentScripts`

webextensions/api/scripting.json
```diff
@@ -0,0 +1,393 @@
...
+                "version_added": "88",
+                "notes": "Available for use in Manifest V3 only."
+              },
+              "edge": {
+                "version_added": "88",
+                "notes": "Available for use in Manifest V3 only."
+              },
+              "firefox": {
+                "version_added": "101",
+                "notes": "Available for use in Manifest V3 only."
```
Rob--W, line 89: Introduced in 101 for use in Manifest V3 only, available to all manifest versions since Firefox 102.

(copy-edit to what makes sense please)

webextensions/api/scripting.json
```diff
@@ -0,0 +1,393 @@
...
+              },
+              "safari_ios": {
+                "version_added": null,
+                "notes": "Available for use in Manifest V3 only."
+              }
+            }
+          }
+        },
+        "executeScript": {
```
Rob--W, line 145: `executeScript` needs extra sections on specific properties:

- `injectImmediately`
   - not supported by Safari
- `world`
   - not supported by Safari
   - Firefox only supports ISOLATED, not MAIN.

webextensions/api/scripting.json
```diff
@@ -0,0 +1,393 @@
...
+                "notes": "Available for use in Manifest V3 only."
+              },
+              "safari_ios": {
+                "version_added": null,
+                "notes": "Available for use in Manifest V3 only."
+              }
+            }
+          }
+        },
+        "insertCSS": {
```
Rob--W, line 215: Need specific note for properties (in `insertCSS` and `removeCSS`):
- `origin` not supported in Safari.

webextensions/api/scripting.json
```diff
@@ -0,0 +1,393 @@
...
+                "notes": "Available for use in Manifest V3 only."
+              },
+              "safari_ios": {
+                "version_added": "15.4",
+                "notes": "Available for use in Manifest V3 only."
+              }
+            }
+          }
+        },
+        "registerContentScripts": {
```
Rob--W, line 250: Need specific note for properties:

- `persistAcrossSessions`:
  - In Firefox, only the `false` value is supported. Support for the true value will be introduced in `<a href="https://bugzilla.mozilla.org/show_bug.cgi?id=1751436">bug 1751436</a>`.

webextensions/api/scripting.json
```diff
@@ -0,0 +1,393 @@
+{
+  "webextensions": {
+    "api": {
+      "scripting": {
+        "ContentScriptFilter": {
```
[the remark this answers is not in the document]
rebloor: Removed

webextensions/api/scripting.json
rebloor: Removed

webextensions/api/scripting.json
rebloor: Updated

webextensions/api/scripting.json
rebloor: Updated

webextensions/api/scripting.json
rebloor: Updated

webextensions/api/scripting.json
rebloor: Added

webextensions/api/scripting.json
```diff
@@ -0,0 +1,351 @@
...
+              },
+              "firefox_android": {
+                "version_added": "101"
+              },
+              "opera": {
+                "version_added": "74"
+              },
+              "safari": {
+                "version_added": "15.4",
+                "notes": "Supported in Manifest V2 and V3 as a parameter of <code>executeScript</code>, <code>insertCSS</code>, and <code>removeCSS</code>."
```
Rob--W, line 26: This note is unnecessary. Let's remove it.

`InjectionTarget` is always a parameter of the mentioned APIs. I previously commented about MV2/MV3 because your previous version of the BCD entry mentioned something about MV3 compat in every note.

webextensions/api/scripting.json
```diff
@@ -0,0 +1,351 @@
...
+                "version_added": "101"
+              },
+              "firefox_android": {
+                "version_added": "101"
+              },
+              "opera": {
+                "version_added": "74"
+              },
+              "safari": {
+                "version_added": null
```
Rob--W, line 55: We know that Safari doesn't support these parts of the API, so use `"version_added": false` instead of `null`.

(this applies to all entries; if you have doubts about the level of support, tag me)

webextensions/api/scripting.json
```diff
@@ -0,0 +1,351 @@
...
+              },
+              "firefox_android": {
+                "version_added": "101"
+              },
+              "opera": {
+                "version_added": "74"
+              },
+              "safari": {
+                "version_added": "15.4",
+                "notes": "Doesn't support <code>origin</code>."
```
Rob--W, line 196: Could you add a new entry for "origin" instead of making it a comment?

(also in removeCSS)

Rob--W (changes requested):

webextensions/api/scripting.json
rebloor: Done

webextensions/api/scripting.json
rebloor: I assume this means that every instance where we have "null" for Safari support it should be replaced with `false`. Done.

webextensions/api/scripting.json
```diff
@@ -0,0 +1,399 @@
...
+                },
+                "firefox_android": {
+                  "version_added": "101",
+                  "notes": "Only supports <code>ISOLATED</code>, not <code>MAIN</code>."
+                },
+                "opera": {
+                  "version_added": "74"
+                },
+                "safari": {
+                  "version_added": false
```
Rob--W, line 136: Correction to what I wrote before: the `"world"` parameter is supported in Safari too, 15.4.

Rob--W (approved):
Thanks, the table looks good. One correction to the "world" property after additional clarification from Apple.

webextensions/api/scripting.json
rebloor: @Rob--W  Safari and Safari iOS?

Rob--W (approved):
Thanks, ready for merging!

PR #15644, opened 2022-04-04
FF100 AbortSignal.timeout() supported

hamishwillee:
`AbortSignal.timeout` was recently added to the spec. 
- FF100 added it in https://bugzilla.mozilla.org/show_bug.cgi?id=1753309
- Deno reports it in release notes for 1.20: https://deno.com/blog/v1.20#support-for-abortsignaltimeoutms
- Chrome shows not yet implemented here: https://chromestatus.com/feature/5768400507764736 (also tested using https://wpt.live/dom/abort/abort-signal-timeout.html on browserstack)
- No signal for Safari
- Not sure about Node. The intent is there, but no clear signal - e.g. this one is closed: https://github.com/nodejs/node/pull/40899
I've marked as not experimental since their are two implementations, and intent from Chromium.

Other docs work for this can be followed in https://github.com/mdn/content/issues/14641

PR #15884, opened 2022-04-19
api.AbortSignal.timeout - node support in 17.3.0

hamishwillee:
From Node docs here https://nodejs.org/api/globals.html#static-method-abortsignaltimeoutdelay `AbortSignal.timeout()` is supported from 17.3.0.

This won't merge because node versions above 17.0.0 are not yet supported.

This follows on from the comment here: https://github.com/mdn/browser-compat-data/pull/15644#issuecomment-1097434698

### DIFF
--- a/browsers/nodejs.json
+++ b/browsers/nodejs.json
@@ -349,14 +349,14 @@
         "17.0.0": {
           "release_date": "2021-10-19",
           "release_notes": "https://nodejs.org/en/blog/release/v17.0.0/",
-          "status": "current",
+          "status": "retired",
           "engine": "V8",
           "engine_version": "9.5"
         },
         "17.2.0": {
           "release_date": "2021-11-30",
           "release_notes": "https://nodejs.org/en/blog/release/v17.2.0/",
-          "status": "current",
+          "status": "retired",
           "engine": "V8",
           "engine_version": "9.6"
         },
--- a/browsers/nodejs.json
+++ b/browsers/nodejs.json
@@ -352,6 +352,20 @@
           "status": "current",
           "engine": "V8",
           "engine_version": "9.5"
+        },
+        "17.2.0": {
+          "release_date": "2021-11-30",
+          "release_notes": "https://nodejs.org/en/blog/release/v17.2.0/",
+          "status": "current",
+          "engine": "V8",
+          "engine_version": "9.6"
+        },
+        "17.3.0": {
+          "release_date": "2021-12-17",
+          "release_notes": "https://nodejs.org/en/blog/release/v17.3.0/",
+          "status": "current",
+          "engine": "V8",
+          "engine_version": "9.6"
         }
       }
     }
--- a/api/AbortSignal.json
+++ b/api/AbortSignal.json
@@ -415,7 +415,7 @@
               "version_added": false
             },
             "nodejs": {
-              "version_added": false
+              "version_added": "17.3.0"
             },
             "opera": {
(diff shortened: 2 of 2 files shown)


queengooborg:
Thanks @hamishwillee!  Would you be up to add NodeJS v17.3.0 to the browser data in this PR (or another PR)?

hamishwillee:
@queengooborg Well, its a little out of scope. 

Happy to give it a go but where is the release information that is needed? I infer that we 
- link to blogs like this one https://nodejs.org/en/blog/release/v17.1.0/ 
- "status": "esr", retired or current
- engine always v8?
- How do I work out engine version?

queengooborg:
These are great questions!  I've been basing the engine version (and yes, the engine's always V8) and release date off of https://nodejs.org/en/download/releases/.  The status should be set to "current" (whereas 17.0.0 should be set to "retired").

hamishwillee:
@queengooborg So do you do it for all point releases, or just the ones you "need" - for example there is at least 17.1.0 and 17.9.0 - if there are ones in between do we do them too?

So the remaining question is - How do I work out engine version? That's not "V8" its a number like `"engine_version": "9.4"`

queengooborg:
We only add the ones that are needed, typically on the following:

- A new major release (v16, v17, etc.)
- A release that changes the V8 engine version (v17.0.0-v17.1.0 uses 9.5, but v17.2.0 and on use 9.6)
- A release that otherwise adds a feature, like in this case

So I would say, add both v17.2.0 and v17.3.0 since v17.2.0 changed V8 engine versions and v17.3.0 added this feature (confirmed by https://nodejs.org/api/globals.html#static-method-abortsignaltimeoutdelay).

As for the engine, I think there's a little bit of confusion -- the engine is actually called "V8" (like the V8 engine in a car, and definitely not the vegetable drink), I'm not saying "v8" like "version 8"!  Based on the link above, it says that NodeJS v17.3.0 uses V8 v9.6.180.15 (which we shorten to "9.6"), so our statement should have the following:

```js
"engine": "v8",
"engine_version": "9.6"
```

hamishwillee:
Thanks for the detailed explanation @queengooborg - hopefully next time I will be able to work it out for myself!

browsers/nodejs.json
```diff
@@ -352,6 +352,20 @@
           "status": "current",
```
queengooborg, line 352:
```suggestion
          "status": "retired",
```

browsers/nodejs.json
```diff
@@ -352,6 +352,20 @@
...
           "engine": "V8",
           "engine_version": "9.5"
+        },
+        "17.2.0": {
+          "release_date": "2021-11-30",
+          "release_notes": "https://nodejs.org/en/blog/release/v17.2.0/",
+          "status": "current",
```
queengooborg, line 359:
```suggestion
          "status": "retired",
```

queengooborg (changes requested):
Thank you for the quick changes!  Just a small change: since v17.0.0 and v17.2.0 have been superseded by a new release, v17.3.0, we should mark them as `retired` status instead.  Browsers (as well as NodeJS) should only have one "current" release.

queengooborg (approved):
Huh, that's kind of odd that the blog still says "current" for them...  Ah well!

Thank you very much, this PR is LGTM!  👍

hamishwillee:
Awesome - and thanks for the education!

PR #9229, opened 2021-02-23
Refactor: move compare-features script test out of linter

bershanskiy:
This PR moves a unit test out of [linter](https://github.com/mdn/browser-compat-data/blob/master/test/lint.js#L161) and into a separate mocha test file similar to [`utils.test.js`](https://github.com/mdn/browser-compat-data/blob/master/test/utils.test.js). This unit test is still called by `npm run test`, since [`test`](https://github.com/mdn/browser-compat-data/blob/master/package.json#L52) script includes `npm run mocha`.

For reviewer convenience, I split it up into two commits: the actual change is in the first commit and the second commit is just `npm run fix`.

### Before:
```
$ npm run test

> @mdn/browser-compat-data@3.1.1 test
> npm run mocha && npm run lint


> @mdn/browser-compat-data@3.1.1 mocha
> mocha "test/**.test.js"



  utils
    √ `escapeInvisibles()` works correctly


  1 passing (3ms)


> @mdn/browser-compat-data@3.1.1 lint
> node test/lint
> [rest of output]
```

### After:
```
$ npm run test

> @mdn/browser-compat-data@3.1.1 test /home/runner/work/browser-compat-data/browser-compat-data
> npm run mocha && npm run lint


> @mdn/browser-compat-data@3.1.1 mocha /home/runner/work/browser-compat-data/browser-compat-data
> mocha "test/**.test.js"



  compare-features script
    ✓ `compareFeatures()` works correctly

  utils
    ✓ `escapeInvisibles()` works correctly


  2 passing (7ms)


> @mdn/browser-compat-data@3.1.1 lint /home/runner/work/browser-compat-data/browser-compat-data
> node test/lint
> [rest of output]
```

A checklist to help your pull request get merged faster:
- [ ] Summarize your changes
- [ ] Data: link to resources that verify support information (such as browser's docs, changelogs, source control, bug trackers, and tests)
- [ ] Data: if you tested something, describe how you tested with details like browser and version
- [ ] Review the results of the linter and fix problems reported (If you need help, please ask in a comment!)
- [ ] Link to related issues or pull requests, if any

PR #15886, opened 2022-04-19
Move compare-features unittest to matching folder

queengooborg:
This PR is a quick follow-up to #9229 that moves the unittest for `compare-features` to the same folder as the script for better organization.  Additionaly, the now-invalid JSDoc is removed.


queengooborg:
I'm self-merging this because it's a very small change that doesn't have any impact on BCD other than organizationally.

Stanislav1975 (approved):

PR #15889, opened 2022-04-19
Add linter for incorrect feature statuses

queengooborg:
This PR is a cherry-pick from #6813 and #11430 to add a new linter to test the status of a feature, enforcing the following:

- Experimental and depcrecated cannot both be `true`
- Standard track must be `true` if there is a spec URL

This should also provide a better foundation for #11506 for long-term linting.  Note that this PR will fail tests for now as we have conflicting data.

### DIFF
--- a/scripts/fix-status.js
+++ b/scripts/fix-status.js
@@ -1,0 +1,100 @@
+#!/usr/bin/env node
+/* Any copyright is dedicated to the Public Domain.
+ * http://creativecommons.org/publicdomain/zero/1.0/ */
+
+/**
+ * Return a new feature object whose first-level properties have been
+ * ordered according to Array.prototype.sort, and so will be
+ * stringified in that order as well. This relies on guaranteed "own"
+ * property ordering, which is insertion order for non-integer keys
+ * (which is our case).
+ *
+ * @param {string} key The key in the object
+ * @param {*} value The value of the key
+ *
+ * @returns {*} The new value
+ */
+
+'use strict';
+const fs = require('fs');
+const path = require('path');
+const { platform } = require('os');
+
+/** Determines if the OS is Windows */
+const IS_WINDOWS = platform() === 'win32';
+
+const fixStatus = (key, value) => {
+  const status = value && value.__compat && value.__compat.status;
+  if (status && status.experimental && status.deprecated) {
+    status.experimental = false;
+  }
+  return value;
+};
+
+/**
+ * @param {Promise<void>} filename
+ */
+const fixStatusFromFile = filename => {
+  let actual = fs.readFileSync(filename, 'utf-8').trim();
+  let expected = JSON.stringify(JSON.parse(actual, fixStatus), null, 2);
+
+  if (IS_WINDOWS) {
+    // prevent false positives from git.core.autocrlf on Windows
+    actual = actual.replace(/\r/g, '');
+    expected = expected.replace(/\r/g, '');
+  }
+
+  if (actual !== expected) {
+    fs.writeFileSync(filename, expected + '\n', 'utf-8');
+  }
+};
+
+if (require.main === module) {
+  /**
+   * @param {string[]} files
+   */
+  function load(...files) {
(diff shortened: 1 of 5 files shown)


scripts/fix-status.js
```diff
@@ -0,0 +1,100 @@
+#!/usr/bin/env node
```
foolip, line 1: Shebang!

scripts/fix-status.js
```diff
@@ -0,0 +1,100 @@
...
+/* Any copyright is dedicated to the Public Domain.
+ * http://creativecommons.org/publicdomain/zero/1.0/ */
+
+/**
+ * Return a new feature object whose first-level properties have been
```
foolip, line 6: Should this doc block be moved to before the function? Although which function, this seems like it's documenting a helper for sorting keys before JSON-serializing, which isn't here?

scripts/fix-status.js
```diff
@@ -0,0 +1,100 @@
...
+ * @returns {*} The new value
+ */
+
+'use strict';
+const fs = require('fs');
+const path = require('path');
+const { platform } = require('os');
+
+/** Determines if the OS is Windows */
+const IS_WINDOWS = platform() === 'win32';
```
foolip, line 24: Import this from utils?

scripts/fix-status.js
```diff
@@ -0,0 +1,100 @@
...
+
+const fixStatus = (key, value) => {
+  const status = value?.__compat?.status;
+  if (status && status.experimental && status.deprecated) {
+    status.experimental = false;
```
foolip, line 29: I disagree that we can autofix this. Why is it necessarily this status that is wrong?

scripts/fix-status.js
```diff
@@ -0,0 +1,100 @@
...
+  }
+  return value;
+};
+
+/**
+ * @param {Promise<void>} filename
```
foolip, line 35: This is sync, no Promise.

scripts/fix-status.js
```diff
@@ -0,0 +1,100 @@
...
+  if (actual !== expected) {
+    fs.writeFileSync(filename, expected + '\n', 'utf-8');
+  }
+};
+
+if (require.main === module) {
+  /**
+   * @param {string[]} files
+   */
+  function load(...files) {
```
foolip, line 56: This piece of code is repeated in lots of places, can it be put in utils?

test/linter/test-status.js
```diff
@@ -0,0 +1,57 @@
...
+    }
+    if (compat.status.experimental && compat.status.deprecated) {
+      logger.error(
+        chalk`{red → Unexpected simultaneous experimental and deprecated status in {bold ${path}}}`,
+      );
+    }
+  }
+
+  // Check children
+  for (const member in data) {
```
foolip, line 29: Something with `Object.entries(data)` would make sense here since we need both key and value.

Although, can we just use utils.walk to visit all the entries?

foolip (commented):
A bunch of comments here would apply to all lint code. If you think they make sense but don't want to fix it now, can you file issues?

test/linter/test-status.js
queengooborg: For this and the above comment, I plan to do as such in the next-gen linter!

queengooborg:
I've now updated this linter to enforce migration 007 and ensure we don't regress there!

queengooborg:
I don't know why but the errors won't display the second time for whatever reason.  #5300 should fix this though.

test/linter/test-status.js
```diff
@@ -0,0 +1,106 @@
...
+
+  const logger = new Logger('Feature Status');
+
+  checkStatus(data, logger);
+
+  logger.emit();
+  return logger.hasErrors();
+}
+
+module.exports = testStatus;
```
saschanaz, line 106: The name is a bit ambiguous as I have #6813 (which should be renamed too) and #7351.

PR #15895, opened 2022-04-19
Add workflow to catch for potential duplicates

queengooborg:
This PR adds a GitHub Actions workflow to alert for potential duplicate issues.

### DIFF
--- a/.github/workflows/potential-duplicate-issues.yml
+++ b/.github/workflows/potential-duplicate-issues.yml
@@ -1,0 +1,31 @@
+name: Potential Duplicates
+on:
+  issues:
+    types: [opened, edited]
+jobs:
+  run:
+    runs-on: ubuntu-latest
+    steps:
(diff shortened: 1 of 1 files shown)


queengooborg:
I tested this out and it's working great!  Only notifying the user once when merge conflicts are first found, and accurately adding or removing the label!

Rumyra:
Awesome I'm excited to see what happens with this one & if it ends up being useful we can move it to mdn/workflows. Ping @schalkneethling to take a look 👍

schalkneethling:
Indeed! If this works well, then please open a PR to add it to mdn/workflows @queengooborg. Thanks!

Rumyra:
Are we good to go with merging?

schalkneethling:
> Are we good to go with merging?

I believe so.

Rumyra:
It's done!

PR #15897, opened 2022-04-19
Add a script to remove redundant flags

queengooborg:
This PR adds a script to remove redundant flags throughout our data.  This is a script I have personally been using for nearly all of the flag removals, and I thought this would be a great resource to share with other contributors as well.


lucalves (approved):

scripts/remove-redundant-flags.js
```diff
@@ -0,0 +1,206 @@
...
+
+  return earliestVersion;
+};
+
+const removeRedundantFlags = (key, value, limitBrowser) => {
+  if (key === '__compat') {
+    for (const [browser, rawSupportData] of Object.entries(value.support)) {
+      if (limitBrowser && browser != limitBrowser) continue;
+
+      if (rawSupportData !== undefined) {
```
foolip, line 53: Is it really possible for `rawSupportData` to be undefined here? (I don't think so.)

scripts/remove-redundant-flags.js
```diff
@@ -0,0 +1,206 @@
...
+        const supportData = Array.isArray(rawSupportData)
+          ? rawSupportData
+          : [rawSupportData];
+        const result = [];
+
+        const simpleStatement = supportData.find(statement => {
```
foolip, line 59: This code felt familiar, looks like this bit from the collector:
https://github.com/foolip/mdn-bcd-collector/blame/4488e1171a2bd83bfe40459aa27add2332449cfe/update-bcd.js#L276-L282

But I don't understand the purpose of looking for a simple statement here, why does this matter to flag removal?

It would be great write in prose as docs for this function what the rules are for when flag data is going to be removed. I think it should be something like this: For each flag entry, determine if there is a later non-flag entry and if the `version_added` of that entry is two years old or more. Such flag entries are candidates for removal. Then, if all of the flag entries are candidates and they are all at end of the array, remove them. Otherwise leave them, as removing them would leave a hole between non-flag entries, which the guideline carves out an exception for.

caugner (changes requested):
I think this PR would need to be updated to use ESM imports.

scripts/remove-redundant-flags.js
```diff
@@ -0,0 +1,187 @@
...
+import { IS_WINDOWS } from '../test/utils.js';
+
+const dirname = fileURLToPath(new URL('.', import.meta.url));
+
+const getEarliestVersion = (...args) => {
+  const versions = args
+    .filter((version) => typeof version === 'string')
+    .map((version) => version.replace('≤', ''));
+
+  let earliestVersion = versions[0];
```
foolip, line 23: What if `versions` is empty? It will end up returning undefined, but that's an "accident" of how the code is written and doesn't look intentional even if it was. How about:

```suggestion
  let earliestVersion;
```

Then tweak the loop so that earliestVersion is set on the first iteration.

scripts/remove-redundant-flags.js
```diff
@@ -0,0 +1,187 @@
...
+import { hideBin } from 'yargs/helpers';
+
+import bcd from '../index.js';
+import { IS_WINDOWS } from '../test/utils.js';
+
+const dirname = fileURLToPath(new URL('.', import.meta.url));
+
+const getEarliestVersion = (...args) => {
+  const versions = args
+    .filter((version) => typeof version === 'string')
```
foolip, line 20: This should make the conditions in the loop easier to reason about:

```suggestion
    .filter((version) => typeof version === 'string' && version !== 'preview')
```

scripts/remove-redundant-flags.js
```diff
@@ -0,0 +1,187 @@
...
+    .map((version) => version.replace('≤', ''));
+
+  let earliestVersion = versions[0];
+
+  for (const version of versions) {
+    if (
+      earliestVersion === 'preview' ||
```
foolip, line 27: Here the condition can be just `!earliestVersion || compareVersions.compare(earliestVersion, version, '>')` if there are no "preview" values left.

foolip (commented):
Did some reviewing but didn't get past the first helper :)

Can you try running this script a bit with the current data and maybe show a draft PR of what changes it makes?

PR #15913, opened 2022-04-20
Fix experimental+deprecated statuses

saschanaz:
#### Summary

Copied from #6813.


foolip:
All of these change experimental. Was there really no case where the mistake was the other way around?

saschanaz:
> All of these change experimental. Was there really no case where the mistake was the other way around?

I assumed that `deprecated` flags were valid, I didn't reviewed the validity of them 🤔

saschanaz:
Better to merge #15932 first I guess.

foolip:
I'd say this one first, because it's easier to reason about and will probably reduce the number of inheritance fixes needed.

foolip:
Hmm, but fixing https://github.com/mdn/browser-compat-data/pull/11506 before this would also be good, it's mechanical and probably makes many of the same changes.

@queengooborg what do you think of the order of all these status changes?

queengooborg:
My preference would be to perform migration 007 (#11506) first.  It's a change I feel a little more confident about and is just about ready to go, plus it follows the migration steps.

foolip:
@saschanaz I tried to resolve the merge conflict but think I messed it up. Do you want to throw away my merge commit and fix it yourself? I expect a lot fewer changes now that #16285 landed.

saschanaz:
Well, I think the number of changed files are still similar 👀

foolip:
Huh, I thought it would make a big dent, but OK then.

foolip (approved):
I've look through all of the changes using `npm run diff` and to my surprise couldn't spot any cases where it actually shouldn't be deprecated.

PR #15914, opened 2022-04-20
Require all URLs in schema to start with https://

foolip:
This isn't fixing any very important problem, mainly done because the
release_notes type was the only one in the schema without a format,
which would thus allow any scheme, including typos.

There was just a single http:// URL that needed to be updated, and that
already redirected to https://.

### DIFF
--- a/browsers/firefox.json
+++ b/browsers/firefox.json
@@ -7,7 +7,7 @@
       "releases": {
         "1": {
           "release_date": "2004-11-09",
-          "release_notes": "http://website-archive.mozilla.org/www.mozilla.org/firefox_releasenotes/en-US/firefox/releases/1.0.html",
+          "release_notes": "https://website-archive.mozilla.org/www.mozilla.org/firefox_releasenotes/en-US/firefox/releases/1.0.html",
           "status": "retired",
           "engine": "Gecko",
           "engine_version": "1.7"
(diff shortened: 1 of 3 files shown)


bershanskiy:
I like this change because it enforces good security and privacy on the internet. If you don't have time, I'll look into upgrading all other URLs to HTTPS (in `notes` and `release_notes`, etc.) myself.

There is a whole bunch of URLs which could/should be upgraded. E.g.:
https://github.com/mdn/browser-compat-data/blob/825f48679f783d54f3ee5d02d87babe0185f7d0a/browsers/firefox.json#L10

If you like, I can open a separate PR to keep this one focused.

queengooborg (approved):
I don't think we should enforce HTTPS on notes in particular, as we may (still) have HTTP-only links for older references we'll want to keep around.  In `release_notes` and `spec_url` though, it seems we've only got that one release still marked as HTTP and that just redirects to HTTPS, so this is LGTM!

(Not merging yet though since it's still a schema change and should get approval from multiple owners.)

foolip:
@bershanskiy the particular instance of http:// that you linked to is the same one that I updated in this PR, but indeed there are lots more. If you want to replace them, I'd suggest doing so manually one-by-one. Most will either redirect or work as HTTPS, but there are probably some that don't work over HTTPS and it's important we don't accidentally break those links.

Elchi3 (approved):
This looks good to me. It does change the schema but not substantially (it mostly adds a validation), so I believe there is no semver bump needed here. If we feel like bumping, it would probably be a minor bump only.

Issue #15854, opened 2022-04-16
api.ScreenOrientation.unlock - NotSupportedError Operation is not supported


m-nathani:
#### What information was incorrect, unhelpful, or incomplete?
`window.screen.orientation.unlock();` does not work in `firefox version 98.0 and above`
OS: `windows 10`.

#### What did you expect to see?
A/c to the https://developer.mozilla.org/en-US/docs/Web/API/ScreenOrientation/unlock#browser_compatibility it should support all the browser above `firefox version 43.0`

#### Did you test this? If so, how?
Check the error message of how to reproduce it..
https://sentry.io/share/issue/6ff319e6eb824b6f84024e2ceefe76b2/

<details>
<summary>MDN page report details</summary>

* Query: `api.ScreenOrientation.unlock`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/ScreenOrientation/unlock
* Report started: 2022-04-16T05:37:33.093Z

</details>

m-nathani:
@queengooborg the issue was closed, however i cant see the fixed in PR.. 🤔

queengooborg:
Do you mean that you don't see the fixes on MDN Web Docs yet, or that the merged PR doesn't resolve your issue?

If it's the first, I release a new version of BCD every Tuesday, which gets merged and released on MDN the following Thursday.

If it's the second, can you explain how the issue isn't resolved?

m-nathani:
Hey thanks for responding..

Since the PR [15918](https://github.com/mdn/browser-compat-data/pull/15918)  was merged yesterday.. i believe it will be live by coming thursday.

Moreover, just to confirm its fix for the issue was that the unlock feature is supposed to work on latest firefox on windows as i am getting an error on unlock function. 

So the the below changes are basically going to update the docs only ?

```json
            "firefox": {
              "version_added": "43",
              "partial_implementation": true,
              "notes": "Always throws <code>NotSupportedError</code>."
            },
```

queengooborg:
> Moreover, just to confirm its fix for the issue that the unlock feature is supposed to work in latest firefox on windows 10 ? . As i am getting an error on `unlock` function.

[...]

PR #15918, opened 2022-04-20
Mark screen orientation unlock as partial in desktop browsers

lukewarlow:
#### Summary

We correctly noted issues on desktop browsers for lock but not unlock, this PR fixes that.

#### Related issues

Fixes https://github.com/mdn/browser-compat-data/issues/15854


lukewarlow:
Noticed https://github.com/mdn/browser-compat-data/issues/15919 while doing this. Not sure on the correct data so will leave for a future PR.

queengooborg (approved):

alanhogan:
What source(s) do you have for the idea that browsers throw errors on unlock?

PR #15922, opened 2022-04-20
Bump contributor requirement to NodeJS v16+

queengooborg:
This PR bumps the contributors' minimum NodeJS requirement to v16.  Based upon a poll in #12792, most of the top (and active) BCD contributors are not only 
fine with switching, but state that they only keep v14 around because of BCD.  Additionally, this bumps all workflows to use v16 to maintain consistency.

### DIFF
--- a/.github/workflows/add-push-artifacts.yml
+++ b/.github/workflows/add-push-artifacts.yml
@@ -8,20 +8,20 @@ jobs:
     name: Enumerate and diff features
     runs-on: ubuntu-latest
     steps:
-    - uses: actions/checkout@v2
-      with:
-        fetch-depth: 0  # get the full repository checkout, not just the inciting commit
-    - uses: actions/setup-node@v2
-      with:
-        node-version: 14.x
-    - run: npm install
-    - run: node ./scripts/enumerate-features.js features.json
-    - run: node ./scripts/diff-features.js --no-github --format=json > features.diff.json
-    - uses: actions/upload-artifact@v2
-      with:
-        name: enumerate-features
-        path: features.json
-    - uses: actions/upload-artifact@v2
-      with:
-        name: diff-features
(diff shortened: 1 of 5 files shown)


.github/workflows/add-push-artifacts.yml
```diff
@@ -8,20 +8,20 @@ jobs:
...
-    - run: node ./scripts/diff-features.js --no-github --format=json > features.diff.json
-    - uses: actions/upload-artifact@v2
-      with:
-        name: enumerate-features
-        path: features.json
-    - uses: actions/upload-artifact@v2
-      with:
-        name: diff-features
-        path: features.diff.json
+      - uses: actions/checkout@v2
```
foolip, line 11: Please undo the indentation changes, to limit this to just the version change. Indentation is currently inconsistent in our YAML files, which I think is fine. If we wanted to make it consistent let's use prettier or something that can change the indentation to whatever we want.

.github/workflows/node.js.yml
```diff
@@ -5,20 +5,20 @@ name: Node.js CI
 
 on:
   push:
-    branches: [ ] # Test on push to every branch
+    branches: [] # Test on push to every branch
```
foolip, line 8: Unrelated formatting changes.

foolip (changes requested):

.github/workflows/add-push-artifacts.yml
queengooborg: Ah sorry, my editor's set to auto-format with Prettier on file save.  Let me see if I can tweak the files without formatting...

queengooborg:
This was discussed during the last BCD meeting with a resounding "yes" from all of the owners, and enough positive feedback from contributors!

foolip (approved):

PR #15924, opened 2022-04-20
Require notes if partial_implementation is true

### LINKED ISSUE #4162: Require notes for partial_implementation

queengooborg:
This PR updates the schema to require notes when `partial_implementation` is true.  This PR will FAIL for the time being as we have conflicting data.  Fixes #4162.


foolip (commented):
Looks good, but do docs also need updating? And how much data needs updating before this could be landed?

queengooborg:
The docs are good, this linter actually [enforces the existing guideline](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#partial_implementation-requires-a-note)!

foolip:
I think the paragraph starting with "For historical reasons, some support statements have the flag set to true without a note" should  be removed in this PR too.

foolip:
Also, if the lint already enforces this, why doesn't it fail for the cases the schema now fails on?

queengooborg:
Sorry, there's a small detail in the wording (I tweaked it to provide slightly better clarity).  It isn't enforced until this PR lands.

foolip:
Aha. I was thinking of the lint as the manually written rules, not the schema.

foolip:
Alright, this looks good now, just need to make it pass before being able to land :)

PR #15932, opened 2022-04-21
Fix status inheritance inconsistency

saschanaz:
#### Summary

Split from #7351


api/BluetoothRemoteGATTCharacteristic.json
```diff
@@ -630,7 +630,7 @@
             }
           },
           "status": {
-            "experimental": false,
+            "experimental": true,
```
saschanaz, line 633: I'm not sure this is right or not. Sounds like `deprecated` should win in this case (#15913), what do others think?

queengooborg (approved):
These changes seem fine to me, thanks!

PR #14090, opened 2021-12-20
Add missing api.WorkerNavigator.fonts feature

queengooborg:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing `fonts` member of the WorkerNavigator API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v4.0.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/WorkerNavigator/fonts

_Check out the [collector's guide on how to review this PR](https://github.com/foolip/mdn-bcd-collector#reviewing-bcd-changes)._

PR #15937, opened 2022-04-21
Remove non-standard and unsupported api.WorkerNavigator.fonts

foolip:
This was added in https://github.com/mdn/browser-compat-data/pull/14090
but appears to have been a mistake or bug.

The now-removed test for this was simple:
```js
(function() {
  var instance = navigator;
  return 'fonts' in instance;
})();
```

However, the test results from v4.0.0 of the collector are all false:
https://github.com/foolip/mdn-bcd-results/tree/1b5607ce99aff7d11bc07134e382930ad83436f5

This was determined by looking at "api.WorkerNavigator.fonts" in the
results, in particular for Chrome 69 and Safari 15, but there's not a
single true test result among the results.

It's unclear what went wrong, but this isn't supported, and never was.

This was removed from the spec here:
https://github.com/WICG/local-font-access/pull/70

This is the only instance of "local-font-access" in BCD so no other bits
of the API were added, at least not with spec URLs.

### DIFF
--- a/api/WorkerNavigator.json
+++ b/api/WorkerNavigator.json
@@ -303,54 +303,6 @@
           }
         }
       },
-      "fonts": {
-        "__compat": {
-          "spec_url": "https://wicg.github.io/local-font-access/#dom-navigatorfonts-fonts",
-          "support": {
-            "chrome": {
-              "version_added": "69"
-            },
-            "chrome_android": {
-              "version_added": "69"
-            },
-            "edge": {
-              "version_added": "79"
-            },
-            "firefox": {
-              "version_added": false
-            },
-            "firefox_android": {
-              "version_added": false
-            },
-            "ie": {
-              "version_added": false
-            },
-            "opera": {
-              "version_added": "56"
-            },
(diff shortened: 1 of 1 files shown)


foolip:
@queengooborg any idea what happened here? One thing that I suspect might have contributed to this is using the same commit messages for both changes produced by scripts and ones made manually. https://github.com/mdn/browser-compat-data/pull/14090 must have been done manually because there's a spec URL, but the description is just like mechanical collector updates. It would help review in the future if machine-made and human-made changes are clearly separated, so that more careful review can be done for manual changes.

It's also possible there are bugs in the scripts, if you think that's the case it would be very good to pin down, so it can be fixed.

queengooborg (approved):
Hmm...I'm unsure what happened here, but I'm happy to remove it since it's no longer a part of the spec or supported!  Most likely I might've forgotten to bump the reported collector version and the results actually came from another version (I don't use the PR-creating script if I add features manually).

Either way, this is LGTM!

PR #15942, opened 2022-04-21
Fix Firefox data for path() CSS basic-shape type

queengooborg:
This PR attempts to make the data for the `path()` value of the `basic-shape` CSS type easier to interpret by separating the notes into individual statements.

### DIFF
--- a/css/types/basic-shape.json
+++ b/css/types/basic-shape.json
@@ -272,13 +272,9 @@
               },
               "firefox": [
                 {
-                  "version_added": "63",
+                  "version_added": "97",
                   "partial_implementation": true,
-                  "notes": [
-                    "From version 97, the <code>d</code> SVG presentation attribute supports <code>path()</code>.",
-                    "From version 71, the <code>clip-path</code> property supports <code>path()</code>.",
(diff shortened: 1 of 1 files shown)


css/types/basic-shape.json
```diff
@@ -272,13 +272,9 @@
...
                 {
-                  "version_added": "63",
+                  "version_added": "97",
                   "partial_implementation": true,
-                  "notes": [
-                    "From version 97, the <code>d</code> SVG presentation attribute supports <code>path()</code>.",
-                    "From version 71, the <code>clip-path</code> property supports <code>path()</code>.",
-                    "Before version 71, only the <code>offset-path</code> property supports <code>path()</code>."
-                  ]
+                  "notes": "Only supported on the <code>d</code> SVG presentation attribute and the <code>clip-path</code> and <code>offset-path</code> properties."
```
foolip, line 277: "Only? What's left?" I wondered. Looks like `shape-outside` is the remaining place that this could work. Can you test if perhaps that actually works and this should be full support here?

queengooborg: It doesn't look like it's supported, no -- I think it would be useful to mention in the notes though!

foolip (approved):

PR #15944, opened 2022-04-21
Update schema to remove "strict mode" warnings

queengooborg:
Since upgrading `ajv`, we are now getting warnings in regards to strict mode compatibility.  This PR aims to fix these issues by adjusting conflicting statements and adding missing ones.

Note: this also performs a Prettier format on the files -- I'd be happy to perform the Prettier formatting in a separate PR first if that helps.

### DIFF
--- a/schemas/compat-data.schema.json
+++ b/schemas/compat-data.schema.json
@@ -35,9 +35,7 @@
               }
             },
             "additionalProperties": false,
-            "required": ["type", "name"],
-            "minProperties": 2,
-            "maxProperties": 3
+            "required": ["type", "name"]
           }
         },
         "partial_implementation": {
--- a/schemas/compat-data.schema.json
+++ b/schemas/compat-data.schema.json
@@ -2,7 +2,6 @@
   "$schema": "http://json-schema.org/schema#",
 
   "definitions": {
-
     "simple_support_statement": {
       "type": "object",
       "properties": {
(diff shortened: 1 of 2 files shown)


foolip:
> I'd be happy to perform the Prettier formatting in a separate PR first if that helps.

Yes please! And adding that to the CI jobs, if we care about keeping the files formatted like this.

queengooborg:
I've opened #15948 to do as such!

foolip:
This isn't entirely trivial to review. Do these changes make the schema stricter overall? Are there any things allowed by the old schema that now aren't allowed, and are those things that we'll really never want to do?

schemas/compat-data.schema.json
```diff
@@ -35,25 +35,38 @@
...
             "additionalProperties": false,
-            "required": ["type", "name"]
+            "required": ["type", "name"],
+            "minProperties": 0,
+            "maxProperties": 3
           }
         },
         "partial_implementation": {
-          "type": "boolean",
+          "const": true,
```
queengooborg, line 44: This is the only change that's a little stricter than what we already allow, but technically speaking we don't really allow any values for `partial_implementation` other than `true` if it's defined anyways.

schemas/compat-data.schema.json
```diff
@@ -35,25 +35,38 @@
...
         },
         "version_added": {
           "description": "A string (indicating which browser version added this feature), the value true (indicating support added in an unknown version), the value false (indicating the feature is not supported), or the value null (indicating support is unknown).",
-          "allOf": [
-            { "type": ["string", "boolean", "null"] },
-            { "not": { "enum": ["true", "false", "null"] } }
+          "anyOf": [
+            {
+              "type": "string",
+              "pattern": "^(≤?(\\d+)(\\.\\d+)*|preview)$"
```
queengooborg, line 52: Well, technically speaking, this also makes our schema "stricter", but all in all this is just enforcing what we already enforce in the schema rather than just linter files.

schemas/compat-data.schema.json
foolip: Does the lint enforce this? If so I think this change is fine.

foolip:
Looking at your comments I realize I had things backwards. Making the schema stricter is actually a compatible change because consumers won't see anything new in the data as a result.

Extending the schema or making it laxer is what could pose a risk to consumers. Is there any change like that?

queengooborg:
There aren't any changes that make our schema more lax, so we're all good there!

schemas/compat-data.schema.json
queengooborg: Surprisingly it isn't enforced by the linter, but just as surprising is that we've not had a value other than `true` show up in our data.  This will enforce it in a linter though!

schemas/compat-data.schema.json
```diff
@@ -35,25 +35,38 @@
               }
             },
             "additionalProperties": false,
-            "required": ["type", "name"]
+            "required": ["type", "name"],
+            "minProperties": 0,
```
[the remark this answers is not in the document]
queengooborg: I thought that Ajv's strict mode didn't like this field without `min` and `max` defined, but I just removed it and the schema test is still happy!

schemas/compat-data.schema.json
```diff
@@ -236,6 +249,7 @@
       "allOf": [
         { "$ref": "#/definitions/identifier" },
         {
+          "type": "object",
```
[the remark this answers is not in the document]
queengooborg: It probably wouldn't be, but Ajv's not happy if this isn't defined:

```
strict mode: missing type "object" for keyword "maxProperties" at "#/allOf/1" (strictTypes)
strict mode: missing type "object" for keyword "minProperties" at "#/allOf/1" (strictTypes)
```

queengooborg:
I found that once we convert to ESM, the warnings regarding strict mode are repeated over and over again, so this has become slightly more critical.

queengooborg:
I'm going to self-merge this because it's soft blocking ESM migration.  While this is a schema change, it isn't one that will impact consumers (or technically, even contributors), so I don't think this needs a semver bump.

PR #15946, opened 2022-04-21
Rename fetchpriority properties to fetchPriority

foolip:
These IDL attributes were renamed in the spec:
https://github.com/WICG/priority-hints/pull/71

However, they remain all-lowercase in Chromium and are about to be
released in Chrome 101, so capture that in the data.

Also mark all of this experimental, per guidelines:
https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#choosing-an-experimental-status

### DIFF
--- a/api/HTMLIFrameElement.json
+++ b/api/HTMLIFrameElement.json
@@ -486,18 +486,21 @@
           }
         }
       },
-      "fetchpriority": {
+      "fetchPriority": {
         "__compat": {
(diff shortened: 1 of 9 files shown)


queengooborg (approved):
LGTM, thanks!

foolip:
https://chromium-review.googlesource.com/c/chromium/src/+/3600925 is backporting the Chromium fix to 102, so the all-lowercase attribute name will perhaps never be released in stable.

PR #15947, opened 2022-04-21
Mention mdn-bcd-collector in contributing doc

queengooborg:
This PR adds a section that describes using the mdn-bcd-collector to update BCD's data.


foolip:
@Elchi3 I'm asking you for second review on everything now. Do you want that, or only for certain riskier changes?

Elchi3 (approved):
I think it is enough to ask multiple owners only on riskier changes. Thanks!

foolip:
All right, I will start using my judgement instead of bothering you for everything 👍

PR #6493, opened 2020-08-06
adding note about the fullscreen directive bug

chrisdavidmills:
As per https://bugzilla.mozilla.org/show_bug.cgi?id=1608358

PR #15950, opened 2022-04-21
Remove remaining Firefox Android flags + old Firefox desktop flags

### LINKED ISSUE #9679: Firefox for Android 79+ doesn't support flags

queengooborg:
This PR removes all of the remaining Firefox Android flag data from BCD that isn't already covered by other PRs, as well as similar Firefox desktop flag data.  Fixes #9679.  Blocked by #15698.


html/global_attributes.json
```diff
@@ -1131,22 +1131,9 @@
...
-                  {
-                    "type": "preference",
-                    "name": "dom.forms.inputmode",
-                    "value_to_set": "true"
-                  }
-                ]
-              }
-            ],
+            "firefox_android": {
+              "version_added": "79"
```
foolip, line 1135: Do you know what explains the big difference to the desktop data above?

html/manifest/icons.json
```diff
@@ -19,15 +19,7 @@
...
-              "version_added": "53",
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "manifest.install.enabled",
-                  "value_to_set": "true"
-                }
-              ],
-              "notes": "By default, used for as the source for \"Add to home screen\" (see bug 1234558)."
+              "version_added": false
```
foolip, line 22: Desktop is null above. Should it be?

http/headers/feature-policy.json
```diff
@@ -576,48 +495,22 @@
...
-                    }
-                  ]
-                }
-              ],
+              "firefox": {
+                "version_added": "74",
+                "partial_implementation": true,
+                "notes": [
+                  "Only supported through the <code>allow</code> attribute on <code>&lt;iframe&gt;</code> elements.",
+                  "Before Firefox 80, applying <code>fullscreen</code> to an <code>&lt;iframe&gt;</code> (i.e. via the <code>allow</code> attribute) does not work unless the <code>allowfullscreen</code> attribute is also present."
```
foolip, line 503: The behavior described here sounds indistinguishable from no support before Firefox 80, assuming just the `allowfullscreen` attribute is enough and adding the `allow` attribute doesn't change anything. Can you check where this note came from and if it makes any sense?

Maybe split this out into a separate PR if it needs tweaking beyond the script-made changes.

http/headers/feature-policy.json
```diff
@@ -15,40 +15,16 @@
...
-                    "name": "dom.security.featurePolicy.header.enabled",
-                    "value_to_set": "true"
-                  }
-                ]
-              }
-            ],
+            "firefox": {
+              "version_added": "74",
+              "partial_implementation": true,
+              "notes": "Only supported through the <code>allow</code> attribute on <code>&lt;iframe&gt;</code> elements."
```
foolip, line 21: Sounds to me like no support for the header! Should this be false with notes instead? If so, can you fix in a separate PR?

http/headers/feature-policy.json
```diff
@@ -385,23 +330,11 @@
...
-                  ]
-                }
-              ],
+              "firefox": {
+                "version_added": "74",
+                "partial_implementation": true,
+                "notes": "Only supported through the <code>allow</code> attribute on <code>&lt;iframe&gt;</code> elements."
+              },
               "firefox_android": {
                 "version_added": false
```
foolip, line 406: No Android?

http/headers/feature-policy.json
```diff
@@ -1703,44 +1499,19 @@
...
-                    }
-                  ]
-                }
-              ],
+              "firefox": {
+                "version_added": "81",
+                "partial_implementation": true,
+                "notes": [
+                  "Only supported through the <code>allow</code> attribute on <code>&lt;iframe&gt;</code> elements.",
+                  "Firefox recognizes the <code>web-share</code> permissions policy, but this has no effect in versions of Firefox that do not support the <a href='https://developer.mozilla.org/docs/Web/API/Navigator/share'><code>share()</code></a> method."
```
foolip, line 1507: For a header, no effect means no support, this note should be removed in favor of the correct version then.

http/headers/origin.json
```diff
@@ -23,17 +23,6 @@
...
-                  {
-                    "name": "network.http.sendOriginHeader",
-                    "type": "preference",
-                    "value_to_set": "true"
-                  }
-                ],
-                "notes": "Sent with non-CORS requests since Fx59, unless they are <code>GET</code> or <code>HEAD</code>."
-              },
               {
                 "version_added": true,
```
foolip, line 38: Here's an "ambiguous gap" that the guideline talks about.

javascript/builtins/Atomics.json
```diff
@@ -38,59 +38,18 @@
...
-              {
-                "version_added": "57",
-                "flags": [
-                  {
-                    "type": "preference",
-                    "name": "javascript.options.shared_memory",
-                    "value_to_set": "true"
-                  }
-                ],
-                "notes": "Support was disabled by default to mitigate <a href='https://blog.mozilla.org/security/2018/01/03/mitigations-landing-new-class-timing-attack/'>speculative execution side-channel attacks (Mozilla Security Blog)</a>."
```
foolip, line 50: How about moving this note to the entry that remains? You did this in one case below, let's have the same everywhere! 😀

And let's rephrase the note to talk about removal rather than "disabled by default".

javascript/builtins/DataView.json
```diff
@@ -193,62 +193,20 @@
...
-                    "version_removed": "55",
-                    "flags": [
-                      {
-                        "type": "preference",
-                        "name": "javascript.options.shared_memory",
-                        "value_to_set": "true"
-                      }
-                    ]
+                    "version_removed": "57",
+                    "notes": "Support was disabled by default to mitigate <a href='https://blog.mozilla.org/security/2018/01/03/mitigations-landing-new-class-timing-attack/'>speculative execution side-channel attacks (Mozilla Security Blog)</a>."
```
foolip, line 199: Same comment about the note phrasing here.

foolip (commented):
Some things here will need multiple round trips I think, can you separate out anything that needs manual fixes to keep this PR simple to review? Grouping everything that has to do with SAB being disabled in a single PR would be great.

html/global_attributes.json
queengooborg: It appears that this was added by #6782.  The [linked bug](https://bugzilla.mozilla.org/show_bug.cgi?id=1631681) shows there was an intentional offset between enabling the feature on desktop and mobile.

html/manifest/icons.json
queengooborg: Desktop support is probably `false` as well, but I'm not totally confident on that.  I'd like to hold off digging into it until we commence our no-nulls project for HTML though?

http/headers/feature-policy.json
queengooborg: It appears the notes came from #6493, as well as #12548.

http/headers/feature-policy.json
queengooborg: I opened #15987 because this seems like something we'd have to dig deeper into first.

http/headers/feature-policy.json
queengooborg: Seems that none of the mobile browsers are marked as supported.  I guess that makes sense.

http/headers/origin.json
queengooborg: I mean, technically, we're not _only_ leaving behind a `version_added: true`; we _do_ have `version_added: 70` up above, so this _is_ technically within the guidelines?  😝 

(I definitely would want to try and find a version number for this, but I'm not yet certain how to find version numbers for HTTP headers and I wouldn't want our flag removal to be blocked on it...)

http/headers/feature-policy.json
```diff
@@ -194,7 +170,7 @@
                 "version_added": "79"
               },
               "firefox": {
-                "version_added": "65",
+                "version_added": "74",
```
foolip, line 173: I'm not sure about removing the "dom.security.featurePolicy.header.enabled" flag from the parent but leaving it here with a bumped version. Can you back out these changes and leave them for https://github.com/mdn/browser-compat-data/issues/15987?

http/headers/origin.json
foolip: OK, let's leave this a mystery for the future :)

foolip (approved):
OK to land with Feature-Policy changes split out.

http/headers/feature-policy.json
foolip: OK, maybe this can be simplified then.

foolip (approved):

foolip:
@queengooborg there are more flags that need to be removed, CI is failing.

http/headers/feature-policy.json
queengooborg: I opened #16537 for us to take a look into this at a later time!

http/headers/feature-policy.json
queengooborg: #15987 was opened to look further into these notes.  I agree that these notes imply `version_added: false`, but would not like to have them block this PR and would love to keep the scope of this PR as small as possible.

queengooborg:
CI is now passing since the separated VR changes have been merged, and all of the questions regarding the questionable notes have issues opened for us to look into after this is merged.  This is ready for re-review, I'd say!

queengooborg:
Oh I just realized that an approval has already been given!

PR #15954, opened 2022-04-22
WebAssembly:Exception.stack and `option.traceStack` and platform support

hamishwillee:
1. This adds support for `WebAssembly:Exception.stack`, and  constructor `WebAssembly:Exception` constructor option `option.traceStack`. 
   - These are NOT in the spec yet but they are in [the explainer](https://github.com/WebAssembly/exception-handling/blob/main/proposals/exception-handling/Exceptions.md#api-additions), they are [intended to go into the spec](https://github.com/WebAssembly/exception-handling/issues/201) and they have been implemented in Firefox, Chromium, node, and probably Safari
   - Because they are not in the spec I indicated non-standard - but perhaps I should just put as standard since things are heading that way? I decided not to so that I can add the spec link later when one exists.

2. Support was added for this in chromium was added in version 95 : https://chromestatus.com/feature/4756734233018368

3. Support was added in node for v8 9.5 which corresponds to v17.0.0 : https://v8.dev/blog/v8-release-95#webassembly

### DIFF
--- a/javascript/builtins/webassembly/Exception.json
+++ b/javascript/builtins/webassembly/Exception.json
@@ -8,16 +8,16 @@
             "spec_url": "https://webassembly.github.io/exception-handling/js-api/#runtime-exceptions",
             "support": {
               "chrome": {
-                "version_added": false
+                "version_added": "95"
               },
               "chrome_android": {
-                "version_added": false
+                "version_added": "95"
               },
               "deno": {
                 "version_added": false
               },
               "edge": {
-                "version_added": false
+                "version_added": "95"
               },
               "firefox": [
                 {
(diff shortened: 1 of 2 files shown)


javascript/builtins/webassembly/Exception.json
```diff
@@ -125,17 +125,85 @@
...
+                  "version_added": "95"
                 }
               },
               "status": {
                 "experimental": false,
                 "standard_track": true,
                 "deprecated": false
               }
+            },
+            "options_parameter_traceStack": {
```
hamishwillee, line 140: Note, it occurs to be that this is not necessary, as it is the only option so far. I included it because it is is non-spec. But even if we decide it is spec I'd keep it - since then other options have something to compare again. Up to you though.

queengooborg (approved):
Thanks, this is LGTM!

hamishwillee:
Thanks! Sorry this is coming in dribs and drabs - taking me a fair bit of time to get my head around it.

queengooborg:
Not a problem!  I'd say it's much easier to review a bunch of smaller PRs than it is to review one big PR as it is.  ;)

PR #15966, opened 2022-04-22
Move Prettier test to package.json scripts

### LINKED ISSUE #7407: test-formatting: keep Prettier check as unittest or move to package.json scripts?
Now that we've introduced Mocha as our test suite, it's time for us to convert our code checkers into unit tests.  The question is, should we keep our Prettier check as a unittest, or would it make more sense for us to put this into a script in `package.json`, say `npm run code-format`?

queengooborg:
This PR fixes #7407 by moving the Prettier commands out of `test/` and `scripts/` and into `package.json` scripts.  It doesn't make too much sense to have these commands inside of the linter scripts since they affect the entire repository.


foolip (approved):
I've tested this a bunch locally, introducing formatting errors and both detecting and fixing them with the various commands.

foolip:
@Elchi3 I think this looks good, but it's touching things I haven't worked with much, PTAL?

Elchi3 (approved):
Looks good to me, but normally you find indefinite verbs in package script commands and not progressive verbs, right?

formatting -> format

Elchi3 (approved):

PR #15967, opened 2022-04-22
Account for newly-added "semver-*-bump" labels in release pulls script

queengooborg:
I recently added two new labels to help us plan out releases and semver bumps: "semver-major-bump" and "semver-minor-bump".  This PR updates the according release script to catch these labels as well.

Note: the intention is that "needs-release-note" will go away in favor of these two labels, since it's implied by either label.


queengooborg:
Notice for the other owners: the intention behind these two new labels is to aid in the automation of releases in the near future.  Instead of marking PRs with "needs release note", we can use these two labels to tell our release script automatically when a major or minor bump is needed.

queengooborg:
I'm self-merging this because it fixes critical infrastructure.

PR #15975, opened 2022-04-22
Update Safari versions for api.ByteLengthQueuingStrategy.ByteLengthQueuingStrategy

queengooborg:
This PR updates and corrects the real values for Safari (Desktop and iOS/iPadOS) for the `ByteLengthQueuingStrategy` member of the `ByteLengthQueuingStrategy` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.0.1).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/ByteLengthQueuingStrategy/ByteLengthQueuingStrategy

### DIFF
--- a/api/ByteLengthQueuingStrategy.json
+++ b/api/ByteLengthQueuingStrategy.json
@@ -95,10 +95,12 @@
               "version_added": "43"
             },
             "safari": {
-              "version_added": "10.1"
+              "version_added": "10.1",
+              "version_removed": "15.4"
             },
             "safari_ios": {
-              "version_added": "10.3"
+              "version_added": "10.3",
+              "version_removed": "15.4"
             },
             "samsunginternet_android": {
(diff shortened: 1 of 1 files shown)


foolip:
This is a surprising removal, any idea what happened here?

foolip:
The test seems broken, returning null. This is why I think we should distinguish false and null more in the updater :)

queengooborg:
Ah, I see what's going on: the test returns `null` for three of the four exposures, and `false` for the last, Shared Worker, since shared workers are unsupported.  I'll reopen this when collector v6.0.3 is released, which will include a fix!

foolip:
@queengooborg can you also file a bug about the bug in the updater here? It shouldn't conclude that it's unsupported from those test results, even if the test was accurate.

PR #15976, opened 2022-04-22
Update Firefox versions for CSSNamespaceRule API

queengooborg:
This PR updates and corrects the real values for Firefox and Firefox Android for the `CSSNamespaceRule` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.0.1).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/CSSNamespaceRule


foolip (commented):
Is there a css.* entry that should also be updated?

queengooborg:
There is `css.at-rules.namespace`, but that's marked as supported since Firefox 1, so I don't think it needs to be updated.

foolip:
http://mdn-bcd-collector.appspot.com/tests/api/CSSNamespaceRule is only testing the interface object, so I tried http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=3314 in Firefox 4 and the interface is supported.

Is there support in the collector to write a custom test that depends on parsing a CSS stylesheet like this?

I think the correct support for Firefox here should probably be 1 just like css.at-rules.namespace, but I haven't tested anything older than Firefox 4.

queengooborg:
I opened https://github.com/foolip/mdn-bcd-collector/pull/2080 to add a custom test for this, so that we can get better results!

PR #15989, opened 2022-04-23
Remove individual browser names from browser schema; enforce one browser per JSON file

queengooborg:
This PR removes the individual browser names from the browser schema file.  It's slightly pointless to enforce this another time in the browser files.  This will allow for slightly easier addition and removal of browsers.

Additionally, this PR adds a limit to the number of browsers a JSON file in `browsers/` may have to one, and only one.

Note: while this is a schema change, it is not one that changes consumer data and only enforces trends we've already stuck to.  As such, this can be merged as a part of a patch release.


foolip:
Can you double check if having this in the schema doesn't allow VS Code to autocomplete browser names and flag errors? Maybe we can add something else to keep the browser names in sync, like extracting them from the schema?

queengooborg:
I didn't realize VS Code had such a feature.  Yikes, that makes things significantly harder...  If we want to keep VS Code's schema-validating functionality (I didn't see an autocompletion prompt when trying out VS Code), we'll have to leave that enum in then -- but at least, we can simplify the browser schema.

PR #15990, opened 2022-04-23
Remove release-level "accepts_flags" property from schema

queengooborg:
This PR removes the release-level `accepts_flags` property from browser data from the schema.  This property isn't used for linting and hasn't been adopted in browser data, as we already have a browser-wide `accepts_flags` statement.

Note: this can be included in a patch release as we have no data affected by this change.

### DIFF
--- a/schemas/browsers-schema.md
+++ b/schemas/browsers-schema.md
@@ -71,10 +71,6 @@ The release objects consist of the following properties:
 
 - An optional `release_notes` property which points to release notes. It needs to be a valid URL.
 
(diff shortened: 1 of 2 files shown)


foolip (approved):
Makes sense, but can you find the issue/PR that introduced this and post a comment that this bit is being reverted? Maybe someone had grand plans for this.

queengooborg:
Done!  This was added in by Daniel when adding the `accepts_flags` in general.

queengooborg:
Merging this since approval's been given!

PR #15992, opened 2022-04-23
Remove double spaces from notes

queengooborg:
This PR replaces all double spaces from our notes.

### DIFF
--- a/javascript/builtins/Object.json
+++ b/javascript/builtins/Object.json
@@ -418,7 +418,7 @@
                 {
                   "version_added": "8",
                   "partial_implementation": true,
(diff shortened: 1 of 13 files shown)


javascript/builtins/Object.json
```diff
@@ -418,7 +418,7 @@
                 {
                   "version_added": "8",
                   "partial_implementation": true,
-                  "notes": "In Internet Explorer 8, this was only supported on DOM objects and with some non-standard behaviors.  This was later fixed in Internet Explorer 9."
+                  "notes": "In Internet Explorer 8, this was only supported on DOM objects and with some non-standard behaviors. This was later fixed in Internet Explorer 9."
```
foolip, line 421: This note makes me wonder if there should be an entry for IE9 with full support. But not in this PR, and not high priority.

foolip (approved):

PR #15995, opened 2022-04-23
Adds version 103 entries for some blink browsers

lukewarlow:
#### Summary

Adds version 103 entries for Chrome, Chrome Android, Webview Android and Edge.

Chrome roadmap does include further future releases but don't know if it's worth adding many ahead of time. 

I seem to remember Firefox used to do this but that appears to not be the case anymore.

https://chromestatus.com/roadmap#:~:text=LATER-,Chrome%20103,-Beta%20coming%20May - Chrome 103 release date

### DIFF
--- a/browsers/chrome.json
+++ b/browsers/chrome.json
@@ -709,6 +709,12 @@
           "status": "nightly",
           "engine": "Blink",
           "engine_version": "102"
+        },
+        "103": {
+          "release_date": "2022-06-21",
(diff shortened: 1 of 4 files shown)


queengooborg (approved):
LGTM, thanks!

Sometimes we'll go through and add planned releases for Firefox, but I guess nobody's done so in a long time.

PR #15997, opened 2022-04-24
Remove html.elements.img.intrinsicsize from BCD

queengooborg:
This PR removes the irrelevant `intrinsicsize` member of the `img` HTML element as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-features).  This was only supported in Chromium behind flags removed over two years ago.

Part of work for #15568.


foolip (approved):

foolip:
Approving but not merging, in case MDN changes are also needed. I'll do the same with the others PRs in this series.

queengooborg:
This PR isn't blocked by content but there was some documentation, so I've opened a PR to remove it.  I'll merge this now since the two can happen independently!

PR #15998, opened 2022-04-24
Remove html.elements.video.intrinsicsize from BCD

queengooborg:
This PR removes the irrelevant `intrinsicsize` member of the `video` HTML element as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-features). This was only supported in Chromium behind flags removed over two years ago.

Part of work for #15568.


foolip (approved):

queengooborg:
There's no mention of this attribute in MDN content, so merging this!

PR #16031, opened 2022-04-26
Skip over a failed PR features diff in release notes

queengooborg:
This PR adds a try-catch to our release-notes script to skip over any pull requests that fail feature enumeration.


queengooborg:
I'm self-merging this without review because this is a _critical_ fix for the next release.  The release script is currently crashing because of conflicts around the NodeJS v16 upgrade.

PR #16034, opened 2022-04-26
Restructure events for Navigation APIs to new structure

queengooborg:
This PR is a quick follow-up to #15951 which restructures the event handlers that were added to the new events structure.

CC @Elchi3


Elchi3 (approved):
Thanks for the quick fix! 👍

PR #16042, opened 2022-04-27
Add subfeatures for RTCDataChannel.binaryType values

### LINKED ISSUE #6770: Chromium does not support RTCDataChannel.binaryType = "blob"
Chromium and Chrome browsers do not support `channel.binaryType = "blob"` https://bugs.chromium.org/p/webrtc/issues/detail?id=2276

That is not clear at 
- https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel
- https://developer.mozilla.org/en-US/docs/Web/API/RTCDataChannel/binaryType
where the compatibility data just states

> `binaryType`
> Yes

which is not a complete dataset. 

The table needs to be expanded to include row and column for `"blob"` and `"arraybuffer"` support.

queengooborg:
This PR fixes #6770 by adding subfeatures for the different `binaryType` values of the RTCDataChannel API.

### DIFF
--- a/api/RTCDataChannel.json
+++ b/api/RTCDataChannel.json
@@ -146,7 +146,7 @@
         },
         "blob_value": {
           "__compat": {
-            "description": "<code>blob</code> value",
+            "description": "<code>Blob</code> value",
             "support": {
               "chrome": {
                 "version_added": false,
--- a/api/RTCDataChannel.json
+++ b/api/RTCDataChannel.json
@@ -95,6 +95,109 @@
             "standard_track": true,
             "deprecated": false
           }
+        },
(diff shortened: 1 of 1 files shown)


api/RTCDataChannel.json
```diff
@@ -95,6 +95,109 @@
             "standard_track": true,
             "deprecated": false
           }
+        },
+        "arraybuffer_value": {
+          "__compat": {
+            "description": "<code>arraybuffer</code> value",
```
foolip, line 101:
```suggestion
            "description": "<code>ArrayBuffer</code> value",
```

foolip (approved):

PR #16046, opened 2022-04-27
updating inert

rachelandrew:
Updating `HTMLElement.inert` as Chrome is shipping in 102. Also added `inert` as a global attribute. 

https://chromestatus.com/feature/5703266176335872

### DIFF
--- a/html/global_attributes.json
+++ b/html/global_attributes.json
@@ -1168,14 +1168,7 @@
               ]
             },
             "firefox_android": {
-              "version_added": "81",
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "html5.inert.enabled",
-                  "value_to_set": "true"
-                }
-              ]
+              "version_added": false
             },
             "ie": {
               "version_added": false
--- a/html/global_attributes.json
+++ b/html/global_attributes.json
@@ -1143,6 +1143,83 @@
           }
         }
       },
+      "inert": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/HTML/Global_attributes/inert",
+          "spec_url": "https://html.spec.whatwg.org/multipage/interaction.html#the-inert-attribute",
+          "support": {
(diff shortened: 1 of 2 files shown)


nt1m:
Not sure with safari_ios is set to false, but this is also supported on iOS (same version as desktop).

nt1m:
Also, Firefox recently enabled it by default only on Nightly/early betas: https://bugzilla.mozilla.org/show_bug.cgi?id=1764134 in version 101, but keeps it disabled for release versions, not sure how to reflect that.

rachelandrew:
@nt1m I've update iOS, as far as Firefox goes I think the extra level of detail isn't that useful as in the stable version it's behind a flag, which is probably what people think of looking at  this.

html/global_attributes.json
```diff
@@ -1143,6 +1143,83 @@
...
+            "firefox_android": {
+              "version_added": "81",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "html5.inert.enabled",
+                  "value_to_set": "true"
+                }
+              ]
+            },
```
queengooborg, line 1179: Firefox Android actually doesn't have flag support anymore!

```suggestion
            "firefox_android": {
              "version_added": false
            },
```

queengooborg (changes requested):
Thanks for your PR!  Just a small detail, but this is LGTM otherwise!

queengooborg:
@nt1m If the feature is available in nightly but disabled in beta/stable, we use ["preview" for the version](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#choosing-preview-values).  I'd be happy to review a separate PR to make those changes, or if you'd like to make the changes here @rachelandrew, I'm also happy to merge them!

rachelandrew:
@queengooborg I've committed your change and also made the preview update.

queengooborg (approved):
Thank you, this is LGTM!

PR #16056, opened 2022-04-28
Tweak options for issue template

queengooborg:
Small tweaks to the options available in the "What type of issue is this?" dropdown.  This PR makes it required, but replaces the last option with "Other" since we can't select a default option.


queengooborg:
I'm going to self-merge this because this fixes something that's currently broken.  I'd be more than happy to make changes as requested!

PR #16063, opened 2022-04-29
Sanitizer API fixes for FF and Chrome

hamishwillee:
This fixes some compatibility data:
1. FF does not support  `Sanitize.sanitizeFor()` as was marked - verified by testing and also you can see it is missing in the IDL: https://searchfox.org/mozilla-central/source/dom/webidl/Sanitizer.webidl
2. FF does not support  `Sanitize.getConfiguration()`  - verified by testing and also you can see it is missing in the IDL: https://searchfox.org/mozilla-central/source/dom/webidl/Sanitizer.webidl
3.  Chrome does support  `Sanitize.getConfiguration()`  - verified by testing on browserstack (came in the same version as everything else).

Note that Chrome 102 is supposed to ship this feature: https://chromestatus.com/feature/5786893650231296
However I have not added that as it is a beta and did not appear to be true during initial testing.

### DIFF
--- a/api/Sanitizer.json
+++ b/api/Sanitizer.json
@@ -154,6 +154,76 @@
           }
         }
       },
+      "getConfiguration": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/Sanitizer/getConfiguration",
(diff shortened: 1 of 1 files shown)


queengooborg (approved):
LGTM, thanks!

PR #16064, opened 2022-04-29
MV3 background event pages

rebloor:
#### Summary
MV3 background event pages related changes including:

- note about non-persistent background page support change in Firefox 101, 
- flagged extension.lastError, extension.getBackgroundPage, and runtime.getBackgroundPage as deprecated

Tested and passed `npm test`


webextensions/api/extension.json
```diff
@@ -82,6 +82,11 @@
               "safari_ios": {
                 "version_added": "15"
               }
+            },
+            "status": {
+              "experimental": false,
+              "standard_track": false,
+              "deprecated": true
```
mixedpuppy, line 89: getBackgroundPage is not deprecated

webextensions/api/runtime.json
```diff
@@ -482,6 +482,11 @@
               "safari_ios": {
                 "version_added": "15"
               }
+            },
+            "status": {
+              "experimental": false,
+              "standard_track": false,
+              "deprecated": true
```
mixedpuppy, line 489: getBackgroundPage is not deprecated

webextensions/manifest/background.json
```diff
@@ -61,17 +61,15 @@
               },
               "firefox": {
                 "version_added": "48",
-                "partial_implementation": true,
                 "notes": [
-                  "Only persistent pages are supported.",
+                  "Before version 101, only persistent pages are supported.",
```
mixedpuppy, line 65: persistent pages are not yet supported.  they are pref'd off and available to developers to pref on for testing.  we might want to reword that, and have tickets to return to these items later.  maybe something like:

"Only persistent pages are supported.  As of 101 addon developers may turn on event pages for testing purposes"

mixedpuppy (changes requested):

webextensions/api/extension.json
rebloor: Sorry, must've misunderstood your comment, reverted

webextensions/api/runtime.json
rebloor: Ditto here too

webextensions/manifest/background.json
rebloor: Using the MV3 preference I assume? e.g. "Only persistent pages are supported. As of version 101, extension developers can turn on event pages for testing purposes using the `extensions.manifestV3.enabled` preference."

mixedpuppy: They can also turn on event pages for MV2 using `extensions.eventPages.enabled` and that will be in the blog posts.

mixedpuppy (approved):
r+ with mention of MV2 support pref

PR #15489, opened 2022-03-23
Added browser compat data for priority-hints

pmeenan:
#### Summary
Added browser compatibility data for the [priority hints](https://wicg.github.io/priority-hints/) attributes and properties.

#### Test results and supporting details
[Chromium status for priority hints](https://chromestatus.com/feature/5273474901737472) - enabled by default in 101.

#### Related issues
Part of the documentation addition for https://github.com/mdn/content/issues/14208

PR #16066, opened 2022-04-29
Add mdn links for fetchPriority properties

HolgerJeromin:
#### Summary
This adds urls to the properties which fixes the missing links in the three tables like:

https://developer.mozilla.org/en-US/docs/Web/API/HTMLIFrameElement#browser_compatibility

They were already missing from the first commits in #15489 from @pmeenan


queengooborg (approved):
Thanks for adding these in, this is LGTM!

PR #16071, opened 2022-04-30
Remove irrelevant notes for api.MediaDevices.getDisplayMedia

### LINKED ISSUE #8821: api.MediaDevices.getDisplayMedia - Browser compatibility is not up to date
In Browser compatibility section, there are annotations for mobile Firefox and Android Web View, indicating, that the `getDisplayMedia` API is exposed but always returns NotAllowedError. I checked that in codesandbox linked below, and also done some research. There were some [bugs in implementation in chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=1038244#c2) (link for chromium bug report) as well as in [Firefox](https://bugzilla.mozilla.org/show_bug.cgi?id=1624181) (link for bugzilla). Both bugs were fixed, and I tested that on WebView and Mobile Firefox `getDisplayMedia` in not a function. Although this bug is still present in mobile Chrome.

Remove annotation regarding `getDisplayMedia` API for WebView and mobile Firefox, but create one for mobile Chrome.
I prepared simple site on codesandbox, that checks, if there is `navigator.MediaDevices.getDisplayMedia` of type "function" -> https://codesandbox.io/s/is-getdisplaymedia-enabled-my091 and checked it in deifferent browsers and devices.

<details>

* Query: `api.MediaDevices.getDisplayMedia`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/MediaDevices/getDisplayMedia
* Report started: 2021-01-19T07:54:07.799Z

</details>

queengooborg:
This PR removes the now-irrelevant notes for Firefox Android and WebView Android regarding availability of MediaDevices.getUserMedia().  The method was removed entirely in Chromium 88 and Firefox Android 79, so it isn't separated into multiple support statements to reflect the method's history (especially since it always rejected).  Fixes #8821.

### DIFF
--- a/api/MediaDevices.json
+++ b/api/MediaDevices.json
@@ -195,8 +195,7 @@
               }
             ],
             "firefox_android": {
-              "version_added": false,
-              "notes": "API is available, but will always fail with <code>NotAllowedError</code>."
+              "version_added": false
             },
             "ie": {
               "version_added": false
@@ -217,8 +216,7 @@
               "version_added": false
             },
             "webview_android": {
-              "version_added": false,
-              "notes": "API is available, but will always fail with <code>NotAllowedError</code>."
+              "version_added": false
             }
           },
           "status": {


foolip (changes requested):
These entries should be `partial_implementation` with a `version_added`.

queengooborg:
Since these have been removed though, is it really worth it to try and research a `version_added` for when the method was first added despite it never being supported?

foolip:
@queengooborg the collector is going to want to add back support for this since the API is present, right? When that happens, it's likely we'll have forgotten about this and won't add `partial_implementation` as we should.

queengooborg:
Alright, in that case, I'm going to close this PR.  I won't be able to get results for all older Android browsers just yet.

PR #16072, opened 2022-04-30
Add more options for issue type in "data-problem"

queengooborg:
This PR offers two additional dropdown options for a "data problem" issue's type, namely "missing compatibility data" and "missing specification".  These two are common occurrences.

### DIFF
--- a/.github/ISSUE_TEMPLATE/data-problem.yml
+++ b/.github/ISSUE_TEMPLATE/data-problem.yml
@@ -27,7 +27,7 @@ body:
         - Incorrect support data (ex. Chrome says "86" but support was added in "40")
         - Browser bug (a bug with a feature that may impact site compatibility)
         - Missing compatibility data
-        - Missing specification
+        - Missing specification link (no spec_url property)
         - Other
   - type: textarea
     id: problem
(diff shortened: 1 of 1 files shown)


.github/ISSUE_TEMPLATE/data-problem.yml
```diff
@@ -26,6 +26,8 @@ body:
       options:
         - Incorrect support data (ex. Chrome says "86" but support was added in "40")
         - Browser bug (a bug with a feature that may impact site compatibility)
+        - Missing compatibility data
+        - Missing specification
```
janbrasna, line 30: I'm trying to understand what cases fall under "Missing specification" exactly; maybe a more verbose description would remove this uncertainty?

- The entry doesn't have a `spec_url` but should?
- The entry doesn't have a `mdn_url` but should?
- The entry has `mdn_url` linking to a 404 not-yet-authored page which shloud be removed?
- The entry has `mdn_url` linking to a 404 not-yet-authored page that shloud be written?
- Something completely different and I'm inferring it just wrong…

janbrasna (commented):
@queengooborg 👋 Trying to see it from a new contributor's perspective, I might need more explanation for the last type to apply it correctly:

.github/ISSUE_TEMPLATE/data-problem.yml
queengooborg: Thank you for your feedback!  I've just pushed an update to clarify better that it is referring to the `spec_url` property!

janbrasna: That is the easiest of all the options (:) no varying opinions here;) 💣 thanks for clarification! Much appreciated, this should help lowering decision paralysis for newcomers.

PR #16075, opened 2022-05-01
Fix data-problem template

queengooborg:
This PR fixes the `data-problem` template by moving the `required` parameter to the `validations` sub-parameter as required by GitHub's schema.


queengooborg:
Self-merging this because it fixes a critical infrastructure issue.

PR #16081, opened 2022-05-02
HTMLInputElement.showPicker() - FF101 and OperaAndroid 68

hamishwillee:
[HTMLInputElement.showPicker()](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/showPicker) support added in FF101 in https://bugzilla.mozilla.org/show_bug.cgi?id=1745005

Opera for Android also supports this in v68 (corresponds to Chrome 99 - tested).

This is part of docs work in https://github.com/mdn/content/issues/15407

### DIFF
--- a/api/HTMLInputElement.json
+++ b/api/HTMLInputElement.json
@@ -2155,12 +2155,10 @@
               "version_added": "99"
             },
             "firefox": {
-              "version_added": false,
-              "notes": "See <a href='https://bugzil.la/1745005'>bug 1745005</a>."
+              "version_added": "101"
             },
             "firefox_android": {
-              "version_added": false,
-              "notes": "See <a href='https://bugzil.la/1745005'>bug 1745005</a>."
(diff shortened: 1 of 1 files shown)


queengooborg (approved):

hamishwillee:
Hi @queengooborg Thanks for the review.

I think I'll need to do a bit more with this and would appreciate your advice. The showPicker entry just tells you that this method exists, and is supported for some input elements. Ideally an input has a picker when you touch them, then showPicker should also work - but in practise it is up to the implementer. 

What that means is that there are differences in behaviour between browsers. Specifically, testing shows:
- Firefox supports input types: color, date, datetime-local, file
- Firefox does not support types: month, week, time and also does not support datalist or autocomplete options.
- Chrome supports all of them. 

I think it is probably worth doing all of these as subfeatures of showPicker, because we can't know what different browsers will support. Thoughts?

PR #16092, opened 2022-05-02
Update Edge versions for Ink API

queengooborg:
This PR updates and corrects the real values for Microsoft Edge for the `Ink` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.0.1).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Ink

### DIFF
--- a/api/Ink.json
+++ b/api/Ink.json
@@ -11,7 +11,7 @@
             "version_added": "94"
           },
           "edge": {
-            "version_added": "94"
+            "version_added": "93"
           },
           "firefox": {
             "version_added": false
@@ -58,7 +58,7 @@
               "version_added": "94"
             },
             "edge": {
-              "version_added": "94"
+              "version_added": "93"
             },
             "firefox": {
               "version_added": false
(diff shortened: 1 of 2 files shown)


foolip (commented):
Should https://wicg.github.io/ink-enhancement/#navigator-interface-extensions also be updated?

Also, what happened here, is there anything public about this shipping a version earlier in Edge than Chrome?

queengooborg:
I found a [post on the Microsoft Edge blog](https://blogs.windows.com/msedgedev/2021/08/18/enhancing-inking-on-the-web/) where they announce that the API is available in the latest Dev Channel of Edge on August 18th, which matches their release date of Edge 93 (September 2nd).  It seemed like Microsoft was taking the lead on this API (the spec editor, who also co-authored the blog post, is from Microsoft).  I can't find anything that specifically mentions a version number (not even in release notes), so that article is our best bet for confirmation.

foolip:
Alright, this was not a mistake then!

There's a risk we'll erase this as a likely off-by-one error when fixing mirroring issues, but checking against collector results  should prevent this.

PR #16094, opened 2022-05-02
Update Firefox versions for api.MediaStream.removetrack_event

queengooborg:
This PR updates and corrects the real values for Firefox and Firefox Android for the `removetrack_event` member of the `MediaStream` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.0.1).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/MediaStream/removetrack_event


api/MediaStream.json
```diff
@@ -789,10 +789,10 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": false
+              "version_added": "59"
```
foolip, line 792: This is from https://bugzilla.mozilla.org/show_bug.cgi?id=1425621 but was adding the `onremovetrack` property only. Can you check when the event itself began firing? A `partial_implementation` range is likely needed.

foolip (changes requested):

api/MediaStream.json
foolip: I found https://bugzilla.mozilla.org/show_bug.cgi?id=1592287 but it doesn't really make sense, the event wasn't fired at all for a long time? (It's more common for the event event to come before the handler.)

queengooborg: Interestingly enough, I can't actually get this event to fire...  I tried using the following script:

```js
navigator.mediaDevices.getUserMedia({audio: true}).then((userMedia) => {
    userMedia.addEventListener('removetrack', (e) => {
        console.log('removetrack event:', e);
    });

    userMedia.onremovetrack = (e) => console.log('onremovetrack:', e);

    var tracks = userMedia.getTracks();

    setTimeout(() => {
        userMedia.removeTrack(tracks[0]);
        console.log('track removed');
    }, 500);
});
```

foolip: I think this event might fire only when the UA removes a track and not when it's removed by script. That might happen when a microphone is unplugged, or something like that.

queengooborg: Crap...if it's not an event that can be triggered by a script, I don't think I can test this.  :/

foolip: One tricky thing is how will we avoid updating the data to say "59" based on the collector later? It will be easy to assume that the collector is right. How about a `partial_implementation` entry from 59-72 that says the property is supported but the event never fires?

api/MediaStream.json
```diff
@@ -563,9 +563,16 @@
...
-              "version_added": "59"
-            },
+            "firefox": [
+              {
+                "version_added": "72"
+              },
+              {
+                "version_added": "59",
+                "partial_implementation": true,
+                "notes": "Before Firefox 72, the event handler was exposed but had no effect."
```
foolip, line 573: How about `version_removed` instead? Otherwise it will look like this note still applies to current Firefox.

foolip (changes requested):

foolip (approved):

foolip (approved):

PR #16101, opened 2022-05-02
HTMLInputElement.showPicker - subfeatures for supported inputs

hamishwillee:
FF101 supports [HTMLInputElement.showPicker()](https://developer.mozilla.org/en-US/docs/Web/API/HTMLInputElement/showPicker) but it doesn't support it for every possible input element.

This is in-line with the spec, which states that _ideally_ all input elements that have a picker should work with showPicker, but that this is up to the implementation.

That is still a compatibility issue for developers since there is no way programmatically to work out whether the method will work (or did succeed) for a particular picker.

So what I've done is added a subfeature for each of the types that the spec indicates are normally supported. At the moment Chromium supports all of them and FF supports a subset.

### DIFF
--- a/api/HTMLInputElement.json
+++ b/api/HTMLInputElement.json
@@ -2189,6 +2189,456 @@
             "standard_track": true,
             "deprecated": false
           }
+        },
+        "color_input": {
+          "__compat": {
+            "description": "<code>color</code> input",
+            "support": {
+              "chrome": {
+                "version_added": "99"
+              },
+              "chrome_android": {
+                "version_added": "99"
+              },
+              "edge": {
+                "version_added": "99"
+              },
+              "firefox": {
+                "version_added": "101"
+              },
+              "firefox_android": {
+                "version_added": "101"
+              },
+              "ie": {
+                "version_added": false
+              },
+              "opera": {
+                "version_added": "85"
+              },
+              "opera_android": {
+                "version_added": "68"
+              },
+              "safari": {
+                "version_added": false,
+                "notes": "See <a href='https://webkit.org/b/234009'>bug 234009</a>."
+              },
+              "safari_ios": {
+                "version_added": false,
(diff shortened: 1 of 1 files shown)


hamishwillee:
PS @queengooborg I get this error in NPM - is there any way to fix it?

```
D:\github\mozilla\browser-compat-data>npm run test

> @mdn/browser-compat-data@4.1.18 test
> npm run format && npm run lint && npm run mocha


> @mdn/browser-compat-data@4.1.18 format
> npx prettier --check "**/*.js" "**/*.ts" "**/*.md" "**/*.json"

npm ERR! Unexpected token '.'

npm ERR! A complete log of this run can be found in:
npm ERR!     C:\Users\hamis.000\AppData\Local\npm-cache\_logs\2022-05-02T23_14_53_858Z-debug-0.log
```

queengooborg:
Hmm, the first thing I'd do is check to make sure you're using NodeJS v16 or higher (we recently bumped the contributor requirement up), and that you're using the latest patch version.  It should be 16.13.0, if I recall off the top of my head properly?

hamishwillee:
@queengooborg Thank you - yes 16.3.0, that fixed it. I was running 17.5.0. 

IMO it is a bug in BCD if I can run the tools with an invalid node version and get no warning.

queengooborg:
v17 should work without a problem actually -- in fact, I just tried out 17.5.0 and didn't experience any issues...  Maybe it was some sort of caching issue, and swapping node versions fixed it?

hamishwillee:
Thanks @queengooborg Hmmm. I tried it with v17.4.0 and same problem. So I have no idea.

However, I have a build that works now, so good to go.

Feel free to review this BCD PR :-)

queengooborg (approved):
Thank you, this is LGTM!

PR #16104, opened 2022-05-03
TransformStreamDefaultController is standard

hamishwillee:
[TransformStreamDefaultController](https://developer.mozilla.org/en-US/docs/Web/API/TransformStreamDefaultController) was marked non-standard, but there is no indication of this in the spec https://streams.spec.whatwg.org/#ts-default-controller-class and it is well supported.

I think therefore that this is an error or out of date information. 

@queengooborg Sorry I didn't catch this in #16103 - was more obvious when working through the docs.

Other docs work for this can be tracked in https://github.com/mdn/content/issues/15469#issuecomment-1115528753

### DIFF
--- a/api/TransformStreamDefaultController.json
+++ b/api/TransformStreamDefaultController.json
@@ -70,7 +70,7 @@
         },
         "status": {
           "experimental": false,
-          "standard_track": false,
+          "standard_track": true,
           "deprecated": false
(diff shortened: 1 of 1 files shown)


queengooborg (approved):
Thanks for catching this, this is looking good to me!

PR #16111, opened 2022-05-03
Update Firefox versions for PushSubscription API

queengooborg:
This PR updates and corrects the real values for Firefox and Firefox Android for the `PushSubscription` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.0.1).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/PushSubscription

### DIFF
--- a/api/PushSubscription.json
+++ b/api/PushSubscription.json
@@ -115,10 +115,10 @@
               "version_added": "17"
             },
             "firefox": {
-              "version_added": false
+              "version_added": "96"
             },
             "firefox_android": {
-              "version_added": false
+              "version_added": "96"
             },
             "ie": {
               "version_added": false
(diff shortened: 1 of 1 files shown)


api/PushSubscription.json
```diff
@@ -314,7 +314,7 @@
               "version_added": "17"
             },
             "firefox": {
-              "version_added": "46"
+              "version_added": "44"
             },
             "firefox_android": {
               "version_added": "48"
```
foolip, line 320: Should this also be updated? It didn't match before, but should it?

queengooborg: It might need to; the original data came from the wiki migration.  I'm not currently able to test older Firefox Android versions to confirm or deny this, though.

foolip: Hmm, there's other data for https://w3c.github.io/push-api/ that says Firefox 48, even with some odd notes, so this is probably correct.

foolip:
I've sent https://github.com/mdn/browser-compat-data/pull/16323 to simplify the Firefox Android data.

PR #16127, opened 2022-05-03
Support fetch in Node.js v16 with flag

vsn4ik:
Refs:
- https://nodejs.org/en/blog/release/v16.15.0/
- https://github.com/mdn/browser-compat-data/pull/15272#issuecomment-1115843060


queengooborg (approved):
Thanks for the quick follow-up, this is LGTM!

PR #16130, opened 2022-05-03
Mirror script: mirror from Chrome by default

queengooborg:
This PR updates our mirroring script to mirror Edge's data from Chromium instead of Internet Explorer by default.

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -82,16 +82,16 @@ Many browsers within BCD can be derived from other browsers given they share the
 
 The <browser> argument is the destination browser that values will be copied to. The script automatically determines what browser to copy from based upon the destination (see table below), but manual specification is possible through the `--source=""` argument.
 
-| Destination      | Default Source    |
-| ---------------- | ----------------- |
-| Chrome Android   | Chrome            |
-| Edge             | Internet Explorer |
-| Firefox Android  | Firefox           |
-| Opera            | Chrome            |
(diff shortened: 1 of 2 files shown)


foolip:
I tried running `npm run mirror edge` and it's making changes that can't be inferred from the Chrome data:

```diff
--- a/html/elements/iframe.json
+++ b/html/elements/iframe.json
@@ -1159,7 +1159,7 @@
                 "version_added": true
               },
               "edge": {
-                "version_added": true
+                "version_added": "≤18"
               },
               "firefox": {
                 "version_added": true
```

Both the IE and Chrome data is true for that entry, but ≤18 doesn't follow. There must be some special logic for Edge still causing these incorrect changes.

queengooborg:
This change makes sense and actually isn't incorrect; the mirroring script is configured to use the original data from Edge to make decisions regarding Chrome mirroring, such as this one.  If Chrome's version is lower than 79 or it's `true`, Edge will be `≤79` if it's `null`, `79` if it's false, and `≤18` if it's true.  (Of course, this logic is undesirable when we transition to build-time mirroring, but that's something we can adjust when we get to that point.)

foolip:
I see, so there's an assumption that the original data has some validity, and also that it's old, since that's when true &rarr; ≤18 made sense. I don't think this logic makes sense, but I agree we can clean it up in future iterations.

foolip (approved):

PR #16132, opened 2022-05-03
Revert "Mirror script: mirror from Chrome by default"

foolip:
Reverts mdn/browser-compat-data#16130

### DIFF
--- a/docs/contributing.md
+++ b/docs/contributing.md
@@ -82,16 +82,16 @@ Many browsers within BCD can be derived from other browsers given they share the
 
 The <browser> argument is the destination browser that values will be copied to. The script automatically determines what browser to copy from based upon the destination (see table below), but manual specification is possible through the `--source=""` argument.
 
-| Destination      | Default Source |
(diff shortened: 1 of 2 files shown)


foolip:
@queengooborg I was too fast to review/merge. It seems like the script doesn't work in the common case of mirroring a recent Chrome version to Edge, in a case like this:
```
            "chrome": {
              "version_added": "102"
            },
            "chrome_android": {
              "version_added": "102"
            },
            "edge": {
              "version_added": false
            },
```

Unless there's a quickfix, shall we revert?

foolip:
@queengooborg edit any existing file to have that data and notice how the script doesn't mirror the Edge data.

There's probably data like this already, I noticed this because I was surprised there weren't any changes based on mirror Chrome to Edge in cases similar to the example given.

queengooborg:
I found out what the problem is, and it's the oldest mistake in the book... 🤦‍♀️   Opening up a PR now!

PR #16133, opened 2022-05-03
Mark ::grammar-error and ::spelling-error as experimental

alattalatta:
#### Summary
Mark CSS `::grammar-error` and `::spelling-error` pseudo-elements as experimental. No browsers have implemented them.

#### Related issues
https://github.com/mdn/content/pull/15615

### DIFF
--- a/css/selectors/grammar-error.json
+++ b/css/selectors/grammar-error.json
@@ -46,7 +46,7 @@
             }
           },
           "status": {
-            "experimental": false,
+            "experimental": true,
             "standard_track": true,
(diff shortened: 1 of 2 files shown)


queengooborg (approved):
Our guidelines for experimental status don't really highlight what to do in the case of "no browser support", but I'm inclined to agree to this should be marked as experimental, since only one browser has shown signs of potentially implementing it.  Plus, the content PR was merged that applies the same change, so we should keep the status consistent across MDN!

PR #16139, opened 2022-05-05
Add spec link for ImageBitmapRenderingContext's canvas attribute

foolip:


Elchi3 (approved):
LGTM
Bonus points if you could set experimental to false for everything in this file.

foolip:
@Elchi3 I've claimed the bonus points.

PR #16148, opened 2022-05-06
DOMException is serialisable 

hamishwillee:
DOMException (and other errors) is serialisable in the current HTML spec: https://github.com/whatwg/html/pull/4665.
- Support was added in Chrome 77 - https://chromestatus.com/feature/5745988251680768
- Support was added in FF101 https://bugzilla.mozilla.org/show_bug.cgi?id=1561357

What this PR does is add a "negative subfeature", which is something Daniel suggested for spec updates that change old behaviour. Rather than saying "X supports serialisation from this version" it says "X does supports not_serialisation from first version to version removed". The theory is that after a time you can remove the subfeature and just have the assumption everything supports serialisation.

Anyway, that's what I did for this. The only one I am not sure how to get is deno. 

IFF you are OK with this approach we can merge and I will update the other serializable errors that are [supported by chrome](https://chromestatus.com/feature/5745988251680768) (but not firefox): Error, EvalError, RangeError, ReferenceError, SyntaxError, TypeError, and URIError.

### DIFF
--- a/api/DOMException.json
+++ b/api/DOMException.json
@@ -259,6 +259,67 @@
             "deprecated": false
           }
         }
+      },
+      "not_serializable": {
+        "__compat": {
+          "spec_url": "https://html.spec.whatwg.org/multipage/structured-data.html#serializable-objects",
+          "description": "Not serializable",
+          "support": {
+            "chrome": {
+              "version_added": "1",
+              "version_removed": "77"
+            },
+            "chrome_android": {
+              "version_added": "18",
+              "version_removed": "77"
+            },
+            "deno": {
+              "version_added": "1.0"
+            },
+            "edge": {
+              "version_added": "12",
+              "version_removed": "79"
+            },
+            "firefox": {
+              "version_added": "1",
+              "version_removed": "101"
+            },
+            "firefox_android": {
+              "version_added": "4",
+              "version_removed": "101"
+            },
+            "ie": {
+              "version_added": "9"
+            },
+            "opera": {
+              "version_added": "≤12.1",
+              "version_removed": "64"
+            },
+            "opera_android": {
+              "version_added": "≤12.1",
+              "version_removed": "55"
+            },
+            "safari": {
+              "version_added": "1"
+            },
+            "safari_ios": {
+              "version_added": "1"
+            },
+            "samsunginternet_android": {
+              "version_added": "1.0",
+              "version_removed": "12.0"
+            },
+            "webview_android": {
+              "version_added": "1",
+              "version_removed": "77"
+            }
+          },
+          "status": {
+            "experimental": false,
+            "standard_track": false,
+            "deprecated": false
+          }
+        }
       }
     }
   }


hamishwillee:
@queengooborg FYI :-)

hamishwillee:
PS Is `code` deprecated? It is marked as deprecated in MDN but not in BCD?

queengooborg (changes requested):
Also, thank you for your PR!  However, I don't think that we should add a negative feature.  While I can see the benefit of eventual removal, what I've found is that marking an anti-feature as "supported" has confused readers ("full support for no support").  Additionally, this is a feature that has been added to the spec, rather than removed, so it feels anti-chronological to have a negative feature.

hamishwillee:
@queengooborg 

> PS Is code deprecated? It is marked as deprecated in MDN but not in BCD?

Sorry, I mean is `DOMException.code` deprecated?

hamishwillee:
OK, I've updated to be a normal "positive" feature. FWIW I don't care either way - just following the patterns I've been taught :-)

queengooborg:
> Sorry, I mean is `DOMException.code` deprecated?

Ah, thanks for clarifying!  AFAICT, it doesn't seem that it is deprecated, there's no mention of deprecation in the linked spec.  I could be wrong though.  😛 

> just following the patterns I've been taught :-)

Oddly they're the opposite patterns of what I had been taught, haha!  All good though, thank you for the quick changes!

api/DOMException.json
```diff
@@ -259,6 +259,58 @@
             "deprecated": false
           }
         }
+      },
+      "serializable_object": {
+        "__compat": {
+          "spec_url": "https://html.spec.whatwg.org/multipage/structured-data.html#serializable-objects",
+          "description": "Exception is serializable",
```
queengooborg, line 266:
```suggestion
          "description": "<code>DOMException</code> is serializable",
```

queengooborg (changes requested):
Just one small description nit, LGTM otherwise!

hamishwillee:
@queengooborg OK, so some evidence :-)

[FF14 release notes](https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/14) say:
> [DOMException.code](https://developer.mozilla.org/en-US/docs/Web/API/DOMException) is now deprecated per latest DOM Level 4 specification.

The doc itself says:

> returns a short that contains one of the [error code constants](https://developer.mozilla.org/en-US/docs/Web/API/DOMException#error_names), or 0 if none match. This field is used for historical reasons. New DOM exceptions don't use this anymore: they put this info in the [DOMException.name](https://developer.mozilla.org/en-US/docs/Web/API/DOMException/name) attribute.

The [spec](https://webidl.spec.whatwg.org/#idl-DOMException-error-names) refers to the Legacy code name and value - and it is the value that is in DOMException.code.

So I think that it is the case that this is not something code should rely on, even though the "deprecation" is at least 10 years old.

Thoughts?

hamishwillee:
Cool. Done in https://github.com/mdn/browser-compat-data/pull/16197

PR #16155, opened 2022-05-06
Bump NodeJS in workflow runs

queengooborg:
This PR bumps the NodeJS version used in workflow runs up to NodeJS v18, the now-current version of Node.


.github/workflows/node.js.yml
```diff
@@ -17,7 +17,7 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-node@v2
       with:
-        node-version: 16.x
+        node-version: 18.x
```
caugner, line 20: The `.x` does not seem to be necessary (see [here](https://github.com/actions/setup-node#supported-version-syntax)):

```suggestion
        node-version: 18
```

.github/workflows/release.yml
```diff
@@ -16,7 +16,7 @@ jobs:
     - uses: actions/checkout@v2
     - uses: actions/setup-node@v2
       with:
-        node-version: 16.x
+        node-version: 18.x
```
caugner, line 19:
```suggestion
        node-version: 18
```

queengooborg:
I've now set up a testing matrix to test our code in all LTS versions of NodeJS!

queengooborg:
We could probably set up a matrix to test multiple NodeJS versions, though I'm not able to modify what checks are required so I can't really make that change here.  😛

PR #16165, opened 2022-05-07
Use new, faster HTML validator and parser for notes test

### LINKED ISSUE #16162: HTML validator is slow
As a part of the new linter that was introduced to validate the HTML within notes and descriptions, an HTML validator was introduced.  However, this validator is _slow_ (as in, super slow).  Running `npm run lint api` without the test took 3.73 seconds, while running with the new test took _49.42 seconds_.  (Running the same test without the HTML validator, but leaving the HTML parser, took 3.74 seconds.)

I feel this validation is essential to BCD so I don't think we should disable the linter, but if possible we should see what we can do to reduce execution time.  Maybe this is a good excuse to switch to Markdown for notes and descriptions?

queengooborg:
This PR fixes #16162 by replacing `html-validate` (and `node-html-parser`) with `@desertnet/html-parser`, which is much, much faster than our current setup.

I tried out a couple of different validators and got the following results when testing `api/AbortController.json` (while still using `node-html-parser` for validating individual tags based on our requirements):
Nothing: 0.56s
`html-validate`: 0.83s
`fast-xml-parser`: 0.56s
`node-html-parser`: 0.55s (does not offer description of issues)
`@desertnet/html-parser`: 0.56s

After completely switching to `@desertnet/html-parser` for all steps, running `npm run lint api` took 3.70 seconds (which is slightly lower than the 3.74 seconds it took including the last parser, and the 49.42 seconds it currently takes with our current HTML validator).)


queengooborg:
Oh wow, and on an added note, it seems to catch more than our current setup does!  This is now blocked by #16166.

caugner (approved):
Very nice improvement. For me, it's 5.7s versus 27.0s before. 🚀

PR #3178, opened 2018-12-05
[Chrome] Updates to Symbol.

jpmedley:

PR #16176, opened 2022-05-09
Fix Chromium versions for Symbol.unscopables

foolip:
Confirmed by testing in Chrome 37+38:
https://mdn-bcd-collector.appspot.com/tests/javascript/builtins/Symbol/unscopables

This now matches Array.@@unscopables:
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/@@unscopables#browser_compatibility

Part of https://github.com/mdn/browser-compat-data/issues/7844.


foolip:
Original data came from https://github.com/mdn/browser-compat-data/pull/3178 without evidence. Most likely the chromium-find-releases-static tool was used, since that's the cause of most errors like this.

PR #16180, opened 2022-05-09
Delete FileException.json

teoli2003:
Not on a standard track, deleted from all browsers since Chrome 29. The associated document is going away in mdn/content#15815.

Time to go…


queengooborg (approved):
Always down for removal of irrelevant docs!  👍

PR #16181, opened 2022-05-09
Remove info about WebVR on experimental Chrome branches

bershanskiy:
#### Summary

This data was only relevant to Canary/Dev branches of Chromium and never released, even behind a flag.

See removed comments and the linked blog.

#### Related issues

N/A


queengooborg (approved):

PR #16184, opened 2022-05-09
Update Firefox versions for Cache API

queengooborg:
This PR updates and corrects the real values for Firefox and Firefox Android for the `Cache` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.0.1).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Cache

### DIFF
--- a/api/Cache.json
+++ b/api/Cache.json
@@ -17,11 +17,11 @@
             "version_added": "16"
           },
           "firefox": {
-            "version_added": "39",
+            "version_added": "41",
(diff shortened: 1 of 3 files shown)


foolip (commented):
Can you check where the original data came from and why this doesn't match the global `caches` property?

queengooborg:
The original data came from the wiki migration.  As for the `caches` global, the collector also reports "41" so I've updated them both here!

foolip (approved):

PR #16190, opened 2022-05-09
Update Chromium versions for Clipboard API

queengooborg:
This PR updates and corrects the real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `Clipboard` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.0.1).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Clipboard

### DIFF
--- a/api/Clipboard.json
+++ b/api/Clipboard.json
@@ -6,10 +6,10 @@
         "spec_url": "https://w3c.github.io/clipboard-apis/#clipboard-interface",
         "support": {
           "chrome": {
-            "version_added": "66"
+            "version_added": "61"
           },
           "chrome_android": {
-            "version_added": "66"
+            "version_added": "61"
           },
           "edge": {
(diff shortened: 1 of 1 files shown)


foolip (changes requested):
This doesn't match `navigator.clipboard`, can you check which is correct?

queengooborg:
Apparently they're both right?  The clipboard API was exposed in Chrome 61, but `navigator.clipboard` wasn't around until Chrome 66.

foolip:
OK, so this sounds like a case of an unusable interface being exposed too early, which can easily happen if someone forgets to add runtime flags to every part of an API. Unless there's some way of using the `Clipboard` interface object before Chrome 61, I think we should match the `navigator.clipboard` data.

PR #16199, opened 2022-05-10
Update Chromium versions for api.Element.createShadowRoot

queengooborg:
This PR updates and corrects the real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `createShadowRoot` member of the `Element` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.0.1).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Element/createShadowRoot

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -2230,6 +2230,7 @@
             "chrome": [
               {
                 "version_added": "35",
+                "version_removed": "88",
                 "notes": "In Chrome 45, the ability to have multiple shadow roots was deprecated."
               },
               {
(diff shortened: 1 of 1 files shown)


foolip (commented):
This removal should have gone together with `document.registerElement`, but that says 80. Which is correct?

foolip:
@queengooborg can you double check that it's still enabled in Sauce and what version of Chrome? It's been removed from the Chromium code base entirely, not sure when, but I think it was a while ago...

foolip (approved):

PR #16226, opened 2022-05-10
Remove long-gone HTMLElement.item*

teoli2003:
These properties got removed from Firefox last, and it was in 2016. Time to go.


queengooborg:
Just looked it up now actually -- the individual pages are removed already, there's just some mentions on the main `HTMLElement` page that have yet to be removed.  Nothing that'll cause missing compatibility tables though!

teoli2003:
Yes, these were deleted long ago, if ever written. Also, I wrote a PR earlier today to remove the mentions on HTMLELement. it is already merged but will be live tomorrow only: mdn/content#15855

PR #16231, opened 2022-05-10
Update URLs throughout BCD

bershanskiy:
#### Summary

Update HTTP with HTTPS links they redirect to.

Opened each URL and checked HTTP response codes.
Note: `http://www.w3.org` URLs redirect to HTTPS, but redirect sequence technically contains 307 Temporary Redirect, but I do not think this is a problem because only protocol is changed.

#### Related issues

N/A


queengooborg (approved):
Thanks for updating these links, LGTM!

PR #16232, opened 2022-05-11
Simplify release bundle by removing index.js

queengooborg:
This PR simplifies the released bundles of BCD by ommitting the redundant `index.js`.  In doing so, this will allow for ESM users to import BCD into their projects without having to specify the `data.json` file in their import.  Tested to make sure this approach works in NodeJS v18, v16, v14, v12, and even v4, all successfully importing BCD.

### DIFF
--- a/scripts/release-build.js
+++ b/scripts/release-build.js
@@ -24,12 +24,6 @@ async function writeData() {
   await fs.writeFile(dest, data);
 }
 
-async function writeIndex() {
-  const dest = path.resolve(directory, 'index.js');
-  const content = `module.exports = require("./data.json");\n`;
-  await fs.writeFile(dest, content);
-}
-
 // Returns an array of promises for copying of all files that don't need transformation
 async function copyFiles() {
   for (const file of verbatimFiles) {
@@ -41,7 +35,7 @@ async function copyFiles() {
 
 function createManifest() {
   const full = require('../package.json');
-  const minimal = { main: 'index.js' };
+  const minimal = { main: 'data.json' };
 
   const minimalKeys = [
     'name',
@@ -89,7 +83,6 @@ async function main() {
 
   await writeManifest();
   await writeData();
-  await writeIndex();
   await copyFiles();
 
   console.log('Data bundle is ready');


README.md
```diff
@@ -14,20 +14,58 @@ Read how this project is [governed](GOVERNANCE.md).
...
-## Installation
+## Installation and Import
+
+### NodeJS
 
 You can install `@mdn/browser-compat-data` as a node package.
 
 ```
 npm install @mdn/browser-compat-data
+...or...
```
caugner, line 25: If you add `bash` on line 23, you could make this a comment:

```suggestion
# ...or...
```

README.md
```diff
@@ -14,20 +14,58 @@ Read how this project is [governed](GOVERNANCE.md).
...
-## Usage
+Then, you can import BCD into your project with either `import` or `require()`:
 
 ```js
+// ESM w/ Import Assertions (NodeJS 16+)
+import bcd from '@mdn/browser-compat-data' assert { type: 'json' };
+// ...or...
+// ESM Wrapper for older NodeJS versions (NodeJS v12+)
+import bcd from '@mdn/browser-compat-data/NodeWrapper';
+// ...or...
```
caugner, line 37: I'd add some newlines for readability, potentially omitting the `// or`s:

```suggestion

// ...or...

// ESM Wrapper for older NodeJS versions (NodeJS v12+)
import bcd from '@mdn/browser-compat-data/NodeWrapper';

// ...or...

```

README.md
```diff
@@ -14,20 +14,58 @@ Read how this project is [governed](GOVERNANCE.md).
...
 npm install @mdn/browser-compat-data
+...or...
+yarn add @mdn/browser-compat-data
 ```
 
-## Usage
+Then, you can import BCD into your project with either `import` or `require()`:
 
 ```js
+// ESM w/ Import Assertions (NodeJS 16+)
```
caugner, line 32: Might be better to avoid the `w/` abbreviation, which is not necessarily commonly known in non-English speaking countries like Germany:

```suggestion
// ESM with Import Assertions (NodeJS 16+)
```

scripts/release/build.js
```diff
@@ -43,7 +43,13 @@ async function copyFiles() {
 
 function createManifest() {
   const full = require('../../package.json');
-  const minimal = { main: 'index.js' };
+  const minimal = {
+    main: 'data.json',
+    exports: {
+      '.': './data.json',
+      './NodeWrapper': './nodewrapper.mjs',
```
caugner, line 50: I find the public facing naming a bit surprising. How about just calling it `legacy`?

```suggestion
      './legacy': './legacy.mjs',
```

scripts/release/build.js
```diff
@@ -26,9 +26,9 @@ async function writeData() {
   await fs.writeFile(dest, data);
 }
 
-async function writeIndex() {
-  const dest = path.resolve(directory, 'index.js');
-  const content = `module.exports = require("./data.json");\n`;
+async function writeWrapper() {
+  const dest = path.resolve(directory, 'nodewrapper.mjs');
+  const content = `import fs from 'node:fs';\nconst bcd = JSON.parse(fs.readFileSync(new URL('./data.json', import.meta.url)));\nexport default bcd;\n`;
```
caugner, line 31: Maybe for readability the newlines could be as is:

```suggestion
  const content = `import fs from 'node:fs';
const bcd = JSON.parse(fs.readFileSync(new URL('./data.json', import.meta.url)));
export default bcd;
`;
```

scripts/release/build.js
```diff
@@ -26,9 +26,9 @@ async function writeData() {
   await fs.writeFile(dest, data);
 }
 
-async function writeIndex() {
-  const dest = path.resolve(directory, 'index.js');
-  const content = `module.exports = require("./data.json");\n`;
+async function writeWrapper() {
```
caugner, line 29:
```suggestion
async function writeLegacy() {
```

caugner (changes requested):
LGTM, just some bike-shedding

scripts/release/build.js
queengooborg: Hmm...I'm not a fan of "legacy" as the name, as it may confuse consumers to think that it's some older version of the data itself, rather than a wrapper for certain NodeJS versions.  Perhaps just "wrapper" could work?

queengooborg: Or actually, what about "legacyNode", to make it clear it's for older NodeJS versions and may be removed in a future major release of BCD?

caugner: The good thing about "legacy" is that people can kind of expect that this will go away some time, whereas "wrapper" sounds like a wrapper object or function, and "legacyNode" also sounds like we're importing a "Node" object or something.

How about `forLegacyNode` to signal "the same, but for legacy NodeJS versions"?

queengooborg: Name has been changed accordingly!

PR #16051, opened 2022-04-28
Add missing spec URLs for API features

queengooborg:
This PR supersedes #12709.  This PR adds many new spec URLs to the API data based upon data from `@webref/idl`. 
 After rebasing #12709, I ran the script introduced in #16052 to fix the property sorting and move the spec URLs up in the JSON, then re-ran https://gist.github.com/gsnedders/0a475683482552a47e65d55df7bd5289 to add the missing spec URLs.

Notice: if there are any changes where other properties are moved around, they were not meant to be included this PR and can be reverted.  Reviewers with edit access to BCD are more than welcome to make changes directly to this PR.

I don't expect this PR to be merged in one go due to its sheer size.  Rather, I'm planning to separate this into multiple PRs and use this one as a breakdown to show what changes remain.

PR #16241, opened 2022-05-11
Add spec URLs for service worker APIs

queengooborg:
This PR is a subset of #16051 that adds spec URLs for the service worker APIs.


teoli2003 (approved):
LGTM

PR #16242, opened 2022-05-11
Add spec URLs for AuthenticatorAttestationResponse API

queengooborg:
This PR is a subset of #16051 that adds spec URLs for the AuthenticatorAttestationResponse API.


teoli2003 (approved):
Links are working, leading to the right place.

PR #1255, opened 2018-03-02
Correct ServiceWorkerRegistration

ExE-Boss:
This makes some corrections to the browser compatibility table data for the [`ServiceWorkerRegistration` API](https://developer.mozilla.org/en-US/docs/Web/API/ServiceWorkerRegistration).

PR #16261, opened 2022-05-12
Fix Chrome/Edge versions for showNotification's options.actions parameter

foolip:
Original source: https://github.com/mdn/browser-compat-data/pull/1255

However, both the orignally suggested 47 and the updated 45 are wrong,
and no evidence is given for either version.

This was enabled together with Notifications.maxActions in 48:
https://storage.googleapis.com/chromium-find-releases-static/2a3.html#2a381fb7a46742b05d199c6614624b1bf1432ebb

We can also confirm 48 is correct with this test:
https://mdn-bcd-collector.appspot.com/tests/api/Notification/maxActions

### DIFF
--- a/api/ServiceWorkerRegistration.json
+++ b/api/ServiceWorkerRegistration.json
@@ -653,13 +653,13 @@
             "description": "<code>options.actions</code> parameter",
             "support": {
               "chrome": {
-                "version_added": "45"
+                "version_added": "48"
               },
               "chrome_android": {
-                "version_added": "45"
+                "version_added": "48"
               },
               "edge": {
-                "version_added": "≤79"
+                "version_added": "18"
               },
               "firefox": {
                 "version_added": false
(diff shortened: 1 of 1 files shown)


api/ServiceWorkerRegistration.json
```diff
@@ -652,13 +652,13 @@
...
-                "version_added": "45"
+                "version_added": "48"
               },
               "chrome_android": {
-                "version_added": "45"
+                "version_added": "48"
               },
               "edge": {
-                "version_added": "≤79"
+                "version_added": "18"
```
foolip, line 661: This matches `maxActions` which I also tested in Edge 17/18 using https://mdn-bcd-collector.appspot.com/tests/api/Notification/maxActions to confirm it was added in Edge 18.

Since we don't have the source we can't be certain they were added together, but it's very likely I would say.

queengooborg (approved):

PR #16265, opened 2022-05-12
Use fast-json-stable-stringify for built package

foolip:
The intention of this is to make the output independent of which order
we traverse files, how we apply mixins, etc. With this change, fewer
internal changes will result in changes to the output, and a lack of
changes to the built output can be used to test if internal refactoring
is really a no-op.


queengooborg (approved):
LGTM, thanks!

PR #16266, opened 2022-05-12
Add tests for core of mirroring script (getMatchingBrowserVersion)

foolip:
There are bugs here which the tests make apparent.


scripts/mirror.test.js
```diff
@@ -0,0 +1,141 @@
...
+    assert.equal(match('chrome', '27', 'opera'), '15');
+    assert.equal(match('chrome', '28', 'opera'), '15');
+    assert.equal(match('chrome', '29', 'opera'), '16');
+    assert.equal(match('chrome', '30', 'opera'), '17');
+  });
+
+  it('opera_android', function () {
+    assert.equal(match('chrome_android', '18', 'opera_android'), '14');
+    assert.equal(match('chrome_android', '26', 'opera_android'), '14');
+    assert.equal(match('chrome_android', '27', 'opera_android'), '14'); // wrong, should be 15
```
foolip, line 58: This is wrong because Opera Android 14 uses Blink 26 as the engine, but there's no such thing. Chrome Android 26 used WebKit 537.31.

scripts/mirror.test.js
```diff
@@ -0,0 +1,141 @@
...
+    assert.equal(match('chrome', '19', 'chrome_android'), '25');
+    assert.equal(match('chrome', '24', 'chrome_android'), '25');
+    assert.equal(match('chrome', '25', 'chrome_android'), '25');
+    assert.equal(match('chrome', '26', 'chrome_android'), '26');
+  });
+
+  it('edge', function () {
+    assert.equal(match('chrome', '1', 'edge'), false);
+    assert.equal(match('chrome', '27', 'edge'), false);
+    assert.equal(match('chrome', '28', 'edge'), '79');
```
foolip, line 31: This doesn't make sense and is because early versions of Chrome had WebKit as the engine, not Blink. It isn't sensible to infer Edge 79 from Edge 28, but not from Edge 27. We should treat all Chrome versions <79 the same in terms of what we infer for Edge.

scripts/mirror.test.js
```diff
@@ -0,0 +1,141 @@
...
+    assert.equal(match('chrome_android', '30', 'opera_android'), '18');
+  });
+
+  it('safari_ios', function () {
+    assert.equal(match('safari', '1', 'safari_ios'), '1');
+    assert.equal(match('safari', '1.1', 'safari_ios'), '1');
+    assert.equal(match('safari', '1.2', 'safari_ios'), '1');
+    assert.equal(match('safari', '1.3', 'safari_ios'), '1');
+    assert.equal(match('safari', '2', 'safari_ios'), '1');
+    assert.equal(match('safari', '3', 'safari_ios'), '2'); // wrong, should be 1
```
foolip, line 70: Actually I'm not sure about this. Safari 3 used WebKit 523.10, and Safari iOS 1 used WebKit 522.11, while iOS 2 used 525.18. So the mapping makes sense. But what Safari version did iOS 1 claim to have in the UA string?

scripts/mirror.test.js
queengooborg: Edge mirroring doesn't use `getMatchingBrowserVersion()` and instead uses custom logic that infers ranges based on the original Edge data.  I imagine we'll simplify the code for build-time mirroring, but we should probably test `bumpVersion()` instead of `getMatchingBrowserVersion()`.

foolip: I'll remove this part of the test then.

scripts/mirror.test.js
```diff
@@ -0,0 +1,134 @@
+/* This file is a part of @mdn/browser-compat-data
+ * See LICENSE file for more information. */
+
+import assert from 'node:assert/strict';
+
+import bcd from '../index.js';
+import getMatchingBrowserVersion from './mirror.js';
+
+describe('mirror', function () {
+  const match = (source_browser, source_version, dest_browser) => {
```
queengooborg, line 10: Could we add a `describe('getMatchingBrowserVersion()'), () => {...}` around all this?

queengooborg (approved):
I went ahead and performed the changes I requested because I'm about to make some more mirroring script changes that will impact this function.  Thanks for adding these tests!

PR #16267, opened 2022-05-12
Move layerX/layerY from UIEvent to MouseEvent

foolip:
These are non-standard properties, but if they are standardized it will
be on MouseEvent, see https://github.com/w3c/uievents/issues/135.

Furthermore, it's only on MouseEvent that these properties ever return
something other than zero.

The apparent removal in Chrome 45 was caused by the move from UIEvent to
MouseEvent in Chromium:
https://bugs.chromium.org/p/chromium/issues/detail?id=503276

Since MouseEvent inherits from UIEvent, we can ignore this prototype
move in BCD, per our guidelines:


teoli2003:
We have these two  pages on mdn/content that needs to be adapted/removed:

https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/layerX
https://developer.mozilla.org/en-US/docs/Web/API/UIEvent/layerY

Setting the _Needs content update_ label.

queengooborg (approved):
LGTM, thanks!  Will merge once the content is updated as well.  👍

foolip:
Content update sent in https://github.com/mdn/content/pull/15993.

sideshowbarker (approved):

PR #16269, opened 2022-05-12
Make individual fix scripts non-executable

queengooborg:
This PR makes the individual fix scripts non-executable, so only the main fix script is executable.  There's not really any reason that someone would want to run an individual fix script without running all of them.  Part of work to fix the ESLint errors in #5904.


caugner (approved):
LGTM, but if these files are no longer scripts, might they need to be moved to the `scripts/lib` folder?

queengooborg:
I've got a PR open to reorganize the scripts within BCD appropriately -- these two files are a part of the fix.js script which is why they're sitting in the root, but the other PR I have will move them all into a subfolder!

PR #16271, opened 2022-05-12
Fix ESLint issues

queengooborg:
This PR fixes all remaining ESLint issues throughout the code.  Part of work for #5904.


test/utils.js
```diff
@@ -64,6 +67,7 @@ function indexToPosRaw(str, index) {
     switch (char) {
       case '\n':
         if (prevChar === '\r') break;
+      // eslint-disable-next-line no-fallthrough
```
caugner, line 70: A comment should suffice (see [here](https://eslint.org/docs/rules/no-fallthrough)):
```suggestion
      // fall through
```

test/utils.js
```diff
@@ -3,6 +3,8 @@
 
 'use strict';
 
+/* eslint no-control-regex: 0 */
```
caugner, line 6: Could this be replaced by `// eslint-disable-next-line no-control-regex` right before the `INVISIBLES_REGEXP` (I didn't see other RegExps with control characters in this file.)

caugner (changes requested):
LGTM, just two nits

test/linter/test-browsers-data.js
```diff
@@ -25,8 +24,8 @@ function processData(data, logger) {
 
   for (const status of ['current', 'beta', 'nightly']) {
     const releasesForStatus = Object.entries(releases)
-      .filter(([version, data]) => data.status == status)
-      .map(([version, data]) => version);
+      .filter(([_, data]) => data.status == status)
```
caugner, line 27: Couldn't we just omit `_` here, to avoid the `no-unused-vars` exception in #5904?

```suggestion
      .filter(([, data]) => data.status == status)
```

test/utils.js
queengooborg: I don't know why but I couldn't get it to work before, now I'm having no trouble with it, odd...

test/linter/test-browsers-data.js
queengooborg: Oh right, I forgot you can do that in JavaScript!

PR #16200, opened 2022-05-10
FF101 RTCRtpEncodingParameters.maxFramerate supported

hamishwillee:
FF101 supports `RTCRtpEncodingParameters.maxFramerate` - added by https://bugzilla.mozilla.org/show_bug.cgi?id=1611957

This is blocked because the [WebRTC Extensions](https://w3c.github.io/webrtc-extensions/#dom-rtcrtpencodingparameters-maxframerate) spec is not in W3C. I've created a PR for that, but I really don't know what I'm doing, so help/update appreciated:  https://github.com/w3c/browser-specs/pull/605.

@queengooborg Are you planning on merging in the `RTCRtpEncodingParameters` dictionary (i.e. we don't do dictionaries any more "as a rule"? I ask because this is an update to it and I'm not clear whether this is one of the few dictionaries we decided to keep, or just hasn't been addressed yet. Either way, I'll make this update first, and any structural fixes can come as a post-process.

Other docs work for this can be tracked in https://github.com/mdn/content/issues/15470

PR #16272, opened 2022-05-13
RTCRtpEncodingParameters.maxFramerate - off spec zero value

hamishwillee:
FF101 added support for [RTCRtpEncodingParameters.maxFramerate](https://developer.mozilla.org/en-US/docs/Web/API/RTCRtpEncodingParameters/maxFramerate), which we updated in #16200

This is supposed to be valid for a 0 value, but FF instead sets this as "no limit on frame rate". This is because the upstream libwebrtc library used will crash if a 0 value is passed (see https://bugzilla.mozilla.org/show_bug.cgi?id=1611957#c10)

To fix this I have marked the API as "partial" and added a note.

Related docs work can be tracked in https://github.com/mdn/content/issues/15470

### DIFF
--- a/test/spec-urls.test.js
+++ b/test/spec-urls.test.js
@@ -54,7 +54,7 @@ describe('spec_url data', () => {
       // Remove if it is in the main ECMA spec
       'https://tc39.es/proposal-hashbang/out.html',
 
-      // Remove if https://github.com/w3c/browser-specs/pull/605 is merged
+      // Remove if https://github.com/w3c/webrtc-extensions/issues/108 is closed
       'https://w3c.github.io/webrtc-extensions/',
 
       // Remove if https://github.com/w3c/mathml/issues/216 is resolved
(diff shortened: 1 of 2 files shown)


test/spec-urls.test.js
```diff
@@ -54,7 +54,7 @@ describe('spec_url data', () => {
       // Remove if it is in the main ECMA spec
       'https://tc39.es/proposal-hashbang/out.html',
 
-      // Remove if https://github.com/w3c/browser-specs/pull/605 is merged
+      // Remove if https://github.com/w3c/webrtc-extensions/issues/108 is closed
```
hamishwillee, line 57: FYI what will happen here is that the parameter will move to the supported spec. So when that changes we'd remove the exception and change the spec.

hamishwillee:
@queengooborg Can you please review this when you have a chance.

queengooborg (approved):

PR #16282, opened 2022-05-13
Remove unused function

queengooborg:
This PR removes an unused function that was accidentally re-added between moving a test from the browsers presence tests to the versions tests and moving all functions to top level.


queengooborg:
Actually, I'm self-merging this because it was the result of a bad merge on my behalf.

PR #16287, opened 2022-05-14
Document when to add version_removed to flag data

### LINKED ISSUE #3318: Choose and document a policy for flagged features
It's been suggested a couple of times now that we should discard data about features behind preferences or flags after those features have been turned on by default. We ought to make a decision about this and document the decision.

For example, see @jpmedley's https://github.com/mdn/browser-compat-data/issues/2992#issuecomment-430999779 on Chrome preferences or @chrisdavidmills's PR #3314, where he tried to delete old data involving Firefox preferences.

queengooborg:
This PR introduces a guideline on when to set flag data as `version_removed`, which fixes #3318.

### DIFF
--- a/docs/data-guidelines.md
+++ b/docs/data-guidelines.md
@@ -242,6 +242,15 @@ These conditions represent minimum requirements for the removal of valid flag da
 
 This guideline was proposed in [#6670](https://github.com/mdn/browser-compat-data/pull/6670).
 
+## When to add `version_removed` to flagged support
+
+A `version_removed` should be added to support statements containing flags under one of the following conditions, whichever comes first:
+
+- The browser has enabled the feature or flag by default in a stable release (not beta or nightly).
+- The feature can no longer be enabled or disabled by toggling the flag.
+- The feature has been removed from the browser.
+- The flag has been removed from the browser.
+
 ## Initial versions for browsers
 
 The schema docs list [initial versions](../schemas/compat-data-schema.md#initial-versions) for BCD browsers. These are the earliest possible version numbers allowed to be used.


docs/data-guidelines.md
```diff
@@ -242,6 +242,17 @@ These conditions represent minimum requirements for the removal of valid flag da
 
 This guideline was proposed in [#6670](https://github.com/mdn/browser-compat-data/pull/6670).
 
+## When to add `version_removed` to flagged support
+
+A `version_removed` should be added to support statements containing flags under one of the following conditions, whichever comes first:
+
+- The browser has enabled the feature or flag by default in a stable release (not beta or nightly).
```
foolip, line 249: Can you say what `version_removed` should be in this case?

docs/data-guidelines.md
```diff
@@ -242,6 +242,17 @@ These conditions represent minimum requirements for the removal of valid flag da
 
 This guideline was proposed in [#6670](https://github.com/mdn/browser-compat-data/pull/6670).
 
+## When to add `version_removed` to flagged support
+
+A `version_removed` should be added to support statements containing flags under one of the following conditions, whichever comes first:
```
foolip, line 247:
```suggestion
A `version_removed` should be added to support statements containing flags under one of the following conditions:
```

Or was the "whichever comes first" significant?

foolip (changes requested):

docs/data-guidelines.md
foolip: Can we express this in terms of the earliest version, rather than what comes first in the list? Then the order is no longer significant and it's easier to interpret.

docs/data-guidelines.md
```diff
@@ -242,6 +242,19 @@ These conditions represent minimum requirements for the removal of valid flag da
...
+## When to add `version_removed` to flagged support
+
+A `version_removed` should be added to support statements containing flags under one of the following conditions:
+
+- The browser has enabled the feature or flag by default in a stable release (not beta or nightly).
+- The feature can no longer be enabled or disabled by toggling the flag.
+- The feature has been removed from the browser.
+- The flag has been removed from the browser.
+
+Set `version_removed` to the earliest applicable version from these conditions. For example, if Chrome 90 enabled the flag by default, then the feature could no longer be toggled off by disabling the flag in Chrome 96, and finally the flag was removed by Chrome 98, set `version_removed` to `90` as it is the earliest applicable version.
```
voxpelli, line 254:
```suggestion
Set `version_removed` to the earliest applicable version from these conditions. For example, if Chrome 90 enabled the flag by default, the feature could no longer be toggled off by disabling the flag in Chrome 96, and finally the flag was removed by Chrome 98, set `version_removed` to `90` as it is the earliest applicable version.
```

Not sure the `then` adds much more than confusion (at least for me)

foolip: I didn't see the problem at first, but @voxpelli explained it on chat. I think:

```suggestion
Set `version_removed` to the earliest applicable version from these conditions. For example, if Chrome 90 enabled the flag by default, the feature could no longer be toggled off by disabling the flag in Chrome 96, and finally the flag was removed by Chrome 98, then set `version_removed` to `90` as it is the earliest applicable version.
```

foolip (approved):

Issue #12001, opened 2021-08-15
api.HTMLMediaElement.volume - Not working on iOS


pagedesigner:
#### What information was incorrect, unhelpful, or incomplete?
When Volume is changed it is now working, either its 0 or 1 (100%)

#### What did you expect to see?
Actual volume must be changed with the change of amount in HTMLMediaElement.volume

#### Did you test this? If so, how?
tested on latest version of iOS Safari and Chrome Browsers

<details>
<summary>MDN page report details</summary>

* Query: `api.HTMLMediaElement.volume`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/volume
* Report started: 2021-08-15T08:36:14.623Z

</details>

queengooborg:
This was fixed by #16311!

PR #16311, opened 2022-05-16
Change HTMLMediaElement.volume support for safari_ios to partial_implementation

JoshuaCWebDeveloper:
#### Summary

`HTMLMediaElement.volume` does not work on iOS browsers. While it returns a value and can be set, the value is always 1, and setting the value has no effect on the actual volume of the media.

I've chosen to reflect this by setting `partial_implementation` on the `safari_ios` browser along with clarifying notes.

This app was tested on Safari iOS 11 - 15: https://0680re.csb.app/. The "Mute (Volume)" button attempts to programmatically set the volume to 0. 

There isn't a lot of official documentation around this. There is a reference to this behavior in https://developer.apple.com/library/archive/documentation/AudioVideo/Conceptual/Using_HTML5_Audio_Video/Device-SpecificConsiderations/Device-SpecificConsiderations.html#//apple_ref/doc/uid/TP40009523-CH5-SW1; however, that doc is quite old, and is only concerned with Safari, even though other iOS browsers (like Chrome) also behave the same way.

This issue has been observed by others many times:
https://stackoverflow.com/questions/12553757/html5-ios-volumechange-event-not-fired
https://stackoverflow.com/questions/71308806/ionic-htmlaudioelement-cant-change-volume-on-ios
https://stackoverflow.com/questions/59417075/javascript-detect-if-volume-can-be-set

#### Related issues

This was reported in https://github.com/mdn/browser-compat-data/issues/12001; however, that issue also brings up Chrome on iOS, which this PR doesn't support.

#### Additional Info & Questions

As mentioned above, this limitation on `volume` is an iOS level restriction, and applies to all browsers running on iOS, not just Safari (which is all this PR covers).

The compat data does not track, for example, Chrome on iOS (I'm curious as to the reason for this), so even if this PR were merged, the docs would still have no indication that setting `volume` would not work on Chrome on iOS. I'd like some input on whether I should also (in addition to this PR) open a PR in the MDN docs repo that discusses this behavior of `volume` on iOS more in-depth.

### DIFF
--- a/api/HTMLMediaElement.json
+++ b/api/HTMLMediaElement.json
@@ -3768,7 +3768,13 @@
               "version_added": "3.1"
             },
             "safari_ios": {
-              "version_added": "3"
+              "version_added": "3",
+              "partial_implementation": true,
+              "notes": [
+                "`volume` returns a value and is writeable; however, the value is always 1, and setting a value has no effect on the volume of the media object.",
+                "On iOS 12 and below, setting a value on `volume` has no effect on the value of `volume`.",
+                "On iOS 13 and up, setting a value on `volume` *will* update the value of `volume`, but it will be reset to 1 asynchronously."
+              ]
             },
             "samsunginternet_android": {
               "version_added": "1.0"
--- a/api/HTMLMediaElement.json
+++ b/api/HTMLMediaElement.json
@@ -3771,7 +3771,7 @@
               "version_added": "3",
               "partial_implementation": true,
               "notes": [
-                "`volume` returns a value and is writeable; however, the value is always 1, and setting a value has no effect on the volume of the media object.",
+                "<code>volume</code> returns a value and is writable; however, the value is always 1, and setting a value has no effect on the volume of the media object.",
                 "On iOS 12 and below, setting a value on `volume` has no effect on the value of `volume`.",
                 "On iOS 13 and up, setting a value on `volume` *will* update the value of `volume`, but it will be reset to 1 asynchronously."
               ]
--- a/api/HTMLMediaElement.json
+++ b/api/HTMLMediaElement.json
@@ -3771,7 +3771,7 @@
               "version_added": "3",
               "partial_implementation": true,
               "notes": [
-                "<code>volume</code> returns a value and is writable; however, the value is always 1, and setting a value has no effect on the volume of the media object.",
+                "<code>volume</code> returns a value and is writable; however, the value is always 1, and setting a value has no effect on the volume of the media object."
               ]
             },
             "samsunginternet_android": {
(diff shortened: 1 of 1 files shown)


api/HTMLMediaElement.json
```diff
@@ -3768,7 +3768,13 @@
               "version_added": "3.1"
             },
             "safari_ios": {
-              "version_added": "3"
+              "version_added": "3",
+              "partial_implementation": true,
+              "notes": [
+                "`volume` returns a value and is writeable; however, the value is always 1, and setting a value has no effect on the volume of the media object.",
```
queengooborg, line 3774: Notes should be HTML formatted.  (Also slight spelling fix.)

```suggestion
                "<code>volume</code> returns a value and is writable; however, the value is always 1, and setting a value has no effect on the volume of the media object.",
```

api/HTMLMediaElement.json
```diff
@@ -3768,7 +3768,13 @@
...
+                "On iOS 12 and below, setting a value on `volume` has no effect on the value of `volume`.",
+                "On iOS 13 and up, setting a value on `volume` *will* update the value of `volume`, but it will be reset to 1 asynchronously."
+              ]
```
queengooborg, line 3777: I think we can skip these notes.  With the async reverts back to "1", it would be perceived as "no effect".

queengooborg (changes requested):
Hey there, thank you for your PR and researching this!  I've yet to test it myself and see if it applies to iPadOS as well, but before I do, I have a few changes to request.

(P.S. we don't have a separate column for other iOS and iPadOS browsers because they all end up using Safari, since it's baked into the OS itself.  That is, regardless of whether you download Chrome, Firefox, or any other browser for your iPhone, in the end it's still just Safari!)

queengooborg:
Hey @JoshuaCWebDeveloper, do you plan to return to this PR?

api/HTMLMediaElement.json
JoshuaCWebDeveloper: I've taken them out.

My reasoning for the extra notes was that having the volume synchronously be changed only to be asynchronously reverted cost my quite a bit of time in debugging (it made me think the volume was working fine and that something else was resetting it).

JoshuaCWebDeveloper:
@queengooborg Sorry for dropping this PR, I submitted it before I went on a vacation, and it got lost in the pileup. I've implemented your requested changes, let me know if there is anything more you need from me. (Thanks for the education on iOS requiring Safari btw, reading about that was... a journey, lol.)

queengooborg (approved):
LGTM, thank you!

Issue #16305, opened 2022-05-16
#5904 broke Windows support


saschanaz:
single-quote is unfortunately not supported by npm `scripts` on Windows, as it uses traditional `cmd`.

queengooborg:
Blegh...that's so dumb.  Fix incoming.

PR #16312, opened 2022-05-17
FF101 HTMLMediaElement.canPlayType() and friends av1 codec support

### LINKED ISSUE #16173: AV1 codec strings
### What type of issue is this?

Browser bug (a bug with a feature that may impact site compatibility)

### What information was incorrect, unhelpful, or incomplete?

These methods can take a string that contains the mime type and optionally a codec parameter, and return whether the media type is supported: [`HTMLMediaElement.canPlayType()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLMediaElement/canPlayType), [`MediaCapabilities.decodingInfo()`](https://developer.mozilla.org/en-US/docs/Web/API/MediaCapabilities/decodingInfo) and [`MediaSource.isTypeSupported()`](https://developer.mozilla.org/en-US/docs/Web/API/MediaSource/isTypeSupported). Info on the [codec string here](https://developer.mozilla.org/en-US/docs/Web/Media/Formats/codecs_parameter#av1). 

Previously FF would parse codecs for AVC/H264 and VP9 video formats properly and give a fairly accurate response on support. 
VP9. However for AV1 it would  treat all codec strings starting with "av01" as av1, ignoring all codec configuration options that could be passed. This might affect support for playback, and will in most systems affect hardware decoding support (returned by `MediaCapabilities.decodingInfo()` as the 'powerEfficient' boolean. 

1. FF101 supports the codecs  param for AV1 according to spec. I was wondering if we should add (for the above methods) a subfeature `codecs_param_av1" (and equivalent for AVC/H264 and VP9) to mark that this works now. 

2. Further, chromium supports parsing this with some off-spec behaviours:
   - (a) optional parameters in the codecs string that are supposed to be "all or none", but chromium allows any subset to be specified.
   - (b) unexpected characters in the string are supposed to be interpreted by evaluating the string up to the unexpected character. Chromium instead just errors. 

### What browsers does this problem apply to, if applicable?

Chromium (Chrome, Edge 79+, Opera, Samsung Internet)

### What did you expect to see?

_No response_

### Did you test this? If so, how?

_No response_

https://bugzilla.mozilla.org/show_bug.cgi?id=1757861#c9

https://github.com/mdn/content/issues/15463

### Do you have anything more you want to share?

I'm after advice on the best way to capture this, if at all. 

Item 1 is not IMO a great concern/need to capture. The methods that check for support are guaranteed to return false if it can't be supported, but `true` is always assumed to be a "probably" answer. So the new behaviour is better/correct but nothing will be broken before or after.

Item 2 though is off spec behaviour which might mean a string would work for FF but fail for Chromium and visa versa.

My thinking is we could for all of these methods add a subfeature "codec parameter av1" to indicate broad support and then for chromium have two notes - one  for each of the item "2" cases.

Of course I'd rather do nothing. Testing media support is a pain :-(

### MDN URL

_No response_

### MDN metadata

_No response_

hamishwillee:
@queengooborg This is my attempt to address the av1 codec string compatibility issues in #16173. 

I abandoned attempt to do this as a subfeature. While we can determine when AV1 is supported (and some subset of the codecs using this test file  [this file](https://bug1757861.bmoattachments.org/attachment.cgi?id=9271759)), we actually can't determine when/if a system has these particular behaviours with respect to the codec parameter options (or at least I do not have time/ability to write test code.

So this add notes for just the `canPlayType()` method for FF and Chrome. If you are OK with the approach, I will roll this out to the other chromiums and then the other methods. 

Note, 
- it _might_ be reasonable to add subfeatures just for broad codec/type/container support - e.g. av1 was introduced in Chrome 70. I think that would be somewhat useful to know and could be tested.
- Unlike previous discussions of media features, this is doable because it is specific to a function, not "across the platform". 
- However I am not sure how _useful_ it is to know that we can use this mime type with this message, if we don't know the detail of the codecs. 

Fixes #16173

### DIFF
--- a/api/HTMLMediaElement.json
+++ b/api/HTMLMediaElement.json
@@ -471,14 +471,14 @@
               "version_added": "3",
               "notes": [
                 "<code>codecs</code> string can contain any subset of optional parameters (should be all or none).",
-                "Errors if <code>codecs</code> string contains unexpected characters (should evaluate string up to character).",
+                "Errors if <code>codecs</code> string contains unexpected characters (should evaluate string up to character)."
               ]
             },
             "chrome_android": {
               "version_added": "18",
               "notes": [
                 "<code>codecs</code> string can contain any subset of optional parameters (should be all or none).",
-                "Errors if <code>codecs</code> string contains unexpected characters (should evaluate string up to character).",
+                "Errors if <code>codecs</code> string contains unexpected characters (should evaluate string up to character)."
               ]
             },
             "edge": {
--- a/api/HTMLMediaElement.json
+++ b/api/HTMLMediaElement.json
@@ -468,17 +468,28 @@
           "description": "<code>canPlayType()</code>",
           "support": {
             "chrome": {
-              "version_added": "3"
+              "version_added": "3",
+              "notes": [
+                "<code>codecs</code> string can contain any subset of optional parameters (should be all or none).",
+                "Errors if <code>codecs</code> string contains unexpected characters (should evaluate string up to character).",
+              ]
             },
             "chrome_android": {
-              "version_added": "18"
+              "version_added": "18",
+              "notes": [
+                "<code>codecs</code> string can contain any subset of optional parameters (should be all or none).",
+                "Errors if <code>codecs</code> string contains unexpected characters (should evaluate string up to character).",
+              ]
             },
             "edge": {
               "version_added": "12"
             },
             "firefox": {
               "version_added": "3.5",
-              "notes": "Before Firefox 28, <code>canPlayType()</code> returned <code>probably</code> when asked about WebM audio or video files without the <code>codecs</code> parameter. Since multiple codecs are supported, this is not enough information to determine if a file can be played, so <code>maybe</code> is now correctly returned."
+              "notes": [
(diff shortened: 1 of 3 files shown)


hamishwillee:
PS did the dependencies change again? I'm on node 16.13.0 and I get this error on WIndows:
```
D:\github\mozilla\browser-compat-data>npm run test

> @mdn/browser-compat-data@4.1.20 test
> npm run format && npm run lint && npm run mocha

> @mdn/browser-compat-data@4.1.20 format
> npx eslint --fix '**/*.js' && npx prettier --check "**/*.js" "**/*.ts" "**/*.md" "**/*.json"

Oops! Something went wrong! :(

ESLint: 8.15.0
```

queengooborg:
Sorry about that @hamishwillee, apparently Windows doesn't like single quotes in commands.  (I don't even run Windows anymore, let alone develop on it.  :P)  I just fixed #16305 so a quick rebase should fix the issue!

hamishwillee:
Thanks very much @queengooborg - that fixed it. I won't move away from Windows because I'm not a masochist :-).

Now that's done, I would appreciate your review/comment of the approach I have used so I can roll it out more widely.

hamishwillee:
And ping re ^^^

Now that's done, I would appreciate your review/comment of the approach I have used so I can roll it out more widely.

queengooborg:
Thanks for waiting on my review -- this approach is LGTM!

hamishwillee:
Thanks @queengooborg - rolled out the notes as appropriate to all the affected methods/version. Read for re-review.

queengooborg (approved):

PR #16317, opened 2022-05-17
Update spec URLs for CSPViolationReportBody

sideshowbarker:
These should all be using the URL for the current spec at https://w3c.github.io/webappsec-csp rather than the outdated https://www.w3.org/TR/CSP3 version.

### DIFF
--- a/api/CSPViolationReportBody.json
+++ b/api/CSPViolationReportBody.json
@@ -3,7 +3,7 @@
     "CSPViolationReportBody": {
       "__compat": {
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/CSPViolationReportBody",
-        "spec_url": "https://www.w3.org/TR/CSP3/#cspviolationreportbody",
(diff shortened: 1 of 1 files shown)


api/CSPViolationReportBody.json
```diff
@@ -590,7 +590,7 @@
       "toJSON": {
         "__compat": {
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/CSPViolationReportBody/toJSON",
-          "spec_url": "https://www.w3.org/TR/CSP3/#cspviolationreportbody",
+          "spec_url": "https://w3c.github.io/webappsec-csp/#cspviolationreportbody",
```
foolip, line 593: This `toJSON` entry should be removed, but I'll take care of that as a follow up.

foolip (approved):

api/CSPViolationReportBody.json
foolip: I've filed https://github.com/w3c/webappsec-csp/issues/546, it should be added to the spec.

PR #16324, opened 2022-05-17
index.js: Always crawl relative to BCD root

queengooborg:
This PR fixes an issue with `index.js` where the file crawling doesn't always occur relative to the root of BCD, specifically when importing the development version for other projects such as the mdn-bcd-collector.


queengooborg:
Self-merging since this is breaking critical infrastructure.

PR #16329, opened 2022-05-17
Order properties within schema to match our ordering

queengooborg:
This PR is a minor change that just reorders the properties in the schema to match our own property ordering.


queengooborg:
Self-merging because this is just reordering properties.  :P

PR #16343, opened 2022-05-18
Mirroring script: remove mirroring IE to Edge

queengooborg:
This PR removes the code to mirror Internet Explorer to Edge.  It is no longer maintained as Edge is Chrome-based now.


scripts/mirror.js
```diff
@@ -282,71 +244,55 @@ const bumpChromeAndroid = (originalData, sourceData) => {
...
 
-  if (source == 'ie') {
-    if (sourceData.version_removed && sourceData.version_removed !== null) {
-      newData.version_added = false;
-    } else if (sourceData.version_added !== null) {
-      newData.version_added = sourceData.version_added ? '12' : null;
-    }
+  let chromeFalse =
+    sourceData.version_removed || sourceData.version_added === false;
+  let chromeNull = sourceData.version_added === null;
```
caugner, line 274:
```suggestion
  const chromeFalse =
    sourceData.version_removed || sourceData.version_added === false;
  const chromeNull = sourceData.version_added === null;
```

scripts/mirror.js
```diff
@@ -282,71 +244,55 @@ const bumpChromeAndroid = (originalData, sourceData) => {
...
+    } else if (compareVersions.compare(sourceData.version_added, '79', '<=')) {
+      newData.version_added =
+        originalData.version_added === null ? '≤79' : '79';
+    } else {
+      newData.version_added = sourceData.version_added;
     }
   }
 
+  let newNotes = updateNotes(sourceData.notes, /Chrome(?! ?OS)/g, 'Edge');
+  if (newNotes) newData.notes = newNotes;
```
caugner, line 314: Maybe just preference, but I think the most common JS coding standards prefer blocks here:
```suggestion
  if (newNotes) {
    newData.notes = newNotes;
  }
```

scripts/mirror.js
```diff
@@ -282,71 +244,55 @@ const bumpChromeAndroid = (originalData, sourceData) => {
 /**
  * @param {SupportStatement} originalData
  * @param {SupportStatement} sourceData
```
caugner: I'm not exactly sure what these contain. Is `sourceData` the _latest_ support statement of the browser from which we're mirroring?

scripts/mirror.js
```diff
@@ -302,71 +264,57 @@ const bumpChromeAndroid = (originalData, sourceData) => {
...
-      newData.notes = newNotes;
+    if (originalData.version_added && !originalData.version_removed) {
+      newData.version_removed = '79';
+    }
+  } else if (chromeNull) {
+    newData.version_added = null;
+  } else {
+    if (sourceData.version_added === true) {
+      newData.version_added =
+        originalData.version_added === true ? '≤18' : true;
```
caugner, line 304: Not entirely sure here (and below).

If Chrome currently supports this feature, but we don't know in which version it was introduced ...

... why do we set `≤18` if we know that Edge currently supports it?

Does this assume that `sourceData.version_added === true` means it was introduced in a very old version of Chrome, because newer versions of Chrome have much more accurate and complete BCD data?

scripts/mirror.js
```diff
@@ -302,71 +264,57 @@ const bumpChromeAndroid = (originalData, sourceData) => {
...
+    } else if (compareVersions.compare(sourceData.version_added, '79', '<=')) {
+      newData.version_added =
+        originalData.version_added === null ? '≤79' : '79';
+    } else {
+      newData.version_added = sourceData.version_added;
```
caugner, line 309: This means for features introduced in Chrome 80 or later, the feature got introduced in the exact same Edge version!?

scripts/mirror.js
```diff
@@ -282,71 +244,55 @@ const bumpChromeAndroid = (originalData, sourceData) => {
...
+    sourceData.version_removed || sourceData.version_added === false;
+  let chromeNull = sourceData.version_added === null;
 
-    if (sourceData.notes) {
-      newData.notes = updateNotes(
-        sourceData.notes,
-        /Internet Explorer/g,
-        'Edge',
-      );
+  if (chromeFalse) {
```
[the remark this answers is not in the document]
queengooborg: This is for either Chrome no longer supporting it, or never having supporting it in the first place.

scripts/mirror.js
queengooborg: Precisely, yep!

scripts/mirror.js
```diff
@@ -282,71 +244,55 @@ const bumpChromeAndroid = (originalData, sourceData) => {
...
-      updateNotes(sourceData.notes, /Chrome/g, 'Edge'),
-      originalData.notes,
-    );
-
-    if (newNotes) {
-      newData.notes = newNotes;
+    if (originalData.version_added && !originalData.version_removed) {
+      newData.version_removed = '79';
+    }
+  } else if (chromeNull) {
```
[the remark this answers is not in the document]
caugner: And this refers to the "most recent" support statement, right?

scripts/mirror.js
caugner: I see. And it's no longer as useful, because we no longer add `version_added: true` to Edge features?

scripts/mirror.js
caugner: Probably some concise code comments for each of the non-trivial branches would be helpful, even if those branches get removed eventually.

queengooborg: Would it be alright to actually hold off on that for now?  These lines of code aren't new, just unindented from the existing lines.  (I'm also planning on following this up _immediately_ with removing usage of the original Edge data, so the comments would last a day or two.  :P)

caugner (approved):
Haven't understood every bit, but LGTM and some parts of it will vanish anyways (so I heard) 😉

scripts/mirror.js
caugner: 👍

PR #16347, opened 2022-05-18
Mirroring script: account for ranges in getMatchingBrowserVersion()

queengooborg:
This PR updates the mirroring script to properly account for ranges.  If the source data contains a range as the version number, then the mirror result will also output a range.


scripts/mirror.js
```diff
@@ -56,37 +56,45 @@ const { argv } = require('yargs').command(
...
- * @param {ReleaseStatement} source_release
+ * @param {string} destBrowser
+ * @param {ReleaseStatement} sourceRelease
+ * @param {string} sourceVersion
  * @returns {ReleaseStatement|boolean}
  */
-const getMatchingBrowserVersion = (dest_browser, source_release) => {
-  const browserData = browsers[dest_browser];
+const getMatchingBrowserVersion = (
+  destBrowser,
```
caugner, line 65: How about just `browser`, like the CLI parameter?

scripts/mirror.js
```diff
@@ -56,37 +56,45 @@ const { argv } = require('yargs').command(
...
+  sourceBrowser,
+  sourceVersion,
```
caugner, line 67: Maybe it wouldn't hurt here either to use an object parameter `source: { browser, version }` instead?

scripts/mirror.js
```diff
@@ -350,7 +358,8 @@ const bumpOpera = (originalData, sourceData, source) => {
   if (typeof sourceData.version_added === 'string') {
     newData.version_added = getMatchingBrowserVersion(
       'opera',
-      browsers[source].releases[sourceData.version_added],
+      source,
```
caugner, line 361: Since you're already renaming variables, might it be a good opportunity to rename `source` to `sourceBrowser` here as well?

caugner (changes requested):

PR #16348, opened 2022-05-18
FullscreenOptions in element.requestFullscreen isn't supported on Firefox

makotokato:
#### Summary
Gecko doesn't support `FullscreenOptions` yet.
See https://searchfox.org/mozilla-central/rev/8fe6930c0832009b3162bebee7d4ede1a4c8c9a8/dom/webidl/Element.webidl#309

WPT such as http://wpt.live/fullscreen/api/element-request-fullscreen-options.html is failure.

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -5905,10 +5905,10 @@
                 "version_added": "79"
               },
               "firefox": {
-                "version_added": "64"
+                "version_added": false
               },
               "firefox_android": {
-                "version_added": "64"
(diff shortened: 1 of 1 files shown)


queengooborg (approved):
LGTM, thank you!  Confirmed lack of support in Firefox 64 and 100 with the test.  Looks like the original data came from #2861, but the bug doesn't mention this option so it looks like a probable mistake.

PR #16350, opened 2022-05-18
Rename blockMany/blockList in version test

queengooborg:
This PR renames `blockMany` and `blockList` in the version tests to something that's more self-explanatory.

### DIFF
--- a/test/linter/test-versions.js
+++ b/test/linter/test-versions.js
@@ -50,7 +50,7 @@ for (const browser of Object.keys(browsers)) {
 }
 
 /** @type {string[]} */
-const blockMany = [
+const realValuesTargetBrowsers = [
   'chrome',
   'chrome_android',
   'edge',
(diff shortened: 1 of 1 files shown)


test/linter/test-versions.js
```diff
@@ -50,7 +50,7 @@ for (const browser of Object.keys(browsers)) {
 }
 
 /** @type {string[]} */
-const blockMany = [
+const realValuesTargetBrowsers = [
```
caugner, line 53: Is this essentially `browsersRequiringNumericalVersions`? Or what does the "real" refer to?

test/linter/test-versions.js
```diff
@@ -66,15 +66,15 @@ const blockMany = [
 ];
 
 /** @type {object.<string, string[]>} */
-const blockList = {
-  api: blockMany,
-  css: blockMany,
+const realValuesRequired = {
```
caugner, line 69: Ditto, is this bascially `browsersRequiringNumericalVersionsBySection` (assuming that "api", "css" etc. are sections, not sure if this is the right term)?

test/linter/test-versions.js
caugner: Related question: Why do onlyu some browsers require "real" values?

queengooborg: That's correct -- we've used the terms "real" and "nonreal" to describe exact version numbers and boolean/null values respectively throughout BCD!

PR #16358, opened 2022-05-19
Link Chromium/WebKit bugs for selectionchange events

foolip:


foolip:
@tkent-google FYI, this is why I asked about a bug in https://github.com/w3c/selection-api/pull/141.

saschanaz:
Should it say the event directly goes to `document` without bubbling on those browsers? (Try `data:text/html,<body onselectionchange="console.log(event)"><input onselectionchange="console.log(event)">`)

foolip:
That's a good idea, yes! How about this note?

> A <code>selectionchange</code> event is fired on <code>document</code>, see <a href='https://developer.mozilla.org/docs/Web/API/Document/selectionchange_event'><code>Document</code>'s <code>selectionchange</code> event</a>. See bug 1234 for firing the event on <code>&lt;input></code> elements.

queengooborg (approved):

saschanaz:
Well, we haven't added the notes...

queengooborg:
Ah my bad, I didn't see the thread before...

saschanaz:
Mind opening a new PR?

foolip:
@queengooborg I couldn't see any open PR from you, so I rolled the changes into https://github.com/mdn/browser-compat-data/pull/16359.

PR #16359, opened 2022-05-19
Adapt selectstart/selectionchange events to new event structure

foolip:
Per https://w3c.github.io/selection-api/#selectstart-event the event target of "selectstart" events is a Node, which will be either an Element or a Text node.

Per https://w3c.github.io/selection-api/#selectionchange-event the event target can be Document, HTMLInputElement or HTMLTextareaElement.

Part of https://github.com/mdn/browser-compat-data/issues/7545.


foolip:
https://mdn-bcd-collector.appspot.com/tests/api/HTMLElement/selectstart_event was used to confirm support in the following releases:

- Chrome 15
- IE 6
- Firefox 52 (not 51)
- Safari 4

The earliest versions of Chrome, IE and Safari can't be confirmed with BrowserStack, so the existing data is assumed to be correct. IE had to be bumped to 5 since Node itself wasn't supported before then. The event itself might have been, but it's not worth trying to capture this.

foolip:
http://mdn-bcd-collector.appspot.com/tests/api/Document/selectionchange_event was used to confirm support in the following releases:

- Chrome 15
- IE 6
- Firefox 52 (not 51)
- Safari 5.1 (not 4)

The collector doesn't have these tests, but they would be needed to keep this data up-to-date in the future:
http://mdn-bcd-collector.appspot.com/tests/api/HTMLTextAreaElement/selectionchange_event
http://mdn-bcd-collector.appspot.com/tests/api/HTMLInputElement/selectionchange_event

queengooborg (approved):
This is LGTM, just waiting on the content updates!

PR #16376, opened 2022-05-20
Marking IE11 as retired

captainbrosset:
The [ie.json](https://github.com/mdn/browser-compat-data/blob/main/browsers/ie.json) browser file still says that IE11 is "current".

This commit changes the status to "retired" given the [formal desktop app retirement coming up on June 15 2022](https://docs.microsoft.com/en-us/lifecycle/announcements/internet-explorer-11-end-of-support).

This PR should, in theory, wait until June 15 to get merged, although I don't think there's a risk with merging it earlier than this.

### DIFF
--- a/browsers/ie.json
+++ b/browsers/ie.json
@@ -56,7 +56,7 @@
         },
         "11": {
           "release_date": "2013-10-17",
-          "status": "current",
+          "status": "retired",
           "engine": "Trident",
           "engine_version": "7.0"
         }


queengooborg (approved):
Thank you for this PR!  I cannot wait for the day we no longer need to worry about IE.

I'm going to hold of on merging this until we approach the official date, however.  While there's not really much risk in merging this a month early, I know that some consumers *do* still care about IE's data.

saschanaz:
Perhaps mark this as a draft to prevent accidental merge?

queengooborg:
Since we're a little closer to the EOL date, I'm going to go ahead and merge this -- it will be included in a release a week early, but I think that's fine.  Thank you for your PR!

PR #16377, opened 2022-05-20
fix: Change link to schema usage example

BANOnotIT:
#### Summary
Old example link of regex_token points to css.properties.color while there's no usage


queengooborg:
Hi there, thank you for your PR!  We're actually about to remove the `matches` object from BCD, though, see #15781.

PR #16378, opened 2022-05-20
Do not use original data to influence mirroring output

queengooborg:
This PR removes the `originalData` variable from the mirroring script.  This was only utilized by Edge's mirroring to aid the transition to Chromium Edge, but now that the migration is complete, we can remove it and greatly simplify our code.

### DIFF
--- a/scripts/mirror.js
+++ b/scripts/mirror.js
@@ -156,20 +156,13 @@ const copyStatement = (data) => {
 };
 
 /**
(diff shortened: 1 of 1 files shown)


queengooborg:
@caugner As promised, the PR that removes a lot of that code that should have been described in comments before!  😛

caugner (approved):
Didn't run it, but LGTM

PR #16379, opened 2022-05-21
Mirroring script: improve note mirroring

queengooborg:
This PR updates the mirroring script to update notes within arrays, as well as update the version numbers within notes.

### DIFF
--- a/scripts/mirror.js
+++ b/scripts/mirror.js
@@ -299,7 +299,7 @@ const bumpGeneric = (sourceData, destination, source, notesRepl) => {
     );
   }
 
-  if (notesRepl && typeof sourceData.notes === 'string') {
+  if (notesRepl && sourceData.notes) {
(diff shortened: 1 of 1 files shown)


lucalves (approved):

scripts/mirror.js
```diff
@@ -160,22 +160,26 @@ const combineNotes = (notes1, notes2) => {
  * @param {string|string[]|null} notes
  * @param {RegExp} regex
  * @param {string} replace
+ * @param {Function} versionBumpFunction
```
caugner, line 163: How about:

```suggestion
 * @param {Function} versionMapper - Receives the source browser version and returns the target browser version.
```

caugner (approved):
LGTM, but I'd rename the mapper function.

Note: The PR description only mentions the Array change, not the version mapping.

PR #16382, opened 2022-05-21
Update position.json

Keyacom:
#### Summary

Added info in `.../position_sticky_table_elements/__compat/support/firefox/notes` about a bug.

[Bug 1727594](https://bugzilla.mozilla.org/show_bug.cgi?id=1727594).

### DIFF
--- a/css/properties/position.json
+++ b/css/properties/position.json
@@ -171,6 +171,7 @@
               },
               "firefox": {
                 "version_added": "59"
+                "notes": "Borders do not display on table headers if <a href='https://developer.mozilla.org/docs/Web/CSS/border-collapse'><code>border-collapse</code></a> is set to <code>collapse</code> (<a href='https://bugzil.la/1727594'>bug 1727594</a>)."
               },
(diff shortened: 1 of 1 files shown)


css/properties/position.json
```diff
@@ -171,6 +171,7 @@
               },
               "firefox": {
                 "version_added": "59"
```
queengooborg, line 173:
```suggestion
                "version_added": "59",
```

queengooborg (changes requested):
Hey there, thank you for your PR, and welcome to BCD!  Before I properly review your PR, I noticed that there's a syntax error in the JSON.

css/properties/position.json
Keyacom: Sorry, I just didn't notice it, but thank you anyways.

queengooborg (approved):
Thanks for the fixes!

I think we should also add the note to Firefox Android -- I'm seeing no indicating that the issue isn't also on Android.

PR #16388, opened 2022-05-22
Fix scripts/fix failures

saschanaz:
It seems `es-main` cannot correctly detect index.js resolved by the directory path and returns false. But not sure it's intended to be a module, does it really need `esMain()`?

### DIFF
--- a/package.json
+++ b/package.json
@@ -64,7 +64,7 @@
     "format": "npx eslint \"**/*.js\" && npx prettier --check \"**/*.js\" \"**/*.ts\" \"**/*.md\" \"**/*.json\"",
     "format:fix": "npx eslint --fix \"**/*.js\" && npx prettier --write \"**/*.js\" \"**/*.ts\" \"**/*.md\" \"**/*.json\"",
     "lint": "node test/lint",
-    "fix": "npm run format:fix && node scripts/fix",
+    "fix": "npm run format:fix && node scripts/fix/index.js",
     "mirror": "node scripts/mirror",
     "stats": "node scripts/statistics",
     "release-build": "node scripts/release/build",
--- a/scripts/fix/index.js
+++ b/scripts/fix/index.js
@@ -66,5 +66,4 @@ if (esMain(import.meta)) {
   }
 }
 
-module.exports = load;
 export default load;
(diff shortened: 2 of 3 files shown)


scripts/fix/index.js
```diff
@@ -67,3 +67,4 @@ if (esMain(import.meta)) {
 }
 
 module.exports = load;
```
queengooborg, line 69:
```suggestion
```

package.json
```diff
@@ -64,7 +64,7 @@
     "format": "npx eslint \"**/*.js\" && npx prettier --check \"**/*.js\" \"**/*.ts\" \"**/*.md\" \"**/*.json\"",
     "format:fix": "npx eslint --fix \"**/*.js\" && npx prettier --write \"**/*.js\" \"**/*.ts\" \"**/*.md\" \"**/*.json\"",
     "lint": "node test/lint",
-    "fix": "npm run format:fix && node scripts/fix",
+    "fix": "npm run format:fix && node scripts/fix/index.js",
```
queengooborg, line 67: Does this only apply to the fix script?  What about the lint script?

queengooborg (changes requested):
The purpose behind `es-main` is to replace `require.main === module`, because as surprising as it may sound, ESM doesn't actually have a proper built-in way to do this!  So yes, `esMain()` is needed in the script to ensure that we're actually _running_ the script, rather than importing it.

package.json
saschanaz: [`stripExt(modulePath) === scriptPath`](https://github.com/tschaub/es-main/blob/8a569c0ca2345221e62a1dedeb00f9a29301178f/main.js#L23), so that one is okay.

saschanaz:
> So yes, `esMain()` is needed in the script to ensure that we're actually _running_ the script, rather than importing it.

I mean, is it intended to be imported by something else? At least nothing inside BCD does that.

package.json
queengooborg: Ah right, thank you!

queengooborg (approved):

saschanaz:
Anyway, filed https://github.com/tschaub/es-main/issues/22

PR #16393, opened 2022-05-23
Add "upstream" property to the browser schema

queengooborg:
This PR adds an "upstream" property to the browser data, which indicates the upstream browser.  This is copied directly from the `getSource()` function of the mirroring script, which has now been refactored to utilize this new property.

### DIFF
--- a/test/linter/test-browsers-data.js
+++ b/test/linter/test-browsers-data.js
@@ -5,6 +5,9 @@ import chalk from 'chalk-template';
 
 import { Logger } from '../utils.js';
 
+import bcd from '../../index.js';
+const { browsers } = bcd;
+
 /**
  * @typedef {import('../../types').Identifier} Identifier
  * @typedef {import('../utils').Logger} Logger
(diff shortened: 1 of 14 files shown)


scripts/mirror.test.js
```diff
@@ -7,136 +7,108 @@ import getMatchingBrowserVersion from './mirror.js';
...
-      assert.equal(match('chrome_android', '25', 'webview_android'), false);
-      assert.equal(match('chrome_android', '26', 'webview_android'), false);
-      assert.equal(match('chrome_android', '27', 'webview_android'), false);
-      assert.equal(match('chrome_android', '28', 'webview_android'), '4.4');
-      assert.equal(match('chrome_android', '36', 'webview_android'), '37');
-      assert.equal(match('chrome_android', '37', 'webview_android'), '37');
-    });
+    for (const [browser, versionMap] of Object.entries(mappings)) {
+      it(browser, () => {
+        for (const pair of versionMap) {
```
caugner, line 108: The downside of this is loop is that it makes it harder to investigate a failure, because the line is the same for all test cases.

I think jest allows parameterised tests or giving each case a description. Not sure if it's worth it?

caugner: I think a combination of [`describe.each`](https://jestjs.io/docs/api#describeeachtablename-fn-timeout) and [`it.each`](https://jestjs.io/docs/api#testeachtablename-fn-timeout) would work here.

caugner: This should do:

```js
describe.each(Object.entries(mappings))('%s', (browser, pairs) => {
  it.each(pairs)('%s, %s', (a, b) => assert.equal(getMatchingBrowserVersion(browser, a, b));
});

File: test/linter/test-browsers-data.js [hunk 2]
```diff
@@ -19,21 +22,36 @@ function processData(data, logger) {
...
+  if (browserData.upstream) {
+    if (browserData.upstream === browser) {
+      logger.error(
+        chalk`{red The upstream for {bold ${browser}} is set to itself.}`,
+      );
+    }
+
+    if (!Object.keys(browsers).includes(browserData.upstream)) {
+      logger.error(
+        chalk`{red The upstream for {bold ${browser}} is an unknown browser (${browserData.upstream}).}`,
```
caugner, line 51: Would it make sense to add a "Must be one of: ..." tip?

For example, if I write `Chrome` instead of `chrome`, this would lead me in the right direction.

types.d.ts
```diff
@@ -55,6 +55,11 @@ export interface BrowserStatement {
    */
   type: BrowserTypes;
 
+  /**
+   * The upstream browser
+   */
+  upstream?: string;
```
caugner, line 61: Oh, this makes me think that that the `meta: { version }` is not yet declared here. 🤔

caugner (approved):
LGTM, I just have a suggestion for the tests, and an unrelated remark about the types.

scripts/mirror.test.js
queengooborg: It looks like Mocha doesn't have `.each()` (we're not using Jest here), but that code looks much nicer and simpler!

Would you mind opening up an issue about switching from Mocha to Jest?  I'm not opposed to switching over to maintain tooling consistency, but that's beyond the scope of this PR.  :P

PR #16396, opened 2022-05-23
Mirroring script: remove executability

queengooborg:
This PR removes the mirroring script's executability, making it an import-only module.  This is part of work to prepare for build-time mirroring.


queengooborg:
For other reviewers who may not be familiar with the end goal: the reason I'm removing the script's executability is to prevent its continued use as a script to directly modify BCD.  This will allow for a much easier transition to the build-time mirroring, because we won't need to undo all the mirroring-based changes.

foolip:
Aren't there people who use this script on a regular basis still, to update some browser data?

queengooborg:
Actually, I just finished the build-time mirroring!  I'll close this PR in favor of a more complete changeset.

PR #16401, opened 2022-05-23
Implement build-time mirroring

queengooborg:
This PR introduces build-time mirroring to BCD, so that contributors may simply specify `"mirror"` as the value, and the data will be mirrored from its upstream counterpart.

Example:
```js
"support": {
  "chrome": {
    "version_added": "66"
  },
  "chrome_android": "mirror", // will become { version_added: "66" }
}
```

In doing so, this also removes the executability of the mirroring script to discourage its standalone use. All new data to mirror should use the new `"mirror"` value, rather than be performed manually.


queengooborg:
> Aren't there cases around copying notes that aren't done correctly yet

Note mirroring was all fixed up in #16379 actually!

> and where manual mirroring + fixups are needed?

If manual intervention is required, then we wouldn't use mirroring anyways.

.github/workflows/release.yml
```diff
@@ -20,7 +20,7 @@ jobs:
           registry-url: 'https://registry.npmjs.org/'
       - run: npm ci
       - run: npm test
-      - run: npm run release-build
+      - run: npm run build
```
caugner, line 23: Is it worth extracting this rename out of this (rather big) PR?

docs/contributing.md
```diff
@@ -76,33 +76,11 @@ If the feature you're interested in is a JavaScript API, you can cross-reference
...
 
-### Optional: Generating data using the mirroring script
+### Optional: Generating data by mirroring
 
-Many browsers within BCD can be derived from other browsers given they share the same engine, for example Opera derives from Chrome, and Firefox Android derives from Firefox. To help cut down time working on copying values between browsers, a mirroring script is provided. You can run `npm run mirror <browser> <feature_or_file> -- [--source=""] [--modify="nonreal"] [--target_version=""]` to automatically copy values.
+Many browsers within BCD can be derived from other browsers given they share the same engine, for example Opera derives from Chrome, and Firefox Android derives from Firefox. To help cut down time working on copying values between browsers, contributors may specify a value in the data to mirror the data for browsers from their upstream counterparts. See the [schema documentation](../schemas/compat-data-schema.md#mirroring-data) for more info.
```
caugner, line 81:
```suggestion
Many browsers within BCD can be derived from other browsers given they share the same engine, for example Opera derives from Chrome, and Firefox Android derives from Firefox. To help cut down time working on copying values between browsers, contributors may specify a special value in the data to automatically mirror the data from their upstream counterparts. See the [schema documentation](../schemas/compat-data-schema.md#mirroring-data) for more info.
```

docs/contributing.md
```diff
@@ -76,33 +76,11 @@ If the feature you're interested in is a JavaScript API, you can cross-reference
...
-
-By default, the mirroring script will only overwrite values in the destination that are `true` or `null`, but can take a `--modify` (or `-m`) argument to specify whether to overwrite values that are `false` as well (`--modify=bool`), or any values (`--modify=always`).
-
-The script can also take a `--target_version` (or `-t`) argument to only perform mirroring if the data affects a specific browser release. This is used to update data when a new Opera or Opera Android version is released, and we wish to update data that only affects the new version.
+Note: originally, this functionality used to be provided as an executable script. However, because the script must be run manually, this meant that mirrored data would become stale rapidly. The proposal to move the mirroring into a build step was proposed in [#15083](https://github.com/mdn/browser-compat-data/issues/15083) to reduce maintenance time.
```
caugner, line 83:
```suggestion
Note: originally, this functionality used to be provided as an executable script. However, because the script had to be run manually, this meant that mirrored data would become stale rapidly. It was proposed in [#15083](https://github.com/mdn/browser-compat-data/issues/15083) to move the mirroring into a build step to reduce maintenance time.
```

schemas/compat-data-schema.md
```diff
@@ -133,6 +133,23 @@ The currently accepted browser identifiers should be declared in alphabetical or
 
 Desktop browser identifiers are mandatory, with the `version_added` property set to `null` if support is unknown.
 
+#### Mirroring data
+
+Most of the browsers are derivatives of other browsers, such as mobile counterparts (Firefox -> Firefox Android) or other forks (Chrome -> Edge, Opera, Samsung Internet). Usually in such cases, the support for a feature is the exact same across the derivatives, or will be when a matching version is released. To make maintenance easier, contributors may specify a simple string, `"mirror"`, as the support statement for the browser, and the version data will be mirrored from its upstream counterpart (as defined in `browsers/<browser>.json`).
+
+An example of this would be the following:
+
+```json
```
caugner, line 142: To avoid the highlighted syntax error:

```suggestion
```js
```

schemas/compat-data-schema.md
```diff
@@ -133,6 +133,23 @@ The currently accepted browser identifiers should be declared in alphabetical or
...
+"support": {
+  "chrome": {
+    "version_added": "66"
+  },
+  "chrome_android": "mirror", // will become { version_added: "66" }
+}
+```
+
+This also helps with maintaining derivatives with a different release schedule than its upstream counterpart. For example, let's say that a new feature was introduced in Chrome 100, but the latest Samsung Internet release is based on Chrome 96. Rather than set Samsung Internet to `{version_added: false}` and then follow up to update the data when a new version is released, we can set it to `"mirror"` instead, which will automatically change to the version number of the new, matching release.
```
caugner, line 151:
```suggestion
This also helps with maintaining derivatives with a different release schedule than their upstream counterpart. For example, let's say that a new feature was introduced in Chrome 100, but the latest Samsung Internet release is based on Chrome 96. Rather than setting Samsung Internet to `{version_added: false}` and then following up to update the data when a new version is released, we can set it to `"mirror"` instead, which will automatically change to the version number of the new, matching release.
```

scripts/release/build.test.js
```diff
@@ -1,26 +0,0 @@
-/* This file is a part of @mdn/browser-compat-data
```
caugner, line 1: No more build test?

types.d.ts
```diff
@@ -146,10 +146,14 @@ export interface ReleaseStatement {
...
  * is only one of them, the array must be omitted.
+ *
+ * Internally, "mirror" may also be used to represent that the data
+ * should be mirrored from its upstream counterpart.
  */
 export type SupportStatement =
   | SimpleSupportStatement
-  | SimpleSupportStatement[];
+  | SimpleSupportStatement[]
+  | 'mirror';
```
caugner, line 156: Is "mirror" in the exported/released data though?

caugner (approved):
LGTM, just one question about the types.

PS: Didn't verify whether `scripts/release/mirror.js` and `scripts/mirror.js` are the same.

scripts/release/build.test.js
queengooborg: The build test is to ensure that the build output and the repository JSON exactly match one another, but since we're now introducing a new step to modify the data...well, this test becomes irrelevant.  😛

types.d.ts
queengooborg: It's not, but I'm adding it in here to prepare for our own internal TypeScript migration!

queengooborg:
Note: the mirroring script is _not_ exactly the same between the two versions. I removed the executable bits from the mirroring script in this PR, and introduced a new function to handle performing the mirroring for the build step, but otherwise it's the same!

scripts/release/build.js
```diff
@@ -18,6 +21,22 @@ const verbatimFiles = ['LICENSE', 'README.md', 'index.d.ts', 'types.d.ts'];
...
   const { default: bcd } = await import('../../index.js');
+
+  const walker = walk(undefined, bcd);
+
+  for (const feature of walker) {
+    for (const [browser, supportData] of Object.entries(
+      feature.compat.support,
+    )) {
+      if (supportData === 'mirror') {
+        feature.data.__compat.support[browser] = bumpData(
```
foolip, line 32: The name `bumpData` doesn't really make sense in this context I think. Maybe `mirrorData` or `mirrorSupport`?

scripts/release/mirror.js
```diff
@@ -0,0 +1,361 @@
...
+
+  return newData;
+};
+
+const bumpData = (destination, data) => {
+  const upstream = browsers[destination].upstream;
+  return bumpVersion(data[upstream], destination);
+};
+
+export default bumpData;
```
foolip, line 361: Since this is the main entry point, can you add tests for it? In particular the `combineStatements` simplification seems important to test.

scripts/release/mirror.js
```diff
@@ -0,0 +1,361 @@
...
+      delete currentStatement.notes;
+    }
+    newData.push(currentStatement);
+  }
+
+  if (newData.length === 1) {
+    return newData[0];
+  }
+
+  // Remove duplicate statements and statements that are only version_added = false
```
foolip, line 201: It looks like there's nothing in `combineStatements` that handles the case where the derived `version_added` and `version_removed` are the same version, which can happen. Such a statement should be removed, and this seems like the place to do it.

foolip (commented):
A post-merge review. In addition to the inline comments, I think it's very important that we test/lint the built results, since we can otherwise have the mirroring step introduce errors which are currently not possible.

I would also like the mirroring script to fail in more cases where it's not safe to make the assumptions that are being made. Right now, it always does something, because it previously relied on a human review step. Without that, we'll have bad data due to mirroring bugs that we will not catch.

PR #16372, opened 2022-05-20
HTMLCanvasElement.mozGetAsFile - removed from all platforms

hamishwillee:
As discussed in https://github.com/mdn/content/issues/16152#issuecomment-1129140999 , `HTMLCanvasElement.mozGetAsFile` has been removed in FF102. It was  behind a flag since FF101 and is not present in any other platform. Therefore it is now old enough (2 years) to remove from BCD.

Other docs work for this can be tracked in https://github.com/mdn/content/issues/16152

PR #16404, opened 2022-05-24
Enforce NodeJS versions that support import assertions for contributors

queengooborg:
This PR updates the contributor engine version to enforce versions of NodeJS that support import assertions (https://nodejs.org/api/esm.html#import-assertions), which should help some contributors troubleshoot issues in their setup.

### DIFF
--- a/package-lock.json
+++ b/package-lock.json
@@ -40,7 +40,7 @@
         "yargs": "~17.5.0"
       },
       "engines": {
-        "node": ">=16.0.0"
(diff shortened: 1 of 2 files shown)


hamishwillee:
We probably should update this to 16.15.0 - since 16.14.0 did not work on windows or ubuntu for me:   https://github.com/mdn/browser-compat-data/pull/16372#issuecomment-1136535629

PR #16406, opened 2022-05-24
Prepare for TypeScript migration

queengooborg:
This PR introduces all of the changes needed to prepare for TypeScript migration.  A side effect of this PR is fixing the TypeScript declarations exported to consumers.


package.json
```diff
@@ -31,6 +31,7 @@
     "@babel/eslint-parser": "~7.17.0",
     "@babel/plugin-syntax-import-assertions": "~7.17.12",
     "@desertnet/html-parser": "~1.0.1",
+    "@types/node": "~17.0.34",
```
caugner, line 34: I think this should be ^16, because that's the minimal required Node version in this project.

caugner (approved):

package.json
queengooborg: I didn't realize the package's versions correlate with major NodeJS versions!

PR #16409, opened 2022-05-24
scheduler global property

hamishwillee:
The prioritised scheduling API was mostly present in BCD and I added FF101 updates in #16177.

BUT, access to the API is via the `scheduler` property on the global object - e.g. `this` or `window` on main thread. This adds BCD entry for the global scheduler.

@queengooborg Question for you. The spec actually adds this to [WindowOrWorderGlobalScope](https://wicg.github.io/scheduling-apis/#sec-patches-html-windoworworkerglobalscope) mixin which we don't document. One place obviously that goes is Window (as documented here). But where do I add the worker entries both for BCD and docs - this could be the top level scope objects (all of them) or somewhere else. Advice appreciated!

### DIFF
--- a/api/_globals/scheduler.json
+++ b/api/_globals/scheduler.json
@@ -1,0 +1,65 @@
+{
+  "api": {
+    "scheduler": {
+      "__compat": {
+        "mdn_url": "https://developer.mozilla.org/docs/Web/API/Window/scheduler",
+        "spec_url": "https://wicg.github.io/scheduling-apis/#dom-windoworworkerglobalscope-scheduler",
+        "support": {
+          "chrome": {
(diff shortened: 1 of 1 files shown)


queengooborg:
Things added to `WindowOrWorkerGlobalScope` should actually be added under `_globals` as per https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#global-apis, would you mind putting it there?

hamishwillee:
@queengooborg Thanks. Fixed! Also added myself a note to check the data guidelines next time I get stuck. Ready for re-review.

queengooborg (approved):

PR #16411, opened 2022-05-24
FF101 CSSStyleSheet supports constructible stylesheets

hamishwillee:
FF101 CSSStyleSheet supports constructible stylesheets in https://bugzilla.mozilla.org/show_bug.cgi?id=1644102

What that essentially means is that `layout.css.constructable-stylesheets.enabled` is now true by default. I've therefore added a new version for support on the three affected methods for firefox and also for firefox android. I've marked the old version with the flag as version removed.

Other docs work can be tracked in https://github.com/mdn/content/issues/16624#issuecomment-1135790000


hamishwillee:
FYI @queengooborg Untested, for obvious reasons. Should be good I hope. Back on Friday.

api/CSSStyleSheet.json
```diff
@@ -63,19 +63,24 @@
...
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "layout.css.constructable-stylesheets.enabled",
-                  "value_to_set": "true"
-                }
-              ]
-            },
+            "firefox": [
+                "version_added": "101"
```
queengooborg, line 67:
```suggestion
              {
                "version_added": "101"
```

hamishwillee:
Thanks @queengooborg  - I merged your suggestions. Ready for a re-review :-)

api/CSSStyleSheet.json
```diff
@@ -63,19 +63,25 @@
...
+                    "type": "preference",
+                    "name": "layout.css.constructable-stylesheets.enabled",
+                    "value_to_set": "true"
+                  }
+                ]
+              }
+            ],
             "firefox_android": {
-              "version_added": false
+              "version_added": 101
```
hamishwillee, line 84:
```suggestion
              "version_added": "101"
```

hamishwillee:
@queengooborg Since you looked at this last I have made another update. My research showed that the `adoptedStyleSheets` property of Document and ShadowRoot is part of the constructible stylesheets story. I did a test and demonstrated that this is also hidden/exposed by the preference `layout.css.constructable-stylesheets.enabled`.

I have therefore updated the value to show that it is supported in FF based on the same versions/prefs as the other affected items. I have not verified this precise feature went in at the same version as the preference, but I think it a very reasonable assumption.

queengooborg (approved):

PR #16415, opened 2022-05-25
Allow optional `implementation_url` in support data

### LINKED ISSUE #126: Support linking to bugs/issues
On https://developer.mozilla.org/en-US/Add-ons/WebExtensions/Browser_support_for_JavaScript_APIs it would be very helpful to have links to bugs that have been filed for issues.

queengooborg:
This PR supersedes #7396 and fixes #126.  This PR updates the schema to allow an `implementation_url` property, which contains a link to a commit/changeset or a bug for that browser.


schemas/compat-data.schema.json
```diff
@@ -63,6 +63,24 @@
             "required": ["type", "name"]
           }
         },
+        "implementation_url": {
```
caugner, line 66: To safe space and to be consistent with `spec_url`, I would suggest a shorter name like `implement_url` or `impl_url`.

caugner (approved):
LGTM, even though the attribute seems a bit long to me.

PR #16417, opened 2022-05-25
Use "mirror" for AbortController data

foolip:
This uses the build-time mirroring recently implemented:
https://github.com/mdn/browser-compat-data/pull/16401

The resulting build/data.json is unchanged.

### DIFF
--- a/api/AbortController.json
+++ b/api/AbortController.json
@@ -8,9 +8,7 @@
           "chrome": {
             "version_added": "66"
           },
-          "chrome_android": {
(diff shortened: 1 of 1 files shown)


queengooborg (approved):
Confirmed the result doesn't change using `npm run diff b4ea4cd68168d84e38ca2ea56d5ad5fb1a69c337 927479112281bd9f23d16d66fc157d8a1c541965` with the diff updates in #16434!

PR #16429, opened 2022-05-25
Add linter for obsolete data

bershanskiy:
#### Summary

Adds linter which checks for guideline called [Removal of irrelevant features](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-features).

It is a linter, not data.

#### Related issues

N/A

Please note: currently this linter checks only the original two rules:
- a feature was implemented and has since been removed from all browsers dating back two or more years ago.
I'm not sure how to interpret the third rule:
- a feature is unsupported in all releases in the past five years.

As written, the third rule seems to imply the second rule: if a feature meets the third rule -- it hasn't been supported for 5 years --  then it seems to me it meets the second rule -- it might have been implemented but hasn't been supported for 2 years.

### DIFF
--- a/test/linter/test-obsolete.js
+++ b/test/linter/test-obsolete.js
@@ -33,23 +33,31 @@ function neverImplemented(support) {
   return true;
 }
 
+/**
+ * @param {*} browsers
+ * @param {*} support
+ * @returns 'warning' | 'error' | false
+ */
 function implementedAndRemoved(browsers, support) {
+  let result = 'error';
+  const warningTime = new Date().getTime() - 2 * 365 * 24 * 60 * 60 * 1000;
+  const errorTime = new Date().getTime() - 2.5 * 365 * 24 * 60 * 60 * 1000;
   for (const browser in support) {
     let data = support[browser];
     if (!Array.isArray(data)) data = [data];
     for (const d of data) {
+      // Feature is still supported
       if (!d.version_removed) return false;
-      if (
-        d.version_removed &&
-        new Date(
-          browsers[browser].releases[d.version_removed].release_date,
-        ).getTime() >
-          new Date().getTime() - 2 * 365 * 24 * 60 * 60 * 1000
-      )
-        return false;
+      const releaseDate = new Date(
+        browsers[browser].releases[d.version_removed].release_date,
+      ).getTime();
+      // Feature was recently supported, no need to show warning
+      if (warningTime < releaseDate) return false;
+      // Feature was supported sufficiently recently to not show an error
+      if (errorTime < releaseDate) result = 'warning';
     }
   }
-  return true;
+  return result;
 }
 
 /**
@@ -67,7 +75,6 @@ function processData(logger, data, browsers, path) {
     const abandoned = status && status.standard_track === false;
     const rule1Fail = abandoned && neverImplemented(support);
     if (rule1Fail && !shouldFail) {
-      console.error(path);
       logger.error(
         chalk`feature was never implemented in any browser and the specification has been abandoned.`,
       );
@@ -76,8 +83,7 @@ function processData(logger, data, browsers, path) {
 
     const rule2Fail = implementedAndRemoved(browsers, support);
     if (rule2Fail && !shouldFail) {
-      console.error(path);
-      logger.error(
+      logger[rule2Fail](
         chalk`feature was implemented and has since been removed from all browsers dating back two or more years ago.`,
       );
       return;
--- a/test/linter/test-obsolete.js
+++ b/test/linter/test-obsolete.js
@@ -5,6 +5,12 @@ import chalk from 'chalk-template';
 import bcd from '../../index.js';
 const { browsers } = bcd;
 
+/**
+ * List of feature keys which are expected to fail this test.
+ * This list is needed for two reasons:
+ *  - some entries indeed should be removed (like api.VR*)
+ *  - some entries are currently just stubs and should be expanded and will pass test later
+ */
 const exceptions = [
   'api.VRDisplay.hardwareUnitId',
   'api.VREyeParameters.recommendedFieldOfView',
@@ -24,7 +30,7 @@ const exceptions = [
   'svg.elements.view.zoomAndPan',
 ];
 
-function neverImplemented(support) {
+export function neverImplemented(support) {
   for (const s in support) {
     let data = support[s];
     if (!Array.isArray(data)) data = [data];
@@ -33,6 +39,11 @@ function neverImplemented(support) {
   return true;
 }
 
+const errorTime = new Date(),
+  warningTime = new Date();
+errorTime.setFullYear(errorTime.getFullYear() - 2.5);
+warningTime.setFullYear(warningTime.getFullYear() - 2);
+
 /**
  * @param {*} browsers
  * @param {*} support
@@ -40,8 +51,6 @@ function neverImplemented(support) {
  */
 function implementedAndRemoved(browsers, support) {
   let result = 'error';
-  const warningTime = new Date().getTime() - 2 * 365 * 24 * 60 * 60 * 1000;
-  const errorTime = new Date().getTime() - 2.5 * 365 * 24 * 60 * 60 * 1000;
   for (const browser in support) {
     let data = support[browser];
     if (!Array.isArray(data)) data = [data];
@@ -50,7 +59,7 @@ function implementedAndRemoved(browsers, support) {
       if (!d.version_removed) return false;
       const releaseDate = new Date(
         browsers[browser].releases[d.version_removed].release_date,
-      ).getTime();
+      );
       // Feature was recently supported, no need to show warning
       if (warningTime < releaseDate) return false;
       // Feature was supported sufficiently recently to not show an error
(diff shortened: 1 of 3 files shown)


test/linter/test-obsolete.js
```diff
@@ -0,0 +1,103 @@
...
+    let data = support[browser];
+    if (!Array.isArray(data)) data = [data];
+    for (const d of data) {
+      if (!d.version_removed) return false;
+      if (
+        d.version_removed &&
+        new Date(
+          browsers[browser].releases[d.version_removed].release_date,
+        ).getTime() >
+          new Date().getTime() - 2 * 365 * 24 * 60 * 60 * 1000
```
queengooborg, line 41: Could we actually hard-code a date?  Performing math to calculate two years difference is more than alright, but I wouldn't want to have the lint suddenly fail on `main` and cause test failures in others' PRs!

queengooborg (changes requested):
Thank you for doing this!  I haven't fully reviewed this PR (I'm on my phone),  but I have one comment first.

By the way, regarding that third rule, I realized the wording is confusing, it took me a while to realize what it's for.  The rule is a clause for IE data: although the feature is supported in IE11, the "current" version, since IE has not gotten an update in over five years and all other browsers don't support it, we would be able to remove it based upon this clause!

bershanskiy:
> By the way, regarding that third rule, I realized the wording is confusing, it took me a while to realize what it's for. The rule is a clause for IE data: although the feature is supported in IE11, the "current" version, since IE has not gotten an update in over five years and all other browsers don't support it, we would be able to remove it based upon this clause!

I believe this rule was introduced in #6429, however, I still do not fully understand the clause even after reading the lined discussion. Since IE data might be [deprecated soon](https://github.com/mdn/browser-compat-data/issues/7458#issuecomment-1136854990), we probably can ignore that clause for now. I'll implement it if MDN team decides to keep IE data around.

queengooborg:
Warnings sound good to me!  Perhaps we should get https://github.com/openwebdocs/project/issues/79 implemented first in this case, though?

bershanskiy:
> Warnings sound good to me!

I'll use warnings then.

> Perhaps we should get https://github.com/openwebdocs/project/issues/79 implemented first in this case, though?

Thanks for letting me know about this project. I don't think this particular test needs to wait for a new linter, but we will be able to migrate this test to a new linter easily.

queengooborg:
Oops, I performed a semi-merge and didn't realize until I did that this was another's PR, sorry about that!

bershanskiy:
I still need to implement this:
> Also, we can use two-tier messaging: warnings for features that were obsolete for more than 2 years but less than 2.5 years and errors for features that have been obsolete for over 2.5 years.

bershanskiy:
> Oops, I performed a semi-merge and didn't realize until I did that this was another's PR, sorry about that!

I just force-pushed a rebase, you rebase is available on [this branch](https://github.com/bershanskiy/browser-compat-data/tree/linter-obsolete-semi).

queengooborg:
> Also, we can use two-tier messaging: warnings for features that were obsolete for more than 2 years but less than 2.5 years and errors for features that have been obsolete for over 2.5 years.

The next-gen linter has been implemented now, and as a part of it, I implemented issue levels, including warnings.  All you need to do is call `logger.warning()` instead!

test/linter/test-obsolete.js
```diff
@@ -0,0 +1,110 @@
...
+      logger.error(
+        chalk`Please remove this file from list of exceptions: ${path}`,
+      );
+    }
+  }
+}
+
+export default {
+  name: 'Obsolete',
+  description: 'Test for osolete data in each support statement',
```
queengooborg, line 102:
```suggestion
  description: 'Test for obsolete data in each support statement',
```

test/linter/test-obsolete.js
```diff
@@ -0,0 +1,110 @@
...
+  scope: 'feature',
+  check(logger, { data, path }) {
+    processData(logger, data, browsers, path.full);
+  },
+  internals: {
+    neverImplemented,
+  },
```
queengooborg, line 109: Rather than export an `internals` property on our default export, I think that we should instead provide a named export of the function and import it that way.

```suggestion
```

test/linter/test-obsolete.js
```diff
@@ -0,0 +1,110 @@
...
+  'http.headers.Feature-Policy.layout-animations',
+  'http.headers.Feature-Policy.legacy-image-formats',
+  'http.headers.Feature-Policy.oversized-images',
+  'http.headers.Feature-Policy.unoptimized-images',
+  'http.headers.Feature-Policy.unsized-media',
+  'svg.attributes.presentation.color-rendering',
+  'svg.elements.view.zoomAndPan',
+];
+
+function neverImplemented(support) {
```
queengooborg, line 27:
```suggestion
export function neverImplemented(support) {
```

test/linter/test-obsolete.test.js
```diff
@@ -0,0 +1,49 @@
+/* This file is a part of @mdn/browser-compat-data
+ * See LICENSE file for more information. */
+
+import assert from 'node:assert/strict';
+
+import testObsolete from './test-obsolete.js';
+const { internals } = testObsolete;
+const { neverImplemented } = internals;
```
queengooborg, line 8:
```suggestion
import { neverImplemented } from './test-obsolete.js';
```

test/linter/test-obsolete.js
```diff
@@ -0,0 +1,110 @@
+/* This file is a part of @mdn/browser-compat-data
+ * See LICENSE file for more information. */
+
+import chalk from 'chalk-template';
+import bcd from '../../index.js';
+const { browsers } = bcd;
+
+const exceptions = [
```
queengooborg, line 8: Can you add a comment describing what `exceptions` means here?  I'm assuming it means "obsolete features that we want to keep", correct?

test/linter/test-obsolete.js
```diff
@@ -0,0 +1,110 @@
...
+  'http.headers.Cache-Control.stale-if-error',
+  'http.headers.Feature-Policy.layout-animations',
+  'http.headers.Feature-Policy.legacy-image-formats',
+  'http.headers.Feature-Policy.oversized-images',
+  'http.headers.Feature-Policy.unoptimized-images',
+  'http.headers.Feature-Policy.unsized-media',
+  'svg.attributes.presentation.color-rendering',
+  'svg.elements.view.zoomAndPan',
+];
+
```
queengooborg, line 26: I think we should define the `warningDate` and `errorDate` variables up top rather than over and over in the function calls.  I also think we should use `setFullYear()`, because it helps clarify what is getting bumped:

```suggestion

let errorTime = new Date();
errorTime.setFullYear(errorTime.getFullYear() - 2.5);
let warningTime = new Date();
warningTime.setFullYear(warningTime.getFullYear() - 2);

```

test/linter/test-obsolete.js
```diff
@@ -0,0 +1,110 @@
...
+/**
+ * @param {*} browsers
+ * @param {*} support
+ * @returns 'warning' | 'error' | false
+ */
+function implementedAndRemoved(browsers, support) {
+  let result = 'error';
+  const warningTime = new Date().getTime() - 2 * 365 * 24 * 60 * 60 * 1000;
+  const errorTime = new Date().getTime() - 2.5 * 365 * 24 * 60 * 60 * 1000;
```
queengooborg, line 44:
```suggestion
```

test/linter/test-obsolete.js
```diff
@@ -0,0 +1,110 @@
...
+  for (const browser in support) {
+    let data = support[browser];
+    if (!Array.isArray(data)) data = [data];
+    for (const d of data) {
+      // Feature is still supported
+      if (!d.version_removed) return false;
+      const releaseDate = new Date(
+        browsers[browser].releases[d.version_removed].release_date,
+      ).getTime();
```
queengooborg, line 53: The `.getTime()` call can be removed since we can compare dates using comparison operators directly:

```suggestion
      );
```

queengooborg (changes requested):
Thank you for your PR -- just got a couple of feedback points, but LGTM overall!

test/linter/test-obsolete.js
bershanskiy: I added the following comment:
```
/**
 * List of feature keys which are expected to fail this test.
 * This list is needed for two reasons:
 *  - some entries indeed should be removed (like api.VR*)
 *  - some entries are currently just stubs and should be expanded and will pass test later
 */
```

I plan to gradually fix these entries and remove exceptions. However, the `exceptions` variable itself is probably necessary since occasionally we have to add feature stubs for upcoming APIs.

bershanskiy:
@queengooborg Thanks for the review, I think I addressed all the comments!

queengooborg (approved):
Thank you for this!  I just ran some quick tests to ensure this is functioning as expected, and all is looking good to me.

For a follow-up, I think we could also enforce the third condition for irrelevant features, where it can be removed if "a feature is unsupported in all releases in the past five years" so we can track IE-only features better.  Something for the future though!

PR #16434, opened 2022-05-25
Update diff script

queengooborg:
This PR updates the diff script to provide a few quality of life updates, including performing mirroring within the diff (which will greatly help in validating upcoming changes to switch version numbers with `mirror`).

Depends on #16432.

### DIFF
--- a/scripts/diff.js
+++ b/scripts/diff.js
@@ -23,16 +23,8 @@ function getBaseAndHeadContents(baseCommit, basePath, headCommit, headPath) {
   return { base, head };
 }
 
-function stringifyValue(value) {
-  return Array.isArray(value)
-    ? 'typeof array'
-    : value && typeof value === 'object'
-    ? JSON.stringify(value)
-    : value;
-}
-
 function stringifyChange(lhs, rhs) {
-  return `${stringifyValue(lhs)} → ${stringifyValue(rhs)}`;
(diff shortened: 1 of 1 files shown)


foolip:
I've skimmed the code but can't do a review now, so asking in case another reviewer beats me to it:

If we change the mirroring rules, will the diff script use the old rules for the previous state, and the new rules for the current state? There would have to be an independent `npm install` and `npm run build` somewhere, but I don't see this.

queengooborg:
I thought about doing a setup that accounts for differences in the mirroring script, but given the number of potential issues with importing the mirroring script (architectural change, incompatible packages with current NodeJS, etc.), I wanted to dig deeper into it at a later time.  This gets us off the ground at least, so we have something to use today -- and then if the mirroring script does change in a way that alters the output, my idea is that unittests will catch this!

foolip:
OK, I guess we can diff build/data.json before/after the changes manually in those cases. Applying the mirroring in the diff is the most important thing, so that it's clear when replacing a version number with "mirror" actually changes the resulting version.

@saschanaz can you review this change?

queengooborg:
I'm going to self-merge this because it involves critical infrastructure changes and is blocking further progress on a change summary bot.

saschanaz:
Oops, somehow missed this. Looks good btw.

PR #16435, opened 2022-05-26
Test style for browsers as well

queengooborg:
This PR adds style testing to the browser JSON.

### DIFF
--- a/test/linter/test-style.js
+++ b/test/linter/test-style.js
@@ -63,9 +63,7 @@ function orderFeatures(key, value) {
 /**
  * Return a new feature object whose first-level properties have been
  * ordered according to compareVersions, and so will be
- * stringified in that order as well. This relies on guaranteed "own"
- * property ordering, which is insertion order for non-integer keys
- * (which is our case).
+ * stringified in that order as well.
  *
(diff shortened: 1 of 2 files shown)


test/linter/test-style.js
```diff
@@ -58,50 +60,99 @@ function orderFeatures(key, value) {
   return value;
 }
 
+/**
+ * Return a new feature object whose first-level properties have been
+ * ordered according to compareVersions, and so will be
+ * stringified in that order as well. This relies on guaranteed "own"
+ * property ordering, which is insertion order for non-integer keys
```
foolip, line 67: This doesn't seem right, we do have "integer" keys like "10" as well as keys like "10.1". These are iterated in the wrong order, and they will be after putting them in a new object as well, right?

Does this lint fail when such versions are out of order?

queengooborg: Description updated -- apparently there was also a typo in the script file so this wasn't updating properly.  What's odd though is that the ordering doesn't seem to be right, hmm...

queengooborg:
It looks like the JavaScript engine simply refuses to order the releases in the specified order, so I don't think this linter is possible anymore.

PR #16445, opened 2022-05-27
Test the default mirror.js export more to show many more bugs

foolip:
This is the code path that the build script uses, and there's a lot
wrong. Most of this should be fixed before we can use built-time
mirroring widely.


foolip:
We have a tricky choice to make now.

We could either replace versions with "mirror" now with all the bugs intact, when more of the data matches, and then fix the bugs to update all the mirrored data. That would probably fix a lot of incorrectly mirrored data, but it might also silently break data that was manually confirmed to be correct with the current version numbers, where correct mirroring would give the wrong result.

Or we fix the bugs in the mirroring script before we do mass replacement, and will be left with a lot of data that doesn't quite match, and after the fact it will be a lot of work to determine if the reason is previous incorrect mirroring, or something else.

Thoughts?

queengooborg:
Hmm...  First off, I think that some of the places where a comment stating mirroring is "wrong" is actually incorrect, specifically where stated that a range should be added vs. the hard number.  I realize this is inconsistent with how WebView is handled, but as I'm working to remove ranges for WebView, this will become less relevant quickly.

For the other bugs, I think they would be helpful to fix before we get deeper into build-time mirroring.

scripts/release/mirror.test.js
```diff
@@ -19,8 +20,10 @@ describe('mirror', () => {
         ['26', '26'],
       ],
       edge: [
-        ['50', '79'],
-        ['78', '79'],
+        ['1', false], // wrong, invalid inference
+        ['27', false], // wrong, invalid inference
+        ['28', '79'], // wrong, could be in Edge <79
+        ['78', '79'], // wrong, could be in Edge <79
```
queengooborg, line 26: Mirroring shouldn't generate ranges (which is inconsistent with WebView below; I'll be fixing that up).

```suggestion
        ['28', '79'],
        ['78', '79'],
```

scripts/release/mirror.test.js
```diff
@@ -36,14 +39,14 @@ describe('mirror', () => {
         ['80', '80'],
       ],
       opera: [
-        ['1', '15'],
+        ['1', '15'], // wrong, could be in Opera <15
```
queengooborg, line 42:
```suggestion
        ['1', '15'],
```

scripts/release/mirror.test.js
```diff
@@ -36,14 +39,14 @@ describe('mirror', () => {
...
         ['27', '15'],
         ['28', '15'],
         ['29', '16'],
         ['30', '17'],
       ],
       opera_android: [
-        ['18', '14'],
+        ['18', '14'], // wrong, could be in Opera Android <14
```
queengooborg, line 49:
```suggestion
        ['18', '14'],
```

queengooborg (changes requested):

scripts/release/mirror.test.js
foolip: My point wasn't that we should use ranges, but that for something implemented long before Chrome 79 it might have been in Edge 18 or earlier too.

But in those cases we shouldn't use "mirror" so I no longer think that's a problem.

foolip:
@queengooborg suggestions applied

queengooborg (approved):

PR #16446, opened 2022-05-27
Chrome 105 adds Sanitizer API

rachelandrew:
The Sanitizer API is in Chrome 102 so I've updated the information for all platforms for this.

https://chromestatus.com/feature/5786893650231296


queengooborg:
The ChromeStatus link says the feature was shipped in Chrome 105 instead, could you confirm which is correct?

queengooborg:
It looks like this is not available in Chrome 103, so I'm going to assume it will be added in Chrome 105.

queengooborg (approved):
I fixed merge conflicts and updated the version numbers accordingly!

PR #16449, opened 2022-05-27
Update length BCD table for new viewport-percentage units

dipikabh:
#### Summary

- In Firefox release 101, new viewport sizes have been introduced: small (s), large (l), and dynamic (d).
- As a result, the existing viewport-percentage length units (vw, vh, vmin, vmax) can have more variants.
Each can be of the type s, l, d resulting in:
dvb, dvh, dvi, dvmax, dvmin, dvw,
lvb, lvh, lvi, lvmax, lvmin, lvw,
svb, svh, svi, svmax, svmin, svw
- Also, vb and vi units are now also supported by default in FF and were already supported in Safari, so are longer experimental

https://bugzilla.mozilla.org/show_bug.cgi?id=1610815

#### Related issues

Doc issue tracking this work: https://github.com/mdn/content/issues/15465
Comment reference: https://github.com/mdn/content/issues/15465#issuecomment-1139175513

### DIFF
--- a/css/types/length.json
+++ b/css/types/length.json
@@ -590,6 +590,150 @@
             }
           }
         },
+        "viewport_percentage_units_dynamic": {
+          "__compat": {
+            "description": "<code>dvb</code>, <code>dvh</code>, <code>dvi</code>, <code>dvmax</code>, <code>dvmin</code>, <code>dvw</code> units",
+            "support": {
+              "chrome": {
+                "version_added": false
+              },
+              "chrome_android": {
+                "version_added": false
+              },
+              "edge": {
+                "version_added": false
+              },
+              "firefox": {
+                "version_added": "101"
+              },
+              "firefox_android": {
+                "version_added": "101"
+              },
+              "ie": {
+                "version_added": false
+              },
+              "opera": {
+                "version_added": false
+              },
+              "opera_android": {
+                "version_added": false
+              },
+              "safari": {
+                "version_added": "15.4"
+              },
+              "safari_ios": {
+                "version_added": "15.4"
+              },
+              "samsunginternet_android": {
+                "version_added": false
+              },
+              "webview_android": {
+                "version_added": false
+              }
+            },
+            "status": {
+              "experimental": false,
+              "standard_track": true,
+              "deprecated": false
+            }
+          }
+        },
+        "viewport_percentage_units_large": {
+          "__compat": {
+            "description": "<code>lvb</code>, <code>lvh</code>, <code>lvi</code>, <code>lvmax</code>, <code>lvmin</code>, <code>lvw</code> units",
+            "support": {
+              "chrome": {
+                "version_added": false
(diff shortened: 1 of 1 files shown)


css/types/length.json
```diff
@@ -557,11 +558,12 @@
...
               },
               "firefox": {
-                "version_added": false,
-                "notes": "See <a href='https://bugzil.la/1287034'>bug 1287034</a>."
+                "version_added": "101",
+                "notes": "Starting with version 21, viewport-percentage lengths are invalid in <code>@page</code>."
               },
               "firefox_android": {
-                "version_added": false
+                "version_added": "19",
```
hamishwillee, line 565: Version 19? Is that correct /where does it come from?

css/types/length.json
```diff
@@ -557,11 +558,12 @@
                 "version_added": false
               },
               "firefox": {
-                "version_added": false,
-                "notes": "See <a href='https://bugzil.la/1287034'>bug 1287034</a>."
+                "version_added": "101",
+                "notes": "Starting with version 21, viewport-percentage lengths are invalid in <code>@page</code>."
```
hamishwillee, line 562: What does this mean? i.e. specifically, this comes in at version 101 so why a note about version 21? (everywhere this appears)?

hamishwillee:
@queengooborg FYI CanIuse seems to think these new values implemented behind a flag on chrome too https://caniuse.com/viewport-unit-variants - also it gets mentioned on twitter: https://twitter.com/jensimmons/status/1499441043930062854?lang=en

But I can't find a chrome status on this. Do we need to verify with test code or we ignore for now?

dipikabh:
Thanks @hamishwillee for catching the errors (and the unnecessary note) - they were all clear oversight on my part.

dipikabh:
@queengooborg, aside from the changes for the new viewport-percentage units, I just updated the table for the older units `lh` and `rlh`. They don't seem to be supported on any browser.

css/types/length.json
```diff
@@ -438,8 +438,8 @@
               }
             },
             "status": {
-              "experimental": false,
-              "standard_track": true,
+              "experimental": true,
```
hamishwillee, line 441: @dipikabh FYI this is better. Note though that normally we don't record things in BCD unless there is some implementation, so (even though this predates you) @queengooborg may suggest you remove the whole entry.

hamishwillee:
Looks great to me, modulo maybe the need to add Chrome (behind flag) and perhaps deleting the features with no implementations.

@queengooborg When you have a chance ...

queengooborg (approved):

css/types/length.json
```diff
@@ -340,8 +340,8 @@
...
+              "standard_track": false,
```
foolip, line 344: The `lh` unit is defined in https://drafts.csswg.org/css-values/#lh

PR #16467, opened 2022-05-30
TaskSignal: prioritychange_event url fix

hamishwillee:
The prioritychange_event is fired on TaskSignal, but the URL assumed it was global. This fixes the URL for the docs to match where they will actually be.


hamishwillee:
@queengooborg When you have a moment!

queengooborg (approved):
👀

PR #16473, opened 2022-05-30
adds Element.isVisible

rachelandrew:
[Element.isVisible](https://drafts.csswg.org/cssom-view/#dom-element-isvisible) is in [Chrome 103](https://chromestatus.com/feature/5163102852087808). This PR adds BCD.


queengooborg (changes requested):
Thank you for adding this!  We recently added a feature that allows derivative browsers to mirror from their upstream counterparts.  Let's utilize it here, if that sounds good (this applies to all other PRs as well)!

foolip:
https://github.com/w3c/csswg-drafts/issues/7317 resolved to rename this to `isHidden`.

rachelandrew:
@foolip I'll update it when I do the other change

queengooborg:
It looks like this has been renamed to `isHidden` in the spec, and it has not been included in Chrome 103 with either name.  I'm going to mark this as a draft since no implementation has been made in any browser yet.

foolip:
It was actually renamed to `checkVisibility` in the end.

queengooborg:
Alright, since there's no implementation and the spec has been renaming this, I think we should wait until a browser has implemented this before creating a PR.

PR #16475, opened 2022-05-30
Don't alter data when performing upstream mirroring

### LINKED ISSUE #16471: mirrorSupport modifies original data only in some cases
https://github.com/mdn/browser-compat-data/pull/16432 introduced this bit of code:

https://github.com/mdn/browser-compat-data/blob/b419f64e0230f0461143bb1b911cb5cf00816376/scripts/release/mirror.js#L369-L372

`data` is the input, and is modified in case the upstream itself is mirrored, as in the case opera_android &larr; chrome_android &larr; chrome mirroring chain.

The API would be simpler if it either always modified the input, or never did.

queengooborg:
This PR fixes #16471 by creating a separate variable to hold the upstream data, which mirroring is then performed on if needed, rather than using and altering the initial data.  Additionally, this includes a check to ensure that upstream data is defined for the compatibility data.

### DIFF
--- a/scripts/release/mirror.js
+++ b/scripts/release/mirror.js
@@ -366,12 +366,20 @@ const mirrorSupport = (destination, data) => {
     );
   }
 
-  if (data[upstream] == 'mirror') {
-    // Perform mirroring upstream if needed
-    data[upstream] = mirrorSupport(upstream, data);
+  let upstreamData = data[upstream];
+
+  if (!upstreamData) {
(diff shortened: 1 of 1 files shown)


queengooborg:
I'm going to self-merge since this is critical infrastructure and the main reviewer for infrastructure PRs is now busy.

foolip (commented):
Fix looks good, thanks!

PR #16480, opened 2022-05-31
Drop exception for WOFF spec

sideshowbarker:
https://github.com/w3c/browser-specs/pull/626 was merged, so we no longer need this exception.


queengooborg:
Marking as "not ready" since this is waiting on a new version of `browser-specs`.

queengooborg (approved):

PR #16484, opened 2022-05-31
Add migration script for mirroring

foolip:


foolip:
This is able to replace ~75k support statements across 2000+ files, but I think the way to use this should be:
- First `node scripts/migrations/008-mirror.js --browsers chrome_android edge firefox_android opera opera_android safari_ios samsunginternet_android webview_android` for the cases that can be mirrored in all browsers, no complications to look into
- Then run things like `node scripts/migrations/008-mirror.js --browsers chrome_android webview_android` and `node scripts/migrations/008-mirror.js opera opera_android` and for smaller groups of things that will often be mirror-able
- Finally go through mirroring the individual browsers. This is the step where a lot of odd discrepancies might be discovered.

queengooborg:
Could we set this to migration 009?  I have a pull request for a migration 008 that's waiting on your review!

foolip:
I've renamed to 009. But also, do we need the numbers? We're missing some in the series, and will also delete scripts that have served their purpose.

scripts/migrations/009-mirror.js
```diff
@@ -0,0 +1,111 @@
...
+
+import bcd from '../../index.js';
+import { walk } from '../../utils/index.js';
+import mirrorSupport from '../release/mirror.js';
+
+const dirname = fileURLToPath(new URL('.', import.meta.url));
+
+export const isMirrorEquivalent = (support, browsers) => {
+  // Make a (shallow) copy of the support data to work around
+  // https://github.com/mdn/browser-compat-data/issues/16471.
```
foolip, line 22: This was fixed, so I should simplify this again.

foolip:
@queengooborg this is ready for review now.

queengooborg (approved):
I haven't tried running the script, but this looks great to me -- let's start migrating browsers to `mirror`!

PR #16485, opened 2022-05-31
Make labeled PRs the default for release-pulls script

queengooborg:
This PR switches the `--labeled` command line argument that can be passed into `release-scripts` with an `--all-prs` argument that performs the opposite.  Essentially, a publisher is more likely to run the script with the flag turned on, so this reverses the flag.


queengooborg:
Self-merging since this is critical infrastructure.

PR #4470, opened 2019-07-09
Fix WebKit versions in Safari

vinyldarkscratch:
Unfortunately, it looks like Safari's user agent string was lying to us regarding the Safari versions -- which I had a hunch about when originally adding browser engines.  After further review, I compared the [release tags](https://trac.webkit.org/browser/webkit/tags) with the [Safari releases](https://trac.webkit.org/browser/webkit/releases/Apple), and found that...well, good thing we caught it sooner than later, right?

By looking in the SVN repository for WebKit and comparing the tags (aka each WebKit release) to the releases (aka each Safari release) by their ChangeLog files, I have determined that the WebKit version outputted by the user agent string outputted by Safari has been incorrect for Safari 10+ (and possibly Safari 9, though I’m unable to confirm this as the repo doesn't have release tags for Safari 9.X):

Screen Shot 2019-07-09 at 22 02 03

In the About panel, I’ve found that the numbers next to the Safari version contain the minor version of macOS (AKA "10.XX"), followed by WebKit version:

Screen Shot 2019-07-09 at 21 58 26

The screenshot below shows that the ChangeLog file for Safari 12.1 exactly matches that for WebKit 607.2.6.1.1, even though the user agent claims it’s running WebKit 605.1.15:

Screen Shot 2019-07-09 at 03 52 45

Boo to bad UA strings... ☹️

PR #16498, opened 2022-06-01
Bump Safari iOS 15 engine version

queengooborg:
This PR bumps the engine version for Safari iOS 15 to match Safari Desktop 15.


foolip:
> a solution I know we have done before.

This made me curious, so I looked up which PRs have previously modified the engine versions and found https://github.com/mdn/browser-compat-data/pull/4470 + https://github.com/mdn/browser-compat-data/pull/4732 + https://github.com/mdn/browser-compat-data/pull/11146.

But none of these say anything about mirroring. Do you still remember what the change was, so we can perhaps revert it after we fix the mirroring in another way?

queengooborg:
Oh huh, I could've sworn one of those involved a tweak to fix up mirroring...  I must have been thinking of #11333, but that was removing excessive parts of the version number.

This is probably a good incentive to implement the `upstream_version` property I mentioned earlier, before adopting build-time mirroring more.

browsers/safari_ios.json
```diff
@@ -149,7 +149,7 @@
           "release_notes": "https://developer.apple.com/documentation/safari-release-notes/safari-15-release-notes",
           "status": "retired",
           "engine": "WebKit",
-          "engine_version": "612.1.27"
+          "engine_version": "612.1.29"
```
gsnedders, line 152: https://github.com/WebKit/WebKit/blob/releases/Apple/Safari-15-iOS-15.0/Source/WebKit/Configurations/Version.xcconfig claims this change is wrong.

scripts/release/mirror.test.js
```diff
@@ -76,7 +76,7 @@ describe('mirror', () => {
         ['13.1', '13.4'],
         ['14', '14'],
         ['14.1', '14.5'],
-        ['15', '15.1'],
+        ['15', '15'],
```
gsnedders, line 79: …but this change seems correct.

browsers/safari_ios.json
foolip: Thanks for confirming!

scripts/release/mirror.test.js
foolip: We'll have to fix it in some other way.

foolip: https://github.com/mdn/browser-compat-data/pull/16594

PR #16510, opened 2022-06-01
Fix version filtering in the traverse script

foolip:
The example `npm run traverse -- -b firefox -f 10` did not work because
10 is interpreted as a number. Use the string option to fix this.

Also remove unused default argument values and fix a typo.

### DIFF
--- a/scripts/traverse.js
+++ b/scripts/traverse.js
@@ -143,6 +143,7 @@ if (esMain(import.meta)) {
           alias: 'f',
           describe: 'Filter by version value. May repeat.',
           type: 'array',
+          string: true,
           nargs: 1,
           default: [],
         })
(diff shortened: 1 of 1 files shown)


scripts/traverse.js
```diff
@@ -89,21 +89,7 @@ function traverseFeatures(obj, browsers, values, depth, identifier) {
...
-    'http',
-    'svg',
-    'javascript',
-    'mathml',
-    'webdriver',
-  ],
-  browsers = Object.keys(bcd.browsers),
-  values = ['null', 'true'],
-  depth = 100,
-) => {
```
queengooborg, line 106: I don't think we should remove the default values here.  While they're mostly duplicated with `yargs`, it would be nice to provide defaults on the function call should we desire to run this function from another script.

foolip: I was bugged that the defaults don't exactly match, but I've now backed out this change. I might be bugged again in the future and fixed it in some other way.

queengooborg (approved):

PR #16514, opened 2022-06-02
Remove api.GlobalEventHandlers.onabort

queengooborg:
This PR demixes the `abort` global event and separates them into their own interfaces.

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -203,56 +203,6 @@
           }
         }
       },
-      "abort_event": {
-        "__compat": {
-          "description": "<code>abort</code> event",
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/Document/abort_event",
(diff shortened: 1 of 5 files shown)


foolip:
These are the event targets for "abort" events I can find in Chromium:
- `XMLHttpRequest`
- `XMLHttpRequestUpload`
- `IDBTransaction`
- `SourceBuffer`
- `AbortSignal`
- `FileReader`
- `HTMLMediaElement`
- `FileWriter`

These all have "abort_event" entries in BCD except `FileWriter`, which isn't in BCD.

`onabort` was on `GlobalEventHandlers` only because of `HTMLMediaElement`, so I don't think there's anything to add, we should just remove the entry, not add anything new.

foolip (approved):

PR #16517, opened 2022-06-02
Mirror Opera support for Object.fromEntries() and .hasOwn()

foolip:
There are lots of entries like this where data hasn't been mirrored, but
this fixes two that came up when looking at the data.

### DIFF
--- a/javascript/builtins/Object.json
+++ b/javascript/builtins/Object.json
@@ -662,9 +662,7 @@
               "opera": {
                 "version_added": "60"
               },
-              "opera_android": {
-                "version_added": false
-              },
+              "opera_android": "mirror",
               "safari": {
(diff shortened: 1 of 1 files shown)


foolip:
cc @voxpelli

voxpelli (approved):
👏

queengooborg (changes requested):
We can also mirror Opera Desktop's data, let's do that while we're here?

foolip:
@queengooborg I'll do that, and also bump for https://github.com/mdn/browser-compat-data/pull/16484 which would do it automatically if the data already matches. I'd prefer to not make many manual "mirror" changes.

queengooborg (approved):

PR #16518, opened 2022-06-02
Drop the sourcemaps.info exception from test/spec-urls.test.js

sideshowbarker:
https://github.com/w3c/browser-specs/issues/618 has been merged, so we no longer need to maintain this exception.


queengooborg:
Marking as "not ready" since this will need a browser-specs update, probably #16521.

queengooborg (approved):
And #16521 is now merged so this is ready!

PR #16522, opened 2022-06-02
Convert spec URLs test to linter unit

queengooborg:
This PR converts the spec URL test from a Mocha unittest to a linter unit, for better logging and to keep all file linting in the linter command.  (I don't know why it was made a Mocha unittest to begin with...)


test/linter/test-spec-urls.js
```diff
@@ -0,0 +1,135 @@
...
+
+  for (const specUrl of featureSpecURLs) {
+    if (!allowedSpecURLs.find((host) => specUrl.startsWith(host))) {
+      logger.error(
+        chalk`Invalid specification host(s) found: {bold ${specUrl}}. Try a more current specification URL and/or check if the specification URL is listed in https://github.com/w3c/browser-specs.`,
+      );
+    }
+  }
+
+  //   let prefixes = [];
```
foolip, line 85: This commented out code was added in this PR, can you remove it and do it as a draft PR or something? Big blocks of commented out code tend to live on and then they don't actually work if uncommented.

test/linter/test-spec-urls.js
```diff
@@ -0,0 +1,135 @@
...
+function processData(data, logger) {
+  if (!data.spec_url) {
+    return;
+  }
+
+  const featureSpecURLs = Array.isArray(data.spec_url)
+    ? data.spec_url
+    : [data.spec_url];
+
+  for (const specUrl of featureSpecURLs) {
```
foolip, line 77:
```suggestion
  for (const specURL of featureSpecURLs) {
```

test/linter/test-spec-urls.js
```diff
@@ -0,0 +1,135 @@
...
+    if (!allowedSpecURLs.find((host) => specUrl.startsWith(host))) {
```
foolip, line 78: I found `host` to be misleading on first read. The host of a URL like https://w3c.github.io/mathml/ is w3c.github.io, and that's not what we're checking, we're looking for a URL prefix which is more strict than that.

```suggestion
    if (!allowedSpecURLs.some((prefix) => specURL.startsWith(prefix))) {
```

test/linter/test-spec-urls.js
```diff
@@ -0,0 +1,135 @@
...
+      logger.error(
+        chalk`Invalid specification host(s) found: {bold ${specUrl}}. Try a more current specification URL and/or check if the specification URL is listed in https://github.com/w3c/browser-specs.`,
```
foolip, line 80:
```suggestion
        chalk`Invalid specification URL found: {bold ${specURL}}. Try a more current specification URL and/or check if the specification URL is listed in https://github.com/w3c/browser-specs.`,
```

foolip (changes requested):

foolip (approved):

PR #16527, opened 2022-06-03
Update LargestContentfulPaint spec URLs

sideshowbarker:
These should be using https://w3c.github.io/largest-contentful-paint URLs rather than https://www.w3.org/TR/largest-contentful-paint URLs.

### DIFF
--- a/api/LargestContentfulPaint.json
+++ b/api/LargestContentfulPaint.json
@@ -3,7 +3,7 @@
     "LargestContentfulPaint": {
       "__compat": {
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/LargestContentfulPaint",
-        "spec_url": "https://www.w3.org/TR/largest-contentful-paint//#sec-largest-contentful-paint-interface",
+        "spec_url": "https://w3c.github.io/largest-contentful-paint/#sec-largest-contentful-paint-interface",
         "support": {
           "chrome": {
             "version_added": "77"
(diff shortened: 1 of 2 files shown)


sideshowbarker:
CI is failing on this because the current Editors Draft URL not in browser-specs generated `index.json` file — and the reason it’s not is that I think browser-spec gets the ED URL from scraping https://www.w3.org/TR/largest-contentful-paint/ and that incorrectly still has https://wicg.github.io/largest-contentful-paint as the ED URL.

The ED moved from wicg.github.io to w3c.github.io a while back. And https://github.com/w3c/largest-contentful-paint/commit/4a0930ca74a98941237635424cd34c8a44825b2d updated the ED URL in the spec source a few days ago, but that’s not yet been pushed/re-published to https://www.w3.org/TR/largest-contentful-paint.

So pinged one of the spec’s editors to ask them to update it. After they update, I guess we need to wait for browser-specs to be rebuild (and re-released?).

foolip:
> So pinged one of the spec’s editors to ask them to update it. After they update, I guess we need to wait for browser-specs to be rebuild (and re-released?).

Yes, and for Dependabot to send a PR to update the dependency in BCD.

organicghemp:
> Estos deben usar [URL https://w3c.github.io/largest-contentful-paint](https://w3c.github.io/largest-contentful-paint) en lugar de [URL https://www.w3.org/TR/largest-contentful-paint](https://www.w3.org/TR/largest-contentful-paint).

queengooborg:
Perhaps we should add this as an exception to the spec URLs test, and then remove it when the spec URLs are updated?

PR #16529, opened 2022-06-03
Demix click events from GlobalEventHandlers

queengooborg:
This PR demixes the click events on the GlobalEventHandlers mixin.


foolip:
The target of these events will always be an `Element` of some sort, according to https://w3c.github.io/uievents/#topmost-event-target. I've also confirmed this with https://software.hixie.ch/utilities/js/live-dom-viewer/?saved=10374 in Chrome, Firefox and Safari &ndash; I can't get the event to fire at `Document` or `Window`.

So I think we should have these entries either on `Element` or on `HTMLElement` + `SVGElement` + `MathMLElement`, and think that it should probably be `Element`. I'll try to test if the event can be fired on an element not in a known namespace.

foolip:
I've confirmed with https://software.hixie.ch/utilities/js/live-dom-viewer/?saved=10375 in Chrome, Firefox and Safari that a plain `Element` instance can be the target of "click" elements, so I think we should put these events on `Element` only.

api/HTMLElement.json
```diff
@@ -597,6 +597,56 @@
           }
         }
       },
+      "auxclick_event": {
```
foolip, line 600: Can you move these entries to Element.json and remove the other additions in this PR?

foolip (changes requested):

foolip:
@queengooborg can you review the `Element` vs. `HTMLElement` + `SVGElement` question?

PR #16530, opened 2022-06-03
Demix blur/focus events from GlobalEventHandlers

queengooborg:
This PR demixes the `blur` and `focus` events from the GlobalEventHandlers mixin.  Note that Window is not modified in this PR because it already has the events.


api/Document.json
```diff
@@ -1001,6 +1001,56 @@
           }
         }
       },
+      "blur_event": {
```
foolip, line 1004: A `Document` can't be the target of a "blur" event, so I don't think we should add this, or indeed any entries. We already have the needed "blur_event" and "focus_event" entries on `Element` and `Window`, which I think is the right place for them. So just removing the GlobalEventHandlers entries in this PR should do it.

foolip (changes requested):

api/GlobalEventHandlers.json
```diff
@@ -597,55 +597,6 @@
           }
         }
       },
-      "onblur": {
-        "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/GlobalEventHandlers/onblur",
-          "spec_url": "https://html.spec.whatwg.org/multipage/webappapis.html#handler-onblur",
```
foolip, line 603: Can you add these spec URLs for the removed entries to the existing entries that will stay?

foolip (approved):

PR #3563, opened 2019-03-05
Add BCD for animation and transition events

wbamberg:
This adds BCD for:

* animation events (`animationcancel`, `animationend`, `animationiteration`, `animationstart`)
* transition events (`transitioncancel`, `transitionend`, `transitionrun`, `transitionstart`)

...to `Document`, `HTMLElement`, and `Window`.

It's part of https://github.com/mdn/sprints/issues/1056.

I've in general taken the data from the corresponding `on-` properties in [GlobalEventHandlers](https://github.com/mdn/browser-compat-data/blob/master/api/GlobalEventHandlers.json), except that I've removed the stuff about webkit prefixes, since that doesn't seem to be true for these events.

PR #16531, opened 2022-06-03
Demix animation events from GlobalEventHandlers

queengooborg:
This PR demixes the `animation` event handlers from the GlobalEventHandlers mixin.  There were some odd discrepancies between the events already on the interfaces.  Here's what I found:

The `animation` events were set to `43`, however the event handlers said they were added in `43` only on `Window` then `79` everywhere else.  After some further research, it appears that the events _were_ added in `43`, and then the event handlers were added later other than for Window.

Additionally, the `Window` event had prefixes added to them, which came from #3563.  However, this appears to have been a mistake; the same PR added the features for `Document` and `HTMLElement` which don't have the prefix data, and the description indicates intention to remove it.  I've removed it here.


foolip (changes requested):
Like with most bubbling events we have too many entries. I've read https://drafts.csswg.org/css-animations/#events and Chromium's implementation, and it looks like the event target is always an `Element` for these events. Can you move the 4 entries to `Element` and remove everything else?

api/Document.json
```diff
@@ -456,9 +456,16 @@
...
-              "version_added": "12"
-            },
+            "safari": [
+              {
+                "version_added": "13.1"
+              },
+              {
+                "version_added": "12",
+                "partial_implementation": true,
+                "notes": "The event handler is exposed but will not actually be called unless both the \"Web Animations\" and \"CSS Animations via Web Animations\" preferences are enabled."
```
foolip, line 466: For this wording, can you follow the example of `onmessageerror`? We don't need to mention the flags I think.

api/Element.json
```diff
@@ -730,6 +730,254 @@
           }
         }
       },
+      "animationcancel_event": {
+        "__compat": {
+          "description": "<code>animationcancel</code> event",
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/Element/animationcancel_event",
+          "spec_url": "https://drafts.csswg.org/css-animations/#eventdef-animationevent-animationcancel",
```
foolip, line 737: Can you link https://drafts.csswg.org/css-animations/#dom-document-onanimationcancel here and do similarly throughout?

foolip (changes requested):

PR #16532, opened 2022-06-04
Simplify and fix WebView Android (and Edge) mirroring

queengooborg:
This PR simplifies the WebView mirroring by utilizing the generic bump function rather than custom logic.  In doing so, I found the custom logic was also broken, setting `4.4` and `4.4.3` incorrectly.


scripts/release/mirror.js
```diff
@@ -29,9 +29,13 @@ export const getMatchingBrowserVersion = (targetBrowser, sourceVersion) => {
   for (const r of releaseKeys) {
     const release = browserData.releases[r];
     if (
-      ['opera', 'opera_android', 'samsunginternet_android'].includes(
-        targetBrowser,
-      ) &&
+      [
+        'edge',
```
queengooborg, line 33: This also provides a fix for Edge's mirroring if attempting to mirror a pre-Blink version of Chrome to Edge.  I didn't feel it was worth placing in a separate PR though.

foolip:
These changes make sense, but I think a plan is needed for finding and fixing all the data that was incorrectly mirrored using the old version of the script.

One way is to replace as much as possible with "mirror" before this change, but I'm sure some incorrectly mirrored data with tweaked notes or something would remain.

So perhaps one could write a script that lists all entries where the data is potentially incorrectly mirrored, and make a burn down list? Hopefully most of it could be fixed automatically, but not all.

foolip:
@queengooborg I had questions in https://github.com/mdn/browser-compat-data/pull/16532#issuecomment-1146671647 about this, can you take a look?

PR #16533, opened 2022-06-04
Demix cancel event from GlobalEventHandlers

queengooborg:
This PR demixes the `cancel` event from GlobalEventHandlers.


foolip:
Per https://html.spec.whatwg.org/multipage/indices.html#event-cancel this event can fire on `<dialog>` and `<input type=file>`. From Chromium source I see there's also a "cancel" event on `Animation`, which is already in BCD.

So a `HTMLInputElement` entry might be missing here.

foolip:
I've sent https://github.com/mdn/browser-compat-data/pull/16847 to sort this out.

PR #16541, opened 2022-06-05
Remove canplay event from GlobalEventHandlers

queengooborg:
This PR demixes the `canplay` events from the GlobalEventHandlers mixin.


api/Document.json
```diff
@@ -1066,6 +1066,130 @@
           }
         }
       },
+      "canplay_event": {
```
foolip, line 1069: There's already a "canplay_event" entry for `HTMLMediaElement` and that's the only one we should have, as the event doesn't bubble. Can you revert the other additions, and then compare the data being removed to that existing entry, noting any discrepancies?

foolip (changes requested):

foolip (approved):
The remaining "canplay_event" on `HTMLMediaElement` looks good.

PR #16544, opened 2022-06-05
Demix change event from GlobalEventHandlers

queengooborg:
This PR demixes the `change` event from the GlobalEventHandlers mixin.


foolip (changes requested):
Here are the place that fire this event in the HTML spec:
https://html.spec.whatwg.org/#checkbox-state-(type=checkbox)
https://html.spec.whatwg.org/#radio-button-state-(type=radio)
https://html.spec.whatwg.org/#file-upload-state-(type=file)
https://html.spec.whatwg.org/#common-input-element-events
https://html.spec.whatwg.org/#send-select-update-notifications
https://html.spec.whatwg.org/#focus-update-steps

All of those concern `<input>` or `<select>` elements.

https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/change_event also mentions `<textarea>`. I can't find where this happens in the spec but have filed https://github.com/whatwg/html/issues/8073.

I suggest adding "change_event" entries to `<input>`, `<select>` and `<textarea>`.

foolip:
Per https://github.com/mdn/browser-compat-data/pull/16651#issuecomment-1174820395 I think maybe it's better to go with a single entry on `HTMLElement` here.

foolip (approved):

api/HTMLElement.json
```diff
@@ -549,6 +549,46 @@
           }
         }
       },
+      "change_event": {
+        "__compat": {
+          "description": "<code>change</code> event",
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLElement/change_event",
+          "spec_url": "https://html.spec.whatwg.org/multipage/webappapis.html#handler-onchange",
```
foolip, line 556:
```suggestion
          "spec_url": [
            "https://html.spec.whatwg.org/multipage/indices.html#event-change",
            "https://html.spec.whatwg.org/multipage/webappapis.html#handler-onchange"
          ],
```

PR #16546, opened 2022-06-05
Remove contextmenu from GlobalEventHandlers

queengooborg:
This PR demixes the `contextmenu` event from the GlobalEventHandlers mixin.


foolip (approved):
There's already a "contextmenu_event" entry for `Element` and that's all we need, so I reverted the other additions. Not merging in case content updates are needed.

PR #16557, opened 2022-06-05
Remove never-implemented WebVR interfaces

bershanskiy:
#### Summary

Remove never-implemented WebVR interfaces.

Apparently, these interfaces existed only within WebVR spec drafts and were never actually implemented.
 - [These interfaces within Firefox source code](https://github.com/mozilla/gecko-dev/blob/master/dom/webidl/VRDisplay.webidl) (please also see "History" of the file, including history prior to renaming)
 - [WebVR spec](https://immersive-web.github.io/webvr/spec/1.1/) (please see IDL and spec history too)

#### Related issues

This is a follow-up to #16429.


bershanskiy:
FYI: related content update had landed, so this PR should not break anything.

bershanskiy:
Should I do anything to advance this PR?

queengooborg (approved):
Thanks, this is LGTM!

PR #16558, opened 2022-06-05
Remove durationchange event from GlobalEventHandlers

queengooborg:
This PR demixes the `durationchange` event from the GlobalEventHandlers mixin.


foolip (changes requested):
This is a media element event, I'd like to ask for the same changes/checks as in https://github.com/mdn/browser-compat-data/pull/16541#pullrequestreview-1002830332.

foolip (approved):
I've compared to the "durationchange_event" entry on `HTMLMediaElement` and that data is earlier than this.

PR #16559, opened 2022-06-05
Demix emptied event from GlobalEventHandlers

queengooborg:
This PR demixes the `emptied` event from the GlobalEventHandlers mixin.


foolip (changes requested):
This is an `HTMLMediaElement` event, can you revert the additions?

foolip (approved):

PR #16571, opened 2022-06-07
Generate TypeScript definitions from schema

### LINKED ISSUE #16494: Generate TypeScript defs from JSON schema
I found https://www.npmjs.com/package/json-schema-to-typescript which takes a JSON schema and converts it to TypeScript type definitions.  I think it would be very helpful for us to utilize this to maintain consistency between our internal schema and external typedefs.

queengooborg:
This PR fixes #16494 by generating TypeScript definitions from the JSON schema files.  The typedef is generated on initial installation, by command, and during the build process.


queengooborg:
Reviewer note: while this pull request does modify the schema, I have been very careful to ensure that updates are within reason and would not impact consumers in any harmful way.  While a top-level `__compat` statement _was_ added as an option, this was to open the door to adding top-level support for a category should we so choose, as well as simplify the schema and TypeScript outputs.

PR #16579, opened 2022-06-07
Linter: add exceptions attribute

### LINKED ISSUE #16567: Linter: add `exceptions` parameter
Mocha has a test hook called [`after()`](https://mochajs.org/#hooks) which runs after all tests in a given block were run. This kind of hook would be useful in the BCD linter. I propose to extend linting rules with one more (optional) hook callback called `after`.

The overall linter entry would look like this:
```
export default {
  name: 'Some linting',
  description: 'Test for some linting',
  scope: 'feature',
  check(logger, { data, path }) {
    // whtewer
  },
  after(logger) { // This is the new hook!
    // Cleanup/summary code goes here
  }
};
```

Use case: The test for obsolete entries has a list of [`exceptions`](https://github.com/mdn/browser-compat-data/blob/main/test/linter/test-obsolete.js#L14) that are expected to fail. The challenge is, that there is no easy way to ensure that once a feature is removed, its exception is also removed. This does not cause any false negatives (since the feature is completely gone in this case), but it would be nice to be able to clean up the linter when needed.

bershanskiy:
#### Summary

Linter: add `exceptions` parameter.

This is a PR to improve tests (linting). I tested it by adding an extra exception and verifying that it was found.

#### Related issues

Fixes #16567.


queengooborg (approved):
LGTM, thank you!

PR #16586, opened 2022-06-07
Updates for Safari 16 beta

gsnedders:
The bulk of this just Safari 16 beta data derived from the mdn-bcd-collector collected data (on both macOS 12 and macOS 13 beta), but in addition:

 * This adds version_removed for SharedWorkerGlobalScope corresponding to when SharedWorker was previously removed from Safari so that we can then document its addition in 16 beta (compare with the version numbers in SharedWorker.json).
 * Adds data for KeyframeEffect, as otherwise we would introduce lint failures from mdn-bcd-collector due to the parent being false. The 13.1/13.4 versions here come straight from mdn-bcd-collector; no further testing has been done there, but given this is when Web Animations support was added this seems highly likely to be correct.

Note this will currently fail the lint:

```
Problems in 2 files:
✖ browsers/safari.json
  → Browser Data – 1 error:
    → safari has multiple beta releases (16, 15.5), which is not allowed.
✖ browsers/safari_ios.json
  → Browser Data – 1 error:
    → safari_ios has multiple beta releases (16, 15.5), which is not allowed.
```

It's unclear to me what the correct solution here is; both 15.5 and 16 are currently beta releases, so the lint is enforcing a property that isn't true.


lukewarlow:
The push and notification changes that are only available on macOS 13 later are currently marked as partial. I don't believe this is correct if that's purely on the basis of OS support.

There's existing entries where IE11 only supports stuff from Windows 8 (see Media Source) and they're not marked as partial.

css/properties/overflow.json
```diff
@@ -93,17 +93,22 @@
...
-                    "name": "CSS overflow: clip support"
-                  }
-                ]
-              },
+              "safari": [
+                {
+                  "version_added": "16"
+                },
+                {
+                  "version_added": "preview",
```
stof, line 101: shouldn't this be replaced entirely instead ? I doubt Safari TP still requires a preference if Safari 16 shipped it.

gsnedders:
> The push and notification changes that are only available on macOS 13 later are currently marked as partial. I don't believe this is correct if that's purely on the basis of OS support.
> 
> There's existing entries where IE11 only supports stuff from Windows 8 (see Media Source) and they're not marked as partial.

https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#operating-system-limitations-imply-partial_implementation states this should be the case.

css/properties/overflow.json
gsnedders: Uh, yes. Fixed! Thanks!

queengooborg (approved):
Thanks for the updates!  I quickly skimmed through the changes to get a general idea of what's getting updated, and this is all looking good to me!

Issue #12925, opened 2021-10-18
api.HTMLElement.focus - `options.preventScroll` supported in Safari 15


fritz-c:
#### What information was incorrect, unhelpful, or incomplete?
The `options.preventScroll` argument is marked as not supported in Safari 15 (desktop and mobile).
#### What did you expect to see?
The `options.preventScroll` argument is now supported in Safari 15 (desktop and mobile). The [bug tracker for adding the option has marked it as resolved as well](https://bugs.webkit.org/show_bug.cgi?id=178583).
#### Did you test this? If so, how?
Yes. I made [a codesandbox to test it](https://i1xni.csb.app/) ([source](https://codesandbox.io/s/htmlelement-focus-preventscroll-i1xni)).

https://user-images.githubusercontent.com/4413963/137767227-c19b0cc8-0180-4fd7-a732-a02a3d3f49b6.MP4

https://user-images.githubusercontent.com/4413963/137767575-8c97f61a-5491-4202-9115-cd4444bdab6c.mov

<details>
<summary>MDN page report details</summary>

* Query: `api.HTMLElement.focus`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus
* Report started: 2021-10-18T15:24:53.483Z

</details>

woody-li:
https://developer.apple.com/safari/technology-preview/release-notes/#r124
Supported in TP 124, and Safari 15 is mapping with TP 126.

queengooborg:
This has been fixed in #16589!

PR #16589, opened 2022-06-07
Update Safari/iOS support for preventScroll option

makotokato:
#### Summary
Safari/iOS supports `https://webkit.org/blog/12669/new-webkit-features-in-safari-15-5/` from 15.5.

- https://webkit.org/blog/12669/new-webkit-features-in-safari-15-5/
- https://bugs.webkit.org/show_bug.cgi?id=236584

I verified this on iOS 15.5 like https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus.

### DIFF
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -1335,8 +1335,7 @@
                 "notes": "See <a href='https://webkit.org/b/178583'>bug 178583</a>."
               },
               "safari_ios": {
-                "version_added": false,
-                "notes": "See <a href='https://webkit.org/b/178583'>bug 178583</a>."
+                "version_added": "15.5"
               },
               "samsunginternet_android": {
                 "version_added": "9.0"


makotokato:
> Thanks for the PR! I just ran a quick test in Safari for macOS 15.5, and it looks like support was added here too -- could you update the data to reflect this?

I think that Safari 15 or 15.1 supports `preventScroll` option (https://github.com/mdn/browser-compat-data/issues/12925), but I cannot check which version supports it from since browserstack doesn't have 15 or 15.1 environment.

makotokato:
Add macOS version per https://github.com/mdn/browser-compat-data/issues/12925#issuecomment-1068709527

queengooborg (approved):
Thank you for the updates!  I'm down to set macOS to "15" and then verify if it's 15 or 15.1 at a later time.  👍

PR #16596, opened 2022-06-08
Mark break-before/break-after: always non-standard/deprecated

foolip:
The spec does not have this property value:
https://drafts.csswg.org/css-break/#break-between
https://drafts.csswg.org/css-break/#break-within

Rather, it's a property value of old the page-break-* properties:
https://drafts.csswg.org/css-break/#page-break-properties

### DIFF
--- a/css/properties/break-after.json
+++ b/css/properties/break-after.json
@@ -118,8 +118,8 @@
               },
               "status": {
                 "experimental": false,
-                "standard_track": true,
-                "deprecated": false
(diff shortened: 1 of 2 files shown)


foolip:
Note that there are no entries for `page-break-before: always` and `page-break-after: always`. I would guess those values were always supported, and it doesn't need to be more fine grained than just the property.

queengooborg (approved):

PR #16599, opened 2022-06-08
Mirror Edge support of RTCRtpTransceiver.currentDirection

foolip:
Confirmed supported in Edge 80 with this test:
http://mdn-bcd-collector.appspot.com/tests/api/RTCRtpTransceiver/currentDirection

### DIFF
--- a/api/RTCRtpTransceiver.json
+++ b/api/RTCRtpTransceiver.json
@@ -60,7 +60,7 @@
               "version_added": "69"
             },
             "edge": {
-              "version_added": false
+              "version_added": "79"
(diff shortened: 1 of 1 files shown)


queengooborg:
Perhaps we could use "mirror" here?

foolip:
Strike that, can't make suggestions in that part of the diff.

queengooborg (approved):
We can merge this as-is now; it's not an incorrect change (confirmed by testing Edge 79 in LambdaTest).  We can follow up with using `mirror` later!

PR #16609, opened 2022-06-09
Demix ended event from GlobalEventHandlers

queengooborg:
This PR demixes the `ended` event from the GlobalEventHandlers mixin.


foolip (changes requested):
There are a few different "ended" events, but it looks like we have all the ones we need already, in `AudioScheduledSourceNode`, `HTMLMediaElement` and `MediaStreamTrack`.

Can you revert the additions?

foolip (approved):

PR #16610, opened 2022-06-09
Demix error event from GlobalEventHandlers

queengooborg:
This PR demixes the `error` event from the GlobalEventHandlers mixin.

### DIFF
--- a/api/GlobalEventHandlers.json
+++ b/api/GlobalEventHandlers.json
@@ -39,48 +39,6 @@
           "deprecated": false
         }
       },
-      "onerror": {
-        "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/GlobalEventHandlers/onerror",
-          "spec_url": "https://html.spec.whatwg.org/multipage/webappapis.html#handler-onerror",
-          "support": {
-            "chrome": {
-              "version_added": "10"
-            },
-            "chrome_android": "mirror",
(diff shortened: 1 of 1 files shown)


foolip:
I've reverted the additions, which aren't specific enough.

foolip (changes requested):
This (the string "error") is probably the most widely used event type, fired in lots of places. For this cleanup we only need to worry about the `Node` hierarchy and `Window`, however.

There's already an entry for `Window`, and these are the elements I can find where there's an "error" event:
- `HTMLImageElement`
- `HTMLLinkElement`
- `HTMLMediaElement` (already in BCD)
- `HTMLObjectElement`
- `HTMLScriptElement`
- `HTMLSourceElement`
- `HTMLStyleElement`
- `HTMLTrackElement`
- `SVGImageElement`
- `SVGScriptElement`
- `SVGStyleElement`
- `SVGUseElement`

Can you add all of these, and remove the "error_event" entry on `Element`? There is no generic "error" event, all of these are separate code paths.

foolip:
Looks like there's also `HTMLInputElement` when `type=image`.

queengooborg:
I've isolated this PR to only focus on removing the event handler from GEH, and have opened an issue for adding the event to the other interfaces.  This should help us remove GEH while we get to the bottom of each interface!

foolip (approved):

PR #16611, opened 2022-06-09
Remove formdata event from GlobalEventHandlers

queengooborg:
This PR demixes the `formdata` event from the GlobalEventHandlers mixin.


foolip (commented):
https://developer.mozilla.org/docs/Web/API/HTMLFormElement/formdata_event already exists, so I'll revert the additions.

PR #16614, opened 2022-06-09
Linter: fix Consistency test, simplify Obsolete test

bershanskiy:
#### Summary

This PR fixes some issues uncovered during the development of #16593. ~~CI will fail because this PR fixes a bug in Consistency linter and uncovers some BCD issues, which I will fix in a separate data-only PR.~~ Please merge #16615 first.

#### Test results and supporting details

N/A

#### Related issues

This PR is blocked on #16615.
This PR blocks #16593.

New lint failures:
```
Consistency - 3 problems (3 errors, 0 warnings):
 ✖ api.MediaMetadata - Error → Basic support in firefox was declared implemented in a later version (82) than the following sub-feature(s):
        → api.MediaMetadata.MediaMetadata: 71
        → api.MediaMetadata.album: 71
        → api.MediaMetadata.artist: 71
        → api.MediaMetadata.artwork: 71
        → api.MediaMetadata.title: 71
 ✖ api.Worker - Error → Basic support in nodejs was declared implemented in a later version (11.7.0) than the following sub-feature(s):
        → api.Worker.Worker: 10.5.0
        → api.Worker.message_event: 10.5.0
        → api.Worker.postMessage: 10.5.0
        → api.Worker.terminate: 10.5.0
 ✖ javascript.statements.export - Error → Basic support in nodejs was declared implemented in a later version (13.2.0) than the following sub-feature(s):
        → javascript.statements.export.namespace: 12.0.0
```

PR #16593, opened 2022-06-08
Convert BCD to TypeScript

queengooborg:
This PR fixes #16333 by converting all of BCD's scripts over to TypeScript.

PR #16615, opened 2022-06-09
Fix: cleanup Consistency errors

bershanskiy:
#### Summary

This fixes consistency errors found by #16614.

This is simple mirroring based on parent feature support.

#### Related issues

This blocks #16614 which in turn blocks #16593.

### DIFF
--- a/api/Worker.json
+++ b/api/Worker.json
@@ -113,17 +113,29 @@
             "ie": {
               "version_added": "10"
             },
-            "nodejs": {
-              "version_added": "10.5.0",
-              "partial_implementation": true,
-              "notes": "Takes an entirely different options object.",
-              "flags": [
-                {
-                  "type": "runtime_flag",
-                  "name": "--experimental-worker"
-                }
-              ]
-            },
+            "nodejs": [
+              {
+                "version_added": "12.17.0",
+                "partial_implementation": true,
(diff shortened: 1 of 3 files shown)


api/Worker.json
```diff
@@ -116,7 +116,13 @@
...
               "partial_implementation": true,
-              "notes": "Takes an entirely different options object."
+              "notes": "Takes an entirely different options object.",
+              "flags": [
+                {
+                  "type": "runtime_flag",
+                  "name": "--experimental-worker"
+                }
+              ]
             },
```
queengooborg, line 126: If we're carrying the flag data down, we should separate these into multiple statements, like so:

```json
"nodejs": [
  {
    "version_added": "11.7.0",
    "notes": "Takes an entirely different options object."
  },
  {
    "version_added": "10.5.0",
    "notes": "Takes an entirely different options object.",
    "flags": [
      {
        "type": "runtime_flag",
        "name": "--experimental-worker"
      }
    ]
  }
]
```

(This applies to all of the NodeJS changes in this file.)

queengooborg (changes requested):
Thanks for catching these!  I just have a few comments, but LGTM otherwise!

queengooborg (approved):

PR #16617, opened 2022-06-09
Fix Opera "fromEntries" compatibility

Burtchen:
#### Summary
Compatibility for fromEntries in Opera seems to be missing.

This was tested using [this Pen](https://codepen.io/barzik/pen/OqdVqW). Most recent opera version to fail via Browserstrack was 58, earliest version to pass was 60.

Screenshot 2022-06-09 at 13 43 31
Screenshot 2022-06-09 at 13 42 26

### DIFF
--- a/javascript/builtins/Object.json
+++ b/javascript/builtins/Object.json
@@ -659,7 +659,9 @@
               "nodejs": {
                 "version_added": "12.0.0"
               },
-              "opera": "mirror",
+              "opera": {
+                "version_added": "60"
+              },
               "opera_android": "mirror",
               "safari": {
                 "version_added": "12.1"


queengooborg:
Hi there, thanks for your PR!  It looks like you're replacing the mirroring statement with a version number, however based on Chrome's data, the mirroring result should be "60" anyways.  Can you post a link to the site where it looks like the compatibility data is missing?

Burtchen:
> Hi there, thanks for your PR! It looks like you're replacing the mirroring statement with a version number, however based on Chrome's data, the mirroring result should be "60" anyways. Can you post a link to the site where it looks like the compatibility data is missing?

Hey, thanks for your quick answer. I was also very surprised by the omission - this is from [caniuse](https://caniuse.com/mdn-javascript_builtins_object_fromentries), which AFAIK uses mdn data under the hood:
Screenshot 2022-06-09 at 17 01 07

queengooborg:
Thanks for the link!  This is interesting -- apparently all of our mirror statements are showing up as "no support" on CanIUse, even though mirroring is confirmed to be working (https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Object/fromEntries).  I wonder if CanIUse is getting data from the repository directly instead of the released packages...

@Fyrd Could you confirm if my theory is correct?

queengooborg:
I'm going to go ahead and close this PR since the problem is with CanIUse rather than BCD.  Thank you for attempting to fix it though!

Fyrd:
@queengooborg Sorry, forgot to respond to this! Yes, can confirm caniuse is currently getting the data directly from the repo. Sounds like it should be the package instead?

queengooborg:
Thanks for the response, @Fyrd!  Yes, data should be obtained from the built packages (or at least from `build/data.json` after `npm run build`), as the built packages perform a few transformations (ex. merging the files in `_mixin` and build-time mirroring of support).

PR #16628, opened 2022-06-10
Add entries for break-before/break-after: avoid (in multicol)

foolip:
This is supported with LayoutNGBlockFragmentation, enabled in M102:
https://storage.googleapis.com/chromium-find-releases-static/1bb.html#1bb5f043274b0c27199c7ddb25e1f906167b5ddd

No entries are added for pages media, since that isn't supported yet.

### DIFF
--- a/css/properties/break-after.json
+++ b/css/properties/break-after.json
@@ -123,6 +123,40 @@
               }
             }
           },
+          "avoid": {
+            "__compat": {
+              "description": "<code>avoid</code>",
+              "support": {
+                "chrome": {
+                  "version_added": "102"
+                },
+                "chrome_android": "mirror",
+                "edge": "mirror",
+                "firefox": {
+                  "version_added": false
+                },
+                "firefox_android": "mirror",
+                "ie": {
+                  "version_added": false
+                },
+                "opera": "mirror",
+                "opera_android": "mirror",
+                "safari": {
+                  "version_added": false
+                },
+                "safari_ios": {
+                  "version_added": false
+                },
+                "samsunginternet_android": "mirror",
+                "webview_android": "mirror"
+              },
+              "status": {
+                "experimental": false,
+                "standard_track": true,
+                "deprecated": false
+              }
+            }
+          },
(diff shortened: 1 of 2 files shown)


foolip:
@mstensho this is me trying to add browser compat data for `break-before: avoid`, based on my limited understanding. Do you know if it's supported in any other browser? And I've noticed in testing that `break-before: avoid` is recognized since before M102, was it recognized but had no effect before? A simple test case should allow me to test and pinpoint exact versions if you don't have them easily available.

mstensho:
What's the "mirror" value?

break-before/break-after:avoid was supported in Opera until they dropped Presto, and I see that some entries say e.g. 11.1 – 12.1 for Opera. Not sure if it makes sense to mention this anymore.

Also: Chromium M102 also supports the 'avoid-column' value.

The 'avoid' (and 'avoid-column') value has "always" been recognized, but treated as 'auto' until M102.

Test (if you still need it):
```html
<p>Test passes if there is a filled green square and <strong>no red</strong>.</p>
<div style="width:100px; height:100px; background:red;">
  <div style="columns:2; column-gap:0; column-fill:auto; height:150px;">
    <div style="contain:size; height:100px; background:green;"></div>
    <div style="contain:size; height:50px; background:green;"></div>
    <div style="break-before:avoid-column; contain:size; height:50px; background:green;"></div>
  </div>
</div>
```

queengooborg:
The "mirror" value means "generate support statements from the upstream browser" to ensure that browsers with the same engines are kept in sync.  In other words, since Chrome's version number is set to 102, Chrome Android's will also be set to 102 since it's set to mirror!

foolip:
I'll need to follow up to incorporate the comments in https://github.com/mdn/browser-compat-data/pull/16628#issuecomment-1153568563

foolip:
I've sent https://github.com/mdn/browser-compat-data/pull/16866 and https://github.com/mdn/browser-compat-data/pull/16867 to further update the data. (This is more complicated than most features...)

PR #16629, opened 2022-06-11
Remove unused util visit()

bershanskiy:
#### Summary

Remove unused util `visit()`. I discovered this util while working on #16593 (making TypeScript declarations stricter).

This util is not used anywhere.

#### Related issues

This util was introduced in #9441, in March 2021. Since then over a year elapsed so probably it is safe to assume there are no concrete plans to use it.

CC @ddbeck


queengooborg:
Removal seems good to me, although it looks like just the test file was removed, not the file containing the utility function?

bershanskiy:
> it looks like just the test file was removed, not the file containing the utility function?

You are right! I fixed this.

Also, this PR will cause another merge conflict with #16593... Sorry about that... 😭

queengooborg (approved):

PR #16650, opened 2022-06-13
Demix got/lostpointercapture events from GlobalEventHandlers

queengooborg:
This PR demixes the `gotpointercapture` and `lostpointercapture` events from the GlobalEventHandlers mixin.


foolip (approved):

foolip:
I've filed https://github.com/mdn/content/issues/17945 about what the content changes will need to be. And that reminds me I also think we should do some additional changes here in BCD. I'll review again.

api/Document.json
```diff
@@ -5185,7 +5146,10 @@
         "__compat": {
           "description": "<code>lostpointercapture</code> event",
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Document/lostpointercapture_event",
-          "spec_url": "https://w3c.github.io/pointerevents/#the-lostpointercapture-event",
+          "spec_url": [
+            "https://w3c.github.io/pointerevents/#dom-globaleventhandlers-onlostpointercapture",
```
foolip, line 5150: Can you switch the order, most relevant link first?

api/Element.json
```diff
@@ -3459,6 +3501,48 @@
           }
         }
       },
+      "lostpointercapture_event": {
+        "__compat": {
+          "description": "<code>lostpointercapture</code> event",
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLElement/lostpointercapture_event",
+          "spec_url": [
+            "https://w3c.github.io/pointerevents/#dom-globaleventhandlers-onlostpointercapture",
```
foolip, line 3509: Switch order.

api/Window.json
```diff
@@ -2746,6 +2746,48 @@
           }
         }
       },
+      "lostpointercapture_event": {
```
foolip, line 2749: Are you sure the event can target a Window as opposed to bubbling up from a Document?

foolip (changes requested):

foolip (approved):

PR #16651, opened 2022-06-13
Demix input event from GlobalEventHandlers

queengooborg:
This PR demixes the `input` event from the `GlobalEventHandlers` mixin.


foolip:
This is a lot like https://github.com/mdn/browser-compat-data/pull/16544, an event that fires on  `<input>`, `<select>` and `<textarea>`, but additionally it can fire on elements using `contenteditable`.

I don't know how we'd represent this perfectly, a single entry on `HTMLElement` is probably the best. I'll suggest doing that in https://github.com/mdn/browser-compat-data/pull/16544 too.

foolip (approved):

PR #16658, opened 2022-06-14
Demix key events from GlobalEventHandlers

queengooborg:
This PR demixes the `key` events from the GlobalEventHandlers mixin.


foolip (changes requested):
Can you keep the entries on `Element` and remove all other entries, including the existing ones on `Document`?

PR #16659, opened 2022-06-14
Demix load event from GlobalEventHandlers

queengooborg:
This PR demixes the `load` event from the GlobalEventHandlers mixin.

### DIFF
--- a/api/GlobalEventHandlers.json
+++ b/api/GlobalEventHandlers.json
@@ -80,48 +80,6 @@
             "deprecated": false
           }
         }
-      },
-      "onload": {
-        "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/GlobalEventHandlers/onload",
-          "spec_url": "https://html.spec.whatwg.org/multipage/webappapis.html#handler-onload",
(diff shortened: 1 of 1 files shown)


api/Document.json
```diff
@@ -5167,6 +5167,253 @@
...
+            "webview_android": "mirror"
+          },
+          "status": {
+            "experimental": false,
+            "standard_track": true,
+            "deprecated": false
+          }
+        }
+      },
+      "loadeddata_event": {
```
foolip, line 5216: Can you break out the loaded* events so that the various "load" events are in one PR, and the rest (media element events) are in another?

foolip (changes requested):
This is going to be similar to https://github.com/mdn/browser-compat-data/pull/16610, lots of events in different places.

This will take time to get right and review, so maybe it makes more sense to handle one interface at a time, doing both "error" and "load" event together?

foolip:
@queengooborg in case you missed it: https://github.com/mdn/browser-compat-data/pull/16659#pullrequestreview-1028272694

foolip (approved):

PR #16660, opened 2022-06-14
Demix lostpointercapture event from GlobalEventHandlers

queengooborg:
This PR demixes the `lostpointercapture` event from the GlobalEventHandlers mixin.


foolip:
I suggest folding this into https://github.com/mdn/browser-compat-data/pull/16650, to keep it all in one place.

foolip (changes requested):
Requesting changes which means folding into https://github.com/mdn/browser-compat-data/pull/16650.

PR #16667, opened 2022-06-14
TSConfig: noFallthroughCasesInSwitch, allowUnreachableCode

bershanskiy:
#### Summary

Enable [`noFallthroughCasesInSwitch`](https://www.typescriptlang.org/tsconfig#noFallthroughCasesInSwitch). I find it easier to just use this rule than insert a `break;` or a comment `// fall through` in every `case`. Also, this particular case of fall-through just happens to be easier to read without a fall-through. Also, note that it's possible to disable this rule on a line-per-line basis via a special comment.

Set [`allowUnreachableCode`](https://www.typescriptlang.org/tsconfig#allowUnreachableCode) to `false` to make TS compiler error out on dead code.

N/A

#### Related issues

N/A

### DIFF
--- a/test/utils.ts
+++ b/test/utils.ts
@@ -90,16 +90,13 @@ export function indexToPosRaw(
     return [null, null];
   }
 
-  let prevChar = '';
   for (let i = 0; i < index; i++) {
     const char = str[i];
     switch (char) {
-      case '\n':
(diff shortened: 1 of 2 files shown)


queengooborg (approved):
Seems good to me, thanks!

PR #16671, opened 2022-06-14
Demix mouse events from GlobalEventHandlers

queengooborg:
This PR demixes the `mouse` events from the GlobalEventHandlers mixin.


foolip (changes requested):
Can you keep / update the entries on `Element` and remove everything else?

foolip (approved):

PR #16673, opened 2022-06-14
Demix play/playing/pause events from GlobalEventHandlers

queengooborg:
This PR demixes the `play`, `playing` and `pause` events from the GlobalEventHandlers mixin.


foolip (changes requested):
These are `HTMLMediaElement` events, can you revert the additions?

foolip (approved):

PR #16674, opened 2022-06-14
Demix pointer events from GlobalEventHandlers

queengooborg:
This PR demixes the `pointer` events from the GlobalEventHandlers mixin.


foolip (changes requested):
Can you move the `HTMLElement` entries to `Element` and remove everything else?

api/Element.json
```diff
@@ -4115,6 +4115,475 @@
           }
         }
       },
+      "pointercancel_event": {
+        "__compat": {
+          "description": "<code>pointercancel</code> event",
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/Element/pointercancel_event",
+          "spec_url": [
+            "https://w3c.github.io/pointerevents/#dom-globaleventhandlers-onpointercancel",
```
foolip, line 4123: Can you reverse the order of these? The first one is the more interesting reference. (I've put the handler definitions last in other PRs I've written/edited.)

foolip (changes requested):

api/Element.json
foolip: Yes, exactly!

foolip (approved):

PR #16678, opened 2022-06-15
Refactor: normalize source_file property

bershanskiy:
#### Summary

On Windows, paths have a separator `\` instead of `/` used on the web and literally everywhere else. This can be confusing and results in differences between published packages built on Windows. For reference, the latest BCD release, 5.0.1, uses `/`.

Please see docs on [`path.sep`](https://nodejs.org/api/path.html#pathsep).

#### Related issues

`source_file` property was introduced in #16675.

### DIFF
--- a/index.ts
+++ b/index.ts
@@ -10,7 +10,7 @@ import { fileURLToPath } from 'node:url';
 import { fdir } from 'fdir';
 
 import extend from './scripts/lib/extend.js';
-import { walk } from './utils/index.js';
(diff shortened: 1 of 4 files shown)


queengooborg (approved):

queengooborg:
Thanks @bershanskiy!  While none of the owners use Windows (the latest published version actually uses `/`), this will be helpful in case we need to pass off the build and release process to a Windows user!

PR #16685, opened 2022-06-15
Fix Identifier type

queengooborg:
This PR fixes the `Identifier` type to ensure the BCD data can be converted to the TypeScript types.

### DIFF
--- a/index.test.ts
+++ b/index.test.ts
@@ -1,0 +1,18 @@
+/* This file is a part of @mdn/browser-compat-data
+ * See LICENSE file for more information. */
+
+import { CompatStatement } from './types/types.js';
+
+import assert from 'node:assert/strict';
+
+import bcd from './index.js';
+
+describe('Using BCD', () => {
+  it('subscript notation', () => {
+    const data: CompatStatement = bcd['api']['AbortController']['__compat'];
+  });
+
+  it('dot notation', () => {
+    const data: CompatStatement = bcd.api.AbortController.__compat;
+  });
+});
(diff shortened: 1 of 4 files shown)


index.test.ts
```diff
@@ -0,0 +1,18 @@
+/* This file is a part of @mdn/browser-compat-data
+ * See LICENSE file for more information. */
+
+import { CompatStatement } from './types/types.js';
+
+import assert from 'node:assert/strict';
```
bershanskiy, line 6: > warning  'assert' is defined but never used         @typescript-eslint/no-unused-vars

queengooborg:
Hmm, apparently with this change, now everything's an "unknown type"...  Argh, I hate TypeScript...

index.test.ts
```diff
@@ -0,0 +1,16 @@
+/* This file is a part of @mdn/browser-compat-data
+ * See LICENSE file for more information. */
+
+import { CompatStatement } from './types/types.js';
+
+import bcd from './index.js';
+
+describe('Using BCD', () => {
+  it('subscript notation', () => {
+    const data: CompatStatement = bcd['api']['AbortController']['__compat'];
```
bershanskiy, line 10: > warning  'data' is assigned a value but never used

The easiest way to fix this is not to declare the variable:
```
  it('subscript notation', () => {
    bcd['api']['AbortController']['__compat'];
  });
```
Similarly on line 14.

queengooborg: The purpose of the variable was to ensure that the types match.  I've added an `assert.ok()` call so this check can still be utilized!

bershanskiy: Ah, I see, yes that makes more sense!

Issue #12261, opened 2021-09-04
Typescript - `data.json` is not of type `CompatData`


j-m:
If you try to
```ts
import * as bcd from '@mdn/browser-compat-data/data.json'
const data = bcd as CompatData
```
you will see the error 

>Conversion of type '{ api: { ANGLE_instanced_arrays: { __compat: { mdn_url: string; spec_url: string; support: { chrome: { version_added: string; }; chrome_android: { version_added: string; }; edge: { version_added: string; }; ... 8 more ...; webview_android: { ...; }; }; status: { ...; }; }; drawArraysInstancedANGLE: { ...; }; drawEle...' to type 'CompatData' may be a mistake because neither type sufficiently overlaps with the other. If this was intentional, convert the expression to 'unknown' first.
  Type '{ api: { ANGLE_instanced_arrays: { __compat: { mdn_url: string; spec_url: string; support: { chrome: { version_added: string; }; chrome_android: { version_added: string; }; edge: { version_added: string; }; ... 8 more ...; webview_android: { ...; }; }; status: { ...; }; }; drawArraysInstancedANGLE: { ...; }; drawEle...' is not comparable to type 'CompatDataIdentifiers'.
    Types of property 'api' are incompatible.
      Type '{ ANGLE_instanced_arrays: { __compat: { mdn_url: string; spec_url: string; support: { chrome: { version_added: string; }; chrome_android: { version_added: string; }; edge: { version_added: string; }; ... 8 more ...; webview_android: { ...; }; }; status: { ...; }; }; drawArraysInstancedANGLE: { ...; }; drawElementsIn...' is not comparable to type 'PrimaryIdentifier'.
        Property '"ANGLE_instanced_arrays"' is incompatible with index signature.
          Type '{ __compat: { mdn_url: string; spec_url: string; support: { chrome: { version_added: string; }; chrome_android: { version_added: string; }; edge: { version_added: string; }; firefox: { version_added: string; }; ... 7 more ...; webview_android: { ...; }; }; status: { ...; }; }; drawArraysInstancedANGLE: { ...; }; dra...' is not comparable to type 'Identifier'.
[...]

PR #16707, opened 2022-06-17
Default TS export: cast to "any" first

queengooborg:
This PR mitigates the ongoing TypeScript bug around the `Identifier` interface.  Note: this is definitely a bad practice and should be properly fixed very soon, but this issue has plagued the past four BCD releases.


bershanskiy:
> this issue has plagued the past four BCD releases.

I think this exact issue was reported back in September 2021: #12261.

PR #16721, opened 2022-06-18
Remove ratechange event from GlobalEventHandlers

queengooborg:
This PR demixes the `ratechange` event from the GlobalEventHandlers mixin.

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -6460,58 +6460,6 @@
           }
         }
       },
-      "ratechange_event": {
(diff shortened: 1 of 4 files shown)


foolip (changes requested):
Please trim this down to just the removal, the "ratechange_event" entry on `HTMLMediaElement` is all we should have here.

queengooborg:
Looks like no content updates were needed for this one!

PR #16722, opened 2022-06-18
Demix reset event from GlobalEventHandlers

queengooborg:
This PR demixes the `reset` event from the GlobalEventHandlers mixin.

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -619,6 +619,46 @@
           }
         }
       },
+      "reset_event": {
+        "__compat": {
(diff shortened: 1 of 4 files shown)


foolip (changes requested):
Looking at Chromium source there are resets events fired for `HTMLFormElement` and `XRReferenceSpace`, and BCD already has both. Can you revert the additions?

foolip (approved):
I've reverted the additions now.

Issue #16694, opened 2022-06-16
Mixed use of `null` and `undefined`


bershanskiy:
I noticed that BCD scripts use `null` and `undefined` interchangeably. This is not a problem in JavaScript, since most of the time the consuming code actually just checks for the truthfulness of these values and both of these are falsy anyway. There is no real issue with this. However, TypeScript is more pedantic and does differentiate `null` from `undefined` so occasionally we have to write code like this:
```
const f = (a: number | null) => { /* do something */ }

let a: number | undefined;
// initialize a somewhere
f(a || null);
```

In my opinion, `f(a || null);` is rather bizarre and it would make sense to pick one of `undefined` or `null`. Do you have a preference?

This is low-priority stuff since no real harm comes from mixing `null` and `undefined`. It's more about readability.

queengooborg:
Thanks for raising this issue!  I'd say that `null` would probably be the better option in most cases?

bershanskiy:
Quoting [comment](https://github.com/mdn/browser-compat-data/pull/16724#issuecomment-1159598243) by @queengooborg:

> Now that I'm seeing these changes, I'm actually thinking that `undefined` would be better. Being able to call `walk()` with no arguments is quite convenient, and contextually, `undefined` makes more sense to me!

PR #16724, opened 2022-06-18
Refactor: use optional arguments consistently

bershanskiy:
#### Summary

Use `null` instead of `undefined` for the non-optional parameters for consistency.

This PR has no visible change, CI passes.

#### Related issues

#16694

### DIFF
--- a/test/lint.ts
+++ b/test/lint.ts
@@ -16,6 +16,7 @@ import linters from './linter/index.js';
 import extend from '../scripts/lib/extend.js';
 import { walk } from '../utils/index.js';
 import { pluralize, LinterMessage, LinterMessageLevel } from './utils.js';
+import { Stats } from 'node:fs';
 
 const dirname = fileURLToPath(new URL('.', import.meta.url));
 
@@ -25,9 +26,7 @@ const dirname = fileURLToPath(new URL('.', import.meta.url));
  * @param {string[]} files The files to test
  * @returns {DataType?}
  */
-const loadAndCheckFiles = async (
-  ...files: string[]
-): Promise<DataType | undefined> => {
+const loadAndCheckFiles = async (...files: string[]): Promise<DataType> => {
   const data = {};
 
   for (let file of files) {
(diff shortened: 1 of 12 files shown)


queengooborg:
Now that I'm seeing these changes, I'm actually thinking that `undefined` would be better.  Being able to call `walk()` with no arguments is quite convenient, and contextually, `undefined` makes more sense to me!

bershanskiy:
Sure, I'll change to `undefined`. I have no strong preference of one over the other, just want the codebase to be consistent.

bershanskiy:
@queengooborg I changed the focus of this PR (updated the PR title, effectively rewrote the whole PR). I probably could have closed it and opened a new one, but somehow ended up sticking with this.

test/lint.ts
```diff
@@ -16,6 +16,7 @@ import linters from './linter/index.js';
 import extend from '../scripts/lib/extend.js';
 import { walk } from '../utils/index.js';
 import { pluralize, LinterMessage, LinterMessageLevel } from './utils.js';
+import { Stats } from 'node:fs';
```
queengooborg, line 19: Small nit: can we move this to right below the main `fs` import?

test/lint.ts
```diff
@@ -25,17 +26,15 @@ const dirname = fileURLToPath(new URL('.', import.meta.url));
  * @param {string[]} files The files to test
  * @returns {DataType?}
  */
-const loadAndCheckFiles = async (
-  ...files: string[]
-): Promise<DataType | undefined> => {
+const loadAndCheckFiles = async (...files: string[]): Promise<DataType> => {
```
queengooborg, line 29: This could still return `undefined` on line 43, but I'm now realizing that this is a bug, and it should really be a `continue;` rather than a `return;`.  Could we fix that while we're here?

queengooborg (changes requested):
Looking good, just a couple of small nits!

bershanskiy:
Done!

queengooborg (approved):
LGTM, thanks!

PR #16727, opened 2022-06-19
Demix scroll event from GlobalEventHandlers

queengooborg:
This PR demixes the `scroll` event from the GlobalEventHandlers mixin.


api/HTMLElement.json
```diff
@@ -2231,6 +2231,46 @@
           }
         }
       },
+      "scroll_event": {
```
foolip, line 2234: Can you move this entry to `Element` instead? I'm pretty sure that with the right CSS, this event could also fire on `SVGElement` or some other type of `Element`.

foolip (changes requested):

api/Document.json
```diff
@@ -6951,7 +6951,10 @@
         "__compat": {
           "description": "<code>scroll</code> event",
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Document/scroll_event",
-          "spec_url": "https://drafts.csswg.org/cssom-view/#eventdef-document-scroll",
+          "spec_url": [
+            "https://drafts.csswg.org/cssom-view/#eventdef-document-scroll",
```
foolip, line 6955: Can you update the `Element` entry with the same URLs? This one is actually the one for elements too, despite "document" in the fragment.

foolip (changes requested):

foolip (changes requested):
Please use the same 2 or 3 URLs for both entries. I'd suggest just keeping 2, https://drafts.csswg.org/cssom-view/#scrolling-events is a section about multiple events.

PR #16728, opened 2022-06-19
Demix securitypolicyviolation event from GlobalEventHandlers

queengooborg:
This PR demixes the `securitypolicyviolation` event from the GlobalEventHandlers mixin.

### DIFF
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -2231,6 +2231,44 @@
           }
         }
       },
+      "securitypolicyviolation_event": {
+        "__compat": {
+          "description": "<code>securitypolicyviolation</code> event",
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLElement/securitypolicyviolation_event",
+          "spec_url": "https://html.spec.whatwg.org/multipage/webappapis.html#handler-onsecuritypolicyviolation",
+          "support": {
+            "chrome": {
+              "version_added": "97"
+            },
+            "chrome_android": "mirror",
+            "edge": "mirror",
+            "firefox": {
+              "version_added": "93"
+            },
+            "firefox_android": "mirror",
+            "ie": {
+              "version_added": false
+            },
+            "opera": "mirror",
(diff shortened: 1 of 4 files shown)


api/HTMLElement.json
```diff
@@ -2231,6 +2231,44 @@
           }
         }
       },
+      "securitypolicyviolation_event": {
```
foolip, line 2234: There's already an entry on `Element`, so please revert this one.

api/Window.json
```diff
@@ -5596,6 +5596,44 @@
           }
         }
       },
+      "securitypolicyviolation_event": {
```
foolip, line 5599: Per https://w3c.github.io/webappsec-csp/#report-violation it looks like the event can never target `Window`, so please revert this.

foolip (changes requested):

foolip:
Per https://w3c.github.io/webappsec-csp/#report-violation and https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/frame/csp/execution_context_csp_delegate.cc;l=287-294;drc=169c6cc102b39295a5bfe2f2a176b42b1c2fe2c4 the possible event targets are `Element`, `Document` and `WorkerGlobalScope`.

The entry on `Element` already exists, but can you add/update the other two to have matching data?

queengooborg:
Are there any other changes that need to be made for this PR?

foolip:
Yes, there are differences in the data across the 3 entries that need to be explained or fixed:
- Chrome 41 or 97?
- Edge 15 or 97?
- Firefox 63 or 93?
- Safari 10 or "preview"?

I would expect that the events on `Element` and `Document` were implemented at the same time, but maybe the event on `WorkerGlobalScope` came later.

Digging through source might be simpler than testing in this case.

queengooborg:
I merged this PR while there were changes requested because I wanted to defer version number updates to a future PR so it wouldn't block the demixing of GEH.  I do plan to follow-up on this using results from the collector!

PR #16731, opened 2022-06-19
Demix select event from GlobalEventHandlers

queengooborg:
This PR demixes the `select` event from the GlobalEventHandlers mixin.


api/Document.json
```diff
@@ -7019,6 +7019,49 @@
           }
         }
       },
+      "select_event": {
```
foolip, line 7022: `Document` isn't the target, I'll revert this.

foolip (changes requested):
In this case I think we should add "select_event" entries to `HTMLInputElement` and `HTMLTextAreaElement`. These are the interfaces that have `selectionStart` and related APIs that can end up firing this event.

Please remove the existing "select_event" entry on `Element`.

foolip (approved):

PR #16732, opened 2022-06-19
Remove show event from GlobalEventHandlers

queengooborg:
This PR demixes the `show` event from the GlobalEventHandlers mixin.


queengooborg:
Looks like no content updates are needed for this one!

foolip:
I think updates are needed to this page, which refer to an `onshow` property:
https://developer.mozilla.org/en-US/docs/Web/API/Element/show_event

foolip (approved):

queengooborg:
I've submitted a content PR (see https://github.com/mdn/content/pull/18167), but I'm going to merge this PR before the content PR is because it does not _block_ this PR.

PR #16733, opened 2022-06-19
Demix slotchange event from GlobalEventHandlers

queengooborg:
This PR demixes the `slotchange` event from the GlobalEventHandlers mixin.


foolip (changes requested):
Can you revert the additions? There are already "slotchange_event" entries for `HTMLSlotElement` and `ShadowRoot`. I can only see it firing on `HTMLSlotElement` in Chromium, but I'm not confident enough about it to suggest removing the `ShadowRoot` entry. I'm asking in https://github.com/whatwg/dom/pull/785#issuecomment-1164539731 but that'll not block this PR.

foolip (approved):

PR #16734, opened 2022-06-19
Demix stalled event from GlobalEventHandlers

queengooborg:
This PR demixes the `stalled` event from the GlobalEventHandlers mixin.

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -7129,6 +7129,58 @@
           }
         }
       },
+      "stalled_event": {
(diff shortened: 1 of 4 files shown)


queengooborg:
This doesn't need content updates, looks like!

foolip (changes requested):
This is another `HTMLMediaElement` event, can you revert the additions?

foolip (approved):

PR #16735, opened 2022-06-19
Demix submit event from GlobalEventHandlers

queengooborg:
This PR demixes the `submit` event from the GlobalEventHandlers mixin.


foolip (approved):
I've made the edits to keep this for just `HTMLFormElement`.

PR #16736, opened 2022-06-19
Demix suspend event from GlobalEventHandlers

queengooborg:
This PR demixes the `suspend` event from the GlobalEventHandlers mixin.


foolip (changes requested):
This is also an `HTMLMediaElement` event. Please revert the additions.

foolip (approved):

PR #16737, opened 2022-06-19
Remove timeupdate event from GlobalEventHandlers

queengooborg:
This PR demixes the `timeupdate` event from the GlobalEventHandlers mixin.

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -7194,58 +7194,6 @@
           }
         }
       },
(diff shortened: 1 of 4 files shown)


foolip (changes requested):
There's already a "timeupdate_event" entry on `HTMLMediaElement` and that's the only one we should have. Can you trim this to just the removal?

foolip (approved):

PR #16749, opened 2022-06-21
Update the CSS color-contrast spec URL

sideshowbarker:
https://drafts.csswg.org/css-color-5/#changes-20220428 notes that `color-contrast()` moved to the Level 6 spec

### DIFF
--- a/css/types/color.json
+++ b/css/types/color.json
@@ -179,7 +179,7 @@
           "__compat": {
             "description": "<code>color-contrast()</code>",
             "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/color_value/color-contrast",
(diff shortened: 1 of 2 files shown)


queengooborg (approved):

queengooborg:
This will need a new `browser-specs` version I believe, but one we've gotten an update I'll merge this PR!

queengooborg:
(Alternatively, we could add this as an exception while we await a `browser-specs` update?)

sideshowbarker:
> (Alternatively, we could add this as an exception while we await a `browser-specs` update?)

👍 Pushed a commit that adds such an exeption

queengooborg (approved):

PR #16753, opened 2022-06-21
Application Cache has been removed

bershanskiy:
#### Summary

Application Cache has been removed from the Web Platform. The history of its removal is pretty convoluted and is probably not worth the effort to track. (There were multiple stages of deprecation, removal from some contexts, reverse/deprecation Origin Trial.) I think we just should set `version_removed` to `true` to let people know it should not be used and then just remove it completely in a few years.

https://web.dev/appcache-removal/

#### Related issues

N/A


queengooborg (approved):
Thanks, LGTM!

PR #16759, opened 2022-06-21
Demix touch events from GlobalEventHandlers

queengooborg:
This PR demixes the `touch` events from the GlobalEventHandlers mixin.


foolip (changes requested):
Can you keep the entries on `Element` and remove everything else?

foolip (approved):
I reverted some stuff and merged the spec URLs. I doubt the entries on `Element` are 100% correct now, but the data is pretty consistent with what's being removed, so if there are errors I think that's best dealt with after this PR.

PR #16761, opened 2022-06-21
Demix waiting event from GlobalEventHandlers

queengooborg:
This PR demixes the `waiting` event from the GlobalEventHandlers mixin.

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -2681,6 +2681,58 @@
           }
         }
       },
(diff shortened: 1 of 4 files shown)


foolip (changes requested):
Another `HTMLMediaElement` event, can you revert the additions?

foolip (approved):

PR #16762, opened 2022-06-21
Demix wheel event from GlobalEventHandlers

queengooborg:
This PR demixes the `wheel` event from the GlobalEventHandlers mixin.

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -6501,7 +6501,10 @@
         "__compat": {
           "description": "<code>wheel</code> event",
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/Element/wheel_event",
-          "spec_url": "https://w3c.github.io/uievents/#event-type-wheel",
+          "spec_url": [
+            "https://w3c.github.io/uievents/#event-type-wheel",
+            "https://html.spec.whatwg.org/multipage/webappapis.html#handler-onwheel"
+          ],
           "support": {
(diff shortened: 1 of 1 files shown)


foolip (changes requested):
Let's keep only the entry on `Element`, that's the target per https://w3c.github.io/uievents/#event-type-wheel and the event bubbles.

These differences in the data need to be investigated:
- Chrome 31 or 61?
- Supported on Safari iOS or not?

foolip:
The question of Chrome 31 or 61 still needs addressing here.

queengooborg:
I just ran the collector and it looks like this is a more complicated issue than originally anticipated.  As reported by the collector, on `Element`, support was introduced in Chrome 61, however on `Document`, it was supported since Chrome 31.

We should probably keep the event on `Document` as well.

foolip:
@queengooborg did you use http://mdn-bcd-collector.appspot.com/tests/api/HTMLElement/wheel_event? There's no test for `Element`...

foolip:
Using http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=10459 I've confirmed Chrome 31 to be the correct version.

foolip (approved):

foolip:
> We should probably keep the event on `Document` as well.

I don't think so, or can you create a test case where the target is a `Document` and not some `Element`?

PR #16764, opened 2022-06-22
Move null from Global Objects to Operators/Expressions

Josh-Cena:
#### Summary

In accordance with https://github.com/mdn/content/pull/17546


queengooborg (approved):
LGTM based upon the issue history over on mdn/content!

PR #16766, opened 2022-06-22
Move import.meta from Statements to Operators/Expressions

Josh-Cena:
#### Summary

In accordance with https://github.com/mdn/content/pull/17547


queengooborg (approved):
This move seems quite odd to me as I wouldn't call this an operator or expression, but I see that the content PR was merged to perform this change, so I'll go along with it!  👍

PR #16787, opened 2022-06-24
Update Samsung Internet versions for AuthenticatorAssertionResponse API

queengooborg:
This PR updates and corrects the real values for Samsung Internet for the `AuthenticatorAssertionResponse` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.0.7).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/AuthenticatorAssertionResponse

### DIFF
--- a/api/AuthenticatorAssertionResponse.json
+++ b/api/AuthenticatorAssertionResponse.json
@@ -40,9 +40,7 @@
             "version_added": "13"
           },
           "safari_ios": "mirror",
-          "samsunginternet_android": {
-            "version_added": false
-          },
+          "samsunginternet_android": "mirror",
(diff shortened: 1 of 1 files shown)


foolip:
I see we have results in https://github.com/foolip/mdn-bcd-results/blob/main/6.0.7-samsunginternet-android-17.0-android-10-5baf37f5d6.json, cool!

@queengooborg it would be useful for the long term I think if we could link to the test results from these PRs. The raw files are better than nothing, but ideally a view similar to what you get after running the tests, before submitting them to the collector storage.

foolip (approved):

PR #16740, opened 2022-06-20
Linter: add message about exceptions

bershanskiy:
#### Summary

Add a nice linter message.

#### Test results and supporting details

This is entirely a cosmetic change, it just adds a message to the test output after `All data passed linting!`:
```
Loading and checking files...
Testing browser data...
Testing feature data...
Testing all features together...
All data passed linting!
Linters have some exceptions, please help us remove them!
 Obsolete has 10 exceptions.
  - css.types.length.lh
  - css.types.length.rlh
  - http.headers.Cache-Control.stale-if-error
  - http.headers.Feature-Policy.layout-animations
  - http.headers.Feature-Policy.legacy-image-formats
  - http.headers.Feature-Policy.oversized-images
  - http.headers.Feature-Policy.unoptimized-images
  - http.headers.Feature-Policy.unsized-media
  - svg.attributes.presentation.color-rendering
  - svg.elements.view.zoomAndPan
```
This message increases visibility of exceptions and I aspire to literally guilt-trip myself into fixing them. 😄 

#### Related issues

#16579

PR #16807, opened 2022-06-27
Linter: fix linter exception logging

### LINKED ISSUE #16785: Unneeded linter exceptions check will display obsolete exceptions when specifying custom fileset
Recently, the ability to add exceptions to linters was implemented, so that linters may allow failures for specific features or files.  Any obsolete exceptions will also be reported if the lint for a specific feature/file _should_ fail, but doesn't.

However, if a custom scope of files is specified when running the lint command (ex. `npm run lint api/AbortController.json`), any features that were not present in the specified file will not throw an error, thus it gets reported as an obsolete exception.

CC @bershanskiy

bershanskiy:
#### Summary

#16740 added a message about linter exceptions, but contained a bug described in #16785. This PR:
 - fixes said bug
 - makes existence of known unnecessary exceptions into a lint error (previously it was just a log statement)

Ran `npm run lint` and `npm run lint api/AbortController.json` and observed expected output. Then added an unneeded exception `api.AbortController` and ran these commands again and observed expected lint failure.

#### Related issues

Fixes #16785.

### DIFF
--- a/test/lint.ts
+++ b/test/lint.ts
@@ -187,12 +187,15 @@ const main = async (
   // Find all unnecessary linting exceptions
   for (const linter of linters.linters) {
     if (linter.exceptions) {
-      const actual = new Set(linters.expectedFailures[linter.name]);
-      for (const exception of linter.exceptions)
-        if (!actual.has(exception))
+      const missingExceptions = linters.missingExpectedFailures[linter.name];
+      for (const exception in missingExceptions) {
+        if (missingExceptions[exception]) {
+          hasErrors = true;
           console.error(
             chalk`{red  ✖ ${linter.name} - Unnecessary exception → ${exception}}`,
           );
+        }
+      }
     }
   }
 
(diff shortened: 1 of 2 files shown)


test/lint.ts
```diff
@@ -187,12 +187,15 @@ const main = async (
   // Find all unnecessary linting exceptions
   for (const linter of linters.linters) {
     if (linter.exceptions) {
-      const actual = new Set(linters.expectedFailures[linter.name]);
-      for (const exception of linter.exceptions)
-        if (!actual.has(exception))
+      const missingExceptions = linters.missingExpectedFailures[linter.name];
+      for (const exception in missingExceptions) {
+        if (missingExceptions[exception]) {
+          hasErrors = true;
```
bershanskiy, line 193: I think unnecessary exceptions should be treated as a lint error, but will be glad to revert this if needed.

queengooborg (approved):
Thanks for tackling this, this is LGTM!

PR #16816, opened 2022-06-28
Chrome 104 has support for individual transform properties

bramus:
#### Summary

Chrome 104 has support for individual transform properties

#### Test results and supporting details\

ChromeStatus Entry: https://chromestatus.com/feature/5705698193178624
Chrome Tracking Bug: https://bugs.chromium.org/p/chromium/issues/detail?id=496550

### DIFF
--- a/css/properties/rotate.json
+++ b/css/properties/rotate.json
@@ -7,8 +7,7 @@
           "spec_url": "https://drafts.csswg.org/css-transforms-2/#individual-transforms",
           "support": {
             "chrome": {
-              "version_added": false,
-              "notes": "See <a href='https://crbug.com/496550'>bug 496550</a>."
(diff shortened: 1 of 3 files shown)


css/properties/rotate.json
```diff
@@ -7,8 +7,7 @@
           "spec_url": "https://drafts.csswg.org/css-transforms-2/#individual-transforms",
           "support": {
             "chrome": {
-              "version_added": false,
-              "notes": "See <a href='https://crbug.com/496550'>bug 496550</a>."
+              "version_added": "104"
```
foolip, line 10: Can you also check the "x_y_z_angle" in this file? What does it mean, and is it supported?

foolip (approved):

PR #16825, opened 2022-06-29
Remove copy/cut/paste events except for on Element

foolip:
MDN was changed to only use the data for Element:
https://github.com/mdn/content/pull/17694

These events are fired on Elements and bubble, so we should have them in
just one place.


foolip:
There's a bit of complexity here around the `oncopy`/`oncut`/`onpaste` event handler properties. As a removed note states, `document.oncopy` and friends were missing in IE and early Edge, which I've confirmed with http://mdn-bcd-collector.appspot.com/tests/api/Document/copy_event.

But the same event handler properties are also missing on `Window`, which is *per spec*, but doesn't really make sense. I filed https://github.com/whatwg/html/issues/8053 about that and expect we will add those to the spec at some point, and that it'll trickle into implementations.

But I don't think the absence of these event handler properties on `document` or `window` is serious enough to call out with notes.

api/Element.json
```diff
@@ -1663,8 +1666,12 @@
               "version_added": "9",
               "notes": "Before Internet Explorer 9, this event is not supported via <code>addEventListener</code>; however, the event handler is supported since IE 5.5. The event can be listened to via <code>element.oncopy</code>."
             },
-            "opera": "mirror",
-            "opera_android": "mirror",
+            "opera": {
+              "version_added": "≤12.1"
```
foolip, line 1670: I tested http://mdn-bcd-collector.appspot.com/tests/api/HTMLElement/copy_event in Opera 12.16 for this.

queengooborg (approved):

PR #16840, opened 2022-07-04
Add Intl.NumberFormat V3 features that are experimental

teoli2003:
In mdn/content#17883 we marked these features as experimental in the docs: they are in the spec and are implemented in Firefox (in Nightly).

This adds them inside bcd.

This is the last bit to fix mdn/content#17848.

### DIFF
--- a/javascript/builtins/intl/NumberFormat.json
+++ b/javascript/builtins/intl/NumberFormat.json
@@ -255,6 +255,120 @@
                 }
               }
             },
+            "options_roundingIncrement_parameter": {
+              "__compat": {
+                "description": "<code>options.roundingIncrement</code> parameter",
+                "support": {
+                  "chrome": {
+                    "version_added": false
+                  },
+                  "chrome_android": "mirror",
+                  "deno": {
+                    "version_added": false
+                  },
+                  "edge": "mirror",
+                  "firefox": {
+                    "version_added": "nightly"
+                  },
+                  "firefox_android": "mirror",
+                  "ie": {
+                    "version_added": false
+                  },
+                  "nodejs": {
+                    "version_added": false
+                  },
+                  "opera": "mirror",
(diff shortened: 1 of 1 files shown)


teoli2003:
@queengooborg Any idea what is wrong here? I don't understand the error message. (First time using `preview` and `mirror`)

javascript/builtins/intl/NumberFormat.json
```diff
@@ -255,6 +255,120 @@
...
+                  },
+                  "chrome_android": "mirror",
+                  "deno": {
+                    "version_added": false
+                  },
+                  "edge": "mirror",
+                  "firefox": {
+                    "version_added": "preview"
+                  },
+                  "firefox_android": "mirror",
```
hamishwillee, line 273: Is there an android nightly?

hamishwillee:
FYI, @queengooborg this can be reproduced locally by just starting from main and changing any of the options to have firefox version set to preview.

javascript/builtins/intl/NumberFormat.json
teoli2003: There is: https://play.google.com/store/apps/details?id=org.mozilla.fenix&hl=en&gl=US

I use it on my phone for ages.

hamishwillee: Thanks. I guess we'll have to be careful with "mirror", I have another PR where stuff is added behind a preference, so "mirror" is no longer true because you can't set prefs on android.

queengooborg:
Lint has been fixed!  The mirroring script didn't account for when the upstream version is set to `preview`...which I'm surprised hasn't been an issue before.

hamishwillee:
Awesome @queengooborg - so can you review this?

queengooborg (approved):

PR #16847, opened 2022-07-04
Demix cancel event from GlobalEventHandlers

foolip:
The event on `<input type=file>` was missing and implemented only in
Firefox:
https://github.com/whatwg/html/pull/6735
https://bugs.chromium.org/p/chromium/issues/detail?id=1227424
https://bugzilla.mozilla.org/show_bug.cgi?id=1719703
https://bugs.webkit.org/show_bug.cgi?id=227799


queengooborg (approved):

foolip:
@queengooborg can you do the content updates for this?

PR #16854, opened 2022-07-05
FF103 Supports Array.findLast and .findLastIndex

hamishwillee:
FF103 adds supports for `Array.findLast` and `.findLastIndex` in https://bugzilla.mozilla.org/show_bug.cgi?id=1704385
 behind a preference.
 
 Other docs work for this can be tracked in https://github.com/mdn/content/issues/17473

### DIFF
--- a/javascript/builtins/Array.json
+++ b/javascript/builtins/Array.json
@@ -535,9 +535,18 @@
               },
               "edge": "mirror",
               "firefox": {
+                "version_added": "103",
+                "flags": [
+                  {
+                    "type": "preference",
+                    "name": "javascript.options.experimental.array_find_last",
+                    "value_to_set": "true"
+                  }
+                ]
+              },
+              "firefox_android": {
                 "version_added": false
               },
-              "firefox_android": "mirror",
               "ie": {
                 "version_added": false
               },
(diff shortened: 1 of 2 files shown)


javascript/builtins/Array.json
```diff
@@ -535,9 +535,18 @@
...
+                    "type": "preference",
+                    "name": "javascript.options.experimental.array_find_last",
+                    "value_to_set": "true"
+                  }
+                ]
+              },
+              "firefox_android": {
                 "version_added": false
               },
-              "firefox_android": "mirror",
```
hamishwillee, line 540: @queengooborg FYI, my assumption is that "mirror" is not smart enough to know that android shouldn't mirror preferenced versions. So removing and marking as false.

foolip: It should do the right thing actually:

https://github.com/mdn/browser-compat-data/blob/b53950916c1aeba50dd0fdcc0f6a9395aed6615b/scripts/release/mirror.ts#L354-L357

hamishwillee: That's actually pretty cool! I'll revert those.

foolip (approved):

PR #16856, opened 2022-07-05
Remove *.toSource() and uneval()

teoli2003:
These methods, only supported in Firefox, got removed more than 2 years ago.

This is the counterpart of mdn/content#17982 and is part of of mdn/content#17962


javascript/builtins/Boolean.json
```diff
@@ -91,93 +91,6 @@
...
-              "webview_android": "mirror"
-            },
-            "status": {
-              "experimental": false,
-              "standard_track": false,
-              "deprecated": false
-            }
-          }
-        },
-        "toString": {
```
foolip, line 136: Oops, this is `toString`, no `toSource`.

javascript/builtins/globals.json
```diff
@@ -709,49 +709,6 @@
             "deprecated": true
           }
         }
-      },
-      "uneval": {
```
foolip, line 713: Can you add this to the PR title/description, or split into a separate PR?

teoli2003: I'm adding it to the title. They have been removed in the same Firefox bug, so it makes sense to keep their removal together here.

javascript/builtins/Boolean.json
teoli2003: Whoopsie, I readded `toString()`; sorry.

queengooborg (approved):
LGTM, thanks!

PR #16862, opened 2022-07-07
Chrome v1.0 supports meta: http-equiv: refresh

Christopher-Hayes:
Chrome has supported `<meta http-equiv="refresh"` since v1.0

## Summary

Currently, Chrome support for the `<meta http-equiv="refresh"` is not specified. Chrome has had support for this feature since version 1.0 release in 2008. Chromium's code-base shows this feature being present as early as 2001.

## Test results and supporting details

The browser logic for this feature can be found in the Chromium source code below.

Ran into trouble trying to actually test a v1.0 Chrome release due to build/dependency issues. Ideally would've tested it against the [W3C spec tests specified here](https://www.w3.org/WAI/GL/WCAG20/WD-WCAG20-TECHS-20081105/H76).

### Source Control
Head element logic for the "refresh" http-equiv property committed to the Chromium project on August 24, 2001.

This logic was found in `third_party/WebKit/WebCore/khtml/html/html_headimpl.cpp` and in later versions of Chromium this logic was moved to `third_party/blink/renderer/core/dom/document.cc`

**Git Commit - Authored by Ken Kocienda on August 24, 2001:** https://github.com/chromium/chromium/commit/d869b93fe74f4d6cb2dd6f6c3e9bf9daee39ba19#diff-236cb39df64dab92a4433fa67f6ed58b3d94d77d8dddafdc8fe2334840aeccdeR260

**More readable version of the relevant file in the commit above:** https://github.com/chromium/chromium/blame/4dcda448a85684cf52451a520ec254288b497fdd/third_party/WebKit/WebCore/khtml/html/html_headimpl.cpp#L266

## Motivation

The motivation to update the version on this now rarely used feature is for caniuse.com to better represent the support of `http-equiv="refresh"`, right now [it's only 65% since Chrome is marked as "Unknown"](https://caniuse.com/mdn-html_elements_meta_http-equiv_refresh).

### DIFF
--- a/html/elements/meta.json
+++ b/html/elements/meta.json
@@ -248,7 +248,7 @@
             "__compat": {
               "support": {
                 "chrome": {
-                  "version_added": true
+                  "version_added": "1"
                 },
                 "chrome_android": "mirror",
                 "edge": {


Christopher-Hayes:
Aside - had a little trouble understanding what exactly `"version_added": true` meant. My assumption with this PR is that it's been there since forever, but it was never tracked down to a specific browser version, considering Firefox does have `"version_added": "1"`. Couldn't find anything about using `true` there in the docs: https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md

foolip:
@Christopher-Hayes thanks for the contribution, and welcome to BCD! The documentation on what `true` means is here:
https://github.com/mdn/browser-compat-data/blob/main/schemas/compat-data-schema.md#version_added

To summarize, it means the feature is supported, but it's not known since when. In part of the data (api/, css/ and javascript/) all `true` values have been replaced with specific versions, but not for HTML.

If your motivation is getting better data into caniuse.com for the availability percentage, then figuring out the right versions of Safari and IE would also be good. I'll see if I can help figure that out.

foolip:
https://github.com/chromium/chromium/blame/4dcda448a85684cf52451a520ec254288b497fdd/third_party/WebKit/WebCore/khtml/html/html_headimpl.cpp#L266 is initial revision of WebKit, so Safari can also be set to "1".

For IE I think some testing will be needed.

foolip (approved):

foolip:
I haven't tested, but found some articles that confirm support in IE6:
https://zoompf.com/blog/2010/03/meta-refresh-nullifies-caching-for-ie6-and-ie7/
https://seclists.org/fulldisclosure/2005/Aug/574

I'll send separate PRs for that and Safari though.

foolip:
I've fixed the data for other browsers in https://github.com/mdn/browser-compat-data/pull/16868.

Christopher-Hayes:
Great, thanks for the info @foolip !

PR #3330, opened 2019-01-23
Splitting up compat data for break-after

rachelandrew:
To match other updated compat data. I am splitting up the data into multicol, paged media and regions i order to better show support in these different contexts.

Related to MDN issue: https://github.com/mdn/sprints/issues/734

Also compat data PRs:

* https://github.com/mdn/browser-compat-data/pull/3327
* https://github.com/mdn/browser-compat-data/pull/3315

PR #16866, opened 2022-07-07
Update support for break-before/break-after: avoid-column

foolip:
These are supported since Chromium M102 according to the implementer:
https://github.com/mdn/browser-compat-data/pull/16628#issuecomment-1153568563

The test case in that comment was used to confirm support for
break-before:avoid-column in IE10 and Edge 18.

IE10 support for break-after:avoid-column was not tested, but set
because it appears to have been set to false by accident:
https://github.com/mdn/browser-compat-data/pull/3330

The same split for break-before didn't make the same change:
https://github.com/mdn/browser-compat-data/pull/3374

### DIFF
--- a/css/properties/break-after.json
+++ b/css/properties/break-after.json
@@ -130,19 +130,24 @@
               "description": "<code>avoid-column</code>",
               "support": {
                 "chrome": {
-                  "version_added": false
+                  "version_added": "102"
                 },
                 "chrome_android": "mirror",
-                "edge": {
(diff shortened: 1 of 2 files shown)


foolip:
@mstensho this is based on your comments.

@rachelandrew I've asked for your review since it was your PR that changed IE10 support for `break-after:avoid-column` to false. It looks like an accident, but if it wasn't (and you remember) please let me know and I'll revert.

PR #16867, opened 2022-07-07
Update support of break-before/after/inside properties in Opera

foolip:
This is according to the implementer of this:
https://github.com/mdn/browser-compat-data/pull/16628#issuecomment-1153568563

### DIFF
--- a/css/properties/break-after.json
+++ b/css/properties/break-after.json
@@ -165,8 +165,24 @@
                   "version_added": false
                 },
                 "oculus": "mirror",
-                "opera": "mirror",
-                "opera_android": "mirror",
+                "opera": [
+                  {
+                    "version_added": "88"
+                  },
+                  {
+                    "version_added": "11.1",
+                    "version_removed": "12.1"
+                  }
+                ],
+                "opera_android": [
+                  {
+                    "version_added": "70"
+                  },
+                  {
+                    "version_added": "11.1",
+                    "version_removed": "12.1"
+                  }
+                ],
                 "safari": {
                   "version_added": false
                 },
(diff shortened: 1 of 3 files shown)


css/properties/break-after.json
```diff
@@ -109,8 +109,24 @@
...
+                    "version_added": "88"
+                  },
+                  {
+                    "version_added": "11.1",
+                    "version_removed": "12.1"
+                  }
+                ],
+                "opera_android": [
+                  {
+                    "version_added": "mirror"
```
foolip, line 123: This doesn't work, but I've put this here because we need a version of Opera Android based on 102 before we can submit this PR.

foolip:
@mstensho was the `avoid-column` keyword also supported? It was recognized by the parser, but the test for it in https://github.com/mdn/browser-compat-data/pull/16628#issuecomment-1153568563 doesn't pass in Opera 12.16, so I'm not sure.

Getting all of this exactly right isn't important, so I wouldn't suggest spending a lot of time on testing if you don' remember.

mstensho:
Yes, Presto also supports avoid-column. The reason why the test that you refer to doesn't pass, is that Presto doesn't support size containment. Replacing contain:size with break-inside:avoid should make it pass in Presto.

css/properties/break-inside.json
```diff
@@ -75,8 +75,24 @@
                 "ie": {
                   "version_added": "10"
                 },
-                "opera": "mirror",
-                "opera_android": "mirror",
+                "opera": [
```
foolip, line 78: @mstensho I confirmed parser support for `break-inside: avoid-column` in Presto and am assuming it was properly supported. Can you confirm?

css/properties/break-after.json
foolip: @queengooborg I don't think this kind of situation will happen often, but any thoughts on how we could support mirroring here too? The problem is that different versions of Opera Android should have different upstream browsers, but we only support a single upstream.

queengooborg: I did have a PR that introduced this functionality a while back, but I closed it as it presented many new problems with little benefit (not to mention, the PR that it depended on, which did the same for IE+Edge, was closed as well).  While we could change the upstream mirroring target to Opera Desktop instead, it's much more likely for compatibility to match Chrome Android.

foolip: I guess I'll just wait for a new Opera Android release, leaving this as draft until then.

foolip:
There's been an Opera Android release, so I've updated this, ready to go now!

queengooborg (approved):

PR #16876, opened 2022-07-08
FF103 transferrable streams

hamishwillee:
FF103 makes some stream types transferrable by default which were previously behind a pref and on nightly. This just changes nightly to FF103.

The associated bug is https://bugzilla.mozilla.org/show_bug.cgi?id=1770627

Other docs work can be tracked in https://github.com/mdn/content/issues/17472

### DIFF
--- a/api/ReadableStream.json
+++ b/api/ReadableStream.json
@@ -378,9 +378,7 @@
                 ]
               }
             ],
-            "firefox_android": {
-              "version_added": "mirror"
-            },
+            "firefox_android": "mirror",
             "ie": {
               "version_added": false
             },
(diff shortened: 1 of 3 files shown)


api/ReadableStream.json
```diff
@@ -379,7 +379,7 @@
               }
             ],
             "firefox_android": {
-              "version_added": false
+              "version_added": "mirror"
             },
```
queengooborg, line 383: This should actually be the following:

```js
"firefox_android": "mirror"
```

(I can't write a suggestion for some odd reason...)

queengooborg (changes requested):

api/ReadableStream.json
hamishwillee: THanks, fixed!

queengooborg (approved):
LGTM, thanks!

PR #16882, opened 2022-07-08
Treat accept_flags as optional (default true) in mirroring

foolip:
This is how the property is used, with oculus.json being the single
outlier of setting accept_flags to true. In test-versions.js,
`browsers[browser].accepts_flags === false` is used as the condition.

This will allow much more data to be mirrored.

PR #16884, opened 2022-07-09
Assume "false" for accepts_flags if not specified

queengooborg:
This PR updates the flag-preventing linter to assume that `accepts_flags` is `false` if not specified.  This retains consistency with how we treat `accepts_webextensions`.  This PR closes #16882 since the other PR will change our scripts to do the opposite.

### DIFF
--- a/test/linter/test-versions.ts
+++ b/test/linter/test-versions.ts
@@ -225,12 +225,10 @@ function checkVersions(
           }
         }
 
-        if ('flags' in statement) {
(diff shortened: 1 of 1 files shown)


foolip:
More changes are needed to fix the mirroring, I'll open a new PR.

foolip:
Oh, you did #16883 too. That will have changed a bunch of the data, I'll comment on that PR to list the changes.

PR #16887, opened 2022-07-09
Run scripts/migrations/009-mirror yet again

foolip:
After https://github.com/mdn/browser-compat-data/pull/16883 a lot more data can be mirrored.


foolip:
I've confirmed that there are no changes whatsoever in the resulting `build/data.json`.

queengooborg (approved):

PR #16894, opened 2022-07-10
Migration 010: Set Oculus to "mirror" for all features

queengooborg:
This PR adds a migration script to set Oculus Browser to "mirror" for all features, so that we can later fill in proper data with the collector.

### DIFF
--- a/scripts/migrations/010-set-oculus-to-mirror.ts
+++ b/scripts/migrations/010-set-oculus-to-mirror.ts
@@ -1,0 +1,112 @@
+/* This file is a part of @mdn/browser-compat-data
+ * See LICENSE file for more information. */
+
+import { CompatStatement, SimpleSupportStatement } from '../../types/types.js';
+import { InternalSupportStatement } from '../../types/index.js';
+
+import fs from 'node:fs';
+import path from 'node:path';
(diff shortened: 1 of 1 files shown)


scripts/migrations/010-set-oculus-to-mirror.ts
```diff
@@ -0,0 +1,112 @@
...
+
+import esMain from 'es-main';
+
+import { IS_WINDOWS } from '../../test/utils.js';
+
+const dirname = fileURLToPath(new URL('.', import.meta.url));
+
+/**
+ * Check to see if the key is __compat and modify the value to remove
+ * flags from WebView Android.
```
foolip, line 19: @queengooborg it looks like this script was based on scripts/migrations/002-remove-webview-flags.ts and there many things that weren't renamed. Spotted as `fixWebViewFlags` first, and that confused me...

queengooborg: Whoops, would you be down to open a PR to fix up these comments?

foolip: It's not a simple search and replace, I'd have to pick new names and explain what functions do what, so I'd rather not.

PR #16902, opened 2022-07-11
Simplify statements for api.BaseAudioContext.createPeriodicWave

queengooborg:
This PR simplifies the Chromium support statements for `api.BaseAudioContext.createPeriodicWave`.  There is no need to use two separate support statements, as a single note and a single statement is sufficient.

### DIFF
--- a/api/BaseAudioContext.json
+++ b/api/BaseAudioContext.json
@@ -640,24 +640,11 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/BaseAudioContext/createPeriodicWave",
           "spec_url": "https://webaudio.github.io/web-audio-api/#dom-baseaudiocontext-createperiodicwave",
           "support": {
-            "chrome": [
-              {
-                "version_added": "59",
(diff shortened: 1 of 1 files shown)


api/BaseAudioContext.json
```diff
@@ -640,24 +640,11 @@
...
-                "version_added": "59",
-                "notes": "Default values supported"
-              },
-              {
-                "version_added": "30"
-              }
-            ],
+            "chrome": {
+              "version_added": "30",
+              "notes": "Before Chrome 59, all arguments must be specified."
```
foolip, line 645: Can you say in more detail what didn't work before Chrome 59? I've looked at the source and can't find the the optionality of arguments changed in 59.

foolip (changes requested):

api/BaseAudioContext.json
foolip: More research needed then :)

foolip (approved):

PR #16909, opened 2022-07-11
Remove never-supported JavaScript features

queengooborg:
This PR uses the script from #6809 to remove JavaScript features that have no support in any browser.


teoli2003:
I know that we likely have merged `Temporal` too early here, but are these features that will never be supported or is this an artifact of being in the progress of being implemented. If it is the latter, I'm not convinced this is a useful removal: it will come back soon(ish).

cc/ @meyerweb for more insight.

PR #16913, opened 2022-07-11
Remove incorrect computed aspect ratio data

jakearchibald:
#### Summary

Computed aspect ratio was never spec'd or implemented for these elements.

https://github.com/whatwg/html/issues/4961#issuecomment-539225576 - it was decided that this feature would only apply to `img`, `video`, and `<input type="image">`.

### DIFF
--- a/html/elements/embed.json
+++ b/html/elements/embed.json
@@ -74,52 +74,6 @@
             }
           }
         },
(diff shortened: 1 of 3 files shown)


foolip (approved):
Thanks Jake!

PR #16923, opened 2022-07-12
NumberFormat/PluralRules updates

hamishwillee:
FF supports a number of `Intl.NumberFormat` items in nightly from https://bugzilla.mozilla.org/show_bug.cgi?id=1648137

This includes:
- `Intl.NumberFormat.format()` - subfeature `number_param_string_is_decimal` to reflect spec update that strings are treated as  Decimal rather than Number - see https://github.com/tc39/proposal-intl-numberformat-v3#interpret-strings-as-decimals-ecma-402-334
- `Intl.NumberFormat.formatRange()`
- `Intl.NumberFormat.formatRangeToParts()`
- `Intl.PluralRules.selectRange()`

### DIFF
--- a/javascript/builtins/intl/NumberFormat.json
+++ b/javascript/builtins/intl/NumberFormat.json
@@ -654,7 +654,7 @@
                 "deprecated": false
               }
             },
-            "number_param_string_is_decimal": {
+            "number_parameter.string_decimal": {
               "__compat": {
                 "description": "<code>number</code> param string value is decimal (not <code>Number</code>)",
                 "support": {
--- a/javascript/builtins/intl/NumberFormat.json
+++ b/javascript/builtins/intl/NumberFormat.json
@@ -908,6 +908,44 @@
                 "standard_track": true,
                 "deprecated": false
               }
+            },
+            "result_useGrouping_property": {
+              "__compat": {
+                "description": "The result <code>useGrouping</code> parameter may contain <code>'always'</code>, ode>'auto'</code>, <code>'min2'</code>, <code>true</code> and <code>false</code> (previous versions may only ntain <code>true</code> and <code>false</code>)",
+                "support": {
+                  "chrome": {
+                    "version_added": false
+                  },
+                  "chrome_android": "mirror",
+                  "deno": {
+                    "version_added": false
+                  },
+                  "edge": "mirror",
+                  "firefox": {
+                    "version_added": "preview"
+                  },
+                  "firefox_android": "mirror",
+                  "ie": {
+                    "version_added": false
+                  },
+                  "nodejs": {
+                    "version_added": false
+                  },
+                  "opera": "mirror",
+                  "opera_android": "mirror",
+                  "safari": {
+                    "version_added": false
+                  },
+                  "safari_ios": "mirror",
+                  "samsunginternet_android": "mirror",
+                  "webview_android": "mirror"
+                },
+                "status": {
+                  "experimental": true,
+                  "standard_track": true,
+                  "deprecated": false
+                }
+              }
             }
           },
           "supportedLocalesOf": {
--- a/javascript/builtins/intl/NumberFormat.json
+++ b/javascript/builtins/intl/NumberFormat.json
@@ -859,44 +859,6 @@
                 "experimental": false,
                 "standard_track": true,
                 "deprecated": false
-              },
-              "result_useGrouping_property": {
-                "__compat": {
-                  "description": "The result property <code>useGrouping</code> may contain <code>'always'</code>, <code>'auto'</code>, <code>'min2'</code>, <code>true</code> and <code>false</code> (previous versions may only contain <code>true</code> and <code>false</code>)",
-                  "support": {
-                    "chrome": {
-                      "version_added": false
-                    },
-                    "chrome_android": "mirror",
-                    "deno": {
-                      "version_added": false
-                    },
-                    "edge": "mirror",
-                    "firefox": {
-                      "version_added": "preview"
-                    },
-                    "firefox_android": "mirror",
-                    "ie": {
-                      "version_added": false
-                    },
-                    "nodejs": {
-                      "version_added": false
-                    },
-                    "opera": "mirror",
-                    "opera_android": "mirror",
-                    "safari": {
-                      "version_added": false
-                    },
-                    "safari_ios": "mirror",
-                    "samsunginternet_android": "mirror",
-                    "webview_android": "mirror"
-                  },
-                  "status": {
-                    "experimental": true,
-                    "standard_track": true,
-                    "deprecated": false
-                  }
-                }
               }
             }
           },
--- a/javascript/builtins/intl/NumberFormat.json
+++ b/javascript/builtins/intl/NumberFormat.json
@@ -653,6 +653,43 @@
                 "standard_track": true,
                 "deprecated": false
               }
+            },
+            "number_param_string_is_decimal": {
+              "__compat": {
+                "description": "<code>number</code> param string value is decimal (not <code>Number</code>)",
+                "support": {
+                  "chrome": {
+                    "version_added": false
+                  },
+                  "chrome_android": "mirror",
+                  "deno": "mirror",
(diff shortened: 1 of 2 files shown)


hamishwillee:
@queengooborg Does adding support for preview count as a supported platform? I ask because if this feature is supported in Chrome "real versions" and FF preview it is still really only widely available in one platform.  I ask because it wants me to change this to not experimental when I consider it still to be.

javascript/builtins/intl/NumberFormat.json
```diff
@@ -571,7 +571,7 @@
             },
             "options_useGrouping_parameter": {
               "__compat": {
-                "description": "<code>options.useGrouping</code> parameter",
+                "description": "<code>options.useGrouping</code> parameter accepts: <code>'always'</code>, <code>'auto'</code>, <code>'min2'</code> (as well as <code>true</code> and <code>false</code>)",
```
hamishwillee, line 574: @queengooborg I need advice on this:

`options_useGrouping_parameter`
-  has existed since the constructor was created but previously could only be values `true` and `false`. 
- Having no feature for this previously was "as expected" because the compatibility only shows things that are non-compatible - and if it is present since day 1.
- But now the [NumberFormat v3](https://github.com/tc39/proposal-intl-numberformat-v3#grouping-enum-ecma-402-367) changes added the options: always, auto and min2. These are only implemented in FF nightly.

The problem with this is/was that with this name it looks like the whole parameter option is not supported in Chrome (say) because you see this subfeature and it appears not to be supported for Chrome. Further it looks like the option can not be used in FF prior to preview. While in fact the values of true and false are supported from a long time ago.

What I have done is added a note about what options are supported by this subfeature. I am not sure this is sufficient because notes are not visible by default. So I might reasonably 
- split this into two features, one just following the constructor, the other being this one.
- something else?

What do you think?

javascript/builtins/intl/NumberFormat.json
```diff
@@ -709,7 +748,7 @@
                 },
                 "edge": "mirror",
                 "firefox": {
-                  "version_added": false
+                  "version_added": "preview"
```
hamishwillee, line 751: @queengooborg . I'm getting errors `Error → Experimental should be set to false for  as the feature is supported in multiple browser engines.`

This is because adding this is seen as adding support for a new browser. But in fact it is kind of "partial support" - it could take ages to really be available. 

Is this correct /should I make these no longer experimental?

javascript/builtins/intl/NumberFormat.json
```diff
@@ -820,6 +859,44 @@
                 "experimental": false,
                 "standard_track": true,
                 "deprecated": false
+              },
+              "result_useGrouping_property": {
```
hamishwillee, line 863: FYI this is a subfeature indicating this property of the reutrned object might have different values than those that were supported in earlier versions. https://github.com/tc39/proposal-intl-numberformat-v3#grouping-enum-ecma-402-367

javascript/builtins/intl/NumberFormat.json
```diff
@@ -820,6 +859,44 @@
...
+                "__compat": {
+                  "description": "The result <code>useGrouping</code> parameter may contain <code>'always'</code>, <code>'auto'</code>, <code>'min2'</code>, <code>true</code> and <code>false</code> (previous versions may only contain <code>true</code> and <code>false</code>)",
```
hamishwillee, line 865:
```suggestion
                  "description": "The result property <code>useGrouping</code> may contain <code>'always'</code>, <code>'auto'</code>, <code>'min2'</code>, <code>true</code> and <code>false</code> (previous versions may only contain <code>true</code> and <code>false</code>)",
```

javascript/builtins/intl/NumberFormat.json
queengooborg: Oops, this will be fixed in #17036!

hamishwillee: Cool. Rebased.

javascript/builtins/intl/NumberFormat.json
```diff
@@ -859,44 +859,6 @@
                 "experimental": false,
                 "standard_track": true,
                 "deprecated": false
-              },
```
hamishwillee: @queengooborg I removed this as there was an error 
- `Error → ADDTIONAL PROPERTY must NOT have additional properties"`
- `result_useGrouping_property is not expected to be here!`
- ` Error → Property sorting error on undefined`
But I still want to have something here:

Essentially three is a useGrouping property that might be specified in the object returned by resolvedOptions. This has changed. This feature was supposed to indicate how.

What do I need to do?

hamishwillee:
@queengooborg Thanks. More advice required on this please.

javascript/builtins/intl/NumberFormat.json
queengooborg: I don't see a removal in the changes so I'm assuming this was reverted?

javascript/builtins/intl/NumberFormat.json
queengooborg: I think that if we should create a subfeature for the specific values (that is, `options_useGrouping_parameter.string_values`), so that the support for the parameter as a whole can be represented separately from the new values.  How does that sound?

javascript/builtins/intl/NumberFormat.json
```diff
@@ -653,6 +653,45 @@
                 "standard_track": true,
                 "deprecated": false
               }
+            },
+            "number_param_string_is_decimal": {
```
hamishwillee, line 657:
```suggestion
            "options_useGrouping_parameter.string_values": {
```

javascript/builtins/intl/NumberFormat.json
```diff
@@ -571,40 +571,88 @@
...
                   "webview_android": "mirror"
                 },
                 "status": {
-                  "experimental": true,
+                  "experimental": false,
                   "standard_track": true,
                   "deprecated": false
                 }
+              },
+              "options_useGrouping_parameter-string_values": {
```
hamishwillee, line 619: @queengooborg SOunds good. Here is the new parameter. Note that the "parent" subfeature for the parameter just copies the constructor versioning. This is certainly true for chrome (i.e. I have tested it exists in version 24) but I have not tested all the other versions.

javascript/builtins/intl/NumberFormat.json
hamishwillee: Yes. Done. Though I called it `options_useGrouping_parameter-string_values` since a period is not allowed as a separator.

javascript/builtins/intl/NumberFormat.json
```diff
@@ -821,6 +908,44 @@
                 "standard_track": true,
                 "deprecated": false
               }
+            },
+            "result_useGrouping_property": {
```
hamishwillee, line 912: Added this back in. Works now.

javascript/builtins/intl/NumberFormat.json
hamishwillee: Fixed and added back.

javascript/builtins/intl/NumberFormat.json
```diff
@@ -571,40 +571,88 @@
             },
             "options_useGrouping_parameter": {
               "__compat": {
-                "description": "<code>options.useGrouping</code> parameter",
```
queengooborg, line 574: This should be added back!

javascript/builtins/intl/NumberFormat.json
queengooborg, line 619: Since this is a subfeature of `options_useGrouping_parameter`, we can simplify the name:

```suggestion
              "string_values": {
```

queengooborg (changes requested):

javascript/builtins/intl/NumberFormat.json
hamishwillee: Done. Sorry - no idea why I would have removed that.

queengooborg (approved):
LGTM, thanks!

PR #16925, opened 2022-07-12
Remove loadstart/loadeddata/loadedmetadata events from GlobalEventHandlers

foolip:
These are media element events already represented on HTMLMediaElement.

Add spec URLs for all media element events while in the area.


queengooborg:
Could we also take care of `loadend` in this PR?

foolip:
@queengooborg no, that's not a media element event, so I left it out.

foolip:
@queengooborg can you make the content updates for this?

PR #16929, opened 2022-07-12
Update Firefox versions for api.FetchEvent.preloadResponse

queengooborg:
This PR updates and corrects the real values for Firefox and Firefox Android for the `preloadResponse` member of the `FetchEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.0.8).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/FetchEvent/preloadResponse

### DIFF
--- a/api/FetchEvent.json
+++ b/api/FetchEvent.json
@@ -208,7 +208,7 @@
               "version_added": "18"
             },
             "firefox": {
-              "version_added": false
+              "version_added": "99"
             },
             "firefox_android": "mirror",
             "ie": {


foolip (commented):
https://bugzilla.mozilla.org/show_bug.cgi?id=1577346 suggests this was in Firefox 97, can you check which is right?

queengooborg:
It looks like it was originally added in Firefox 97 behind a flag, "dom.serviceWorkers.navigationPreload.enabled", as suggested by the linked commit (https://hg.mozilla.org/integration/autoland/rev/4717024d2987), so my guess is that the flag was enabled by default in Firefox 99.

foolip (approved):

PR #16939, opened 2022-07-12
Update WebView versions for api.NetworkInformation.typechange_event

queengooborg:
This PR updates and corrects the real values for WebView Android for the `typechange_event` member of the `NetworkInformation` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.0.8).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/NetworkInformation/typechange_event

### DIFF
--- a/api/NetworkInformation.json
+++ b/api/NetworkInformation.json
@@ -437,7 +437,7 @@
             "safari_ios": "mirror",
             "samsunginternet_android": "mirror",
             "webview_android": {
-              "version_added": false
+              "version_added": "50"
             }
           },
           "status": {


queengooborg:
The event handler is supported in the latest version of WebView, and then the version number comes from mirroring + fixing consistency (since the API itself is supported since WebView 50)!

foolip:
I see, thanks! It would be great to have two commits in cases like this, one with the change made by script, and another with the manual fixes. Otherwise it looks like the script can do more than it actually can.

PR #16961, opened 2022-07-13
Fix Safari support for sizes/srcset attributes/properties

foolip:
The `<picture>` element was in WebKit trunk version 602.1.14:
https://github.com/WebKit/WebKit/commit/ee1f2663e07eb21f6ea8e0b0d01fce127d66e9a7
https://github.com/WebKit/WebKit/blob/ee1f2663e07eb21f6ea8e0b0d01fce127d66e9a7/Source/WebCore/Configurations/Version.xcconfig

The reflected properties were in WebKit trunk version 603.1.6:
https://github.com/WebKit/WebKit/commit/96c370b3601a5427ba1a4dd2ef78657d59729721
https://github.com/WebKit/WebKit/blob/96c370b3601a5427ba1a4dd2ef78657d59729721/Source/WebCore/Configurations/Version.xcconfig

Add notes to reflect the discrepancy, in case we try to merge attributes
and reflected properties in the future.

### DIFF
--- a/api/HTMLSourceElement.json
+++ b/api/HTMLSourceElement.json
@@ -141,7 +141,8 @@
             "opera": "mirror",
             "opera_android": "mirror",
             "safari": {
-              "version_added": "10.1"
+              "version_added": "10.1",
+              "notes": "The <a href='https://developer.mozilla.org/docs/Web/HTML/Element/source#attr-sizes'><code>sizes</code> attribute is supported since Safari 9.1."
             },
             "safari_ios": "mirror",
             "samsunginternet_android": "mirror",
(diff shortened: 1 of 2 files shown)


api/HTMLSourceElement.json
```diff
@@ -141,7 +141,8 @@
             "opera": "mirror",
             "opera_android": "mirror",
             "safari": {
-              "version_added": "10.1"
+              "version_added": "10.1",
+              "notes": "The <a href='https://developer.mozilla.org/docs/Web/HTML/Element/source#attr-sizes'><code>sizes</code> attribute is supported since Safari 9.1."
```
queengooborg, line 145:
```suggestion
              "notes": "The <a href='https://developer.mozilla.org/docs/Web/HTML/Element/source#attr-sizes'><code>sizes</code></a> attribute is supported since Safari 9.1."
```

queengooborg (changes requested):
Got some problems with the notes, but otherwise looking good!

queengooborg (approved):

PR #16962, opened 2022-07-13
Remove api.SVGImageElement.crossOrigin from BCD

queengooborg:
This PR removes the irrelevant `crossOrigin` member of the `SVGImageElement` API as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-features). The lack of current support has been confirmed by the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.0.8), even if the current BCD suggests support.

### DIFF
--- a/api/SVGImageElement.json
+++ b/api/SVGImageElement.json
@@ -43,40 +43,6 @@
           "deprecated": false
         }
       },
-      "crossOrigin": {
-        "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLImageElement/crossOrigin",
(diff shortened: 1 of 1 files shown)


foolip:
I think this is a tricky case, because there could be some value in documenting that this doesn't work for `SVGImageElement` specifically, even though it does work for other interfaces. https://bugs.chromium.org/p/chromium/issues/detail?id=842321 is the Chromium bug for this. How about linking that (and other bugs) in notes instead?

PR #16970, opened 2022-07-13
Improve structure of CacheStorage entries for Chromium

foolip:
This follows the structure of Request and allow more of the data to be
mirrored.


queengooborg (approved):
Version numbers in notes are actually bumped as well, which is what I'm assuming you're referring to -- but separating this into `partial_implementation` feels much more appropriate for this!

foolip:
> Version numbers in notes are actually bumped as well, which is what I'm assuming you're referring to -- but separating this into `partial_implementation` feels much more appropriate for this!

No, I'm actually referring to dropping the statement with the note entirely for browsers that forked after that point. That's not possible with a single entry with a note.

PR #16594, opened 2022-06-08
Manually map Safari versions in getMatchingBrowserVersion

foolip:
This is not ideal, but it fixes the problem now. Tests are in place to
ensure that if we fix it in some other way, there's no observable
difference.

PR #16994, opened 2022-07-14
Add upstream_version browser key; move Safari iOS mapping to browser JSON

queengooborg:
This PR is a follow-up to #16594 that adds an `upstream_version` property to the browser JSON and moves the manual mapping of Safari to Safari iOS into the browser JSON.


foolip:
I don't know why CI isn't failing, `npm run build` is broken for me locally. Is that not run in CI?

The problem:

```
scripts/release/mirror.ts:55:17 - error TS2339: Property 'upstream_version' does not exist on type 'ReleaseStatement'.

55     if (release.upstream_version == sourceVersion.replace('≤', '')) {
                   ~~~~~~~~~~~~~~~~
```

queengooborg:
Have you re-run `npm run gentypes`?

foolip:
I've never run `npm run gentypes`, but it looks like it's done when running `npm install`, which makes sense.

With that I can confirm that this doesn't change any of the compat data, so no mishaps.

foolip:
So with this approach, `upstream_version` is only used where the engine version would lead to the wrong mapping. But it makes it hard to see what the complete mapping is. The test does this, but won't be updated for future releases. WDYT about someone writing the complete mapping for all browsers to a file, and having the tests confirm all of those, and assert that the mapping is complete?

queengooborg:
I'm planning to add more `upstream_version` entries, not just to Safari iOS but to other browsers as well (and effectively replace the engine version for Blink/Gecko since they just match the browser's version), in a future PR!

foolip:
> replace the engine version for Blink/Gecko

Can you expand on this a bit? Do you intend to remove `engine` and `engine_version` fields from most of browsers/*.json?

queengooborg:
> Can you expand on this a bit? Do you intend to remove `engine` and `engine_version` fields from most of browsers/*.json?

Just the `engine_version`, since Chromium and Firefox don't have engine versions that differ from the browser's version number!

foolip:
> > Can you expand on this a bit? Do you intend to remove `engine` and `engine_version` fields from most of browsers/*.json?
> 
> Just the `engine_version`, since Chromium and Firefox don't have engine versions that differ from the browser's version number!

For pre-Blink WebKit, would you keep WebKit? I'm not sure the it's worth optimizing away redundant engine versions at the expense of consistently having one, but we can look at that in a future PR.

foolip:
@queengooborg about the "Fix mirroring with ranges" change, that will result in "≤3.2" versions in the built data, right? Can you arrange for the built data to be linted if it isn't already, to avoid this?

queengooborg:
I'm going to close this PR because it has sat inactive for years, and there has not been any noticeable interest.

PR #16995, opened 2022-07-14
Add data for MathML global attributes

fred-wang:
#### Summary

This is adding data for MDN articles describing MathML global attributes.

This was done by checking the source code of browsers, bug entries as well as the commits.

Note: `npm test` fails with "version_added: 'true' is NOT a valid version number" but that seems to be what other files do to.

#### Related issues

https://github.com/mdn/content/pull/17812

### DIFF
--- a/mathml/global_attributes.json
+++ b/mathml/global_attributes.json
@@ -1,6 +1,53 @@
 {
   "mathml": {
     "global_attributes": {
+      "dir": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/MathML/Global_attributes/dir",
+          "spec_url": "https://w3c.github.io/mathml-core/#dfn-dir",
+          "support": {
+            "chrome": {
+              "version_added": "80",
+              "impl_url": "https://crrev.com/c/1908533",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "#enable-experimental-web-platform-features",
+                  "value_to_set": "Enabled"
+                }
+              ]
+            },
+            "chrome_android": "mirror",
+            "edge": "mirror",
+            "firefox": {
+              "version_added": "70",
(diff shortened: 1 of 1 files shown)


queengooborg:
Hey @fred-wang, thank you for your PR, and welcome to BCD!

To answer your comment about `true` being disallowed as a `version_added`, it is because a [long-term goal for BCD](https://github.com/mdn/browser-compat-data/issues/3555) is to remove all "non-real values" (that is, data without a specific version number), and `mathml` is one of the categories where this is enforced.

I'd be more than happy to help research version numbers for Firefox and Safari for you, if you'd like!

fred-wang:
Thanks for the information @queengooborg ; I should be able to figure out which version these attributes were implemented. I'll update the PR later to get rid of the failures with version_added.

fred-wang:
@queengooborg I've updated the PR but I'm not sure whether there is an easy way to get Safari's version from the WebKit commit.

fred-wang:
@queengooborg Great, thanks I think I've updated the data and tests pass locally now.

Note that I haven't added data for MathML global attributes that are currently using the equivalent HTML pages. However, I've done that for the `dir` attribute because its syntax is slightly different from HTML's `dir` and I plan to create https://developer.mozilla.org/docs/Web/MathML/Global_attributes/dir

queengooborg (approved):
Thank you, this is LGTM!  I agree with not duplicating HTML global attributes here.  Welcome to BCD, @fred-wang!  🎉

fred-wang:
@queengooborg Thanks!

fred-wang:
@queengooborg extra question: when is that new data going to be available from MDN? even when testing MDN locally, data is not reachable.

queengooborg:
I release a new version of BCD every Tuesday, and it usually gets updated in Yari within the following two days.  These changes will go live by the end of the week, promise!

fred-wang:
Thanks for the information.

PR #16997, opened 2022-07-14
Remove unreachable code from diff script

foolip:
ts-node refuses to run this script with the line present:

> TSError: ⨯ Unable to compile TypeScript:
> scripts/diff.ts:86:3 - error TS7027: Unreachable code detected.


foolip:
@saschanaz FYI

PR #13940, opened 2021-12-10
An array of cookieStoreId supported in the tabs.query API

rebloor:
#### Summary
Added information about the support for an array of `cookieStoreId` in the `tabs.query` API.

#### Test results and supporting details
`npm test`  found no errors.

#### Related issues
Addresses documentation requirements of [Bug 1730931](https://bugzilla.mozilla.org/show_bug.cgi?id=1730931) Supporting an array of cookieStoreId in the tabs.query API.

Supporting MDN documentation updates made in An array of cookieStoreId supported in the tabs.query API [#11123 ](https://github.com/mdn/content/pull/11123)

PR #17006, opened 2022-07-14
Simplify multiple support statements with just notes

queengooborg:
This PR simplifies all of the BCD statements with multiple statements that only have notes (no `prefix`, `alternative_name`, `flags`, etc.).  Fixes the issues reported in #17005.


api/Clients.json
```diff
@@ -130,25 +130,17 @@
...
-                "notes": "<code>Client</code> objects returned in most recent focus order."
-              },
-              {
-                "version_added": "44"
-              }
-            ],
+            "firefox": {
+              "version_added": "44",
+              "notes": [
+                "Since Firefox 54, <code>Client</code> objects are returned in most recent focus order.",
```
foolip, line 136: Is this the per-spec behavior or not? Presumably there was a bug before 54, and we should have a note or `partial_implementation` for that?

api/Console.json
```diff
@@ -306,15 +306,10 @@
...
-                "version_added": "8.10.0",
-                "notes": "Alias for <code>console.log</code>"
-              },
-              {
-                "version_added": "8.0.0"
-              }
-            ],
+            "nodejs": {
+              "version_added": "8.0.0",
+              "notes": "Since NodeJS 8.10.0, this method is an alias for <code>console.log</code>."
```
foolip, line 311:
```suggestion
              "notes": "Since Node.js 8.10.0, this method is an alias for <code>console.log</code>."
```
```suggestion
              "notes": "Since NodeJS 8.10.0, this method is an alias for <code>console.log</code>."
```

api/FetchEvent.json
```diff
@@ -318,25 +318,17 @@
...
-                "notes": "NetworkError thrown if request mode is same-origin and response type is cors (see <a href='https://bugzil.la/1222008'>bug 1222008</a>)."
-              },
-              {
-                "version_added": "44"
-              }
-            ],
+            "firefox": {
+              "version_added": "44",
+              "notes": [
+                "Since Firefox 59, a <code>NetworkError</code> thrown if the request mode is <code>same-origin</code> and the response type is <code>cors</code> (see <a href='https://bugzil.la/1222008'>bug 1222008</a>).",
```
foolip, line 324: Is this per-spec or not? If it's correct to throw `NetworkError` and that now happens, then no note at all is needed, because the old behavior is a subset of the new correct behavior. Nobody needs to do anything special to support the older version in this case.

api/RTCRtpSender.json
```diff
@@ -291,15 +291,10 @@
...
-                "version_added": "64",
-                "notes": "Changes to parameters that should update live now do so starting in Firefox 64."
-              },
-              {
-                "version_added": "46"
-              }
-            ],
+            "firefox": {
+              "version_added": "46",
+              "notes": "Changes to parameters that should update live now do so starting in Firefox 64."
```
foolip, line 296: Should this actually be `partial_implementation` for 46-63?

api/_globals/atob.json
```diff
@@ -15,15 +15,10 @@
...
-              "version_added": "27",
-              "notes": "<code>atob()</code> ignores all space characters in the argument to comply with the latest HTML5 spec (see <a href='https://bugzil.la/711180'>bug 711180</a>)."
-            },
-            {
-              "version_added": "1"
-            }
-          ],
+          "firefox": {
+            "version_added": "1",
+            "notes": "Since Firefox 27, <code>atob()</code> ignores all space characters in the argument to comply with the latest HTML5 spec (see <a href='https://bugzil.la/711180'>bug 711180</a>)."
```
foolip, line 20: This again is describing the correct and expected behavior, and doesn't say what the problem was before this.

css/properties/border-bottom-left-radius.json
```diff
@@ -150,15 +150,10 @@
...
-                  "version_added": "4"
-                },
-                {
-                  "version_added": "1",
-                  "notes": "Before Firefox 4, the <a href='https://developer.mozilla.org/docs/Web/CSS/percentage'><code>&lt;percentage&gt;</code></a> was relative to the width of the box even when specifying the radius for a height. This implied that <code>-moz-border-radius-bottomleft</code> was always drawing an arc of circle, and never an ellipse, when followed by a single value."
-                }
-              ],
+              "firefox": {
+                "version_added": "1",
+                "notes": "Before Firefox 4, the <a href='https://developer.mozilla.org/docs/Web/CSS/percentage'><code>&lt;percentage&gt;</code></a> was relative to the width of the box even when specifying the radius for a height. This implied that <code>-moz-border-radius-bottomleft</code> was always drawing an arc of circle, and never an ellipse, when followed by a single value."
```
foolip, line 155: A `partial_implementation` range for this would be better I think.

webextensions/api/tabs.json
```diff
@@ -3163,15 +3163,10 @@
...
-                      "notes": "Supports an array of strings to return tabs associated with any of the list of <code>cookieStoreId</code>s.",
-                      "version_added": "97"
-                    },
-                    {
-                      "version_added": "52"
-                    }
-                  ],
+                  "firefox": {
+                    "version_added": "52",
+                    "notes": "Since Firefox 97, this property supports an array of strings to return tabs associated with any of the list of <code>cookieStoreId</code>s."
```
foolip, line 3168: partial_implementation for the range before?

api/FetchEvent.json
queengooborg: It looks like this is [per-spec behavior](urlhttps://github.com/whatwg/fetch/pull/655), but I wasn't sure since the note was added to Chromium browsers as well.  Maybe a subfeature would be appropriate here?

webextensions/api/tabs.json
queengooborg: It's hard to determine what qualifies as `partial_implementation` for web extensions, since there are no standards to refer to!  This was added in #13940.

api/FetchEvent.json
foolip: If you can identify some behavior that web developers need to understand today, then maybe a subfeature would be appropriate. A test showing what kind of code triggers this `NetworkError` might help illuminate.

webextensions/api/tabs.json
foolip: Maybe comparing to other browsers would help answer what's noteworthy and not? This note isn't too bad as it is (now with your edits) but one still wonders if this works in all browsers or not.

queengooborg: No other browsers support this feature it looks like, so we have nothing to compare it to.  However, after reading the documentation on MDN, I've updated this with a partial implementation and rewrote the notes to indicate what it does and does not support.

api/Clients.json
```diff
@@ -133,20 +133,27 @@
             "firefox": [
               {
                 "version_added": "54",
-                "notes": "<code>Client</code> objects returned in most recent focus order."
+                "notes": "<a href='https://www.mozilla.org/en-US/firefox/organizations/'>Extended Support Releases (ESR)</a> before Firefox 78 ESR do not support service workers and the Push API."
```
foolip, line 136: I don't love duplicating this note, but that's what we have to do when the version ranges for two notes overlap partially. I've sent https://github.com/mdn/browser-compat-data/pull/17060 to delete these notes in particular, which should then make this cleanup cleaner.

queengooborg:
Since all feedback has been addressed on this PR, I'm going to go ahead and merge it before any additional merge conflicts occur.  This should unblock the linter PR!

foolip:
@queengooborg I hadn't reviewed the whole PR, just left a few comments on the most obvious things I could spot. Can you please revert so all changes are reviewed? (By someone other than me is fine.)

queengooborg:
Argh, I already merged the linter updates…

foolip:
Maybe revert https://github.com/mdn/browser-compat-data/pull/17005 too for now? Skimming these changes there's plenty that needs investigation IMHO to ensure the resulting notes make sense, and some minor style things like tense and redundant naming of versions in ranged statements.

queengooborg:
Argh...  Alright, I'll revert both PRs.  I'll open two separate PRs to redo this one: one for what's already been reviewed so we can merge that immediately, and then another for the remaining stuff.  Hopefully we can get this and the linter re-merged quickly.

PR #17013, opened 2022-07-15
background-position-* 2-value syntax support in Safari 15.4

nt1m:
Safari 15.4 now supports the 2-value syntax for background-position-x/background-position-y

See: https://bugs.webkit.org/show_bug.cgi?id=202148 for testcase and bug fix.


foolip (approved):
Thanks @nt1m!

PR #17014, opened 2022-07-15
Unprefixed user-select not in Safari 16

nt1m:
This has been reverted in https://github.com/WebKit/WebKit/commit/e4b63beaeb743e2d311d689590ec49bb0728b3ff


foolip (approved):
That's a shame, but thanks for updating the data @nt1m!

PR #17021, opened 2022-07-16
Update Safari versions for Animation API

queengooborg:
This PR updates and corrects the real values for Safari (Desktop and iOS/iPadOS) for the `Animation` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.0.8).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Animation

Note: due to issues with `update-bcd`, these changes were performed manually by comparing the results output for Safari 10.0 and 10.1.

### DIFF
--- a/api/Animation.json
+++ b/api/Animation.json
@@ -21,7 +21,7 @@
           "opera": "mirror",
           "opera_android": "mirror",
           "safari": {
-            "version_added": "13.1"
+            "version_added": "10.1"
           },
           "safari_ios": "mirror",
           "samsunginternet_android": "mirror",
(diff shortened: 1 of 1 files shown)


foolip (commented):
This doesn't match the data for https://developer.mozilla.org/en-US/docs/Web/API/Element/animate. Maybe the interface was exposed before it could actually be used for anything?

queengooborg:
Ah, yep, that's what's going on.  Because two of the properties were supported in that early version of Safari as well, the collector was setting them to Safari 10.1.  Between 10.1 and 13.1, they all returned `null`, so the collector didn't know what to do.

PR #17043, opened 2022-07-19
Update mstyle compatibility data

fred-wang:
#### Summary
* Update data for Chrome, based on info from mathml.global_attribute.

* Update notes for Firefox, taken from Gecko-specific notes of
  https://developer.mozilla.org/en-US/docs/Web/MathML/Element/mstyle

* Remove dir, displaystyle, scriptlevel since they now belong to
  mathml.global_attribute

* Update scriptminsize/scriptsizemultiplier, removed in Firefox 103
  and not part of MathML Core.

This was essentially copying info from https://github.com/mdn/browser-compat-data/pull/16995, https://bugzilla.mozilla.org/show_bug.cgi?id=1772697 and https://developer.mozilla.org/en-US/docs/Web/MathML/Element/mstyle

#### Related issues

This is needed for https://github.com/mdn/content/pull/18273

### DIFF
--- a/mathml/elements/mstyle.json
+++ b/mathml/elements/mstyle.json
@@ -7,13 +7,30 @@
           "spec_url": "https://w3c.github.io/mathml-core/#style-change-mstyle",
           "support": {
             "chrome": {
-              "version_added": false,
-              "notes": "See <a href='https://crbug.com/6606'>bug 6606</a>."
+              "version_added": "80",
+              "impl_url": "https://crrev.com/c/1908533",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "#enable-experimental-web-platform-features",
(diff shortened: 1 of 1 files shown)


fred-wang:
cc @queengooborg @teoli2003

fred-wang:
@queengooborg any update on the review here?

queengooborg (approved):
Eventually, we should probably get those attributes separated into subfeatures, but we can worry about that later.  For now, LGTM!

fred-wang:
> Eventually, we should probably get those attributes separated into subfeatures, but we can worry about that later. For now, LGTM!

I'm not sure I understand this? All attributes are already handled as subfeatures no? And all the ones removed here are in mathml.global_attribute?

Issue #17024, opened 2022-07-17
Remove _mixins folder and merge features?


queengooborg:
The `_mixins` folder under `api/` holds partial feature data for certain APIs, separated by their original mixin.  It was a great way to quickly demix numerous mixins that we had in BCD, however now it feels redundant.  We're not exposing mixins to readers, and for anyone who cares about whether a feature is implemented through a mixin or not, there are better tools available to determine that, such as Webref.  Additionally, tooling for BCD has to perform additional work to account for the data in `_mixins`.

I'm thinking we should simply remove the `_mixins` folder and merge the features into the files for the rest of the APIs.  This should reduce the amount of work tooling must do, and reduce contributor confusion.

foolip:
This seems fine to me.

queengooborg:
This was agreed upon during the last BCD meeting, so I've opened a PR to do this and will merge it shortly!

PR #17053, opened 2022-07-20
Combine ARIAMixin members into individual interfaces

queengooborg:
This PR combines the separated mixin files for `ARIAMixin` into the main interface files.  Part of work for #17024.


queengooborg:
This was approved during the BCD meeting call earlier today!

PR #17057, opened 2022-07-21
Remove data for MathML global attributes from MathML elements.

fred-wang:
#### Summary

The information for MathML global attributes has been added to a
separate file [1] so it does not need to be repeated for each
individual MathML element anymore. That also makes things
consistent with what HTML does. Note that the case of mstyle is
handled in [2].

N/A this is just removing duplicated data.

#### Related issues

[1] https://github.com/mdn/browser-compat-data/pull/16995
[2] https://github.com/mdn/browser-compat-data/pull/17043


fred-wang:
cc @queengooborg

fred-wang:
@queengooborg can you please review this?

queengooborg (approved):

PR #17058, opened 2022-07-21
mfrac@bevelled is removed from Firefox

fred-wang:
#### Summary

This commit marks mfrac@bevelled as deprecated since it is no longer
part of MathML Core. Also indicates that it was removed from
Firefox 71. It seems this was already done for MDN in the past.

https://bugzilla.mozilla.org/show_bug.cgi?id=1587572
https://w3c.github.io/mathml-core/#fractions-mfrac

#### Related issues

Fixes https://github.com/mdn/content/issues/15359 again.

### DIFF
--- a/mathml/elements/mfrac.json
+++ b/mathml/elements/mfrac.json
@@ -53,7 +53,8 @@
               "chrome_android": "mirror",
               "edge": "mirror",
               "firefox": {
-                "version_added": "1"
+                "version_added": "1",
+                "version_removed": "71"
               },
               "firefox_android": "mirror",
               "ie": {
@@ -72,7 +73,7 @@
             "status": {
               "experimental": false,
               "standard_track": true,
-              "deprecated": false
+              "deprecated": true
             }
           }
         },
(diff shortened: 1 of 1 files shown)


fred-wang:
cc @queengooborg

fred-wang:
Mmh actually checking https://github.com/mdn/content/issues/15359#issuecomment-1109089637 there is some doubt about this. So let's wait a bit before taking this.

mathml/elements/mfrac.json
```diff
@@ -72,7 +73,7 @@
             "status": {
               "experimental": false,
               "standard_track": true,
-              "deprecated": false
+              "deprecated": true
```
queengooborg, line 76: Looking at the thread on mdn/content, it seems that there is no formal decision to deprecate this yet, so we should probably keep this as `false`:

```suggestion
              "deprecated": false
```

queengooborg (changes requested):

mathml/elements/mfrac.json
fred-wang: Yes, let's just take the firefox change for now.

queengooborg (approved):

Issue #16890, opened 2022-07-10
mathml.elements.math - Supported in Chrome 106


FryingggPannn:
### What type of issue is this?

Incorrect support data (ex. Chrome says "86" but support was added in "40")

### What information was incorrect, unhelpful, or incomplete?

MathML support was added in Chromium 106, and Chrome 106 is based on this version. Other Chromium 106-based browsers may also have MathML support in the future. See https://caniuse.com/mathml

### What browsers does this problem apply to, if applicable?

Chromium (Chrome, Edge 79+, Opera, Samsung Internet)

### What did you expect to see?

MathML supported in Chrome 106

### Did you test this? If so, how?

I tested this on my MacBook Air, and also checked the support table on Can I Use.

### MDN URL

https://developer.mozilla.org/en-US/docs/Web/MathML

### MDN metadata

<details>
<summary>MDN page report details</summary>

* Query: `mathml.elements.math`
* Report started: 2022-07-10T01:15:49.193Z

</details>

fred-wang:
Note that MathML is still under a flag in Chrome, so we need to wait before saying it is supported. Pull request to update the status: https://github.com/mdn/browser-compat-data/pull/17059

fred-wang:
BTW, I'm not sure why CanIuse said 106 ; initially chromestatus.com was using that milestone but we delayed that a bit after discussion on https://groups.google.com/a/chromium.org/g/blink-dev/c/n4zf_3FWmAA ; browser-compat-data should still be updated when the flag is on by default, but that's not the case yet.

PR #17059, opened 2022-07-21
Update MathML implementation status for Blink-based browsers

fred-wang:
#### Summary

A new MathML implementation started in Blink in 2019 and very basic
support was implemented under the "experimental web platform features"
flag. This corresponds to Blink 80 [1]. As a consequence, update the
Blink-based browsers accordingly:

- Chrome switched to Blink in version 28 [2], before that there was a
  WebKit implementation that was implemented enabled only in Chrome 24.
  We keep that information and add link to why it was removed in
  Chrome 25. Also add the new implementation in version 80.

- Edge switched to Blink in version 79 [3], add the new
  implementation in version 80.

- Opera switched to Blink in version 15 [4], before that there was a
  Presto implementation. We keep that information and add the new
  implementation in version 67 (corresponding to Blink version 80).

- Opera Android does not support flags [5], so MathML cannot be
  enabled for now. Don't change anything for this entry.

[1] https://chromiumdash.appspot.com/commit/ee4ac810348a8dd42fb0f5bc7353c507e4ab17c6
    https://chromiumdash.appspot.com/commit/6a1f248788b300a9d2b96c977976e747c5ec4aaa
    https://chromiumdash.appspot.com/commit/e4f3a0881c7a63f5ed0ff9dd68ef04d84a535969
    https://chromiumdash.appspot.com/commit/63dd456d6e73f1d41aadf211aefa3456f5825f47
    https://chromiumdash.appspot.com/commit/9bbdc694b075ea78b913f89954571f55ad14f0d8
    https://chromiumdash.appspot.com/commit/8b8b60fb32ffcc41c3ae41224389822969e172f3
[2] Cf chrome.json
[3] Cf edge.json
[4] Cf opera.json
[5] npm test is reporting that.

#### Related issues

Fixes https://github.com/mdn/browser-compat-data/issues/16890

### DIFF
--- a/mathml/elements/math.json
+++ b/mathml/elements/math.json
@@ -10,15 +10,13 @@
               {
                 "version_added": "80",
                 "impl_url": "https://crbug.com/6606",
-                "partial_implementation": true,
                 "flags": [
                   {
                     "type": "preference",
                     "name": "#enable-experimental-web-platform-features",
                     "value_to_set": "Enabled"
                   }
-                ],
-                "notes": "This is a new Blink-based implementation started in 2019."
+                ]
               },
               {
                 "version_added": "24",
@@ -32,15 +30,13 @@
             "edge": {
               "version_added": "80",
               "impl_url": "https://crbug.com/6606",
-              "partial_implementation": true,
               "flags": [
                 {
                   "type": "preference",
                   "name": "#enable-experimental-web-platform-features",
                   "value_to_set": "Enabled"
                 }
-              ],
-              "notes": "This is a new Blink-based implementation started in 2019."
+              ]
             },
             "firefox": [
               {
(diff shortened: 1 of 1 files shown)


fred-wang:
cc @queengooborg

mathml/elements/math.json
```diff
@@ -6,15 +6,41 @@
...
+                "impl_url": "https://crbug.com/6606",
+                "partial_implementation": true,
+                "flags": [
+                  {
+                    "type": "preference",
+                    "name": "#enable-experimental-web-platform-features",
+                    "value_to_set": "Enabled"
+                  }
+                ],
+                "notes": "This is a new Blink-based implementation started in 2019."
```
queengooborg, line 21: We should remove this note; it doesn't provide any useful information.

queengooborg (approved):
LGTM, let's remove the "new Blink-based implementation" notes and then this is good to go!

queengooborg (changes requested):
(Oops, meant to "request changes" instead of "approve"!)

mathml/elements/math.json
fred-wang: Done (It seems the scheme requires it when `partial_implementation` is true, so I removed that key too)

queengooborg (approved):

PR #17062, opened 2022-07-21
Add Firefox implementation notes for menclose@notation

fred-wang:
#### Summary

Import Firefox implementation notes for menclose@notation from https://developer.mozilla.org/en-US/docs/Web/MathML/Element/menclose#gecko-specific_notes

This was taken from MDN's Gecko-specific notes, I haven't checked again whether that's accurate.

#### Related issues

https://github.com/mdn/content/pull/18589

### DIFF
--- a/mathml/elements/menclose.json
+++ b/mathml/elements/menclose.json
@@ -42,27 +42,9 @@
               },
               "chrome_android": "mirror",
               "edge": "mirror",
-              "firefox": [
-                {
-                  "version_added": "32",
-                  "notes": "Notation <code>phasorangle</code> was implemented."
-                },
-                {
-                  "version_added": "24",
-                  "partial_implementation": true,
-                  "notes": "Notation <code>updiagonalarrow</code> was implemented."
-                },
-                {
-                  "version_added": "2",
-                  "partial_implementation": true,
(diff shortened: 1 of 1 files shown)


mathml/elements/menclose.json
```diff
@@ -145,9 +145,27 @@
...
+                  "version_added": "2",
+                  "partial_implementation": true,
+                  "notes": "Notation <code>madruwb</code> was implemented."
+                },
+                {
+                  "version_added": "1",
+                  "partial_implementation": true,
+                  "notes": "Only notations from MathML2 were implemented."
+                }
+              ],
```
queengooborg, line 168: I think it would be better to place these into subfeatures, so that the compatibility statements don't get so long here?

fred-wang: OK, but I guess that means adding keys for each browser again? Also menclose is not in MathML Core and its future in browsers is uncertain, so not sure it's worth spending too much time on it... Moreover, these notes are for very old versions of Firefox. Maybe we can just keep version 1 (partial) and 32 (complete) instead?

fred-wang: Added subfeatures.

queengooborg (approved):

PR #17034, opened 2022-07-18
Update support for :has() in Chrome

MrBrain295:
#### Summary

Fix version.
#### Test results and supporting details

https://chromestatus.com/feature/5794378545102848
#### Related issues

This fixes: #17020

PR #17072, opened 2022-07-22
Link Android bug for navigator.mediaDevices.getDisplayMedia()

foolip:
The WebView data is mirrored. It ended up like this seemingly due to an
incorrect interpretation of a "This also applies to webview_android"
review comment for Chrome Android, which was resolved by copying the
Firefox Android data:
https://github.com/mdn/browser-compat-data/pull/3290

### DIFF
--- a/api/MediaDevices.json
+++ b/api/MediaDevices.json
@@ -146,7 +146,8 @@
               "version_added": "72"
             },
             "chrome_android": {
-              "version_added": false
+              "version_added": false,
+              "notes": "See <a href='https://crbug.com/487935'>bug 487935</a>."
             },
             "edge": [
               {
@@ -154,6 +155,7 @@
               },
               {
                 "version_added": "17",
+                "version_removed": "79",
                 "notes": "Available as a member of <code>Navigator</code> instead of <code>MediaDevices</code>."
               }
             ],
@@ -184,10 +186,7 @@
               "version_added": false
             },
             "samsunginternet_android": "mirror",
-            "webview_android": {
-              "version_added": false,
-              "notes": "API is available, but will always fail with <code>NotAllowedError</code>."
-            }
+            "webview_android": "mirror"
           },
           "status": {
             "experimental": false,


api/MediaDevices.json
```diff
@@ -146,14 +146,16 @@
               "version_added": "72"
             },
             "chrome_android": {
-              "version_added": false
+              "version_added": false,
+              "notes": "See <a href='https://crbug.com/487935'>bug 487935</a>."
```
jpmedley, line 150: This bug looks like it's finally moving forward. If so, I have the same objection I have to listing bugs generally. The time they're included after fixing tends to far exceed the time they're useful to readers.

jpmedley (changes requested):

api/MediaDevices.json
foolip: I remove the links when I'm the reviewer as in https://github.com/mdn/browser-compat-data/pull/17034 and also clean up stale links when I find them, as in https://github.com/mdn/browser-compat-data/pull/17061.

What lint or process would be required for you to be happy with bug links for cases like this one?

jpmedley: You're assuming there should be something here. I'm saying that the cost benefit analysis of a link's value doesn't favor including bug links of any kind.

foolip: Is there a guideline on this? No other reviewers have pushed back on links, and I've added a lot of them, pretty sure each of @queengooborg @Elchi3 @ddbeck @sideshowbarker have reviewed such changes from me.

The benefit is that developers can click through to the bug and star/subscribe. And the Chrome team uses stars as a prioritization signal. For BCD maintainers, it also makes it clear that it's not just false by oversight, and changes to it should be reviewed with some care.

queengooborg: We don't have a guideline regarding bug links, but typically we have encouraged adding bug URLs.  Along with the benefits that @foolip has described above, adding bug links also has the benefit of telling web developers that some consideration (at the very least) has been given to implement a feature.  In other words, adding a bug link changes a "not supported" field to "not supported, but may possibly be implemented in the future".  Additionally, if the bug is fixed, a developer can see that with the link and open a BCD issue saying "hey, this feature has now been implemented so the tables should be updated" -- without it, there are no references that a developer can refer to.

In regards to the delay between a bug getting fixed and BCD updating the version number, this is something @foolip and I have been working hard on with the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project.

queengooborg (approved):
Thanks for adding the bug link by the way, LGTM!

PR #17077, opened 2022-07-22
Mirror SpeechGrammar data for Chrome Android

foolip:
This API is related to SpeechRecognition's grammars property, which is a
SpeechGrammarList. The data for that is all mirrored, so it's
implausible that SpeechGrammar alone was held back until Chrome 64.

There seems to have been no discussion about the difference in the PR
that added the data:
https://github.com/mdn/browser-compat-data/pull/976

### DIFF
--- a/api/SpeechGrammar.json
+++ b/api/SpeechGrammar.json
@@ -9,10 +9,7 @@
             "version_added": "25",
             "prefix": "webkit"
           },
-          "chrome_android": {
-            "version_added": "64",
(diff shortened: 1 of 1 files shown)


queengooborg (approved):
The difference in data is from the wiki migration, and I'm not seeing anything on [Chrome Status](https://chromestatus.com/feature/4782875580825600) that suggests this was true, so LGTM!

PR #17091, opened 2022-07-23
Add a subfeature for api.FetchEvent.respondWith returning NetworkError

queengooborg:
This PR adds a subfeature to `api.FetchEvent.respondWith` for if a NetworkError is thrown if the request mode is `same-origin` and the response type is `cors`.  This moves the information from the original notes.  The version number for Chromium comes from the linked [ChromeStatus feature](https://www.chromestatus.com/feature/5694278818856960).


queengooborg:
I'm going to self-merge this because it's blocking #17089.

PR #17092, opened 2022-07-23
Simplify multiple support statements with just notes (redo reviewed parts)

queengooborg:
This PR is a cherry-pick of #17006 that only includes the already-reviewed changes from that PR (hence the quick merge).

### DIFF
--- a/api/Clients.json
+++ b/api/Clients.json
@@ -123,11 +123,13 @@
             },
             "firefox": [
               {
-                "version_added": "54",
-                "notes": "<code>Client</code> objects returned in most recent focus order."
+                "version_added": "54"
               },
               {
-                "version_added": "44"
+                "version_added": "44",
+                "version_removed": "54",
+                "partial_implementation": true,
+                "notes": "Before Firefox 54, <code>Client</code> objects were not returned in the most recent focus order as required by the specification."
               }
             ],
             "firefox_android": "mirror",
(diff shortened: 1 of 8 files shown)


api/Clients.json
```diff
@@ -123,11 +123,13 @@
...
-                "version_added": "54",
-                "notes": "<code>Client</code> objects returned in most recent focus order."
+                "version_added": "54"
               },
               {
-                "version_added": "44"
+                "version_added": "44",
+                "version_removed": "54",
+                "partial_implementation": true,
+                "notes": "Before Firefox 54, <code>Client</code> objects were not returned in the most recent focus order as required by the specification."
```
foolip, line 132: We shouldn't have "Before Firefox 54" here.

api/RTCRtpSender.json
```diff
@@ -293,11 +293,13 @@
...
-                "version_added": "64",
-                "notes": "Changes to parameters that should update live now do so starting in Firefox 64."
+                "version_added": "64"
               },
               {
-                "version_added": "46"
+                "version_added": "46",
+                "version_removed": "64",
+                "partial_implementation": true,
+                "notes": "Before Firefox 64, changes to parameters that should update live would not do so."
```
foolip, line 302: Should not have "Before Firefox 64"

api/Clients.json
queengooborg: We actually have a fair number of notes that prefix "Before [Browser] [Version]" where the version matches the `version_removed` of a partial implementation (ex. `HTMLOptionElement` and `XMLDocument`), and we don't have a guideline on this.  Maybe we should create one?

PR #17095, opened 2022-07-23
Simplify multiple support statements with just notes for http/

queengooborg:
This PR is a cherry-pick of #17006 for the unreviewed features in `http/`.


http/data-url.json
```diff
@@ -50,22 +50,19 @@
             "chrome_android": "mirror",
             "edge": {
               "version_added": "12",
-              "notes": "The maximum size supported is 4GB"
+              "notes": "Before Edge 79, the maximum size supported was 4GB."
```
foolip, line 53: @ddbeck recommends a different style, current tense and not past tense.

queengooborg: I've updated the notes according to the proposed style, however I'd like to point out it hasn't been made into a guideline.

I also happened to notice that these notes were duplicated on both the parent and a few children.  I've moved these notes to the parent feature and removed them from the children for better placement/consistency.

http/data-url.json
```diff
@@ -12,15 +12,18 @@
...
           "firefox": {
             "version_added": true
           },
           "firefox_android": "mirror",
           "ie": {
             "version_added": "8",
-            "notes": "The maximum size supported is 32kB"
+            "notes": [
+              "Since Internet Explorer 9, the maximum size supported is 4GB.",
+              "In Internet Explorer 8, the maximum size supported is 32kB."
```
foolip, line 25: Hmm, this makes me wonder if the point of these notes is really just to document the very limited size in IE 8?

The 4GB limit doesn't really seem worth documenting TBH. But if it is, we should test it to confirm, and probably some other browser also has such a limit without a note...

queengooborg: I don't really think it's significant enough to document, but I'd like to leave the notes for now, since determining whether they're removed feels out-of-scope for this PR (and I'd like to unblock the linter PR as soon as possible).

queengooborg:
I'm merging this since review comments have been addressed and additional changes would be better deferred to another PR.

PR #17103, opened 2022-07-24
Update Firefox versions for api.CompositionEvent.CompositionEvent

queengooborg:
This PR updates and corrects the real values for Firefox and Firefox Android for the `CompositionEvent` member of the `CompositionEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/CompositionEvent/CompositionEvent

### DIFF
--- a/api/CompositionEvent.json
+++ b/api/CompositionEvent.json
@@ -51,7 +51,7 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "52"
+              "version_added": "53"
             },
             "firefox_android": "mirror",
             "ie": {


foolip:
Can you check where the original data came from? With changes this small, we don't want to be flipping back and forth based on some detail which may already have been settled in a past review.

foolip:
What kind of collector bug do you think this might have been? Are the results suspect, or the update script at the time?

queengooborg:
It looks like this was a results problem -- querying the results for Firefox 52 at the time I opened the PR, it states that the constructor was supported with no message attached.  Considering that it was the only Windows 7-based result within that range of browser versions, I think it was a fluke, or a problem in harness.js that has since been resolved.  (I did test Firefox 52 on Windows 7 in every CT to double-check.)

foolip:
Thanks for confirming! I would have been more concerned if it was an updater bug.

foolip (approved):

PR #17105, opened 2022-07-24
backdrop-filter: Update FF compatibility to 103

reinhart1010:
As [Bug 1578503](https://bugzilla.mozilla.org/show_bug.cgi?id=1578503) has been closed as fixed in Firefox 103, which will be released on July 26, 2022, The compatibility table for this feature needs to be updated.


foolip (approved):

foolip:
Thanks @reinhart1010 for the fix!

PR #17107, opened 2022-07-24
Update Firefox versions for ContactAddress API

queengooborg:
This PR updates and corrects the real values for Firefox and Firefox Android for the `ContactAddress` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/ContactAddress

### DIFF
--- a/api/ContactAddress.json
+++ b/api/ContactAddress.json
@@ -12,7 +12,8 @@
           },
           "edge": "mirror",
           "firefox": {
-            "version_added": false
+            "version_added": "13",
+            "version_removed": "23"
           },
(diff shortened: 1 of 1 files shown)


foolip (commented):
I feel like we've looked at this before. Was this really the same API, or just one by the same name but different in every other way?

queengooborg:
Ah, yes it was: https://github.com/mdn/browser-compat-data/pull/8071#discussion_r556514771

PR #17109, opened 2022-07-24
Update Firefox notes for api.CountQueuingStrategy.highWaterMark

queengooborg:
This PR updates and corrects the real values for Firefox and Firefox Android for the `highWaterMark` member of the `CountQueuingStrategy` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/CountQueuingStrategy/highWaterMark

### DIFF
--- a/api/CountQueuingStrategy.json
+++ b/api/CountQueuingStrategy.json
@@ -93,11 +93,17 @@
             "edge": {
               "version_added": "16"
             },
-            "firefox": {
-              "version_added": "65",
-              "partial_implementation": true,
-              "notes": "The property is defined on the instance instead of the prototype object. See <a href='https://bugzil.la/1684316'>bug 1684316</a>."
-            },
+            "firefox": [
+              {
+                "version_added": "89"
+              },
+              {
+                "version_added": "65",
+                "version_removed": "89",
+                "partial_implementation": true,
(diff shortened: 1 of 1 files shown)


api/CountQueuingStrategy.json
```diff
@@ -93,11 +93,17 @@
...
-            },
+            "firefox": [
+              {
+                "version_added": "89"
+              },
+              {
+                "version_added": "65",
+                "version_removed": "89",
+                "partial_implementation": true,
+                "notes": "The property is defined on the instance instead of the prototype object. See <a href='https://bugzil.la/1684316'>bug 1684316</a>."
```
foolip, line 104: If this note is right we need a custom test that detects support on the instance. Did you verify the ranges for this in some way?

queengooborg: I've added a custom test to the collector which will help us test this using an instance!

Although, maybe we should just remove the notes?  I don't think it's that important to web developers to mention the property isn't on the prototype, especially for an interface that would solely be accessed through creating instances.

foolip: I agree! If the property can be used on an instance, that's all that matters in practice.

foolip (approved):
Changes look good, but can you update the PR title/description to reflect what it's doing now?

PR #17120, opened 2022-07-25
Remove all Microsoft proprietary features from BCD

queengooborg:
This PR removes all of the Microsoft proprietary events, properties, etc. from BCD.  They can be removed per our guidelines, as they have been removed from Edge over two years ago and have not been included in a browser version released within the last five years.


foolip (commented):
Are none of the CSS properties variants of standard properties so that we should add them with prefix or alt. name to those?

queengooborg:
No, the `-ms-` prefixed CSS properties that remain in BCD now are proprietary.  In regards to the Grid properties, Microsoft's implementation of CSS Grids is based upon an old version of the specification, and is mostly incompatible with the CSS grid layout we know today.  Rachel has written a [great article about this](https://rachelandrew.co.uk/archives/2016/11/26/should-i-try-to-use-the-ie-implementation-of-css-grid-layout/).

foolip (approved):

queengooborg:
Although these should also be removed from the content side, I'm going to go ahead and merge this PR a little early since there is still debate on whether to keep them documented or not -- but if they are kept in MDN content, it would strictly be for the purposes of updating old websites and should not be encouraged for new development.

Issue #17112, opened 2022-07-24
css.properties.offset-path - Supported in Safari 15.5


fdelaneau:
### What type of issue is this?

Incorrect support data (ex. Chrome says "86" but support was added in "40")

### What information was incorrect, unhelpful, or incomplete?

Safari 15.5 (17613.2.7.1.8) seems to fully support CSS motion path.

### What browsers does this problem apply to, if applicable?

Safari

### What did you expect to see?

offset-path should not work but does (yeah !)

### Did you test this? If so, how?

Yes, I compared this page on Chrome 103.0.5060.134 and Safari 15.5 (17613.2.7.1.8) on a Codepen that uses both offset-path and offset-rotate.

https://codepen.io/Chokcoco/pen/gOgqoem

### MDN URL

https://developer.mozilla.org/en-US/docs/Web/CSS/CSS_Motion_Path

### MDN metadata

<details>
<summary>MDN page report details</summary>

* Query: `css.properties.offset-path`
* Report started: 2022-07-24T08:34:07.739Z

</details>

PR #17129, opened 2022-07-26
Fix: #17112

MrBrain295:
#### Summary

Fix Safari version.
#### Related issues

This fixes: #17112

### DIFF
--- a/css/properties/offset-path.json
+++ b/css/properties/offset-path.json
@@ -44,7 +44,7 @@
               }
             ],
             "safari": {
-              "version_added": "15.5"
+              "version_added": "15.4"
             },
             "safari_ios": "mirror",
             "samsunginternet_android": [
--- a/css/properties/offset-path.json
+++ b/css/properties/offset-path.json
@@ -44,7 +44,7 @@
               }
             ],
             "safari": {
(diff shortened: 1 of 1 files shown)


css/properties/offset-path.json
```diff
@@ -44,7 +44,7 @@
               }
             ],
             "safari": {
-              "version_added": false
+              "version_added": "15.5"
```
queengooborg, line 47: Looks like this should actually be 15.4 based upon https://bugs.webkit.org/show_bug.cgi?id=231801 and https://github.com/WebKit/WebKit/commit/d3d6c2754702263730e05ec7627746d1f0b766f7 (see [Version.xcconfig](https://github.com/WebKit/WebKit/blob/d3d6c2754702263730e05ec7627746d1f0b766f7/Source/WebKit/Configurations/Version.xcconfig) which indicates this was added in WebKit 613.1.8, where Safari 15.4 uses WebKit 613.1.17).

```suggestion
              "version_added": "15.4"
```

queengooborg (changes requested):

queengooborg (approved):
Thanks, LGTM!

PR #17130, opened 2022-07-26
Add __compat entries for all flex context CSS properties (+ fix discrepancies)

### LINKED ISSUE #13804: Regularize feature namespaces for CSS, HTML, HTTP, and JavaScript

queengooborg:
This PR adds top-level `__compat` entries for all of the CSS properties that have flex and grid contexts, so that every feature in BCD has a compat statement.  This fixes #13804.

Additionally, this PR fixes a few small discrepancies between browsers where mirroring should be applied but isn't.

### DIFF
--- a/css/properties/align-content.json
+++ b/css/properties/align-content.json
@@ -2,43 +2,97 @@
   "css": {
     "properties": {
       "align-content": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/align-content",
+          "spec_url": [
+            "https://drafts.csswg.org/css-align/#align-justify-content",
+            "https://drafts.csswg.org/css-flexbox/#align-content-property"
+          ],
+          "support": {
+            "chrome": [
+              {
+                "version_added": "29"
+              },
+              {
+                "prefix": "-webkit-",
+                "version_added": "21"
+              }
+            ],
+            "chrome_android": "mirror",
+            "edge": [
+              {
+                "version_added": "12"
+              },
+              {
+                "prefix": "-webkit-",
+                "version_added": "12"
+              }
+            ],
+            "firefox": [
+              {
+                "version_added": "28"
+              },
+              {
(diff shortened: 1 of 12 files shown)


queengooborg:
I'm self-merging this because other than Opera Android going from false to truthy, there are no version number changes introduced in this PR, and the changes are simple.  This is also blocking a longer update I'm looking to do to enforce file names.

Issue #13716, opened 2021-11-24
css.selectors.autofill - Chrome 96, Edge 96 not actually supporting non-vendor prefixed version


wojtekmaj:
#### What information was incorrect, unhelpful, or incomplete?

All Chromium-based browsers do NOT support `:autofill`, just `:-webkit-autofill`; even though it was [announced](https://blog.chromium.org/2021/10/chrome-96-beta-conditional-focus.html) on Chromium blog that support was added in Chrome 96 Beta. On [Chrome Platform Status](https://www.chromestatus.com/feature/5592445322526720) it still appears as "In development" and the practical tests I've done confirm that the latter information is not out of date.

#### What did you expect to see?

All Chromium-based browsers not to be marked as supporting `:autofill` without vendor prefix.

#### Did you test this? If so, how?

The simplest way is to execute `document.body.querySelector(':autofill');`.

Chrome 96:
![image](https://user-images.githubusercontent.com/5426427/143241670-e72b527c-2072-41bd-a4e1-d147ee70ccf2.png)

Edge 96:
![image](https://user-images.githubusercontent.com/5426427/143241744-4b177afc-9834-4118-9b93-9da2368d93af.png)

Firefox 95 (the only one listed here _actually_ supporting `:autofill` - but simply not matching it on `about:blank`):
![image](https://user-images.githubusercontent.com/5426427/143241880-0d1e982b-0a89-4416-ab09-d3f1e22b759b.png)

<details>
<summary>MDN page report details</summary>

* Query: `css.selectors.autofill`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/CSS/:autofill
* Report started: 2021-11-24T12:45:00.379Z

</details>

PR #17139, opened 2022-07-26
Fix `:autofill` status for Chrome and Chromium-based browsers

chevsky:
#### Summary
Some time ago `:autofill` pseudo-class was considered working in Chrome without prefix and updated in https://github.com/mdn/browser-compat-data/pull/12835. Despite the feature status page information, the pseudo-class is still behind feature-flag, so only prefixed version works.

- https://chromestatus.com/feature/5592445322526720 (`:autofill` status page)
- https://bugs.chromium.org/p/chromium/issues/detail?id=1164311#c8 (feature-flag discussion)
- https://ngc27m.csb.app (`:autofill` support demo)

#### Related issues
Fixes https://github.com/mdn/browser-compat-data/issues/13716

### DIFF
--- a/css/selectors/autofill.json
+++ b/css/selectors/autofill.json
@@ -7,15 +7,10 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/:autofill",
           "spec_url": "https://html.spec.whatwg.org/multipage/semantics-other.html#selector-autofill",
           "support": {
-            "chrome": [
-              {
-                "version_added": "96"
-              },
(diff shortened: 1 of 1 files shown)


queengooborg (approved):
Hey there, thanks for your PR!  This had surprised me, but I now see that ChromeStatus says that the shipping dates are "estimated milestones" (pretty easy to miss that little part).  With your test page and looking deeper into the source, this is LGTM!

Welcome to BCD!  🎉

PR #17167, opened 2022-08-01
Update Chrome/Opera ProgressEvent data

foolip:
The lengthComputable property was in the initial implementation of
ProgressEvent in WebKit:
https://github.com/WebKit/WebKit/commit/13e3659e55eeec5a96acf0f2e85298ed89ad1316

That was before Chrome 1.

Support for the constructor and lengthComputable was manually confirmed
by testing in Opera 12.16 on macOS.

### DIFF
--- a/api/ProgressEvent.json
+++ b/api/ProgressEvent.json
@@ -68,8 +68,12 @@
               "version_added": false
             },
             "oculus": "mirror",
-            "opera": "mirror",
-            "opera_android": "mirror",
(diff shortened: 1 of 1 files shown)


foolip:
Opera 12.16 support also confirmed with the custom tests added in https://github.com/foolip/mdn-bcd-collector/pull/2190

queengooborg (approved):

PR #17172, opened 2022-08-01
Remove MediaStream's stop() method

foolip:
This has been removed from all browsers for more than 2 years.


foolip:
There's nothing to update on MDN for this, I checked all occurrences of `MediaStream.*stop` and none of them are about this. Most are for the method on `MediaStreamTrack`.

queengooborg (approved):

PR #6829, opened 2020-10-05
Add/update WebKit + Blink data for Navigator API

vinyldarkscratch:
This PR updates the Chromium, Safari, and Opera data for the Navigator API using the mdn-bcd-collector project, aided with a little manual testing, to achieve more real data for the Navigator API.

Tests used: https://mdn-bcd-collector.appspot.com/tests/api/Navigator

PR #17173, opened 2022-08-01
Fix Safari version for navigator.requestMediaKeySystemAccess()

foolip:
Support in Safari 12.1 confirmed with this test:
https://mdn-bcd-collector.appspot.com/tests/api/Navigator/requestMediaKeySystemAccess

This also matches related APIs like MediaKeys.

The source of this data is also the collector tests:
https://github.com/mdn/browser-compat-data/pull/6829

It's not clear what went wrong then, some manual changes are possible.

### DIFF
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -3710,7 +3710,7 @@
               ]
             },
             "safari": {
-              "version_added": "13.1"
+              "version_added": "12.1"
             },
             "safari_ios": "mirror",
             "samsunginternet_android": "mirror",


foolip:
@queengooborg this data is from your PR, can you check what might have gone wrong and if it could affect other data?

A situation like this is why I suggest separating script-made and manual changes into separate commits, to make it easier to debug long after the fact.

queengooborg (approved):
I think that this was either an issue with the test code, or some HTTP vs. HTTPS issue that I didn't catch back in the day (since I didn't know to look out for accidental HTTP results).  Either way, current results show this is supported in Safari 12.1, so LGTM!

PR #17177, opened 2022-08-02
Update Safari versions for api.GestureEvent.GestureEvent

queengooborg:
This PR updates and corrects the real values for Safari macOS/iOS/iPadOS for the `GestureEvent` member of the `GestureEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/GestureEvent/GestureEvent

### DIFF
--- a/api/GestureEvent.json
+++ b/api/GestureEvent.json
@@ -58,7 +58,7 @@
               "version_added": "9"
             },
             "safari_ios": {
-              "version_added": "2"
+              "version_added": false
             },
             "samsunginternet_android": "mirror",
             "webview_android": "mirror"


foolip (commented):
What about Safari macOS support? Seems to not be supported there either, in which case the whole entry should be removed?

queengooborg:
I've confirmed a lack of support in Safari macOS as well, surprised the collector didn't update it -- I've opened a removal PR: https://github.com/mdn/browser-compat-data/pull/17215

PR #5254, opened 2019-12-03
Add data for OVR_multiview2.

jpmedley:
Addition to #5216. 

* The interface was [added in 70](https://storage.googleapis.com/chromium-find-releases-static/ad8.html#ad806a38015518e7361ab70adc4696b2e45c449d).
* It's current alternative name was [given in 75](https://storage.googleapis.com/chromium-find-releases-static/c58.html#c58c8fafa6c39a7cf5420849b6b58ca4c029272c).

PR #17200, opened 2022-08-02
Update Chromium versions for OVR_multiview2 API

queengooborg:
This PR updates and corrects the real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `OVR_multiview2` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/OVR_multiview2

Note: all of the original data is removed because this extension was not supported at all before Chrome 93, as tested in various versions on BrowserStack.  The alternative names only applied to the name of the interface, which are irrelevant in this case.

### DIFF
--- a/api/OVR_multiview2.json
+++ b/api/OVR_multiview2.json
@@ -5,17 +5,9 @@
         "mdn_url": "https://developer.mozilla.org/docs/Web/API/OVR_multiview2",
         "spec_url": "https://www.khronos.org/registry/webgl/extensions/OVR_multiview2/",
         "support": {
-          "chrome": [
-            {
-              "version_added": "75",
-              "alternative_name": "OVRMultiview2"
-            },
-            {
-              "version_added": "70",
-              "version_removed": "75",
-              "alternative_name": "WebGLMultiview"
-            }
-          ],
+          "chrome": {
+            "version_added": "93"
+          },
           "chrome_android": "mirror",
           "edge": "mirror",
           "firefox": {
@@ -26,9 +18,7 @@
             "version_added": false
           },
           "oculus": "mirror",
-          "opera": {
-            "version_added": false
-          },
+          "opera": "mirror",
           "opera_android": {
             "version_added": false
           },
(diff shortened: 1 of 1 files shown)


foolip:
https://mdn-bcd-collector.appspot.com/tests/api/OVR_multiview2 indicates no support on macOS for me.

Also, https://storage.googleapis.com/chromium-find-releases-static/6ae.html#6ae28fa19fbf35d6c929879a741ece420465ae6a suggests this was enabled in Chrome 79, much earlier.

Can you investigate more how support changed over time here?

queengooborg:
I also see no macOS support, so I've updated the data accordingly.

As for the difference between Chrome 79 and 93, I can't see anything in the [linked bug](https://bugs.chromium.org/p/chromium/issues/detail?id=949249) that suggests Chrome 93, however I've confirmed that there is a lack of support between Chrome 73 and 92 for this extension in BrowserStack, SauceLabs and LambdaTest, both from the collector test and manually running the following:

```js
var canvas1 = document.createElement('canvas');
var ctx1 = canvas1.getContext('webgl');
ctx1.getSupportedExtensions(); // <-- Looking for "multiview" in any of the results
var canvas2 = document.createElement('canvas');
var ctx2 = canvas2.getContext('webgl2');
ctx2.getSupportedExtensions(); // <-- Same search
```

foolip:
I think this is correct then, but https://storage.googleapis.com/chromium-find-releases-static/6ae.html#6ae28fa19fbf35d6c929879a741ece420465ae6a still leaves some room for doubt.

@kenrussell you were close to the implementation/shipping of this, do you know if it's correct that `OVR_multiview2` was first enabled on most platforms in Chrome 93, but still not on macOS? If so, what disabled it between Chrome 79 and 92, since it looks enabled in the code?

queengooborg:
My best guess is that it might've had something to do with hardware support or something along those lines.  Either way, I think we're good to merge for now and then we can make corrections if someone raises an issue later.  WDYT?

foolip:
Hmm, so the data came from https://github.com/mdn/browser-compat-data/pull/5254 based on source inspection, so let's go with test results instead. I wish I understood what explains the difference, but it's not a big deal if this data is a little bit wrong, so I'm not going to dig more.

foolip (approved):

PR #17212, opened 2022-08-02
scroll-timeline at-rule removed from spec; drop from BCD

sideshowbarker:
https://github.com/w3c/csswg-drafts/commit/fc6766d dropped the scroll-timeline at-rule from the “Scroll-linked Animations” spec


queengooborg:
This should also be set to `standard_track: false` accordingly, right?

foolip (commented):
Should `standard_track` also change to false, and deprecated to true? Or just delete this whole entry unless Firefox is likely to actually ship it?

sideshowbarker:
> Or just delete this whole entry unless Firefox is likely to actually ship it?

Let’s do that. I’ve pushed an update — and I opened https://github.com/mdn/content/pull/19056 and https://github.com/mdn/content/pull/19087 to drop the Specifications and Browser Compatibility sections from the MDN article, as well as marking the article with the Deprecated and Non-standard banners.

foolip (approved):

queengooborg:
Copying my comment from the content changes here:

> Since Firefox _does_ support the feature behind a flag, then I think that retaining the BCD data/tables would be better to show there is a browser with experimental, flagged implementation. I think we're more likely to catch the removal from the BCD side.

queengooborg:
Just saw the content PR was merged, so once this PR is rebased, I'll merge it!

PR #17214, opened 2022-08-02
Remove Chrome support for Sanitizer functions.

jpmedley:
#### Summary
Only part of the Sanitizer API is shipping in 105. The Chrome Status entry linked to below is less than clear. I've had several conversations with the feature developer. Also [a Chrome article](https://web.dev/sanitizer/) was recently updated to cover only the portions of the API that are actually shipping.

**This is why I always oppose putting anything on MDN before It reaches at least beta.**

#### Related issues
https://chromestatus.com/feature/5786893650231296

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -8509,14 +8509,7 @@
           "spec_url": "https://wicg.github.io/sanitizer-api/#dom-element-sethtml",
           "support": {
             "chrome": {
-              "version_added": "93",
-              "flags": [
-                {
-                  "type": "preference",
-                  "name": "enable-experimental-web-platform-features",
-                  "value_to_set": "enabled"
-                }
-              ]
+              "version_added": "105"
             },
             "chrome_android": "mirror",
             "edge": "mirror",
--- a/api/Sanitizer.json
+++ b/api/Sanitizer.json
@@ -133,7 +133,7 @@
           "support": {
             "chrome": [
               {
-                "version_added": "105"
+                "version_added": false
               },
               {
                 "version_added": "93",
@@ -189,7 +189,7 @@
           "support": {
             "chrome": [
               {
-                "version_added": "105"
+                "version_added": false
               },
               {
                 "version_added": "93",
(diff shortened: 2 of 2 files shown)


foolip:
This data came from https://github.com/mdn/browser-compat-data/pull/16446. Requesting review from @rachelandrew.

> This is why I always oppose putting anything on MDN before It reaches at least beta.

It would be technically straightforward to enforce this if we (the Chrome team) agreed on a policy we want and proposed it as a BCD guidelines.

Support is also tested by https://mdn-bcd-collector.appspot.com/tests/api/Sanitizer so this would also be addressed by the process described in https://github.com/openwebdocs/project/issues/89 which is a current BCD project, which @queengooborg will work on.

jpmedley:
>?This data came from https://github.com/mdn/browser-compat-data/pull/16446. Requesting review from @rachelandrew.

I'm aware of where it came from. I've been exchanging emails from the feature owner and with Chris Harrelson for the last two days. I'm nailing down the state of flags now. I'll correct the structure when I correct the flags.

rachelandrew:
Yeah this was a bit of a weird one, I didn't go back to change it because I wasn't sure what the resolution was but it looks like it got merged anyway. I should have probably marked it as a draft or something once I realised there was confusion. I must have thought it was in a beta initially for some reason.

jpmedley:
The original plan was to go all the way to beta with the whole thing.

jpmedley:
@foolip It looks like this [is governed by two flags](https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/sanitizer_api/README.md;drc=565fc592a7b5b795b4414e942910ddad4fc3a1d9;bpv=1;bpt=0) and has been since November. Do we have a way to represent that, or should I add a note?

queengooborg:
> Do we have a way to represent that, or should I add a note?

The flag property is an array of individual flag objects, so adding another object to the array to represent the second flag will do the trick!

jpmedley:
If the tests pass, this is finished.

jpmedley:
Chrome 105 is about to ship. Can someone please approve this?

rachelandrew (approved):

PR #17215, opened 2022-08-02
Remove api.GestureEvent.GestureEvent from BCD

queengooborg:
This PR removes the irrelevant `GestureEvent` member of the `GestureEvent` API as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-features). The lack of current support has been confirmed by the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.0), even if the current BCD suggests support.

### DIFF
--- a/api/GestureEvent.json
+++ b/api/GestureEvent.json
@@ -34,42 +34,6 @@
           "deprecated": false
         }
       },
-      "GestureEvent": {
-        "__compat": {
-          "description": "<code>GestureEvent()</code> constructor",
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/GestureEvent/GestureEvent",
-          "support": {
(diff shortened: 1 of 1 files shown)


foolip (approved):

queengooborg:
No content update is needed, actually; the page doesn't exist!

foolip:
https://developer.mozilla.org/en-US/docs/Web/API/GestureEvent#constructor does exist though, and needs removing.

queengooborg:
Ah okay, so there's content referencing the constructor, but nothing that is broken by the removal in BCD -- so the label doesn't apply, but still something to be done.  I'll open up a content PR for that!

PR #17226, opened 2022-08-03
Update WebKit support for innerHTML/outerHTML/innerText/outerText

foolip:
innerHTML and innerText was supported on HTMLElement from the beginning of WebKit:
https://github.com/WebKit/WebKit/blob/a063c2b75ddebd1edfe9c54c428d762bdb61f794/WebCore/khtml/ecma/kjs_html.cpp#L908-L911

outerHTML was implemented on 2004-05-05, after the Safari 1.2 release
but long before the Safari 1.3 release, so assume 1.3:
https://github.com/WebKit/WebKit/commit/09b1ec859efb384981e6f6f35e27363d83f78c38

outerText was implemented shortly after, also for 1.3:
https://github.com/WebKit/WebKit/commit/206f85c759065dc3df66c8818ef2eb3682ee7a22

Support for all 4 properties was manually confirmed in Safari 4 by
evaluating `document.body.innerHTML` and similar. This means any
remaining errors are before that time.

All of this was well before the initial releases of WebView Android and
Chrome, so set those versions to 1.

The notes about Node are wrong. innerHTML and outerHTML were originally
on HTMLElement, but that really doesn't matter, since even now when they
are on Element, they throw InvalidStateError for XML documents per spec:
https://w3c.github.io/DOM-Parsing/#ref-for-dom-innerhtml-innerhtml-2
https://w3c.github.io/DOM-Parsing/#ref-for-dom-element-outerhtml-2

### DIFF
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -27,12 +27,12 @@
             "version_added": "10.1"
           },
           "safari": {
-            "version_added": "1.3"
+            "version_added": "1"
           },
           "safari_ios": "mirror",
           "samsunginternet_android": "mirror",
           "webview_android": {
-            "version_added": "≤37"
+            "version_added": "1"
           }
         },
         "status": {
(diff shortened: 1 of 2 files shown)


api/HTMLElement.json
```diff
@@ -27,12 +27,12 @@
             "version_added": "10.1"
           },
           "safari": {
-            "version_added": "1.3"
+            "version_added": "1"
```
foolip, line 30: This change is needed to pass the lint, and is supported by the link given in the description:
https://github.com/WebKit/WebKit/blob/a063c2b75ddebd1edfe9c54c428d762bdb61f794/WebCore/khtml/ecma/kjs_html.cpp

queengooborg (approved):

PR #17213, opened 2022-08-02
Fix Firefox Android WebRTC data

foolip:
Clamp to 24 and mirror.

PR #4340, opened 2019-06-19
Add audio/video track support data to RTCRtpReceiver source get…

a2sheppy:
Firefox 68 adds support for video tracks to `RTCRtpReceiver`'s methods
`getContributingSources()` and `getSynchronousSources()`. This patch
adds all the data needed to present this.

Source:
* https://bugzilla.mozilla.org/show_bug.cgi?id=1534466

PR #17227, opened 2022-08-03
Mirror support for RTCRtpReceiver *_tracks_supported for Firefox Android

foolip:
After https://github.com/mdn/browser-compat-data/pull/17213 these are
the only remaining non-mirrored values in WebRTC APIs.

The data came from here:
https://github.com/mdn/browser-compat-data/pull/4340
https://bugzilla.mozilla.org/show_bug.cgi?id=1534466

Since there is no discussion about Android support at all, this seems
like a mistake.

### DIFF
--- a/api/RTCRtpReceiver.json
+++ b/api/RTCRtpReceiver.json
@@ -152,9 +152,7 @@
               "firefox": {
                 "version_added": "59"
               },
-              "firefox_android": {
-                "version_added": false
-              },
+              "firefox_android": "mirror",
               "ie": {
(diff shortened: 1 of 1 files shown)


foolip:
@Rumyra this is an educated guess on my part. Might you be able to find someone working on WebRTC in Gecko who could confirm this is supported on Android?

Rumyra:
I can confirm our assumptions are correct @foolip - so this is good to merge 👍

Rumyra (approved):

PR #17235, opened 2022-08-03
Mirror Android/Opera support for TypedArray copyWithin() and of()

foolip:
Chrome 45 was confirmed with these tests:
https://mdn-bcd-collector.appspot.com/tests/javascript/builtins/TypedArray/copyWithin
https://mdn-bcd-collector.appspot.com/tests/javascript/builtins/TypedArray/of

Then support in Opera 89 and Chrome Android 100 for the same were
confirmed in BrowserStack, and the data mirrored.

### DIFF
--- a/javascript/builtins/TypedArray.json
+++ b/javascript/builtins/TypedArray.json
@@ -349,9 +349,7 @@
               "chrome": {
                 "version_added": "45"
               },
-              "chrome_android": {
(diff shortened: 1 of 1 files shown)


queengooborg (approved):
Seems implausible there's differences in core JS features between OSes.  👍

PR #17244, opened 2022-08-04
Update Chromium versions for api.RTCRtpEncodingParameters.maxFramerate

bc-lee:
#### Summary

According to https://chromestatus.com/feature/6216116758118400 and
https://storage.googleapis.com/chromium-find-releases-static/9fe.html#9fe620ffcfaf81dc6ab3838a7d1ac5368fc1c53d,
this has been shipped since Chromium 81.

### DIFF
--- a/api/RTCRtpEncodingParameters.json
+++ b/api/RTCRtpEncodingParameters.json
@@ -210,7 +210,7 @@
           "spec_url": "https://w3c.github.io/webrtc-extensions/#dom-rtcrtpencodingparameters-maxframerate",
           "support": {
             "chrome": {
-              "version_added": "83"
+              "version_added": "81"
             },
             "chrome_android": "mirror",
             "edge": "mirror",
--- a/api/RTCRtpEncodingParameters.json
+++ b/api/RTCRtpEncodingParameters.json
@@ -210,7 +210,7 @@
           "spec_url": "https://w3c.github.io/webrtc-extensions/#dom-rtcrtpencodingparameters-maxframerate",
           "support": {
             "chrome": {
-              "version_added": false
+              "version_added": "83"
             },
             "chrome_android": "mirror",
             "edge": "mirror",


foolip:
https://chromestatus.com/feature/5703036164898816 and https://bugs.chromium.org/p/chromium/issues/detail?id=857048 don't mention `maxBitrate`, and the entry you're updating here is `maxFramerate`. Can you confirm if this was the intended change?

api/RTCRtpEncodingParameters.json
```diff
@@ -210,7 +210,7 @@
           "spec_url": "https://w3c.github.io/webrtc-extensions/#dom-rtcrtpencodingparameters-maxframerate",
           "support": {
             "chrome": {
-              "version_added": false
+              "version_added": "74"
```
foolip, line 213: So for the actual change here, it looks like Chrome 81 is the right version for `maxFramerate`:
https://storage.googleapis.com/chromium-find-releases-static/9fe.html#9fe620ffcfaf81dc6ab3838a7d1ac5368fc1c53d

```suggestion
              "version_added": "81"
```

foolip (changes requested):

api/RTCRtpEncodingParameters.json
bc-lee: I was in a rush for this PR, obviously a few things are wrong. I've updated the commit to reflect the changes.

api/RTCRtpEncodingParameters.json
```diff
@@ -210,7 +210,7 @@
           "spec_url": "https://w3c.github.io/webrtc-extensions/#dom-rtcrtpencodingparameters-maxframerate",
           "support": {
             "chrome": {
-              "version_added": false
+              "version_added": "83"
```
foolip, line 213: @bc-lee this now says 83 instead of 81. I see that https://chromestatus.com/feature/6216116758118400 says 83, but I don't think that's right.

```suggestion
              "version_added": "81"
```

foolip (changes requested):

foolip (approved):

bc-lee:
Is it possible to squash two commits on merge?

PR #17246, opened 2022-08-04
Remove incorrect notes on `MediaQueryList.addEventListener`

36degrees:
#### Summary

`MediaQueryList.addEventListener` was **not** based on EventTarget before Safari 14, otherwise it would implement the `addEventListener` and `removeEventListener` methods and so you would not need to use `addListener`.

It looks like [this might have been accidentally flipped in code review](https://github.com/mdn/browser-compat-data/pull/6403#discussion_r463062928) ('is not' => 'was not' => 'is').

N/A

#### Related issues

N/A

### DIFF
--- a/api/MediaQueryList.json
+++ b/api/MediaQueryList.json
@@ -27,12 +27,10 @@
             "version_added": "12.1"
           },
           "safari": {
-            "version_added": "5.1",
-            "notes": "Before Safari 14, <code>MediaQueryList</code> is based on <code>EventTarget</code>, so you must use <code>addListener()</code> and <code>removeListener()</code> to observe media query lists."
+            "version_added": "5.1"
           },
           "safari_ios": {
-            "version_added": "5",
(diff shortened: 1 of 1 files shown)


foolip (changes requested):
I think we should remove these notes instead. There's already a "EventTarget_inheritance" subfeature that captures this, and *all* browsers have gone through this change, not just Safari.

36degrees:
As requested, I've revised this PR to remove the incorrect notes instead, and updated the PR title.

I've force-pushed which means the original edit has been lost, but have tried to capture the history in the commit message instead.

36degrees:
Rebased again to bring up to date with changes in the `main` branch.

foolip (approved):

foolip:
Thanks @36degrees, and sorry this was left sitting for so long!

36degrees:
No worries! Thanks for approving and merging 🎉

Issue #17258, opened 2022-08-05
Make status required for webextensions?


foolip:
https://github.com/mdn/browser-compat-data/pull/5816 brought the schema and documentation out of sync with regards to the `status` object. The documentation still says it's mandatory:
https://github.com/mdn/browser-compat-data/blob/main/schemas/compat-data-schema.md#the-__compat-object

But in the schema, it's optional for webextensions. By making it required and counting errors, it looks like there are 1364 entries where it's not present, so about 10% of all of BCD.

The reason I noticed is because the TypeScript bindings treat it as optional, so all code has to pretend it can be missing, even if only dealing with API data.

Could we just add status blocks to all webextensions entries instead?

cc @wbamberg @Elchi3 @ddbeck from https://github.com/mdn/browser-compat-data/pull/5816 and @saschanaz from https://github.com/mdn/browser-compat-data/issues/6851. And @queengooborg.

queengooborg:
The reason that webextensions don't have a status block is because the status block would have the same `standard_track`, and there's effectively no real way to determine deprecation status, since web extensions are a non-standard feature to begin with.  While requiring status blocks for all features would certainly simplify some of the linter and typedef code, I feel that it would create more redundant work for contributors.

foolip:
Can we require status blocks to be absent where we know it doesn't make sense, rather than letting it be optional? Right now there's a mix in some files. Alternatively, we can lint that all of the values are false for those entries, to have some more consistency.

queengooborg:
Yeah that sounds good to me -- would you be down to open a PR to implement that?

PR #17259, opened 2022-08-05
Tighten up the types of walk utils to avoid redundant checks

foolip:
`walk()` can never yield an object without a `compat` property, and yet
the common WalkOutput type made it look like it at all call sites. Fix
this by specializing the types and simplifying the call sites.

### DIFF
--- a/scripts/migrations/007-experimental-false.ts
+++ b/scripts/migrations/007-experimental-false.ts
@@ -26,7 +26,7 @@ const dirname = fileURLToPath(new URL('.', import.meta.url));
  */
 export const fixExperimental = (bcd: CompatData | Identifier): void => {
   for (const { compat } of walk(undefined, bcd)) {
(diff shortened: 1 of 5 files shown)


scripts/migrations/007-experimental-false.ts
```diff
@@ -26,7 +26,7 @@ const dirname = fileURLToPath(new URL('.', import.meta.url));
  */
 export const fixExperimental = (bcd: CompatData | Identifier): void => {
   for (const { compat } of walk(undefined, bcd)) {
-    if (!compat?.status?.experimental) {
+    if (!compat.status?.experimental) {
```
foolip, line 29: See https://github.com/mdn/browser-compat-data/issues/17258 about the remaining question mark here.

lucalves (approved):

queengooborg (approved):

PR #17263, opened 2022-08-05
Mirror data where it was true and upstream had a real version

foolip:
This is relatively safe to do for browsers that only been based on one
browser engine, which excludes Edge and Opera. Also, anything with notes
was not mirrored, since that might require a closer review.

These changes were made with an ad-hoc script but would be best reviewed
by comparing the output.

### DIFF
--- a/html/elements/abbr.json
+++ b/html/elements/abbr.json
@@ -9,9 +9,7 @@
             "chrome": {
               "version_added": "2"
             },
-            "chrome_android": {
-              "version_added": true
-            },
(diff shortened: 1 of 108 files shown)


foolip:
Aha, so I didn't consider that this would end up changing the `opera_android` data when it was mirrored, from true to "14" a lot of the time. That might not be correct and should be reverted somehow.

foolip:
This approach doesn't make sense, probably better to just do this as part of a push to get real values for HTML etc.

PR #17264, opened 2022-08-05
Remove <audio> and <video> attributes that don't exist

foolip:
There are properties on the DOM APIs for most of these, but these aren't
attributes. While there's a muted attribute, there is no volume
attribute. The possible content attributes are listed here:
https://html.spec.whatwg.org/multipage/media.html#the-audio-element
https://html.spec.whatwg.org/multipage/media.html#the-video-element

### DIFF
--- a/html/elements/audio.json
+++ b/html/elements/audio.json
@@ -87,40 +87,6 @@
             }
           }
         },
-        "buffered": {
-          "__compat": {
-            "support": {
-              "chrome": {
-                "version_added": null
-              },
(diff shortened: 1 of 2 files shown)


foolip:
I've checked https://developer.mozilla.org/en-US/docs/Web/HTML/Element/audio and https://developer.mozilla.org/en-US/docs/Web/HTML/Element/video and don't see the same confusion there.

sideshowbarker (approved):

PR #17278, opened 2022-08-09
FF104 HTMLElement.focus() param focusVisible

hamishwillee:
FF104 adds support for [`HTMLElement.focus()`](https://developer.mozilla.org/en-US/docs/Web/API/HTMLElement/focus#parameters) param `focusVisible` in https://bugzilla.mozilla.org/show_bug.cgi?id=1765083. 

This is not supported in chromium or webkit yet, though there are bugs https://bugs.chromium.org/p/chromium/issues/detail?id=1317039 and https://bugs.webkit.org/show_bug.cgi?id=242456

Related docs work can be tracked in https://github.com/mdn/content/issues/19302

PR #17279, opened 2022-08-09
Replace backslashes with forward slashes for filepaths on Windows

queengooborg:
This should fix the issue reported by @hamishwillee in https://github.com/mdn/browser-compat-data/pull/17278#issuecomment-1208799737.

### DIFF
--- a/test/lint.ts
+++ b/test/lint.ts
@@ -54,9 +54,9 @@ const loadAndCheckFiles = async (...files: string[]): Promise<DataType> => {
         full: path.relative(process.cwd(), file),
         category: '',
       };
-      if (path.sep !== '/') {
+      if (path.sep === '\\') {
         // Normalize file paths for Windows users
-        filePath.full = filePath.full.replace(path.sep, '/');
+        filePath.full = filePath.full.replace(/\\/g, '/');
       }
       if (filePath.full.includes('/')) {
         filePath.category = filePath.full.split('/')[0];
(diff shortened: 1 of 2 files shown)


hamishwillee:
@queengooborg Exactly same behaviour  as before:  ran `npm install` then `npm test`

queengooborg:
Argh...  I don't have a way to try out things in Windows, but it's definitely something that's related to the filepaths.

test/lint.ts
```diff
@@ -54,8 +54,12 @@ const loadAndCheckFiles = async (...files: string[]): Promise<DataType> => {
         full: path.relative(process.cwd(), file),
         category: '',
       };
-      if (filePath.full.includes(path.sep)) {
-        filePath.category = filePath.full.split(path.sep)[0];
+      if (path.sep !== '/') {
+        // Normalize file paths for Windows users
+        filePath.full = filePath.full.replace(path.sep, '/');
```
hamishwillee, line 59: @queengooborg I added some crappy logging . It seems to be showing that this only replaces the first separator, and you have several.

```
Before replace: webextensions\manifest\storage.json 
After replace: webextensions/manifest\storage.json 
```

Could this be the cause? If not, add logging where you need and I'll run this for you.

queengooborg: Ack, right, I forgot that JavaScript's `.replace()` function is dumb... 🤦‍♀️

hamishwillee: Yeah, lets move to Python for all of this :-)

hamishwillee (approved):
Fix is good thanks. I can't merge though :-)

PR #17302, opened 2022-08-11
Update X-Frame-Options spec links

domenic:
Also mark ALLOW-FROM as deprecated and nonstandard.

https://wpt.fyi/results/x-frame-options?label=experimental&label=master&aligned


teoli2003 (approved):
Thank you!

PR #17316, opened 2022-08-12
css/selectors/backdrop/dialog/firefox corrections

queengooborg:
- Mark Firefox Android as partial implementation
- Add Firefox versions for css.selectors.backdrop.dialog


Elchi3 (approved):
Makes sense given https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/98#html

PR #17322, opened 2022-08-12
Mark navigator 'platform' and 'vendor' deprecated

OnkarRuikar:
### Summary
They are non-normative.They've been marked deprecated on MDN Web Docs in ~2016. Also, they are non-normative.
Browser implementers have been advised not to provide much info out of them:
>  Any information in this API that varies from user to user can be used to profile the user. In fact, if enough such information is available, a user can actually be uniquely identified. For this reason, user agent implementers are _**strongly urged to include as little information in this API as possible**_.

#### Vendor
- https://developer.mozilla.org/en-US/docs/Web/API/Navigator#navigator.vendor
- https://github.com/mdn/content/pull/12526
- https://html.spec.whatwg.org/multipage/system-state.html#dom-navigator-vendor-dev

#### Platform
- https://developer.mozilla.org/en-US/docs/Web/API/Navigator#navigator.platform
- https://html.spec.whatwg.org/multipage/system-state.html#dom-navigator-platform-dev

### DIFF
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -2174,7 +2174,7 @@
           "status": {
             "experimental": false,
             "standard_track": true,
-            "deprecated": false
+            "deprecated": true
           }
         }
       },
@@ -4462,7 +4462,7 @@
           "status": {
             "experimental": false,
             "standard_track": true,
-            "deprecated": false
+            "deprecated": true
           }
         }
       },


queengooborg (approved):

SyntaxColoring:
Over in https://github.com/mdn/content/issues/14429#issuecomment-3092444674, I mentioned that this change looked questionable. I'll clarify why here, to try to keep things organized.

> They are non-normative.

I'm not sure what was meant by this. The provided links (https://html.spec.whatwg.org/multipage/system-state.html#dom-navigator-vendor-dev, https://html.spec.whatwg.org/multipage/system-state.html#dom-navigator-platform-dev) do point inside a "For web developers (non-normative)" block. However, the surrounding text also describes them normatively, as far as I can tell. Maybe the spec was different when this PR was written.

> They've been marked deprecated on MDN Web Docs in ~2016

But un-deprecated in 2022, a few months before this PR, in https://github.com/mdn/content/pull/14452.

> Browser implementers have been advised not to provide much info out of them:
> 
> > Any information in this API that varies from user to user can be used to profile the user. In fact, if enough such information is available, a user can actually be uniquely identified. For this reason, user agent implementers are _**strongly urged to include as little information in this API as possible**_.

This doesn't seem relevant. This is advice to browser vendors to implement the feature with privacy in mind. It is not advice to web developers to avoid using the feature. (It's certainly good for web developers to be aware of the limits, though!)

PR #17326, opened 2022-08-12
FF103 supports the CSS contain:style option

hamishwillee:
FF103 adds support for the contain: style property in https://bugzilla.mozilla.org/show_bug.cgi?id=1463600.

Other docs work for this can be tracked in https://github.com/mdn/content/issues/19452

@queengooborg I updated this to make the previous support marked as "partial implementation" and add full support in 103. Seems to make more sense than having a subfeature just for the style option.

### DIFF
--- a/css/properties/contain.json
+++ b/css/properties/contain.json
@@ -11,10 +11,16 @@
             },
             "chrome_android": "mirror",
             "edge": "mirror",
-            "firefox": {
-              "version_added": "69",
-              "notes": "Firefox does not support the <code>style</code> value."
-            },
+            "firefox": [
+              {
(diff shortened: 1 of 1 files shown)


hamishwillee:
PS @queengooborg BCD is working locally on Windows since you added #17279 but I am still seeing lots of warnings. I assume I am intended to, though they are just white noise to me. They look like:
image

queengooborg:
They're warnings that I've eventually got to fix, sorry about that!  You can run `npm run lint` instead of `npm test` to specifically run the data linters instead of every check, which will keep those warnings from appearing!

queengooborg (approved):

PR #17352, opened 2022-08-15
Adding subfeatures guidance

hamishwillee:
The current guidance explains how you ad subfeatures, but omits how you decide when a subfeature is required.

My understanding is that you only add subfeatures for compatibility differences. So someone seeing that an interface is supported knows that all the browsers implement that subfeature in the same way in those particular versions. 

BUT they don't know exactly what methods, properties etc are exactly implemented from that data. While they might "hope" that everything in the spec is implemented, what they actually know is that they are all the same implementation in those indicated versions. 

I wanted to check this because the use case [`ImageData`] used widely in the doc does not comply with these rules - I'll comment below.

This comes out of the discussion here: https://github.com/mdn/browser-compat-data/issues/7849#issuecomment-1211067916

FYI @queengooborg @foolip


docs/data-guidelines.md
```diff
@@ -12,6 +12,16 @@ If it's helpful to understanding the rule, summarize the rationale. Definitely c
...
 
+
+## Adding subfeatures
+
+Subfeatures should only be added to indicate a  _difference_ in browser compatibility versions from the parent feature; they should not be added just to indicate that a particular feature exists.
+
+For example, the `ImageData` interface was initially defined with `data`, `height` and `width` properties.
+As these properties were defined in the same versions as the parent feature [`ImageData`](https://github.com/mdn/browser-compat-data/blob/main/api/ImageData.json), they do not require a subfeature entry.
+However the `colorSpace` property, the constructor, and support for using the API in workers were added in later versions.
+This indicates a compatibility break, which is why a subfeature entry is required.
```
hamishwillee, line 23:
```suggestion
For example, the `ImageData` interface was initially defined with `data`, `height` and `width` properties.
These do not require a subfeature entry because they were defined in the same versions as the parent feature [`ImageData`](https://github.com/mdn/browser-compat-data/blob/main/api/ImageData.json).

The subfeatures for the `colorSpace` property, the constructor, and support for using the API in workers, are required, because they appear in different versions than the original `ImageData` feature.
```

docs/data-guidelines.md
```diff
@@ -12,6 +12,16 @@ If it's helpful to understanding the rule, summarize the rationale. Definitely c
 
 -- END TEMPLATE -->
 
+
+## Adding subfeatures
+
+Subfeatures should only be added to indicate a  _difference_ in browser compatibility versions from the parent feature; they should not be added just to indicate that a particular feature exists.
+
+For example, the `ImageData` interface was initially defined with `data`, `height` and `width` properties.
```
hamishwillee, line 20: This is not actually true ImageData has these values defined in BCD. We should remove them right @queengooborg ?

FWIW I understand the reason for not documenting "to indicate support", but it does make it hard to know whether you  have just omitted something or there is no known compatibility break.

docs/data-guidelines.md
```diff
@@ -12,6 +12,16 @@ If it's helpful to understanding the rule, summarize the rationale. Definitely c
 
 -- END TEMPLATE -->
 
+
+## Adding subfeatures
+
+Subfeatures should only be added to indicate a  _difference_ in browser compatibility versions from the parent feature; they should not be added just to indicate that a particular feature exists.
+
```
hamishwillee, line 19:
```suggestion
Subfeatures should only be added to indicate a  _difference_ in browser compatibility versions from the parent feature; they should not be added just to indicate that a particular feature exists.

As a rule of thumb, if a subfeature would have the same versions as its parent feature, then it should not be added.

```

docs/data-guidelines.md
queengooborg: It's actually the opposite: all subfeatures should be added regardless of whether they match the parent version!  Each subfeature should be added to BCD, even if the data matches its parent, to help with clarity.  Say `ImageData` has recently added a new `format` subfeature -- consumers of BCD wouldn't be able to tell whether or not `format` has been supported if the assumption is that all undefined subfeatures match the parent.

Behavioral differences, on the other hand, are more subjective.  If the spec has changed the behavior of a feature and browsers needed to update, then we should add a subfeature for that behavior change.

There are a number of subfeatures we are missing from BCD at the moment, which I'm hoping to fix in the future!

hamishwillee: @queengooborg OK. That makes sense. It concerns me that this was not communicated to me. Or the new way of adding bug URLs. 

I'll try an update then.

hamishwillee: P.S. We could reduce duplication by allowing `"support": mirror` for the whole compat section of a subfeature implemented at the same versions as its parent.

docs/data-guidelines.md
```diff
@@ -12,6 +12,17 @@ If it's helpful to understanding the rule, summarize the rationale. Definitely c
 
 -- END TEMPLATE -->
 
+
+## Adding subfeatures
+
+Subfeatures should be added for all methods, properties, tags, attributes, method arguments, and other specification entities, provided that they have an implementation in one or more of the supported browser engines.
+
+In addition, subfeatures should be added to indicate support for specific behaviours, such as support in workers, or that and interface is serializable.
```
hamishwillee, line 20:
```suggestion
In addition, subfeatures should be added to indicate support for specific behaviours, such as support in workers, or that an interface is serializable.
```

docs/data-guidelines.md
```diff
@@ -12,6 +12,17 @@ If it's helpful to understanding the rule, summarize the rationale. Definitely c
 
 -- END TEMPLATE -->
 
+
+## Adding subfeatures
```
hamishwillee, line 16: @queengooborg @foolip OK, I have reworded this to be the opposite. This says "always add a subfeature for any entity or behaviour where there is an implementation in a browser."

docs/data-guidelines.md
```diff
@@ -12,6 +12,45 @@ If it's helpful to understanding the rule, summarize the rationale. Definitely c
...
+
+A feature may contain _subfeatures_, which document the browser versions in which parts of the parent feature were first supported or removed.
+For example, a method subfeature entry would indicate the browser versions in which the method was added to the parent interface.
+Subfeatures are also used to reflect changes in behaviour in particular versions, such as "support in workers" or "returns a promise".
+A subfeature might nest its own subfeatures, and so on. For example, an interface feature might nest a method subfeature, which nests an option parameter, which then nests a subfeature for a value of that option.
+
+Subfeatures entries are added for all:
+
+- interface methods and properties that have an implementation in any of the supported browsers.
+- ?
```
hamishwillee, line 30: Are there an other exceptions?

docs/data-guidelines.md
```diff
@@ -12,6 +12,45 @@ If it's helpful to understanding the rule, summarize the rationale. Definitely c
 
 -- END TEMPLATE -->
 
+
+## Adding features and subfeatures
```
hamishwillee, line 16: @queengooborg Updated as per our discussion:

- features required for all released web platform features. Term feature is now defined.
- subfeatures term now defined
- subfeature required for interface methods and properties by default if there is any released implementation.
- subfeatures not requried for anything else unless there is a change in version support. This has nothing to do with what the spec is doing and is entirely due to some implementation change.

I think this is at least close, though there might be other exceptions like interface methods and properties. See comment below.

There is also the slight problem of things like HTTP headers which are not necessarily implemented in a browser, and for which the term "supported" is very loose. I think we should get this in and then address that separately.

docs/data-guidelines/api.md
```diff
@@ -2,6 +2,49 @@
 
 This file contains guidelines that are specific to the web API features (`api/`).
 
+
+## Adding features and subfeatures
+
+A json file should be created for every _top level feature_ that has a released public implementation on one of the supported set of browsers (and not before).
```
foolip, line 8: The "and not before" here should come with a lint I think, and I think running that lint will reveal that more things are removed by this rule than perhaps we really want to see removed.

But if we can sort through that, I think that we should add things to BCD when they're in a beta browser release, to not have to wait until stable before documenting things, if someone is willing to do the work earlier.

docs/data-guidelines/api.md
```diff
@@ -2,6 +2,49 @@
...
+For Web APIs the top level features are individual Web API interfaces and the children of namespace objects like `Intl`.
+
+A possible structure for a Web API is shown below. 
+
+```
+api                           // category - always api for Web APIs
+  interface                   // top level feature
```
foolip, line 15: Also globals like `fetch()`.

docs/data-guidelines/api.md
```diff
@@ -2,6 +2,49 @@
...
+    methods                  // feature: required
+       parameters             //   subfeature: compatibility is different
+         parameter options    //   subfeature: if compatibility is different
+    support in workers     // subfeature: if compatibility is different (might also just be on specific methods)
+```
+
+The interface must include subnodes for _all_ its properties and methods (that have an implementation in at least one of the supported browsers), whether or not they were added at the same time as the parent interface, and hence share the same compatibility information.
+
+All other code and behavioral aspects of the interface are given nested nodes **only** if their compatibility is _different_ to that of their parent feature.
+This applies to method arguments, method argument options, and behaviour changes like the addition of support for workers, or the type of return value.
```
foolip, line 26:
```suggestion
This applies to method parameters, and behaviour changes like the addition of support for workers, or the type of return value.
```

docs/data-guidelines/api.md
```diff
@@ -2,6 +2,49 @@
...
+
+A possible structure for a Web API is shown below. 
+
+```
+api                           // category - always api for Web APIs
+  interface                   // top level feature
+    properties               // feature: required
+    methods                  // feature: required
+       parameters             //   subfeature: compatibility is different
+         parameter options    //   subfeature: if compatibility is different
```
foolip, line 19: What's this? We don't nest parameters, they have a structure like `options_capture_parameter`.

docs/data-guidelines/api.md
foolip, line 8: I don't think we should talk about the file structure here. That's actually very complicated, and better done as a lint I think. Plenty of little edge cases.

docs/data-guidelines/api.md
```diff
@@ -2,6 +2,49 @@
...
+    support in workers     // subfeature: if compatibility is different (might also just be on specific methods)
+```
+
+The interface must include subnodes for _all_ its properties and methods (that have an implementation in at least one of the supported browsers), whether or not they were added at the same time as the parent interface, and hence share the same compatibility information.
```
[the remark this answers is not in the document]
hamishwillee: @foolip Part of the reason for this update was that the term subfeatures was undefined. I thought it just meant a sub node of some other node, but @queengooborg was using it to mean a property or a method below the top level. 

Sub-node on the other hand is unambiguous. I think we should use that term if we mean "create another level of nesting" unless we absolutely define subfeature to mean "nested definition at any level".

docs/data-guidelines/api.md
hamishwillee: It also applies to parameter options.

docs/data-guidelines/api.md
```diff
@@ -2,6 +2,49 @@
...
+
+All other code and behavioral aspects of the interface are given nested nodes **only** if their compatibility is _different_ to that of their parent feature.
+This applies to method arguments, method argument options, and behaviour changes like the addition of support for workers, or the type of return value.
+
+> **Note:** Items that must be given nodes are referred to as "features", while those that are only created when there is a compatibility difference to the parent are called _subfeatures_.
```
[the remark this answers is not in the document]
hamishwillee: @foolip See https://github.com/mdn/browser-compat-data/pull/17352/files#r972516073

I would be happy with this if we used feature and subfeature sanely. But we don't. So define it properly - otherwise node and sub node are better.

If I sound frustrated, it is because I am. So much time wasted on this. IMO if we just said "feature" is the top level item and subfeature is any nested item at any nesting level this would be fine. But you'll have to get @queengooborg to agree - I couldn't

docs/data-guidelines/api.md
```diff
@@ -2,6 +2,49 @@
...
+
+> **Tip:** A good rule of thumb is that if a subfeature has the same compatibility information as its parent, then the subfeature is not needed.
```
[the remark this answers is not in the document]
hamishwillee: I don't think so. See the comment above:

> **Note:** Items that must be given nodes are referred to as "features", while those that are only created when there is a compatibility difference to the parent are called _subfeatures_.

So its a subfeature only if there is a compatibility difference.

docs/data-guidelines/api.md
```diff
@@ -2,6 +2,49 @@
...
+
+- An interface is defined in a specification with a method (feature) that has two parameters (subfeatures).
+- A supported browser implements the method and both the parameters:
+  - A feature node is created for the method (by default)
+  - Subfeature nodes are not created for the method parameters because they were created in the same version as the methods; there is no compatibility change.
+- Another browser implements the methods, but only one of the parameters.
+  - A subfeature node is added to the method parameter that was _not_ supported by the second browser.
+- Yet another browser implements the method with a new parameter. 
+  - A method subfeature is added for the new parameter because it was added in a different version than the parent method.
+- The specification changes to require that the interface now can be used in workers:
```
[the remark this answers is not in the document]
hamishwillee:
```suggestion
- The browser changes to require that the interface now can be used in workers:
```

docs/data-guidelines/api.md
hamishwillee: OK. Edit the page how you like.

hamishwillee:
@foolip Thanks for reviewing this.

I'm a bit frustrated with how this ended up - mostly because we had no definitions of our terminology, and now we have ones that I don't like. 

How about you and vinyl fight this out and just make the edits you need.

docs/data-guidelines/api.md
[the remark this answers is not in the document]
foolip: @queengooborg can you clarify the definitions you'd like to use?

What we have in BCD is a tree, where some nodes in the tree have `__compat`. These nodes are called "features" in `traverse.js`, and I've often called them entries as well.

I think there are two potentially useful concepts:
- Any feature that sits below another one in the tree, like "api.AbortController.abort" (because "api.AbortController" is a feature)
- Things that are behaviors of another feature and only worth adding if support differs, like "api.fetch.blob_data_support".

I don't know what the arguments are, but my initial preference would be to just call everything features, and talk about "behavioral features" or "subfeatures" where needed to be precise.

foolip:
@hamishwillee I hope we can land on some definitions that make sense to everyone.

@queengooborg It looks like we are invited to duel. Given your beat saber skills I think I'd be quickly shredded if we do it with swords, but let's see if we can work it out in comments instead 😄

hamishwillee:
@foolip Yes, you two must duel; let the gods of clarity prevail.

Apologies that I am taking my grumps out on you two. This seemed like this task should be so easy and it has not been.

I'm OK with everything being a feature, and using "behavioral features" to mean what it means. 

The problem is that for this to work we need to explain the conditions when you need to create a sub-node for a new feature -and when you do not. So we create a sub node by default for methods and properties of an interface but we do not for the arguments of a function or constructor.

I wanted to use the name "subfeature" to refer to any subnode of the method. If we can't do that, then find an acceptable substitute you'll be happy with and this should not end up being hard.

docs/data-guidelines/api.md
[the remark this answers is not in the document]
queengooborg: I'd like to define the terms as such:
- Feature: any object with a `__compat` statement (ex. api.Foo)
- Subfeature: any feature that is a level below another feature (ex. api.Foo.bar)
- Behavioral feature: any feature that describes a behavior, parameter, object, etc. (ex. api.Foo.worker_support, or api.Foo.bar.frozen_value)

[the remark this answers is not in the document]
foolip: I think those definitions are good.

docs/data-guidelines/api.md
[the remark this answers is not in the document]
queengooborg: Huh?  But that's exactly the definition of "subfeature" I was hoping for...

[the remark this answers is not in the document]
hamishwillee: @queengooborg I really hope I haven't been in "violent agreement" with you for most of this discussion :-0.

queengooborg:
This PR is the oldest PR currently open, and I don't see us coming to a resolution any time soon (this is unfortunately a low priority for us right now).  What should we do with this PR?

Elchi3:
I still want to review this. Recently, there were discussions about features, subfeatures, behavioral features in the context of groups.

docs/data-guidelines/features.md
```diff
@@ -0,0 +1,55 @@
...
+  - A feature node is created for the method (by default)
+  - Subfeature nodes are not created for the method parameters because they were created in the same version as the methods; there is no compatibility change.
+- Another browser implements the methods, but only one of the parameters.
+  - A subfeature node is added to the method parameter that was supported by _both_ browsers.
+  - Another subfeature node is added to the method parameter that was _not_ supported by the second browser.
+- Yet another browser implements the method with a new parameter.
+  - A method subfeature is added for the new parameter because it was added in a different version than the parent method.
+- The browser changes to require that the interface now can be used in workers:
+
+  - A "supported in workers" subfeature is added to the interface feature because this is a change in compatibility.
```
caugner, line 52: Here's how I would restructure the example for better readability:

```suggestion
Given: A specification defines an interface with a method that has two parameters.

Assume: One supported browser implements the method and both the parameters.

Then:
- We create a feature node for the method (by default)
- We don't create subfeature nodes for the method parameters, because they were created in the same version as the methods; there is no compatibility change.

Assume: Another browser implements the method, but only one of the parameters.

Then:
- We add one subfeature node for the parameter that is supported by _both_ browsers.
- We add another subfeature node for the parameter that is only supported by the first, but _not_ by the second browser.

Assume: Yet another browser implements the method with a new parameter.

Then:
  - We add yet another subfeature node for the new parameter, because it was added in a different version than the parent method.
- 

Assume: One of the browsers starts supporting the interface in workers:

Then:
  - We add a `worker_support` subfeature (`"Available in workers"`) to the interface (see [Web Workers](./api.md#web-workers-worker_support)).
```

docs/data-guidelines/features.md
```diff
@@ -0,0 +1,55 @@
...
+category                // category name (always `api` for API interfaces)
+  interface               // top level feature
+    properties              // subfeature, add if supported
+      property values         // parameter subfeature, add all of them if not all values were supported since property was introduced
+    methods                // subfeature, add if supported
+      parameters              // parameter subfeature, add all of them if not all parameters were supported since method was introduced
+        parameter options       // parameter subfeature, add all of them if not all parameter options were supported since property was introduced
+        parameter type          // parameter subfeature, add all of them if not all parameter types were supported since property was introduced
+    support in workers     // behavioral subfeature, add if compatibility is different (might also just be on specific methods)
+```
```
caugner, line 29: Ca we use nested lists here, rather than a code block?

```suggestion
- _category_ ← category name (always `api` for API interfaces)
  - _interface_ ← top-level feature
    - _properties_ ← subfeature, add if supported
      - _property values_ ← parameter subfeature, add all of them if not all values were supported since property was introduced
    - _methods_ ← subfeature, add if supported
      - _parameters_ ← parameter subfeature, add all of them if not all parameters were supported since method was introduced
        - _parameter options_ ← parameter subfeature, add all of them if not all parameter options were supported since property was introduced
        - _parameter type_ ← parameter subfeature, add all of them if not all parameter types were supported since property was introduced
    - _worker support_ ← behavioral subfeature, add if compatibility is different (might also just be on specific methods)
```

caugner (changes requested):
LGTM overall, here are three suggestions to hopefully bring this towards a consensus.

docs/data-guidelines/features.md
hamishwillee: @caugner I have no problem with this alternative approach, but I think I'd like the change the rule, not that I've been around a while.

The "rule" for API properties and methods is that you always add entries for all that are supported.

The current rule for method arguments (as I understand it and stated there) is that you only document the incompatibility. So the assumption is that if something is not documented then it is compatible.

I'd like to change the rule to "As soon as there is an incompatibility you should document all parameters including those that were in the original version.

The reason is that as it is now there is no way to spot whether something is compatible or has just been omitted.

hamishwillee: Upshot, if we know the rules I'm open to any presentation at all.

caugner: If this PR explains the rule different than the current status quo, then let's document the status quo first, and then propose the change afterwards. That way we separate concerns and make it easier to move forward.

Anyways, I'll put this on the agenda for the next BCD call.

hamishwillee: Well, I wrote it to "test" the rules. So right now it reflects what I think the rules were, but not necessarily what they are.

caugner:
@hamishwillee Thank you for taking the time to author this guidance PR. Meanwhile, we added [guidelines for behavioral subfeatures](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines/README.md#behavioral-subfeatures), so I think your proposed subfeature guidelines would need to be reworked to avoid overlap. Should we close this chapter?

hamishwillee:
> @hamishwillee Thank you for taking the time to author this guidance PR. Meanwhile, we added [guidelines for behavioral subfeatures](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines/README.md#behavioral-subfeatures), so I think your proposed subfeature guidelines would need to be reworked to avoid overlap. Should we close this chapter?

@caugner As far as I can tell there isn't _any_ overlap, and there is still a need for this information.

Specifically, the data guidelines cover how and what data you can add, but not when you should do so. 

By way of example, see https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines/README.md#parameters-and-parameter-object-features. This explains the format for a parameter. However it doesn't say if you should create a parameter record for every parameter in the method, or just those that were added in different versions.

Similarly I know, for example, that a Web API interface should document all the methods irrespective of whether they were all present in the first version. But I am not sure that is stated anywhere.

We probably close this and someone in the BCD team should take the time to own the problem and solution.

PR #17355, opened 2022-08-15
Update Firefox versions for api.Document.readystatechange_event

queengooborg:
This PR updates and corrects the real values for Firefox and Firefox Android for the `readystatechange_event` member of the `Document` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Document/readystatechange_event

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -5967,7 +5967,7 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "4"
+              "version_added": "9"
             },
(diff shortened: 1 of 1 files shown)


foolip (changes requested):
I don't think that collector results are enough to justify a change like this, since the event may have been supported before the event handler property. Can you either test for the event itself, or track this down in source history?

PR #17359, opened 2022-08-15
Mark 'Report' and 'ReportBody' experimental

OnkarRuikar:
Resolves https://github.com/mdn/content/pull/19570#discussion_r945429779

#### Summary
Only Chromium supports them. So, we need to mark them experimental.


teoli2003 (approved):
👍

PR #17361, opened 2022-08-15
Update Firefox versions for api.EventSource.EventSource

queengooborg:
This PR updates and corrects the real values for Firefox and Firefox Android for the `EventSource` member of the `EventSource` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/EventSource/EventSource

### DIFF
--- a/api/EventSource.json
+++ b/api/EventSource.json
@@ -59,7 +59,7 @@
             },
             "edge": "mirror",
             "firefox": {
-              "version_added": "6"
+              "version_added": "20"
             },
             "firefox_android": {
               "version_added": "45"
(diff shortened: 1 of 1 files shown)


foolip (changes requested):
This seems fishy. The constructor is the only way to use this API AFAIK, so if it's not supported the API as a whole isn't.

https://website-archive.mozilla.org/www.mozilla.org/firefox_releasenotes/en-us/firefox/6.0/releasenotes/ lists `EventSource`, so I guess there's a testing problem here.

queengooborg:
Figured out the cause (I don't know why I didn't dig deeper): the Window exposure returns `null` but the Worker exposure returns `false`, so the resolved value becomes `false`.  I'll fix the collector test.

PR #17362, opened 2022-08-15
Update Chromium/Safari versions for api.EventSource.url

queengooborg:
This PR updates and corrects the real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `url` member of the `EventSource` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/EventSource/url

### DIFF
--- a/api/EventSource.json
+++ b/api/EventSource.json
@@ -445,11 +445,9 @@
               "version_added": "12"
             },
             "safari": {
-              "version_added": "5"
+              "version_added": "6"
             },
-            "safari_ios": {
-              "version_added": "5"
-            },
+            "safari_ios": "mirror",
             "samsunginternet_android": "mirror",
             "webview_android": "mirror"
           },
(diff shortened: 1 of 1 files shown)


foolip (changes requested):
Confirmed that this was added in WebKit trunk version 535.19:
https://github.com/WebKit/WebKit/commit/75f75ebe0d41d3f9fbb8e156f3dd5ced83067275
https://github.com/WebKit/WebKit/blob/75f75ebe0d41d3f9fbb8e156f3dd5ced83067275/Source/WebCore/Configurations/Version.xcconfig

That maps to Chrome 18, but also Safari 6. Can you check if the collector also wants to change the Safari data here?

Note that there was an uppercase `URL` property from the start, if you'd like to add that as an `alternative_name` that'd be nice. It wasn't removed at the same time as `url` was added though, so the removal version would need researching.

queengooborg:
The collector didn't update Safari's data since there was no results for versions older than 6, so I manually tested Safari 5.1 and confirmed a lack of support there.

foolip (approved):

PR #17365, opened 2022-08-15
Update Chromium versions for FontFaceSetLoadEvent API

queengooborg:
This PR updates and corrects the real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `FontFaceSetLoadEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/FontFaceSetLoadEvent

### DIFF
--- a/api/FontFaceSetLoadEvent.json
+++ b/api/FontFaceSetLoadEvent.json
@@ -6,7 +6,7 @@
         "spec_url": "https://drafts.csswg.org/css-font-loading/#fontfacesetloadevent",
         "support": {
           "chrome": {
-            "version_added": "35"
+            "version_added": "57"
           },
           "chrome_android": "mirror",
           "edge": "mirror",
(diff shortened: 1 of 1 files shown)


foolip (changes requested):
This interface has been around in some shape since way earlier:
https://source.chromium.org/chromium/chromium/src/+/2651ddc644941a8ad501b399e4d8ffb2d36cb53c

The test here depends on the constructor being exposed. Can you check which event uses this interface, and see when support for that was added? This data should match, most likely.

queengooborg:
It looks like the existing data matches `api.FontFaceSet.loading_event`, and the changes in this PR agree with when the [constructor was exposed](https://chromestatus.com/feature/4784993343635456).  This is tricky since the interface wasn't exposed beforehand, but we should probably add in some overrides on the collector side.

Issue #14052, opened 2021-12-16
Add data for import assertions


lucacasonato:
Import assertions were shipped to Deno 1.17 and Chrome 91 already. We should add support data for this feature somewhere.

Issue #15123, opened 2022-02-25
javascript.statements.import - Node version incomplete


lucaswerkmeister:
#### What information was incorrect, unhelpful, or incomplete?

[MDN JavaScript modules § import](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules#import) reports that modules are available since Node 13.2.0, or behind `--experimental-modules` between 12.0.0 and 13.2.0. I think this is incomplete: the flag was removed in [12.17.0](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V12.md#2020-05-26-version-12170-erbium-lts-targos), and the feature is considered stable since [12.22.0](https://github.com/nodejs/node/blob/master/doc/changelogs/CHANGELOG_V12.md#2021-03-30-version-12220-erbium-lts-richardlau), so the Node 12 line actually has stable module support nowadays if you’re on a sufficiently recent version.

#### What did you expect to see?

I guess it’s a bit tough to express this in a version range – there are Node 12 versions with better support than some Node 13 versions – but as far as I’m aware there are ways to represent this here, so I’d expect some statement that Node ≥12.22.0 has module support.

#### Did you test this? If so, how?

I tested it on Debian Bullseye, which ships Node 12.22.5; named and default imports/exports work fine.

<details>
<summary>MDN page report details</summary>

* Query: `javascript.statements.import`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/JavaScript/Guide/Modules
* Report started: 2022-02-25T21:26:41.396Z

</details>

Issue #14053, opened 2021-12-16
Add data for JSON module imports


lucacasonato:
The web recently added support for JSON module imports: `import foo from "./foo.json" assert { type: "json" };`. We should document compatibility somewhere. This has been shipped to at least Chrome 91 and Deno 1.17.

Blocked on #14052.

lucacasonato:
I am unsure if this should get the `data:js` or `data:api` label. This feature is specified in the HTML spec, so proably `data:api`. Where would it fit in though?

PR #17366, opened 2022-08-15
Add import assertions

Josh-Cena:
#### Summary

Fix https://github.com/mdn/browser-compat-data/issues/14052, part of https://github.com/mdn/content/issues/19220

First time doing actual compat data changes... I don't really know how this ought to be done.

See https://v8.dev/features/import-assertions

Node versions are hand-tested.

### DIFF
--- a/javascript/operators/import.json
+++ b/javascript/operators/import.json
@@ -92,7 +92,7 @@
                 },
                 {
                   "version_added": "16.14",
-                  "notes": "The second parameter no longer throws a parser error, but the <code>--experimental-json-modules</code> is still needed to load JSON modules."
+                  "notes": "The second parameter no longer throws a parser error, but the <code>--experimental-json-modules</code> flag is still needed to load JSON modules."
                 },
                 {
                   "version_added": "16.0",
--- a/javascript/operators/import.json
+++ b/javascript/operators/import.json
@@ -65,6 +65,65 @@
             "standard_track": true,
             "deprecated": false
           }
+        },
+        "options": {
+          "__compat": {
+            "description": "The second options parameter",
+            "support": {
+              "chrome": {
+                "version_added": "91"
+              },
+              "chrome_android": "mirror",
+              "deno": {
+                "version_added": "1.17"
+              },
+              "edge": "mirror",
+              "firefox": {
+                "version_added": false,
+                "notes": "See <a href='https://bugzil.la/1736059'>bug 1736059</a>."
+              },
+              "firefox_android": "mirror",
+              "ie": {
+                "version_added": false
+              },
+              "nodejs": [
+                {
+                  "version_added": "16.15"
+                },
+                {
+                  "version_added": "16.14",
+                  "notes": "The second parameter no longer throws a parser error, but the <code>--experimental-json-modules</code> is still needed to load JSON modules."
+                },
+                {
+                  "version_added": "16.0",
+                  "flags": [
+                    {
+                      "type": "runtime_flag",
+                      "name": "--harmony-import-assertions"
+                    }
+                  ]
+                }
+              ],
+              "oculus": "mirror",
+              "opera": {
+                "version_added": false
+              },
+              "opera_android": {
+                "version_added": false
+              },
+              "safari": {
+                "version_added": "15"
+              },
+              "safari_ios": "mirror",
+              "samsunginternet_android": "mirror",
+              "webview_android": "mirror"
+            },
+            "status": {
+              "experimental": true,
+              "standard_track": false,
+              "deprecated": false
+            }
+          }
         }
       }
     }
--- a/javascript/statements.json
+++ b/javascript/statements.json
@@ -1377,6 +1377,65 @@
               "deprecated": false
             }
           }
+        },
+        "assertions": {
+          "__compat": {
+            "description": "Import assertions",
+            "support": {
+              "chrome": {
+                "version_added": "91"
+              },
+              "chrome_android": "mirror",
+              "deno": {
+                "version_added": "1.17"
+              },
+              "edge": "mirror",
+              "firefox": {
+                "version_added": false,
+                "notes": "See <a href='https://bugzil.la/1736059'>bug 1736059</a>."
+              },
+              "firefox_android": "mirror",
+              "ie": {
+                "version_added": false
+              },
+              "nodejs": [
+                {
+                  "version_added": "16.15"
+                },
+                {
+                  "version_added": "16.14",
+                  "notes": "<code>assert</code> no longer throws a parser error, but the <code>--experimental-json-modules</code> is still needed to load JSON modules."
+                },
+                {
+                  "version_added": "16.0",
+                  "flags": [
+                    {
+                      "type": "runtime_flag",
+                      "name": "--harmony-import-assertions"
+                    }
+                  ]
+                }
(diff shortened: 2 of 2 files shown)


javascript/operators/import.json
```diff
@@ -65,6 +65,65 @@
             "standard_track": true,
             "deprecated": false
           }
+        },
+        "options": {
+          "__compat": {
+            "description": "The second options parameter",
```
Josh-Cena, line 71: I called it "options" instead of "assertions", because unlike the `import` statement, the options here permit any unknown attributes from the start, so other keys won't throw parser errors either. In this case, should "assertions" be separately listed as a subfeature to increase discoverability?

queengooborg, line 71: The spec seems to call it `opt` for "options" so this sounds good to me -- let's call it the "options" parameter to adhere to the existing structure of these parameters throughout BCD!

(Note: we'll want to make sure to call it as such on MDN as well.)

```suggestion
        },
        "options_parameter": {
          "__compat": {
            "description": "The <code>options</code> parameter",
```

javascript/statements.json
```diff
@@ -1377,6 +1377,65 @@
               "deprecated": false
             }
           }
+        },
+        "assertions": {
```
queengooborg, line 1381: Let's give it a name with an underscore to keep with BCD's data structure:

```suggestion
        "import_assertions": {
```

javascript/operators/import.json
```diff
@@ -65,6 +65,65 @@
...
+                "version_added": false,
+                "notes": "See <a href='https://bugzil.la/1736059'>bug 1736059</a>."
+              },
+              "firefox_android": "mirror",
+              "ie": {
+                "version_added": false
+              },
+              "nodejs": [
+                {
+                  "version_added": "16.15"
```
queengooborg, line 91:
```suggestion
                  "version_added": "16.15.0"
```

javascript/operators/import.json
```diff
@@ -65,6 +65,65 @@
...
+                },
+                {
+                  "version_added": "16.14",
```
queengooborg, line 94:
```suggestion
                  "version_added": "16.14.0",
```

javascript/operators/import.json
```diff
@@ -65,6 +65,65 @@
...
+                  "notes": "The second parameter no longer throws a parser error, but the <code>--experimental-json-modules</code> flag is still needed to load JSON modules."
+                },
+                {
+                  "version_added": "16.0",
```
queengooborg, line 98:
```suggestion
                  "version_added": "16.0.0",
```

javascript/statements.json
```diff
@@ -1377,6 +1377,65 @@
...
+                "version_added": false,
+                "notes": "See <a href='https://bugzil.la/1736059'>bug 1736059</a>."
+              },
+              "firefox_android": "mirror",
+              "ie": {
+                "version_added": false
+              },
+              "nodejs": [
+                {
+                  "version_added": "16.15"
```
queengooborg, line 1403:
```suggestion
                  "version_added": "16.15.0"
```

javascript/statements.json
```diff
@@ -1377,6 +1377,65 @@
...
+                },
+                {
+                  "version_added": "16.14",
```
queengooborg, line 1406:
```suggestion
                  "version_added": "16.14.0",
```

javascript/statements.json
```diff
@@ -1377,6 +1377,65 @@
...
+                  "notes": "<code>assert</code> no longer throws a parser error, but the <code>--experimental-json-modules</code> flag is still needed to load JSON modules."
+                },
+                {
+                  "version_added": "16.0",
```
queengooborg, line 1410:
```suggestion
                  "version_added": "16.0.0",
```

queengooborg (changes requested):
Thanks!  Just a couple of quick updates, and then this will be LGTM!

queengooborg (approved):
Thanks for the quick fixes!

Josh-Cena:
> "16.14.0" is NOT a valid version number for nodejs

Okay that looks hilarious...

Also, I'm pretty confident it should be called experimental, since we only have V8 support.

queengooborg:
Experimental set to true sounds good to me!

I've opened up https://github.com/mdn/browser-compat-data/pull/17367 to add v16.14.0 to the browser data, and will merge that promptly so this PR passes!

queengooborg (changes requested):
Ah, we should actually set the second of the three NodeJS statements to `partial_implementation`, since there are extra details to note regarding compatibility.

By the way, should we create a subfeature for the `json` assertion type?  It might be a little easier to represent the fact that NodeJS added support for the type in `16.15.0` and `17.5.0` (AKA `17.0.0` through `17.4.0` did not have support).

Josh-Cena:
It's complicated for JSON modules. IIRC it's unflagged in 17.5, but backported to 16.15. I don't think we've reached resolution on what to do in that case—do we? I also don't plan to address https://github.com/mdn/browser-compat-data/issues/14053 — just focusing on the syntax in this PR. JSON modules would also involve web API I think. (Again it's complicated.)

queengooborg:
In a case like that, I'd say that we should do the following:

```json
[
  {
    "version_added": "17.5.0"
  },
  {
    "version_added": "16.15.0",
    "version_removed": "17.0.0"
  },
  {
    "version_added": "17.0.0",
    "version_removed": "17.5.0",
    "flags": [
      {
        "type": "runtime_flag",
        "name": "--experimental-json-modules"
      }
    ]
  },
  {
    "version_added": "16.0.0",
    "version_removed": "16.15.0",
    "flags": [
      {
        "type": "runtime_flag",
        "name": "--experimental-json-modules"
      }
    ]
  }
]
```

It's kind of messy, especially with the flag data, but it's the best we can do with Node's LTS release cycle.  WDYT?

Josh-Cena:
This problem is kind of non-unique—also present in https://github.com/mdn/browser-compat-data/issues/15123.

Also, do you want to me to do the `--experimental-json-modules` flag in this PR as well? Frankly I didn't check that. In fact JSON modules and import assertions are two separate things: the latter is merely syntax. I only included a partial implementation note for convenience.

queengooborg:
Yeah, it's an ugly solution, but I feel it is a necessary one due to how NodeJS' release cycle works.  On the plus side, doing this allows us to better represent exactly which versions support the feature!

As for the flag, if you're down to incorporate it, that would be great!  Otherwise, no big deal; flags quickly become useless to document once they're no longer needed for a feature.  ;P

Josh-Cena:
Sorry @queengooborg I think I'm a bit lost on this. Could you update the PR in the way to want so we can get to merge this?

hamishwillee:
@queengooborg ^^^ Could you please propose fixes as suggestions to unblock this?

queengooborg:
I have proposed a solution already, actually -- see https://github.com/mdn/browser-compat-data/pull/17366#issuecomment-1214894126!

Josh-Cena:
The flags data for Node is a bit of a mess—there are two flags: `--harmony-import-assertions` and `--experimental-json-modules`. The first one enables the parser feature, the second one makes `assert { type: "json" }` actually "work" (turns on the JSON loader). I don't really know what to do here...

PR #17371, opened 2022-08-15
Update Chromium versions for api.HTMLAreaElement.origin

queengooborg:
This PR updates and corrects the real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `origin` member of the `HTMLAreaElement` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/HTMLAreaElement/origin

### DIFF
--- a/api/HTMLAreaElement.json
+++ b/api/HTMLAreaElement.json
@@ -369,10 +369,10 @@
             "opera": "mirror",
             "opera_android": "mirror",
             "safari": {
-              "version_added": "5.1"
+              "version_added": "6"
             },
             "safari_ios": {
-              "version_added": "5"
+              "version_added": "6"
             },
             "samsunginternet_android": "mirror",
             "webview_android": {
--- a/api/HTMLAreaElement.json
+++ b/api/HTMLAreaElement.json
@@ -369,10 +369,10 @@
             "opera": "mirror",
             "opera_android": "mirror",
             "safari": {
-              "version_added": "6"
+              "version_added": "10.1"
             },
             "safari_ios": {
-              "version_added": "6"
+              "version_added": "10.3"
             },
             "samsunginternet_android": "mirror",
             "webview_android": {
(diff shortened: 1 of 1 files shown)


api/HTMLAreaElement.json
```diff
@@ -351,7 +351,7 @@
           "spec_url": "https://html.spec.whatwg.org/multipage/links.html#dom-hyperlink-origin-dev",
           "support": {
             "chrome": {
-              "version_added": "8"
+              "version_added": "32"
```
foolip, line 354: Manually confirmed with https://mdn-bcd-collector.appspot.com/tests/api/HTMLAreaElement/origin and the test is so simple there's little room for doubt.

But if the Chrome data was wrong, so was Safari most likely. Here's the commit I think added this in Safari:
https://github.com/WebKit/WebKit/commit/fe3dc0b4f4d5406eeb5e77e962570bd8a9921cad
https://github.com/WebKit/WebKit/blob/fe3dc0b4f4d5406eeb5e77e962570bd8a9921cad/Source/WebCore/Configurations/Version.xcconfig

Can you test Safari too?

foolip (changes requested):

api/HTMLAreaElement.json
```diff
@@ -369,10 +369,10 @@
             "opera": "mirror",
             "opera_android": "mirror",
             "safari": {
-              "version_added": "5.1"
+              "version_added": "6"
```
foolip, line 372: This doesn't seem right, http://mdn-bcd-collector.appspot.com/tests/api/HTMLAreaElement/origin fails in Safari 9.1.

foolip (changes requested):

api/HTMLAreaElement.json
```diff
@@ -369,10 +369,10 @@
             "opera": "mirror",
             "opera_android": "mirror",
             "safari": {
-              "version_added": "5.1"
+              "version_added": "10.1"
```
foolip, line 372: Sauce Labs has Safari 10, and https://mdn-bcd-collector.appspot.com/tests/api/HTMLAreaElement/origin passes there. That's what WebKit version mapping also points at.

PR #17385, opened 2022-08-15
Update Chromium versions for api.Location.toString

queengooborg:
This PR updates and corrects the real values for Chromium for the `toString` member of the `Location` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Location/toString

### DIFF
--- a/api/Location.json
+++ b/api/Location.json
@@ -649,7 +649,7 @@
           "spec_url": "https://html.spec.whatwg.org/multipage/history.html#dom-location-href-dev",
           "support": {
             "chrome": {
-              "version_added": "52"
+              "version_added": "29"
             },
             "chrome_android": "mirror",
             "deno": {


api/Location.json
```diff
@@ -649,7 +649,7 @@
           "spec_url": "https://html.spec.whatwg.org/multipage/history.html#dom-location-href-dev",
           "support": {
             "chrome": {
-              "version_added": "52"
+              "version_added": "29"
```
foolip, line 652: This is not Firefox. Was it the Firefox data that was supposed to be updated to 29?

foolip (changes requested):

api/Location.json
queengooborg: No, Firefox is fine, I just selected the wrong browser option in my PR-creating script.  :P

foolip: I've failed to confirm that Chrome 29 is correct using https://mdn-bcd-collector.appspot.com/tests/api/Location/toString. The test also passes on Chrome 28.

foolip (changes requested):

queengooborg:
It looks like a custom test was added in collector v6.1.1 so new results need to be collected.  I'll hold off until v6.1.2 is released.

PR #17386, opened 2022-08-15
Update Safari versions for MerchantValidationEvent API

queengooborg:
This PR updates and corrects the real values for Safari (Desktop and iOS/iPadOS) for the `MerchantValidationEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.gooborg.com) project (v6.1.0).

Tests Used: https://mdn-bcd-collector.gooborg.com/tests/api/MerchantValidationEvent

### DIFF
--- a/api/PaymentRequest.json
+++ b/api/PaymentRequest.json
@@ -276,7 +276,7 @@
             "opera": "mirror",
             "opera_android": "mirror",
             "safari": {
-              "version_added": "11.1"
+              "version_added": "12.1"
             },
             "safari_ios": "mirror",
             "samsunginternet_android": "mirror",
--- a/api/MerchantValidationEvent.json
+++ b/api/MerchantValidationEvent.json
@@ -28,7 +28,7 @@
           "opera": "mirror",
           "opera_android": "mirror",
           "safari": {
-            "version_added": "11.1"
+            "version_added": "12.1"
           },
           "safari_ios": "mirror",
           "samsunginternet_android": "mirror",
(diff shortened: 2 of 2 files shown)


foolip (changes requested):
This should match https://developer.mozilla.org/docs/Web/API/PaymentRequest/merchantvalidation_event, can you update that data too?

foolip:
But also confirm if the merchantvalidation event is actually supported, not just assume that the interface data is correct.

api/PaymentRequest.json
```diff
@@ -276,7 +276,7 @@
             "opera": "mirror",
             "opera_android": "mirror",
             "safari": {
-              "version_added": "11.1"
+              "version_added": "12.1"
```
foolip, line 279: I've confirmed support with https://mdn-bcd-collector.appspot.com/tests/api/PaymentRequest/merchantvalidation_event in Safari 11.1 on BrowserStack. Making an edit like this which is contradicted by collector results will likely result in it surfacing again later, and likely making the change without noticing it's inconsistent with `MerchantValidationEvent`.

foolip (changes requested):

foolip:
> But also confirm if the merchantvalidation event is actually supported, not just assume that the interface data is correct.

This turned out to be important in this case. This is a case of the interface not being exposed even though the event was:
https://github.com/WebKit/WebKit/commit/90a5b4d7316500b38d277d8bca709b33d6ec027e

The event and interface was originally prefixed with `apple`/`Apple` and that's something we could optionally record as well:
https://github.com/WebKit/WebKit/commit/7678653a3cb439d11074cbe8c5b7d47aa31d1905

queengooborg:
This is for pre-2020 browser releases, so I'm going to close this and worry about it another time.

PR #17388, opened 2022-08-15
Update Chrome/Safari versions for MessageEvent API

queengooborg:
This PR updates and corrects the real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `MessageEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.0).

### DIFF
--- a/api/MessageEvent.json
+++ b/api/MessageEvent.json
@@ -54,7 +54,7 @@
           "spec_url": "https://html.spec.whatwg.org/multipage/comms.html#the-messageevent-interface:dom-event-constructor",
           "support": {
             "chrome": {
-              "version_added": "1"
+              "version_added": "16"
             },
             "chrome_android": "mirror",
             "deno": {
--- a/api/MessageEvent.json
+++ b/api/MessageEvent.json
@@ -81,7 +81,7 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "4"
+              "version_added": "6"
             },
             "safari_ios": "mirror",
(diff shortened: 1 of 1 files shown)


foolip (dismissed):

api/MessageEvent.json
```diff
@@ -54,7 +54,7 @@
           "spec_url": "https://html.spec.whatwg.org/multipage/comms.html#the-messageevent-interface:dom-event-constructor",
           "support": {
             "chrome": {
-              "version_added": "1"
+              "version_added": "16"
```
foolip, line 57: Hmm, how did you determine this? BrowserStack doesn't have Chrome 15. I wanted to check how the test fails in Chrome 15, to confirm it's not actually supported.

queengooborg: It does have Chrome 15, actually!  That's why the collector's lowest version to run Selenium on is Chrome 15.

foolip: Hmm, I guess I was looking on the wrong platforms then. Can you paste the error message from Chrome 15 when trying this constructor?

queengooborg: The resulting error is pretty straightforward: `threw TypeError: Illegal constructor`

foolip (approved):

foolip (commented):
Before merging could you also check if Safari data doesn't need updating?

foolip (approved):

PR #17389, opened 2022-08-15
Update Edge/IE versions for MessageEvent API

queengooborg:
This PR updates and corrects the real values for Microsoft Edge for the `MessageEvent` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.0).

### DIFF
--- a/api/MessageEvent.json
+++ b/api/MessageEvent.json
@@ -68,7 +68,7 @@
             },
             "firefox_android": "mirror",
             "ie": {
-              "version_added": "9"
+              "version_added": false
             },
             "nodejs": {
               "version_added": "15.0.0"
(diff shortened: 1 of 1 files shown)


api/MessageEvent.json
```diff
@@ -302,7 +302,7 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "3"
+              "version_added": "28"
```
foolip, line 305: This is not Edge.

foolip (changes requested):

foolip (changes requested):
IE should probably also change for this to make sense.

PR #17410, opened 2022-08-17
Add data for Intl.[[FallbackSymbol]]

Josh-Cena:
#### Summary

This is something I wish to document. To do this we first need compat data.

I shamelessly stole the data from here: https://github.com/ljharb/intl-fallback-symbol

I've confirmed the Node versions myself.

I think the description looks a bit awful. Let me know if there're any concerns.

### DIFF
--- a/javascript/builtins/Intl/DateTimeFormat.json
+++ b/javascript/builtins/Intl/DateTimeFormat.json
@@ -116,7 +116,9 @@
                   "firefox": {
                     "version_added": "54"
                   },
-                  "firefox_android": "mirror",
+                  "firefox_android": {
+                    "version_added": "56"
+                  },
                   "ie": {
                     "version_added": false
                   },
--- a/javascript/builtins/Intl/DateTimeFormat.json
+++ b/javascript/builtins/Intl/DateTimeFormat.json
@@ -99,7 +99,7 @@
             },
             "IntlLegacyConstructedSymbol": {
               "__compat": {
-                "description": "Supports normative optional ChainDateTimeFormat behavior",
+                "description": "Supports normative optional <code>ChainDateTimeFormat</code> behavior",
                 "support": {
                   "chrome": [
                     {
--- a/javascript/builtins/Intl/DateTimeFormat.json
+++ b/javascript/builtins/Intl/DateTimeFormat.json
@@ -97,6 +97,59 @@
                 "deprecated": false
               }
             },
+            "IntlLegacyConstructedSymbol": {
+              "__compat": {
+                "description": "Supports normative optional ChainDateTimeFormat behavior",
+                "support": {
+                  "chrome": [
+                    {
+                      "version_added": "91"
+                    },
+                    {
+                      "version_added": "57",
+                      "partial_implementation": true,
+                      "notes": "The fallback symbol property has description <code>IntlFallback</code>."
+                    }
+                  ],
+                  "chrome_android": "mirror",
+                  "deno": {
+                    "version_added": null
+                  },
+                  "edge": "mirror",
+                  "firefox": {
+                    "version_added": "54"
+                  },
+                  "firefox_android": "mirror",
+                  "ie": {
+                    "version_added": false
+                  },
+                  "nodejs": [
+                    {
+                      "version_added": "16.0.0"
+                    },
+                    {
+                      "version_added": "8.0.0",
+                      "partial_implementation": true,
+                      "notes": "The fallback symbol property has description <code>IntlFallback</code>."
+                    }
+                  ],
+                  "oculus": "mirror",
+                  "opera": "mirror",
+                  "opera_android": "mirror",
+                  "safari": {
+                    "version_added": "14.1"
+                  },
+                  "safari_ios": "mirror",
+                  "samsunginternet_android": "mirror",
+                  "webview_android": "mirror"
+                },
+                "status": {
+                  "experimental": false,
+                  "standard_track": true,
+                  "deprecated": true
+                }
+              }
+            },
             "locales_parameter": {
               "__compat": {
                 "description": "<code>locales</code> parameter",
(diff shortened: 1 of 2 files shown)


queengooborg:
Hey @Josh-Cena, thanks for adding this in!  Would you mind describing to me what it this feature is used for?

Josh-Cena:
Sure, here's my draft for the content update:

> ### Return value
> 
> A new `Intl.DateTimeFormat` object.
> 
> > **Note:** The text below describes behavior that is marked by the specification as "optional". It may not work in all environments. Check the [browser compatibility table](#browser_compatibility).
> 
> Normally, `Intl.DateTimeFormat()` can be called with or without [`new`](/en-US/docs/Web/JavaScript/Reference/Operators/new), and a new `Intl.DateTimeFormat` instance is returned in both cases. However, if the [`this`](/en-US/docs/Web/JavaScript/Reference/Operators/this) value is an object that is [`instanceof`](/en-US/docs/Web/JavaScript/Reference/Operators/instanceof) `Intl.DateTimeFormat` (doesn't necessarily mean it's created via `new Intl.DateTimeFormat`; just that it has `Intl.DateTimeFormat.prototype` in its prototype chain), then the value of `this` is returned instead, with the newly created `Intl.DateTimeFormat` object hidden in a `[Symbol(IntlLegacyConstructedSymbol)]` property.
> 
> ```js
> const formatter = Intl.DateTimeFormat.call(
>   { __proto__: Intl.DateTimeFormat.prototype },
>   "en-US",
>   { dateStyle: "full" },
> );
> console.log(Object.getOwnPropertyDescriptors(formatter));
> // {
> //   [Symbol(IntlLegacyConstructedSymbol)]: {
> //     value: DateTimeFormat [Intl.DateTimeFormat] {},
> //     writable: false,
> //     enumerable: false,
> //     configurable: false
> //   }
> // }
> ```
> 
> Note that there's only one actual `Intl.DateTimeFormat` instance here: the one hidden in `[Symbol(IntlLegacyConstructedSymbol)]`. When calling the [`format()`](/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/format) and [`resolvedOptions()`](/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/resolvedOptions) methods on `formatter`, it would correctly use the options stored in that instance, but calling all other methods (e.g. [`formatRange()`](/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl/DateTimeFormat/formatRange)) would fail: "TypeError: formatRange method called on incompatible Object", because those methods don't consult the hidden instance's options.
> 
> This behavior does not happen when `Intl.DateTimeFormat()` is called without `new` but with `this` set to anything else that's not an `instanceof Intl.DateTimeFormat`. If you call it directly as `Intl.DateTimeFormat()`, the `this` value is [`Intl`](/en-US/docs/Web/JavaScript/Reference/Global_Objects/Intl), and a new `Intl.DateTimeFormat` instance is created normally.

javascript/builtins/Intl/DateTimeFormat.json
```diff
@@ -97,6 +97,55 @@
                 "deprecated": false
               }
             },
+            "IntlLegacyConstructedSymbol": {
+              "__compat": {
+                "description": "Supports normative optional ChainDateTimeFormat behavior",
```
queengooborg, line 102:
```suggestion
                "description": "Supports normative optional <code>ChainDateTimeFormat</code> behavior",
```

javascript/builtins/Intl/NumberFormat.json
```diff
@@ -97,6 +97,55 @@
                 "deprecated": false
               }
             },
+            "IntlLegacyConstructedSymbol": {
+              "__compat": {
+                "description": "Supports normative optional ChainNumberFormat behavior",
```
queengooborg, line 102:
```suggestion
                "description": "Supports normative optional <code>ChainNumberFormat</code> behavior",
```

queengooborg (changes requested):
Alright, thank you for the info!  Just got a couple of small tweaks, and this'll be LGTM!

javascript/builtins/Intl/DateTimeFormat.json
```diff
@@ -97,6 +97,55 @@
...
+                      "version_added": "57",
+                      "partial_implementation": true,
+                      "notes": "The fallback symbol property has description <code>IntlFallback</code>."
+                    }
+                  ],
+                  "chrome_android": "mirror",
+                  "edge": "mirror",
+                  "firefox": {
+                    "version_added": "54"
+                  },
```
queengooborg, line 118:
```suggestion
                  },
                  "firefox_android": "mirror",
```

javascript/builtins/Intl/NumberFormat.json
```diff
@@ -97,6 +97,55 @@
...
+                      "version_added": "57",
+                      "partial_implementation": true,
+                      "notes": "The fallback symbol property has description <code>IntlFallback</code>."
+                    }
+                  ],
+                  "chrome_android": "mirror",
+                  "edge": "mirror",
+                  "firefox": {
+                    "version_added": "54"
+                  },
```
queengooborg, line 118:
```suggestion
                  },
                  "firefox_android": "mirror",
```

queengooborg (approved):
Whoops, look like we're missing Firefox Android here...which the linter should have caught?  🤔

queengooborg (approved):
Thanks, LGTM!

javascript/builtins/Intl/DateTimeFormat.json
```diff
@@ -97,6 +97,56 @@
...
+                  "firefox_android": "mirror",
```
queengooborg, line 119:
```suggestion
                  "firefox_android": {
                    "version_added": "56"
                  },
```

javascript/builtins/Intl/NumberFormat.json
```diff
@@ -97,6 +97,56 @@
...
+                  "firefox_android": "mirror",
```
queengooborg, line 119:
```suggestion
                  "firefox_android": {
                    "version_added": "56"
                  },
```

queengooborg (changes requested):
Ah, wait, almost there -- it looks like Firefox Android support came a little bit later for this feature!

Josh-Cena:
Ah, I think I removed it because I have no idea when it was introduced...

queengooborg (approved):
I don't know why the linter isn't verifying the presence of the browser here, because it definitely should be, hmm...  I'll look into it later, thank you for your PR!

PR #17418, opened 2022-08-17
Update support info for Node.js support of `AbortSignal`

aduh95:
#### Summary

Data was wrong (it showed no support on the Node.js cell, when it fact Node.js does support it), so I updated it to be conform to reality.

- `reason` property and parameter were added with https://github.com/nodejs/node/pull/40807
- `thorwIfAborted` was added with https://github.com/nodejs/node/pull/40951
- `timeout` was added with https://github.com/nodejs/node/pull/40899

### DIFF
--- a/api/AbortSignal.json
+++ b/api/AbortSignal.json
@@ -100,9 +100,15 @@
               "ie": {
                 "version_added": false
               },
-              "nodejs": {
-                "version_added": "17.2.0"
-              },
+              "nodejs": [
+                {
+                  "version_added": "17.2.0",
+                },
+                {
+                  "version_added": "16.14.0",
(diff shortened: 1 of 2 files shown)


queengooborg (changes requested):
Hey @aduh95, thank you for your PR!  It looks like the feature was also [backported to 16.14.0](https://nodejs.org/en/blog/release/v16.14.0/).  Could we create two separate release statements to represent this, like so?

```json
{
  "nodejs": [
    {
      "version_added": "17.2.0",
    },
    {
      "version_added": "16.14.0",
      "version_removed": "17.0.0",
    }
  ]
}
```

queengooborg (approved):
Thank you, LGTM!

PR #17420, opened 2022-08-17
Update Chromium versions for ReadableStream API

queengooborg:
This PR updates and corrects the real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `ReadableStream` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/ReadableStream

### DIFF
--- a/api/ReadableStream.json
+++ b/api/ReadableStream.json
@@ -52,7 +52,7 @@
           "spec_url": "https://streams.spec.whatwg.org/#ref-for-rs-constructor⑤",
           "support": {
             "chrome": {
-              "version_added": "43"
+              "version_added": "52"
             },
             "chrome_android": "mirror",
             "deno": {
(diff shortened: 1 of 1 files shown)


api/ReadableStream.json
```diff
@@ -52,7 +52,7 @@
           "spec_url": "https://streams.spec.whatwg.org/#ref-for-rs-constructor⑤",
           "support": {
             "chrome": {
-              "version_added": "43"
+              "version_added": "52"
```
foolip, line 55: How did things behave in Chrome 43-51? Was the interface really exposed, but it didn't work? I suspect more changes are needed to make the data consistent around this.

foolip (changes requested):

api/ReadableStream.json
queengooborg: That seems to be the case, yes.  In Chrome 51, `new ReadableStream()` simply returns `Illegal constructor`.

foolip: I see, thanks for checking!

foolip (approved):

PR #17422, opened 2022-08-18
Update all browsers versions for SVGAltGlyphElement API

queengooborg:
This PR updates and corrects the real values for all browsers for the `SVGAltGlyphElement` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/SVGAltGlyphElement

### DIFF
--- a/api/SVGAltGlyphElement.json
+++ b/api/SVGAltGlyphElement.json
@@ -6,12 +6,14 @@
         "spec_url": "https://www.w3.org/TR/SVG11/text.html#InterfaceSVGAltGlyphElement",
         "support": {
           "chrome": {
-            "version_added": "1"
+            "version_added": "1",
+            "version_removed": "40"
           },
           "chrome_android": "mirror",
           "edge": "mirror",
           "firefox": {
             "version_added": "4",
+            "version_removed": "49",
             "notes": "partial support, see <a href='https://bugzil.la/456286'>bug 456286</a> and <a href='https://bugzil.la/571808'>bug 571808</a>."
           },
           "firefox_android": "mirror",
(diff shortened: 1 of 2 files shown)


api/SVGAltGlyphElement.json
```diff
@@ -6,35 +6,31 @@
...
             "version_added": false
           },
           "oculus": "mirror",
-          "opera": {
-            "version_added": "10.6"
-          },
-          "opera_android": {
-            "version_added": "11"
-          },
+          "opera": "mirror",
```
foolip, line 24: This will change `version_added` to "15". Wasn't this in 10.6 as the previous data claims?

queengooborg: It was not supported before Opera 15, no; running the test in the collector on Opera 12.16 with https://github.com/foolip/mdn-bcd-collector/pull/2301 shows that the object is an `SVGElement` object, AKA a generic object.

foolip: That checks out. I also confirmed that `<altGlyph>` parses to an `SVGElement`.

Can you update `svg/elements/altGlyph.json` as well?

foolip (changes requested):

svg/elements/altGlyph.json
```diff
@@ -22,12 +22,8 @@
               "version_added": null
             },
             "oculus": "mirror",
-            "opera": {
-              "version_added": "10.6"
-            },
-            "opera_android": {
-              "version_added": "11"
-            },
+            "opera": "mirror",
```
foolip, line 25: Can you also make the other data consistent with the interface? In particular, it was removed from Chrome.

foolip (approved):

PR #17424, opened 2022-08-18
Update Firefox versions for SVGAnimatedRect API

queengooborg:
This PR updates and corrects the real values for Firefox and Firefox Android for the `SVGAnimatedRect` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/SVGAnimatedRect

### DIFF
--- a/api/SVGAnimatedRect.json
+++ b/api/SVGAnimatedRect.json
@@ -13,7 +13,7 @@
             "version_added": "12"
           },
           "firefox": {
-            "version_added": "1.5"
+            "version_added": "15"
           },
           "firefox_android": "mirror",
           "ie": {


foolip:
The test here could give false negatives if `viewBox` wasn't supported back then. Can you check when the interface was first exposed? If that's also FF15 I'm convinced :)

queengooborg:
It looks like the problem resides with the interface of choice: `SVGViewElement` wasn't supported until Firefox 15, but `SVGSVGElement`, which also has `viewBox`, was around since Fx 1.5.  I'll tweak the test!

PR #17426, opened 2022-08-18
Update Safari versions for api.SVGStringList.length

queengooborg:
This PR updates and corrects the real values for Safari (Desktop and iOS/iPadOS) for the `length` member of the `SVGStringList` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/SVGStringList/length

### DIFF
--- a/api/SVGStringList.json
+++ b/api/SVGStringList.json
@@ -280,7 +280,7 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "13"
+              "version_added": "13.1"
             },
             "safari_ios": "mirror",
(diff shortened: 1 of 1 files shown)


foolip:
My question in https://github.com/mdn/browser-compat-data/pull/17428#pullrequestreview-1093324704 applies here too, except that the previous edit is from https://github.com/mdn/browser-compat-data/pull/12118.

foolip (approved):
Just assuming this is correct without closer review.

PR #17428, opened 2022-08-18
Update Safari versions for api.SVGTransformList.length

queengooborg:
This PR updates and corrects the real values for Safari (Desktop and iOS/iPadOS) for the `length` member of the `SVGTransformList` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/SVGTransformList/length

### DIFF
--- a/api/SVGTransformList.json
+++ b/api/SVGTransformList.json
@@ -367,7 +367,7 @@
               "version_added": "≤12.1"
             },
             "safari": {
-              "version_added": "13"
+              "version_added": "13.1"
             },
             "safari_ios": "mirror",
             "samsunginternet_android": {


foolip (commented):
This was last updated in https://github.com/mdn/browser-compat-data/pull/12119 also based on collector results, and the test seems to have been the same then:
https://v3-dot-mdn-bcd-collector.appspot.com/tests/api/SVGTransformList

Can you look into what's changed? I don't want to churn the data without understanding why it's changing, since there's always a risk that collector bugs are the explanation.

queengooborg:
I remember that we were doing some changes to the UA parser that involved minor Safari versions, so I have a feeling this was one of the effects that came from the previous bug.

foolip:
> I remember that we were doing some changes to the UA parser that involved minor Safari versions, so I have a feeling this was one of the effects that came from the previous bug.

I don't remember there being a 13 vs. 13.1 issue, but in any case 13.1 can be confirmed by source:
https://github.com/WebKit/WebKit/commit/575ac5551fcea1e5cf5b66a3370fbe3ad1aae8c8
https://github.com/WebKit/WebKit/blob/575ac5551fcea1e5cf5b66a3370fbe3ad1aae8c8/Source/WebCore/Configurations/Version.xcconfig

foolip (approved):

PR #17430, opened 2022-08-18
Update Chromium versions for api.ServiceWorkerContainer.controller

queengooborg:
This PR updates and corrects the real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `controller` member of the `ServiceWorkerContainer` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/ServiceWorkerContainer/controller

### DIFF
--- a/api/ServiceWorkerContainer.json
+++ b/api/ServiceWorkerContainer.json
@@ -48,7 +48,7 @@
               "version_added": "17"
             },
             "firefox": {
-              "version_added": "74"
+              "version_added": "44"
             },
             "firefox_android": "mirror",
             "ie": {
--- a/api/ServiceWorkerContainer.json
+++ b/api/ServiceWorkerContainer.json
@@ -41,14 +41,14 @@
           "spec_url": "https://w3c.github.io/ServiceWorker/#navigator-service-worker-controller",
           "support": {
             "chrome": {
-              "version_added": "40"
+              "version_added": "54"
             },
             "chrome_android": "mirror",
             "edge": {
               "version_added": "17"
             },
             "firefox": {
-              "version_added": "44"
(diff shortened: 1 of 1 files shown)


api/ServiceWorkerContainer.json
```diff
@@ -41,7 +41,7 @@
           "spec_url": "https://w3c.github.io/ServiceWorker/#navigator-service-worker-controller",
           "support": {
             "chrome": {
-              "version_added": "40"
+              "version_added": "54"
```
foolip, line 44: I tracked this down to https://storage.googleapis.com/chromium-find-releases-static/77a.html#77a9444d1b574bb603e5a79d588e6f9e93361b1e

The `controller` property has been there since way before:
https://chromium.googlesource.com/chromium/src/+/d51c935c71766b703d7adca88ef1cdc947c90699

So this is giving an incorrect result because we're testing the prototype, but this property had `[Unforgeable]` at one point.

foolip (changes requested):

api/ServiceWorkerContainer.json
queengooborg: Should this be left as "40" then?

foolip: I've sent https://github.com/foolip/mdn-bcd-collector/pull/2283.

foolip: https://mdn-bcd-collector.appspot.com/tests/api/ServiceWorkerContainer/controller is now updated, and I've confirmed 40 is the right version on BrowserStack.

PR #17440, opened 2022-08-18
Update Chromium versions for api.WritableStreamDefaultWriter.WritableStreamDefaultWriter

queengooborg:
This PR updates and corrects the real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `WritableStreamDefaultWriter` member of the `WritableStreamDefaultWriter` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/WritableStreamDefaultWriter/WritableStreamDefaultWriter

### DIFF
--- a/api/WritableStreamDefaultWriter.json
+++ b/api/WritableStreamDefaultWriter.json
@@ -65,7 +65,7 @@
           "spec_url": "https://streams.spec.whatwg.org/#ref-for-default-writer-constructor①",
           "support": {
             "chrome": {
-              "version_added": "59"
+              "version_added": "78"
             },
             "chrome_android": "mirror",
             "deno": {


foolip:
Is this an API that makes sense to say that it's supported without the constructor? How did one use the API before the constructor was supported?

queengooborg:
It looks like instances of this interface are also returned by [`WritableStream.getWriter()`](https://developer.mozilla.org/en-US/docs/Web/API/WritableStream/getWriter)!

foolip:
Aha, then it makes sense!

foolip (approved):

PR #17454, opened 2022-08-18
Update Chromium versions for api.Window.appinstalled_event

queengooborg:
This PR updates and corrects the real values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `appinstalled_event` member of the `Window` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Window/appinstalled_event

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -180,7 +180,7 @@
           "spec_url": "https://wicg.github.io/manifest-incubations/#dom-window-onappinstalled",
           "support": {
             "chrome": {
-              "version_added": "64"
+              "version_added": "61"
             },
             "chrome_android": {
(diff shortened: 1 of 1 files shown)


foolip (changes requested):
Looks like the properties were added after the event itself in this case:
https://storage.googleapis.com/chromium-find-releases-static/d0d.html#d0d175487822ae4ec8a217a6aa8abe8ab78da02e

More digging would be needed to determine the correct versions for an initial `partial_implementation` statement.

PR #17464, opened 2022-08-18
Fix incorrect sentence

tkrotoff:
Fix sentence `this header was not sent with sent with` => `this header was not sent with`


Elchi3 (approved):
Yes! Thank you and welcome to mdn-browser-compat-data! 🎉

PR #17468, opened 2022-08-19
Remove api.PresentationRequest.startWithDevice from BCD

queengooborg:
This PR removes the irrelevant `startWithDevice` member of the `PresentationRequest` API as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-features). The lack of current support has been confirmed by the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.0), even if the current BCD suggests support.

### DIFF
--- a/api/PresentationRequest.json
+++ b/api/PresentationRequest.json
@@ -302,46 +302,6 @@
             "deprecated": false
           }
         }
-      },
-      "startWithDevice": {
-        "__compat": {
-          "support": {
-            "chrome": {
-              "version_added": false
-            },
-            "chrome_android": "mirror",
-            "edge": "mirror",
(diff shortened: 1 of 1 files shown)


foolip (approved):

foolip:
> even if the current BCD suggests support

It doesn't though, was this a template commit message?

queengooborg:
It is, yes, it's another function of my auto-pr script -- numerous times I've gone to update the existing data per collector results, and then find out that no browsers support it, so rather than modify the data, wait for a review, then remove the feature, I added that instead to indicate that removal was not a mistake!

PR #17469, opened 2022-08-19
CSS update contain: style and add contain: inline-size

hamishwillee:
Updates to the contain property

1. `contain: style`
   - FF103 support (already added as a note in #17326) now done properly as a subfeature). 
   - Also added in Safari 15.4 (tech preview https://developer.apple.com/safari/technology-preview/release-notes/)
   - Chrome - browser test of WPT shows works in original version.

3. Added `contain: inline-size`
   - FF101: https://bugzilla.mozilla.org/show_bug.cgi?id=1755565
   - Chrome 105: https://chromestatus.com/feature/6525308435955712
   - Safari tech preview: https://developer.apple.com/safari/technology-preview/release-notes/#:~:text=Enabled%20CSS%20Container%20Queries%20by%20default

Also verified that the other cases: strict, paint, content, size layout all appear in the first version of respective browser. This was done using spot checks of https://wpt.live/css/css-contain/ . Note that chrome had spotty support against the tests for `paint` up until Chrome 70, but can't find associated tests, and that's a while ago. So no need to add note I don't think.

### DIFF
--- a/css/properties/contain.json
+++ b/css/properties/contain.json
@@ -11,24 +11,15 @@
             },
             "chrome_android": "mirror",
             "edge": "mirror",
-            "firefox": [
-              {
-                "version_added": "103"
-              },
-              {
-                "version_added": "69",
-                "partial_implementation": true,
-                "notes": "Firefox does not support the <code>style</code> value."
-              }
-            ],
+            "firefox": {
+                "version_added": "69"
+            },
             "firefox_android": "mirror",
             "ie": {
               "version_added": false
             },
             "oculus": "mirror",
-            "opera": {
-              "version_added": "40"
-            },
+            "opera": "mirror",
             "opera_android": "mirror",
             "safari": {
               "version_added": "15.4"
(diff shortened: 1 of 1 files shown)


hamishwillee:
@queengooborg Please review!

css/properties/contain.json
```diff
@@ -11,24 +11,15 @@
...
+            },
             "firefox_android": "mirror",
             "ie": {
               "version_added": false
             },
             "oculus": "mirror",
-            "opera": {
-              "version_added": "40"
-            },
+            "opera": "mirror",
```
hamishwillee, line 22: Note, 39 is the matching version with chrome, and testing shows this is indeed the version that the options start working in.

css/properties/contain.json
```diff
@@ -42,6 +33,76 @@
             "standard_track": true,
             "deprecated": false
           }
+        },
+        "inline-size_value": {
+          "__compat": {
+            "description": "<code>inline-size</code> value",
```
queengooborg, line 39: For CSS properties, we don't need to suffix the subfeatures with `_value` -- like properties and methods of interfaces, there's a natural implication that the regular subfeatures of a CSS property is a possible value.

```suggestion
        "inline-size": {
          "__compat": {
```

css/properties/contain.json
```diff
@@ -42,6 +33,76 @@
...
+            "status": {
+              "experimental": false,
+              "standard_track": true,
+              "deprecated": false
+            }
+          }
+        },
+        "style_value": {
+          "__compat": {
+            "description": "<code>style</code> value",
```
queengooborg, line 74:
```suggestion
        "style": {
          "__compat": {
```

queengooborg (changes requested):

hamishwillee:
THanks . I assume the pattern for CSS values is documented?

queengooborg (approved):
Thank you, LGTM!

(And yes, I believe this is documented as a data guideline!)

PR #17472, opened 2022-08-19
Remove status blocks from webextensions

queengooborg:
This PR removes the status blocks from the `webextensions/` folder, unblocking #17471.


xeenon:
I think you meant issue #17258 in the commit message.

queengooborg:
Oops, actually I meant a different PR, I swapped two of the digits around when typing the number out -- fixed!

PR #17473, opened 2022-08-19
Add data for 2d canvas willReadFrequently parameter

foolip:
Support in Firefox 28 and Chrome 99, and lack of support in Safari, was
confirmed with this test:
https://software.hixie.ch/utilities/js/live-dom-viewer/?saved=10586

Additional evidence:
https://bugzilla.mozilla.org/show_bug.cgi?id=884226
https://chromestatus.com/feature/6051647656558592

A WebKit bug was filed and linked here.

### DIFF
--- a/api/HTMLCanvasElement.json
+++ b/api/HTMLCanvasElement.json
@@ -751,6 +751,41 @@
                 "deprecated": false
               }
             }
+          },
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
Nice, I remember documenting this for Firefox at some point 👍

PR #17474, opened 2022-08-19
Add data for 2d canvas colorSpace parameter

foolip:
Support in Chrome 92 and Safari 15.6 was confirmed with this test:
https://software.hixie.ch/utilities/js/live-dom-viewer/?saved=10584

(Chrome doesn't support the color() function, and still paints black.)

Assume that Safari 15.2 is correct, matching ImageData's colorSpace
property.

Supporting evidence:
https://developer.apple.com/documentation/safari-release-notes/safari-15_2-release-notes
https://storage.googleapis.com/chromium-find-releases-static/f1e.html#f1e5b1b1561a449e1d6096bce01fcc6cd7e690ff

### DIFF
--- a/api/HTMLCanvasElement.json
+++ b/api/HTMLCanvasElement.json
@@ -704,6 +704,40 @@
               }
             }
           },
+          "options_colorSpace_parameter": {
+            "__compat": {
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
👍

PR #17476, opened 2022-08-19
Add Safari versions for css.properties.font-family.system-ui

### LINKED ISSUE #17337: `system-ui` seems fully supported from iOS / Safari 11
https://github.com/mdn/browser-compat-data/blob/d6184383e1e6fccb1421c91ec478a2b65db45403/css/properties/font-family.json#L91-L99

caniuse lists version 11 as fully supported : https://caniuse.com/?search=system-ui

queengooborg:
This PR adds real values for Safari (Desktop and iOS/iPadOS) for the `system-ui` member of the `font-family` CSS property.  This updates the data to match CanIUse: https://caniuse.com/?search=system-ui (Fixes #17337)  Also, this changes the key from `system_ui` to `system-ui` to adhere to our data structure.

### DIFF
--- a/css/properties/font-family.json
+++ b/css/properties/font-family.json
@@ -80,19 +80,10 @@
                 },
                 {
                   "alternative_name": "-apple-system",
-                  "version_added": "9",
-                  "notes": "Supported since macOS 10.11."
-                }
-              ],
-              "safari_ios": [
-                {
-                  "version_added": "11"
-                },
-                {
-                  "alternative_name": "-apple-system",
                   "version_added": "9"
                 }
               ],
+              "safari_ios": "mirror",
               "samsunginternet_android": "mirror",
               "webview_android": "mirror"
             },
--- a/css/properties/font-family.json
+++ b/css/properties/font-family.json
@@ -46,7 +46,7 @@
             "deprecated": false
           }
         },
-        "system_ui": {
+        "system-ui": {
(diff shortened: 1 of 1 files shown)


css/properties/font-family.json
```diff
@@ -74,15 +74,25 @@
...
-                "version_added": "9"
-              },
+              "safari": [
+                {
+                  "version_added": "11"
+                },
+                {
+                  "alternative_name": "-apple-system",
+                  "version_added": "9",
+                  "notes": "Supported since macOS 10.11."
```
foolip, line 84: I don't think we should have notes like these. It's often the case that a feature depends on a new macOS release, but we don't say so in BCD.

Per https://en.wikipedia.org/wiki/Safari_version_history#Safari_9 Safari 9 was released at the same time as macOS 10.11, but also supported on 10.9 and 10.10.

foolip (changes requested):

css/properties/font-family.json
xeenon: Correct, this would have been supported on all OS releases that Safari 9 shipped with. Only certain rare cases does a feature lock with the OS version (usually image/video encoding formats).

foolip: Thanks for confirming, @xeenon!

foolip (approved):

PR #17481, opened 2022-08-19
Add missing api.AudioContext.createConstantSource feature

queengooborg:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing `createConstantSource` member of the AudioContext API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/AudioContext/createConstantSource

### DIFF
--- a/api/AudioContext.json
+++ b/api/AudioContext.json
@@ -294,6 +294,37 @@
           }
         }
       },
+      "createConstantSource": {
+        "__compat": {
+          "support": {
+            "chrome": {
+              "version_added": "56"
+            },
+            "chrome_android": "mirror",
+            "edge": "mirror",
+            "firefox": {
+              "version_added": "52"
+            },
+            "firefox_android": "mirror",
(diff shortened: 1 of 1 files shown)


api/AudioContext.json
```diff
@@ -294,6 +294,38 @@
...
+            "safari": {
+              "version_added": "14"
+            },
+            "safari_ios": "mirror",
+            "samsunginternet_android": "mirror",
+            "webview_android": "mirror"
+          },
+          "status": {
+            "experimental": false,
+            "standard_track": true,
```
foolip, line 324: I don't think so. This is already on `BaseAudioContext`.

foolip (changes requested):

PR #17483, opened 2022-08-19
Add missing api.AudioTrackList.item feature

queengooborg:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing `item` member of the AudioTrackList API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/AudioTrackList/item

### DIFF
--- a/api/AudioTrackList.json
+++ b/api/AudioTrackList.json
@@ -277,6 +277,7 @@
             "ie": {
               "version_added": false
             },
+            "oculus": "mirror",
             "opera": "mirror",
             "opera_android": "mirror",
             "safari": {
--- a/api/AudioTrackList.json
+++ b/api/AudioTrackList.json
@@ -259,6 +259,40 @@
           }
         }
       },
+      "item": {
+        "__compat": {
+          "support": {
+            "chrome": {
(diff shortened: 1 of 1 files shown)


api/AudioTrackList.json
```diff
@@ -259,6 +259,41 @@
...
+            "safari": {
+              "version_added": "7"
+            },
+            "safari_ios": "mirror",
+            "samsunginternet_android": "mirror",
+            "webview_android": "mirror"
+          },
+          "status": {
+            "experimental": false,
+            "standard_track": true,
```
foolip, line 292: It's not in https://html.spec.whatwg.org/multipage/media.html#audiotracklist-and-videotracklist-objects

foolip (changes requested):

queengooborg:
This PR is a good example of why https://github.com/foolip/mdn-bcd-collector/pull/1292 would be helpful!

PR #17489, opened 2022-08-19
Add missing CSS font-palette-values at-rule

queengooborg:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing `CSSFontPaletteValuesRule` API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/CSSFontPaletteValuesRule

### DIFF
--- a/api/CSSFontPaletteValuesRule.json
+++ b/api/CSSFontPaletteValuesRule.json
@@ -1,0 +1,164 @@
+{
+  "api": {
+    "CSSFontPaletteValuesRule": {
+      "__compat": {
+        "support": {
+          "chrome": {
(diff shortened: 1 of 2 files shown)


foolip (changes requested):
Looks good, but I think we should add the at-rule (https://drafts.csswg.org/css-fonts-4/#at-ruledef-font-palette-values) at the same time.

foolip (approved):

PR #17494, opened 2022-08-19
Add missing api.DataCue.data feature

queengooborg:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing `data` member of the DataCue API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/DataCue/data

### DIFF
--- a/api/DataCue.json
+++ b/api/DataCue.json
@@ -69,6 +69,38 @@
           }
         }
       },
+      "data": {
+        "__compat": {
+          "support": {
+            "chrome": {
+              "version_added": false
+            },
+            "chrome_android": "mirror",
+            "edge": "mirror",
+            "firefox": {
+              "version_added": false
+            },
+            "firefox_android": "mirror",
+            "ie": {
+              "version_added": false
+            },
+            "oculus": "mirror",
+            "opera": "mirror",
+            "opera_android": "mirror",
+            "safari": {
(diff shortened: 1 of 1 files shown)


api/DataCue.json
```diff
@@ -69,6 +69,38 @@
...
+            "safari": {
+              "version_added": "11"
+            },
+            "safari_ios": "mirror",
+            "samsunginternet_android": "mirror",
+            "webview_android": "mirror"
+          },
+          "status": {
+            "experimental": false,
+            "standard_track": true,
```
foolip, line 99: It's not in https://wicg.github.io/datacue/#datacue-interface.

queengooborg: Oops, closing.

foolip: In a case like this when it's in the only implementation of the API I think it might still make sense to add it, just correctly marked as non-standard.

queengooborg: That's a fair point -- for now though, I'd like to focus on adding standard track features, and then later look into the non-standard ones.  Standard features will be much easier to review!

PR #17496, opened 2022-08-19
Add missing Directory API

queengooborg:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing `Directory` API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Directory

### DIFF
--- a/test/linter/test-spec-urls.ts
+++ b/test/linter/test-spec-urls.ts
@@ -50,6 +50,8 @@ const specsExceptions = [
 
   // Remove if https://github.com/w3c/browser-specs/pull/667#issuecomment-1200089758 is resolved
   'https://w3c.github.io/device-memory',
+
+  'https://wicg.github.io/directory-upload/proposal.html',
 ];
 
 const allowedSpecURLs = [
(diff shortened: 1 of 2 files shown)


foolip (changes requested):
Can you add spec URLs? It'll be much easier to review then.

test/linter/test-spec-urls.ts
```diff
@@ -50,6 +50,8 @@ const specsExceptions = [
 
   // Remove if https://github.com/w3c/browser-specs/pull/667#issuecomment-1200089758 is resolved
   'https://w3c.github.io/device-memory',
+
+  'https://wicg.github.io/directory-upload/proposal.html',
```
foolip, line 54: This would need an issue filed. I'm not sure this spec is going anywhere though, how does it relate to https://fs.spec.whatwg.org/#api-filesystemdirectoryhandle?

queengooborg:
I now see that the repository was actually archived (https://github.com/wicg/directory-upload) so this should be removed from the custom IDL on the collector side -- I'll append this to https://github.com/foolip/mdn-bcd-collector/pull/1292!

PR #17497, opened 2022-08-19
Add missing features for Element API

queengooborg:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing features of the Element API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Element

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -2231,6 +2231,38 @@
           }
         }
       },
+      "attributeStyleMap": {
+        "__compat": {
+          "support": {
+            "chrome": {
+              "version_added": "66"
+            },
+            "chrome_android": "mirror",
+            "edge": "mirror",
+            "firefox": {
+              "version_added": false
+            },
+            "firefox_android": "mirror",
+            "ie": {
+              "version_added": false
+            },
+            "oculus": "mirror",
+            "opera": "mirror",
+            "opera_android": "mirror",
+            "safari": {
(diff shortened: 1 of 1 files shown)


api/Element.json
```diff
@@ -6623,6 +6655,38 @@
...
+            "safari": {
+              "version_added": false
+            },
+            "safari_ios": "mirror",
+            "samsunginternet_android": "mirror",
+            "webview_android": "mirror"
+          },
+          "status": {
+            "experimental": false,
+            "standard_track": true,
```
foolip, line 6685: Are you sure? Spec URL?

api/Element.json
```diff
@@ -2231,6 +2231,38 @@
           }
         }
       },
+      "attributeStyleMap": {
```
foolip, line 2234: This is already on HTMLElement. Where should it be?

queengooborg: Found this is in the `custom-idl` with a link to a resolved BCD issue, so I'll undo this!

api/Element.json
queengooborg: Looks like neither of these are standard, so closing.

PR #17500, opened 2022-08-19
Add missing api.FetchEvent.targetClientId feature

queengooborg:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing `targetClientId` member of the FetchEvent API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/FetchEvent/targetClientId

### DIFF
--- a/api/FetchEvent.json
+++ b/api/FetchEvent.json
@@ -432,6 +432,38 @@
             "deprecated": false
           }
         }
+      },
+      "targetClientId": {
+        "__compat": {
+          "support": {
+            "chrome": {
+              "version_added": false
+            },
+            "chrome_android": "mirror",
+            "edge": "mirror",
+            "firefox": {
+              "version_added": false
+            },
+            "firefox_android": "mirror",
+            "ie": {
+              "version_added": false
+            },
+            "oculus": "mirror",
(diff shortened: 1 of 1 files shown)


api/FetchEvent.json
```diff
@@ -432,6 +432,38 @@
...
+            },
+            "firefox_android": "mirror",
+            "ie": {
+              "version_added": false
+            },
+            "oculus": "mirror",
+            "opera": "mirror",
+            "opera_android": "mirror",
+            "safari": {
+              "version_added": "11.1"
```
foolip, line 455: This was removed in https://github.com/WebKit/WebKit/commit/f3f63d6b8f665f0d702ab30455778069ab879413, let's not add it.

foolip (changes requested):

api/FetchEvent.json
queengooborg: I see this is also set as an alternative name for `replacesClientId` as well.

PR #17503, opened 2022-08-20
Update spec URL for css.properties.contain.inline-size

sideshowbarker:
This should be the drafts.csswg.org URL rather than the www.w3.org/TR one.

### DIFF
--- a/css/properties/contain.json
+++ b/css/properties/contain.json
@@ -37,7 +37,7 @@
         "inline-size": {
           "__compat": {
             "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/contain#inline-size",
-            "spec_url": "https://www.w3.org/TR/css-contain-3/#valdef-contain-inline-size",
+            "spec_url": "https://drafts.csswg.org/css-contain-3/#valdef-contain-inline-size",
             "support": {
               "chrome": {
                 "version_added": "105"


queengooborg:
Hey @sideshowbarker!  I actually feel a lot less confident about adding draft CSSWG spec URLs to BCD now, due to how frequently the site goes down.  We've had users comment about how the URL is broken and it needs to be fixed.  I feel that if there is an applicable spec anywhere else besides drafts.csswg.org, then we should use it until the website is guaranteed not to go down any longer (or at least, nearly as frequently).

What do you think?

sideshowbarker:
I think we should keep using the drafts.w3.org URLs.

Doing otherwise doesn’t help get the drafts.w3.org instability problems fixed.

The drafts.w3.org versions are the only versions that are guaranteed to be up to date with the git sources.

The w3.org/TR URLs are not guaranteed to be up to date with the git sources and in general are often wildly outdated.

queengooborg:
While the W3 URLs may be out of date and not reflect the recent updates to the Git repo, they also don't go down nearly as frequently.  The CSSWG drafts website has been broken for over a year now, and every time I've personally tried to take a look at a CSSWG draft spec, it's been down.  Personally, when I'm working on BCD, I would rather refer to an older version of the spec, than not be able to refer to one at all.

I'm happy to merge this anyways and link to the CSSWG draft spec instead in the end.  I just want to voice my thoughts as an MDN reader and contributor.

foolip (approved):

foolip:
I'm also frustrated that the site is down so often, but also feel strongly that we shouldn't link to TR unless we absolutely have to.

@past FYI here's some impact of the instability out in the ecosystem. Is there any public info about the efforts to fix it?

queengooborg:
It looks like, based upon https://github.com/w3c/csswg-drafts/issues/6528#issuecomment-1221232095, work should be starting soon to finally fix this!  Earlier today (before I saw this), I reached out to Schalk earlier about redirecting links on MDN to the [backup server](https://andreubotella.com/csswg-auto-build/) while waiting for implementation.

Anyways, I don't want to be a blocker on this, so let's merge this PR!

PR #17532, opened 2022-08-23
WebExtension tweaks for Safari and Manifest V3

xeenon:
* Add better version specific notes for Safari support.
* Add Safari 16 version support for `runtime.getFrameId` and drop STP reference.
* Add more consistent manifest version availability notes.
    * Features that are dropped in V3 noted as "Manifest V2 only".
    * Features that are added in V3 noted as "Manifest V3 or later" to avoid prematurely limiting to V3 when future versions are supported.
    * Features that are dropped in V3 by some browsers but are supported in full in others still noted as "Manifest V2 or later".
* More consistent use of required/optional instead of mandatory.

### DIFF
--- a/webextensions/api/scripting.json
+++ b/webextensions/api/scripting.json
@@ -7,7 +7,8 @@
             "mdn_url": "https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/API/scripting/InjectionTarget",
             "support": {
               "chrome": {
-                "version_added": "88"
+                "version_added": "88",
+                "notes": "Available for use in Manifest V3 or later."
               },
               "edge": "mirror",
               "firefox": {
@@ -16,7 +17,8 @@
               "firefox_android": "mirror",
               "opera": "mirror",
               "safari": {
-                "version_added": "15.4"
+                "version_added": "15.4",
+                "notes": "Available for use in Manifest V2 or later."
               },
               "safari_ios": "mirror"
             }
--- a/webextensions/manifest/browser_specific_settings.json
+++ b/webextensions/manifest/browser_specific_settings.json
@@ -12,16 +12,10 @@
               "version_added": "15",
               "version_removed": "79"
             },
-            "firefox": [
-              {
-                "version_added": "48"
-              },
-              {
-                "version_added": "42",
-                "alternative_name": "applications",
-                "notes": "Mandatory before Firefox 48."
-              }
-            ],
+            "firefox": {
+              "version_added": "42",
+              "notes": "Before Firefox 48, this property was required."
+            },
             "firefox_android": "mirror",
             "opera": "mirror",
             "safari": {
--- a/webextensions/manifest/theme.json
+++ b/webextensions/manifest/theme.json
@@ -62,8 +62,8 @@
                   "version_added": "55",
                   "version_removed": "70",
                   "notes": [
-                    "Before version 59, the RGB array form was not supported for this property.",
-                    "Before version 63, this property was mandatory.",
+                    "Before Firefox 59, the RGB array form was not supported for this property.",
+                    "Before Firefox 63, this property was required.",
                     "Use <code>frame</code> instead."
                   ]
                 },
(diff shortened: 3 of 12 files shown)


xeenon:
@queengooborg let me know if this is okay, or if there are somethings I should do differently.

queengooborg:
Throughout MDN, the terms "required" and "optional" are used more frequently than "mandatory" -- could we keep using the term "required" and replace "not required"/"not mandatory" with optional?

webextensions/manifest/browser_specific_settings.json
```diff
@@ -12,16 +12,10 @@
               "version_added": "15",
               "version_removed": "79"
             },
-            "firefox": [
-              {
-                "version_added": "48"
-              },
-              {
-                "version_added": "42",
-                "alternative_name": "applications",
```
xeenon, line 21: `"alternative_name": "applications"` was a dummy entry to allow per-version notes here. I removed it to avoid confusion.

webextensions/manifest/browser_specific_settings.json
```diff
@@ -12,16 +12,10 @@
...
-                "notes": "Mandatory before Firefox 48."
-              }
-            ],
+            "firefox": {
+              "version_added": "42",
+              "notes": "Required before Firefox 48."
```
queengooborg, line 17:
```suggestion
              "notes": "Before Firefox 48, this property was required."
```

webextensions/manifest/theme.json
```diff
@@ -990,7 +990,7 @@
               },
               "firefox": {
                 "version_added": "55",
-                "notes": "Mandatory before Firefox 60"
+                "notes": "Required before Firefox 60."
```
queengooborg, line 993:
```suggestion
                "notes": "Before Firefox 60, this property was required."
```

queengooborg (changes requested):
Thanks for tackling this!  While we're here, let's also tweak some of the wording to follow the typical formula for notes:

queengooborg (approved):
Thank you, this is LGTM!

xeenon:
Rebased and ready to merge when you can, thanks @queengooborg!

queengooborg:
By the way, no need to rebase and force push for BCD PRs -- we squash merge all pull requests to this repository!

webextensions/api/scripting.json
```diff
@@ -7,7 +7,8 @@
             "mdn_url": "https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/API/scripting/InjectionTarget",
             "support": {
               "chrome": {
-                "version_added": "88"
+                "version_added": "88",
+                "notes": "Available for use in Manifest V3 or later."
```
sublimator, line 11: @xeenon 

I noted that Safari is missing scripting.registerContentScripts
```
  .registerContentScripts([
    {
      matches: ['https://*/*', 'http://*/*'],
      id: 'wm-polyfill.js',
      world: 'MAIN',
      js: ['wm-polyfill.js'],
      runAt: 'document_start'
    }
  ])
```

It's currently the only way to inject a polyfill in the Main context at document_start

xeenon: Yes, `registerContentScripts` was added to Chrome after Safari shipped v3 support. It is on our radar too add.

sublimator: @xeenon 
Great to hear! :) 
Thanks for the info

webextensions/manifest/browser_specific_settings.json
```diff
@@ -12,16 +12,10 @@
...
-              {
-                "version_added": "42",
-                "alternative_name": "applications",
-                "notes": "Mandatory before Firefox 48."
-              }
-            ],
+            "firefox": {
+              "version_added": "42",
+              "notes": "Before Firefox 48, this property was required."
+            },
```
willdurand, line 18: Why has this been changed? This breaks our tooling and it took me a while to figure that out since it is not mentioned anywhere in the changelog. This patch is for Safari, it would have been good to not change Firefox data or to move these changes to a separate issue/PR.

queengooborg: This PR is actually for Manifest V3 entirely as well, and if you look at the diff, you'll notice that this is simply consolidating the statements together to adhere to the typical structure of BCD (where a behavioral difference that only applies to one browser should be a single note, rather than two separated statements).

We apologize in advance for the changes causing an issue with your tooling; however, please do note that features may be added, removed or modified at any time to match developer interest, BCD data guidelines, or for better accuracy, as explained in the [Semantic versioning policy](https://github.com/mdn/browser-compat-data#Semantic-versioning-policy).  In other words, changes such as this are expected to happen in any release and will not result in any changelog notes.

PR #17545, opened 2022-08-24
Remove incorrect note about Firefox popstate event

wbamberg:
Fixes https://github.com/mdn/browser-compat-data/issues/17544.

As far as I can see this note is incorrect: https://stackoverflow.com/questions/6421769/popstate-on-pages-load-in-chrome and it is contradicted in the page.

### DIFF
--- a/api/Window.json
+++ b/api/Window.json
@@ -3770,8 +3770,7 @@
               "version_added": "12"
             },
             "firefox": {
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
Seems likely that this is wrong. Thanks! 👍

PR #17563, opened 2022-08-26
FF104 CSS Font Loading API supports workers

hamishwillee:
FF104 adds support for workers to CSS Font Loading API in https://bugzilla.mozilla.org/show_bug.cgi?id=1072107 behind pref layout.css.font-loading-api.workers.enabled

Other docs work can be tracked in https://github.com/mdn/content/issues/18769

FYI @queengooborg

### DIFF
--- a/api/FontFaceSetLoadEvent.json
+++ b/api/FontFaceSetLoadEvent.json
@@ -72,6 +72,48 @@
           }
         }
       },
+      "worker_support": {
+        "__compat": {
+          "description": "Available in workers",
+          "support": {
+            "chrome": {
+              "version_added": "69"
+            },
+            "chrome_android": "mirror",
+            "edge": "mirror",
+            "firefox": {
+              "version_added": "104",
+              "flags": [
+                {
+                  "type": "preference",
+                  "name": "layout.css.font-loading-api.workers.enabled",
+                  "value_to_set": "true"
+                }
+              ]
+            },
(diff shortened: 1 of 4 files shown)


api/FontFaceSetLoadEvent.json
```diff
@@ -72,6 +72,48 @@
           }
         }
       },
+      "worker_support": {
```
hamishwillee, line 75: Note, this interface indicates worker support in the spec: https://drafts.csswg.org/css-font-loading/#fontfacesetloadevent

I mirrored the BCD for chrome as it would all have been added at the same time.

Note, I did not add worker support for the events that were fired.

api/WorkerGlobalScope.json
```diff
@@ -126,6 +126,47 @@
           }
         }
       },
+      "fonts": {
```
hamishwillee, line 129: This is the new accessor for the API in Workers. I mirrored the Chrome support for this API in other docs.

queengooborg (approved):

api/FontFaceSetLoadEvent.json
caugner: FYI @hamishwillee It looks like mirroring BCD for Chrome was incorrect in this case (see [this PR](https://github.com/mdn/browser-compat-data/pull/24653)).

PR #17580, opened 2022-08-28
chore: use Dependabot v3 for GitHub Actions

nschonni:
#### Summary
Dependabot is already running for NPM dependencies, but noticed some of the core GitHub Action version were running v2 rather than v3 version, so this seems like a better way than submitting a bunch of separate PRs


queengooborg (approved):
This change looks good to me -- thank you for your PR, and welcome to BCD! 🎉

PR #17581, opened 2022-08-28
fix: various typos

nschonni:
#### Summary

Ran Vale to find duplicates, then cSpell for typos

### DIFF
--- a/RELEASE_NOTES.md
+++ b/RELEASE_NOTES.md
@@ -4427,7 +4427,7 @@ May 20, 2021
 
 ### Notable changes
 
-- The [definition of the `experimental` status](https://github.com/mdn/browser-compat-data/blob/main/schemas/compat-data-schema.md#status-information) has been revised to more closely reflect its actual use. For more information on choosing values for `experimental`, see the [the new data guideline for `experimental`](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines/index.md#choosing-an-experimental-status). ([#9933](https://github.com/mdn/browser-compat-data/pull/9933))
+- The [definition of the `experimental` status](https://github.com/mdn/browser-compat-data/blob/main/schemas/compat-data-schema.md#status-information) has been revised to more closely reflect its actual use. For more information on choosing values for `experimental`, see [the new data guideline for `experimental`](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines/index.md#choosing-an-experimental-status). ([#9933](https://github.com/mdn/browser-compat-data/pull/9933))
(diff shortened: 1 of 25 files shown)


RELEASE_NOTES.md
```diff
@@ -108,7 +108,7 @@ August 16, 2022
 - `browsers.webview_android.releases.108` ([#17373](https://github.com/mdn/browser-compat-data/pull/17373))
 - `browsers.webview_android.releases.109` ([#17373](https://github.com/mdn/browser-compat-data/pull/17373))
 - `browsers.webview_android.releases.110` ([#17373](https://github.com/mdn/browser-compat-data/pull/17373))
-- `css.types.image.paint.additional_paramters` ([#17314](https://github.com/mdn/browser-compat-data/pull/17314))
+- `css.types.image.paint.additional_parameters` ([#17314](https://github.com/mdn/browser-compat-data/pull/17314))
```
queengooborg, line 111: This is a feature identifier, so we shouldn't modify it.

```suggestion
- `css.types.image.paint.additional_paramters` ([#17314](https://github.com/mdn/browser-compat-data/pull/17314))
```

queengooborg (changes requested):
It feels awful seeing all the typos laid out in one PR like this... 😆  Thanks for fixing them up!

There's one instance that should remain a typo because it's a feature identifier, but otherwise LGTM!

RELEASE_NOTES.md
nschonni: I believe I change that one in the JSON file too. Since it was only a few weeks old, I thought it might have been unintentional

queengooborg: It was definitely an unintentional typo, however we still added it in with a typo.  When the next release comes around, its release notes will clarify that the typo version had been removed in favor of the fixed one!

nschonni: OK, but should the fix in the JSON file stay?
I'll probably ignore the RELEASE_NOTES.md file from the typo checking going forward anyway

queengooborg: Yes, we should definitely change the identifier in the actual JSON file!  The reason we're not fixing the same typo in the release notes is because we want to represent that the feature was added with the broken name initially.  When this PR is merged to fix the name, the release notes will show that the identifier with the typo was removed and the fixed identifier was added!

I imagine that the release notes will still have potential for typos, but I can say with certainty that anything in the release notes in inline code (`foo`) should be ignored, if you're able to do that?

queengooborg (approved):
Thank you for doing this, this is all LGTM!

PR #17582, opened 2022-08-28
ci: use setup-node NPM cache

nschonni:
#### Summary

actions/setup-node now wraps actions/cache to allow caching of node_modules between builds


queengooborg (approved):
Should help give the workflows a slight performance boost, thanks!

Issue #16163, opened 2022-05-07
Use Markdown for notes and descriptions?


queengooborg:
I think it would be beneficial to use Markdown for our notes and descriptions, rather than HTML.  Currently, we're spending a lot of resources (see #16162) to validate the HTML within the notes/descriptions to ensure that there is no HTML we don't want, such as scripts, headers, etc.  However, Markdown already has a limited set of potential elements which, when accounting for the fact that notes cannot contain newlines, line up perfectly with our selection.  Perhaps we should use Markdown for notes and descriptions rather than HTML?

One way that we can keep this change non-breaking is by adding an HTML-generating step in the build process, so consumers will still have HTML-formatted notes, but will increase consumer-contributor inconsistency which may not be desirable.

PR #17592, opened 2022-08-28
Add Markdownlint

nschonni:
#### Summary

Prettier is already running on the repo to do the basic formatting, but this can catch additional unintended formatting issues.


queengooborg:
Hey @nschonni!  I'm not sure if I/we want to add Markdownlint to this particular project, but if we do, there are some changes being made that I don't like.  Can we revert the file changes and just keep the linter addition for now?  (I understand it will cause test failures, but that's alright: it helps show the linter is working!)

nschonni:
I pulled out the changes, but I didn't add the problem matcher here, since like you said, there is less markdown in this repo than than mdn/content

queengooborg:
Hey again!  So after disabling rules MD026, MD034 and MD036, the only tweaks that are left are the ones I just covered in #17612 and some missing alt. text for images in a document we're looking to rewrite anyways.  With this in mind, I don't think that we should introduce a brand new dependency for what will only catch a few minor problems.  I'll be asking the rest of the BCD owners what they think during the meeting tomorrow, and will let you know what we decide!

queengooborg:
I talked about this PR during the BCD meeting earlier today, and we decided that adding Markdownlint isn't really important to us in BCD's current stage.  We have very little Markdown in the repository, so just sticking with Prettier is sufficient enough for our needs.

We've been considering converting all of our notes from HTML to Markdown, see https://github.com/mdn/browser-compat-data/issues/16163.  If we decide to go for it, then I think introducing some Markdown linters would be great, but for now, it seems like little benefit for introducing a new dependency.  Thank you anyways!

PR #17594, opened 2022-08-28
Format .github folder

nschonni:
#### Summary

- Remove empty arrays
- Remove unessessary quoting
- Rename and remove default values from Node.js action template

### DIFF
--- a/.github/workflows/node.js.yml
+++ b/.github/workflows/node.js.yml
@@ -1,13 +1,8 @@
-# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
-# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
-
 name: Node.js CI
 
 on:
   push:
-    branches: [] # Test on push to every branch
   pull_request:
-    branches: [] # Test on push to every branch
 
 permissions:
   contents: read
--- a/.github/ISSUE_TEMPLATE/data-problem.yml
+++ b/.github/ISSUE_TEMPLATE/data-problem.yml
@@ -16,9 +16,9 @@ body:
 
         **Need help with a browser?**
         🙋 To get help with [Firefox](https://support.mozilla.org/en-US/kb/file-bug-report-or-feature-request-mozilla), [Chrome](https://support.google.com/chrome/answer/95315?hl=en-GB&ref_topic=7439544), [Safari](https://www.apple.com/feedback/safari.html), or another browser, check the browser's support site.
-        
+
         ### Issue etiquette
-        
+
         When opening an issue, please:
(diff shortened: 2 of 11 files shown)


.github/workflows/node.js.yml
```diff
@@ -1,25 +0,0 @@
-# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
```
queengooborg, line 1: Why is this file getting deleted?

queengooborg: Ah, I see why now -- let's not rename this file, please!

nschonni: Do you have an alternate suggestion? `node.js.yml` is an odd default that comes from the Actions UI wizard

queengooborg: Let's think about a new name at a later time.  The filename doesn't make much of a difference, so I'd rather keep the filename the same for now to retain the Git history!

.github/workflows/node.js.yml
```diff
@@ -17,7 +12,7 @@ jobs:
       - uses: actions/checkout@v3
       - uses: actions/setup-node@v3
         with:
-          node-version: 16
+          node-version: lts/*
```
nschonni, line 15: Changed this because of the line above `name: Active LTS`, but I'm thinking that might just be more boilerplate from the GitHub Actions wizard, rather than an intentional name

.github/workflows/node.js.yml
nschonni: There doesn't really seem to be any important history to retain for the file https://github.com/mdn/browser-compat-data/commits/main/.github/workflows/node.js.yml but I've reverted the name to avoid this getting stalled with bikeshedding the name

.github/workflows/node.js.yml
queengooborg: Let's keep this hard-coded as "16" -- I've found that it's better for us to bump the Node.js version ourselves!

.github/workflows/node.js.yml
```diff
@@ -1,13 +1,8 @@
-# This workflow will do a clean install of node dependencies, build the source code and run tests across different versions of node
-# For more information see: https://help.github.com/actions/language-and-framework-guides/using-nodejs-with-github-actions
-
-name: Node.js CI
+name: CI
```
queengooborg, line 1: Let's keep the name the same just for continuity and to keep the scope of this PR solely to formatting changes!

nschonni:
@queengooborg I've reverted those renames, but updated the prettier config so it'll apply to workflow files if the other glob PR lands.

queengooborg (approved):
Formatting seems good to me -- we should probably be running Prettier on these files as it is.  Thank you!

Issue #12760, opened 2021-10-09
css.properties.outline - In Chrome 94, outline does follow the shape of border-radius.


HM100:
#### What information was incorrect, unhelpful, or incomplete?

Does not mention that in Chrome 94, outline does follow the shape of border-radius.

#### What did you expect to see?

I've should see something like this: ``Before Chrome 93, outline does not follow the shape of border-radius.``

#### Did you test this? If so, how?

I tested it on the example of https://developer.mozilla.org/en-US/docs/Web/CSS/outline

<details>
<summary>MDN page report details</summary>

* Query: `css.properties.outline`
* MDN URL: https://developer.mozilla.org/en-US/docs/Web/CSS/outline
* Report started: 2021-10-09T18:14:12.613Z

</details>

PR #17598, opened 2022-08-29
Consider `outline` not following border-radius a partial implementation

cpmsmith:
#### Summary
This information is currently in the data, but as a note. It's not tagged to specific versions, either, which means when scanning, for instance, the browser compatibility table on MDN, the Safari column looks the same as browsers which have implemented this. Now that the [spec says this should be done](https://www.w3.org/TR/css-ui-3/#ref-for-border-edge), I think it's reasonable to consider it a partial implementation if it isn't done.

image

image

The WebKit bug in the note got marked as a duplicate of [this bug](https://bugs.webkit.org/show_bug.cgi?id=20807), so I updated the link.

Also note that on the browsers that have fixed it, I didn't put a `version_removed` property on the partial implementations, only a newer support statement. That seems to be valid, and I'm not totally clear on the semantics of having it vs. not having it, so let me know if I should add it.

#### Related issues
Related to mdn/browser-compat-data#12760, Fyrd/caniuse#6099

### DIFF
--- a/css/properties/outline.json
+++ b/css/properties/outline.json
@@ -12,6 +12,7 @@
               },
               {
                 "version_added": "1",
+                "version_removed": "94",
                 "partial_implementation": true,
                 "notes": "Before Chrome 94, <code>outline</code> does not follow the shape of <code>border-radius</code>."
               }
@@ -23,6 +24,7 @@
               },
               {
                 "version_added": "12",
+                "version_removed": "94",
                 "partial_implementation": true,
                 "notes": "Before Edge 94, <code>outline</code> does not follow the shape of <code>border-radius</code>."
               }
@@ -33,6 +35,7 @@
               },
               {
                 "version_added": "1.5",
+                "version_removed": "88",
                 "partial_implementation": true,
                 "notes": "Before Firefox 88, <code>outline</code> does not follow the shape of <code>border-radius</code>."
               },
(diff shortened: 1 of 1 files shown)


css/properties/outline.json
```diff
@@ -6,18 +6,34 @@
...
-            "chrome": {
-              "version_added": "1",
-              "notes": "Before Chrome 94, <code>outline</code> does not follow the shape of <code>border-radius</code>."
-            },
+            "chrome": [
+              {
+                "version_added": "94"
+              },
+              {
+                "version_added": "1",
```
queengooborg, line 14:
```suggestion
                "version_added": "1",
                "version_removed": "94",
```

queengooborg (changes requested):
Hi there, thank you for your PR and welcome to BCD!

I'm down for converting this from a note to `partial_implementation`; it's a pretty severe caveat.  We should also add a `version_removed` for when the partial implementation no longer applies.

css/properties/outline.json
cpmsmith: Done!

queengooborg (approved):
Thank you, LGTM!

kylemh:
Hello! This needs to be updated to show just for versions below 16.4: https://webkit.org/blog/13966/webkit-features-in-safari-16-4/#:~:text=Outline%20%2B%20Border%20Radius

cpmsmith:
> Hello! This should be updated to show just for versions below 16.4: https://webkit.org/blog/13966/webkit-features-in-safari-16-4/#:~:text=Outline%20%2B%20Border%20Radius

Good point—opened #20424

PR #17619, opened 2022-08-31
Adds the @container rule

rachelandrew:
Shipped in Chrome 105, and also in Safari 16 beta,

### DIFF
--- a/css/at-rules/container.json
+++ b/css/at-rules/container.json
@@ -1,0 +1,41 @@
+{
+  "css": {
+    "at-rules": {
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
Confirmed with https://chromestatus.com/feature/6525308435955712 (although there is a comment that it will work better in Chrome 106 because "Table fragmentation" isn't shipped in 105) and https://developer.apple.com/documentation/safari-release-notes/safari-16-release-notes.

PR #17624, opened 2022-08-31
Remove compat data for deprecated math@mode attribute

fred-wang:
#### Summary

Remove compat data for deprecated math@mode attribute
This was only implemented in Gecko and removed in version 70.
Currently, we state support on Firefox Android, which is incorrect.

See below.

#### Related issues

https://bugzilla.mozilla.org/show_bug.cgi?id=1573438#c10
https://github.com/mdn/browser-compat-data/pull/5047

### DIFF
--- a/mathml/elements/math.json
+++ b/mathml/elements/math.json
@@ -130,43 +130,6 @@
               "deprecated": false
             }
           }
-        },
-        "mode": {
-          "__compat": {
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):

fred-wang:
Thanks!

PR #17625, opened 2022-08-31
Add Husky for pre-commit formatting hooks

queengooborg:
This PR adds `husky` as a dependency for pre-commit hooks to format all of the files before creating a commit.


queengooborg:
That's the eventual plan!

.lintstagedrc
```diff
@@ -0,0 +1,5 @@
+{
+    "*": "prettier --ignore-unknown --write",
+    "*.{js,jsx,tsx}": "eslint --fix",
+    "*.{json}": "npm run fix"
```
nschonni, line 4: Prettier stuff will already be run on the wildcard
```suggestion
    "*.{json}": "ts-node scripts/fix/index.ts"
```

Elchi3 (commented):
Let's ask in the next owner's meeting about pre-commit hooks. I'm not against it per se, just want to make sure the majority appreciates these ergonomics. Also, might be nice to mention this in the contribution docs if we introduce them.

queengooborg:
MDN content just landed this, so let's land this here as well!

PR #17626, opened 2022-08-31
Fix spec_url of margin-block-start

teoli2003:
This was reported in https://github.com/mdn/content/discussions/20175.

The spec_url was completely wrong (to scroll-snap).


Elchi3 (approved):
Thanks 👍

PR #17631, opened 2022-08-31
chore: add cSpell for spellchecking

nschonni:
#### Summary

- Generated a custom TXT based dictionary rather than using the built-in `words` array for performance reasons.
- Excluded a few instances of `Analyser` that are baked into the API and can't be changed, but shouldn't be used again.
- Add a CI workflow and problem matcher
- Add a script to recreate the dictionary to shrink it down as some of the words are upstreamed to cspell-dicts as appropriate


package.json
```diff
@@ -82,6 +83,8 @@
     "release": "ts-node scripts/release/index.ts",
     "remove-redundant-flags": "ts-node scripts/remove-redundant-flags.ts",
     "show-errors": "npm test 1> /dev/null",
+    "spelling": "cspell \"**\"",
+    "spelling:regen": " cspell --no-progress --words-only --unique \"**\" > project-words.txt",
```
nschonni, line 87: This leaves the file unsorted, so I've just been resorting with VSCode for now

PR #17638, opened 2022-09-01
MathML: Mark 'maction' deprecated

OnkarRuikar:
It has been deprecated.

#### Supporting details
- https://developer.mozilla.org/en-US/docs/Web/MathML/Element/maction
- https://github.com/mdn/content/pull/18650
- https://github.com/w3c/mathml-core/pull/25

### DIFF
--- a/mathml/elements/maction.json
+++ b/mathml/elements/maction.json
@@ -76,7 +76,7 @@
             "status": {
               "experimental": false,
               "standard_track": true,
-              "deprecated": false
+              "deprecated": true
             }
           },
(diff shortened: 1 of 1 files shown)


Elchi3 (commented):
Should we mark everything in this file as deprecated?
 cc @fred-wang

fred-wang:
Yes, sounds good to me.

Elchi3 (approved):
Thanks @OnkarRuikar and @fred-wang 👍

PR #17641, opened 2022-09-01
Correcting and updating WCO compat data

captainbrosset:
#### Summary

The **Window Controls Overlay** feature is useful for Progressive Web Apps to take advantage of the full window area when installed on desktop.

This was documented on MDN and BCD about a year ago and the chrome/edge version numbers set at the time were 98. Unfortunately the feature did not ship with 98. It was available only behind a flag at that time, and remained like this until 105 (which just shipped for Chrome, and is about to ship for Edge).

This PR updates the 98 version numbers to 105, for the corresponding APIs (navigator.windowControlsOverlay) and removes the `experimental` flag since the feature is now stable on those browsers (it went through origin trial already and the spec hasn't changed for about a year, and is now shipping on-by-default).

I tested today with Chrome 105 (I made sure the flag was set to its default value). To test it yourself: open Chrome 105, go to an app that uses the feature: https://microsoftedge.github.io/Demos/pwamp/ install the app, and click the "hide title bar" button in the title bar (chevron icon).

See the chrome 105 release notes: https://developer.chrome.com/blog/new-in-chrome-105/#more

### DIFF
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -4658,7 +4658,7 @@
           "spec_url": "https://wicg.github.io/window-controls-overlay/#windowcontrolsoverlay-interface",
           "support": {
             "chrome": {
-              "version_added": "98"
+              "version_added": "105"
             },
             "chrome_android": {
               "version_added": false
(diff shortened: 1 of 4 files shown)


Elchi3 (approved):
Thank you! 👍

PR #17651, opened 2022-09-02
Add missing api.HTMLElement.HTMLElement feature

queengooborg:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing `HTMLElement` member of the HTMLElement API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.0).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/HTMLElement/HTMLElement

### DIFF
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -41,6 +41,40 @@
           "deprecated": false
         }
       },
+      "HTMLElement": {
+        "__compat": {
+          "description": "<code>HTMLElement()</code> constructor",
+          "spec_url": "https://html.spec.whatwg.org/multipage/dom.html#htmlelement",
+          "support": {
+            "chrome": {
+              "version_added": false
+            },
+            "chrome_android": "mirror",
+            "edge": "mirror",
+            "firefox": {
+              "version_added": false
+            },
+            "firefox_android": "mirror",
+            "ie": {
+              "version_added": false
+            },
+            "oculus": "mirror",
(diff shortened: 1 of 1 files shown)


foolip (commented):
Have you tried using this constructor to verify it works as advertised? It's somewhat surprising it's been in Safari for so long but not Firefox or Chrome.

Can you look for bugs to link for this?

lucalves (approved):

queengooborg:
Running `new HTMLElement()` throws `TypeError: new.target is not a valid custom element constructor`.  I don't know enough about this constructor to determine if that error is fine or not.

foolip:
That error is from here:

https://github.com/WebKit/WebKit/blob/ee1be476e83b744d0d91887b48953cb2552fcfc0/Source/WebCore/bindings/js/JSHTMLElementCustom.cpp#L63-L64

So it looks like `HTMLElement` is special cased to *not* work as a constructor. This is all about this bit in the spec:
https://html.spec.whatwg.org/multipage/dom.html#html-element-constructors

So I don't think we should have these in BCD, and actually shouldn't generate tests for them in the collector either. I'll make a PR to remove these tests.

foolip:
https://mdn-bcd-collector.appspot.com/tests/api/HTMLElement/HTMLElement is no more!

PR #17662, opened 2022-09-02
Add compat data for font-family: math

fred-wang:
This is implemented in Chromium on non-Android platforms since version 103,
under the experimental web platform features flag. Not in any other web
engine.

Spec:
https://drafts.csswg.org/css-fonts/#valdef-font-family-math

Chromium:
https://chromiumdash.appspot.com/commit/52ec74ca677b34eff97085592dc2e1e7cb8c1749
https://docs.google.com/document/d/1bvY_Npe2zLW_705KXdmecsH6P9I9wBMpHRsZ9CxWNOI/edit

Firefox:
https://bugzilla.mozilla.org/show_bug.cgi?id=1788937

WebKit:
https://bugs.webkit.org/show_bug.cgi?id=209595

### DIFF
--- a/css/properties/font-family.json
+++ b/css/properties/font-family.json
@@ -72,19 +72,8 @@
                 "version_added": false
               },
               "oculus": "mirror",
-              "opera": {
-                "version_added": "89",
-                "flags": [
-                  {
-                    "type": "preference",
-                    "name": "#enable-experimental-web-platform-features",
-                    "value_to_set": "Enabled"
-                  }
-                ]
-              },
-              "opera_android": {
-                "version_added": false
-              },
+              "opera": "mirror",
+              "opera_android": "mirror",
               "safari": {
                 "version_added": false
               },
--- a/css/properties/font-family.json
+++ b/css/properties/font-family.json
@@ -46,6 +46,59 @@
             "deprecated": false
           }
         },
+        "math": {
+          "__compat": {
+            "description": "<code>math</code>",
+            "support": {
+              "chrome": {
+                "version_added": "103",
+                "flags": [
+                  {
+                    "type": "preference",
+                    "name": "#enable-experimental-web-platform-features",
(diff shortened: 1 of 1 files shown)


fred-wang:
cc @Elchi3 @queengooborg

css/properties/font-family.json
```diff
@@ -46,6 +46,59 @@
...
+                  {
+                    "type": "preference",
+                    "name": "#enable-experimental-web-platform-features",
+                    "value_to_set": "Enabled"
+                  }
+                ]
+              },
+              "opera_android": {
+                "version_added": false
+              },
```
queengooborg, line 87:
```suggestion
              "opera": "mirror",
              "opera_android": "mirror",
```

css/properties/font-family.json
```diff
@@ -46,6 +46,59 @@
...
+                  {
+                    "type": "preference",
+                    "name": "#enable-experimental-web-platform-features",
+                    "value_to_set": "Enabled"
+                  }
+                ]
+              },
+              "chrome_android": {
+                "version_added": false
+              },
```
queengooborg, line 65:
```suggestion
              "chrome_android": "mirror",
```

queengooborg (approved):

queengooborg (changes requested):
Looks good, though we can mirror the other Chromium browsers!

css/properties/font-family.json
fred-wang: As explained on https://docs.google.com/document/d/1bvY_Npe2zLW_705KXdmecsH6P9I9wBMpHRsZ9CxWNO on Blink generally does not use font-family keywords for font selection (it only has a few special handlings) and internal font settings are not used. However parsing, serialization, etc works as expected. Note that this limitation is not specific to `font-family: math` though, so I guess let's merge this...

queengooborg (approved):

PR #17675, opened 2022-09-04
Add missing api.WebGL[2]RenderingContext.[drawingBuffer/unpack]ColorSpace feature

queengooborg:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing `drawingBufferColorSpace` and `unpackColorSpace` members of the WebGLRenderingContext 1 and 2 APIs, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.2).

Tests Used:
https://mdn-bcd-collector.appspot.com/tests/api/WebGLRenderingContext/drawingBufferColorSpace
https://mdn-bcd-collector.appspot.com/tests/api/WebGLRenderingContext/unpackColorSpace
https://mdn-bcd-collector.appspot.com/tests/api/WebGL2RenderingContext/drawingBufferColorSpace
https://mdn-bcd-collector.appspot.com/tests/api/WebGL2RenderingContext/unpackColorSpace

### DIFF
--- a/api/WebGL2RenderingContext.json
+++ b/api/WebGL2RenderingContext.json
@@ -3127,6 +3127,39 @@
           }
         }
       },
+      "drawingBufferColorSpace": {
+        "__compat": {
+          "spec_url": "https://registry.khronos.org/webgl/specs/latest/1.0/#DOM-WebGLRenderingContext-drawingBufferColorSpace",
+          "support": {
+            "chrome": {
+              "version_added": "104"
+            },
+            "chrome_android": "mirror",
+            "edge": "mirror",
+            "firefox": {
+              "version_added": false
+            },
+            "firefox_android": "mirror",
+            "ie": {
+              "version_added": false
+            },
+            "oculus": "mirror",
+            "opera": "mirror",
+            "opera_android": "mirror",
+            "safari": {
+              "version_added": false
+            },
+            "safari_ios": "mirror",
+            "samsunginternet_android": "mirror",
+            "webview_android": "mirror"
+          },
+          "status": {
+            "experimental": true,
+            "standard_track": true,
+            "deprecated": false
+          }
+        }
+      },
       "drawingBufferHeight": {
         "__compat": {
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/WebGLRenderingContext/drawingBufferHeight",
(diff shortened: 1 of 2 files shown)


api/WebGL2RenderingContext.json
```diff
@@ -3127,6 +3127,39 @@
           }
         }
       },
+      "drawingBufferColorSpace": {
+        "__compat": {
+          "spec_url": "https://registry.khronos.org/webgl/specs/latest/1.0/#DOM-WebGLRenderingContext-drawingBufferColorSpace",
```
foolip, line 3132: Almost all other spec URLs in this file are to numbered sections. Can you match that?

foolip (approved):
104 confirmed by https://storage.googleapis.com/chromium-find-releases-static/e6b.html#e6bc3505c69a5a758adc3651dfda29d1e72a4696

api/WebGL2RenderingContext.json
queengooborg: There's actually a number of spec URLs that point to places like this, and I feel that's better because it points to the exact feature. We should probably change the numbered ones accordingly.

foolip: The numbered sections style is still completely dominant. I'd suggest sticking to that, and then opening a PR to change everything to your preferred style, asking review from whoever added the current spec URLs.

queengooborg: I decided to look into this, and I figured out why there's a lot of numbered sections used rather than the exact method or property.  It's because the link to the section is the closest available!  Features like `activeTexture` or `attachShader` don't have named anchors that point directly to them, so the section header has to be used in its place.

foolip: I see, so it's the exception that there's a direct link. In fact, just 11 cases like this one. `Array.from(document.querySelectorAll('[id]')).map(e => e.id).filter(id => id.startsWith('DOM'))` gives:

```
DOM-WebGLActiveInfo-size
DOM-WebGLActiveInfo-type
DOM-WebGLActiveInfo-name
DOM-WebGLShaderPrecisionFormat-rangeMin
DOM-WebGLShaderPrecisionFormat-rangeMax
DOM-WebGLShaderPrecisionFormat-precision
DOM-WebGLRenderingContext-canvas
DOM-WebGLRenderingContext-drawingBufferWidth
DOM-WebGLRenderingContext-drawingBufferHeight
DOM-WebGLRenderingContext-drawingBufferColorSpace
DOM-WebGLRenderingContext-unpackColorSpace
```

If you want to use these IDs where available, can you submit a PR updating the rest?

foolip (approved):

api/WebGL2RenderingContext.json
queengooborg: Just did a check, and all of the other features have the appropriate URLs!

foolip: Cool! I already approved, I'll merge as well now then.

PR #17679, opened 2022-09-04
Add missing FontData API

queengooborg:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing `FontData` API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.2).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/FontData

### DIFF
--- a/api/FontData.json
+++ b/api/FontData.json
@@ -1,0 +1,202 @@
+{
+  "api": {
+    "FontData": {
+      "__compat": {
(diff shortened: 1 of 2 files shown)


foolip:
This depends on `queryLocalFonts`, which should be shipped at the same time, can you add that as well to this PR?

foolip (approved):

PR #17692, opened 2022-09-05
Add compat data for the math-shift property

fred-wang:
#### Summary

Add compat data for the math-shift property
It's implemented in Chrome 87 under the "experimental web platform features" flag
but not in other browsers.

Chrome:
https://chromiumdash.appspot.com/commit/c82a21d4e24e53d4adf915a6c1f27005db1b5850

### DIFF
--- a/css/properties/math-shift.json
+++ b/css/properties/math-shift.json
@@ -1,0 +1,47 @@
+{
+  "css": {
+    "properties": {
+      "math-shift": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/math-shift",
+          "spec_url": "https://w3c.github.io/mathml-core/#the-math-shift",
+          "support": {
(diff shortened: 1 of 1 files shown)


fred-wang:
cc @Elchi3 

Note: the MDN page does not exist yet, will upload a separate PR

Elchi3 (approved):
Looks good to me but I like to double check with @queengooborg about what our latest policy is when a feature doesn't ship anywhere yet. 
Vinyl, is this good to get in or should it have an unflagged implementation somewhere before we accept it? (experimental is set to true here, so that's good, too)

fred-wang:
FWIW, I believe the same happens for `math-depth`.

queengooborg (approved):
This looks good to me -- as long as some browser has some implementation of the feature, whether behind a flag or enabled by default, the feature meets minimum requirements for inclusion!

PR #17695, opened 2022-09-05
Remove useless line

MrBrain295:
#### Summary
It looks to be copied from content.


queengooborg:
I think that this line is still relevant -- after submitting a PR, one should review the output of the CI checks and make changes as needed.  I don't think that we should remove this line from the PR template.

MrBrain295:
Oh, sorry I thought it was copied from content and was encouraging people to check the formatting of the check boxes. But yeah that makes sense.

PR #17700, opened 2022-09-06
Update Safari versions for AudioBufferSourceNode API

queengooborg:
This PR updates and corrects the real values for Safari (Desktop and iOS/iPadOS) for the `AudioBufferSourceNode` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.3).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/AudioBufferSourceNode

### DIFF
--- a/api/AudioBufferSourceNode.json
+++ b/api/AudioBufferSourceNode.json
@@ -218,7 +218,7 @@
             "opera": "mirror",
             "opera_android": "mirror",
             "safari": {
-              "version_added": "6"
+              "version_added": "7"
             },
             "safari_ios": "mirror",
             "samsunginternet_android": "mirror",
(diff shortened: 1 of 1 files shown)


foolip:
Which exact Safari versions were tested here? I didn't think we have the right versions available to distinguish between 6 and 7.

queengooborg:
I've tested with Safari 6, 7 and 8 -- I got my old laptop set up for virtual machines again and have run older Safari, Chrome and Firefox versions to collect their results!

foolip:
Cool, I see that https://github.com/foolip/mdn-bcd-results has 6.0 and 7.0 currently, so this should be reliable.

foolip (approved):

queengooborg:
By the way, if we get results for any of the Safari backported releases (6.1/6.2/7.1), the collector will ignore them and not use them in any BCD updates)!

foolip:
Oh that's right, we check `inBcd` and don't use the data at all. That's good!

PR #17707, opened 2022-09-06
Move api.Element.openOrClosedShadowRoot to webext/

### LINKED ISSUE #16986: api.Element.openOrClosedShadowRoot - Chrome 88+ has it
### What type of issue is this?

Incorrect support data (ex. Chrome says "86" but support was added in "40")

### What information was incorrect, unhelpful, or incomplete?

[Chrome 88+ has it as Web extension API](https://developer.chrome.com/docs/extensions/reference/dom/)

### What browsers does this problem apply to, if applicable?

Chromium (Chrome, Edge 79+, Opera, Samsung Internet)

### What did you expect to see?

Same as FIrefox, but with version 88 and the wild char to explain is extension only.

### Did you test this? If so, how?

No, but I am sure who wrote [the extension API](https://developer.chrome.com/docs/extensions/reference/dom/) did.

https://developer.chrome.com/docs/extensions/reference/dom/

### Do you have anything more you want to share?

_No response_

### MDN URL

https://developer.mozilla.org/en-US/docs/Web/API/Element/openOrClosedShadowRoot

### MDN metadata

<details>

* Query: `api.Element.openOrClosedShadowRoot`
* Report started: 2022-07-14T10:14:57.556Z

</details>

queengooborg:
This PR moves `api.Element.openOrClosedShadowRoot` to `webextensions.api.dom.openOrClosedShadowRoot` to better match where this function can be used.  The placement was chosen based on Chrome's implementation (see https://github.com/mdn/browser-compat-data/pull/17414#issuecomment-1220022812).

Fixes #16986.

### DIFF
--- a/webextensions/api/dom.json
+++ b/webextensions/api/dom.json
@@ -7,7 +7,7 @@
             "mdn_url": "https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/API/dom/openOrClosedShadowRoot",
             "support": {
               "chrome": {
-                "version_added": false
+                "version_added": "88"
               },
               "edge": "mirror",
               "firefox": {
--- a/webextensions/api/dom.json
+++ b/webextensions/api/dom.json
@@ -9,29 +9,16 @@
               "chrome": {
                 "version_added": false
               },
-              "chrome_android": "mirror",
               "edge": "mirror",
               "firefox": {
                 "version_added": "63"
               },
               "firefox_android": "mirror",
-              "ie": {
-                "version_added": false
-              },
-              "oculus": "mirror",
               "opera": "mirror",
-              "opera_android": "mirror",
               "safari": {
                 "version_added": false
               },
-              "safari_ios": "mirror",
-              "samsunginternet_android": "mirror",
-              "webview_android": "mirror"
-            },
-            "status": {
-              "experimental": false,
-              "standard_track": false,
-              "deprecated": false
+              "safari_ios": "mirror"
             }
           }
         }
--- a/webextensions/api/dom.json
+++ b/webextensions/api/dom.json
@@ -1,0 +1,41 @@
+{
+  "webextensions": {
+    "api": {
+      "dom": {
+        "openOrClosedShadowRoot": {
+          "__compat": {
+            "mdn_url": "https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/API/dom/openOrClosedShadowRoot",
+            "support": {
+              "chrome": {
+                "version_added": false
+              },
+              "chrome_android": "mirror",
+              "edge": "mirror",
+              "firefox": {
+                "version_added": "63"
+              },
+              "firefox_android": "mirror",
+              "ie": {
+                "version_added": false
+              },
+              "oculus": "mirror",
+              "opera": "mirror",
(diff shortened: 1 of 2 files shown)


webextensions/api/dom.json
```diff
@@ -0,0 +1,28 @@
+{
+  "webextensions": {
+    "api": {
+      "dom": {
+        "openOrClosedShadowRoot": {
+          "__compat": {
+            "mdn_url": "https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/API/dom/openOrClosedShadowRoot",
+            "support": {
+              "chrome": {
+                "version_added": false
```
Rob--W, line 10: Supported since Chrome 88 - https://bugs.chromium.org/p/chromium/issues/detail?id=778816#c29

Which means that Opera 74 supports this too (https://en.wikipedia.org/wiki/History_of_the_Opera_web_browser#Opera_2021)

and Edge 88 supports this method too.

webextensions/api/dom.json
```diff
@@ -0,0 +1,28 @@
+{
+  "webextensions": {
+    "api": {
+      "dom": {
+        "openOrClosedShadowRoot": {
+          "__compat": {
+            "mdn_url": "https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/API/dom/openOrClosedShadowRoot",
```
Rob--W, line 7: Is there a mdn/content ticket that tracks the documentation?

webextensions/api/dom.json
```diff
@@ -0,0 +1,28 @@
...
+            "support": {
+              "chrome": {
+                "version_added": false
+              },
+              "edge": "mirror",
+              "firefox": {
+                "version_added": "63"
```
Rob--W, line 14: Firefox does not support the `dom` API namespace.

But it does support the `element.openOrClosedShadowRoot` property.

Rob--W (changes requested):

webextensions/api/dom.json
queengooborg: This PR will act as the tracker for the document move with the `needs content update` label -- I forgot to add it before, so I've added it now!

webextensions/api/dom.json
queengooborg: How do you suppose we should handle this?  Notes?  Alt. names?

lucalves (approved):

webextensions/api/dom.json
Rob--W: Is `alternative_name` typically used for same API signatures? The difference here is a method vs a getter property.

How about `partial_implementation` with the note that it's available as `element.isOpenOrClosedShadowRoot` instead of a method in the extension namespace?

If customary, we could even use alternative_name + partial_implementation + notes.

queengooborg: It's hard to classify what is partial implementation, since that property refers to "an incomplete implementation of the spec that causes significant browser compatibility issues"...but since web extensions have no spec to refer to...!

However, if this method has a completely different structure between Firefox and Chrome, I'm actually thinking that we should split it into two separate features.

queengooborg:
I'm going to go ahead and merge this since it's been sitting around for a while and has continued to cause issues.  Any changes can be made as follow-up PRs.

PR #17708, opened 2022-09-06
Update Firefox versions for api.CanvasRenderingContext2D.measureText

queengooborg:
This PR updates and corrects the real values for Firefox and Firefox Android for the `measureText` member of the `CanvasRenderingContext2D` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.3).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/CanvasRenderingContext2D/measureText

### DIFF
--- a/api/CanvasRenderingContext2D.json
+++ b/api/CanvasRenderingContext2D.json
@@ -2010,7 +2010,7 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "2"
+              "version_added": "3.5"
             },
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/3.5 says "Canvas elements now support the HTML 5 text API." so I think this makes sense.

PR #17710, opened 2022-09-06
Update Firefox versions for Console API

queengooborg:
This PR updates and corrects the real values for Firefox and Firefox Android for the `Console` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.3).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Console

### DIFF
--- a/api/Console.json
+++ b/api/Console.json
@@ -300,7 +300,7 @@
               "version_added": "12"
             },
             "firefox": {
-              "version_added": "4"
+              "version_added": "5"
             },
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
Confirmed with https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/5#developer_tools for `debug`.

And `group` shipped with `groupEnd` and so I think 9 is correct, too.

PR #17712, opened 2022-09-06
Mirror javascript.builtins to Opera

queengooborg:
This PR mirrors the JavaScript builtins data from Chrome to Opera where it differs and the collector states the support should match.


Elchi3 (approved):
Thanks, mirroring makes sense for these core features.

PR #17714, opened 2022-09-06
Mirror javascript.builtins to WebView

queengooborg:
This PR mirrors the JavaScript builtins data to WebView as needed.  Support is confirmed using the mdn-bcd-collector project.


Elchi3 (approved):
Thanks, I can't think of a reason why these things shouldn't be mirrored.

PR #17734, opened 2022-09-08
Add Firefox version removed information for menuitem element

bsmth:
__Description:__

Support for `menuitem` was removed in Fx103. The related pref `dom.menuitem.enabled` is also removed.

__Links:__

- Bugzilla: https://bugzilla.mozilla.org/show_bug.cgi?id=1372276

__Related:__

- https://github.com/mdn/content/pull/20421
- Parent task: https://github.com/mdn/content/issues/19707


api/Element.json
```diff
@@ -8438,19 +8438,20 @@
...
                     "type": "preference",
                     "name": "dom.menuitem.enabled",
                     "value_to_set": "true"
                   }
                 ]
-              },
-              {
-                "version_added": "8",
-                "version_removed": "85"
               }
```
queengooborg, line 8455: We've updated the [flag data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines/index.md#removal-of-irrelevant-flag-data) so that flag data can be removed right away once the flag is removed.

And with that in mind, a lot of these features would now be considered irrelevant altogether and can be removed entirely [per the guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines/index.md#removal-of-irrelevant-features)!

queengooborg:
Oh I forgot to say: welcome to BCD, Brian, it's great to have you here!  🎉   Please let me know if you have any questions!

api/Element.json
bsmth: Hi @queengooborg, vthanks a lot. 

So if I understand correctly, I can find all occurrences of `dom.menuitem.enabled` (which is deprecated) and remove it. If it's the only item in an array of flags, I presume I just remove the whole flags arr:

```json
                "flags": [
                  {
                    "type": "preference",
                    "name": "dom.menuitem.enabled",
                    "value_to_set": "true"
                  }
                ]
```

queengooborg: Actually, the whole support statement, including `version_added` and `version_removed`, should be removed, so this:

```json
"firefox": [
  {
    "version_added": "8",
    "version_removed": "85"
  },
  {
    "version_added": "85",
    "version_removed": "103",
    "flags": [
      {
        "type": "preference",
        "name": "dom.menuitem.enabled",
        "value_to_set": "true"
      }
    ]
  }
],
```

...becomes this:

```json
"firefox": [
  {
    "version_added": "8",
    "version_removed": "85"
  }
],
```

...which then resolves to this in the end, since support statements with only one member should no longer be an array:

```json
"firefox": {
  "version_added": "8",
  "version_removed": "85"
},
```

bsmth: OK thank you, that makes sense. I will follow up with changes shortly.

bsmth:
> Welcome to BCD, Brian, it's great to have you here!

Thanks so much, delighted to join 🙌🏻

queengooborg (approved):
Thank you for the quick updates, this is LGTM!

PR #17736, opened 2022-09-08
chore: unignore more paths for Prettier

nschonni:
#### Summary

Now that the prettier globs have been expanded and run on everything, this unignores the rest of the non-generated files since they mostly already follow the formating


nschonni:
@queengooborg the lint test failed because I think the diff is using JSON.stringify, so I ended up having to use the other parser.
Feel free to close if it doesn't make sense

queengooborg:
We're actually using a slightly custom JSON stringification that is handled in `npm run fix`, so I don't think we should have two separate formatters running on the same files.

PR #17743, opened 2022-09-09
Fix obsolete test

### LINKED ISSUE #6431: Linter: prevent an all "no" row

queengooborg:
This PR fixes the `implementedAndRemoved` test of the obsolete feature linter to accommodate mirrored statements and handle when other browsers have `version_added: false`.

Closes #6809 as irrelevant, since this performs the same logic.  Naturally fixes #6431.


queengooborg:
Note: the tests are failing because there are a lot of features that can be considered obsolete (or are not supported).

PR #17745, opened 2022-09-09
Fix broken links in browser data guidelines

ddbeck:
#### Summary

The browser data guidelines' links are broken (I suppose because they're nested on level deeper than before).

I clicked on them and 💔 

#### Related issues

#17519


queengooborg (approved):
Great catch, thank you!

PR #17763, opened 2022-09-11
Update Safari versions for javascript.builtins.Function.name

queengooborg:
This PR updates and corrects the real values for Safari (Desktop and iOS/iPadOS) for the `name` member of the `Function` JavaScript builtin, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.1.3).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/javascript/builtins/Function/name

### DIFF
--- a/javascript/builtins/Function.json
+++ b/javascript/builtins/Function.json
@@ -534,7 +534,7 @@
                 "version_added": "11"
               },
               "safari": {
-                "version_added": "6"
+                "version_added": "4"
               },
               "safari_ios": "mirror",
               "samsunginternet_android": "mirror",


foolip (commented):
This is a slightly bigger change, can you do some manual testing as well to see if there are any differences in Safari 4 and 6?

```js
function myName() {}
console.log(myName.name);
```

Perhaps it didn't work initially.

queengooborg:
In this case, we should probably add a custom test, but I don't think the collector is currently equipped for custom tests for JavaScript.

foolip:
Custom tests for JS are actually possible, see here for an example:
https://github.com/foolip/mdn-bcd-collector/blob/d5ccc5002c4ad001df81edbc4b5152ab0afe099b/custom-js.json#L512-L513

PR #17772, opened 2022-09-13
Make Safari 5.1 mirror to iOS 5 (not 6)

foolip:
Safari 5.1 has engine version 534.48, while iOS 5 has engine version 534.46. That's slightly before the Safari version, and this has caused mirroring so far to map Safari 5.1 to iOS 6. But 534.48 and 534.46 are close enough that we can consider them the same, and the Safari macOS release as an upstream for the iOS release.

Existing data mirroring from 5.1 is updated so that there are no changes to build/data.json.

This will allow a lot of 5.1+5 pairs to be mirrored in a follow-up PR.


queengooborg (approved):

foolip:
I've sent https://github.com/mdn/browser-compat-data/pull/17813 to burn down the changes made here, since the data was likely wrong.

PR #17793, opened 2022-09-14
Amend menuitem fx103 menuitem and context menu

bsmth:
This PR reverts changes that landed in a252dd9 and adds a version removed for the `<meniutem>` element and context menu.

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -8437,10 +8437,22 @@
             },
             "chrome_android": "mirror",
             "edge": "mirror",
-            "firefox": {
-              "version_added": "8",
-              "version_removed": "85"
-            },
+            "firefox": [
+              {
+                "version_added": "85",
+                "flags": [
+                  {
+                    "type": "preference",
+                    "name": "dom.menuitem.enabled",
+                    "value_to_set": "true"
+                  }
+                ]
+              },
+              {
+                "version_added": "8",
+                "version_removed": "85"
+              }
+            ],
             "firefox_android": "mirror",
             "ie": {
               "version_added": false
--- a/api/Element.json
+++ b/api/Element.json
@@ -8440,6 +8440,7 @@
             "firefox": [
               {
                 "version_added": "85",
+                "version_removed": "103",
                 "flags": [
                   {
                     "type": "preference",
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -585,6 +585,7 @@
             "firefox": [
               {
                 "version_added": "85",
+                "version_removed": "103",
                 "flags": [
                   {
                     "type": "preference",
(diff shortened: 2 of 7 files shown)


queengooborg:
As stated in Slack, #17734 is good because we don't need to keep the flag data around after it's become irrelevant!

bb010g:
@queengooborg I just got really thrown off by this while debugging a WebExtension that broke from the mentioned change, and tracked this down because [Bug 1372276](https://bugzilla.mozilla.org/show_bug.cgi?id=1372276) documented its removal in Firefox 103, not Firefox 85. MDN now states that [`<menuitem>` was "\[r\]emoved in Firefox 85 and later"](https://developer.mozilla.org/en-US/docs/Web/HTML/Element/menuitem#browser_compatibility), which is false. I get that it was behind a flag and not something the general internet should have been using, but the feature was behind a flag for Firefox ESR 91 and is still behind a flag for Firefox ESR 102. The next ESR, Firefox ESR 115, will be the first ESR with the flag removed, but that isn't happening until [Q3 2023](https://wiki.mozilla.org/Release_Management/Calendar). I don't think the flag data should be completely removed from BCD, at least until Firefox ESR 102 is no longer supported (which is scheduled for 2023-09-26). The data isn't yet irrelevant, as the flag was/is being used for its intended purpose of provisional extended support after deprecation & before removal, and removal has only happened so far on mainline, not yet on ESR. If the flag had been fully made unnecessary before its removal in mainline, the the presence of this data would indeed be irrelevant, but I wouldn't be here if everything had been properly patched.

Would it be okay if I opened a new PR with the changes here? (I'd make sure @bsmth is still the commit author.)

queengooborg:
Hi @bb010g -- the policy of removing flag data from BCD as soon as it is no longer applicable to a feature can be found here: https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines/index.md#removal-of-irrelevant-flag-data  Originally, the policy was to wait two years before performing flag data removal, however after running a pull and discussing the results with the rest of the MDN team, it was decided that once a flag will no longer affect the availability of a feature, it should not be recorded in BCD.

A big part of the reasoning behind this is because flags are meant to allow advanced users to test new, bleeding-edge features.  No normal user should be enabling flags as they pose potential stability and security vulnerabilities.  I feel this is especially true for enterprise users in the case of ESR releases.

Unfortunately, ESR releases of browsers are not included when considering the flag removal policy.  I realize this is inconvenient for ESR release tracking, and as such I apologize for the hassle.  However, I don't think that we should add the flag data back in as our script will simply remove it all over again, and the context of ESR releases, I feel this is a niche case and not worth adding an exception for.

bb010g:
> Unfortunately, ESR releases of browsers are not included when considering the flag removal policy. I realize this is inconvenient for ESR release tracking, and as such I apologize for the hassle. However, I don't think that we should add the flag data back in as our script will simply remove it all over again,

Could the flag removal policy & script be updated to deal with ESRs? This seems like a pretty visible flaw in the policy, and it's likely to recur.

PR #17794, opened 2022-09-14
Update early support for CanvasGradient and CanvasPattern

foolip:
Support for CanvasPattern is implied by support for ctx.createPattern(),
since that's how a CanvasPattern instance is created. Similarly, support
for CanvasGradient is implied by support for either
createLinearGradient() or createRadialGradient().

The support data for CanvasRenderingContext2D looks reliable based on
the PRs that updated it to the current state:
https://github.com/mdn/browser-compat-data/pull/7465
https://github.com/mdn/browser-compat-data/pull/8666

Update CanvasGradient and CanvasPattern to match. The versions now match
for all browsers except Opera, where some ≤12.1 versions were left alone
rather than trying to confirm the correct versions.

### DIFF
--- a/api/CanvasGradient.json
+++ b/api/CanvasGradient.json
@@ -6,7 +6,7 @@
         "spec_url": "https://html.spec.whatwg.org/multipage/canvas.html#canvasgradient",
         "support": {
           "chrome": {
-            "version_added": "6"
+            "version_added": "1"
           },
           "chrome_android": "mirror",
           "edge": {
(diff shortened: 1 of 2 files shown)


api/CanvasGradient.json
```diff
@@ -6,7 +6,7 @@
         "spec_url": "https://html.spec.whatwg.org/multipage/canvas.html#canvasgradient",
         "support": {
           "chrome": {
-            "version_added": "6"
+            "version_added": "1"
```
foolip, line 9: I'm hoping this will be confirmed by https://github.com/foolip/mdn-bcd-collector/pull/2334 but I don't have Chrome 1 to test with myself.

lucalves (approved):

queengooborg (approved):

PR #17799, opened 2022-09-14
Label any PR changing files starting with a dot as "infra"

queengooborg:
This PR updates the labeler rules to mark any files starting with a dot as "infra".


nschonni:
Might be worth pairing this with the https://github.com/mdn/browser-compat-data/blob/main/.github/CODEOWNERS

PR #17800, opened 2022-09-14
Update data for HTTP header Expect-CT

bershanskiy:
#### Summary
Update information for HTTP header `Expect-CT`.

1. Changed `chrome_android` to `mirror` because `chrome.notes` describes [bug 786563](https://crbug.com/786563) that also applies to Android.
2. Changed `ie.version_added` and `safari.version_added` from `null` to `false` because IE and Safari never supported `Expect-CT` at all. ([Source](https://chromestatus.com/feature/5098783126323200): "No other browser [besides Chromium] has implemented `Expect-CT` so removing it is not an interoperability concern.")
3. Changed `status.deprecated` to `true` because [Chrome Platform Status](https://chromestatus.com/feature/5098783126323200) deprecated `Expect-CT`. Eventually, Chromium is probably going to remove `Expect-CT` entirely.

#### Related issues
Matching content PR: https://github.com/mdn/content/pull/20742


http/headers/Expect-CT.json
```diff
@@ -41,7 +39,7 @@
           "status": {
             "experimental": false,
             "standard_track": true,
-            "deprecated": false
+            "deprecated": true
```
queengooborg, line 42: We shouldn't change `deprecated` to true if a _browser_ deprecates the feature, but rather if the _spec_ deprecates it.

```suggestion
            "deprecated": false
```

bershanskiy: Thanks for the clarification, I updated this!

queengooborg (approved):

PR #17811, opened 2022-09-15
Remove "default values" notes for Web Audio constructors

foolip:
These notes have been in BCD since the very beginning:
https://github.com/mdn/browser-compat-data/pull/433

Chrome 55 is the right version for the constructors overall:
https://storage.googleapis.com/chromium-find-releases-static/579.html#57909df69be0cf0b2ac42e5c9c018d4a00ee9766

The Chrome 59 change is this one:
https://storage.googleapis.com/chromium-find-releases-static/ef0.html#ef006156fd62853a893617ca104f4ac2045e19c1

Adding default values to dictionary members is by itself not necessarily
a web observable changes at all. This was confirmed with one case
comparing Chrome 58 and 59:
http://software.hixie.ch/utilities/js/live-dom-viewer/?saved=10708

One part of this was observable, with the OscillatorNode constructor:
https://bugs.chromium.org/p/chromium/issues/detail?id=710471

This was a spec bug, and fixed in the spec around the same time:
https://github.com/WebAudio/web-audio-api/issues/1173

Overall, this change does not seem noteworthy enough to record in BCD.

### DIFF
--- a/api/AudioBufferSourceNode.json
+++ b/api/AudioBufferSourceNode.json
@@ -44,8 +44,7 @@
           "spec_url": "https://webaudio.github.io/web-audio-api/#dom-audiobuffersourcenode-constructor-audiobuffersourcenode",
           "support": {
             "chrome": {
-              "version_added": "55",
-              "notes": "Before version 59, the default values were not supported."
+              "version_added": "55"
             },
             "chrome_android": "mirror",
             "edge": {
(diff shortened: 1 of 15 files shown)


Elchi3 (approved):
I agree it is not that important. This could have been put in a range but given how long ago this has been fixed, that's not worth it here either. LGTM 👍

PR #17814, opened 2022-09-15
Update Chromium compat data for math@display

fred-wang:
#### Summary

- Add new implementation performed in [1].

- Also remove legacy implementation from Chrome 24, there is already some documentation about this MathML implementation for the mathml.elements.math feature and not documenting Chrome 24 is consistent with what we do for other MathML subfeatures that have been re-implemented later.

[1] https://chromiumdash.appspot.com/commit/94e1142bd73bebfdce14bc779601bad8277fc4c6


fred-wang:
cc @Elchi3

Elchi3 (approved):

PR #17824, opened 2022-09-17
NetworkInformation is removed from Firefox Desktop in version 32

saschanaz:
#### Summary

It's exposed in 31 but that was accidental; it was soon removed in version 32 by https://bugzilla.mozilla.org/show_bug.cgi?id=1020758.

And it was enabled by default.

Version 32 indeed has no `navigator.connection` while v31 does, per builds from mozregression.


queengooborg:
Looks like you've got some consistency issues with the NetworkInformation API.  Could you fix those up real quick?

queengooborg (approved):

PR #17831, opened 2022-09-18
Chrome support unprefixed hyphenate-character

rachelandrew:
Chrome support for unprefixed hyphenate-character is in 106: https://chromestatus.com/feature/5169156928831488


foolip (changes requested):
Can you also change `webview_android` to "mirror"? Otherwise the data will claim it's still prefixed-only in WebView.

rachelandrew:
Missed that one, done now.

queengooborg (approved):
Thanks for the quick changes, LGTM!

PR #17834, opened 2022-09-19
FF105 CSP script- and style- src-elem and src-attr directives

hamishwillee:
FF105 adds support for  CSP 'script-src-elem' and 'script-src-attr' directives and CSP 'style-src-elem' and 'style-src-attr' directives in  https://bugzilla.mozilla.org/show_bug.cgi?id=1529337 and https://bugzilla.mozilla.org/show_bug.cgi?id=1529338 respectively - enabled in nightly and behind preferences.

Other docs work tracked in https://github.com/mdn/content/issues/20878

FYI  @jpmedley I noticed that Opera Android support is marked as Null - there is nothing to indicate why that would be in https://chromestatus.com/feature/5141352765456384 or the issue in which it was updated: https://github.com/mdn/browser-compat-data/pull/4529

### DIFF
--- a/http/headers/Content-Security-Policy.json
+++ b/http/headers/Content-Security-Policy.json
@@ -1132,10 +1132,21 @@
               },
               "chrome_android": "mirror",
               "edge": "mirror",
-              "firefox": {
-                "version_added": false,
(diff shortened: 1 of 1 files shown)


hamishwillee:
FYI @queengooborg

queengooborg (approved):

PR #4448, opened 2019-07-06
Set Chromium/Safari versions for various CSS selectors (part 2)

vinyldarkscratch:
This sets Chromium and Safari versions for various CSS selectors within spec, as to aid #4303.

<details>
	<summary>css.selectors.host</summary>
	<ol>
		<li>Implemented in 6d9e2418d7d7310dcc0d59dd7abb0641d21406c8</li>
		<li>Implemented in WebKit 602.1.7</li>
		<li>Chrome already has "54" in data</li>
		<li><strong>Safari 10, Chrome 54</strong></li>
	</ol>
</details>

<details>
	<summary>css.selectors.in-range, css.selectors.out-of-range</summary>
	<ol>
		<li>Implemented in f55645df30dfadef723a4c727f8b6ad5ba05caa3</li>
		<li>Implemented in WebKit 534.16</li>
		<li><strong>Safari 5.1, Chrome 10</strong></li>
	</ol>
</details>

<details>
	<summary>css.selectors.indeterminate, css.selectors.indeterminate.checkbox</summary>
	<ol>
		<li>Implemented in f3fd7741eece9d277e7ecb9b165b8ce14513e273</li>
		<li>Safari already has "3" in data for checkbox</li>
		<li><strong>Mirrored onto parent and Chrome (Safari 3, Chrome 1)</strong></li>
	</ol>
</details>

<details>
	<summary>css.selectors.indeterminate.progress</summary>
	<ol>
		<li>Chrome already has "6" in data</li>
		<li><strong>Mirrored onto Safari (Safari 5.1)</strong></li>
	</ol>
</details>

<details>
	<summary>css.selectors.invalid.form, css.selectors.valid.form</summary>
	<ol>
		<li>Supported in Safari 9 (manual testing)</li>
		<li>Chrome already has "40" in data</li>
		<li><strong>Safari 9, Chrome 40</strong></li>
	</ol>
</details>

<details>
	<summary>css.selectors.left, css.selectors.right</summary>
	<ol>
		<li>Implemented in 1fcfaa5241a51365843a2c422fa2187ec5c3848a</li>
		<li>Implemented in WebKit 533.7</li>
		<li><strong>Safari 5.1, Chrome 6</strong></li>
	</ol>
</details>

<details>
	<summary>css.selectors.placeholder-shown.non_text_types</summary>
	<ol>
		<li>Supported in Safari 9 (manual testing)</li>
		<li>Chrome already has "47" in data</li>
[...]

PR #17852, opened 2022-09-21
Update support for @page :left/:right pseudo-classes

foolip:
The Safari data for this comes from research in this PR:
https://github.com/mdn/browser-compat-data/pull/4448

If WebKit 533.7 is correct, that maps to Safari 5.

### DIFF
--- a/css/selectors/left.json
+++ b/css/selectors/left.json
@@ -29,11 +29,9 @@
               "version_added": "10.1"
             },
             "safari": {
-              "version_added": "5.1"
+              "version_added": "5"
             },
-            "safari_ios": {
-              "version_added": "6"
-            },
+            "safari_ios": "mirror",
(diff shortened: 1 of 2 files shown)


foolip:
@queengooborg in https://github.com/mdn/browser-compat-data/pull/4448 you wrote:

<ol>
<li>Implemented in 1fcfaa5241a51365843a2c422fa2187ec5c3848a</li>
<li>Implemented in WebKit 533.7</li>
<li><strong>Safari 5.1, Chrome 6</strong></li>
</ol>

I can't find commit 1fcfaa5241a51365843a2c422fa2187ec5c3848a, do you know which repo that was in?

queengooborg (approved):
Probably would have been the WebKit repository on Git if I had to guess, but I don't really trust the version numbers from back then -- so this is LGTM!

PR #17858, opened 2022-09-22
Remove onvisibilitychange notes from webkitvisibilitychange

foolip:
The prefixed event shouldn't be assoicated with the unprefixed event
handler property.

### DIFF
--- a/api/Document.json
+++ b/api/Document.json
@@ -6778,9 +6778,7 @@
               },
               {
                 "version_added": "13",
-                "prefix": "webkit",
-                "partial_implementation": true,
-                "notes": "The <code>onvisibilitychange</code> event handler property is not supported."
+                "prefix": "webkit"
               }
             ],
             "chrome_android": "mirror",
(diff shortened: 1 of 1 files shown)


queengooborg:
Are we sure that the notes are irrelevant though?  Based upon the delay in a lack of support for the event handler property, shouldn't the notes be changed to reference the prefixed event handler?

lucalves (approved):

foolip:
> Are we sure that the notes are irrelevant though? Based upon the delay in a lack of support for the event handler property, shouldn't the notes be changed to reference the prefixed event handler?

My perspective is that with non-standard APIs, we can't say that something is partially implemented because there's nothing to compare it to. We could have a note, but not `partial_implementation`. Would that work?

queengooborg:
This isn't non-standard though, it's just prefixed.  But even if it was non-standard, we should still retain a `partial_implementation` statement per our [guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines/api.md#dom-events-eventname_event), as part of the feature (the event handler) is known to be unsupported.

foolip:
At the time it was implemented it might have been non-standard, but I guess that doesn't matter now. OK, so I'll have to research whether the event handler properties were supported or not.

PR #17861, opened 2022-09-22
Add data guideline on `partial_implementation`

jugglinmike:
#### Summary

Because the `partial_implementation` property describes an incomplete feature, it is somewhat ambiguous whether the feature it describes should be considered "supported." As of this commit, the property is exclusively included in affirmative support statements. Formalize this trend with an explicit data guideline.

n/a

### DIFF
--- a/docs/data-guidelines/index.md
+++ b/docs/data-guidelines/index.md
@@ -68,6 +68,12 @@ If you set `"partial_implementation": true`, then write a note describing how th
 
 This guideline was proposed in [#7332](https://github.com/mdn/browser-compat-data/pull/7332).
 
+## `"partial_implementation"` requires an affirmative `"version_added"`
+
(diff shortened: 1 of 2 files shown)


jugglinmike:
This patch is based on my interpretation of BCD's data guidelines. [I've recently suggested making it a formal policy.](https://github.com/mdn/browser-compat-data/pull/17861)

queengooborg:
I'm not sure if this is explicitly needed.  Since we are working towards 100% real values everywhere, we are planning to eliminate all uncertain version data anyways, including those on `partial_implementation`.  I'm not against merging this though, so I'll leave it open for another to review!

jugglinmike:
That makes sense to me! I didn't realize that `partial_implementation` is expected to be removed. Maybe it would be best to abandon this patch and instead add a note discouraging its use.

queengooborg:
What I mean is that we're working on replacing all non-real values (that is, "true" and "null") with real values (proper version numbers or "false").  Eventually, the requirement for real values will apply everywhere in BCD.  It's not that this patch is _wrong,_ it's just a subset of the bigger goal!

PR #17880, opened 2022-09-26
FF105 gfx.offscreencanvas.enabled enabled by default

hamishwillee:
FF105 enables `OffScreenCanvas` by default in https://bugzilla.mozilla.org/show_bug.cgi?id=1779009.

@queengooborg 

1. I have removed the preference versions `gfx.offscreencanvas.enabled` now that 105 is released. I can put it back if needed (all done in one commit).
2. Note that the preference was "partial implementation" and points to the meta bug: https://bugzil.la/1390089 - that appears to be mostly done, and the outstanding stuff is just bugs. So I have NOT marked the released version as partial. 

Other docs work for this tracked in https://github.com/mdn/content/issues/11591.

FYI @bsmth

### DIFF
--- a/api/OffscreenCanvas.json
+++ b/api/OffscreenCanvas.json
@@ -82,7 +82,7 @@
             "chrome_android": "mirror",
             "edge": "mirror",
             "firefox": {
-              "version_added": false
+              "version_added": "105"
             },
             "firefox_android": "mirror",
             "ie": {
@@ -99,7 +99,7 @@
             "webview_android": "mirror"
           },
           "status": {
-            "experimental": true,
+            "experimental": false,
             "standard_track": true,
             "deprecated": false
           }
@@ -119,7 +119,7 @@
             "chrome_android": "mirror",
             "edge": "mirror",
             "firefox": {
-              "version_added": false
+              "version_added": "105"
             },
             "firefox_android": "mirror",
             "ie": {
@@ -136,7 +136,7 @@
             "webview_android": "mirror"
           },
           "status": {
-            "experimental": true,
+            "experimental": false,
             "standard_track": true,
             "deprecated": false
           }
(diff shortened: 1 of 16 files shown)


api/OffscreenCanvas.json
```diff
@@ -461,7 +468,7 @@
               "chrome_android": "mirror",
               "edge": "mirror",
               "firefox": {
-                "version_added": false
+                "version_added": "105"
```
hamishwillee, line 471: Whoops. This one is the accessor for 2d context.

api/OffscreenCanvas.json
```diff
@@ -214,7 +217,7 @@
             "webview_android": "mirror"
           },
           "status": {
-            "experimental": true,
+            "experimental": false,
             "standard_track": true,
             "deprecated": false
           }
```
hamishwillee, line 223: @queengooborg I'm getting an error for lines below

```
api.OffscreenCanvas.convertToBlob - Error → Basic support in firefox was declared implemented in a later version (105) than the following sub-feature(s):
        → api.OffscreenCanvas.convertToBlob.option_type_parameter_webp: 96
```

Presumably the solution is just to update support from 96 to 105 (this should have been behind the pref too "in theory")

hamishwillee: Actually, I'll just update it to 105. That makes sense given that's when this release is "public"

api/OffscreenCanvasRenderingContext2D.json
```diff
@@ -273,8 +274,42 @@
...
+            "webview_android": "mirror"
+          },
+          "status": {
+            "experimental": false,
+            "standard_track": true,
+            "deprecated": false
+          }
+        }
+      },
+      "commit": {
```
hamishwillee, line 300: Note, this was added in FF105. The chrome version isn't certain but it was certainly around this time based on various chromium bugs - and this matches the parent.

bsmth:
Thanks @hamishwillee

queengooborg (approved):

PR #17895, opened 2022-09-27
Add object-view-box

rachelandrew:
#### Summary

Adds BCD for the `object-view-box` property.

https://chromestatus.com/feature/5213032857731072

### DIFF
--- a/css/properties/object-view-box.json
+++ b/css/properties/object-view-box.json
@@ -3,7 +3,7 @@
     "properties": {
       "object-view-box": {
         "__compat": {
-          "spec_url": "https://drafts.csswg.org/css-images-4/#the-object-view-box",
+          "spec_url": "https://w3c.github.io/csswg-drafts/css-images-4/#the-object-view-box",
           "support": {
             "chrome": {
               "version_added": "104"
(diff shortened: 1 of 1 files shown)


css/properties/object-view-box.json
```diff
@@ -0,0 +1,39 @@
...
+            },
+            "firefox_android": "mirror",
+            "ie": {
+              "version_added": false
+            },
+            "oculus": "mirror",
+            "opera": "mirror",
+            "opera_android": "mirror",
+            "safari": {
+              "version_added": "false"
```
queengooborg, line 24:
```suggestion
              "version_added": false
```

css/properties/object-view-box.json
```diff
@@ -0,0 +1,39 @@
+{
+  "css": {
+    "properties": {
+      "object-view-box": {
+        "__compat": {
+          "spec_url": "https://drafts.csswg.org/css-images-4/#the-object-view-box",
```
queengooborg, line 6: Since `drafts.csswg.org` is down so often, we require the use of another official mirror instead:

```suggestion
          "spec_url": "https://w3c.github.io/csswg-drafts/css-images-4/#the-object-view-box",
```

queengooborg (changes requested):
Looks like there's a couple of issues with the data, but otherwise LGTM!

queengooborg (approved):

PR #17901, opened 2022-09-28
Removal of support for the object-src directive

rebloor:
#### Summary

Adds a note to provide documentation for [Bug 1766881](https://bugzilla.mozilla.org/show_bug.cgi?id=1766881) Remove object-src requirement from the extension CSP, at least in MV3

#### Related issues

Related changes to the release notes to be completed.

### DIFF
--- a/webextensions/manifest/content_security_policy.json
+++ b/webextensions/manifest/content_security_policy.json
@@ -6,7 +6,8 @@
           "mdn_url": "https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/manifest.json/content_security_policy",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": true,
+              "notes": "The <code>object-src</code> directive is required."
             },
             "edge": {
               "version_added": "14",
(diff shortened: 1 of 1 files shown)


webextensions/manifest/content_security_policy.json
```diff
@@ -39,6 +39,7 @@
               "edge": "mirror",
               "firefox": {
                 "version_added": "72",
+                "notes": "From Firefox 105, the directive <code>object-src<code> is not supported.",
```
Rob--W, line 42:
```suggestion
                "notes": "Up until Firefox 104, the <code>object-src</code> directive was required with a secure source. Since Firefox 105, the <code>object-src</code> directive is optional.",
```

Rob--W (changes requested):
The note was wrong; Firefox didn't stop supporting it, it just stopped requiring unnecessary boilerplate.

Note that Safari doesn't require it either.
So you can also add a note to Chrome (mirrored to Edge and Opera) that they still require object-src to be specified, with a link to the crbug that I shared on Bugzilla.

webextensions/manifest/content_security_policy.json
rebloor: @Rob--W could you check again, I think I got the version wrong. Also "since" may be ambiguous as it also means "as" etc. so changed.

queengooborg: @rebloor You edited @Rob--W's suggestion to change "Since Firefox 105" to "106", but you didn't tweak "Until Firefox 104".  What happens in Firefox 105?  Does the other part need to be modified as well?

By the way, @rebloor, please avoid modifying others' comments in such ways unless agreed upon or otherwise needed to remove language that violates code of conduct.  If there's a discrepancy in the version numbers, it's better to add your own comment so that it can be further discussed.  :)

queengooborg: Ah, I see that this suggestion wasn't merged directly and instead another commit was made to make the requested changes with an answer to my question, never mind!

webextensions/manifest/content_security_policy.json
```diff
@@ -6,25 +6,31 @@
           "mdn_url": "https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/manifest.json/content_security_policy",
           "support": {
             "chrome": {
-              "version_added": true
+              "version_added": true,
+              "notes": "The <code>object-src</code> directive is required."
             },
             "edge": {
               "version_added": "14",
               "notes": "Only the default content security policy is supported: \"script-src 'self'; object-src 'self';\"."
```
Rob--W, line 14: Edge 79+ is based on chromium and mirrors the same behavior.

Rob--W (approved):

Bdieir (approved):

Rob--W (approved):

PR #17906, opened 2022-09-28
Update based on mdn-bcd-collector results from Safari Technology Preview 154

gsnedders:
#### Summary

Based on Safari 16.0 and Safari Technology Preview 154 on macOS 12.6, run mdn-bcd-collector's update-bcd.

Tested with mdn-bcd-collector.

There are various release notes for these changes; I've not collected them all myself, but hopefully we trust mdn-bcd-collector to do its job.

### DIFF
--- a/api/ElementInternals.json
+++ b/api/ElementInternals.json
@@ -21,10 +21,12 @@
           "opera": "mirror",
           "opera_android": "mirror",
           "safari": {
-            "version_added": false,
+            "version_added": "preview",
             "notes": "See <a href='https://webkit.org/b/197960'>bug 197960</a>."
           },
-          "safari_ios": "mirror",
+          "safari_ios": {
+            "version_added": false
+          },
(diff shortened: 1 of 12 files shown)


api/ElementInternals.json
```diff
@@ -21,10 +21,12 @@
           "opera": "mirror",
           "opera_android": "mirror",
           "safari": {
-            "version_added": false,
+            "version_added": "preview",
```
rniwa, line 24: This isn't right. ElementInternals isn't enabled by default in STP right now.

gsnedders:
> Hey @gsnedders -- thanks for submitting this PR, but since it's been sitting around for a while and involves TP releases (plus Safari 16.1 has been released), I think it would be much better to open a new PR with data from the most recent TP, especially as it seems that some features were already disabled or released since this PR was created.

Yeah, that was my plan, but I've been travelling for weeks which hasn't helped!

PR #17912, opened 2022-09-29
Add subfeatures for menclose@notation

fred-wang:
#### Summary

This add subfeatures for each value of the menclose@notation attribute as discussed in [1], except "radical" which is considered irrelevant [2]. Additonally, "mirror" is used for Safari iOS (as it has the same MathML support as macOS) and the notes for Safari is removed (as that's redundant information).

- This was just done by copying existing implementation status of the notation feature, and mirroring for Safari iOS. These values were already in MathML2 so hopefully they were in the initial implementations.
- I also checked on https://developer.mozilla.org/en-US/docs/Web/MathML/Element/menclose current support for WebKit and Firefox.

#### Related issues

[1] https://github.com/mdn/browser-compat-data/pull/17062
[2] https://github.com/mdn/content/pull/21179

### DIFF
--- a/mathml/elements/menclose.json
+++ b/mathml/elements/menclose.json
@@ -53,12 +53,9 @@
               "opera": "mirror",
               "opera_android": "mirror",
               "safari": {
-                "version_added": "7",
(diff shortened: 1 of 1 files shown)


fred-wang:
cc @Elchi3 @queengooborg

Elchi3 (approved):
Thank you! 👍

PR #17916, opened 2022-09-30
Update support data for Meta Quest browser

jugglinmike:
#### Summary
    
These changes were generated procedurally by the mdn-bcd-collector project using data collected from the Meta Quest browser.

This patch is reproducible using the mdn-bcd-collector project at revision 4c430c1bb8bc2fcd3f72778e1dbe62bbb565587b and the data submitted to mdn-bcd-results here: https://github.com/foolip/mdn-bcd-results/pull/3278

#### Related issues

This is based on gh-17908 which has not been accepted at the time of submission. Although it may be technically possible to express this patch as an independent change set, the semantics of the "mirror" support value mean that there will be less churn if the Chrome for Android data is modified first. Feedback regarding the changes to Chrome for Android may be better shared in gh-17908.

### DIFF
--- a/api/AmbientLightSensor.json
+++ b/api/AmbientLightSensor.json
@@ -49,7 +49,9 @@
             "chrome": {
               "version_added": "56"
             },
-            "chrome_android": "mirror",
+            "chrome_android": {
+              "version_added": false
+            },
             "edge": "mirror",
             "firefox": {
               "version_added": false
--- a/api/AudioProcessingEvent.json
+++ b/api/AudioProcessingEvent.json
@@ -7,7 +7,9 @@
           "chrome": {
             "version_added": "14"
           },
-          "chrome_android": "mirror",
+          "chrome_android": {
+            "version_added": false
+          },
           "edge": {
             "version_added": "12"
           },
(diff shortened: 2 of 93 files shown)


jugglinmike:
I've added a commit that updates the guidance on the compat data schema to permit the version range specifier "≤23.0" for data on Quest. That value occurs 8 times in this patch. I'm suggesting allowing this because some aspects of the browser's maintenance make it particularly difficult to find more precise information (the source code is not publicly available, the maintainers do not provide old builds, the popular automation services do not include the browser, and [the release notes omit some recent releases](https://developer.oculus.com/documentation/web/browser-release-notes/)).

queengooborg:
I noticed that this also updates Chrome Android data along with Oculus Quest data.  Could we isolate just the Oculus Quest changes?

jugglinmike:
Yeah, that's the complexity I was trying to describe in the "Related issues" section above. Besides the process-related inefficiency (or "churn" as I referred to it), I'm also concerned about how the data will be interpreted by others in the interim. With Quest "mirroring" Chrome for Android, inaccuracies in Chrome for Android will appear as feature additions/removals in Quest.

Do you think it'll be feasible to land gh-17908 before this?

PR #17920, opened 2022-09-30
Update compat data for mfrac@linethickness

fred-wang:
This commit updates compat data for the linethickness attribute of the <mfrac> element:

- Replace notes about partial implementation in Safari with the use of of array of implementations.
- Use "mirror" for Safari iOS, its MathML implementation matches the macOS one.
- Add a subfeature for deprecated named values thin, medium, thick values that are not part of MathML Core. These were removed in Firefox 83 [1] but are still supported by WebKit [2].

[1] https://bugzilla.mozilla.org/show_bug.cgi?id=1664488
[2] https://github.com/WebKit/WebKit/blob/0b0b865747d48a813eee3c2f52da026714bd61aa/Source/WebCore/mathml/MathMLFractionElement.cpp#L59

#### Related issues

N/A

### DIFF
--- a/mathml/elements/mfrac.json
+++ b/mathml/elements/mfrac.json
@@ -100,13 +100,17 @@
               "oculus": "mirror",
               "opera": "mirror",
               "opera_android": "mirror",
-              "safari": {
-                "version_added": "5.1",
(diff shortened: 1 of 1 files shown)


fred-wang:
cc @Elchi3

Elchi3 (approved):
Thanks! 👍

PR #17924, opened 2022-09-30
Split 'color keywords' into named-color and system-color

wbamberg:
This PR splits the "Color keywords" subfeature into two:

- "named-color", reflecting the `<named-color>` type documented at https://developer.mozilla.org/en-US/docs/Web/CSS/named-color, 
- "system-color", reflecting the `<system-color>` type documented (currently) at https://developer.mozilla.org/en-US/docs/web/css/color_value/system_color_keywords, but which I am in the process of making into a proper CSS data type page and moving to https://developer.mozilla.org/en-US/docs/web/css/system-color . When I do that it will want its own BCD table, which is why I need a "system-color" subfeature.

This is also more accurate, since per spec https://w3c.github.io/csswg-drafts/css-color-4/#color-keywords includes four features, two of which (`transparent` and `currentcolor`) are already split out as subfeatures in BCD.

### DIFF
--- a/css/types/color.json
+++ b/css/types/color.json
@@ -84,11 +84,11 @@
             }
           }
         },
-        "color_keywords": {
+        "named-color": {
           "__compat": {
-            "description": "Color keywords",
-            "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/color_value/color_keywords",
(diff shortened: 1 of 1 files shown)


teoli2003 (approved):
👍

PR #17927, opened 2022-10-03
Add subfeature for selection the visible child of semantics

fred-wang:
#### Summary

Add subfeature for the proprietary algorithm to select the visible child of a <semantics> element and remove corresponding notes. This makes easier to compare browser implementations.

N/A (no changes)

#### Related issues

N/A


fred-wang:
cc @Elchi3

PR #17934, opened 2022-10-04
Add compatibility data for legacy MathML lengths

fred-wang:
#### Summary

MDN contains a section about legacy MathML lengths [1] and refers to a browser compatibility table `mathml.attribute_values` that is currently missing. This commit adds relevant compatibility data for two kinds of accepted values mentioned in this section:

- non-zero unitless length values (e.g. `5` to mean 500%),
- named spaces (e.g. `thinmathspace` to mean 3/18em)

The edge case numbers ending with a dot (e.g. `34.px`) is ignored, that seems a bug in MathML 3 that is not even mention in version 4 [2] and is not super important for web developers.

Corresponding data is added for length attributes, namely mfrac's linethickness, mo's lspace/rspace/maxsize/minsize, mspace's depth/height/width, mtable's width and the global mathsize attribute (note that non-zero unitless is not added for mspace, it's not relevant since the default dimensions are 0 anyway).

These features have been implemented a long time in Gecko, before the first release of Firefox. Unitless values have been removed in Firefox 70 [3] and the there is plan to disable the pref for named spaces [4]. The two features have also been implemented in Safari 10 [5] and there is no plan to implement them in Chromium since they are not part of MathML Core.

[1] https://developer.mozilla.org/en-US/docs/Web/MathML/Attribute/Values#legacy_mathml_lengths
[2] https://w3c.github.io/mathml/#fund_units
|3] https://bugzilla.mozilla.org/show_bug.cgi?id=1574749#c8
[4] https://bugzilla.mozilla.org/show_bug.cgi?id=1793549
[5] https://commits.webkit.org/136199@main

#### Related issues

N/A

### DIFF
--- a/mathml/attribute_values.json
+++ b/mathml/attribute_values.json
@@ -1,0 +1,76 @@
+{
+  "mathml": {
+    "attribute_values": {
+      "named_space": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/MathML/Attribute/Values#legacy_mathml_lengths",
+          "support": {
+            "chrome": {
+              "version_added": false
+            },
+            "chrome_android": "mirror",
+            "edge": "mirror",
+            "firefox": {
+              "version_added": "1"
+            },
+            "firefox_android": "mirror",
+            "ie": {
+              "version_added": false
+            },
+            "oculus": "mirror",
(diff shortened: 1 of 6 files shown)


fred-wang:
cc @Elchi3

PR #17947, opened 2022-10-06
docs(CSP) add compat data for 'wasm-unsafe-eval' source expression

bsmth:
This PR adds the `wasm-unsafe-eval` source expression as a child of `script-src`.

```http
Content-Security-Policy: script-src 'wasm-unsafe-eval';
```

__Open question:__

Although `'wasm-unsafe-eval'` can only be used with `script-src`, it's one of many [sources](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/Sources) that can be used for the `script-src` directive. Should `'wasm-unsafe-eval'` live under `script-src` or should sources have a separate list? This would be something like:

```
sources
├── <host-source> (including <scheme-source>?)
├── 'self'
├── 'unsafe-eval'
├── 'wasm-unsafe-eval'
├── 'unsafe-hashes'
├── 'unsafe-inline'
├── 'none'
├── 'nonce-<base64-value>'
├── '<hash-algorithm>-<base64-value>'
├── 'strict-dynamic'
└── 'report-sample'
```

The compat table for this would probably live on [the sources page on MDN](https://developer.mozilla.org/en-US/docs/Web/HTTP/Headers/Content-Security-Policy/Sources).

### Browser support details

* Safari (141) https://webkit.org/blog/12434/release-notes-for-safari-technology-preview-141/
* Chrome (97) https://chromestatus.com/feature/5499765773041664
* Fx (102) https://bugzilla.mozilla.org/show_bug.cgi?id=1740263

### Related issues and pull requests

- Parent task: https://github.com/mdn/content/issues/17619
- Content & Fx relnote changes: https://github.com/mdn/content/pull/16943


queengooborg (changes requested):
Since Edge and Opera are Chrome-based, I think we can safely mirror the data from Chrome to them.

bsmth:
> Since Edge and Opera are Chrome-based, I think we can safely mirror the data from Chrome to them.

Seems to be breaking some tests with:

```logs
    Consistency - 1 problem (1 error, 0 warnings):
 ✖ http.headers.Content-Security-Policy.script-src - Error → Unknown support in parent for opera_android, but support is declared in the following sub-feature(s):
        → http.headers.Content-Security-Policy.script-src.wasm-unsafe-eval: 
```

I've reverted in https://github.com/mdn/browser-compat-data/pull/17947/commits/6eb7fe1a3a2350d99165b52c05f5a2eacc9fee82

queengooborg (approved):
Looks like a lot of Opera Android is set to `null` in this file, sorry.  I'll submit a PR to fix it all up.  To keep this PR unblocked, I'll go ahead and merge this and then make the corrections to Opera later!

PR #17948, opened 2022-10-06
Update version ranges for Safari snap scroll implementation bug

jplatte:
#### Summary

Update the version ranges for scroll margin / padding not applying for scrolls to fragment target.

Tested manually by @thibaultamartin while working on a site with me.

#### Notes

There seem to be some discrepancies in wording and metadata between the margin and padding properties (padding has `version_removed`, margin doesn't but has the version in the note text). I'm happy to fix that as well if you tell me which version is preferred.

### DIFF
--- a/css/properties/scroll-margin-bottom.json
+++ b/css/properties/scroll-margin-bottom.json
@@ -23,13 +23,13 @@
             "opera_android": "mirror",
             "safari": [
               {
-                "version_added": "14.1"
+                "version_added": "16"
               },
               {
                 "version_added": "11",
                 "alternative_name": "scroll-snap-margin-bottom",
                 "partial_implementation": true,
-                "notes": "Before version 14.1, scroll margin is not applied for scrolls to fragment target or <code>scrollIntoView()</code>, see <a href='https://webkit.org/b/189265'>bug 189265</a>."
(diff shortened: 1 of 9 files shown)


jplatte:
Oh, to be clear this was tested on `Version 15.6.1 (17613.3.9.1.16)` (broken) and `Version 16.0 (17614.1.25.9.10, 17614)` (works). Only navigation to fragment was tested, not `scrollIntoView()`.

queengooborg:
Hey @jplatte, thank you for your PR and welcome to BCD!  Would you mind providing barebones test code so we may verify this on our end as well?

jplatte:
Do you mean an HTML file where the bug can be observed?

queengooborg:
That's correct, or a Codepen or something similar!

queengooborg:
Hey again!  Just checking to see if you're still planning to follow up on this PR?

jplatte:
It doesn't have very high priority to me and I would be happy if somebody else took care of creating test cases, but I do plan to get back to it some time if nobody else does that.

queengooborg:
I'm going to close this PR in that case.  If someone else reports this issue is was still present, I'll happily revisit this!

PR #17952, opened 2022-10-07
CSS scroll-timelines-*

hamishwillee:
FF103 adds  the following CSS properties behind a preference in https://bugzilla.mozilla.org/show_bug.cgi?id=1754897:
- scroll-timeline-name
- scroll-timeline-axis
- scroll-timeline (shorthand)

This adds the CSS properties to BCD

Note, normally I wouldn't document these, but they are in the spec, FF is moving fast on them, and Chrome also has this in Canary 107 behind experimental web preferences.

Note, I did not include the Chrome stuff because it is still early days. Tracking but is here though https://bugs.chromium.org/p/chromium/issues/detail?id=1317765

Other docs work for this can be tracked in https://github.com/mdn/content/issues/21069

### DIFF
--- a/css/properties/scroll-timeline-axis.json
+++ b/css/properties/scroll-timeline-axis.json
@@ -1,0 +1,47 @@
+{
+  "css": {
+    "properties": {
+      "scroll-timeline-axis": {
+        "__compat": {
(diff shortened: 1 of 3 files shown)


hamishwillee:
@queengooborg FYI.

PR #17953, opened 2022-10-07
Update compat data for the mpadded element

fred-wang:
#### Summary

- State that Chrome and Safari fully supports mpadded attributes as defined in MathML Core and remove old notes for partial implementation.
- Also remove old note about a bug in Firefox too that is not super important.
- Add subfeatures to describe non-standard features from legacy MathML implementations and indicate support for Safari and Firefox (see below).
- Add missing subfeature for the depth attribute.

As explained on MDN [1], MathML Core accepts `<length-percentage>` for mpadded attributes but some browsers also support more values. More precisely:

- Chrome parses them as `<length-percentage>` [2]
- WebKit parses them as a legacy MathML length [3] [4]
- Firefox parses them witht the complete legacy syntax [5] [6]

[1] https://developer.mozilla.org/en-US/docs/Web/MathML/Element/mpadded
[2] https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/mathml/mathml_padded_element.cc;l=15;drc=047c7dc4ee1ce908d7fea38ca063fa2f80f92c77
[3] https://searchfox.org/wubkat/rev/e8e9be7353991aee7df45aa62f0e0148f16e6f9f/Source/WebCore/mathml/MathMLPaddedElement.cpp#51
[4] https://developer.mozilla.org/en-US/docs/Web/MathML/Attribute/Values#legacy_mathml_lengths
[5] https://searchfox.org/mozilla-central/rev/d416a7b827c2d76b12848e355a4e03dde49ece25/layout/mathml/nsMathMLmpaddedFrame.cpp#117
[6] https://developer.mozilla.org/en-US/docs/Web/MathML/Element/mpadded#legacy_syntax

#### Related issues

https://github.com/mdn/content/pull/21258

### DIFF
--- a/mathml/elements/mpadded.json
+++ b/mathml/elements/mpadded.json
@@ -20,11 +20,7 @@
             "chrome_android": "mirror",
             "edge": "mirror",
             "firefox": {
-              "version_added": "1",
-              "notes": "<code>&lt;mpadded&gt;</code> within <code>&lt;mtable&gt;</code> is not supported before Firefox 20, see <a href='https://bugzil.la/459363'>bug 459363</a>."
-            },
-            "firefox_android": {
-              "version_added": "4"
+              "version_added": "1"
             },
             "ie": {
               "version_added": false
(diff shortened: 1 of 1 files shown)


fred-wang:
cc @Elchi3

Elchi3 (approved):

PR #17966, opened 2022-10-09
Improve ZoomSettings data

gregorypappas:
#### Summary

In  <a href='https://bugzil.la/1772166'>bug 1772166</a> I am updating Firefox's implementation of defaultZoomFactor to align with other browsers, I would like to document these changes on MDN.

Firefox's current implementation, which is hardcoded to always return `1`
https://searchfox.org/mozilla-central/rev/6f006e854adf794a7e17c136662d69d81cb6ca97/browser/components/extensions/parent/ext-tabs.js#1233

### DIFF
--- a/webextensions/api/tabs.json
+++ b/webextensions/api/tabs.json
@@ -1259,9 +1259,8 @@
                   },
                   {
                     "version_added": "45",
-                    "version_removed": "107",
                     "partial_implementation": true,
-                    "notes": "Will always return 1, regardless of the actual value."
+                    "notes": "Returns 1, regardless of the default zoom factor setting."
                   }
                 ],
                 "firefox_android": {
--- a/webextensions/api/tabs.json
+++ b/webextensions/api/tabs.json
@@ -1243,6 +1243,83 @@
               },
               "safari_ios": "mirror"
             }
+          },
+          "defaultZoomFactor": {
+            "__compat": {
+              "support": {
+                "chrome": {
+                  "version_added": true
+                },
+                "edge": {
+                  "version_added": "79"
+                },
+                "firefox": [
+                  {
+                    "version_added": "107"
+                  },
+                  {
+                    "version_added": "45",
+                    "version_removed": "107",
+                    "partial_implementation": true,
+                    "notes": "Will always return 1, regardless of the actual value."
+                  }
+                ],
+                "firefox_android": {
+                  "version_added": false
+                },
+                "opera": "mirror",
+                "safari": {
+                  "version_added": false
+                },
+                "safari_ios": "mirror"
+              }
+            }
+          },
+          "mode": {
+            "__compat": {
+              "support": {
+                "chrome": {
+                  "version_added": true
+                },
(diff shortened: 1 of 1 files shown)


webextensions/api/tabs.json
```diff
@@ -1243,6 +1243,83 @@
...
+                "firefox": [
+                  {
+                    "version_added": "107"
+                  },
+                  {
+                    "version_added": "45",
+                    "version_removed": "107",
+                    "partial_implementation": true,
+                    "notes": "Will always return 1, regardless of the actual value."
+                  }
```
rebloor, line 1265: I think it might be confusing to consider this feature removed. Elsewhere we handle similar situations with a note along the lines of:
```suggestion
                  {
                    "version_added": "45",
                    "notes": "Until Firefox 107, returns 1 regardless of the default zoom factor setting."
                  }
```

rebloor (changes requested):

webextensions/api/tabs.json
gregorypappas: Thanks for reviewing! Seems like a reasonable change to make. FWIW, my inspiration was the compatibility data for [CSS.supports()](https://developer.mozilla.org/Web/API/CSS/supports#browser_compatibility), is a different approach usually taken for WebExtension compat data or is CSS.supports an anomaly?

rebloor: I balked at the idea of marking it as removed and the [scheme guide isn't forthcoming on a standard] (https://github.com/mdn/browser-compat-data/blob/main/schemas/compat-data-schema.md#partial_implementation). However, I've done some further digging and found this example in the web extension data:
```
              "safari": [
                {
                  "version_added": "15.4",
                  "notes": "Since Safari 15.4, this event is also fired in response to a message from webpages allowed in <a href='https://developer.mozilla.org/docs/Mozilla/Add-ons/WebExtensions/manifest.json/externally_connectable'><code>externally_connectable</code></a>."
                },
                {
                  "version_added": "14",
                  "notes": "This event is only fired in response to a message from an extension's containing app, not webpages nor other extensions."
                }
              ],
```
So, I'm now inclined to think that you were probably along the right lines but I would suggest:
```suggestion
                  {
                    "version_added": "107"
                  },
                  {
                    "version_added": "45",
                    "partial_implementation": true,
                    "notes": "Returns 1, regardless of the default zoom factor setting."
                  }
```

rebloor: I also note that under [`version_removed`](https://github.com/mdn/browser-compat-data/blob/main/schemas/compat-data-schema.md#version_removed) that "If the feature has not been removed from the browser, this property is omitted, rather than being set to false." In this case, the feature was never removed rather it was fully implemented in 107.

gregorypappas: Thanks, your suggestion is perfect!

rebloor (approved):

rebloor:
Many thanks @gregorypappas

PR #17967, opened 2022-10-10
SameSite: Corrections on "schemeful" support status

kiding:
#### Summary

Corrected SameSite `schemeful` statuses for Chromium-based browsers.

* Enforcement rolled out to Chrome 88+ in an out-of-bounds manner
  * Reference: [Chrome Platform Status](https://chromestatus.com/feature/5096179480133632)
  * Chrome for Desktop, Chrome for Android
* Flags in the Chromium source code were changed in 91
    * Reference: [Commit `9b8d9f0`](https://chromium.googlesource.com/chromium/src/+/9b8d9f03aa453b7b39f0bef436009c4310799b0d) ([Diff between Chromium 90 and 91](https://chromium.googlesource.com/chromium/src/+log/90.0.4430.85..91.0.4472.19?n=10000))
* Browsers based on Chromium 91
    * Edge 91
    * Opera 77: [Release notes](https://blogs.opera.com/desktop/2021/04/opera-77-developer/)
    * Opera for Android 64: [Release notes](https://www.apkmirror.com/apk/opera-software-asa/opera/opera-64-0-3282-59726-release/)
    * Samsung Internet 16: [Release notes](https://r2.community.samsung.com/t5/Tech-Talk/Samsung-Internet-Browser-Latest-Version-16-0/td-p/10174163)

#### Related issues

A follow-up on #9352.


queengooborg:
Hey @kiding, thanks for your PR!  However, there's a few discrepancies I'd like to clear up:

The ChromeStatus.com link that you included says that the feature was shipped in Chrome 89, not 88.  Where did you get "88" from?

I'm also confused as to how the "91" plays into the data here.  If the feature was enabled by default in Chromium 91, then Chrome should also be set to 91, right?  Why are only the derivative browsers getting updated but not the main Chromium browser?

Have you also done any testing to verify these version numbers?

kiding:
> The ChromeStatus.com link that you included says that the feature was shipped in Chrome 89, not 88. Where did you get "88" from?

The link says: *"Releasing to Stable, as part of a gradual rollout, starting in M88."* In Google's case, there is an out-of-bounds channel where practically browsers ask Google which flags should be turned on. Google does experiments on it -- for example, 50% of the browsers should have the flag on, etc. 

As an example, see [SameSite Updates](https://www.chromium.org/updates/same-site/) for what it can do.

> Why are only the derivative browsers getting updated but not the main Chromium browser?

It's because how the flags are applied depends on the browser vendor, and the derivative browser vendors usually follow the Chromium source code for the default flag values. 

Of course they can implement their own policy too: for some flags, [Edge turns them on in a later version than Chrome/Chromium.](https://learn.microsoft.com/en-us/microsoft-edge/web-platform/site-impacting-changes)

> Have you also done any testing to verify these version numbers?

Yes!

queengooborg:
> The link says: "Releasing to Stable, as part of a gradual rollout, starting in M88."

I would not trust the accuracy of that statement on ChromeStatus, and in fact I even question the "89" down below (notice how it reads "Estimated milestones").

I'm also skeptical about trusting commits to determine when a feature was enabled or disabled by default.  They can provide a good idea on what version a feature was added into Chrome, but I've found quite often that features can get pushed back or pulled into earlier releases.

---

I'm going to mark this as "info needed" for now because I'd like to do some more research on this (although I don't have much experience testing HTTP headers).

kiding:
While I don't agree, I do agree with the sentiment that testing is the most trustworthy method. That says, it is easy to test this without any server setup.

1. Prepare the target version: If it is a fresh installation, wait for 5~10 mins for the out-of-bounds update to complete, then restart the app.
2. Go to `http://example.com`
3. Turn on DevTools, then execute the following.
```
document.cookie='key=value;SameSite=Strict';
img = document.createElement('img');
img.src = 'https://example.com';
document.body.append(img);
```
4. Look for the image request under the "Network" tab. If the request does not have `Cookie` header, it means the "schemeful" support is enabled for the browser.

#### Examples

* Firefox 107, not enabled by default 
image

* Safari 16.1, not supported
image

* Chrome 107, enabled by default
image

Elchi3:
> Flags in the Chromium source code were changed in 91
> * Reference: [Commit `9b8d9f0`](https://chromium.googlesource.com/chromium/src/+/9b8d9f03aa453b7b39f0bef436009c4310799b0d) ([Diff between Chromium 90 and 91](https://chromium.googlesource.com/chromium/src/+log/90.0.4430.85..91.0.4472.19?n=10000))

Yeah, so this hints towards 91 and his also confirms 91: https://storage.googleapis.com/chromium-find-releases-static/9b8.html#9b8d9f03aa453b7b39f0bef436009c4310799b0d

I think we should set chrome to 91 and leave the rest as "mirror".

Elchi3:
@kiding sorry for the delay here. If you are happy with 91, then I can merge this PR. Let me know!

Elchi3 (approved):
No response, I'm going with 91 here then. Thanks all!

kiding:
Sorry for a delayed response. 

As I have mentioned throughout the PR, Chrome and Chromium can have different flags enabled by default, and this is one of those cases. Chrome turned it on first, and Chromium with other downstream browsers did next.

But I understand the difficulty proving it since Chrome does not provide old versions to test with. I remember when I wrote this PR, I did some testing with [Slimjet’s archive](https://www.slimjet.com/chrome/google-chrome-old-version.php) (limitation: only some versions are being archived) and [APKMirror](https://www.apkmirror.com/apk/google-inc/chrome-canary-unstable/) (limitation: does not reflect the desktop version). They were in line with what I described, but at the same time not exact tests for sure. 

Thanks for all the work!

PR #18004, opened 2022-10-16
Update Safari versions for AudioNode API

queengooborg:
This PR updates and corrects the real values for Safari (Desktop and iOS/iPadOS) for the `AudioNode` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.2.6).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/AudioNode

### DIFF
--- a/api/AudioNode.json
+++ b/api/AudioNode.json
@@ -60,7 +60,7 @@
             "opera": "mirror",
             "opera_android": "mirror",
             "safari": {
-              "version_added": "6"
+              "version_added": "7"
             },
             "safari_ios": "mirror",
             "samsunginternet_android": "mirror",
@@ -98,7 +98,7 @@
             "opera": "mirror",
             "opera_android": "mirror",
             "safari": {
-              "version_added": "6"
+              "version_added": "7"
             },
             "safari_ios": "mirror",
             "samsunginternet_android": "mirror",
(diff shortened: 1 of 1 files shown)


foolip (changes requested):
The custom test for this uses `AnalyserNode`. Since `AudioNode` is the base class for lots of other concrete nodes, it doesn't make sense for support to be later than any of the concrete node types. I think the custom test will need updating here.

queengooborg:
I picked `AnalyserNode` because it was one of the first `AudioNode` interfaces to be supported, along with `BiquadFilter`, `Convolver`, `DynamicsCompressor` and others.  Additionally, for the properties being changed in this PR, Chrome is set to 27 (when it is 14 otherwise), so it doesn't make sense for Safari to be 6 here as well.

I believe that this is a Safari 6.1 case, especially so when I find that the current data was from the wiki migration.

foolip:
@queengooborg sorry, I read the diff sloppily and assumed that the api.AudioNode feature itself was being updated, but it's only members of the `AudioNode` interface. There's nothing suspicious about that.

Indeed https://github.com/mdn/browser-compat-data/pull/18301 changed these same features, so it's all very plausible.

foolip:
Oh, and a tool to summarize changes made by a PR would make it more obvious which entries are being updated, and avoid misreadings like I did here.

foolip (approved):

PR #18021, opened 2022-10-16
Update Safari versions for MutationRecord API

queengooborg:
This PR updates and corrects the real values for Safari (Desktop and iOS/iPadOS) for the `MutationRecord` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.2.6).

### DIFF
--- a/api/MutationRecord.json
+++ b/api/MutationRecord.json
@@ -58,7 +58,7 @@
             "opera": "mirror",
             "opera_android": "mirror",
             "safari": {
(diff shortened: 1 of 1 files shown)


foolip (changes requested):
The test here is just testing the interface object, so it might be wrong. The data should match https://mdn-bcd-collector.appspot.com/tests/api/MutationObserver/takeRecords but doesn't.

PR #18022, opened 2022-10-16
Update Safari versions for api.Navigator.registerProtocolHandler

queengooborg:
This PR updates and corrects the real values for Safari (Desktop and iOS/iPadOS) for the `registerProtocolHandler` member of the `Navigator` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.2.6).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/Navigator/registerProtocolHandler

### DIFF
--- a/api/Navigator.json
+++ b/api/Navigator.json
@@ -2413,7 +2413,8 @@
             },
             "opera_android": "mirror",
             "safari": {
-              "version_added": false
+              "version_added": "5",
+              "version_removed": "5.1"
             },
             "safari_ios": "mirror",
             "samsunginternet_android": "mirror",


foolip (commented):
Did it actually work? It seems likely that it was exposed by accident, but if the OS integration wasn't done, it would need a note. I don't think it's worth digging into this, so just overriding it in the collector would be OK I think.

lucalves (approved):

queengooborg:
I'm not going to worry about this too much.  It's just one single version and eventually we probably won't collect data for this version of Safari anyways.

PR #18031, opened 2022-10-16
RTCRtpEncodingParameters.maxFramerate now part of proper spec

hamishwillee:
`RTCRtpEncodingParameters.maxFramerate` as in an extension spec; it just got promoted to being part of the proper spec. See https://github.com/w3c/webrtc-extensions/issues/108

Other implemented features are intended to move too (as they are implemented).


Elchi3 (approved):
Nice! Only two more features until we can remove the exception in https://github.com/mdn/browser-compat-data/blob/349264727d351be383060918e21b9ea186f0ba02/test/linter/test-spec-urls.ts#L39

PR #18046, opened 2022-10-19
Include render-blocking status browser compat data

abinpaul1:
#### Summary

`renderBlockingStatus` in PerfomanceResourceTiming to indicate the render-blocking nature of a resource. The attribute takes away the burden of having to rely on complex heurestics to determine which resources were actually render-blocking and provides a direct signal from the browser.

#### Related issues

https://github.com/mdn/content/pull/19325

### DIFF
--- a/api/PerformanceResourceTiming.json
+++ b/api/PerformanceResourceTiming.json
@@ -480,6 +480,40 @@
           }
         }
       },
+      "renderBlockingStatus": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/PerformanceResourceTiming/renderBlockingStatus",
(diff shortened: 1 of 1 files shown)


bsmth (approved):
Thanks 👍🏻

bsmth:
Hi @abinpaul1 is this correct for Edge and Opera? I think it's only in Chrome 107, right:

See: 
https://pr19325.content.dev.mdn.mozit.cloud/en-US/docs/Web/API/PerformanceResourceTiming/renderBlockingStatus

---

image

abinpaul1:
Hey @bsmth , apologies for delay in responding. I think its accurate, because all three of them share the same rendering engine Blink and the feature is available from blink 107 onwards. I have confirmed the field is present in Opera 93 and Edge 107.

bsmth:
> I have confirmed the field is present in Opera 93 and Edge 107

OK very good, tnx

PR #18064, opened 2022-10-24
FF101 CSS scroll() function on animation_timeline

hamishwillee:
FF101 added support for the `scroll()` CSS functional notation in https://bugzilla.mozilla.org/show_bug.cgi?id=1737918 behind a preference.

I haven't added data for one of these before, but the `scroll()` appears to only be applicable on an `animation-timeline` as an option, so I have added it there. I would have made the "key" into `scroll()` but the linter wouldn't let me, so I followed the pattern of `font-variant-alternates` and used the description to clarify it as a function - that OK?

Other docs work for this can be tracked in https://github.com/mdn/content/issues/15467

FYI @queengooborg

### DIFF
--- a/css/properties/animation-timeline.json
+++ b/css/properties/animation-timeline.json
@@ -54,7 +54,7 @@
               "edge": "mirror",
               "firefox": {
                 "version_added": "101",
-                "notes": "Zero scroll range is treated as 100% but should be 0% (see <a href='https://github.com/w3c/csswg-drafts/issues/7401'>csswg-drafts#7401</a>).",
+                "notes": "Zero scroll range is treated as 100% but should be 0% (see <a href='https://bugzil.la/1780865'>bug 1780865</a>).",
                 "flags": [
                   {
                     "type": "preference",
--- a/css/properties/animation-timeline.json
+++ b/css/properties/animation-timeline.json
@@ -40,6 +40,48 @@
             "standard_track": true,
             "deprecated": false
           }
+        },
+        "scroll": {
+          "__compat": {
+            "description": "<code>scroll()</code>",
+            "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/animation-timeline/scroll",
+            "spec_url": "https://w3c.github.io/csswg-drafts/scroll-animations-1/#funcdef-scroll",
+            "support": {
+              "chrome": {
+                "version_added": false
+              },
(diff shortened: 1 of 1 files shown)


css/properties/animation-timeline.json
```diff
@@ -40,6 +40,49 @@
...
+            "spec_url": "https://w3c.github.io/csswg-drafts/scroll-animations-1/#funcdef-scroll",
+            "support": {
+              "chrome": {
+                "version_added": false
+              },
+              "chrome_android": "mirror",
+              "edge": "mirror",
+              "firefox": {
+                "version_added": "101",
+                "notes": "Zero scroll range is treated as 100% but should be 0% (see <a href='https://github.com/w3c/csswg-drafts/issues/7401'>csswg-drafts#7401</a>).",
```
hamishwillee, line 57: @queengooborg FYI, this note comes from discussion at https://bugzilla.mozilla.org/show_bug.cgi?id=1737918#c16

Essentially if an element is the source of the scroll timeline, and it does not have a scroll bar in the axis you specify, then the timeline has zero scroll range and is treated as inactive. The spec says that this should indicate 0 progress, but FF currently indicates 100 progress. 

I added the note to record the spec incompatibility, but it still isn't clear whether FF plans to revert or Chrome does something different or whatever. So for now, just noting the spec difference. When the FF engineer gets back from holiday sometime next month might update this.

lucalves (approved):

hamishwillee:
@queengooborg Can we please merge this?

hamishwillee:
Note, Chrome also implements this behind a preference, but I don't know how to work out what that preference IS - the associated bug is https://bugs.chromium.org/p/chromium/issues/detail?id=1317765

queengooborg:
> Note, Chrome also implements this behind a preference, but I don't know how to work out what that preference IS - the associated bug is https://bugs.chromium.org/p/chromium/issues/detail?id=1317765

It might actually not -- https://bugs.chromium.org/p/chromium/issues/detail?id=1317765#c53 says that "Making animation-delay a shorthand based on a runtime flag is a bit complicated since our code generation does not support it. That work is therefore deferred to a later CL."  But it's not a big concern to me to research, if someone else wants to dig deeper and make a PR, I won't say no!  ;P

queengooborg (approved):

PR #18092, opened 2022-10-27
add InkPresenter interface

chrisdavidmills:
#### Summary

This PR adds the Ink API's [InkPresenter](https://wicg.github.io/ink-enhancement/#ink-presenter) interface to BCD.

The other features of the API were already in BCD, so I just copied the support data from them. It looks fairly accurate tho — the Chrome folks have confirmed to me that it is definitely supported in Chrome 94 and above, and I also tested it in Edge and found it to be working 

See [Enhancing Inking on the Web](https://blogs.windows.com/msedgedev/2021/08/18/enhancing-inking-on-the-web/) for more explanation and a demo.

### DIFF
--- a/api/InkPresenter.json
+++ b/api/InkPresenter.json
@@ -1,0 +1,144 @@
+{
+  "api": {
+    "InkPresenter": {
+      "__compat": {
+        "spec_url": "https://wicg.github.io/ink-enhancement/#ink-presenter",
+        "support": {
+          "chrome": {
+            "version_added": "94"
+          },
+          "chrome_android": "mirror",
+          "edge": {
+            "version_added": "93"
+          },
+          "firefox": {
+            "version_added": false
+          },
+          "firefox_android": "mirror",
+          "ie": {
+            "version_added": false
+          },
+          "oculus": "mirror",
+          "opera": "mirror",
+          "opera_android": "mirror",
+          "safari": {
+            "version_added": false
+          },
+          "safari_ios": "mirror",
+          "samsunginternet_android": "mirror",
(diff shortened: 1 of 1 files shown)


api/InkPresenter.json
```diff
@@ -0,0 +1,144 @@
...
+    "InkPresenter": {
+      "__compat": {
+        "spec_url": "https://wicg.github.io/ink-enhancement/#ink-presenter",
+        "support": {
+          "chrome": {
+            "version_added": "94"
+          },
+          "chrome_android": "mirror",
+          "edge": {
+            "version_added": "93"
```
teoli2003, line 12: This is weird. How is it that the Edge version is before the Chrome version? (Here and for the other ink).

@queengooborg Can we test this?

teoli2003 (changes requested):

api/InkPresenter.json
chrisdavidmills: It is a bit strange...maybe it should just be 94.

queengooborg: Microsoft led the development of Ink APIs so Edge getting the feature early actually does make sense.  I think the latest mdn-bcd-collector version has the IDL for this interface, so if desired, we can confirm the data through it!

(By the way, welcome back Chris, it's great to see you again!)

chrisdavidmills: > (By the way, welcome back Chris, it's great to see you again!)

👋 ❤️ !

Nice to see you too ;-)

teoli2003 (approved):
I think your explanation… explain why. Thanks @queengooborg

PR #18094, opened 2022-10-27
fix: broken code tag

victor-homyakov:
#### Summary

The `<code>` tag was missing the first two chars, resulting in broken HTML markup


queengooborg (approved):
Hi there, thank you for your PR and welcome to BCD!  I'm surprised our HTML validation linter didn't catch this, hopefully migrating to Markdown in the future will help reduce issues like this!

PR #18096, opened 2022-10-28
FF106 - i18n.detectLanguage supported on Android

hamishwillee:
FF106 fixes a long standing bug that `i18n.detectLanguage` existed on Android but always returned an error in https://bugzilla.mozilla.org/show_bug.cgi?id=1712214#c5

As per that bug, effectively Android support for this exists from FF106 and not 48 as indicated by the data.

### DIFF
--- a/webextensions/api/i18n.json
+++ b/webextensions/api/i18n.json
@@ -38,7 +38,7 @@
                 "version_added": "47"
               },
               "firefox_android": {
-                "version_added": "48"
+                "version_added": "106"
               },
               "opera": "mirror",
               "safari": {


queengooborg:
Since the method existed but just returned an error, we should document that like so:

```json
{
  "firefox_android": [
    {
      "version_added": "106"
    },
    {
      "version_added": "48",
      "version_removed": "106",
      "notes": "The method existed but always threw an error."
    }
  ]
}
```

hamishwillee:
Thanks very much @queengooborg - fixed.

queengooborg (approved):

PR #18101, opened 2022-10-28
Remove percentage subfeature from word-spacing property

wbamberg:
I'm not 100% sure about this but I think it is probably the right thing to do.

BCD currently has a subfeature for [`word-spacing`](https://developer.mozilla.org/en-US/docs/Web/CSS/word-spacing) taking a percentage. The subfeature has only Firefox supporting it. All that is still true.

However in 2019 percentages were removed as a value for this property (https://www.w3.org/TR/2019/WD-css-text-3-20191113/#changes-2018-12-06), see also https://github.com/mdn/content/pull/21861. 

So it seemed better to remove the subfeature, since it's no longer specced behavior. We could I suppose add a note to the main feature, like "Firefox also supports percentage values", but I wasn't sure it is worth it.

### DIFF
--- a/css/properties/word-spacing.json
+++ b/css/properties/word-spacing.json
@@ -38,6 +38,39 @@
             "deprecated": false
           }
         },
+        "percentages": {
+          "__compat": {
+            "description": "<code>&lt;percentage&gt;</code> values",
+            "support": {
+              "chrome": {
+                "version_added": false
+              },
+              "chrome_android": "mirror",
+              "edge": "mirror",
+              "firefox": {
+                "version_added": "45"
+              },
+              "firefox_android": "mirror",
+              "ie": {
+                "version_added": false
+              },
+              "oculus": "mirror",
+              "opera": "mirror",
+              "opera_android": "mirror",
+              "safari": {
+                "version_added": "7"
+              },
+              "safari_ios": "mirror",
+              "samsunginternet_android": "mirror",
+              "webview_android": "mirror"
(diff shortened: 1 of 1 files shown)


css/properties/word-spacing.json
```diff
@@ -38,39 +38,6 @@
...
-                "version_added": "7"
-              },
-              "safari_ios": "mirror",
-              "samsunginternet_android": "mirror",
-              "webview_android": "mirror"
-            },
-            "status": {
-              "experimental": false,
-              "standard_track": true,
-              "deprecated": false
```
teoli2003, line 70: I think we should keep it but set `standard_track` and `deprecated` to (respectively) `false` and `true`.

Also Safari seems to support it according to this data.

teoli2003 (changes requested):

css/properties/word-spacing.json
queengooborg: I agree with this -- browsers still support this so we shouldn't remove it from BCD just yet.

wbamberg: Yes, that does sound better :).

queengooborg (approved):

OnkarRuikar:
I think it's coming back: [[css-text-4] Add percentages to letter-spacing and word-spacing.](https://github.com/w3c/csswg-drafts/commit/67d5372594560e69d0ed9a3f9f021c61790062db)

PR #18114, opened 2022-10-31
Minor: Content-Security-Policy.json - restrict notes to versions.

jeroenhabets:
Restrict "Before Firefox 5#" note to before 5# versions. (The notes alerted me unnecessarily :) )

#### Summary

Minor usability change, now these couple of notes should not appear for recent versions.

Not tested but I did validate the JSON


jeroenhabets:
I've read [compat-data-schema.md#the-support_statement-object](https://github.com/mdn/browser-compat-data/blob/main/schemas/compat-data-schema.md#the-support_statement-object)  but cannot see why this array is not accepted here (while it is in other places)...

queengooborg:
Hi @jeroenhabets -- the reason that the linter is failing from those statements is because an array cannot have multiple statements with just `version_added` and an optional note.  Other entries in an array must be marked as a prefixed/alt. name statement or partial implementation.

With that being said, I'm wondering if the statements with the notes should be marked as partial implementation...

jeroenhabets:
@queengooborg thank you for your suggestion! In lieu of the notes marking the earlier versions a partial implementation is applicable.

NOTE: Even after a 2nd scan AFAICT this requirement for version array's is not deducible from [compat-data-schema.md#the-support_statement-object](https://github.com/mdn/browser-compat-data/blob/main/schemas/compat-data-schema.md#the-support_statement-object).

queengooborg:
> NOTE: Even after a 2nd scan AFAICT this requirement for version array's is not deducible from [compat-data-schema.md#the-support_statement-object](https://github.com/mdn/browser-compat-data/blob/main/schemas/compat-data-schema.md#the-support_statement-object).

The constraints on arrays are too complex to put in the JSON schema, so instead we perform checks using a JavaScript linter (see `test/linter/test-version.ts`).  Basically, we have some logic that prevents a contributor from adding an array that is simply a bunch of `version_added` statements with optional notes, and instead requires that all minus one statement in an array must have some special statement (`partial_implementation`, `prefix`, `version_removed`, etc.) added to it as well.  This linter rule was added because certain contributors used multiple statements with notes to represent support for subfeatures, rather than simply creating subfeatures, which looked ugly and wasn't machine-friendly.

> @queengooborg thank you for your suggestion! In lieu of the notes marking the earlier versions a partial implementation is applicable.

Thank you!  After reading the notes again, `partial_implementation` makes a lot more sense for these statements!

queengooborg (approved):
Thank you for your work on this PR, this is looking good to me!  Welcome to BCD!  🎉

PR #18126, opened 2022-11-02
Add Link header specification

sofalse:
#### Summary

Added Link header specification, based on RFC8288 and to be used with 103 Early Hint status.

RFC8288 added in browser-specs@3.32.0

As references used the [test page](https://early-hints.fastlylabs.com/), [chromium docs](https://chromium.googlesource.com/chromium/src/+/master/docs/early-hints.md#using-origin-trial) and [RFC8288](https://httpwg.org/specs/rfc8288.html#header).


http/headers/Link.json
```diff
@@ -0,0 +1,42 @@
...
+          "mdn_url": "https://developer.mozilla.org/docs/Web/HTTP/Headers/Link",
+          "spec_url": "https://httpwg.org/specs/rfc8288.html#header",
+          "support": {
+            "chrome": {
+              "version_added": "103"
+            },
+            "chrome_android": "mirror",
+            "edge": {
+              "version_added": null
+            },
```
queengooborg, line 15:
```suggestion
            "edge": "mirror",
```

http/headers/Link.json
```diff
@@ -0,0 +1,42 @@
...
+            "firefox": {
+              "version_added": null
+            },
+            "firefox_android": "mirror",
+            "ie": {
+              "version_added": null
+            },
```
queengooborg, line 22:
```suggestion
            "ie": {
              "version_added": false
            },
```

queengooborg (changes requested):
Hey there, thank you for your PR and welcome to BCD!

I'm unable to verify that the test actually works properly, as the test states that it passes as far back as Chrome 30 and Firefox 32.  This doesn't match what [Chrome Status](https://chromestatus.com/feature/5207422375297024) reports, nor the documentation you've linked, and it appears that the test is just bad altogether because it doesn't actually test for the status code's support in browsers:

> Passing means that the client doesn't break when receiving a non-final status code like 103; it doesn't mean that they'll actually do anything with the hints.

But regardless of the test's functionality, if ChromeStatus says 103 then I'll go with that number.  👍   We can also mirror this to Edge (since it's a Chromium fork) and set IE to `false` because there's no way that IE can support a new feature like this!

sofalse:
Hi @queengooborg, thanks for the review!
I obviously agree to the suggested changes and looking forward for further feedback or an approve.

PR #18133, opened 2022-11-03
Update CSS <length> viewport units in Chrome 108

rachelandrew:
Chrome 108 has support sv* units, lv* units, dv* units and the logical vi/vb units.

https://chromestatus.com/feature/5170718078140416

### DIFF
--- a/css/types/length.json
+++ b/css/types/length.json
@@ -340,7 +340,7 @@
             "description": "<code>vb</code> unit",
             "support": {
               "chrome": {
-                "version_added": false
+                "version_added": "108"
               },
               "chrome_android": "mirror",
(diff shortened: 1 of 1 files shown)


queengooborg:
It looks like Chrome 108 is just an "estimated milestone" and not a guaranteed version.  I think we should hold off on making these changes until Chrome 108 is confirmed, as the version number can change unexpectedly.

(I wish ChromeStatus made it more clear whether it's guaranteed or estimated; it's just a small bit of text that's way too easy to miss.)

queengooborg:
Ah whoops, looks like I had merged similar PRs already, so I'll merge this one too!

PR #18138, opened 2022-11-04
FF107 ships CSS contains-intrinsic-* properties

hamishwillee:
FF107 ships the CSS `contains-intrinsic-*` properties in https://bugzilla.mozilla.org/show_bug.cgi?id=1792886 (these were added in FF104 in #17890).

This changes the release to 107 for the properties, and marks them as no longer experimental.

@queengooborg Given that we now don't keep preference after a feature is released I removed the pref version. Is this the right thing to do? I mean it will be correct in 2 weeks when FF107 releases, but should I instead keep pref version and have a separate entry for FF107?

Other docs work for this tracked in https://github.com/mdn/content/issues/21274

### DIFF
--- a/css/properties/contain-intrinsic-block-size.json
+++ b/css/properties/contain-intrinsic-block-size.json
@@ -12,14 +12,7 @@
             "chrome_android": "mirror",
             "edge": "mirror",
             "firefox": {
-              "version_added": "104",
-              "flags": [
-                {
-                  "type": "preference",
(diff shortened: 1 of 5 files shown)


queengooborg (approved):
This looks good to me -- technically, our guidelines say that the flag should be removed once the feature is enabled in a _stable_ release, but since the release in question is just one release away, I'm not too worried about it!

PR #18169, opened 2022-11-07
Change preview -> false for outdated preview values

queengooborg:
I decided to verify all of the features that are set to preview to see if they’re still accurate or if they were left behind.  Turns out, nearly EVERY instance of `version_added: preview` was outdated and simply incorrect.  Most of the ones for Safari were added a year ago and never touched since, and hardly any of the features marked as preview for Firefox were supported in the latest nightly build.

This PR is the result of this verification.  This PR sets all of the invalid `preview` values to `false` for all categories, and then updates derivative browsers to mirror where applicable.  The Temporal JS API was left alone in this PR as I am not yet sure how to test it.

### DIFF
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -1287,11 +1287,9 @@
             "chrome_android": "mirror",
             "edge": "mirror",
             "firefox": {
-              "version_added": "preview"
-            },
-            "firefox_android": {
               "version_added": false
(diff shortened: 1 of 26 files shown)


queengooborg:
I'm going to merge this before any merge conflicts occur, and since nobody seems to be available to review this.

PR #18177, opened 2022-11-08
feat(js): update ICU 69 `trailingZeroDisplay` compat data

VictorWesterlund:
#### Summary

Chromium 106 and WebKit 15.6 (?) adds support for the `options.trailingZeroDisplay` parameter in `Intl.NumberFormat()`.

**Chromium ref:** https://bugs.chromium.org/p/v8/issues/detail?id=12550

**Webkit: ??**

Unable to find supporting details that this has left experimental on WebKit. https://bugs.webkit.org/show_bug.cgi?id=215438 is still open. But it seems to work in the later versions of Safari:

Tested on Safari|Works
--|--
16|✅
15.6|✅
14.1|❌

I currently don't have a way to test the versions between 14.1 and 15.6. Perhaps someone with more info can chip in.

#### Related issues

N/A

### DIFF
--- a/javascript/builtins/Intl/NumberFormat.json
+++ b/javascript/builtins/Intl/NumberFormat.json
@@ -530,7 +530,7 @@
                   "opera": "mirror",
                   "opera_android": "mirror",
                   "safari": {
-                    "version_added": false
+                    "version_added": "15.4"
                   },
                   "safari_ios": "mirror",
                   "samsunginternet_android": "mirror",
--- a/javascript/builtins/Intl/NumberFormat.json
+++ b/javascript/builtins/Intl/NumberFormat.json
@@ -527,9 +527,7 @@
                   "nodejs": {
                     "version_added": false
                   },
-                  "opera": {
-                    "version_added": false
-                  },
+                  "opera": "mirror",
                   "opera_android": "mirror",
                   "safari": {
                     "version_added": false
--- a/javascript/builtins/Intl/NumberFormat.json
+++ b/javascript/builtins/Intl/NumberFormat.json
@@ -510,7 +510,7 @@
                 "description": "<code>options.trailingZeroDisplay</code> parameter",
                 "support": {
                   "chrome": {
-                    "version_added": false
+                    "version_added": "106"
                   },
                   "chrome_android": "mirror",
                   "deno": {
@@ -527,7 +527,9 @@
                   "nodejs": {
                     "version_added": false
                   },
-                  "opera": "mirror",
+                  "opera": {
+                    "version_added": false
+                  },
                   "opera_android": "mirror",
                   "safari": {
                     "version_added": false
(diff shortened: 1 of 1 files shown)


javascript/builtins/Intl/NumberFormat.json
```diff
@@ -527,7 +527,9 @@
                   "nodejs": {
                     "version_added": false
                   },
-                  "opera": "mirror",
+                  "opera": {
+                    "version_added": false
+                  },
```
queengooborg, line 532: Opera should still mirror its data from Chrome; I would be VERY surprised if a core JS feature like this was intentionally disabled in Opera.
```suggestion
                  "opera": "mirror",
```

javascript/builtins/Intl/NumberFormat.json
```diff
@@ -510,7 +510,7 @@
                 "description": "<code>options.trailingZeroDisplay</code> parameter",
                 "support": {
                   "chrome": {
-                    "version_added": false
+                    "version_added": "106"
```
queengooborg, line 513: Did you test this?  If so, could you provide your testing code so we may verify?

queengooborg (changes requested):
Hi there, thank your PR and welcome to BCD!  I'm not seeing anything in the linked bug that indicates the feature is supported by default in Chrome, but [comment 4](https://bugs.chromium.org/p/v8/issues/detail?id=12550#c4) suggests that a runtime flag can be passed to enable the feature.

Regarding Safari's support, would you mind opening up an issue with your findings?  I can follow up using the VMs I've set up and pinpoint the exact version number!

javascript/builtins/Intl/NumberFormat.json
VictorWesterlund: You're absolutely right, my apologies. Must've screwed up my initial testing somehow.

javascript/builtins/Intl/NumberFormat.json
VictorWesterlund: Sure, [here's a page for that](https://cattegat.github.io/test-trailingzerodisplay/).

```js
Intl.NumberFormat(navigator.language, {
    style: "currency",
    currency: "SEK",
    currencyDisplay: "symbol",
    trailingZeroDisplay: "stripIfInteger"
}).format(123.00);
```

VictorWesterlund:
Thanks, glad to be here!

> I'm not seeing anything in the linked bug that indicates the feature is supported by default in Chrome

I'm also quite puzzled by this. That was the only reference I could find. Not sure what to make of it since I can very clearly see that it works (no flags enabled).

> Regarding Safari's support, would you mind opening up an issue with your findings?

Will do! Thank you.

queengooborg:
Thanks for linking the test code!  I was able to confirm that support is present in Chrome 106 and not 105 using the code.  I also was able to confirm support was added in Safari 15.4 as well, if you'd like to add that in!

VictorWesterlund:
There we go!

queengooborg (approved):

PR #18189, opened 2022-11-10
add MSE in workers data

chrisdavidmills:
#### Summary

This PR adds the required new BCD for Chrome's additions to Media Source Extensions, which allows them to be used from inside dedicated workers.

See my research document for more details: https://docs.google.com/document/d/1_zwIr7c9_O3_PEoAnaS5M7yExx9VtjMSUJ6U5yiEias/edit#

#### Notes

* To make it so that the linter will pass the data, I have changed the support for the following members to "supported by default in Chrome 70". I don't actually know what the version is, and will ask some of the Chrome folk if they know.
  * `SourceBuffer.audioTracks`
  * `SourceBuffer,textTracks`
  * `SourceBuffer.videoTracks`
* I didn't add "supported in workers" data to the following members, as they were false in Chrome, and the data pattern suggested that they were experimental features in other browsers that didn't make it?
  * `SourceBuffer.appendBufferAsync` - experimental Fx feature?
  * `SourceBuffer.appendStream` - experimental feature supported in pre-Chromium Edge?
  * `SourceBuffer.removeAsync` - experimental Fx feature?
* I tried to make the notes for these subfeatures say "Available in dedicated workers", AIUI they are available in dedicated, but not shared or service workers, but the linter won't let me. How should I deal with this detail instead?

### DIFF
--- a/api/MediaSource.json
+++ b/api/MediaSource.json
@@ -294,6 +294,40 @@
           }
         }
       },
+      "canConstructInDedicatedWorker": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/MediaSource/canConstructInDedicatedWorker",
+          "spec_url": "https://w3c.github.io/media-source/#dom-mediasource-canconstructindedicatedworker",
+          "support": {
+            "chrome": {
+              "version_added": "108"
+            },
+            "chrome_android": "mirror",
+            "edge": "mirror",
+            "firefox": {
+              "version_added": false
+            },
+            "firefox_android": "mirror",
+            "ie": {
+              "version_added": false
+            },
+            "oculus": "mirror",
+            "opera": "mirror",
+            "opera_android": "mirror",
+            "safari": {
+              "version_added": false
+            },
+            "safari_ios": "mirror",
+            "samsunginternet_android": "mirror",
+            "webview_android": "mirror"
+          },
+          "status": {
+            "experimental": true,
+            "standard_track": true,
+            "deprecated": false
+          }
+        }
+      },
       "clearLiveSeekableRange": {
         "__compat": {
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/MediaSource/clearLiveSeekableRange",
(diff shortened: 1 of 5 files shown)


queengooborg (approved):
LGTM, thank you!

chrisdavidmills:
@queengooborg yay! ❤️

wolenetz:
These are not supported in Chrome yet (their implementation is incomplete and experimental, behind the blink::RuntimeEnabledFeature "AudioVideoTracks" which is marked experimental still):
```
SourceBuffer.audioTracks
SourceBuffer.videoTracks
HTMLMediaElement.audioTracks
HTMLMediaElement.videoTracks
```
And Chrome does not have support at all for `SourceBuffer.textTracks` - the partial experimental implementation of MSE and progressive src= inband text track sourcing was removed due to it never stabilizing and over time had diverged to become less stable. It might be added back in future, but currently there is no IDL binding, even experimental, for SourceBuffer.textTracks.

wolenetz:
Is it true that MSE-in-Workers is available in Opera 93? I looked at that changelog and it didn't indicate anything to me on skimming it that related to MSE-in-Workers. I ask because the BCD table for at least MediaSource has Opera version 93 as the version supporting MSE-in-Workers.

wolenetz:
Please revert the portion of this BCD data that is specific to VideoPlaybackQuality: that portion of the API is specific to the HTMLMediaElement, which itself is only available on the main thread context. There is no worker HTMLMediaElement in Chrome (or on any browser); only a main thread/context HTMLMediaElement. It's unfortunate that originally MSE specification was used to get VideoPlaybackQuality specified (I think it was due to spec workgroups being more amenable in W3C for this change at the time long ago). More recently, the VideoPlaybackQuality specification (Media Playback Quality, as referenced on MDN) is precise about this API only being exposed on Window context (not worker).

queengooborg:
Hi @wolenetz -- rather than comment on an already-merged pull request, would you mind opening a new issue and linking to this PR instead?  This will allow us to better track what potentially needs to be done to correct the data, as threads on closed PRs will most certainly get lost in everyone's inboxes and forgotten.  :)

wolenetz:
I filed #18220

PR #16222, opened 2022-05-10
Update Edge versions for IdleDetector API

queengooborg:
This PR updates and corrects the real values for Microsoft Edge for the `IdleDetector` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v6.0.1).

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/IdleDetector

_Check out the [collector's guide on how to review this PR](https://github.com/foolip/mdn-bcd-collector#reviewing-bcd-changes)._

PR #18196, opened 2022-11-11
Add `idle-detection` to `http.headers.Feature-Policy`

ddbeck:
#### Summary

Complete the features required to represent all of the features implied by the Idle Detection spec.

I didn't actually test this. The data here is cribbed from the existing `api.IdleDetector` features, but I did a little independent verification via:

- https://chromestatus.com/feature/4590256452009984
- https://storage.googleapis.com/chromium-find-releases-static/487.html#487196e61e193ee2511231e3cbe4229c26687ec2

#### Related issues

- https://github.com/mdn/browser-compat-data/issues/9556 — There's whole _thing_ here about how Permissions Policy (PP) is the new Feature Policy (FP). Nominally, _this_ feature ought to be a PP feature too, but that's a can of worms affecting many, many features and I refuse to do any more than acknowledge that fact here. A close inspection of the links above reveal that `idle-detection` is enabled for both PP and FP, so this PR is merely incomplete, not wrong.

### DIFF
--- a/http/headers/Feature-Policy.json
+++ b/http/headers/Feature-Policy.json
@@ -509,6 +509,43 @@
             }
           }
         },
+        "idle-detection": {
+          "__compat": {
+            "mdn_url": "https://developer.mozilla.org/docs/Web/HTTP/Headers/Feature-Policy/idle-detection",
+            "spec_url": "https://wicg.github.io/idle-detection/#api-permissions-policy",
+            "support": {
+              "chrome": {
+                "version_added": "94"
+              },
+              "chrome_android": "mirror",
+              "edge": {
+                "version_added": "94",
+                "version_removed": "96"
+              },
+              "firefox": {
+                "version_added": false
+              },
+              "firefox_android": "mirror",
+              "ie": {
+                "version_added": false
+              },
+              "oculus": "mirror",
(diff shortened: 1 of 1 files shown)


lucalves (approved):

http/headers/Feature-Policy.json
```diff
@@ -509,6 +509,43 @@
...
+            "mdn_url": "https://developer.mozilla.org/docs/Web/HTTP/Headers/Feature-Policy/idle-detection",
+            "spec_url": "https://wicg.github.io/idle-detection/#api-permissions-policy",
+            "support": {
+              "chrome": {
+                "version_added": "94"
+              },
+              "chrome_android": "mirror",
+              "edge": {
+                "version_added": "94",
+                "version_removed": "96"
```
queengooborg, line 523: Where did this `version_removed` come from?  Could you link the source for this?

ddbeck: You're the source for this: https://github.com/mdn/browser-compat-data/pull/16222#issuecomment-1128636469

queengooborg: Ah right!  Thanks for the link!

queengooborg (approved):

PR #18199, opened 2022-11-12
JavaScript block statement: verified working in IE8.

MasterInQuestion:

### DIFF
--- a/javascript/statements.json
+++ b/javascript/statements.json
@@ -127,7 +127,7 @@
             },
             "firefox_android": "mirror",
             "ie": {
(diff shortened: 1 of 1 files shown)


queengooborg:
Hi there, we have actually frozen IE data a while back and are not accepting changes involving IE.  Thank you anyways.

MasterInQuestion:
Does not such much defied the purpose of even maintaining the browser-compat-data anyway?
    Thank you anyway.

MasterInQuestion (commented):
Also verified working in IE6.

PR #18204, opened 2022-11-14
Array group()/groupByToMap() available behind a flag

hamishwillee:
From FF108 the `Array.group()` and `groupByToMap()` are not enabled by default https://bugzilla.mozilla.org/show_bug.cgi?id=1799619

On further investigation there was a flag introduced in FF98, that was enabed in nightly prior to 108. So all I've done is added that, and omitted the fact this was ever enabled in preview.

FYI @queengooborg


hamishwillee:
@queengooborg Can you please merge this. It was removed from nightly in FF108 (while the API gets looked at), but there is actually a preference that should have been added much earlier.

PR #18205, opened 2022-11-14
Ff018 csp script src

hamishwillee:
FF108 ships the CSP HTTP header directives `script-src-elem`, `script-src-attr`, `style-src-elem` and style-src-attr` - https://bugzilla.mozilla.org/show_bug.cgi?id=1790621.

This changes "preview" release to "108". Also removes the preferenced version from 108.
- I assume it is OK to lose the information that this is enabled by default in Nightly. 
- I left the preference stuff in there. Release isn't until this time next month, so I'd rather leave it in so people can have the preference if they need it.

Other docs work can be tracked in https://github.com/mdn/content/issues/22113

FYI @queengooborg

### DIFF
--- a/http/headers/Content-Security-Policy.json
+++ b/http/headers/Content-Security-Policy.json
@@ -1079,7 +1079,7 @@
               "edge": "mirror",
               "firefox": [
                 {
-                  "version_added": "preview"
+                  "version_added": "108"
                 },
                 {
                   "version_added": "105",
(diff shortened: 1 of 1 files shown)


queengooborg (changes requested):
Let's remove the flag data while we're here!

hamishwillee:
@queengooborg To me it doesn't make sense to remove these before the release is out - we should use tooling to sweep the data afterwards. As per Please see my comments on https://github.com/mdn/browser-compat-data/pull/18081#issuecomment-1319374501

hamishwillee:
@queengooborg Updated as requested. I'm not particularly concerned about this one being pulled out.

queengooborg (approved):

Issue #18220, opened 2022-11-15
MSE-in-Workers BCD data update PR #18189 contained a few items needing correction


wolenetz:
### What type of issue is this?

Incorrect support data (ex. Chrome says "86" but support was added in "40")

### What information was incorrect, unhelpful, or incomplete?

I was a bit late on completing my review of PR #18189, so unfortunately it landed with a few items that I found only after it merged.
(Below is copied verbatim from the PR #18189 conversations where I originally noted the issues on my eventual review of the PR):

https://github.com/mdn/browser-compat-data/pull/18189#issuecomment-1315936925:
These are not supported in Chrome yet (their implementation is incomplete and experimental, behind the blink::RuntimeEnabledFeature "AudioVideoTracks" which is marked experimental still):
```
SourceBuffer.audioTracks
SourceBuffer.videoTracks
HTMLMediaElement.audioTracks
HTMLMediaElement.videoTracks
```
And Chrome does not have support at all for `SourceBuffer.textTracks` - the partial experimental implementation of MSE and progressive src= inband text track sourcing was removed due to it never stabilizing and over time had diverged to become less stable. It might be added back in future, but currently there is no IDL binding, even experimental, for `SourceBuffer.textTracks`.

https://github.com/mdn/browser-compat-data/pull/18189#issuecomment-1315949977:
Also, for at least the "MediaSource" (https://developer.mozilla.org/en-US/docs/Web/API/MediaSource) page, there are two "Available in workers" rows of BCD table data at the very bottom, with none following the ctor at the top - was one of those two rows meant for the ctor, or was it extra - it's confusing to me.

https://github.com/mdn/browser-compat-data/pull/18189#issuecomment-1315952913:
Is it true that MSE-in-Workers is available in Opera 93? I looked at that changelog and it didn't indicate anything to me on skimming it that related to MSE-in-Workers. I ask because the BCD table for at least MediaSource has Opera version 93 as the version supporting MSE-in-Workers.

https://github.com/mdn/browser-compat-data/pull/18189#issuecomment-1315959788:
[...]

Issue #18224, opened 2022-11-16
Chrome 108 should mirror to Opera 94, but it doesn't


chrisdavidmills:
In https://developer.mozilla.org/en-US/docs/Web/API/MediaSource#browser_compatibility, The Chrome 108 data mirrors Opera support to version 93, but I've just checked and Opera 94 is based on Chrome 108.

chrisdavidmills:
Hrm, I'm confused now. Looking at https://github.com/mdn/browser-compat-data/blob/main/browsers/opera.json, Opera 94's engine version is set to 108, which is where I'm assuming where the mirror data gets taken from...so how come version 93 is mirrored in the version referenced above?

Is this because the new Opera version was only just added... see https://github.com/mdn/browser-compat-data/commit/bc145de20f2180455f37c73ad2b490492d676799 ?

queengooborg:
I believe this is because of the way that the mirroring script handles beta/nightly releases, see https://github.com/mdn/browser-compat-data/blob/b626adc774876722f2e49fa6f2e10a0989c2883a/scripts/release/mirror.ts#L108-L111.  Since Chrome 108 and Opera 93 are both the beta release of the browser, it's mapping them together.  This was done because of Safari iOS beta releases, I believe, but now that we have `preview`, we no longer need this, so I'll submit a PR to fix this!

chrisdavidmills:
@queengooborg this is great, thanks a lot!

PR #18225, opened 2022-11-16
add fixes for wolenetz comments

chrisdavidmills:
#### Summary

This PR adds fixes in response to the review comments provided by @wolenetz in https://github.com/mdn/browser-compat-data/issues/18220

I did the following:

* Removed the separate `worker_support` items from the top-level interface support items, as the extra rows in the support tables were confusing and unnecessary
* Marked `audioTracks` and `videoTracks` support as experimental, for `SourceBuffer` and `HTMLMediaElement`
* Marked `SourceBuffer.textTracks` support as `false` for Chromium
* Reverted `VideoPlaybackQuality` changes
* Verified that MSE-in-workers works in Opera 94, not Opera 93, by testing @wolenetz's demo in beta and developer versions. To fix this, I'm not going to change every Opera support number individually, as that is not maintainable, and the linter complained. Instead, I'm going to fix the mirroring code so that it reports the correct Opera version for Chrome 106/107/108. I've filed an issue for that at https://github.com/mdn/browser-compat-data/issues/18224, which I'll have a go at fixing next.

Testing was done using the demo available at https://wolenetz.github.io/mse-in-workers-demo/mse-in-workers-demo.html

#### Related issues

Fixes https://github.com/mdn/browser-compat-data/issues/18220

### DIFF
--- a/api/HTMLMediaElement.json
+++ b/api/HTMLMediaElement.json
@@ -2871,17 +2871,20 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/API/HTMLMediaElement/srcObject",
           "spec_url": "https://html.spec.whatwg.org/multipage/media.html#dom-media-srcobject-dev",
           "support": {
-            "chrome": {
-              "version_added": "52",
-              "partial_implementation": true,
-              "notes": "Only supports <code>MediaStream</code> objects (see <a href='https://crbug.com/506273'>bug 506273</a>)."
-            },
+            "chrome": [
+              {
+                "version_added": "108",
+                "partial_implementation": true,
+                "notes": "Support added for <code>MediaSource</code> objects transferred from dedicated workers via <code>MediaSource.handle</code> (see <a href='https://crbug.com/878133'>bug 878133</a>)."
+              },
+              {
+                "version_added": "52",
+                "partial_implementation": true,
+                "notes": "Support added for <code>MediaStream</code> objects (see <a href='https://crbug.com/506273'>bug 506273</a>)."
+              }
+            ],
             "chrome_android": "mirror",
-            "edge": {
-              "version_added": "12",
-              "partial_implementation": true,
-              "notes": "Only supports <code>MediaStream</code> objects (see <a href='https://crbug.com/506273'>bug 506273</a>)."
-            },
+            "edge": "mirror",
             "firefox": [
               {
                 "version_added": "42",
@@ -2906,11 +2909,7 @@
             },
             "safari_ios": "mirror",
             "samsunginternet_android": "mirror",
-            "webview_android": {
-              "version_added": "52",
-              "partial_implementation": true,
-              "notes": "Only supports <code>MediaStream</code> objects (see <a href='https://crbug.com/506273'>bug 506273</a>."
-            }
+            "webview_android": "mirror"
           },
           "status": {
             "experimental": false,
--- a/api/HTMLMediaElement.json
+++ b/api/HTMLMediaElement.json
@@ -2875,7 +2875,7 @@
               {
                 "version_added": "108",
                 "partial_implementation": true,
-                "notes": "Support added for <code>MediaSource</code> objects transferred from dedicated workers via <code>MediaSource.handle</code> (see <a href='https://crbug.com/878133'>bug 878133</a>)."
+                "notes": "Support added for <code>MediaSourceHandle</code> objects transferred from dedicated workers where they were obtained from <code>MediaSource.handle</code> (see <a href='https://crbug.com/878133'>bug 878133</a>)."
               },
               {
                 "version_added": "52",
--- a/api/HTMLMediaElement.json
+++ b/api/HTMLMediaElement.json
@@ -153,8 +153,8 @@
               "flags": [
                 {
                   "type": "preference",
-                  "name": "enable-experimental-web-platform-features",
-                  "value_to_set": "enabled"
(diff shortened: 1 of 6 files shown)


api/HTMLMediaElement.json
```diff
@@ -2871,17 +2871,20 @@
...
-            "chrome": {
-              "version_added": "52",
-              "partial_implementation": true,
-              "notes": "Only supports <code>MediaStream</code> objects (see <a href='https://crbug.com/506273'>bug 506273</a>)."
-            },
+            "chrome": [
+              {
+                "version_added": "108",
+                "partial_implementation": true,
+                "notes": "Support added for <code>MediaSource</code> objects transferred from dedicated workers via <code>MediaSource.handle</code> (see <a href='https://crbug.com/878133'>bug 878133</a>)."
```
wolenetz, line 2878: Probably best to be precise here: ...Support added for <code>MediaSource**Handle**</code> objects transferred from dedicated workers where they were obtained from <code>MediaSource.handle</code>...

api/MediaSource.json
```diff
@@ -103,38 +103,38 @@
...
-            "safari_ios": "mirror",
-            "samsunginternet_android": "mirror",
-            "webview_android": "mirror"
-          },
-          "status": {
-            "experimental": true,
-            "standard_track": true,
-            "deprecated": false
+            "status": {
+              "experimental": true,
```
wolenetz, line 134: I noticed that MSE-in-Workers in MDN in these changes is being marked "experimental". It is shipping on-by-default in Chrome, and is implemented to follow the corresponding changes in the current MSE spec working draft. What does "experimental" mean here, given this information?

api/SourceBuffer.json
```diff
@@ -1723,7 +1678,14 @@
             "description": "Available in workers",
             "support": {
               "chrome": {
-                "version_added": "108"
+                "version_added": "108",
```
wolenetz, line 1681: Is there a way to note that both SourceBuffer.audioTracks and .videoTracks in a *dedicated worker* MediaSource's SourceBuffer is *not* expected to succeed even if the experimental flag is provided (there is further spec work and implementation work required to support those track lists from a worker context, and currently the implementation - if experimental flag is enabled and the app invokes the audioTracks() or videoTracks() getter on a *worker* SourceBuffer, then Chrome crashes the renderer process to avoid potentially worse behavior. This is known work in both spec and impl (MSE spec, HTML spec, Chrome impl), partially captured by this MSE spec issue: https://github.com/w3c/media-source/issues/280

wolenetz (commented):
Looks pretty good. A few nits/questions. Thanks again for working on this!

api/HTMLMediaElement.json
chrisdavidmills: Good call, updated!

api/MediaSource.json
chrisdavidmills: This is the definition that MDN uses for experimental: https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Experimental_deprecated_obsolete#experimental. In the time I've been contributing to MDN, we've gone back and forth on this a lot, and it still isn't perfect, but I think it's OK. MSE-in-workers is marked as experimental because it is still only available in one rendering engine, so the advice to devs would be to think carefully when considering using it. If you want to support Fx and WebKit browsers, then you might need to think about a fallback, alternative, or gracefully degrading experience. If you only need Chromium support, then you are OK. This certainly isn't the same as "don't use it".

api/SourceBuffer.json
chrisdavidmills: I've added a note to `audioTracks`/`videoTracks` to hopefully cover this; let me know what you think. I wasn't sure that the spec issue you linked to really explained what the issue was, so I decided not to link to it. IS there a better link available, or do you think I should link here, as that is where explanation will be added?

api/MediaSource.json
wolenetz: Thanks for the explanation. https://developer.mozilla.org/en-US/docs/MDN/Writing_guidelines/Experimental_deprecated_obsolete#experimental could probably use clarification around the meaning of "It is implemented and enabled by default in less than two modern major browsers" -- I understand that it is being applied as if s/major browsers/major distinct underlying rendering engines in browsers/, but on the surface it is confusing.

api/SourceBuffer.json
wolenetz: Good point - I've reactivated the most closely related Chromium issue to this piece of the spec. Please link to https://bugs.chromium.org/p/chromium/issues/detail?id=487288#c24 (it also links transitively to the MSE spec issue 280 that I mentioned, above).

wolenetz (commented):
Looks good to me - please add the link to https://bugs.chromium.org/p/chromium/issues/detail?id=487288#c24 as I just commented about a few minutes ago.

api/MediaSource.json
chrisdavidmills: I've filed this: https://github.com/mdn/content/issues/22391. Please add to it if you have any other thoughts on the subject. I'll try to fix this at some point next week, if no one else grabs it.

api/SourceBuffer.json
chrisdavidmills: Perfect, that is a useful explanation. Link added.

queengooborg (approved):
Thanks for following up on this, LGTM!

PR #18233, opened 2022-11-17
Firefox 106 is retired

ramiy:
Firefox 107 is the current version. Firefox 106 was retired.

This PR fixes #18221


queengooborg (approved):
Good catch, thanks!

PR #18250, opened 2022-11-20
Add Permissions-Policy

chrisdavidmills:
#### Summary

The `Feature-Policy` header changed to `Permissions-Policy` some time ago. This PR adds `Permission-Policy` BCD, keeping the existing `Feature-Policy` data but classifying it as support under an alternative name.

This is part of the work described at https://github.com/mdn/content/issues/22347

See my research document at https://docs.google.com/document/d/14xHANU3n43PfJR_dCpEwVCiVYv12BAfGqOICNvd79R4/edit#heading=h.mahbkxhoku5t for a list of the `Permissions-Policy` directives I am intending to implement on MDN as part of this work. This gives us parity with what was previously documented under `Feature-Policy`, but it adds a few more that are known to be supported, and it removes mention of some that were obviously being worked on at some point, but never seem to have gotten any traction.

Note — I have not yet deleted the `Feature-Policy.json` file, so that the existing docs don't break if the new BCD is merged and imported sooner than the MDN content is updated. We can always delete that later on.

New information was largely gotten from https://github.com/w3c/webappsec-permissions-policy/blob/main/features.md.

### DIFF
--- a/http/headers/Permissions-Policy.json
+++ b/http/headers/Permissions-Policy.json
@@ -6,9 +6,15 @@
           "mdn_url": "https://developer.mozilla.org/docs/Web/HTTP/Headers/Permissions-Policy",
           "spec_url": "https://w3c.github.io/webappsec-permissions-policy/#permissions-policy-http-header-field",
           "support": {
-            "chrome": {
-              "version_added": "60"
-            },
+            "chrome": [
+              {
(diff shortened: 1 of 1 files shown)


queengooborg (approved):
LGTM, thank you!

PR #18326, opened 2022-11-27
Mark the u element as depreciated

MrBrain295:
#### Summary
The `u` element is depreciated (see https://developer.mozilla.org/en-US/docs/Web/HTML/Element/u#usage_notes).

### DIFF
--- a/html/elements/u.json
+++ b/html/elements/u.json
@@ -36,7 +36,7 @@
           "status": {
             "experimental": false,
             "standard_track": true,
-            "deprecated": false
+            "deprecated": true
           }
         }
       }


queengooborg:
Hey @MrBrain295, thank you for digging into this, however reading the linked text, I don't believe we should mark this feature as deprecated.  The linked text reads:

> Along with other pure styling elements, the original HTML Underline (<u>) element was deprecated in HTML 4; however, <u> was restored in HTML 5 with a new, semantic, meaning: to mark text as having some form of non-textual annotation applied.

In other words, HTML 4 deprecated the feature, but then HTML 5 un-deprecated it.  Therefore, it is _not_ deprecated in the latest spec.

PR #18351, opened 2022-12-01
All RegExp legacy features are standard and deprecated

Josh-Cena:
#### Summary

https://github.com/mdn/browser-compat-data/pull/10273 added a link to a stage-3 proposal, which is fine, because this feature is de-facto implemented by all browsers anyway. However:

1. The edit isn't propagated to all legacy features.
2. The feature isn't marked as deprecated, causing it to appear as if it's fully standard.


Elchi3 (approved):
Thanks! 👍

PR #18352, opened 2022-12-01
Fix missing Firefox version number for MIDIInputMap.@@iterator

saschanaz:
#### Summary

#18278 somehow missed this. https://mdn-bcd-collector.gooborg.com/tests/api/MIDIInputMap/@@iterator says it's supported in Firefox, so not sure if it was a tooling bug or something.


queengooborg:
Looks like you've got a failing test, due to Firefox Android being set to just `false` rather than `{version_added: false}`.  Could you fix the errors?  Additionally, could we remove the flag data since it's now irrelevant?

PR #18369, opened 2022-12-05
add missing data for hyphenate-limit-chars

chrisdavidmills:
#### Summary

As part of my work on the [CSS `hyphenate-limit-chars` property](https://docs.google.com/document/d/1aLdSuLdZEMjdnMpeyBRdwwZyLr1JMbQ39Nk0ZWumLBU/edit#) (added to Chrome in version 109), I checked out the browser-compat-data. I found that the Chrome data was already there, but the [CSS Tricks article](https://css-tricks.com/almanac/properties/h/hyphenate-limit-chars/) mentions support in IE 10 and pre-Chromium versions of Edge with an `-ms-` prefix. I took this opportunity to add that data.

### DIFF
--- a/css/properties/hyphenate-limit-chars.json
+++ b/css/properties/hyphenate-limit-chars.json
@@ -9,13 +9,22 @@
               "version_added": "109"
             },
             "chrome_android": "mirror",
-            "edge": "mirror",
+            "edge": [
+              {
+                "version_added": "109"
+              },
+              {
+                "version_added": "≤79",
+                "prefix": "-ms-"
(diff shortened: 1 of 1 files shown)


queengooborg:
A while back, we actually ended up freezing IE data so we're not doing intentional updates to it.  I think we should just skip this PR, WDYT?

chrisdavidmills:
I'm more than happy to not worry about IE/Early Edge data ;-)

Thanks for letting me know; I'll close this PR.

PR #18421, opened 2022-12-13
Update Edge to support 103 Early Hints

tunetheweb:
#### Summary

Edge appears to support 103 Early Hints from WPT, I presume from same version as Chrome (it's not noted at all in the Edge release notes)

https://wpt.fyi/results/loading/early-hints?label=experimental&label=master&aligned

You can also see it working from this test, where the fonts load early:
https://www.webpagetest.org/result/221213_BiDcDR_DN3/1/details/

### DIFF
--- a/http/status.json
+++ b/http/status.json
@@ -46,9 +46,7 @@
               "version_added": "103"
             },
             "chrome_android": "mirror",
-            "edge": {
-              "version_added": false
-            },
+            "edge": "mirror",
             "firefox": {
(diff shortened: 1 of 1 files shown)


queengooborg (approved):
Hey there, thank you for your PR and welcome to BCD!  This makes total sense to me and was most likely a mistake when the version number was added to Chrome.  I don't see any reason Edge would disable support for a core feature like this.  👍

PR #18426, opened 2022-12-14
22118

dletorey:
#### Summary

I have updated the spec_url links to point to draft.csswg.org/

N/A

#### Related issues

N/A


dipikabh:
I don't think we should add `font-variant-emoji` to the font-variant BCD table. Other constituent longhand `font-variant-*` properties are also not part of the table.

dletorey:
>I don't think we should add font-variant-emoji to the font-variant BCD table. Other constituent longhand font-variant-* properties are also not part of the table. 
@dipikabh I agree and have removed this

css/properties/font-variant-emoji.json
```diff
@@ -4,7 +4,7 @@
       "font-variant-emoji": {
         "__compat": {
           "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/font-variant-emoji",
-          "spec_url": "https://w3c.github.io/csswg-drafts/css-fonts/#font-variant-emoji-prop",
+          "spec_url": "https://w3c.github.io/csswg-drafts/css-fonts-4/#font-variant-emoji-prop",
```
sideshowbarker, line 7:
```suggestion
          "spec_url": "https://w3c.github.io/csswg-drafts/css-fonts/#font-variant-emoji-prop",
```

css/properties/font-variant.json
```diff
@@ -4,7 +4,7 @@
       "font-variant": {
         "__compat": {
           "mdn_url": "https://developer.mozilla.org/docs/Web/CSS/font-variant",
-          "spec_url": "https://w3c.github.io/csswg-drafts/css-fonts/#font-variant-prop",
+          "spec_url": "https://w3c.github.io/csswg-drafts/css-fonts-4/#font-variant-prop",
```
sideshowbarker, line 7:
```suggestion
          "spec_url": "https://w3c.github.io/csswg-drafts/css-fonts/#font-variant-prop",
```

sideshowbarker (changes requested):
BCD policy is to use the un-leveled spec URLs rather than the leveled ones — unless the anchor doesn’t exist in the un-leveled spec but does exist in some leveled version.

In this case, the `#font-variant-emoji-prop` anchor does exist in the un-leveled spec. So we should use the un-leveled spec URL unless there’s some other specific reason for the using the leveled spec. Is there such a reason in this case?

queengooborg:
It looks like there are three different things this PR is doing.  May we split them into separate PRs for easier review and merging, please?

PR #16020, opened 2022-04-26
Add data for CSS HDR media queries

queengooborg:
This PR adds the `dynamic-range` and `video-dynamic-range` CSS media query features to BCD to go along with https://github.com/mdn/content/issues/14405.

Data comes from the following:
https://bugzilla.mozilla.org/show_bug.cgi?id=1751217
https://chromestatus.com/feature/5680926106320896
https://webkit.org/blog/10247/new-webkit-features-in-safari-13-1/#:~:text=for%20more%20information.-,More%20CSS%20Additions,-There%20are%20a

PR #18462, opened 2022-12-20
Remove support for video-dynamic-range

NiedziolkaMichal:
#### Summary
I have run [test](https://codepen.io/NiedziolkaMichal/pen/YzjPGRr) on Chrome, Firefox, and Safari. Only firefox has this feature working.

#### Related issues

[Chromium issue](https://bugs.chromium.org/p/chromium/issues/detail?id=1224711&q=video-dynamic-range&can=2)
[Webkit issue](https://bugs.webkit.org/show_bug.cgi?id=249678)

### DIFF
--- a/css/at-rules/media.json
+++ b/css/at-rules/media.json
@@ -1475,7 +1475,7 @@
             "spec_url": "https://w3c.github.io/csswg-drafts/mediaqueries-5/#video-dynamic-range",
             "support": {
               "chrome": {
-                "version_added": "98"
+                "version_added": false
               },
               "chrome_android": "mirror",
               "edge": "mirror",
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
This data originally came from https://github.com/mdn/browser-compat-data/pull/16020 but the sources https://chromestatus.com/feature/5680926106320896 and https://webkit.org/blog/10247/new-webkit-features-in-safari-13-1/#:~:text=for%20more%20information.-,More%20CSS%20Additions,-There%20are%20a only talk about `dynamic-range` and not `video-dynamic-range`. Thus I think your PR is correct and thanks for mentioning/filing the browser bugs.

PR #18515, opened 2023-01-03
FF109: CSS system-color add Mark, MarkText, ButtonBorder

hamishwillee:
FF109 adds support for new [`<system-color>`](https://developer.mozilla.org/en-US/docs/Web/CSS/system-color#mark) values `Mark`, `MarkText`, `ButtonBorder` in https://bugzilla.mozilla.org/show_bug.cgi?id=1638052

There is no support in Chrome according to https://bugs.chromium.org/p/chromium/issues/detail?id=1260079.

There is no evident of support in Safari et al.

Other docs work for this can be tracked in https://github.com/mdn/content/issues/23334

### DIFF
--- a/css/types/color.json
+++ b/css/types/color.json
@@ -1027,6 +1027,44 @@
               "standard_track": true,
               "deprecated": false
             }
+          },
+          "mark_marktext_buttonborder": {
+            "__compat": {
+              "description": "Mark, MarkText, ButtonBorder",
+              "spec_url": [
+                "https://w3c.github.io/csswg-drafts/css-color/#valdef-system-color-mark",
+                "https://w3c.github.io/csswg-drafts/css-color/#valdef-system-color-marktext",
+                "https://w3c.github.io/csswg-drafts/css-color/#valdef-system-color-buttonborder"
+              ],
+              "support": {
+                "chrome": {
+                  "version_added": false
+                },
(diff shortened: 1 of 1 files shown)


css/types/color.json
```diff
@@ -1027,6 +1027,44 @@
               "standard_track": true,
               "deprecated": false
             }
+          },
+          "mark_marktext_buttonborder": {
```
hamishwillee, line 1031: @queengooborg I did this as a subfeature combining all three system colors. I can split them up if you want. The reason I did them combined is that they were added to the spec at the same time and both FF and Chrome had them in grouped bugs. So I would expect them to always deliver together. 

IF that proves incorrect we can split them later. Or if you want we can split now?

hamishwillee:
@queengooborg FYI I'm away for a couple of weeks after today. So if you need changes and have no response, that's why.

css/types/color.json
queengooborg: I think that grouping is fine if these three colors are expected to always be implemented at the same time, and we can separate them later if we need to!

queengooborg (approved):
Sorry for the wait, I hadn't seen this PR -- this is LGTM!

hamishwillee:
Thanks very much! All good.

PR #18537, opened 2023-01-04
Add data for FileSystemHandle.remove()

chrisdavidmills:
#### Summary

This PR adds data for the new `FileSystemHandle.remove()` method. See my [research document](https://docs.google.com/document/d/1ymUjOxI8yL2krSrR3Mn3ZEAGhfQ1S592qQzPV89VSyg/edit#) for more information.

### DIFF
--- a/api/FileSystemHandle.json
+++ b/api/FileSystemHandle.json
@@ -189,6 +189,41 @@
           }
         }
       },
+      "remove": {
+        "__compat": {
+          "mdn_url": "https://developer.mozilla.org/docs/Web/API/FileSystemHandle/remove",
+          "support": {
+            "chrome": {
+              "version_added": "110"
+            },
+            "chrome_android": "mirror",
+            "edge": "mirror",
+            "firefox": {
+              "version_added": false
+            },
+            "firefox_android": "mirror",
+            "ie": {
+              "version_added": false
+            },
+            "oculus": "mirror",
+            "opera": "mirror",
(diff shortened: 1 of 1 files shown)


api/FileSystemHandle.json
```diff
@@ -189,6 +189,41 @@
...
+            },
+            "safari_ios": "mirror",
+            "samsunginternet_android": "mirror",
+            "webview_android": {
+              "version_added": false
+            }
+          },
+          "status": {
+            "experimental": true,
+            "standard_track": false,
```
a-sully, line 222:
```suggestion
            "experimental": false,
            "standard_track": true,
```
I'm not too familiar with the format here, but it seems like these bits should be flipped? This is being standardized in [whatwg/fs](https://github.com/whatwg/fs/pull/9)

queengooborg: "Experimental" should be left as `true` because only one browser engine, Chromium, supports it.  Also, we don't set "Standard Track" to `true` until it has landed in a spec's standards track, so the pull request must be merged first.

queengooborg (approved):
LGTM, thanks!

api/FileSystemHandle.json
chrisdavidmills: Cheers for providing the explanation @queengooborg 

@a-sully I'm happy to update this as soon as it lands in the spec.

PR #18511, opened 2023-01-02
Add Chrome version numbers for tabs webextensions API

gregorypappas:
#### Summary

Fix up the version_added field for a couple properties

#### Test results and supporting details

https://web.archive.org/web/20150609090643/https://developer.chrome.com/extensions/tabs#method-getZoomSettings
https://web.archive.org/web/20150609090643/https://developer.chrome.com/extensions/tabs#type-ZoomSettings

#### Related issues

https://bugzilla.mozilla.org/show_bug.cgi?id=1772166

PR #18538, opened 2023-01-05
defaultZoomFactor fix version update

rebloor:
#### Summary

The issue with defaultZoomFactor not showing the correct value was fixed in Firefox 110, as per [Bug 1772166](https://bugzilla.mozilla.org/show_bug.cgi?id=1772166)


rebloor:
Necessary change completed in [#18511](https://github.com/mdn/browser-compat-data/pull/18511)

PR #18540, opened 2023-01-05
Update Chromium versions for CSPViolationReportBody API

queengooborg:
This PR updates and corrects version values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `CSPViolationReportBody` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.gooborg.com) project (v7.1.2).

Tests Used: https://mdn-bcd-collector.gooborg.com/tests/api/CSPViolationReportBody

Note: this was originally set to "74" in #10925 based upon commit history.  However, while Chrome seems to have IDL for this feature, it's not exposed.

### DIFF
--- a/api/CSPViolationReportBody.json
+++ b/api/CSPViolationReportBody.json
@@ -6,7 +6,7 @@
         "spec_url": "https://w3c.github.io/webappsec-csp/#cspviolationreportbody",
         "support": {
           "chrome": {
-            "version_added": "74"
+            "version_added": false
           },
           "chrome_android": "mirror",
           "edge": "mirror",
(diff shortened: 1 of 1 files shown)


queengooborg:
After researching this further, I found that Chrome actually does support this, but it's not directly exposed like it should be.  I created mdn/content#23433 to add an example on how to obtain an instance of this interface, and I pushed https://github.com/GooborgStudios/mdn-bcd-collector/commit/c0161af1bf72d5220d3c50335382caec6d732a2c to disable this interface in the collector for now (until I can figure out how to implement a test in the collector).

PR #18546, opened 2023-01-05
Remove api.Element.aria[Col/Row]IndexText from BCD

queengooborg:
This PR removes the irrelevant `ariaColIndexText` and `ariaRowIndexText` members of the `Element` API as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-features). The lack of current support has been confirmed by the [mdn-bcd-collector](https://mdn-bcd-collector.gooborg.com) project (v7.1.2), even if the current BCD suggests support.

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -945,40 +945,6 @@
           }
         }
       },
-      "ariaColIndexText": {
-        "__compat": {
-          "mdn_url": "https://developer.mozilla.org/docs/Web/API/Element/ariaColIndexText",
-          "spec_url": "https://w3c.github.io/aria/#dom-ariamixin-ariacolindextext",
-          "support": {
-            "chrome": {
-              "version_added": "81"
-            },
-            "chrome_android": "mirror",
-            "edge": "mirror",
-            "firefox": {
-              "version_added": false
-            },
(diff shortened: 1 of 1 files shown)


Elchi3 (commented):
hey hey I added them in https://github.com/mdn/browser-compat-data/pull/17642 and you approved! :D Why is the collector thinking they don't exist?

queengooborg:
I probably should've done more research first, but as far as I can tell, there isn't actually any support for these.  Using the [example for `ariaColIndexText`](https://developer.mozilla.org/en-US/docs/Web/API/Element/ariaColIndexText#examples) in Chrome 108, `ariaColIndexText` is `undefined` when it should be `"Aria Role column"`.

May I ask where Chrome 81 came from in your PR?

Elchi3 (approved):
hm I can't find evidence, I guess I was just fixing MDN pages with no compat tables and thought they were supported. I think it is a bit sad to loose these pages, but that's our policy. Approving but not merging since MDN content updates are needed.

sideshowbarker:
MDN changes merged https://github.com/mdn/content/pull/23489

PR #18552, opened 2023-01-05
Remove InterventionReportBody API from BCD

queengooborg:
This PR removes the irrelevant `InterventionReportBody` API as per the corresponding [data guidelines](https://github.com/mdn/browser-compat-data/blob/main/docs/data-guidelines.md#removal-of-irrelevant-features). The lack of current support has been confirmed by the [mdn-bcd-collector](https://mdn-bcd-collector.gooborg.com) project (v7.1.2), even if the current BCD suggests support.


Elchi3 (commented):
There is https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/core/frame/intervention_report_body.idl;l=9?q=InterventionReportBody&ss=chromium&start=11

Also https://chromestatus.com/feature/5544632075157504 says it's shipping in version 70.

queengooborg:
I'm not sure how to programmatically get support for this interface due to the nature of the interface, but I think this PR is wrong, so I'm going to close it.

PR #18557, opened 2023-01-05
releases(108): Fx 108 adds support for 'effective-directive' and 'status-code' props

bsmth:
CSP violation reports sent by Firefox include `effective-directive` and `status-code` properties in the payload.

The page referring to the API doesn't exist -> https://developer.mozilla.org/docs/Web/API/CSPViolationReportBody

Pointing to https://developer.mozilla.org/docs/Web/HTTP/CSP#violation_report_syntax instead

__Bugs:__
- https://bugzilla.mozilla.org/show_bug.cgi?id=1192684

__Parent issue:__
- [ ] https://github.com/mdn/content/issues/22907

__See also:__
- [x] New API pages added in https://github.com/mdn/content/pull/23433


api/CSPViolationReportBody.json
```diff
@@ -2,7 +2,7 @@
   "api": {
     "CSPViolationReportBody": {
       "__compat": {
-        "mdn_url": "https://developer.mozilla.org/docs/Web/API/CSPViolationReportBody",
+        "mdn_url": "https://developer.mozilla.org/docs/Web/HTTP/CSP#violation_report_syntax",
```
queengooborg, line 5: We shouldn't be changing this URL, and instead create a new page that talks about the API itself.

api/CSPViolationReportBody.json
```diff
@@ -11,7 +11,7 @@
           "chrome_android": "mirror",
           "edge": "mirror",
           "firefox": {
-            "version_added": false
+            "version_added": "63"
```
queengooborg, line 14: How did you get this version number?  Did you use test code, or a bug?

queengooborg (changes requested):

api/CSPViolationReportBody.json
bsmth: OK, can do

api/CSPViolationReportBody.json
bsmth: Based on 

* https://developer.mozilla.org/en-US/docs/Mozilla/Firefox/Releases/63#apis
* https://bugzilla.mozilla.org/show_bug.cgi?id=1472661

queengooborg: Bug 1472661 seems to refer to the `SecurityPolicyViolationEvent` API rather than this one (`CSPViolationReportBody`).  I don't see any IDL for this interface in Firefox either.

I think that the file you actually want to modify is `http/headers/Content-Security-Policy.json`, since that's specific to the HTTP header?

bsmth: I was also thinking that, but there are two parts to this topic:

* HTTP headers which set policies (`Content-Security-Policy: default-src 'self'; report-uri http://mycoolurl.com/reports`)
* Violation reports -> JSON payloads the browser will send to a configurable URL if a policy in the page headers are violated:

```json
{
  "csp-report": {
    "document-uri": "http://mycoolurl.com/somepage.html",
    "referrer": "",
    "blocked-uri": "http://example.com/css/style.css",
    "violated-directive": "style-src",
    "original-policy": "default-src 'self'; report-uri http://mycoolurl.com/reports"
  }
}
```

bsmth:
@queengooborg - I think we can merge, what do you think? 👍🏻

queengooborg (approved):

saschanaz:
Are you sure that Firefox has `CSPViolationReportBody`? It does expose `ReportBody` but not the CSP one.

PR #18561, opened 2023-01-06
Permissions - reorder permission strings together

hamishwillee:
This prefixes rather than suffixes permission keys with `permission_` so that they will be grouped. Otherwise you end up with the case we have now where the permissions are interleaved with methods - see https://developer.mozilla.org/en-US/docs/Web/API/Permissions_API#browser_compatibility

This is first part of work on https://github.com/mdn/content/issues/13003. We should also add spec links for each of these permissions, as they are defined in their own specs, but will leave that as a follow up.

### DIFF
--- a/api/Permissions.json
+++ b/api/Permissions.json
@@ -36,7 +36,7 @@
           "deprecated": false
         }
       },
-      "accelerometer_permission": {
+      "permission_accelerometer": {
         "__compat": {
           "description": "<code>accelerometer</code> permission",
           "support": {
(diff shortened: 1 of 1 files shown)


hamishwillee:
@marcoscaceres If you had to create a spec link for these, I presume we'd be best linking to the precise spec they are defined in rather than the permission registry?

queengooborg (approved):
This looks so much better and follows standard BCD naming conventions, thanks!

PR #18566, opened 2023-01-07
Adding `nth-col` and `nth-last-col` data

yarusome:
#### Summary

Adds missing data for `:nth-col()` and `:nth-last-col()` pseudo-classes.

These two selectors are in an identical situation to that of the [column combinator (`||`)](https://developer.mozilla.org/docs/Web/CSS/Column_combinator), i.e. no implementation by any browser yet but having a Firefox bug ([Bug 371323](https://bugzil.la/371323)).

#### Related issues

Fully fixes https://github.com/mdn/content/issues/23365.

### DIFF
--- a/css/selectors/nth-col.json
+++ b/css/selectors/nth-col.json
@@ -1,0 +1,39 @@
+{
+  "css": {
+    "selectors": {
+      "nth-col": {
+        "__compat": {
+          "description": "<code>:nth-col()</code>",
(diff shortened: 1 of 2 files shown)


yarusome:
I thought IE data has been frozen since a long time ago. Do I have to add placeholders for IE?

Elchi3 (commented):
Hey @yarusome, thanks for your PR. It looks like this feature isn't implemented anywhere yet, so it doesn't qualify for addition to BCD unfortunately.

Issue #18565, opened 2023-01-06
"ping-other-repos" workflow fails on Dependabot commits


queengooborg:
### What type of issue is this?

Infrastructure issue

### What is the issue?

The `ping-other-repos` workflow fails on all commits authored by Dependabot.  Looking closely, it seems that the token is not passed into the workflow for Dependabot commit.

Dependabot: https://github.com/mdn/browser-compat-data/actions/runs/3854711506/jobs/6568995481
Other Author: https://github.com/mdn/browser-compat-data/actions/runs/3854109447/jobs/6567746489

CC @sideshowbarker

### What behavior were you expecting?

It's odd to see the token not getting passed into commits authored by Dependabot.  I wonder why...

### What version(s) of BCD is the issue present in?

- [ ] The current BCD release
- [X] The current version of the `main` branch

sideshowbarker:
Since we don’t need the "ping-other-repos" workflow to run for dependabot merges anyway, I think we can deal with this by having the workflow check if the `github.actor` is `dependabot[bot]` — and if so, then doing an early return to bail out of the workflow and completely skip the actual workflow steps. So I opened #18568 with a change which does that.

PR #18568, opened 2023-01-07
[gh-actions] Ignore dependabot for ping-other-repos workflow

sideshowbarker:
Fixes https://github.com/mdn/browser-compat-data/issues/18565

### DIFF
--- a/.github/workflows/ping-other-repos.yml
+++ b/.github/workflows/ping-other-repos.yml
@@ -11,7 +11,7 @@ permissions:
 jobs:
   ping:
     runs-on: ubuntu-latest
-    if: ${{ github.actor == 'dependabot[bot]' }}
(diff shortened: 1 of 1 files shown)


.github/workflows/ping-other-repos.yml
```diff
@@ -11,6 +11,7 @@ permissions:
 jobs:
   ping:
     runs-on: ubuntu-latest
+    if: ${{ github.actor == 'dependabot[bot]' }}
```
queengooborg, line 14: I think this should be reversed:

```suggestion
    if: ${{ github.actor != 'dependabot[bot]' }}
```

sideshowbarker: oofs yeah, thanks for catching that

queengooborg (approved):
Thanks for the quick update, this is LGTM!

PR #9190, opened 2021-02-19
Add Edge versions for CSSTransition API

vinyldarkscratch:
This PR adds real values for Microsoft Edge for the `CSSTransition` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.appspot.com) project (v1.3.3).  Results are manually confirmed for accuracy.

Tests Used: https://mdn-bcd-collector.appspot.com/tests/api/CSSTransition

PR #18570, opened 2023-01-08
Update Chromium versions for CSSTransition API

queengooborg:
This PR updates and corrects version values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `CSSTransition` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.gooborg.com) project (v7.1.3).

Tests Used: https://mdn-bcd-collector.gooborg.com/tests/api/CSSTransition

---

This data was originally set to "78" in #6372.  This data had come from reading through the IDL and using commit history to determine when the feature was implemented (see https://github.com/mdn/browser-compat-data/pull/6372#discussion_r459623074).  Since Edge was set to "84" in #9190, it makes less sense to keep Chrome at "78".

### DIFF
--- a/api/CSSTransition.json
+++ b/api/CSSTransition.json
@@ -6,12 +6,10 @@
         "spec_url": "https://w3c.github.io/csswg-drafts/css-transitions-2/#the-CSSTransition-interface",
         "support": {
           "chrome": {
-            "version_added": "78"
-          },
-          "chrome_android": "mirror",
-          "edge": {
             "version_added": "84"
           },
+          "chrome_android": "mirror",
+          "edge": "mirror",
           "firefox": {
             "version_added": "75"
           },


Elchi3 (commented):
It's unclear to me where 84 comes from. 

I think 78 is still the best and proofed version number https://storage.googleapis.com/chromium-find-releases-static/d69.html#d692c061c5a4f07611af04d53a032c8ac4dde7c5

captainbrosset:
It's weird that Edge and Chrome do not agree currently. I did a bit of research and found the following:

The CSSTransition IDL was added to Chromium in 78: https://chromiumdash.appspot.com/commit/d692c061c5a4f07611af04d53a032c8ac4dde7c5
(tracking bug: https://bugs.chromium.org/p/chromium/issues/detail?id=981900)

But it seems to be part of the larger WebAnimations implementation in Chromium which got enabled in 84:
https://chromiumdash.appspot.com/commit/9cdcb106b0bcfee75cf3e94576fd48c055cc4bd6
(tracking bug: https://bugs.chromium.org/p/chromium/issues/detail?id=978551)

If I'm understanding this correctly, the correct versions here should be 84 for both chrome and edge.

Elchi3 (approved):
> But it seems to be part of the larger WebAnimations implementation in Chromium which got enabled in 84:
https://chromiumdash.appspot.com/commit/9cdcb106b0bcfee75cf3e94576fd48c055cc4bd6
(tracking bug: https://bugs.chromium.org/p/chromium/issues/detail?id=978551)

Thanks, this was the missing information here.

queengooborg:
Thanks @captainbrosset!  And thanks @Elchi3 for approving!

PR #18572, opened 2023-01-08
Update Chromium versions for CryptoKey API

queengooborg:
This PR updates and corrects version values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `CryptoKey` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.gooborg.com) project (v7.1.3).

Tests Used: https://mdn-bcd-collector.gooborg.com/tests/api/CryptoKey

### DIFF
--- a/api/CryptoKey.json
+++ b/api/CryptoKey.json
@@ -6,7 +6,7 @@
         "spec_url": "https://w3c.github.io/webcrypto/#cryptokey-interface",
         "support": {
           "chrome": {
-            "version_added": "37"
+            "version_added": "41"
           },
           "chrome_android": "mirror",
           "deno": {
@@ -47,7 +47,7 @@
           "spec_url": "https://w3c.github.io/webcrypto/#dom-cryptokey-algorithm",
           "support": {
             "chrome": {
-              "version_added": "37"
+              "version_added": "41"
             },
             "chrome_android": "mirror",
             "deno": {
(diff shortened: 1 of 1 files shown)


Elchi3 (commented):
So this seems to have been "Key" first and then "CrytoKey". I know how to map SVN commits to a Chrome version number https://src.chromium.org/viewvc/blink/trunk/Source/modules/crypto/CryptoKey.idl?pathrev=176587

queengooborg:
Was it just the interface name that changed, or did the way to get a key change as well?

Either way, this is pre-2020 data so I'm not going to bother with this for now.

foolip:
This far back (before Chrome 45) I use https://www.chromium.org/developers/calendar/ and map the date, but there's no point in mapping https://src.chromium.org/viewvc/blink/trunk/Source/modules/crypto/CryptoKey.idl?pathrev=176587 because:
- `NoInterfaceObject` is used, so the interface name doesn't matter
- https://mdn-bcd-collector.gooborg.com/tests/api/CryptoKey uses `crypto.subtle.generateKey()`, so the interface doesn't matter

The right move here is to ensure the data matches the earliest other API that involves `CryptoKey`, and that seems to be 37:
https://developer.mozilla.org/en-US/docs/Web/API/SubtleCrypto#browser_compatibility

But why do the tests say 41 then? Something is still wrong here. (But not important.)

PR #18591, opened 2023-01-09
Add CHIPs data additions

chrisdavidmills:
#### Summary

This PR adds data for features related to the new Cookies Having Independent Partitioned State (CHIPS) cookie partitioning. Specifically:

* The `Set-Cookie` header `Partitioned` attribute, used to specify that a cookie's storage should be partitioned
* New options/properties for `CookieStore` API features.

See my [research document](https://docs.google.com/document/d/1NCl3Ut9LePUhRP98jrL1HC-n42ZjDoyMKtoZcZMZicg/edit#heading=h.frl8u3cowbcx) for more information about this feature.

Note: I've not yet added a spec URL for this data point. I've requested that the spec be added to the browser-specs repo in here: https://github.com/w3c/browser-specs/issues/822


queengooborg:
I notice that the ChromeStatus feature that is linked states that Chrome 109 is an "estimated milestone" (and unfortunately, the distinction is VERY easy to miss).  In my experience, if ChromeStatus says "estimated milestone," the version numbers can't be trusted.  May we hold off until this feature has shipped or has a definite version number set?

chrisdavidmills:
> I notice that the ChromeStatus feature that is linked states that Chrome 109 is an "estimated milestone" (and unfortunately, the distinction is VERY easy to miss). In my experience, if ChromeStatus says "estimated milestone," the version numbers can't be trusted. May we hold off until this feature has shipped or has a definite version number set?

Doesn't it always say "estimated milestones" in that section of ChromeStatus? I'm not really sure how this particular one is any less reliable than the others I've submitted recently. This one is also "enabled by default in 109", and has available by default in 109 since it waas Chrome Canary. 109 is meant to be released today, so we shall know soon enough any way ;-)

queengooborg:
> Doesn't it always say "estimated milestones" in that section of ChromeStatus?

I was about to say "no" because I've seen ChromeStatus say "enabled by default in:" rather than "estimated milestone", but looking back at a page I knew said that (https://chromestatus.com/feature/6461631328419840 vs. http://web.archive.org/web/20210126051908/https://chromestatus.com/feature/6461631328419840), I see this was a general UI change.

Because of this being a UI change, though, I'm more concerned about using ChromeStatus as an information source.  First of all, why did they change the UI to always say "estimated"?  Shouldn't the Chrome team have definitive data for their own browser, rather than estimating when a feature is released/will release?  Additionally, as I've mentioned above, I've not had good luck with recent version data from ChromeStatus and have often times found the data to be incorrect.  I haven't dug into whether the problematic data is only for future features, or if it applies to all features, but it's definitely a problem.

foolip:
@jrobbins perhaps you can help explain how the chromestatus data is maintained?

queengooborg:
Chrome 109 has been released.  Testing with https://cr2.kungfoo.net/cookies/v2/ (found through https://developer.chrome.com/docs/privacy-sandbox/chips/#try-it-out -> https://www.chromium.org/updates/chips/#demo-chrome-103-or-later), I can confirm this is implemented.  👍

queengooborg (approved):

PR #18598, opened 2023-01-09
Update Samsung Internet versions for NavigatorUAData API

queengooborg:
This PR updates and corrects version values for Samsung Internet for the `NavigatorUAData` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.gooborg.com) project (v7.1.3).

Tests Used: https://mdn-bcd-collector.gooborg.com/tests/api/NavigatorUAData

### DIFF
--- a/api/NavigatorUAData.json
+++ b/api/NavigatorUAData.json
@@ -2,6 +2,7 @@
   "api": {
     "NavigatorUAData": {
       "__compat": {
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
Yeah, this seems like something where mirroring shouldn't be done as it is probably very user agent specific if and how this is implemented. 

Approving but I would love if you could add mdn_urls to this file before merging 🥇

PR #18601, opened 2023-01-09
Update Firefox versions for api.PerformanceEntry.toJSON

queengooborg:
This PR updates and corrects version values for Firefox and Firefox Android for the `toJSON` member of the `PerformanceEntry` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.gooborg.com) project (v7.1.3).

Tests Used: https://mdn-bcd-collector.gooborg.com/tests/api/PerformanceEntry/toJSON

### DIFF
--- a/api/PerformanceEntry.json
+++ b/api/PerformanceEntry.json
@@ -280,7 +280,7 @@
               "version_added": "16"
             },
             "firefox": {
-              "version_added": "35"
+              "version_added": "34"
             },
             "firefox_android": "mirror",
             "ie": {


Elchi3 (commented):
https://bugzilla.mozilla.org/show_bug.cgi?id=1047848 confirms 34 is correct but the parent feature says 35, so the linter fails. Maybe 35 is better given the whole API was introduced in https://bugzilla.mozilla.org/show_bug.cgi?id=1002855.

Why does the linter think it is 34?

PR #18619, opened 2023-01-11
Chrome 109 is released

queengooborg:
This PR marks Chrome 109 as the current version of Chrome.


Elchi3 (changes requested):
Webview android 108 needs to be marked "retired"

Elchi3 (approved):

PR #18620, opened 2023-01-11
Make mirroring linter an error vs. a warning

queengooborg:
This PR converts the "mirrorable data" warning into an error as suggested in https://github.com/mdn/browser-compat-data/pull/18617#issuecomment-1378381044.


Elchi3 (approved):
Thanks! 👍

PR #18624, opened 2023-01-11
Enable css lenght type lh for chrome

sp-schoen:
#### Summary

Enable the css length unit "lh" for chrome. Referenced bug is also close (Nov 2022) and tested in chromium 109. 

Bug was fixed https://crbug.com/937104
Implementation tested in chromium 109 (Did not work before in 108)

### DIFF
--- a/css/types/length.json
+++ b/css/types/length.json
@@ -267,8 +267,7 @@
             "description": "<code>lh</code> unit",
             "support": {
               "chrome": {
-                "version_added": false,
-                "notes": "See <a href='https://crbug.com/937104'>bug 937104</a>."
+                "version_added": "109"
               },
               "chrome_android": "mirror",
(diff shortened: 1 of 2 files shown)


Elchi3 (changes requested):
Thanks for your PR! Can you do the same for rlh? That seems to be fixed in Chromium 109 as well.

sp-schoen:
@Elchi3 I think this is in a separate issue https://bugs.chromium.org/p/chromium/issues/detail?id=1381037 . I also tested rlh and it is shown as invalid in the dev tools

Elchi3 (approved):
Oh you are absolutely right, see also the thread in https://groups.google.com/a/chromium.org/g/blink-dev/c/E3Q7qOCk7A4

PR #18628, opened 2023-01-11
Add data points for new FedCM features

chrisdavidmills:
#### Summary

This PR adds data for the [FedCM API](https://fedidcg.github.io/FedCM/).

See my [research document](https://docs.google.com/document/d/1_vvGjUePmEHP7adoF8cTRh8e2M1_haa8H4Us74LEQDY/edit#) for more information on the exact changes required for this addition.


queengooborg:
I'm not finding any support for IdentityCredential.logoutRPs or IdentityProvider in Chrome 108-109.  We should remove them.

chrisdavidmills:
@queengooborg this is a bit of a weird one. Both features you mention exist in the spec and in the Chrome source code:

* https://source.chromium.org/chromium/chromium/src/+/refs/tags/110.0.5475.0:third_party/blink/renderer/modules/credentialmanagement/identity_provider.idl
* https://source.chromium.org/chromium/chromium/src/+/main:third_party/blink/renderer/modules/credentialmanagement/identity_credential.idl

But they don't appear to be exposed in the browser from tests I've done, so unless I'm missing something like a flag, they should probably be removed, yes.

I've asked the relevant team at Google for clarification

chrisdavidmills:
@queengooborg I got a definitive answer from the Google folks. Those features aren't implemented yet, so I've removed them from the PR.

queengooborg (approved):

PR #18636, opened 2023-01-11
Update Safari versions for WEBGL_compressed_texture_astc API

queengooborg:
This PR updates and corrects version values for Safari (Desktop and iOS/iPadOS) for the `WEBGL_compressed_texture_astc` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.gooborg.com) project (v7.1.3).

Tests Used: https://mdn-bcd-collector.gooborg.com/tests/api/WEBGL_compressed_texture_astc

_Check out the [collector's guide on how to review this PR](https://github.com/GooborgStudios/mdn-bcd-collector#reviewing-bcd-changes)._

PR #18637, opened 2023-01-11
Update Safari versions for WEBGL_compressed_texture_etc API

queengooborg:
This PR updates and corrects version values for Safari (Desktop and iOS/iPadOS) for the `WEBGL_compressed_texture_etc` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.gooborg.com) project (v7.1.3).

Tests Used: https://mdn-bcd-collector.gooborg.com/tests/api/WEBGL_compressed_texture_etc

### DIFF
--- a/api/WEBGL_compressed_texture_etc.json
+++ b/api/WEBGL_compressed_texture_etc.json
@@ -21,7 +21,7 @@
           "opera": "mirror",
           "opera_android": "mirror",
           "safari": {
-            "version_added": false
+            "version_added": "16.2"
           },
           "safari_ios": {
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
Same story as https://github.com/mdn/browser-compat-data/pull/18636
Seems like is is implemented for a while already per https://bugs.webkit.org/show_bug.cgi?id=197900 but maybe it only shows up now due to hardware configurations/drivers/libraries. Indicating support is probably correct, we can always change to an earlier version if needed.

PR #18642, opened 2023-01-12
Update Chromium versions for Worklet API

queengooborg:
This PR updates and corrects version values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `Worklet` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.gooborg.com) project (v7.1.3).

Tests Used: https://mdn-bcd-collector.gooborg.com/tests/api/Worklet

### DIFF
--- a/api/Worklet.json
+++ b/api/Worklet.json
@@ -6,7 +6,7 @@
         "spec_url": "https://html.spec.whatwg.org/multipage/worklets.html#worklets-worklet",
         "support": {
           "chrome": {
-            "version_added": "65"
+            "version_added": "66"
           },
           "chrome_android": "mirror",
           "edge": "mirror",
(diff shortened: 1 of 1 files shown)


Elchi3 (changes requested):
I think for this one the collector test has to change to use Paint worklets instead of audio worklets.

https://bugs.chromium.org/p/chromium/issues/detail?id=567358#c55 says that paint was there earlier and so that should be the baseline test for the Worklet class.

queengooborg:
Got it -- I updated the test in https://github.com/GooborgStudios/mdn-bcd-collector/commit/96a394f47a2726cc0480b7aca63764d3276ac1bb to use `PaintWorklet` as a fallback if `AudioWorklet` isn't available (only Chrome supports `PaintWorklet` so far.)

PR #18644, opened 2023-01-12
Update Chromium versions for api.Bluetooth.referringDevice

queengooborg:
This PR updates and corrects version values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `referringDevice` member of the `Bluetooth` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.gooborg.com) project (v7.1.3).

Tests Used: https://mdn-bcd-collector.gooborg.com/tests/api/Bluetooth/referringDevice

### DIFF
--- a/api/Bluetooth.json
+++ b/api/Bluetooth.json
@@ -193,11 +193,9 @@
           "spec_url": "https://webbluetoothcg.github.io/web-bluetooth/#dom-bluetooth-referringdevice",
           "support": {
             "chrome": {
-              "version_added": "56"
-            },
-            "chrome_android": {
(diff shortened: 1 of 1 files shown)


Elchi3 (commented):
I can't find referringDevice in the chromium source, so this seems correct. It becomes an all-false feature as a consequence. So, should we remove entirely?

queengooborg:
I wonder why the linter didn't throw an error that this was a feature without support...  But anyways, I opened up #18646 to remove this feature entirely.

PR #18665, opened 2023-01-16
webRequestFilterResponse Permission

rebloor:
### Description

Details of the `"webRequestfilterResponse"` and `"webRequestFilterResponse.serviceWorkerScript"` API permissions.

### Motivation

Address of the BCD requirements of [Bug 1809235](https://bugzilla.mozilla.org/show_bug.cgi?id=1809235).

### Related issues and pull requests

PR for related content changes: https://github.com/mdn/content/pull/23699

### DIFF
--- a/webextensions/manifest/optional_permissions.json
+++ b/webextensions/manifest/optional_permissions.json
@@ -719,6 +719,46 @@
               "safari_ios": "mirror"
             }
           }
+        },
+        "webRequestfilterResponse": {
+          "__compat": {
+            "description": "<code>webRequestfilterResponse</code>",
+            "support": {
+              "chrome": {
+                "version_added": false
+              },
+              "edge": "mirror",
+              "firefox": {
+                "version_added": "110"
+              },
+              "firefox_android": "mirror",
+              "opera": "mirror",
+              "safari": {
+                "version_added": false
+              },
+              "safari_ios": "mirror"
+            }
+          },
+          "serviceWorkerScript": {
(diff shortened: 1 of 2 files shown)


queengooborg (approved):

webextensions/manifest/optional_permissions.json
```diff
@@ -719,6 +719,46 @@
...
+              },
+              "firefox_android": "mirror",
+              "opera": "mirror",
+              "safari": {
+                "version_added": false
+              },
+              "safari_ios": "mirror"
+            }
+          },
+          "serviceWorkerScript": {
```
rpl, line 742: Technically `serviceWorkerScript` isn't a property of a `webRequestFilterReponse` object, but a separate permission string (one is `"webRequestFilterResponse" and the other one is `"webRequestFilterResponse.serviceWorkerScript"`.

There are at least two other pre-existing WebExtensions API permissions that are using a similar convention: `"downloads"` and `"downloads.open"` and those two seems to not be nested as I'm suggesting for these two new ones:

- https://github.com/mdn/browser-compat-data/blob/8c311cba4f58c34d2c6396e65362772af23f21d5/webextensions/manifest/permissions.json#L375-L397

@rebloor given that this PR seems to have been already merged, would you mind to open a separate issue and PR to track following up about that?

rebloor: Fixed
image

PR #18675, opened 2023-01-18
add entry for : css.at-rules.import.supports

romainmenke:
#### Summary

There is no indication on [the `@imports` docs](https://developer.mozilla.org/en-US/docs/Web/CSS/@import#browser_compatibility) that support conditions are not supported.
My intention is the have an entry in the browser support overview showing that `@import url(foo.css) supports(...)` is not supported by any browser at this time.

~~I was unsure how to express "not supported by any browser" in the JSON structure.
I've omitted the `support` field entirely.~~

No browser supports `supports()` in `@import`.
Relevant WPT tests : https://wpt.fyi/results/css/css-cascade/import-conditions.html?label=master&label=experimental&aligned&view=subtest&q=import

#### Related issues

Fixes https://github.com/mdn/browser-compat-data/issues/12404

### DIFF
--- a/css/at-rules/import.json
+++ b/css/at-rules/import.json
@@ -79,6 +79,29 @@
         "supports": {
           "__compat": {
             "description": "<code>supports([ <supports-condition> | <declaration> ])</code>",
+            "support": {
+              "chrome": {
+                "version_added": false
+              },
+              "chrome_android": "mirror",
+              "edge": "mirror",
+              "firefox": {
+                "version_added": false
+              },
+              "firefox_android": "mirror",
+              "ie": {
+                "version_added": false
+              },
+              "oculus": "mirror",
+              "opera": "mirror",
+              "opera_android": "mirror",
+              "safari": {
+                "version_added": false
+              },
+              "safari_ios": "mirror",
+              "samsunginternet_android": "mirror",
+              "webview_android": "mirror"
+            },
             "status": {
               "experimental": false,
               "standard_track": true,
--- a/css/at-rules/import.json
+++ b/css/at-rules/import.json
@@ -75,6 +75,16 @@
               "deprecated": false
             }
           }
+        },
+        "supports": {
+          "__compat": {
+            "description": "<code>supports([ <supports-condition> | <declaration> ])</code>",
+            "status": {
+              "experimental": false,
+              "standard_track": true,
+              "deprecated": false
+            }
+          }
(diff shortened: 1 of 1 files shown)


queengooborg:
Our policy is not to add features to BCD if no browser supports it yet, so we should hold off on this until a browser has implemented support.

romainmenke:
But the issue is that the documentation does include `supports()` in `@import`.
This is confusing because the feature is fully documented and it isn't clear that this doesn't work at all.

Alternatively a clear note could be added in text to indicate that the feature isn't implemented?

queengooborg:
Ah, if that's the case then we should remove documentation for `supports()` and only add it once a browser has implemented support.  Would you be down to open an issue in mdn/content for this?

romainmenke:
@rachelandrew Requested that the docs are not removed in the linked issue : https://github.com/mdn/browser-compat-data/issues/12404#issuecomment-899066899

> Would you be down to open an issue in mdn/content for this?

Yes!

queengooborg:
> @rachelandrew Requested that the docs are not removed in the linked issue : [#12404 (comment)](https://github.com/mdn/browser-compat-data/issues/12404#issuecomment-899066899)

I see, thanks for linking that!  I'll discuss this with her in person.

queengooborg:
> @rachelandrew Requested that the docs are not removed in the linked issue : [#12404 (comment)](https://github.com/mdn/browser-compat-data/issues/12404#issuecomment-899066899)

I see, thank you for linking this!  Looking through the comment, the response was to the issue author's question about whether to document features based upon the spec's status, rather than the browser support.  Additionally, we made the "no unsupported features in BCD" policy after this comment was written.  Considering the fact this feature has been documented since 2021 and still has not gotten support, I doubt it ever will at this point, which is all the more reason to remove its documentation.

rachelandrew:
The thing is this is in a spec that's in CR (therefore suggesting you don't think it will ever be supported is odd), so completely missing it out of the docs means that anyone looking into this is going to find it in a spec, but it be completely missing from MDN. It's one thing to not document an entire spec that has no implementation, but this missing out of one piece of a spec, especially a well progressed spec, is problematic in my opinion.

queengooborg:
This had just been removed from MDN a little bit ago!

romainmenke:
Thank you @queengooborg

PR #18680, opened 2023-01-19
Fixed to webRequestFilterResponse.serviceWorkerScript

rebloor:
#### Summary

Fixes incorrect presentation of webRequestFilterResponse.serviceWorkerScript made in #18665

### DIFF
--- a/webextensions/manifest/optional_permissions.json
+++ b/webextensions/manifest/optional_permissions.json
@@ -738,25 +738,25 @@
               },
               "safari_ios": "mirror"
             }
-          },
-          "serviceWorkerScript": {
-            "__compat": {
-              "description": "<code>webRequestFilterResponse.serviceWorkerScript</code>",
-              "support": {
-                "chrome": {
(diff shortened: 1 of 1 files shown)


queengooborg:
Rather than prefixing the name of the feature, it should actually be indented one level and made as a sub feature, since we already represent `webRequestFilterResponse`.

queengooborg:
Actually, I see that's how it already is now, so the feature is already good and its placement does not need to be changed.

rebloor:
@queengooborg however, this is not the way we present information about dot permissions. For example, see https://github.com/mdn/browser-compat-data/blob/11c77b5ccffa8519513d45110c5a24f180326f7a/webextensions/manifest/permissions.json#L397

queengooborg:
Ah, I see what's going on.  These aren't properties, but string values that can be used.  We definitely should update the format to better represent this.

queengooborg (approved):

PR #18689, opened 2023-01-20
Fix representation of permissions in webextensions.manifest.optional_permissions

queengooborg:
This PR fixes the representation of the permissions within `webextensions.manifest.optional_permissions` to make it more clear that they are not properties.  This was inspired by #18680, where the PR had resolved the `webRequestFilterResponse.serviceWorkerScript` from being a `serviceWorkerScript` feature nested under `webRequestFilterResponse`, because it is not a `serviceWorkerScript` property of `webRequestFilterResponse`, but rather a separate string value of `webRequestFilterResponse.serviceWorkerScript`.  This PR attempts to resolve this confusion by suffixing all of the features with `_permission`.

### DIFF
--- a/webextensions/manifest/optional_permissions.json
+++ b/webextensions/manifest/optional_permissions.json
@@ -24,9 +24,9 @@
             }
           }
         },
-        "activeTab": {
+        "activeTab_permission": {
           "__compat": {
-            "description": "<code>activeTab</code>",
+            "description": "<code>activeTab</code> permission",
             "support": {
               "chrome": {
                 "version_added": false
@@ -46,9 +46,9 @@
             }
           }
         },
-        "background": {
+        "background_permission": {
           "__compat": {
-            "description": "<code>background</code>",
+            "description": "<code>background</code> permission",
             "support": {
               "chrome": {
                 "version_added": true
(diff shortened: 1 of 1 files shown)


rebloor (changes requested):
I would like @rpl and @Rob--W to comment on this. Also, if we are going to make this change, it also needs to be applied to the permissions.json file

Rob--W:
Luca and I discussed this briefly; We're not keen on this approach because the key would not match the permission name any more.

`permissions` and `optional_permissions` are already arrays that cannot contain anything else than strings by convention. Firefox only accepts strings, but there are examples of long-deprecated Chrome Apps (not extensions!) to accept objects ([`chrome.fileSystem`](https://developer.chrome.com/docs/extensions/reference/fileSystem/) API for example - [example of manifest.json file](https://source.chromium.org/chromium/chromium/src/+/main:chrome/test/data/extensions/api_test/file_system_provider/unresponsive_extension/manifest.json;l=9-13;drc=b197a62f22383255558f2a417ee96967f53e9ac7)). In any case, there is no ambiguity on the meaning of keys under `optional_permissions` (and `permissions`).

There are some permissions separated by a dot, but the dot does not convey any special meaning. At best the dot can be interpreted as suggesting a relation between the permission name and the namespace/permission before the dot. The current solution, of using `downloads_open` + "description": "<code>downloads.open</code>"` to represent permissions with dots looks sufficient for all intents and purposes.

queengooborg:
I ran into a similar situation with an API recently, and suggested a structure similar to what's currently in BCD for this feature.  As such, I'm going to close this.

PR #18696, opened 2023-01-20
Mark 'lh' and 'rlh' units as standard compliant

Th3S4mur41:
#### Summary

'lh' and 'rlh' are part of the CSS Values and Units Module Level 4 draft just like small, large, and dynamic viewport units.

They should therefor also be marked as following the standard track
Both lengths were added as [supported in Chromium 109 in a previous PR](https://github.com/mdn/browser-compat-data/pull/18624)

https://www.w3.org/TR/css-values-4/#lh
https://www.w3.org/TR/css-values-4/#rlh

### DIFF
--- a/css/types/length.json
+++ b/css/types/length.json
@@ -265,6 +265,7 @@
         "lh": {
           "__compat": {
             "description": "<code>lh</code> unit",
+            "spec_url": "https://www.w3.org/TR/css-values-4/#lh",
             "support": {
               "chrome": {
                 "version_added": "109"
@@ -290,7 +291,7 @@
               "webview_android": "mirror"
             },
             "status": {
-              "experimental": false,
+              "experimental": true,
               "standard_track": true,
               "deprecated": false
             }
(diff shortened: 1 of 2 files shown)


css/types/length.json
```diff
@@ -290,8 +290,8 @@
               "webview_android": "mirror"
             },
             "status": {
-              "experimental": true,
-              "standard_track": false,
+              "experimental": false,
```
queengooborg, line 293: We do not mark "experimental" based upon spec status, but rather based upon browser support.

```suggestion
              "experimental": true,
```

css/types/length.json
```diff
@@ -368,8 +368,8 @@
...
```
queengooborg, line 371:
```suggestion
              "experimental": true,
```

queengooborg (changes requested):
Along with the below changes, could you also add those URLs as `spec_url` properties?

Th3S4mur41:
Spec URL added 😊

queengooborg (approved):
Thanks, looking good!

PR #18738, opened 2023-01-25
icons.json: mirrored Opera and Edge to Chrome

shadow-identity:
#### Summary

Since SVG icons are [not supported in Chromium](https://bugs.chromium.org/p/chromium/issues/detail?id=29683) they are not supported in derivatives like Edge and Opera.
I was asked to update these browsers data to "mirror" in change request to corresponding MDN page update: https://github.com/mdn/content/pull/22845

Unfortunatelly I did not tested my changes.

#### Related issues

It is related to https://github.com/mdn/content/pull/22845

### DIFF
--- a/webextensions/manifest/icons.json
+++ b/webextensions/manifest/icons.json
@@ -9,16 +9,12 @@
               "version_added": true,
               "notes": "Chrome does not support SVG format for icons. It is recommended to use PNG images."
             },
-            "edge": {
-              "version_added": "14"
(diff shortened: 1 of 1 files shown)


shadow-identity:
I'm not sure if we can add a link to the [bug](https://bugs.chromium.org/p/chromium/issues/detail?id=29683) into the "notes" section, but if we can, it would be useful.

PR #18762, opened 2023-01-28
Add missing api.HTMLElement.virtualKeyboardPolicy feature

queengooborg:
This PR is a part of a project to add missing interfaces and interface features to BCD that are from an active spec (including WICG specs) and is supported in at least one browser.  This particular PR adds the missing `virtualKeyboardPolicy` member of the HTMLElement API, populating the results using data from the [mdn-bcd-collector](https://mdn-bcd-collector.gooborg.com) project (v7.1.5).

Tests Used: https://mdn-bcd-collector.gooborg.com/tests/api/HTMLElement/virtualKeyboardPolicy

### DIFF
--- a/api/HTMLElement.json
+++ b/api/HTMLElement.json
@@ -2022,6 +2022,39 @@
             "deprecated": false
           }
         }
+      },
+      "virtualKeyboardPolicy": {
+        "__compat": {
+          "spec_url": "https://w3c.github.io/virtual-keyboard/#dom-elementcontenteditable-virtualkeyboardpolicy",
+          "support": {
+            "chrome": {
+              "version_added": "94"
+            },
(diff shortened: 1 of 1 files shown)


Elchi3 (commented):
LGTM but the same entry should be present in html/ here: https://github.com/mdn/browser-compat-data/blob/main/html/global_attributes.json for https://developer.mozilla.org/en-US/docs/Web/HTML/Global_attributes/virtualkeyboardpolicy to render compat and spec tables.

Someone should probably also create an MDN page at https://developer.mozilla.org/en-US/docs/Web/API/HTMLELement/virtualKeyboardPolicy 

cc @captainbrosset

Elchi3 (approved):

PR #18768, opened 2023-01-28
Combine SVGMatrix and WebKitCSSMatrix into DOMMatrix

### LINKED ISSUE #6726: Does WebKitCSSMatrix deserve a separate support statement?
Currently `WebKitCSSMatrix` exists as a separate interface while it's currently merely an alias of `DOMMatrix`. I wonder it could be replaced by something similar with `alternative_name` in DOMMatrix.

queengooborg:
This PR takes WebKitCSSMatrix and SVGMatrix and combines them into DOMMatrix.  Both of these are defined as [`LegacyWindowAlias`es](https://drafts.fxtf.org/geometry-1/#dommatrix) in the IDL, and should not have separate files.  (The subfeatures will probably need to be updated; a future collector version will help fix this.)

Fixes #6726.


api/DOMMatrix.json
```diff
@@ -40,16 +146,18 @@
           "spec_url": "https://drafts.fxtf.org/geometry/#dom-dommatrix-dommatrix",
           "support": {
             "chrome": {
-              "version_added": "61"
+              "version_added": "2"
```
foolip, line 149: I don't think we should change the data for the constructor entry. Adding aliases that go back would be OK, and look just like the parent feature, but this data I would interpret as `new DOMMatrix()` being available since Chrome 2, which it isn't. The data isn't always shown with the parent feature after all:
https://developer.mozilla.org/docs/Web/API/DOMMatrix/DOMMatrix

foolip (changes requested):

foolip:
Also, the twists and turns that finally led up to `DOMMatrix` are kind of hilarious. Possibly a record in slowest and weirdest API rollout ever?

api/DOMMatrix.json
```diff
@@ -5,27 +5,133 @@
...
+            {
+              "version_added": "5",
+              "alternative_name": "SVGMatrix"
+            },
+            {
+              "version_added": "4",
+              "alternative_name": "WebKitCSSMatrix"
+            }
+          ],
+          "safari_ios": [
```
foolip, line 108: Can we mirror here, or what's the difference if we do?

api/DOMMatrix.json
```diff
@@ -5,27 +5,133 @@
...
+              "alternative_name": "SVGMatrix"
+            },
+            {
+              "version_added": "3",
+              "alternative_name": "WebKitCSSMatrix"
+            }
+          ],
           "samsunginternet_android": "mirror",
-          "webview_android": "mirror"
+          "webview_android": [
```
foolip, line 122: Mirror?

foolip (approved):
Looks like we could mirror more, but approving so you can land after updating, if the change is to just mirror.

api/DOMMatrix.json
queengooborg: There are some slight differences regarding Safari to Safari iOS mapping around there, so I'm skeptical about mirroring.  (If the mirroring output was the same as what's currently shown, the linter would freak out and say we should use `"mirror"`!)

foolip: Hmm, OK. Thanks for checking!

PR #18771, opened 2023-01-29
Correct data for AudioContext.createMediaStream* and MediaStreamAudioDestinationNode

queengooborg:
This PR corrects the data for both `AudioContext.createMediaStream*` methods, as well as `MediaStreamAudioDestinationNode`, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.gooborg.com) project (v7.1.5).

### DIFF
--- a/api/AudioContext.json
+++ b/api/AudioContext.json
@@ -387,7 +387,7 @@
             "opera": "mirror",
             "opera_android": "mirror",
             "safari": {
-              "version_added": "6"
+              "version_added": "11"
             },
             "safari_ios": "mirror",
             "samsunginternet_android": "mirror",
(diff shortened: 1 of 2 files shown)


queengooborg:
It's not within the scope of the 2020+ project, but the collector has been consistently making this change since v6, so I wanted to get this change in, or figure out how the test might be wrong.

foolip (approved):
This seems right, these bits were implemented long before Safari 11 but were behind the MEDIA_STREAM ifdef:
https://github.com/WebKit/WebKit/commit/29e4c6bf7a344fca1ecc2bdc2971ae789d15951d#diff-a77a250e10ae90ab910941cccf7e0ed0c5c856e2e72e01216935e1300aa86243R26

And MediaStream did ship in Safari 11.

PR #18772, opened 2023-01-29
Update Chromium versions for AudioParam API

queengooborg:
This PR updates and corrects version values for Chromium (Chrome, Opera, Samsung Internet, WebView Android) for the `AudioParam` API, based upon results from the [mdn-bcd-collector](https://mdn-bcd-collector.gooborg.com) project (v7.1.5).

Tests Used: https://mdn-bcd-collector.gooborg.com/tests/api/AudioParam

---

Note: although the `AudioParam` API was introduced all the way back in [late 2010](https://source.chromium.org/chromium/chromium/src/+/1e8047393acec1d927b8aeb56fbbf7ebf5d84aa2:third_party/WebKit/WebCore/webaudio/AudioParam.idl;dlc=368be040da67709e2023171654ed89436a351085), this was never usable until `GainNode` and similar was introduced.

### DIFF
--- a/api/AudioParam.json
+++ b/api/AudioParam.json
@@ -190,7 +190,7 @@
               "version_added": "24"
             },
             "chrome_android": {
-              "version_added": "18",
+              "version_added": "25",
               "partial_implementation": true,
               "notes": "This sets the target volume at the specified time, but it doesn’t ramp to it, causing this function to behave like <code>setValueAtTime()</code>."
             },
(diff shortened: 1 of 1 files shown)


foolip (commented):
This matches https://developer.mozilla.org/en-US/docs/Web/API/GainNode/gain#browser_compatibility but that's because the test uses `gainNode.gain` as the instance.

But there are other APIs that use `AudioParam` which go back to Chrome 14:
https://developer.mozilla.org/en-US/docs/Web/API/BiquadFilterNode#browser_compatibility

So I think the problem here is with the test. Tweaking it to try all different kinds of `AudioParam` APIs doesn't seem worth it to fix this data however. Maybe just override for this one?

queengooborg:
Ah, whoops, I didn't think that `BiquadFilterNode` had any `AudioParam` instances -- fixed the tests!

PR #18776, opened 2023-01-29
Add data for `Element.animate()` `options.timeline`

ninevra:
#### Summary

Adds support data for the `timeline` property of the `options` parameter of `Element.animate()`.

Firefox does not support this option. Chromium does, likely since version 85. Safari likely does, since version 16.

Manually tested in Firefox 111 and Chromium 109. A simple test case is:

```js
console.assert(
  document.body.animate([], { timeline: new DocumentTimeline() }).timeline !== document.timeline
);
```

Chromium support was added in [d393163](https://chromium.googlesource.com/chromium/src/+/d3931637492464d7781d2699c3c549c8f845aec6), which released in [Chromium 85](https://storage.googleapis.com/chromium-find-releases-static/d39.html#d3931637492464d7781d2699c3c549c8f845aec6).

Safari support was added in https://github.com/WebKit/WebKit/commit/1a2775d4de27be9e08d6c54a5d7ae03abc315168, included in WebKit since 614.1.6, which is Safari 16.

I don't have a setup to test Safari 16 or any additional browsers at the moment; the `"mirror"` entries are pure speculation, since the linter disallows `null`. I apologize if that makes this PR not useful.

### DIFF
--- a/api/Element.json
+++ b/api/Element.json
@@ -445,6 +445,40 @@
               "deprecated": false
             }
           }
+        },
+        "options_timeline_parameter": {
+          "__compat": {
+            "description": "<code>options.timeline</code> parameter",
+            "spec_url": "https://w3c.github.io/csswg-drafts/web-animations-1/#dom-keyframeanimationoptions-timeline",
+            "support": {
+              "chrome": {
+                "version_added": "85"
+              },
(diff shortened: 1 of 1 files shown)


Elchi3 (approved):
LGTM 👍

PR #18780, opened 2023-01-29
Fix Firefox compatibility data for the Link HTTP header

urkerab:
Support was added to Gecko 0.9 (predates Firefox) by [bug 3248](https://bugzilla.mozilla.org/show_bug.cgi?id=3248).

### DIFF
--- a/http/headers/Link.json
+++ b/http/headers/Link.json
@@ -12,7 +12,7 @@
             "chrome_android": "mirror",
             "edge": "mirror",
             "firefox": {
-              "version_added": true
+              "version_added": "1"
             },
             "firefox_android": "mirror",
             "ie": {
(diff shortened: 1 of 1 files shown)


http/headers/Link.json
```diff
@@ -12,7 +12,7 @@
             "chrome_android": "mirror",
             "edge": "mirror",
             "firefox": {
-              "version_added": null
+              "version_added": true
```
queengooborg, line 15:
```suggestion
              "version_added": "1"
```

queengooborg (changes requested):
Hey there, thank you for your PR and welcome to BCD!  Since we know what version number this was introduced in, we should use that version number instead of `true` (which means "supported, but we don't know when"):

http/headers/Link.json
urkerab: I almost missed this because you left a review comment rather than a suggestion comment...

queengooborg: I did both, though I meant to add this to the review.  🙃

queengooborg (approved):
Thank you, LGTM!

PR #18785, opened 2023-01-30
Revert "Add CHIPs data additions"

queengooborg:
Reverts mdn/browser-compat-data#18591.

When looking deeper into the code for https://cr2.kungfoo.net/cookies/v2/ in an attempt to reverse engineer test code for the collector, I had realized that what I thought was functional code was actually just names set by the server, and did not actually reflect partitioning.  The "Opt-out of partitioned cookies" button does nothing in the example.  Additionally, in my test code, I find that the `partitioned` option is completely ignored:

```js
var accessed = false;
cookieStore.set({name: 'foo', value: 'bar', partitioned: {
  get: function () {
    accessed = true;
    return true;
  }
}}).then(() => {
  console.log(accessed);
});
// > false
```

Additionally, the [ChromeStatus feature's](https://chromestatus.com/feature/5179189105786880) linked bug is [set to "Assigned (Open)", not "Fixed (Closed)"](https://crbug.com/1225444).

CC @chrisdavidmills

### DIFF
--- a/api/CookieStore.json
+++ b/api/CookieStore.json
@@ -105,39 +105,6 @@
             "standard_track": true,
             "deprecated": false
           }
-        },
-        "partitioned_option": {
-          "__compat": {
-            "description": "<code>partitioned</code> option",
-            "support": {
-              "chrome": {
-                "version_added": "109"
-              },
-              "chrome_android": "mirror",
-              "edge": "mirror",
-              "firefox": {
-                "version_added": false
-              },
-              "firefox_android": "mirror",
-              "ie": {
-                "version_added": false
-              },
-              "oculus": "mirror",
-              "opera": "mirror",
-              "opera_android": "mirror",
-              "safari": {
-                "version_added": false
-              },
-              "safari_ios": "mirror",
-              "samsunginternet_android": "mirror",
-              "webview_android": "mirror"
-            },
-            "status": {
-              "experimental": true,
-              "standard_track": false,
-              "deprecated": false
-            }
-          }
         }
       },
       "get": {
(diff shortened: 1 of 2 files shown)


chrisdavidmills:
Ah. Thanks for the extra testing @queengooborg. I got told to add these because they were supported, but I wonder if anything has changed since then.

Ah, looking at the ChromeStatus it now says shipping in 110. Did you test in Chrome beta (110), or just the current release (109)? 110 is supposed to become current today. If it's working in that version, it would be a shame to revert this all at this point...we could just set the version to 110 instead?

Also, did you find evidence of non-support for the header too? Or just the CookieStore stuff? You've reverted everything.

queengooborg:
I'm not seeing any support in the current builds of Chrome 110 or 111 either, both with the above code and using the test website.  Maybe we could get some feedback from the Chromium bug owners?

queengooborg:
(FYI, I still don't see any support in the stable release of Chrome 111!)

Elchi3 (commented):
I'm not really sure how to review this one. Can we get a subject matter expert or someone from the Chromium side to comment on whether this is implemented or not?

chrisdavidmills:
Hi @DCtheTall! When you get a second, can you please help us get to the bottom of this? We are struggling to find a definitive answer as to the current state of support for the CHIPs additions I documented at the start of the year (`partitioned` in the Cookie Store API and the `Set-Cookie` header).

Thanks in advance.

queengooborg:
It appears that the ChromeStatus page has been updated and now says that the feature is shipping in Chrome 114!  https://chromestatus.com/feature/5179189105786880

chrisdavidmills:
> It appears that the ChromeStatus page has been updated and now says that the feature is shipping in Chrome 114! https://chromestatus.com/feature/5179189105786880

Crikey, it really has slipped. I guess we could test it in the latest Canary. Do you have some test cases to use?

queengooborg:
> Do you have some test cases to use?

Probably the best test I currently have is the code written up above, which tests to see if the `partitioned` option is accessed -- this can result in false positives, but it at least gives us a general idea about support!

chrisdavidmills:
An update — Google officially reported this as released in Chrome 114 beta, so I've [updated the BCD](https://github.com/mdn/browser-compat-data/pull/19754) to reflect that.

queengooborg:
Closing in favor of https://github.com/mdn/browser-compat-data/pull/19754, then!

PR #18795, opened 2023-01-31
Remove redundant worker_support from each MediaSource subfeature

queengooborg:
This PR removes the redundant `worker_support` subfeatures applied to all of the MediaSource subfeatures, and adds a top-level `worker_support` feature in their place.


Elchi3 (approved):
👍

PR #18814, opened 2023-02-03
last baseline data

rachelandrew:
The `last baseline` alignment value is now implemented in all three engines. The data was a bit unwieldy as it represented the fact that for a time the value was parsed but didn't do anything in Chrome and Safari. I have just removed that information, but let me know if you want it back.

I've tested this: https://codepen.io/rachelandrew/pen/MWBKPrR/1ec83980993b3d01d160f9438fbfca4e 

- Safari release notes: https://developer.apple.com/documentation/safari-release-notes/safari-16_2-release-notes
- Chrome Status: https://chromestatus.com/feature/5093352798683136

### DIFF
--- a/css/properties/align-items.json
+++ b/css/properties/align-items.json
@@ -197,9 +197,7 @@
                   "version_added": "108"
                 },
                 "chrome_android": "mirror",
-                "edge": {
-                  "version_added": "mirror"
-                },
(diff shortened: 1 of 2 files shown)


Elchi3 (approved):
Thanks! 👍

PR #18833, opened 2023-02-07
RTCRtpEncodingParameters.priority is standard

hamishwillee:
`RTCRtpEncodingParameters.priority` had this note against the property:

> The standard version of this property is <a href='https://developer.mozilla.org/docs/Web/API/RTCRtpSendParameters/priority'><code>RTCRtpSendParameters.priority</code></a>.

However if you look at the spec, `RTCRtpSendParameters.priority` does not exist. However this recent spec does include the priority in the object: https://www.w3.org/TR/webrtc-priority/#encoding-parameters

Further FF110 adds this priority and updates it to have a default value "low".

It is likely that there was some play in the spec at some point. I've removed the note but noted the version in which FF added the default value. 

Other docs work for this can be tracked in https://github.com/mdn/content/issues/23677

FYI @queengooborg


hamishwillee:
PS I know this is a dictionary update, and we don't do dictionaries any more. However I don't want to untangle this now - could be done as a post process.

api/RTCRtpEncodingParameters.json
```diff
@@ -239,7 +239,7 @@
           },
           "status": {
             "experimental": false,
-            "standard_track": false,
+            "standard_track": true,
             "deprecated": true
```
hamishwillee, line 243:
```suggestion
            "deprecated": false
```

api/RTCRtpEncodingParameters.json
```diff
@@ -239,8 +239,8 @@
           },
           "status": {
             "experimental": false,
-            "standard_track": false,
-            "deprecated": true
+            "standard_track": true,
+            "deprecated": false
```
hamishwillee, line 243: This is in the spec, and implemented everywhere. There is no evidence that it is deprecated.

api/RTCRtpEncodingParameters.json
```diff
@@ -209,19 +209,19 @@
...
             "chrome_android": "mirror",
             "edge": {
-              "version_added": "≤79",
-              "notes": "The standard version of this property is <a href='https://developer.mozilla.org/docs/Web/API/RTCRtpSendParameters/priority'><code>RTCRtpSendParameters.priority</code></a>."
+              "version_added": "≤79"
             },
             "firefox": {
               "version_added": "46",
-              "notes": "The standard version of this property is <a href='https://developer.mozilla.org/docs/Web/API/RTCRtpSendParameters/priority'><code>RTCRtpSendParameters.priority</code></a>."
+              "notes": "Version 110 sets to spec default priority to <code>low</code>."
```
teoli2003, line 224: What was the previous default priority? I guess 46-110 has a different default priority than the modern spec and this has been fixed?

hamishwillee: Firefox assigned no default priority at all prior to 110. I don't know what the spec had previously, which is why I just added this as a note to FF. My thinking is that unless we end up with data on other browsers then in "not too long" everyone will use this default. 

But if you think this warrants further investigation, let me know.

teoli2003: Looks fine to me. I had trouble understanding the sentence, so I tried to rewrite it:
```suggestion
              "notes": "From version 110 on the default priority is <code>low</code>."
```

hamishwillee: Thanks! Minor variation on your change done
```suggestion
              "notes": "In version 110 and later the default priority is <code>low</code>."
```

teoli2003 (approved):

teoli2003 (approved):
👍

